PR #112, opened 2014-03-11
added option to overwrite existing files (not adding #XX)

schriftgestalt:


behdad:
Dupe of https://github.com/behdad/fonttools/pull/208

PR #181, opened 2014-10-05
[subset] added --recalc-unicode-ranges option through TTFont argument

anthrotype:
Hi, 

I tried again to add a --recalc-unicode-ranges option to subset.py, this time via a new 'recalcUnicodeRanges' argument to the TTFont constructor (default is False).

With recalcUnicodeRanges=True, the OS/2 'ulUnicodeRange' values will be recalculated upon save/compile according to the Unicode code points specified in the font's cmap tables.

I put the ranges data and the corresponding updateUnicodeRanges function inside the OS/2 table module.

Please, let me know what you think.
Thank you.

Cosimo

### DIFF
--- a/Lib/fontTools/subset.py
+++ b/Lib/fontTools/subset.py
@@ -261,6 +261,12 @@ Glyph naming and encoding options:
       Drop the 3.0 symbol 'cmap'. [default]
 
 Other font-specific options:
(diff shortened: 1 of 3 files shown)


behdad:
Ok based on the discussion, lets continue with #191.

PR #191, opened 2014-10-14
[subset] added option to prune unused ulUnicodeRanges bits

anthrotype:
after @roozbehp suggestion, I rewrote this thing once again. 

The subset option "--recalc-unicode-ranges" (default is False) now only works in "negative": that is, it prunes (turn OFF) all the ulUnicodeRange bits which are no longer needed, as no intersection is found between the ranges and the code points that are left in the font after subsetting. It does *not* also turn the bits ON in case there's any intersection, like I did before.

This approach seems more sensible, as there isn't universal agreement on when those bits should be set on. In the specs the determination of "functional" unicode ranges is left to the font designer.

I kept the unicode ranges data from the specs inside the OS/2 table module, as it may also be recycled in other scripts.

Let me know what you think about it.
Thank you,

Cosimo

### DIFF
--- a/Lib/fontTools/subset.py
+++ b/Lib/fontTools/subset.py
@@ -266,9 +266,9 @@ Other font-specific options:
       ranges defined in the OpenType specification v1.6 are intersected with
       the Unicode code points specified in the font's 'cmap' tables: when
       no overlap is found, the bit will be switched off. However, it will
-      *not* be switched on if an intersection is found.
+      *not* be switched on if an intersection is found.  [default]
   --no-update-unicode-ranges
-      Don't change the 'OS/2 ulUnicodeRange' bits. [default]
+      Don't change the 'OS/2 ulUnicodeRange' bits.
   --recalc-bounds
       Recalculate font bounding boxes.
   --no-recalc-bounds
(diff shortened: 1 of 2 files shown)


anthrotype:
@behdad should this option be on by default then?
Currently it's off.

behdad:
Yes, lets have this always on in the subsetter.  We'll address the possibility to have it in ttLib later.

anthrotype:
Ok, but I'm off now. Could you fix it after you merge it?
Thanks.

twardoch:
It should be on by default. The bits are just a "cache" for quick lookups instead of parsing the entire cmap. That's how they're used by Windows. Even if the algorithm just "removes" the bits, it should be on by default. I am not aware of any situation where one of these bits should be on if no single Unicode from a given range is present in the cmap :)

anthrotype:
recalc-unicode-ranges is now on by default.
let me know if it requires any more changes.

anthrotype:
I'm closing this PR because it contains too many merge conflicts after the formatting changes introduced lately. I shall open a new PR branch from the tip of the current master.

behdad:
Just so you know, you can force-push reworked branches to github. Eg:

  git push origin -u +uniranges3_reworked:uniranges3

anthrotype:
I'll do that next time, thank you!

PR #206, opened 2015-01-03
Regard very low datetimes as unix timestamps

olivierberten:
These used to be converted to 0.


behdad:
Just wondering, do these also result in an underflow?  Why were they converted to 0 before?  I agree that it's useful though.  Merging.

olivierberten:
fonttools used to use time.mktime for this, which used to have problems on some systems with early dates. That problem never existed with gmtime though...

PR #207, opened 2015-01-06
Add comments with names of TrueType instruction to assembly

schriftgestalt:


behdad:
Thanks.  I cherry-picked one commit.  Here it is: https://github.com/behdad/fonttools/commit/5d66d59a1f3a25f31fd696bf6024da71251a977b

PR #208, opened 2015-01-06
add option to overwrite existing files

schriftgestalt:


schriftgestalt:
We discussed this before. You suggested a solution to use .tt{x,f} as extension. But when converting back and forth, this is confusing, I would not know what direction of conversion it would do.

grzegorzrolek:
Hi Georg, it was I who suggested the brace expansion. Brace expansion is actually dead simple: `.tt{x,f}` always expands in order of what’s within the braces, `.ttx .ttf`, so there really shouldn’t be any confusion what direction the conversion will take.

My argument agains the option is two-fold. First, because your use case, that is, dealing with long filenames, is actually a problem of the command line interpreter, not the command itself (a classic Unix example). Second, a file overwrite can be really destructive for users not accustomed with the command line, and it would be better for such users to type the filenames explicitly for both input and output just to be sure they really mean it. Users who know what they’re doing can as well use various shortcuts like the brace expansion.

That said, it’s Behdad who decides on such issues in the end, so feel free to lobby him otherwise.

behdad:
I merged this, but added it as -f.

PR #209, opened 2015-01-08
add some missing return values

schriftgestalt:
That was somehow missed by the last commit.


behdad:
Merged manually in adfe7f293a41da4ba0188e6a9688af9eaaf0726e.  Thanks.

PR #212, opened 2015-01-27
Handle the exception of IndexError when list item access is out of range

vitalyvolkov:
I am not strong with font tables but this commit is a try to fix small bug and catch case when accessing array item out of range


vitalyvolkov:
https://github.com/googlefonts/fontbakery/issues/501

vitalyvolkov:
I have removed more white spaces changes and leave only needed ones

PR #222, opened 2015-03-29
implement WOFF2 encoder/decoder

anthrotype:
Hi all,
sorry if this has taken so long to share. I thought it was better to submit at least an incomplete draft, than risking to duplicate efforts with someone else, who may be also working on the same thing.

I decided to make a separate woff2 module instead of trying to fit it inside sfnt module, since it was too difficult to conciliate the many differences between the SFNT/WOFF1 and WOFF2.

The current version can create WOFF2 fonts which are compatible with the OpenType Sanitiser that ships in current browsers. However, as the WOFF2 specs and reference implementation are constantly changing, we may need to modify some aspects once those changes are implemented.

The encoder still doesn't support WOFF2 collections. I decided to wait until @behdad implements his new TTC API.

The woff2 module requires the brotli Python extension from https://github.com/google/brotli.

Thank you.

Cosimo

### DIFF
--- a/Lib/fontTools/ttLib/woff2.py
+++ b/Lib/fontTools/ttLib/woff2.py
@@ -1,0 +1,899 @@
+from __future__ import print_function, division, absolute_import
+from fontTools.misc.py23 import *
+import sys
+
+try:
+	import brotli
+except ImportError:
+	print('ImportError: No module named brotli.\n'
+		  'The WOFF2 encoder requires the Brotli Python extension, available at:\n'
+		  'https://github.com/google/brotli', file=sys.stderr)
+	sys.exit(1)
+
+import array
+import struct
+from fontTools.misc import sstruct
+from fontTools.misc.arrayTools import calcIntBounds
+from fontTools.ttLib import (TTFont, TTLibError, getTableModule, getTableClass,
+	getSearchRange)
+from fontTools.ttLib.sfnt import (SFNTReader, SFNTWriter, DirectoryEntry,
+	WOFFFlavorData, sfntDirectoryFormat, sfntDirectorySize, SFNTDirectoryEntry,
+	sfntDirectoryEntrySize, calcChecksum)
+from fontTools.ttLib.tables import ttProgram
+
+
+def normaliseFont(ttFont):
+	""" The WOFF 2.0 conversion is guaranteed to be lossless in a bitwise sense
+	only for 'normalised' font files. Normalisation occurs before any transforms,
+	and involves:
+		- removing the DSIG table, since the encoding process can invalidate it;
+		- recalculating simple glyph bounding boxes so they don't need to be stored
+		  in the bboxStream, but are recalculated by the decoder;
+		- padding glyph offsets to multiple of 4 bytes;
+		- setting bit 11 of head 'flags' field to indicate that the font has
+		  undergone a 'lossless modifying transform'.
+	"""
+	if "DSIG" in ttFont:
+		del ttFont["DSIG"]
+
+	if ttFont.sfntVersion == '\x00\x01\x00\x00':
+		ttFont.recalcBBoxes = True
+		ttFont.padGlyphData = True
+		# don't be lazy so that glyph data is 'expanded' on decompile
+		ttFont.lazy = False
+		# force decompile glyf table to perform normalisation steps above
+		if not ttFont.isLoaded('glyf'):
+			ttFont['glyf']
+		# XXX Not clear if this also applies to CFF-flavoured fonts
+		ttFont['head'].flags |= 1 << 11
+
+
+class WOFF2Reader(SFNTReader):
+
+	flavor = "woff2"
+
+	def __init__(self, file):
+		self.file = file
+
+		signature = self.file.read(4)
+		if signature != b"wOF2":
+			raise TTLibError("Not a WOFF2 font (bad signature)")
+
+		self.file.seek(0)
+		self.DirectoryEntry = WOFF2DirectoryEntry
+		sstruct.unpack(woff2DirectoryFormat, self.file.read(woff2DirectorySize), self)
+
+		self.tables = {}
+		self.tableOrder = []
+		offset = 0
+		for i in range(self.numTables):
+			entry = self.DirectoryEntry()
+			entry.fromFile(self.file)
+			tag = Tag(entry.tag)
+			self.tables[tag] = entry
+			self.tableOrder.append(tag)
+			entry.offset = offset
+			offset += entry.length
+
+		totalUncompressedSize = offset
+		compressedData = self.file.read(self.totalCompressedSize)
+		decompressedData = brotli.decompress(compressedData)
+		if len(decompressedData) != totalUncompressedSize:
+			raise TTLibError(
+				'unexpected size for decompressed font data: expected %d, found %d'
+				% (totalUncompressedSize, len(decompressedData)))
+		self.transformBuffer = StringIO(decompressedData)
+
+		self.flavorData = WOFF2FlavorData(self)
+
+	def __getitem__(self, tag):
+		entry = self.tables[Tag(tag)]
+		rawData = entry.loadData(self.transformBuffer)
+		if tag not in woff2TransformedTableTags:
+			return rawData
+
+		if hasattr(entry, 'data'):
+			# table already reconstructed, return compiled data
+			return entry.data
+
+		if tag == 'glyf':
+			# reconstruct both glyf and loca
+			self.glyfTable = WOFF2GlyfTable()
+			data = self.glyfTable.reconstruct(rawData)
+		elif tag == 'loca':
+			assert len(rawData) == 0, "expected 0, received %d bytes" % len(rawData)
+			if not hasattr(self, 'glyfTable'):
+				# make sure glyf is loaded first
+				self['glyf']
+			data = self.glyfTable.getLocaData()
+
+		if len(data) != entry.origLength:
+			raise TTLibError(
+				"reconstructed '%s' table doesn't match original size: expected %d, found %d"
+				% (tag, entry.origLength, len(data)))
+		entry.data = data
+		return data
+
+
+class WOFF2Writer(SFNTWriter):
+
+	flavor = "woff2"
+
+	def __init__(self, file, numTables, sfntVersion="\000\001\000\000",
+				 flavorData=None):
+		self.file = file
+		self.numTables = numTables
+		self.sfntVersion = Tag(sfntVersion)
+		self.flavorData = flavorData
+
+		self.directoryFormat = woff2DirectoryFormat
+		self.directorySize = woff2DirectorySize
+		self.DirectoryEntry = WOFF2DirectoryEntry
+
+		self.signature = "wOF2"
+
+		self.nextTableOffset = 0
+		self.transformBuffer = StringIO()
+
+		self.tables = {}
+		self.tableOrder = []
+
+	def __setitem__(self, tag, data):
+		"""Associate new entry named 'tag' with raw table data."""
+		if tag in self.tables:
+			raise TTLibError("cannot rewrite '%s' table" % tag)
+
+		entry = self.DirectoryEntry()
+		entry.tag = Tag(tag)
+		if tag == 'head':
+			entry.checkSum = calcChecksum(data[:8] + b'\0\0\0\0' + data[12:])
+		else:
+			entry.checkSum = calcChecksum(data)
+		entry.flags = getKnownTagIndex(entry.tag)
+		entry.origLength = len(data)
+		# WOFF2 table data are written to disk only on close(), after all tags
+		# have been specified
+		entry.data = data
+
+		self.tables[tag] = entry
+		self.tableOrder.append(tag)
+
+	def close(self):
+		""" All tags must have been specified. Now write the table data and directory.
+		"""
+		# According to the specs, the WOFF2 table directory must reflect the 'physical
+		# order' in which the tables have been encoded. Moreover, the glyf and loca
+		# tables must be placed 'as a pair with loca table following the glyf table'.
+		"""\
+		if 'loca' in self.tableOrder and 'glyf' in self.tableOrder:
+			glyfIndex = self.tableOrder.index('glyf')
+			locaIndex = self.tableOrder.index('loca')
+			self.tableOrder.insert(glyfIndex+1, self.tableOrder.pop(locaIndex))
+		tables = [(tag, self.tables[tag]) for tag in self.tableOrder]
+		"""
+		# However, to pass the legacy OpenType Sanitiser currently included in browsers,
+		# we must sort the table directory and data alphabetically by tag.
+		# See:
+		# https://github.com/google/woff2/pull/3
+		# https://lists.w3.org/Archives/Public/public-webfonts-wg/2015Mar/0000.html
+		# TODO(user): change to match spec once browsers are on newer OTS
+		tables = sorted(self.tables.items())
+
+		if len(tables) != self.numTables:
+			raise TTLibError("wrong number of tables; expected %d, found %d" % (self.numTables, len(tables)))
+
+		# compute the original SFNT offsets for checksum adjustment calculation
+		offset = sfntDirectorySize + sfntDirectoryEntrySize * len(tables)
+		for tag, entry in tables:
+			entry.origOffset = offset
+			offset += (entry.origLength + 3) & ~3
+
+		# size of uncompressed font
+		self.totalSfntSize = offset
+
+		self.signature = b"wOF2"
+		self.reserved = 0
+
+		for tag, entry in tables:
+			if tag == "loca":
+				data = b""
+			elif tag == "glyf":
+				indexFormat, = struct.unpack(">H", self.tables['head'].data[50:52])
+				numGlyphs, = struct.unpack(">H", self.tables['maxp'].data[4:6])
+				glyfTable = WOFF2GlyfTable()
+				glyfTable.setLocaData(self.tables['loca'].data, indexFormat, numGlyphs)
+				data = glyfTable.transform(entry.data)
+			else:
+				data = entry.data
+			entry.offset = self.nextTableOffset
+			entry.saveData(self.transformBuffer, data)
+			self.nextTableOffset += entry.length
+
+		# update head's checkSumAdjustment
+		self.writeMasterChecksum()
+
+		# compress font data
+		self.transformBuffer.seek(0)
+		uncompressedData = self.transformBuffer.read()
+		compressedData = brotli.compress(uncompressedData, brotli.MODE_FONT)
+		self.totalCompressedSize = len(compressedData)
+
+		# start calculating total size of WOFF2 font
+		offset = woff2DirectorySize
+		for tag, entry in tables:
+			offset += len(entry.toString())
+		offset += self.totalCompressedSize
+		offset = (offset + 3) & ~3
+
+		# calculate offsets and lengths for any metadata and/or private data
+		compressedMetaData = privData = b""
+		flavorData = self.flavorData or WOFF2FlavorData()
+		if flavorData.majorVersion is not None and flavorData.minorVersion is not None:
+			self.majorVersion = flavorData.majorVersion
+			self.minorVersion = flavorData.minorVersion
+		else:
+			self.majorVersion, self.minorVersion = struct.unpack(
+				">HH", self.tables['head'].data[4:8])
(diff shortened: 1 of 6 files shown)


anthrotype:
For those who wishes to test the woff2 encoder on Windows or OSX and don't have the developer tools required to build the Brotli extension, I've set up a Github repository which uses Travis and Appveyor to build pre-compiled [Python wheels](pythonwheels.com) compatible with Python 2.7 and 3.4:

https://github.com/anthrotype/brotli-wheels/releases

To install a wheel, just download it and install it with pip.
For example:

    pip install Brotli-0.1-cp27-none-macosx_10_9_intel.whl

So far, the supported Python versions and architectures include:

- OS X 10.9 Python 2.7, "intel" architecture (system)
- OS X 10.9 Python 2.7, "x86_64" architecture (homebrew)
- OS X 10.9 Python 3.4, "x86_64" architecture (homebrew)

- Windows Python 2.7, "win32" architecture (32 bit)
- Windows Python 3.4, "win32" architecture (32 bit)
- Windows Python 2.7, "amd64" architecture (64 bit)
- Windows Python 3.4, "amd64" architecture (64 bit)

Cheers,

C.

behdad:
Thanks @anthrotype!  I haven't checked the full patchset yet.  My one feedback so far is, I like to avoid adding more arguments to TTFont constructor and instead switch to a more sophisticated configuration system.  Something like yml, but ideally without requiring PyYML.  Such that, for example, user would control the padding by setting options['glyf']['compile']['padding'] = 2.  WDYT?

behdad:
Very neat use of Travis and Appveyor BTW!  I wish I was more well-versed with them :D.

anthrotype:
hey @behdad, thanks for you comments.
I don't know what exactly you have in mind about this YML-based configuration system. I guess that would allow more versatility in configuration options. I shall wait to see more examples of that, and see how I could adapt it to the woff2 encoder. 

As for the glyph data padding, I understand your reluctance to add a new argument to TTFont's signature. The problem I'm faced with is the following -- /cc @khaledhosny @rsheeter.

In the latest [Editors draft](http://dev.w3.org/webfonts/WOFF2/spec/) of the WOFF2 spec they decided to remove the ambiguous concept of "nominal size" which required encoders, among other things, to pad glyph data to multiples of 4 bytes.
However, both the WOFF2 decoder in the [OpenType Sanitiser](https://github.com/khaledhosny/ots), and the [reference implementation](https://github.com/google/woff2), still rely on the glyf table's `origLength` for allocating memory during the glyf reconstruction process. This means that, if fontTools doesn't add padding to the sfnt glyf table and set the `origLength` to the actual table length, then when one tries to load a WOFF2 font in Chrome or Firefox, the process will fail since the decoder, on the one hand, will read `origLength` strictly as the reconstructed table's `dst_length`, and on the other, will add padding to the reconstructed glyf table, hence exceeding the allocated length.

All this means that, until decoders expect the `origLength` to represent the "nominal size" of sfnt glyf table, the loca offsets will have to be padded to 4 bytes. So for WOFF2, this is not like a user's option, but rather a de facto requirement which arises from other "user agents" implementations.

The reason I decided to add a public `padGlyphData` argument and attribute to TTFont, is that by doing so I can generate a "normalised" sfnt font (e.g. with padded glyphs, without DSIG, etc.), and then compare this and the output of the WOFF2 decompression, thus making sure the two are bitwise identical.

Alternatively, I could have written in the glyf table module:

    if ttFont.flavor == "woff2":
        # pad glyph data

but then I would not be able to pad glyph data for non WOFF2-fonts, for which flavor is `None`.

Of course, I'm happy to receive further suggestions from any of you.
Thanks again for taking the time to review the patch.

C.

anthrotype:
I made a quick gist to exemplify the roundtrip test I was referring to:
https://gist.github.com/anthrotype/7372e958a459547d7ea5

anthrotype:
OK, maybe an explicit "padGlyphData" attribute just to do a roundtrip testing was a bit too much -- and that test of mine was quite tautological...
So I decide to remove the TTFont argument. The glyf table module now simply checks if the font's flavor is "woff2" in order to decide whether to pad glyph data to 4 bytes or not.

If I want to test that the reconstructed glyf and loca tables are identical to the original ('normalised') ones, I can compare them directly, like in this gist:

https://gist.github.com/anthrotype/98dfa654be415a7d0cfc

There's no need to create a temporary sfnt font to do the comparison, because the only thing that actually changes are the tables that are subject to modifying transform; all the rest is guaranteed to be bitwise identical by the Brotli compressor.
 
C.

anthrotype:
The woff2 encoder was always discarding the bounding box information in the original font (`recalcBBoxes=True`), but this could create problems with fonts which have non-standard bounding boxes (cf. a similar issue in the reference implementation https://github.com/google/woff2/issues/12)

I've patched it to follow the [#conform-mustCalculateOmitBBoxValues](http://dev.w3.org/webfonts/WOFF2/spec/#conform-mustCalculateOmitBBoxValues) conformance requirement.

Cheers,

C.

anthrotype:
I just added some simple doctest strings to the helper functions (also discovered how tricky it is to make doctest play nicely with both python 2 and 3!).
I'm wondering if a full unittest may be required for the main reader and writer classes...

anthrotype:
make check was aborting because the brotli extension is not installed on the Travis virtual machine.
I moved the "import brotli" statements away from the main module namespace, nesting them inside the methods which actually use the extension.
but perhaps the setup.py needs to be modified so that it builds brotli extension from a submodule.
What do you think?

anthrotype:
and what about pypy? I haven't tested it yet, but I imagine the current brotli Python bindings will need to be adapted using CFFI or ctypes (or even rewritten in pure Python?).

khaledhosny:
I think you should just try to import brotli early on when encountering WOFF2 files and use its presence to decide whether WOFF2 can be supported or not. This way FontTools degrades gracefully when brotli is absent.

anthrotype:
@khaledhosny I tried to do what you've suggested in 262c8612678d04e005497e34b81f7fcc21ea0a9b.

Also, I tried to build the current brotli extension under PyPy on Win, Mac and Linux, and all seems good: it passes both the compatibility and roundtrip tests (apart from a minor issue with the Windows PyPy, already fixed with https://github.com/google/brotli/pull/71).

The question remains as to how to meet this new installation requirement for fontTools.
Should brotli be imported as a git submodule in fontTools' repository, and the setup.py modified so that it tries to builds the brotli extension upon install? 
Or should we leave WOFF2 support as optional for now, and just remind the users about where to download and how to build the extension (in the README.md maybe?)

adrientetar:
Should remain optional imo.

jzempel:
@anthrotype i'm using your [PR branch](https://github.com/anthrotype/fonttools/tree/woff2_12) (along with brotli installed on osx 10.10) to subset about 25 vendor fonts. i.e. `$ pyftsubset font.woff2 --unicodes="U+0370-03FF,U+1F00-1FFF"`. I get the following error on 50% of them:

```
fontTools.ttLib.TTLibError: reconstructed 'loca' table has wrong index format: expected 1, found 0
```

raised [here](https://github.com/anthrotype/fonttools/blob/woff2_12/Lib/fontTools/ttLib/woff2.py#L642). I don't have much font format expertise, but I'm happy to help however I can.

anthrotype:
hey @jzempel 
thanks for reporting this. It seems there is a mismatch between the reconstructed loca's indexFormat and the value which is encoded in the glyf table of the WOFF2 font that you are trying to read.

Could you please tell me the name of the font which cannot be decoded? And also whether the WOFF2 font was originally encoded through fontTools or through another WOFF2 encoder? 

By the way, in your example you are trying to subset the WOFF2 font directly. I would suggest you to subset the original sfnt font (TTF or OTF) and use the --flavor=woff2 option to compress the result, provided you have a copy of the original uncompressed font.

jzempel:
@anthrotype i'm subsetting a bunch of different faces of proxima nova. The WOFF2 files were encoded by the provider.

Great tip on subsetting TTF => WOFF2. That ends up fixing my issue for all the fonts in question.

anthrotype:
@jzempel I tried to fix the issue in 96ee22429e51ec683b7cc27b236d2406f293ac27
could you `git pull` and see if it works?
thanks,

C.

jzempel:
@anthrotype - thanks. That fixes all WOFF2 => WOFF2.subset fonts that I'm attempting.

adrientetar:
@behdad Is this ready to land?

behdad:
> @behdad Is this ready to land?

Yes, mostly.  I need to find a couple of hours to understand and land it.

rolfnl:
Yes, it's been working fine for me. Not sure if there are some additional tools to verify the generated files? Other then reading the file signature. Anyway, I had no problems so far, the generated files work as expected in the various browsers etc.

benkiel:
Is there any concern that the WOFF2 spec seems to be changing a bit — is this premature? Last edits to the working draft changed several things. I'm not following that mailing list, but perhaps better to wait until it hits CR to see if any more changes happen?

anthrotype:
Regarding @benkiel's legitimate concerns, I'd like to point out that I have not followed strictly the old Working Draft from 2014, but I always tried to keep up with the Editors Drafts, as well as the discussions on the public-webfont-wg list and of course Google's reference woff2 implementation.
Therefore, the new Working Draft published early this month didn't come as a surprise.

The only major thing that is still missing is the support for TrueType Collections, which I deliberately put on hold (I hear @behdad is also working on a new TTC API). Anyway, no browsers currently support these fonts.

Apart from TTC, there still a couple of things in my woff2 encoder which may change once other implementations catch up with the spec:

1) the padding of glyph data inside glyf table to multiples of 4 bytes is no longer required, after the notion of "nominal size" was dropped from the WOFF2 specification (nor is it forbidden). The problem is other decoders still expect a "normalised" glyf table with [4-byte aligned](https://github.com/google/woff2/issues/15) local offsets (or at least [2-byte aligned](https://github.com/khaledhosny/ots/issues/60) in the case of OTS). My fontTools branch can decode glyf tables both with and without padding; the encoder however must produce a 4-byte padded glyf tables to avoid its fonts being rejected by browsers.

2) the spec says it is the responsibility of WOFF2 decoder to reorder the table directory alphabetically as per OpenType spec. Unlike SFNT or WOFF1, in WOFF2 the table directory must follow the same physical order in which the tables were encoded, as there are no explicit table offsets. The order can be any one an encoder finds most convenient -- except for the requirement that loca follows glyf, since the former can only be reconstructed from the latter. The reference woff2 encoder, for simplicity, just sorts the table alphabetically (ie. "l" follows "g"). There is a problem, however, in the reference decoder, as well as in the OpenType Sanitiser that ships in current browsers (but not the latest HEAD from @khaledhosny), whereby the decoder fails to decompress a WOFF2 font whose table directory/data is not already sorted alphabetically (I tried to address it myself [here](https://github.com/google/woff2/pull/8)). So again, to avoid encoded fonts being rejected by browsers, I set the fontTools encoder to do this unnecessary table reordering.

Lastly, the WOFF2 fonts made with fontTools encoder would be 100% bitwise identical with those produced by the reference encoder, if it was not for a minor difference in the way fontTools and the reference implementation handle the [compression of repeating flags](https://github.com/behdad/fonttools/blob/master/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L611) inside glyph coordinates: when the current flag is the same as the last one, but the number of repetitions is only 1, fontTools prefers to simply append the second flag as is, whereas the reference implementation sets the "flagRepeat" bit in the preceding byte and encodes the number of repetitions (1). From my tests, it looks like repeating the flag value like fontTools does produces slightly better Brotli compression.

all best,

Cosimo

Nantonos:
@benkiel changes are few, and always evaluated with careful consideration of impact on existing running code and on fonts in the wild. There is a list at
http://www.w3.org/TR/WOFF2/#changes

Some are clarifications, some are additions made to support Font Collections (previously known as TrueType Collections). And yes, we found a bug in OT Sanitizer (now fixed, but it needs to propogate) which caused valid fonts to be discarded.

The idea time to point out errors and omissions is now, not waiting until CR. The latest draft was published specifically to get final feedback before we go into CR. Chrome, Chrome for Android, Opera, and Firefox nightly are using WOFF2 now.

benkiel:
@Nantonos @anthrotype Understood completely. I didn't define my question well above. I was more wondering if it would be better to leave this branch unmerged into FontTools until CR, so that it was clear to users that WOFF2 was still somewhat in flux, rather then merge it now and have it feel more 'official' than it is. 

This concern comes from the point of view of someone making self hosted web fonts for clients. Unlike a service that can rebuild WOFF2 files when the spec changes, convincing folks to update self hosted fonts when/if the spec changes is a greater cost of time/effort. I've learned, often the hard way, that fonts live long lives once sent to clients, so I tend to be conservative with things that have a high chance for changes.

I saw the issue with the OT Sanitizer and WOFF2, but had misread that it was caused by a spec change, not a bug. Admittedly, I've not been tracking the development of WOFF2 as close as WOFF, and had connected the OTS issue with the raft of changes I saw in the latest Editor's Draft of the spec. My apologies for this misconception, and I also think that my concerns with releasing WOFF2 fonts and the need of the Working Group to have tested implementations of WOFF2 are different issues that I didn't frame well in my initial question.

behdad:
> 1) the padding of glyph data inside glyf table to multiples of 4 bytes is no longer required, after the notion of "nominal size" was dropped from the WOFF2 specification (nor is it forbidden). The problem is other decoders still expect a "normalised" glyf table with 4-byte aligned local offsets (or at least 2-byte aligned in the case of OTS). My fontTools branch can decode glyf tables both with and without padding; the encoder however must produce a 4-byte padded glyf tables to avoid its fonts being rejected by browsers.

So is this all about coming up with a "nominal size" that is large enough?  Then can't we just add 3 bytes per glyph to make sure?  Or do we really need to encode with padding?  Not that it really matters.  Just trying to follow the latest status of the issue.

> 2) the spec says it is the responsibility of WOFF2 decoder to reorder the table directory alphabetically as per OpenType spec. Unlike SFNT or WOFF1, in WOFF2 the table directory must follow the same physical order in which the tables were encoded, as there are no explicit table offsets. The order can be any one an encoder finds most convenient -- except for the requirement that loca follows glyf, since the former can only be reconstructed from the latter. The reference woff2 encoder, for simplicity, just sorts the table alphabetically (ie. "l" follows "g"). There is a problem, however, in the reference decoder, as well as in the OpenType Sanitiser that ships in current browsers (but not the latest HEAD from @khaledhosny), whereby the decoder fails to decompress a WOFF2 font whose table directory/data is not already sorted alphabetically (I tried to address it myself here). So again, to avoid encoded fonts being rejected by browsers, I set the fontTools encoder to do this unnecessary table reordering.

Thanks.  That's unfortunate.  Guess we can remove this later and try to find an order that works better for Brotli.


> Lastly, the WOFF2 fonts made with fontTools encoder would be 100% bitwise identical with those produced by the reference encoder, if it was not for a minor difference in the way fontTools and the reference implementation handle the compression of repeating flags inside glyph coordinates: when the current flag is the same as the last one, but the number of repetitions is only 1, fontTools prefers to simply append the second flag as is, whereas the reference implementation sets the "flagRepeat" bit in the preceding byte and encodes the number of repetitions (1). From my tests, it looks like repeating the flag value like fontTools does produces slightly better Brotli compression.

Right.  While we can change that, I don't think that's a requirement, and slightly better Brotli compression sounds great to me.

I'm a bit confused though.  The glyf table transform throws away the flagRepeat bit and repeats flags individually and let Brotli take care of compression.  No?

I don't think the spec not being final is an issue given that many of use CC'ed on this issue also are involved in the changes going into the spec and the other implementations.

anthrotype:
> I'm a bit confused though. The glyf table transform throws away the flagRepeat bit and repeats flags individually and let Brotli take care of compression. No?

you're absolutely right. I was misled into thinking that the random size differences between the WOFF2 fonts produced by woff2_compress, on the one hand, and my fontTools woff2 encoder had to do directly with with the compressed flags' packing algorithm.

It turns out, the transformed tables produced by the two encoders are bitwise identical (despite the original sfnt glyf tables pack coordinate flags differently).

Using an hex editor, I verified that the only difference found when comparing the uncompressed "transformBuffers" of the two WOFF2 fonts (i.e. the actual data that goes in and out Brotli compressor) is, well... the head table's checkSumAdjustment -- and that's how it should be.
Anyway, the difference is negligible (around 28 bytes, for example, in the case of Roboto-Regular.ttf).

behdad:
Why do we expect checkSumAdjustment to be different?

anthrotype:
Because the original sfnt glyf tables use different coordinate flags, and their checksum will be different, and the checksum adjustment refers not to the WOFF2 font but to the original uncompressed sfnt font.

behdad:
Ah, right.  Thanks.

anthrotype:
> So is this all about coming up with a "nominal size" that is large enough? Then can't we just add 3 bytes per glyph to make sure? Or do we really need to encode with padding?

We could. But.. I don't quite like this approach.
If one simply increments `glyf.origLength += 3*numGlyphs` without also padding the table with null bytes between glyphs, then technically we are lying about the size of the original table.
Besides, the calculated checkSumAdjustment in the head table would also change because of that (ok, hardly anybody cares about checkSumAdjustment, but still).

Anyway, I tried to make such a WOFF2 test font with "incremented" origLength, but with unpadded glyf table. I then decoded it with fontTools, ot-sanitise and woff2_decompress.
They all successfully decompressed such WOFF2 without any warnings. However:

- woff2_decompress will write the wrong, incremented origLength to the glyf directory entry of the decompressed font. This means when one subsequently tries to decompile the resulting sfnt glyf table using fontTools, fontTools will raise a warning (not an exception), saying "too much 'glyf' table data";

- both fontTools and ot-sanitise correctly update the sfnt glyf directory entry's `length` with the reconstructed length.

All in all, I believe it's better to actually pad the glyf table, at least temporarily, before other decoders are updated.

What they basically need to be able to do is writing the sfnt table directory at the end of the whole glyf reconstruction process, rather than at the beginning like they both currently do.

behdad:
Thanks for the thorough investigation. :)

anthrotype:
I noticed a bug in the woff2 encoder whereby private data was not being written to file. This would result in a reported length different from the actual file size. So I fixed that in the last commit https://github.com/anthrotype/fonttools/commit/065470c7c681b74219308d7ada97c883e6482316. 

The problem is that git has found merge conflicts in the woff2 branch's ttLib/__init__.py.
So I'm forced to do a rebase on the tip of the current master, and fix each merge conflicts manually.

Sorry about that.

twardoch:
I haven't checked, but please make sure that the build scripts allow fontTools to build without any Brotli dependencies. Is that so?

anthrotype:
@twardoch yes, Adam. I haven't changed the setup.py. Brotli is currently required only if one tries to instantiate the WOFF2 reader/writer classes.

twardoch:
OK, cool.

Lib/fontTools/ttLib/woff2.py
```diff
@@ -0,0 +1,1023 @@
...
+		elif glyph.isComposite():
+			self._encodeComponents(glyph)
+		else:
+			self._encodeCoordinates(glyph)
+		self._encodeBBox(glyphID, glyph)
+
+	def _encodeComponents(self, glyph):
+		lastcomponent = len(glyph.components) - 1
+		more = 1
+		haveInstructions = 0
```
khaledhosny: Should’t this be `True` and `False`?

anthrotype: Isn't it the same thing in Python?
I took it from here:
https://github.com/behdad/fonttools/blob/master/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L421

anthrotype: I meant this line actually:
https://github.com/behdad/fonttools/blob/master/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L535
But you got the point.
Thanks for reviewing the code btw :-)

khaledhosny: It just makes the code a bit obscure. That referenced code probably just too old that it predates the introduction of `bool` in Python 2.3 or something.

khaledhosny: Indeed, this code was committed in 1999, four years before Python 2.3.

anthrotype: Yeah, maybe you're right.
We can fix that.

anthrotype:
this branch contains too many merge conflicts. I shall close it now, and reopen as a new PR with one commit per file.

PR #223, opened 2015-04-01
[sfnt] fix WOFF checksum adjustment calculation

anthrotype:
As suggested in a comment by @behdad inside the sfnt module's source code,

> the checksum for woff is not working correctly

The calculation is broken in WOFF1 because it's using the compressed `length` and the `offset` values from the WOFF directory entries, instead of using the `origLength` and the corresponding  `origOffset` from the original SFNT font.
The head table's checksum adjustment value of a WOFF font should refer not to the WOFF font itself, but to the input SFNT font.
This patch fixes that.

Now, if you compare the `checkSumAdjustment` of the input SFNT font with that of the resulting WOFF, they should finally match -- of course provided that the table data keeps the same order in the two fonts (i.e. using same `reorderTable` value for both).

Cheers,

C.

### DIFF
--- a/Lib/fontTools/ttLib/sfnt.py
+++ b/Lib/fontTools/ttLib/sfnt.py
@@ -111,6 +111,9 @@ class SFNTWriter(object):
 			self.DirectoryEntry = WOFFDirectoryEntry
 
 			self.signature = "wOFF"
+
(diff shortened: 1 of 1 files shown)


behdad:
Thanks!

PR #230, opened 2015-04-15
Support unittest in addition to doctest

brawer:


behdad:
Please rebase on top of master, so it doesn't show up your ltag commit here.

behdad:
Merged manually.  Thanks.

Issue #232, opened 2015-04-15
[ttx] Cannot decompile Apple’s Helvetica.dfont


brawer:
Since a few hours ago, ttx fails to process the 'name' table in Helvetica.dfont of MacOS X 19.9.5. Before, ttx would handle this font perfectly fine. This might be cased by the following commit:
https://github.com/behdad/fonttools/commit/8b0152f0a1ced963775e3b8d1b938a55c6c64b7d

$ ./Tools/ttx Helvetica.dfont
[...]
Dumping 'name' table...
Traceback (most recent call last):
  File "./Tools/ttx", line 6, in <module>
    ttx.main(sys.argv[1:])
  File "/Users/sascha/src/fonttools2/fonttools/Tools/fontTools/ttx.py", line 310, in main
    process(jobs, options)
  File "/Users/sascha/src/fonttools2/fonttools/Tools/fontTools/ttx.py", line 295, in process
    action(input, output, options)
  File "/Users/sascha/src/fonttools2/fonttools/Tools/fontTools/ttx.py", line 206, in ttDump
    bitmapGlyphDataFormat=options.bitmapGlyphDataFormat)
  File "/Users/sascha/src/fonttools2/fonttools/Tools/fontTools/ttLib/__init__.py", line 283, in saveXML
    self._tableToXML(tableWriter, tag, progress, quiet)
  File "/Users/sascha/src/fonttools2/fonttools/Tools/fontTools/ttLib/__init__.py", line 323, in _tableToXML
    table.toXML(writer, self)
  File "/Users/sascha/src/fonttools2/fonttools/Tools/fontTools/ttLib/tables/_n_a_m_e.py", line 72, in toXML
    name.toXML(writer, ttFont)
  File "/Users/sascha/src/fonttools2/fonttools/Tools/fontTools/ttLib/tables/_n_a_m_e.py", line 168, in toXML
    writer.write(unistr)
  File "/Users/sascha/src/fonttools2/fonttools/Tools/fontTools/misc/xmlWriter.py", line 48, in write
    self._writeraw(escape(string), indent=indent)
  File "/Users/sascha/src/fonttools2/fonttools/Tools/fontTools/misc/xmlWriter.py", line 73, in _writeraw
    self.file.write(s)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xa9' in position 0: ordinal not in range(128)

brawer:
This is on Python 2.7.5.

behdad:
[...]

PR #233, opened 2015-04-15
[xmlWriter] fix UnicodeEncodeError in #232

anthrotype:
in Python 2.x, writing a unicode string to a file that was opened in binary format does not raise TypeError -- unlike Python 3.x.
So under Python 2.x, the xmWriter always uses `tounicode` to coerce bytestrings to unicode strings. However, when xmlWriter tries to write them to the file opened in binary format, Python 2 uses the default system encoding ("ascii") to implicitly convert unicode to bytes, hence it fails (cf. #232).

A better approach to figure out whether writer expects bytes or unicodes could be to test if the file's `mode` attribute contains the string `"b"`, as in "wb", "wb+", "rb", etc.

Cheers,

Cosimo

### DIFF
--- a/Lib/fontTools/misc/xmlWriter.py
+++ b/Lib/fontTools/misc/xmlWriter.py
@@ -24,12 +24,10 @@ class XMLWriter(object):
 			self.file = fileOrPath
 
 		# Figure out if writer expects bytes or unicodes
-		try:
-			self.file.write(tounicode(''))
-			self.totype = tounicode
-		except TypeError:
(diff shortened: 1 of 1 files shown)


behdad:
The problem is: the file object can be anything that has a write().  We really just want to check for whether we can write a bytes / unicode.  So I'll just swap the test order.  Thanks.

behdad:
Fixed.  Please try.

anthrotype:
thanks!

PR #235, opened 2015-04-16
Emit XML even for names whose platform encoding is unsupported

brawer:
When dumping the Macintosh Skia font to XML, the TTX tool crashed
with a Python exception. After this change, the font can be dumped
fine. This was caused by a name record with platformID=1 (Macintosh)
and platEncID=2 (Traditional Chinese).


behdad:
Thanks.  I'm merging this.  We can adjust for the exception if you agree, on a followup commit.

behdad:
Merged manually.

PR #237, opened 2015-04-16
Raise UnicodeDecodeError for structurally invalid names

brawer:
This will make it impossible to process fonts with super-broken
name tables. But we do not handle arbitrary broken fonts anyway,
so this is arguably better than silently ignoring junk content.
Resolves a review comment in #235.

In the unit test, replace calls of deprecated unittest.assertEquals()
by calls of unittest.assertEqual().

### DIFF
--- a/Lib/fontTools/ttLib/tables/_n_a_m_e.py
+++ b/Lib/fontTools/ttLib/tables/_n_a_m_e.py
@@ -154,10 +154,7 @@ class NameRecord(object):
 		encoding = self.getEncoding()
 		if encoding == None:
 			return None
-		try:
-			return tounicode(self.string, encoding=encoding)
-		except UnicodeDecodeError:
-			return None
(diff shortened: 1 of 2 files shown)


behdad:
I have a better suggestion now:  lets add a errors argument, that we pass to decode.  Then user can choose between ignore, replace, etc.

As for the XML writing, I think we still should catch the case and write as bytes instead of Unicode.

Lets follow this unwritten rule: it's ok to not be able to load arbitrarily broken fonts.  But when we load a font, we should successfully write it to XML, load the XML back, and compile it to *a* font.

brawer:
Not sure I fully understand... mind taking over the change?

behdad:
Yep.  Thanks.  I'll send a PR for you to review.

behdad:
I fixed this differently.  Please check.

PR #240, opened 2015-04-16
Instruct git to ignore emacs backup files

brawer:


behdad:
The recommended way to do this is to add it to your home directory's config.  From man gitignore:

       ·   Patterns which a user wants Git to ignore in all situations (e.g.,
           backup or temporary files generated by the user’s editor of choice)
           generally go into a file specified by core.excludesfile in the
           user’s ~/.gitconfig. Its default value is
           $XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set
           or empty, $HOME/.config/git/ignore is used instead.

That said, I don't have a strong reason to oppose this, so I'll merge.  Thanks.

behdad:
Merged manually.  Thanks.

PR #242, opened 2015-04-18
[bezierTools] fix bug in splitLine function

anthrotype:
https://github.com/behdad/fonttools/issues/241

good catch @typemytype!

Indeed the function is not splitting the line when the two points are vertically aligned (same x coordinate) and `isHorizontal=True`.

While looking at this, I also noticed another inconsistency in the code.
When the value `t == 1`, i.e. when the `where` coordinate falls on `pt2`, the line is not split -- which is ok.
However, if `t == 0`, i.e. when `where` is on top of `pt1`, then a zero-length segment is created, with two overlapping points, which I think doesn't make sense.
 
Here's my fix. Please let me know what you think.
All best,

Cosimo

### DIFF
--- a/Lib/fontTools/misc/bezierTools.py
+++ b/Lib/fontTools/misc/bezierTools.py
@@ -80,11 +80,15 @@ def splitLine(pt1, pt2, where, isHorizontal):
         >>> printSegments(splitLine((0, 0), (100, 100), 100, True))
         ((0, 0), (100, 100))
         >>> printSegments(splitLine((0, 0), (100, 100), 0, True))
-        ((0, 0), (0, 0))
         ((0, 0), (100, 100))
         >>> printSegments(splitLine((0, 0), (100, 100), 0, False))
-        ((0, 0), (0, 0))
         ((0, 0), (100, 100))
+        >>> printSegments(splitLine((100, 0), (0, 0), 50, False))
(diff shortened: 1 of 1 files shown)


behdad:
Thanks.  The fix looks right.  Removing the degenerate segment also looks good but the segment was probably there intentionally.  This code isn't used in fontTools itself.  I wonder if any users make assumptions about that segment being there...

The way this was intended might have been: the line segment includes the first point but not the last point (ie, it's half-open).  With that definition, I can see how a split at t=0 might be desirable as to signal that there is a hit.

I see a copy of this code in robofab, which doesn't have the bug in question.  The use of that function in robofab.pens.marginPen indeed relies on "hitting" at t==0: https://github.com/robofab-developers/robofab/blob/master/Lib/robofab/pens/marginPen.py#L48

I suggest we leave that part as is, and document it so.

cc @letterror.

anthrotype:
@behdad you're right. It's not a mistake, but it was there intentionally. I could not figure out why it was like that, but now I see.
I reverted to the the original `0 <= t < 1` in 663d3eeb778d00b3d248583a515f61c81f4e5a18
thanks

behdad:
Thanks.  Pushed a squashed version of the two.

typemytype:
thanks!

Issue #246, opened 2015-04-20
xmlWriter.py: Capitalize UTF-8


frankrolf:
Historically, FontTools has written the first line of an XML file like this:
`<?xml version="1.0" encoding="UTF-8"?>`, 
(with UTF capitalized, as Unicode.org and W3C do).

In this fork, the UTF has changed to lowercase. 
Since Robofont now has moved to use this fork of FontTools, thousands of glyph files change in source control (I am sure not only on my system).

Please change UTF back to uppercase.

benkiel:
+1

See a patch I submitted for a workaround to this.

benkiel:
Note, python3.x isn't something I deal with, so the string formatting I did in my patch may no longer be legal, but the idea is there.

behdad:
Hum.  The fonttools that I inherited was writing XML in latin1, not UTF-8.  I fixed that in 4b3df49b5b70b3618e91e3ab954cedca773831d6.

I can make the change you are requesting, but want to understand the issue completely first.

@benkiel which patch is that?

frankrolf:
Does the line`parser = ParserCreate("latin1")` really specify the encoding of the resulting XML file? 
I don’t think it does. When I write a TTX file with the sourceforge fork, it comes out as UTF-8.

I think the XML encoding is passed to the function here:
http://sourceforge.net/p/fonttools/code/ci/master/tree/Lib/xmlWriter.py#l25

Thanks for considering this fix.

behdad:
> Does the lineparser = ParserCreate("latin1") really specify the encoding of the resulting XML file?

No.  That's for the xmlReader.  I was talking about this line:

```patch
-	def __init__(self, fileOrPath, indentwhite=INDENT, idlefunc=None, encoding="ISO-8859-1"):
+	def __init__(self, fileOrPath, indentwhite=INDENT, idlefunc=None, encoding="utf-8"):
```

> I don’t think it does. When I write a TTX file with the sourceforge fork, it comes out as UTF-8.

I really 

> I think the XML encoding is passed to the function here:
> http://sourceforge.net/p/fonttools/code/ci/master/tree/Lib/xmlWriter.py#l25

Right.  And I don't see where "UTF-8" is being passed to it in the sourceforge version.

> Thanks for considering this fix.

miguelsousa:
@behdad This difference:
https://github.com/behdad/fonttools/blob/master/Lib/fontTools/misc/xmlWriter.py#L44
[...]

PR #247, opened 2015-04-21
Test that XMLWriter preserves capitalization of passed encoding

brawer:
Unit test for #246


behdad:
Thanks Sascha.  I'm actually going to hardcode "UTF-8" in the output.  So I'll land this, make that change and update the tests.

behdad:
Merged manually.

PR #256, opened 2015-05-03
[OS/2] update fsFirstCharIndex and fsLastCharIndex upon compile

anthrotype:
this addresses a [request](https://github.com/behdad/fonttools/issues/255) by @lemzwerg to recalculate the `fsFirstCharIndex` and `fsLastCharIndex` of OS/2 table while reassembling a font.

I added a method to the OS/2 table class, which is called upon compiling that table. It simply gathers all codes specified in cmap subtables (3, 0) or (3, 1) and takes the min/max among these.

Should this be the default behaviour when compiling the OS/2 table (like in the current patch), or it should be controlled through some sort of option?

### DIFF
--- a/Lib/fontTools/ttLib/tables/O_S_2f_2.py
+++ b/Lib/fontTools/ttLib/tables/O_S_2f_2.py
@@ -121,6 +121,7 @@ class table_O_S_2f_2(DefaultTable.DefaultTable):
 		self.panose = sstruct.unpack(panoseFormat, self.panose, Panose())
 
 	def compile(self, ttFont):
+		self.updateFirstAndLastCharIndex(ttFont)
 		panose = self.panose
 		self.panose = sstruct.pack(panoseFormat, self.panose)
 		if self.version == 0:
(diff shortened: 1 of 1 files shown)


lemzwerg:
Very nice!  Thanks for the quick patch.  My opinion is that a font shouldn't lie.  Consequently, automatically computed values should be used as much as possible to avoid human failure.  I plead for your code to be integrated unconditionally.

behdad:
Can you add a comment to the XML output, ala the maxp table, to say that these are autocomputed?

anthrotype:
do you mean a general comment at the top of the table, like in maxp or loca:

    writer.comment("Most of this table will be recalculated by the compiler")

or a specific one which only refers to `fsFirstCharIndex` and `fsLastCharIndex`?

behdad:
Maybe be more specific, instead of "Most of this table", say which fields are recalculated.

anthrotype:
OK, I have added a comment at the top OS/2 table XML output , which now reads like that:

```xml
 <OS_2>
    <!-- The fields 'fsFirstCharIndex' and 'fsLastCharIndex'
         will be recalculated by the compiler -->
 <version value="4"/>
```

anthrotype:
OK, now using isUnicode() and with the commits squashed into one

lemzwerg:
Thanks!

PR #260, opened 2015-05-07
Change whitespace in subset.py to conform with Python style guide

brawer:
The Python style (outside of Google) is four spaces for indentation.
After this change, pylint still has complaints about this file,
but these formatting changes will probably improve the health
metrics of landscape.io. Personally, I couldn't care less about
whitespace, but it is hard to spot actual problems in a sea
of lint warnings.

### DIFF
--- a/Lib/fontTools/subset.py
+++ b/Lib/fontTools/subset.py
@@ -306,361 +306,362 @@ Example:
 
 
 def _add_method(*clazzes):
-  """Returns a decorator function that adds a new method to one or
-  more classes."""
-  def wrapper(method):
-    for clazz in clazzes:
-      assert clazz.__name__ != 'DefaultTable', 'Oops, table class not found.'
-      assert not hasattr(clazz, method.__name__), \
-          "Oops, class '%s' has method '%s'." % (clazz.__name__,
-                                                 method.__name__)
-      setattr(clazz, method.__name__, method)
-    return None
-  return wrapper
+    """Returns a decorator function that adds a new method to one or
+    more classes."""
+    def wrapper(method):
+        for clazz in clazzes:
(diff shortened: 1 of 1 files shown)


behdad:
I remember I had all tabs; for the Open Source release, Roozbeh convinced me that it's Google recommendation to use two spaces, so I did that.  When I merged the subsetter back into fonttools, I wish I had kept the tabs, which is what the rest of the codebase uses.

I'll merge this now, but we should change our landscape.io setting to ignore whitespace issues.  There are a couple of outstanding Pull Requests against the subsetter that I would need to rebase.

behdad:
Should we just change to 8, and use tabs now that we touched it?

brawer:
Or, perhaps, we should try to comply with PEP8?
https://www.python.org/dev/peps/pep-0008/

I wouldn't want to spend too much time on style things (the discussions can
be such a time sink), but there's certainly some advantage to being in line
with other Python code out there. We could try to run this on the codebase:
https://pypi.python.org/pypi/autopep8/1.1.1

Shall I give it a try?

-- Sascha

2015-05-07 11:16 GMT+02:00 Behdad Esfahbod <notifications@github.com>:

mvasilkov:
I second PEP8 so very much.

brawer:
Here's a sample change generated by autopep8:
https://github.com/behdad/fonttools/pull/261

-- Sascha

2015-05-07 14:15 GMT+02:00 я котик пур-пур <notifications@github.com>:

brawer:
And here's the same PEP8 reformatting for an entire directory:
https://github.com/behdad/fonttools/pull/263

2015-05-07 14:25 GMT+02:00 Sascha Brawer <sascha@brawer.ch>:

PR #262, opened 2015-05-07
[macCreatorType] disable workaround in py27 for setting as well getting

anthrotype:
I know nowadays one no longer needs to set the creator and type on a mac, but because we already fixed the "getting" part, we may as well fix the "setting" one...

### DIFF
--- a/Lib/fontTools/misc/macCreatorType.py
+++ b/Lib/fontTools/misc/macCreatorType.py
@@ -29,10 +29,4 @@ def getMacCreatorAndType(path):
 
 def setMacCreatorAndType(path, fileCreator, fileType):
 	if MacOS is not None:
-		if sys.version_info[:2] < (2, 7) and sys.byteorder == "little":
-			# work around bug in MacOS.SetCreatorAndType() on intel:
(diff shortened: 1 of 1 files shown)


behdad:
The Python bug you referred to before, suggested that only one of set and get was affected.  Can you check again?

anthrotype:
Yes, I did check. The setter is indeed currently broken on 2.7, unless we disable that workaround.

behdad:
No no.  What I'm saying is this paragraph [0]:

"""
Annoyingly enough only 
GetCreatorAndType needed fixing, SetCreatorAndType already did the right 
thing (makeing the two functions inconsistent with each other).
""""

So I think we should remove the hack for Set, not contain it to < 2.7.

[0] http://bugs.python.org/issue1594

anthrotype:
I see what you mean now. I'll check with pyhton 2.6 as well and let you know

anthrotype:
Indeed, you're right. With Python 2.6.9, the current setting function (the one with the workaround) writes the creator and type with reversed strings. So we should disable the workaround for all python versions, not just 2.7.

PR #263, opened 2015-05-07
Use PEP-8 whitespace conventions for fontTools/ttLib/tables

brawer:
Changes were automatically generated with autopep8.


behdad:
I'm rather hesitant to make such huge cosmetic changes, specially given that there are still people transitioning from the old fonttools, so I don't want to disrupt the code base drastically.

Should we just tune landscape.io re whitespace instead?

brawer:
Sure. Personally, I'd rather spend energy on actual changes than on style
issues anyway.

-- Sascha

behdad:
Agreed.  Lets let this go for now and I'll try to adjust landscape.io settings a bit.

behdad:
I'm watching this now:
https://www.youtube.com/watch?v=wf-BqAjZb8M

PR #264, opened 2015-05-08
[py23] simulate "wide" unichr and ord on "narrow" UCS2-only Python 2

anthrotype:
I drop this here in case anybody may be interested.

As you know, when working with non-BMP unicode characters on a Python 2 which was built with UCS-2 Unicode support (i.e. the so called "narrow" builds), then the built-in `unichr` and the corresponding `ord` functions are limited to the range(0x10000). This is the case of most Python 2 configurations, including the OS X system Python and the Windows installer from Python.org.
This is quite annoying, given that Python 3 is always configured with UCS-4 support.

The py23 compatibility layer inside the misc sub-package could be the right place to add two simple workarounds which extends the unichr and byteord range to the full Unicode (0x110000), even on narrow Python 2 builds.

For example, think about looking up the built-in unicodedata module for the name of a Unicode character:

```python
>>> unicodedata.name(unichr(0x10000))
'LINEAR B SYLLABLE B008 A'
```

The line above works fine on Python 3 (of course, using our `unichr` alias of the `chr` built-in), but will fail on the majority of Python 2 builds.

What do you think? Does it seem too hackish?

### DIFF
--- a/Lib/fontTools/misc/py23.py
+++ b/Lib/fontTools/misc/py23.py
@@ -23,7 +23,6 @@ try:
 
 		def unichr(i):
 			"""
-			
 			Return the unicode character whose Unicode code is the integer 'i'.
 			The valid range is 0 to 0x10FFFF inclusive.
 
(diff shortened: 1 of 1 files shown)


behdad:
Thanks!

twardoch:
In the same spirit, if fontTools uses unicodedata now (does it?), you could somehow incorporate https://pypi.python.org/pypi/unicodedata2/ into it.

behdad:
The only thing we use unicodedata for is to get character names to print out in comments in the XML.  So, not worth it.

behdad:
I suppose we can try importing unicodedata2 before unicodedata...

PR #267, opened 2015-05-09
[subset] add option to prune unused ulUnicodeRange bits

anthrotype:
this is the same as https://github.com/behdad/fonttools/pull/191
all the merge, rebase or cherry-pick would fail, so I was forced to redo the changes manually from a new branch.

From the previous thread, the consensus was that we would have this option always on in the subsetter, and maybe we could think about having it in ttLib later.

@behdad in the other PR about OS/2 fsFirstCharIndex, you suggested to use isUnicode() method of cmap subtables. I wonder if I should use that here as well. Currently I'm only checking platformID 3 and platEncID 1 and 10, because those are the only ones which are mentioned in the relevant [paragraph](http://www.microsoft.com/typography/otspec/os2.htm#ur) of OS/2 specification.

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -288,6 +288,14 @@ Other font-specific options:
       required by the standard, nor by any known implementation.
   --no-canonical-order
       Keep original order of font tables. This is faster. [default]
+  --prune-unicode-ranges
+      Update the 'OS/2 ulUnicodeRange*' bits after subsetting. The Unicode
+      ranges defined in the OpenType specification v1.7 are intersected with
+      the Unicode codepoints specified in the font's Unicode 'cmap' subtables:
+      when no overlap is found, the bit will be switched off. However, it will
+      *not* be switched on if an intersection is found.  [default]
+  --no-prune-unicode-ranges
+      Don't change the 'OS/2 ulUnicodeRange*' bits.
 
 Application options:
   --verbose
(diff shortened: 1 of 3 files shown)


behdad:
Yes, no hardcoding please.  Use isUnicode().  We started considering symbol cmaps as Unicode as well, as there are MS fonts that have a symbol cmap and PUA encoding, which technically, is still Unicode.

anthrotype:
Ok, I just fixed it and squashed into the previous commit.

behdad:
Please rename all instances of "recalc" to "update".  Right now there's a discrepancy in the docstring changes where you have --recalc-... and --no-update-...

I'm not 100% sure which name I like more.  "recalc" doesn't sound right given that we rely on the previous setting.  So "update" or "prune" are better.  Maybe "prune".  That's much closer to what this actually does.

anthrotype:
Good point. Sorry about that.
I just changed all to "prune".

anthrotype:
I had forgotten about this PR.

The consensus was to add an option to the subsetter to prune unused unicode range bits, and to have it enabled by default.

So I have cleaned this up and rebased on master.

The subsetter has a new `--prune-unicode-ranges` default option (which can be disabled with `--no-prune-unicode-ranges`).

The OS/2 table module contains the mapping between Unicode blocks and the `ulUnicodeRange*` bits defined in the current OT spec.

The OS/2 table class has some new methods:
- `getUnicodeRanges()`: get the sequence of `ulUnicodeRange*` bits currently enabled;
- `setUnicodeRanges(bits)`: set these manually given a sequence of integers;
- `recalcUnicodeRanges(pruneOnly=False)`: recalculate them based on the codepoints from the cmap's Unicode subtables, using the same algorithm used by FontLab (cc @twardoch): which is, "if any codepoint falls in the range, set that bit (plus set bit special bit 57 if any codepoint is beyond BMP)". A keyword argument `pruneOnly` allows to simply clear the bits for which there's no intersection, without also activating the ones for which there are (that's what subset.py is using).

PR #270, opened 2015-05-11
implement WOFF2 encoder/decoder as standalone module

anthrotype:
This is the same code as in https://github.com/behdad/fonttools/pull/222, with two differences:

1) The normalisation step, which was scattered between the `woff2`, `ttLib.__init__` and the `_g_l_y_f` modules, is all done inside the woff2 module, which is now self-contained.

2) I'm using a new trick in order to initialise a compatible reader/writer instance based on the given 'flavor'. I can override the `__new__` method of SFNTReader and SFNTWriter and use that as a sort of object factory which return an instance of a sub-class inheriting from SFNTReader or SFNTWriter. 

This is very handy. Think about it: we could add support for even more flavours (e.g. TTC, DFONT, EOT?) without needing to modify the API, but simply sub-classing the base reader/writer classes. All other scripts that use SFNTReader/Writer directly are immediately compatible with the new flavour (provided the new sub-classes keeps the same interface).

I believe such an approach would allow us more flexibility not only in implementing new formats, but also in reorganising the existing ones in a more object-oriented fashion. I'm thinking about the current SFNTReader and SFNTWriter classes, which do too many things at once: i.e. TTFs, OTFs, mac suitcase fonts, single fonts in TTC, and WOFF 1.0. This means the code got so complex that it turned out to be impossible to fit WOFF2 as well inside it.

On the other hand, a much cleaner situation would be having two abstract reader and writer classes which define the API, and more specialised subclasses inheriting from them, while also using the `__new__` method of the base class as a factory.

For example, the WOFF 1.0 reader/writer currently embedded in the SFNT one, shares several things not only with the original SFNT itself, but obviously also with its WOFF 2.0 brother. Duplicate code was all but inevitable, given the impossibility to squeeze WOFF2 inside the sfnt module for risk of exploding, as well as the impossibility of inheriting from/override its methods in a clean, simple way.

In the end, all these cross-relationships could be better captured by splitting the current catch-all mother class into different classes and modules linked by inheritance and imports.

what do you guys think?
Please, let me know.
Thanks.


Lib/fontTools/ttLib/woff2.py
```diff
@@ -0,0 +1,1083 @@
...
+			raise TTLibError("Transform for table '%s' is unknown" % tag)
+		if tag == 'glyf':
+			# reconstruct both glyf and loca
+			self.glyfTable = WOFF2GlyfTable()
+			data = self.glyfTable.reconstruct(rawData)
+		elif tag == 'loca':
+			assert len(rawData) == 0, "expected 0, received %d bytes" % len(rawData)
+			if not hasattr(self, 'glyfTable'):
+				# make sure glyf is loaded first
+				self['glyf']
```
adrientetar: Does this do anything sensible?

anthrotype: Because the transformed loca table is zero bytes, it must be reconstructed alongside the glyf table. The SFNTReader (from which WOFF2Reader inherits) allows to access table data randomly, but a WOFF 2.0 font is meant to be read as a stream, as there are no offsets to specific tables. If the users tries to get "loca" before "glyf", then we must reconstruct the latter as well. But we don't want to do it again and again, so the result is stored as "data" attribute.

adrientetar: I see, thanks for the explanation.

anthrotype:
I just added two little scripts in the Snippets folder. They work the same way as woff2_compress and woff2_decompress from https://github.com/google/woff2. The only difference is they use fontTools ;-)

behdad:
Cosimo, would you consider this branch ready to merge now?

anthrotype:
@behdad it's kind of ready, but ideally we would add a unit test.
I'll create a "TestData" folder and put a TTX file from an open source font. I'll use Pablo Impallari's Lobster, like you suggested in another thread.
It need not be a full test suite for the format -- i.e. https://github.com/w3c/woff2-tests --, just a unit test for the main features of the woff2 module.

anthrotype:
I'm not sure where I should palce the folder containing the test fonts in TTX format to be used by unit tests.

Currently we have the `*_test.py` modules in the same location as the modules they refer to.
If I do the same for woff2_test.py, then for consistency I should also place the folder with the test fonts somewhere inside "Lib/fontTools". In fact, I can't use a path relative to the test module to the data folder if the latter is located outside the source distribution, otherwise the test would fail if it is run from a different location (e.g. site-packages folder after installation?).
Besides, I don't know if it's a good idea to include such test fonts along with the python source code...

Wouldn't it be better if we move all our test modules in a separate "tests" folder located in the root of the repo, outside the "Lib/fontTools" folder, and create a "data" subfolder which would contain the resources needed for the unittest fixtures, such as TTX files, etc.?
Then the test suite will no longer be included in the source distribution, but again this is not such a bad thing.

What do guys you think? /cc @behdad @brawer @djdm

brawer:
Sure, moving the tests into a separate directory sounds good.

anthrotype:
I think the woff2 branch is finally ready to be merged.
I have added the unit tests and substantially refactored the code.
Sorry this has taken so long...

The travis build passes because brotli is not installed on the virtual machine, so the woff2_test module is skipped. I will update the .travis.yml file to `pip install` brotli in another PR.

Please have a look and let me know if you have any comments or suggestions.
Thanks.

anthrotype:
Just added a few more commits to enable the WOFF2 tests on Travis:
- add pip's `requirements.txt` containing a link to [Brotli repository](https://github.com/google/brotli); 
- update Travis' GCC to version 4.8 as required by Brotli (see: https://github.com/travis-ci/travis-ci/issues/1379)
- use pip to install Brotli from source before running tests;

All the WOFF2 tests now run successfully on Travis' Ubuntu, as well as (locally) on my Windows 7 virtual machine and OS X Yosemite, both python 2.7 and 3.4.

anthrotype:
I rebased on top of master and replaced all `StringIO` with `BytesIO` in woff2 and woff2_test modules.

behdad:
I have a feeling that tomorrow is the day this shall land...

anthrotype:
I just set up a branch over at https://github.com/daltonmaag/brotli/tree/ci to build compiled wheel packages for Windows and OS X using Appveyor and Travis respectively.

If you want to try the woff2 module, you can install Brotli by simply doing
```
pip install --find-links https://github.com/daltonmaag/brotli/releases/latest brotli
```

I also submitted a PR to upstream https://github.com/google/brotli/pull/136. If and when they merge that, they could have the wheels uploaded automatically on the Python Package Index (PyPI). This way, users will be able to do a simple `pip install brotli` and have the precompiled wheel installed for each target platform (without needing any developer tools).

anthrotype:
@behdad what else is left to do here? Are we waiting for brotli to be uploaded to PyPI?

behdad:
Nothing left, just me reviewing.  I'll try to do today.

behdad:
Thanks Cosimo!

benkiel:
Forgive if this isn't the best place to ask this; I'm wondering if this pull has the htmx transformation bit that isn't in the WOFF2 spec yet, but is being added. I've looked through the code and I don't think it does (which is what I want to confirm, as I don't think any browsers support this), but in case I missed something in my reading of the code I wanted to check.

anthrotype:
@benkiel No, it doesn't (yet). There's a branch on my fork where I have experimented with it. But it's best to wait until it's finalised in the spec.

benkiel:
@anthrotype 100% agree, just wanted to confirm. I would vote that when it is finalized in WOFF2 spec, fontools has, for a bit, a switch to turn it on/off until one can be sure that WOFF2 readers support it.

anthrotype:
I was referring to this: https://github.com/anthrotype/fonttools/tree/woff2_hmtx
It was just a quick and dirty sketch to assess whether there is any gain is in terms of file size (it turned out to be around 1% on average).
Anyway, yes I agree we'll need to wait for a while before enabling that by default.
By the way, thank you for reviewing the code! :)

behdad:
We definitely won't ever have code in master that would produce fonts that work on no platform. Ie, we are not crazy. :-)

PR #273, opened 2015-05-20
[setup.py] use setuptools if available, or fall back to distutils

anthrotype:
This patch addresses issue https://github.com/behdad/fonttools/issues/272

If setuptools raises an ImportError, then it falls back to the old distutils method.

I modified the four scripts so that they can work as setuptools 'console_scripts', in which the main function takes no arguments.

I also added an empty identifier "" in the 'packages' argument of setup function, otherwise the 
standalone sstruct.py and xmlWriter.py module -- which we have put back at the root of Lib for legacy -- would not be correctly installed.

I also removed py2exe since that is made redundant by the setuptools 'console_scripts' approach. This in fact will automagically create Windows executables (.exe) for each script using the specified entry point (I think this is really cool!).

Finally, I removed the custom 'build_ext' command which is no longer needed, since the "Src/eexecOpmodule.c" extension module was removed some time ago from behdad's fork.

### DIFF
--- a/Lib/fontTools/inspect.py
+++ b/Lib/fontTools/inspect.py
@@ -251,7 +251,9 @@ class Inspect(object):
 		self.scrolled_window.add(self.treeview)
 		self.window.show_all()
 
-def main(args):
+def main(args=None):
+	if args is None:
+		args = sys.argv[1:]
 	if len(args) < 1:
 		print("usage: pyftinspect font...", file=sys.stderr)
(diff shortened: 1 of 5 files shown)


anthrotype:
I pushed a new commit to fix this:

    WARNING: '' not a valid package name; please use only.-separated package names in setup.py

The "py_modules" argument works with both setuptools and distutils.

PR #274, opened 2015-05-23
[maxp] calculate correct maxSizeOfInstructions

anthrotype:
Fixes an issue whereby VTT calculates an incorrect value, and MS Font
Validator raises a warning message (W1900).

VTT wrongly includes the length of font-wide hinting tables (fpgm, prep)
in the computation, whereas the spec clearly says that this value is:

"the maximum size in bytes for all of the instructions associated with a
particular glyph."

https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6maxp.html

### DIFF
--- a/Lib/fontTools/ttLib/tables/_m_a_x_p.py
+++ b/Lib/fontTools/ttLib/tables/_m_a_x_p.py
@@ -73,6 +73,7 @@ class table__m_a_x_p(DefaultTable.DefaultTable):
 		maxCompositeContours = 0
 		maxComponentElements = 0
 		maxComponentDepth = 0
+		maxSizeOfInstructions = 0
 		allXMaxIsLsb = 1
 		for glyphName in ttFont.getGlyphOrder():
 			g = glyfTable[glyphName]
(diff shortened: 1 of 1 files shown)


anthrotype:
@behdad I'm now using `getBytecode` instead of accessing `bytecode` attribute directly. I also squashed the commits and pushed a reworked branch with the `git push -u +<new_branch>:<old_branch>` trick.

anthrotype:
I noticed that ttfautohint also does the same as VTT: that is, it includes the fpgm and prep in the calculation of the maxSizeOfInstructions field.
If the two major implementations agree on disagreeing from the spec, there must be a reason.
So it's better to put this PR on hold while we get more clarifications on the OpenType mailing list.

behdad:
At the end of the day, no one should care about this number these days.  I'm fine with doing the right thing though.

Also, this is still all controlled by recalcBBoxes.  I think that's fine though.

anthrotype:
on the OT mailing, Greg Hitchcock confirmed that the `maxSizeOfInstructions`

> should be a maximum of the glyph instructions, the pre-program, and font program.

He also said that neither Apple's nor Microsoft's current rasterizers use this value any more.

I think we can close this PR.

PR #276, opened 2015-05-23
warn when fsSelection bits aren’t defined or don’t match head.macStyle bits

moyogo:


behdad:
Humm.  Right now we don't do any automatic version detection here.  If we are going to change version, then we should do a more involved act.

moyogo:
Ah yes, this only works going from version 2 or 3 to 4.
I guess it can be both ways, if a version from 1 to 4, sxHeight, etc. need to be added as well but if one adds sxHeight, etc. the version should be valid.

moyogo:
I changed the commits so instead we just warn when fsSelection bits 7, 8 or 9 are on a version < 4.
It will also warn if bit 6 is on and bits 0 and 5 are not of, and also if bits 0 and 5 don’t agree with head.macStyle bits 1 and 0.

Lib/fontTools/ttLib/tables/O_S_2f_2.py
```diff
@@ -125,6 +127,19 @@ def decompile(self, data, ttFont):
 	def compile(self, ttFont):
 		self.updateFirstAndLastCharIndex(ttFont)
 		panose = self.panose
+		head = ttFont["head"]
+		if self.fsSelection & 1 != head.macStyle & 1<<1:
+			log.warning("fsSelection bit 0 (italic) and "
+				"head table macStyle bit 1 (italic) should match")
+		if self.fsSelection & 1<<6 != head.macStyle & 1:
```
anthrotype:
```diff
-if self.fsSelection & 1<<6 != head.macStyle & 1:
+if self.fsSelection & 1<<5 != head.macStyle & 1:
```

PR #277, opened 2015-05-23
[OS/2] fix misspellings 'usMaxContext', 'usFirstCharIndex', 'usLastCh…

anthrotype:
Fixes https://github.com/behdad/fonttools/issues/74

### DIFF
--- a/Lib/fontTools/ttLib/tables/O_S_2f_2.py
+++ b/Lib/fontTools/ttLib/tables/O_S_2f_2.py
@@ -60,8 +60,8 @@ OS2_format_0 = """
 	ulUnicodeRange4:        L       # character range
 	achVendID:              4s      # font vendor identification
 	fsSelection:            H       # font selection flags
-	fsFirstCharIndex:       H       # first unicode character index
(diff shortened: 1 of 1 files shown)


davelab6:
I think this came up before and there was a concern that fixing it would break legacy scripts, so perhaps you could add a 2nd commit that allows legacy scripts to continue to work?

anthrotype:
You're right. What about this https://github.com/behdad/fonttools/commit/e63cbe8d3c16ef0460e52ad8880c2b9998a2e182?

davelab6:
Looks perfect to me :)

anthrotype:
I tried to fix misspellings with other misspellings! Thank you @djdm for spotting them.
I rebased and squashed the new commit into the previous one.

behdad:
I'm also concerned about people who build tables in code.  Perhaps better way to fix this is to provide properties for the wrong values.  Or just ignore this problem and accept them as historical artifacts.

anthrotype:
Thanks for the suggestion. Yes, it makes more sense to use properties.
I changed it accordingly and rebased. Please let me know if it's OK now.

PR #279, opened 2015-05-23
[ttx] 'touch' output file to avoid race condition in choosing file names

anthrotype:
Fixes https://github.com/behdad/fonttools/issues/45


behdad:
Why is the utime() needed?

anthrotype:
yeah, it's not needed... I was just trying to emulate the `touch` command from python.
I have simplified it now.

anthrotype:
I just noticed that after this patch, when one does `ttx -l font.ttf` to list the table directory, an empty file is being created. Sorry about that, the fix is coming shortly.

PR #285, opened 2015-06-04
Move unit tests to separate "tests" folder

anthrotype:
I think it makes more sense to keep the tests separate from the source code.

I created a new "tests" folder in the root of the repository, mirroring the folder structure of the fontTools package. I moved there all the unittest modules, each in its respective subfolder.

I changed all relative imports in the test scripts to absolute imports, since they are no longer part of the package.

I modified the `run-tests.sh` shell script so that it performs all the doctests first, which are still located in the ./Lib folder, and then all the unittests in the ./tests folder.

The shell script extends the PYTHONPATH automatically. However, if one wishes to run individual test scripts, then the fontTools package must be importable -- e.g. by using setuptools "develop" or pip "editable" mode.

Please let me know what you think,
Thanks.

Cosimo


behdad:
So originally I didn't have any objection with tests being in-tree, because they don't have *any* overhead on the library itself, and test files that are close to code they test are much more likely to be updated over time.  So, why do you think we should move?

anthrotype:
I see what you mean. But then also the test data used by the tests should be located somewhere in the package, right?
See also my comment in the woff2 thread: https://github.com/behdad/fonttools/pull/270#issuecomment-107532229

anthrotype:
Actually, if we don't want the *_test.py and the respective test fonts to be installed in the Python site-packages we could always not include them in the setup.py.
To make sure the test data is reachable from the unit test's location, the two have to go together. Either we install them both or none.
Anyway, if you think it's not a good idea, I'll close the PR and try another strategy.

anthrotype:
OK, let's keep the unittests in the same folder as the modules they test.
I will create a "test_data" folder inside Lib/fontTools/ttLib and place two test fonts in TTX format, one TT-flavoured, the other CFF-flavoured, which I'll use for testing the woff2 module.

One last thing and then I stop: what about putting all the *_test.py inside a "tests" subfolder, one for each sub-package? E.g. fontTools/ttLib/tests, fontTools/misc/tests, etc. The relative import would still work, by just adding another dot (e.g. from ..woff2 import *).

PR #288, opened 2015-06-10
fix xmlWriter and _l_t_a_g tests failing on Windows because of \r\n

anthrotype:
@brawer the explicit "\n" newline separator was making some of the unittests fail on Windows, where python uses "\r\n" instead. I replaced them with `os.linesep`.

cheers,

C.


anthrotype:
do you think is it worth to set up Appveyor so that we can run tests on Windows, like we do for Linux on Travis?

behdad:
Yes, Appveyor would be great.  Can you and @ebramino work on that?

PR #289, opened 2015-06-10
[.travis.yml] pip install unittest2 for python2.6

anthrotype:
I'm struggling to find a single unittest idiom that would work for all python versions between 2.6 and 3.4.
The best solution I found is to install the `unittest2` package when Python version is 2.6.
That will backport all the new unittest features introduced in Python 2.7 and 3.x.
For example, I want to skip tests in woff2 module if brotli is not installed, instead of fail. Or I want to assert that an exception message matches a specific regular expression pattern. All this is not available in python2.6 unittest built-in module.
The pip installer is already on the Travis virtual machine, so we simply need to check if the Travis python version is 2.6, and then do a pip install `unittest2`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ python:
   - "pypy"
   - "pypy3"
 install:
+  - if [[ $TRAVIS_PYTHON_VERSION == 2.6 ]]; then pip install unittest2; fi
   - make install
 script:
   - make check


djdm:
Do we still need to support Python 2.6?

behdad:
I like to use this opportunity to drop python 2.6 support...

anthrotype:
If it were up to me, I'd drop python2.6 support almost instantly. Maybe @behdad could poll other users' opinions on the fontTools mailing list...

behdad:
Thinking about it today I decided that I'll check my 1and1 Linux hosting box and if that's 2.7, we definitely drop.  Now checked and it's 2.6.6...  That's a Debian 6 machine.

Let me ask around, see what latest version of RHEL ships.

anthrotype:
RHEL 7 comes with python 2.7 (previous version had python 2.6)
http://developerblog.redhat.com/2014/06/26/rhel-7-is-for-developers/

behdad:
Ok, lets drop 2.6 then.  Which means, we can use dictionary and set comprehensions:

```
>>> {k:v for k,v in [(1,2)]}
{1: 2}
>>> {x for x in [1, 2]}
set([1, 2])

```

behdad:
Dropped 2.6.  Lets close.

anthrotype:
:+1:

PR #293, opened 2015-06-13
[setup.py] monkeypatch py_compile to halt compilation on syntax errors

anthrotype:
This is an attempt to fix issue https://github.com/behdad/fonttools/issues/290.
It uses a trick found [here](http://stackoverflow.com/a/2240549) which monkeypatches the py_compile.compile function used by distutils, so that it will raise an exception instead of a warning on any errors found while bytecompiling the modules.
With this the `make all` command will properly fail.

### DIFF
--- a/Lib/fontTools/merge.py
+++ b/Lib/fontTools/merge.py
@@ -528,42 +528,42 @@ def mapLookups(self, lookupMap):
 	     otTables.ChainContextPos)
 def __merge_classify_context(self):
 
-  class ContextHelper(object):
-    def __init__(self, klass, Format):
-      if klass.__name__.endswith('Subst'):
-	Typ = 'Sub'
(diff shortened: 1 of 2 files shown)


anthrotype:
the Python 3 builds on Travis were failing because of mixed tabs and spaces (in Python 3 this raises a TabError). I fixed it a subsequent commit.
Shall I open a separate PR for that, is it ok like this?

anthrotype:
btw, it was actually thanks to the patch that the TabError in merge.py could be found.. I guess one more reason to have `make` return failure on compilation errors.

behdad:
Thanks Cosimo.

I also added you as a collaborator on the repo so you can push trivial patches directly.

anthrotype:
Thank you!

PR #300, opened 2015-06-24
[GX] Make it easier to construct ‘fvar’ tables from code

brawer:
In the previous table-driven implementation, client code had to
know the internals of the ‘fvar’ structure for correctly adding
variation axes to a font. In the new implementation, clients
do not have to futz around with binary offsets (which makes it
more likely that tools build correct fonts).

### DIFF
--- a/Lib/fontTools/ttLib/tables/_a_v_a_r.py
+++ b/Lib/fontTools/ttLib/tables/_a_v_a_r.py
@@ -29,8 +29,7 @@ class table__a_v_a_r(DefaultTable.DefaultTable):
         self.segments = {}
 
     def compile(self, ttFont):
-        fvarAxes = ttFont["fvar"].table.VariationAxis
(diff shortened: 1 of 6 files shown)


anthrotype:
hey Sascha, this is much better than the generic BaseTTXConverter that was used before for fvar table!
I guess you could also get rid of the fvar stuff which is inside otData.py, since it is no longer used here, no?

brawer:
@anthrotype: Thanks for noticing! Removed in https://github.com/behdad/fonttools/commit/5bd9ade3990ff5409ba5d955252ded6a793dc896.

behdad:
I agree that the existing code does not produce the prettiest objects.  In the long term I like to find a middle-ground, where we decompile using data, then transform into more usable objects.  For now, I'm merging this, since everything will be in the git history.

And Thanks!

PR #303, opened 2015-06-25
[ttProgram] add missing __len__ method to Program class

anthrotype:
Fixes https://github.com/behdad/fonttools/issues/301

### DIFF
--- a/Lib/fontTools/ttLib/tables/_f_p_g_m.py
+++ b/Lib/fontTools/ttLib/tables/_f_p_g_m.py
@@ -22,9 +22,6 @@ class table__f_p_g_m(DefaultTable.DefaultTable):
 		program.fromXML(name, attrs, content, ttFont)
 		self.program = program
 
-	def __len__(self):
-		return len(self.program)
-
 	def __bool__(self):
 		return hasattr(self, 'program') and bool(self.program)
 
--- a/Lib/fontTools/ttLib/tables/_f_p_g_m.py
+++ b/Lib/fontTools/ttLib/tables/_f_p_g_m.py
@@ -24,3 +24,8 @@ class table__f_p_g_m(DefaultTable.DefaultTable):
 
 	def __len__(self):
 		return len(self.program)
+
+	def __bool__(self):
+		return hasattr(self, 'program') and bool(self.program)
+
+	__nonzero__ = __bool__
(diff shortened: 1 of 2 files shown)


behdad:
I don't like that this transparently causes compilation of the program.  How about we get rid of it and implement __bool__/__nonzero__ only?

anthrotype:
yes, we could do that. Though it's no longer going to be `__len__`...

anthrotype:
I added `__bool__` and `__nonzero__` to both `ttProgram.Program` and to `_f_p_g_p.table__f_p_g_m`.
However, I left the `__len__` methods as well, in case one is interested in getting the actual length of the program, rather than its existence.
Please let me know if this is ok now.

anthrotype:
no wait.. one can't have zero-length tables.
The fpgm or prep must contain at least 1 instruction, or they are not valid tables.
I believe the `__bool__` approach is not needed. If one wants to know the actual length of the fpgm or prep programs, one calculates the length of the compiled bytecode.
Else, if one simply wants to know whether a fpgm or prep table are non-empty, then one simply needs to check `'fpgm' in font`. If True, then the table must be non-null.
Or do we want to allow writing zero-length fpgm and prep tables? (i.e. whose `program` contains an empty array, or whose `assembly` is an empty string). I don't know if that makes sense..

anthrotype:
so if we don't want to silently cause the compilation of the program when the user requests its length, then we need to remove `__len__` from fpgm table class, and leave ttProgram.Program as is.
We could leave the `__bool__` methods, but I think they are not very useful.

behdad:
The reason I don't like the __len__ on the program is that it can be very confusing, if, say, someone does this:

```
  for i in range(program):
    op = program.assembly[i]
```

Sure, that's not very Pythonic.  Still, I think it's safer to just not provide that __len__.

Re fpgm possibly being zero, I think that can happen.  I like it the way you have it right now.

behdad:
Yeah, removing len from fpgm and adding bool there sounds better to me.

anthrotype:
ok, I'll do that

behdad:
Thanks.

anthrotype:
OK, I removed __len__ from fpgm, and added __bool__ and __nonzero__ to both Program and fpgm.

behdad:
Sorry to keep adding more work for you.  But, trying to be more conscious about testing.  Care to add tests for ttProgram change?

anthrotype:
yes, definitely. A doctest should be enough for this.
I'll do that tomorrow.

behdad:
Thanks.

PR #304, opened 2015-06-25
Add list-chars-in-font.py snippet

ysangkok:
This snippet shows how to use fonttools to tell if a character is contained in a TTF font.

Requested for inclusion here: https://twitter.com/davelab6/status/608808066428100609

### DIFF
--- a/Snippets/list-chars-in-font.py
+++ b/Snippets/list-chars-in-font.py
@@ -1,0 +1,19 @@
+#!/usr/bin/env python
+from itertools import chain
+import sys
+
+from fontTools.ttLib import TTFont
+from fontTools.unicode import Unicode
+
+ttf = TTFont(sys.argv[1], 0, verbose=0, allowVID=0,
+                ignoreDecompileErrors=True,
+                fontNumber=-1)
+
+chars = chain.from_iterable([y + (Unicode[y[0]],) for y in x.cmap.items()] for x in ttf["cmap"].tables)
+print(list(chars))
+#Use this for just checking if the font contains the codepoint given as second argument:
+#char = int(sys.argv[2],0)
+#print(Unicode[char])
+#print(char in (x[0] for x in chars))
+
+ttf.close()


Snippets/list-chars-in-font.py
```diff
@@ -0,0 +1,19 @@
...
+import sys
+
+from fontTools.ttLib import TTFont
+from fontTools.unicode import Unicode
+
+ttf = TTFont(sys.argv[1], 0, verbose=0, allowVID=0,
+                ignoreDecompileErrors=True,
+                fontNumber=-1)
+
+chars = chain.from_iterable([y + (Unicode[y[0]],) for y in x.cmap.items()] for x in ttf["cmap"].tables)
```
anthrotype: since you are displaying the unicode character names, it'd be better to exclude non-Unicode cmap subtables, using the `isUnicode()` method: 
```python
chars = chain.from_iterable([y + (Unicode[y[0]],) for y in x.cmap.items()] for x in ttf["cmap"].tables if x.isUnicode())
```

Snippets/list-chars-in-font.py
```diff
@@ -0,0 +1,19 @@
...
+print(list(chars))
```
anthrotype: you could prettify the output by formatting it like a table, and maybe using hex digits for character codes instead of int
```
0x0021 exclam EXCLAMATION MARK
```

anthrotype: your `chars` chain is likely to include lots of duplicate entries. It's better if you do `sorted(set(chars))`. This way you get a sorted list of unique elements.

davelab6:
When I found this I ended up writing something similar and more simple, https://github.com/google/fonts/blob/master/tools/encodings/namelist.py

anthrotype:
@davelab6 your snippet is more readable (both the code and its output). Why do you exclude characters like "SPACE" or "NO-BREAK SPACE" from the list?
You're also not filtering duplicate entries referenced in more than one cmap subtable (but maybe that's exactly what you want).

anthrotype:
it'd be also nice to use the `setUnicodeData` function from `fontTools.unicode` module, to use a custom, and presumably more up-to-date "UnicodeData.txt" file... ok, now I'm going to far! :)

davelab6:
> your snippet is more readable (both the code and its output)

Thanks! I thought the use of itertools would be better, though

> Why do you exclude characters like "SPACE" or "NO-BREAK SPACE" from the list?

Its specific to the needs of the Google Fonts lists in that directory 

> You're also not filtering duplicate entries referenced in more than one cmap subtable (but maybe that's exactly what you want)

Probably a mistake! :)

behdad:
This:
```
chars = set()
chars.update(*(st.cmap.keys() for st in font['cmap'].tables if st.isUnicode()))
```

anthrotype:
:+1:

ysangkok:
Closing because @davelab6 snippet is superior.

PR #309, opened 2015-06-26
[GX] Code snippet for building interpolated fonts

brawer:


anthrotype:
:+1:

behdad:
As discussed, please change the weights to OpenType scheme (400, 700, ...) and add avar if nonlinearity is needed.

brawer:
Done in https://github.com/behdad/fonttools/commit/6cab1e6c1d2348089844a0f7e3716ffccca60504

It seems to work surprisingly well without an ‘avar’ table.

PR #316, opened 2015-07-03
[cffLib] Fix encoding of Notice and Copyright values in XML on Python 2

mashabow:
In `Notice` and `Copyright` values, Latin-1 characters such as '©' or 'ç' are replaced with U+FFFD on Python 2.7.9:

```xml
<ttFont sfntVersion="OTTO" ttLibVersion="2.5">

  <CFF>
    <CFFFont name="SabonNextLTPro-Regular">
      <version value="001.001"/>
      <Notice value="Copyright � 2002 - 2010 Linotype GmbH, www.linotype.com. All rights reserved. This font software may not be reproduced, modified, disclosed or transferred without the express written approval of Linotype GmbH. Sabon is a trademark of Linotype Corp. registered in the U.S. Patent and Trademark Office and may be registered in certain other jurisdictions in the name of Linotype Corp. or its licensee Linotype GmbH. This typeface is original artwork of Jean Fran�ois Porchez. The design may be protected in certain jurisdictions."/>
      <Copyright value="Copyright � 2002 - 2010 Linotype GmbH, www.linotype.com. All rights reserved. This font software may not be reproduced, modified, disclosed or transferred without the express written approval of Linotype GmbH. Sabon is a trademark of Linotype Corp. registered in the U.S. Patent and Trademark Office and may be registered in certain other jurisdictions in the name of Linotype Corp. or its licensee Linotype GmbH. This typeface is original artwork of Jean Fran�ois Porchez. The design may be protected in certain jurisdictions."/>
```
I think it should be as below, like on Python 3.4.0:

```xml
<ttFont sfntVersion="OTTO" ttLibVersion="2.5">

  <CFF>
    <CFFFont name="SabonNextLTPro-Regular">
      <version value="001.001"/>
      <Notice value="Copyright © 2002 - 2010 Linotype GmbH, www.linotype.com. All rights reserved. This font software may not be reproduced, modified, disclosed or transferred without the express written approval of Linotype GmbH. Sabon is a trademark of Linotype Corp. registered in the U.S. Patent and Trademark Office and may be registered in certain other jurisdictions in the name of Linotype Corp. or its licensee Linotype GmbH. This typeface is original artwork of Jean François Porchez. The design may be protected in certain jurisdictions."/>
      <Copyright value="Copyright © 2002 - 2010 Linotype GmbH, www.linotype.com. All rights reserved. This font software may not be reproduced, modified, disclosed or transferred without the express written approval of Linotype GmbH. Sabon is a trademark of Linotype Corp. registered in the U.S. Patent and Trademark Office and may be registered in certain other jurisdictions in the name of Linotype Corp. or its licensee Linotype GmbH. This typeface is original artwork of Jean François Porchez. The design may be protected in certain jurisdictions."/>
```

### DIFF
--- a/Lib/fontTools/misc/xmlWriter.py
+++ b/Lib/fontTools/misc/xmlWriter.py
@@ -142,7 +142,9 @@ class XMLWriter(object):
 			return ""
 		data = ""
 		for attr, value in attributes:
-			data = data + ' %s="%s"' % (attr, escapeattr(str(value)))
+			if not isinstance(value, (bytes, unicode)):
+				value = str(value)
+			data = data + ' %s="%s"' % (attr, escapeattr(value))
 		return data
 
 
--- a/Lib/fontTools/cffLib.py
+++ b/Lib/fontTools/cffLib.py
@@ -673,7 +673,7 @@ class ASCIIConverter(SimpleConverter):
 	def write(self, parent, value):
 		return tobytes(value, encoding='ascii')
 	def xmlWrite(self, xmlWriter, name, value, progress):
-		xmlWriter.simpletag(name, value=tostr(value, encoding="ascii"))
+		xmlWriter.simpletag(name, value=tounicode(value, encoding="ascii"))
 		xmlWriter.newline()
 	def xmlRead(self, name, attrs, content, parent):
 		return tobytes(attrs["value"], encoding=("ascii"))
@@ -684,7 +684,7 @@ class Latin1Converter(SimpleConverter):
 	def write(self, parent, value):
 		return tobytes(value, encoding='latin1')
 	def xmlWrite(self, xmlWriter, name, value, progress):
-		xmlWriter.simpletag(name, value=tostr(value, encoding="latin1"))
+		xmlWriter.simpletag(name, value=tounicode(value, encoding="latin1"))
 		xmlWriter.newline()
 	def xmlRead(self, name, attrs, content, parent):
 		return tobytes(attrs["value"], encoding=("latin1"))


Lib/fontTools/misc/xmlWriter.py
```diff
@@ -142,7 +142,9 @@ def stringifyattrs(self, *args, **kwargs):
 			return ""
 		data = ""
 		for attr, value in attributes:
-			data = data + ' %s="%s"' % (attr, escapeattr(str(value)))
+			if not isinstance(value, (bytes, unicode)):
```
brawer: Mind adding a test case to xmlWriter_test.py?

anthrotype:
it seems to be a Windows-only issue. At least, I wasn't able to replicate it neither on OS X nor on Linux.
When I tried to dump the CFF table of SabonNextLTPro-Regular like in your example, with fontTools checked at the latest HEAD, and using python 2.7.10 on OS X, or 2.7.6 on Ubuntu, the copyright characters were correctly encoded...

anthrotype:
Sorry, forget what I just wrote...
I was misled by Sublime Text which automatically opened my ttx file using "Windows 1252" encoding (it was too clever!). When I force Sublime to reopen the file with UTF-8 encoding, the bug shows up in OS X and Linxu as well.

Because xmlWriter only supports writing in UTF-8, the input it receives must be either utf-8 encoded bytestrings, or unicode literals (which are then encoded as utf-8 inside the `escape` function).

the problem is the cffLib module is sending latin-1 encoded bytestrings. Because the latter are already instances of `bytes`, the `tostr` function (alias of `py23.tobytes` when running in python 2) does not encode them to utf-8, but leaves them encoded as latin-1.

@mashabow patch correctly fixes this, by sending the xmlWriter unicode literals, and by skipping `str` coercion if the value is already an instance of bytes or unicodes.

Lib/fontTools/misc/xmlWriter.py
anthrotype: I'll do that

mashabow:
Thanks!

PS. I found that I also have a misunderstanding in my first comment:

> Latin-1 characters such as '©' or 'ç' are replaced with U+FFFD on Python 2.7.9

It is not true. xmlWriter writes Latin-1 characters as raw bytes (e.g. `0xA9` for '©'). Because they are invalid as UTF-8, of cource, then my editor (Atom) replaces them with U+FFFD.

twardoch:
Well, it's a tricky issue. The strings inside the CFF Font Info structures are "PostScript strings", and, strictly speaking, these strings should only contain ASCII characters. Non-ASCII character codes should be escaped via an octal code (`\ddd`), and, strictly speaking, the interpretation of those is undefined (i.e. it’s not specified anywhere which “encoding” they represent). Strictly speaking, raw bytes with values >127 are not permitted at all there. 

By convention, both the `\ddd` escaped codes and the (technically invalid) raw bytes with values >127 are best interpreted as ISO 8859-1 (ISO Latin 1). 

Since the first 256 codes of ISO Latin 1 are identical to Unicode, the `\ddd` octal code convention could probably be extended to cover all of Unicode (using `\dddd` etc.), but I haven’t tested it. 

Also, parentheses are dangerous. When the CFF is converted to a Type 1 (which happens in some printing environments), `)` is actually interpreted as a terminator for the string, and only “by luck” a pair of parentheses as in `(c)` is interpreted as part of text. If one makes a CFF where the Copyright string contains only a closing paranthesis, such fonts may not print on some devices (it’s really funny). 

However, parentheses can be escaped using `\`, i.e. `\(`, `\)` an `\\` are permitted in PostScript strings. 

The new FontLab codebase obeys the following guidelines within all Type 1 and CFF “Font Info” strings, such as Copyright or Notice:

### Reading

* `\\` should be interpreted as “\”
* `\(` should be interpreted as “\”
* `\)` should be interpreted “)”
* `\ddd` should be interpreted as octal numerical code in ISO 8859-1 encoding
* all other ASCII characters should be interpreted as is
* non-ASCII characters should be interpreted as ISO 8859-1
* when inside a string an unescaped `(` is found, it should be interpreted as “(”, and the matching closing `)` should also be interpreted as “)” rather than as a terminator for the string

### Writing

* Only ASCII characters are permitted
* The following characters must be escaped: “(” → `\(`, “)” → `\)`, “\” → `\\`
* The following characters should be unescaped:  `!"#$%&'*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[]^_`abcdefghijklmnopqrstuvwxyz{|}~`
* Three characters should undergo a special treatment: “©” → `\(C\)`, “®” → `\(R\)` and “™” → `\(TM\)`
* All other characters should be removed

twardoch:
Ps. It’s even more tedious. It seems that Adobe’s MakeOTF these days writes UTF-8 into the CFF Font Info fields. Which means, if I have `0xA9` in the CFF Notice field, a tool can interpret it as the ISO 8859-1 code for “©” but then another tool may write the UTF-8 representation of “©” into CFF, which is `0xC2 0xA9`. Then, the tool that thinks it’s ISO 8859-1 will interpret that as “Â©”.

brawer:
How about the following?

* when _reading_ such a string, check whether it is structurally valid UTF-8. If yes, decode as UTF-8; otherwise, decode as ISO-8859-1. Finally, unescape PostScript sequences such as `\(`.  
* when _writing_ such a string, first escape for PostScript (replace `(` by `\(`, etc.) Then, check whether it contains any characters above U+00FF. If yes, encode as UTF-8; otherwise, encode as ISO-8859-1.

I’ve seen that logic a few times in codecs for other ancient file formats, unrelated to fonts. This heuristic can obviously fail, but it works reasonably well in practice because `Â©` or `Â£` are extremely rare.

It’s pretty easy to check whether a byte sequence is structurally valid UTF-8; see [section 4 of RFC 3629](https://tools.ietf.org/html/rfc3629#section-4) for a grammar.

One complication to keep in mind: A lot of libraries claim to write UTF-8 but actually they produce an esoteric encoding called [http://www.unicode.org/reports/tr26/](CESU-8). There, codepoints above U+0FFFF are first encoded as UTF-16 surrogate pairs, and then the surrogates are sent separately to a naive UTF-8 encoder. The result is in explicit violation of the UTF-8 spec, but a lot of conversion code is doing it. When fonttools is going to decode UTF-8, it would be definitely worth testing how CESU-8 gets handled. (Probably Python’s UTF-8 codec will handle this transparently, which would be great; but I’d still test it).

anthrotype:
@twardoch thanks for your investigation!

> It seems that Adobe’s MakeOTF these days writes UTF-8 into the CFF Font Info fields.

Where did you get this information? I checked the [afdko source code](https://github.com/adobe-type-tools/afdko) and could not find any reference of MakeOTF or any other FDK module re-encoding input strings to UTF-8 when writing CFF Font Info fields, neither in the python scripts nor in the C source files.
Are you sure it isn't something else which is passing UTF-8 bytes to MakeOTF, and the latter simply writes them like that in the CFF Font Info fields?

I like the idea of adding support for PostScript escape sequences when parsing/writing strings from/to CFF binary table (of course, the TTX output should always be written as UTF-8).

@brawer to test whether the input is structurally valid UTF-8, wouldn't it be enough to see if it raises or not a `UnicodeDecodeError`?

twardoch:
True, it looks it’s not FDK after all. I've noticed that Glyphs writes Unicode characters into the CFF copyright string as UTF-8 while RoboFont filters all non-ASCII out. So it seems like the tools do different things themselves when preparing the data that is then passed to MakeOTF.

anthrotype:
If, like Adam says, the established convention is to interpret both non-ASCII characters in CFF copyright strings (either octal \ddd escaped codes or as raw bytes >127) as ISO 8859-1 (ISO Latin 1), and if Glyphs is the only editor which writes UTF-8 bytes in CFF, while others like Robofont or Fontlab filters out non-ASCII characters, wouldn't it be better to raise the issue with Glyphs app (cc @schriftgestalt), instead of allowing fontTools to also write such UTF-8 bytes in the binary CFF table?

by the way, ufo2fdk filters out not just  `()` parentheses, but also `[](){}<>/%` when normalising string for PostScript:
https://github.com/typesupply/ufo2fdk/blob/master/Lib/ufo2fdk/fontInfoData.py#L166

schriftgestalt:
I fixed it. Thanks for the notice. It should follow Adams recommendations.

anthrotype:
Thanks Georg! :+1:

PR #320, opened 2015-07-12
Write a font to stdout when compiling with "-o -"

mashabow:
Write a font to stdout when compiling with `-o -`.
Dumping XML to stdout is already supported (365b0bfce17d0eae765a6f432928c5a075871027).

### DIFF
--- a/Lib/fontTools/ttx.py
+++ b/Lib/fontTools/ttx.py
@@ -246,6 +246,7 @@ def ttCompile(input, output, options):
 			output = sys.stdout
 			if sys.platform == "win32":
 				# set binary flag on python2.x (Windows)
+				import platform
 				runtime = platform.python_implementation()
 				if runtime == "PyPy":
 					# the msvcrt trick doesn't work in pypy, so I use fdopen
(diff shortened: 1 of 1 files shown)


anthrotype:
I like the idea. However the standard I/O streams are opened in "text" mode by default, while you want to write binary data to stdout. You need to reopen it in "binary" mode, otherwise on Windows all `\n` characters in your binary data will be converted to `\r\n`, and you don't want that to happen.
Have a look [here](https://github.com/google/brotli/blob/master/python/bro.py#L24) for an example of how to open stdin/stdout/stderr in binary mode, which works across platforms (Unix and Windows) and implementations (both CPython and PyPy).

mashabow:
Thank you for your advice! Now I have tried to fix it, although unfortunately I have no Windows machine to test this code...

behdad:
Cosimo, what's the status of this?  Thanks.

anthrotype:
I'm working on a better solution that uses `io.open`. I'll send in a new PR as soon as it's ready.

behdad:
Sounds good.  Thanks!

PR #326, opened 2015-07-24
[xmlReader] open file for reading bytes (and revert two previous commits...)

anthrotype:
In commit 6b4567f, I submitted a patch to fix a `UnicodeDecodeError` raised by xmlReader on Windows Python 3 (https://github.com/behdad/fonttools/issues/323). I had xmlReader open files (still in "text" mode) with UTF-8 encoding, to prevent Windows from implicitly opening files using the locale cp1252 encoding.

However, the expat XML parser seems to require bytestrings as input. I you feed unicode strings, these will be implicitly converted to bytes using 'ascii' encoding. So in commit fe76598, I ended up re-encoding to UTF-8 the unicode strings as read from the file, before passing them to the parser.

I think it would be simpler and cleaner if we just open the file as `rb`, so we read in bytes and feed those directly to the expat parser. The latter will take care of applying the correct encoding (i.e. UTF-8) as declared in the xml header.

### DIFF
--- a/Lib/fontTools/misc/xmlReader.py
+++ b/Lib/fontTools/misc/xmlReader.py
@@ -26,7 +26,7 @@ class XMLReader(object):
 		if self.progress:
 			import stat
 			self.progress.set(0, os.stat(self.fileName)[stat.ST_SIZE] // 100 or 1)
-		file = open(self.fileName)
+		file = open(self.fileName, 'rb')
 		self._parseFile(file)
 		file.close()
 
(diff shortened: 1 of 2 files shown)


brawer:
Mind writing a test case? Perhaps it could create a temporary file, writing something like the following to it, then parse it with xmlReader.

    <doc>fôôbär</doc>

anthrotype:
@brawer I added a test case. Is that alright?

brawer:
Thanks, @anthrotype, for the test case! Merged.

PR #327, opened 2015-07-27
[xmlWriter] escape carriage returns '\r' using XML character entity '&13#;'

anthrotype:
by explicitly escaping any carriage returns `\r` that occurr within XML element tags, we make sure that TTX can roundtrip name records containing Windows-style line endings (`\r\n`).
If we don't escape carriage returns, these will get automatically normalised into Unix-style line endings `\n` by the expat XML parser.

See https://github.com/behdad/fonttools/issues/318.

### DIFF
--- a/Lib/fontTools/misc/xmlReader_test.py
+++ b/Lib/fontTools/misc/xmlReader_test.py
@@ -46,6 +46,36 @@ class TestXMLReader(unittest.TestCase):
 		content = strjoin(reader.contents[0]).strip() 
 		self.assertEqual(expected, content)
 
+	def test_normalise_newlines(self):
+
+		class DebugXMLReader(XMLReader):
+
+			def __init__(self, fileName, ttFont, progress=None, quiet=False):
+				super(DebugXMLReader, self).__init__(
+					fileName, ttFont, progress, quiet)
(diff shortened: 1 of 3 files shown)


brawer:
Isn't it a little surprising that this works? According to [section 2.11 of the XML specification](http://www.w3.org/TR/REC-xml/#sec-line-ends), XML parsers are actually required to convert Windows-style newlines to UNIX-style newlines before passing them to the application.

anthrotype:
in fact that's what is already happening in the xmlReader, via the expat parser (see https://github.com/behdad/fonttools/issues/318#issuecomment-119663239).

Any unescaped `\r` or `\r\n` are read in as `\n` by the XML parser.

In turn, fontTools' xmlWriter module writes platform-specific newlines (via `os.linesep`). This means a TTX file made on Windows will have `\r\n` newlines, instead of just `\n`.

The problem reported by @mashabow occurs in the character data content *within* XML elements (e.g. name records), not *between* elements, where the OS-specific newlines are written unescaped, and are correctly normalised while being read.

I could actually add a test case in the xmlReader that makes sure platform-specific newlines are correctly normalised -- of course, when not explicitly escaped.

Issue #328, opened 2015-07-31
traceback.print_exc(file) fails on Python 3 when file = BytesIO()


anthrotype:
whenever a table decompilation fails, for any reason, the normal behaviour is to retain the binary table data and fall back to the `DefaultTable`. The `traceback` module is used to print the error to a `StringIO` stream in order to save that message in a table attribute.

The problem is that in Python3, the `traceback.print_exc(file)` expects a stream of *unicode* text, but instead it's receiving a stream of *bytes*. This is because, when running fontTools under Python 3, the `StringIO` class, as imported from `fontTools.misc.py23`, is actually an alias of `io.BytesIO`.

This specific issue  could be easily fixed by simply importing the *true* `io.StringIO`, which represents an in-memory stream of unicode text, just before calling  `traceback.print_exc(file)`:

However, maybe we should consider modifying the `py23` bridge itself, so that all references to `StringIO` in the codebase will actually mean `io.BytesIO` on __both__ Python 2 and 3; and use the name `StringIO` (from `io` module) to mean a stream of unicode text on both 2 and 3. 

Here's the traceback of `traceback`: 
```
Traceback (most recent call last):
  File "/usr/local/var/pyenv/versions/fonttools3/bin/ttx", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/Users/cosimolupo/Documents/Github/fonttools/Tools/ttx", line 6, in <module>
    ttx.main(sys.argv[1:])
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/ttx.py", line 323, in main
    process(jobs, options)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/ttx.py", line 308, in process
    action(input, output, options)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/ttx.py", line 217, in ttDump
    bitmapGlyphDataFormat=options.bitmapGlyphDataFormat)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/ttLib/__init__.py", line 283, in saveXML
    self._tableToXML(tableWriter, tag, progress, quiet)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/ttLib/__init__.py", line 298, in _tableToXML
    table = self[tag]
[...]

PR #330, opened 2015-08-07
[py23] disambiguate bytes vs unicode in-memory streams

anthrotype:
This fixes issue #328, whereby an exception is raised when running `traceback.print_exc` function.

The error occurs because the `file` argument is expecting a stream of `str` objects: on Python 2, this must be a stream of *bytes*, whereas on Python 3 it must be a stream of *unicode* strings.

Currently we are using the name `StringIO` to mean an in-memory stream of bytes on both python versions. 

This patch redefines the import statements inside `py23` so that, on both 2 and 3:
- the name `BytesIO` will unambiguously mean a stream of bytes;
- the name `UnicodeIO` means a stream of unicodes.

Finally, the name `StringIO` is declined in a version-specific sense: that is, as a stream of **bytes** on Python 2 and a stream of **unicodes** on Python 3. 

So we give streams a similar treatment as the one reserved for strings themselves (cf. `tobytes`, `tostr` and `tounicode` functions).

Of course, I replaced all usages of `StringIO` in the codebase with `BytesIO` -- except where it means "a stream of 'str', whatever that means in the Python version I'm running".

I hope this makes sense.

### DIFF
--- a/Lib/fontTools/misc/psLib.py
+++ b/Lib/fontTools/misc/psLib.py
@@ -38,7 +38,7 @@ class PSTokenError(Exception): pass
 class PSError(Exception): pass
 
 
-class PSTokenizer(StringIO):
+class PSTokenizer(BytesIO):
 
 	def getnexttoken(self,
 			# localize some stuff, for performance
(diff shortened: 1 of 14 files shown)


behdad:
lgtm.  Please merge.  Thanks!

PR #332, opened 2015-08-09
fix "indentation contains mixed spaces and tabs" errors reported by Landscape

miguelsousa:


anthrotype:
I'm not 100% sure it's a good idea to "pep8ify" everything just to make Landscape.io happy..
Have you watched Raymond Hettinger's talk "Beyond PEP8"?
http://youtu.be/wf-BqAjZb8M

miguelsousa:
@anthrotype I did watched that, and I'm not trying to pep8ify or make Landscape happy. The goal was to help surface the real errors.

anthrotype:
Thank Miguel, I'll have a closer look then.
I was mainly concerned about possible merge conflicts in other open pull requests.

behdad:
pep8 thread is https://github.com/behdad/fonttools/pull/263.

I like this though.  Should be safe to merge even.

PR #334, opened 2015-08-11
Parser for OpenType feature files

brawer:
Currently, this is just a lexer and parser for OpenType feature files;
building actual tables will come as the next step. Also, a few constructs
still remain to be implemented.


anthrotype:
that's fantastic, thank you Sascha! :100:

davelab6:
:8ball: !

twardoch:
This is great! 
One thing that I should mention is that the AFDKO maintainers have agreed to change the FEA format as follows: 
To denote a glyph range with a hyphen, the hyphen must be surrounded by whitespace. Otherwise, the hyphen is interpreted as a part of the glyphname. 

So previously, the FEA syntax [a-z] and [a - z] both denoted a class made of a range of glyphs: [a b c d e ... x y z]. After the proposed change gets implemented, only [a - z] willdenote the range while [a-z] willstand for a one-glyph class with the glyph named "a-z". 

This is an effort to make the hyphen a first-class citizen at least in the font development stage (and sources). Glyphs uses hyphens in glyphnames and there is widespread agreement that they're a good idea.

davelab6:
+1 for that

miguelsousa:
@readroberts is this true?

>One thing that I should mention is that the AFDKO maintainers have agreed to change the FEA format as follows: 
To denote a glyph range with a hyphen, the hyphen must be surrounded by whitespace. Otherwise, the hyphen is interpreted as a part of the glyphname.

anthrotype:
@brawer I tried the lexer/parser on a few FEA files and it seems to working quite well.
I noticed that `pos` or `markClass` statements are not supported yet -- but I'm sure you're working on it already :)
Btw, do you have a list of constructs which haven't been implemented yet?
Keep up the good work!

brawer:
Yes, I’ll soon implement `pos` and `markClass` (and all the other remaining constructs). As the next step, I thought I’ll build some actual OpenType tables (GSUB) from the syntax tree, but I was hoping to collect some general feedback before the code grows too much.

readroberts:
Whoa, wait a minute! I have always been strongly against allowing a glyph name to contain a hyphen. This is not because of ranges in the feature file syntax, it is because there is a long history of using a hyphen to denote ranges in command line arguments that represent glyph lists. Most of the FDK tools will break if a glyph name contains a hyphen. If  you enter two glyph names separated by a hyphen surrounded by spaces, command line parsers will see that as three separate arguments. This is a strong argument against hyphens in glyph names. There are workflows in which being able to enter glyph ranges in command-line argument lists  is frequent and useful.

I have also never seen an argument for hyphens in glyph names, other than some people like them. I do understand that Georg Seifert likes to use them to add script and language suffixes, enough to maintain his separate branch of the makeotfexe tool that supports them, but I have never understood why another period clause wouldn't work as well.

davelab6:
Escape the spaces on the command line? :)

There's a LOT of sources out there with these glyph names due to Glyphs...

grzegorzrolek:
> There's a LOT of sources out there with these glyph names due to Glyphs…

Still, these are sources produced around Glyphs, not the FDK.

I’m totally with Read on this. Glyphs is an app that’s targeted for the kind of people who probably won’t even see a command line throughout their entire life. From that standpoint, Glyphs can implement whatever makes sense as far as those people are concerned, like allowing emojis as glyph names or scrapping the concept of a user-defined glyph name entirely: that, as long as it has some form of file export for a specific format or purpose that does the necessary normalisation. But FDK is a command line suite, thus should probably adhere to what’s most appropriate and useful in that specific environment.

twardoch:
Grzegorz, I think you're missing the point: for the most people, FDK is not a set of commandline tools. It's a set of libraries that are used by FontLab products, by RoboFont, by Glyphs and by its own commandline tools to compile feature definitions. 

Read, 

you say that various FDK tools would fail if glyph names contained hyphens. The UFO format does not impose any limitations on glyph names used in UFO (as they're not necessarily the same as PostScript glyph names). So in UFO, glyph names can be very flexible. 

I'm not advocating hyphens to be added to PostScript glyph names in the final fonts — only to the FEA compiler. Tools that produce final fonts should of course ensure that the fonal fonts don't have any non-compliant characters in the PostScript glyph names.

benkiel:
I agree with @readroberts and @grzegorzrolek. The Adobe FDK syntax has always said that hyphens are not allowed in glyph names (http://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#2.f.i). 

That Glyphs decided to implement behavior that was contrary to the spec isn't something for the FDK to solve, but for Glyphs to normalize and solve. @twardoch The UFO spec doesn't limit glyph names, but it does reference the Adobe FDK spec for how to handle fea code. Again, if an application wants to allow emoji or hyphens or whatever, it has to do the work to normalize to the FDK syntax.

twardoch:
OK, in that case I retract my request.

brawer:
Regarding hyphens in glyph names, this is currently not permitted according to [section 2.f.i of the OpenType Feature File Specification](http://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#2.f.i). Once the spec changes, it should be straightforward to modify the implementation in fonttools.

davelab6:
https://github.com/davelab6/fontqa reports glyph names with multiple periods as a problem. I don't know why.

twardoch:
@davelab6 The tests for FontQA were written with the point of view of FontFont's internal font production system. They had tighter internal specs than the wider industry standards. Also, lots of these tests are outdated.

brawer:
From @behdad over e-mail: “feel free to merge and continue on” — merging now

anthrotype:
:+1:

behdad:
Thanks Sascha!  Looks great.

readroberts:
I will change  makeotf to allow characters which are not permitted  in glyph names by the PostScript spec. Miguel Sousa and Frank Grießhammer discussed this with me, and I do see that there are advantages in using these in the development-mode glyph names.  I am still reluctant to support hyphen because the FDK has been in use for over 10 years, and allowing a hyphen in glyph names may break existing feature files. Is there any reason not to use another character? I currently plan to allow "+ * : ~ ^ !".

readroberts:
PS I do not plan to allow double-quote in a glyph name in makeotf.

davelab6:
Awesome!

twardoch:
Read: that's great news, as I have some plans for the characters you’ve listed :)

PR #338, opened 2015-08-19
change behaviour of reorderTables: allow True, None and False

anthrotype:
I believe the current behaviour of `reorderTables` argument in `TTFont.save` function is wrong, or at least is a bit misleading.
It's clear what `reorderTables=True` means: i.e. set the table order to the one recommended by the OpenType specification.
However, when one sets `reorderTables=False`, one may wish that the "original" input font's table order would be kept upon saving. However, that's not the case because the input font's table order as stored in the SFNTReader entries `offset` attribute is simply ignored. Actually, the final order is just random. It basically is whatever order fonttools finds more convenient when recompiling the tables: i.e. based on whether a table has been loaded, and whether it's compilation depends on some other table being compiled first.

OK, I know the table order is meaningless from a functional point of view. And of course, such an "original" table order would be even less meaningful if a font was loaded from XML, in which case there's no reader object to read the table offsets from.

But anyway, I would like to propose to slightly change the current behaviour of `reorderTables`, so that it accepts three instead of the current two values:

- `True`: the current default, reordering by OT spec recommended order.
- `None`: don't care about table order, use whatever order is faster and more convenient (this is equivalent to the current `False`). 
- `False`: try to keep the input font's table order if it was loaded from binary file, or else same as `None`.

For example, say I load a font and save it back to a new file with `reorderTables=False` without doing anything else in between. If the original sfnt directory was valid, and inter-table padding was correct, etc., I would expect the input and output fonts to be identical to each other.

```shell
>>> font = TTFont("font.ttf", recalcBBoxes=False, recalcTimestamp=False)
>>> font.save("font_new.ttf", reorderTables=False)
>>> exit()
$ diff font.ttf font_new.ttf
$ echo $?
0
```

WDYT?

### DIFF
--- a/Lib/fontTools/ttLib/__init__.py
+++ b/Lib/fontTools/ttLib/__init__.py
@@ -212,11 +212,8 @@ class TTFont(object):
 		if "GlyphOrder" in tags:
 			tags.remove("GlyphOrder")
 		numTables = len(tags)
-		if reorderTables:
-			import tempfile
-			tmp = tempfile.TemporaryFile(prefix="ttx-fonttools")
-		else:
-			tmp = file
+		# write to a temporary stream to allow saving to unseekable streams
+		tmp = BytesIO()
 		writer = sfnt.SFNTWriter(tmp, numTables, self.sfntVersion, self.flavor, self.flavorData)
 
 		done = []
@@ -225,11 +222,25 @@ class TTFont(object):
 
 		writer.close()
 
-		if reorderTables:
+		if (reorderTables is None or
+				(reorderTables is False and not hasattr(self, 'reader'))):
+			# don't reorder tables and save as is
+			file.write(tmp.getvalue())
+			tmp.close()
+		else:
+			if reorderTables is False:
+				# sort tables using the original font's order
+				tableOrder = self.reader.keys()
+			else:
+				# use the recommended order from the OpenType specification
+				tableOrder = None
 			tmp.flush()
 			tmp.seek(0)
-			reorderFontTables(tmp, file)
+			tmp2 = BytesIO()
+			reorderFontTables(tmp, tmp2, tableOrder)
+			file.write(tmp2.getvalue())
 			tmp.close()
+			tmp2.close()
 
 		if closeStream:
 			file.close()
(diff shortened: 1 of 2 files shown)


behdad:
SGTM.  How about using an OrderedDict?

anthrotype:
I forgot to mention another thing about this patch, which is the ability to save a font to an unseekable stream such as stdout.
The SFNTWriter needs to seek back and forth. So I just pass the writer a BytesIO stream instead of the output file, and write to the latter only once at the end.
I should have probably split this in the two separate commits..

behdad:
Yeah, just verified what you are doing there.  Looks good.

anthrotype:
do you mean, the TTFont.tables dictionary or the SFNTReader and SFNTWriter tables dict?
Changing the former would produce too many changes. Yes, the latter could be turned into OrderedDicts if you like (then I should also change WOFF2Reader and WOFF2Writer for consistency)

anthrotype:
since I'm changing the `save` function, can I also get rid of the `makeSuitecase` argument?
Remember we have removed the `SFNTResourceWriter` class from the `macUtils` module, as it's no longer supported on OSX.

behdad:
> do you mean, the TTFont.tables dictionary or the SFNTReader and SFNTWriter tables dict?
> Changing the former would produce too many changes. Yes, the latter could be turned into OrderedDicts if you like (then I should also change WOFF2Reader and WOFF2Writer for consistency)

I meant the latter.  It looks rather redundant to me to keep tables and table_order separately, whereas OrderedDict is exactly the combination of those two.  Use your judgement though, I don't have a strong opinion here.

behdad:
> since I'm changing the save function, can I also get rid of the makeSuitecase argument?

Don't touch in the same commit definitely.  Also, who knows which downstream library is sending us makeSuitecase=False right now.  Lets not break anyone unnecessarily...

anthrotype:
ok, i'll do it separately. However, I believe we should at least `raise NotImplementedError`

anthrotype:
OK, I now use OrderedDict in sfnt now and get the sorted `keys()`.

behdad:
Thanks!

PR #341, opened 2015-08-20
[travis] compile Brotli using `build_ext --libraries=stdc++` on pypy3

anthrotype:
I use pip `--global-option` to run distutils `build_ext` command with a custom `--libraries=stdc++` argument when compiling brotli for pypy3.
This should fix the issue with importing the brotli module in pypy3, which in turn makes it skip the woff2 tests: https://github.com/behdad/fonttools/issues/339
Finger crossed...


anthrotype:
it worked!

PR #342, opened 2015-08-22
Add roboto_name_fix script to snippets.

codeman38:
@davelab6 suggested that I contribute this script to the fonttools repo as an example snippet.

This script fixes the name tables in the Roboto font binaries from https://github.com/google/roboto to be compatible with Windows, so that styles outside the base 4 are split off into separate families. The preferred family name is left intact for OSes that support it.

Original source: https://gist.github.com/codeman38/9d77f650a875639a55e6

### DIFF
--- a/Snippets/roboto_name_fix.py
+++ b/Snippets/roboto_name_fix.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python2.7
+
+"""Update the 'name' table in the extra Roboto weights to be
+more compatible with Windows' font chooser."""
+__author__="codeman38"
+
+import argparse
+import glob
+import os
+from fontTools import ttLib
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Update the 'name' table in the extra Roboto weights "
+                    "to be more compatible with Windows' font chooser.")
+    parser.add_argument('filename', nargs='+',
+                        help='file name(s) of fonts to convert')
+    args = parser.parse_args()
+    for filespec in args.filename:
+        for fname in glob.glob(filespec):
+            process_font(fname)
+
+def process_font(fname):
+    outname = fname+'.tmp'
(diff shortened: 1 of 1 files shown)


behdad:
cc @twardoch @brawer 
Adam, Dave, is there something more universal / generic to make out of this?  Eg, something to check the various font names are "consistent" and fix stuff?

davelab6:
Reading the code, I actually am not sure how you want this to be more universal/generic; the code isn't tied to Roboto fonts, and while it doesn't check other things about the NAME table that could be checked, wouldn't scripts to check those things be other snippets, or even part of another project (like adam's sfntqa)

behdad:
I'm fine if Cosimo wants to merge this. ;)

brawer:
@anthrotype, what do you think?

anthrotype:
the short answer is "I don't like it, but I don't really care".

anthrotype:
I don't see the point of this script. It simply appends to the family nane (nameID=1) the first word from the subfamily (nameID=2), and forces the latter to be either "Regular" or "Italic" (if the last word of subfamily is "Italic").
I don't think this approach can be generalised, so I suggest we close this.
BTW, the issue with the legacy subfamily has already been fixed in the current Roboto.

davelab6:
> I don't see the point of this script. 

The point of contributing it here is to show 'real world' examples of scripts that use fontTools to 'hotfix' font binaries. 

> It simply appends to the family name (nameID=1) the first word from the subfamily (nameID=2), and forces the latter to be either "Regular" or "Italic" (if the last word of subfamily is "Italic").

But it also shows how to read and write fonts, use argparse, etc. These are novice-level basics, but the type design community are overall novice-level. 

If its rejected, that's fine; but I would then like clarification about what should be submitted to Snippets, because I thought this was a good fit :)

PR #346, opened 2015-08-31
[snippets] add gbk codepoint conversion in convert cmap format process

bihicheng:
add gbk codepoint conversion in convert cmap format process


behdad:
Humm.  Where does the module gbk come from?

bihicheng:
from me. :)

bihicheng:
come from a lua script.

behdad:
So, Python codecs didn't have that?

bihicheng:
I tried the python codecs first, but it have not method to do that,  then I use this way now.

davelab6:
@behdad what additional steps are needed to merge this, or should it be closed?

behdad:
I'm not sure if this is suitable and don't have to check it more.

PR #349, opened 2015-09-04
Support non-BMP characters for synthetic glyph names

brawer:
When a font supplies no glyph names in its 'post' table, fontTools
builds synthetic glyph names by reversing the 'cmap' table.
After this change, the library looks at all 'cmap' subtables for
Unicode, irrespective of format or platform. For example, glyph #4
in NotoSansOldItalic-Regular.ttf gets now named "u10300" instead of
"glyph00004".

Moved the code for building a reversed 'cmap' table into the cmap class,
for easier testing.


anthrotype:
the travis test passed. this can be merged, no?

brawer:
OK, merging.

PR #351, opened 2015-09-09
add 'padding' attribute to glyf table

anthrotype:


anthrotype:
The attribute is required by WOFF2, and controls the amount of padding applied to glyph data upon compile. Glyph lengths are aligned to multiples of the specified value.
Allowed values are (0, 1, 2, 4). '0' means no padding; '1' (default) similarly means no padding, except for when padding would allow the use of short loca offsets.

PR #352, opened 2015-09-09
[ttx] add --flavor option to compile to "woff" or "woff2"

anthrotype:
This adds a `--flavor` option similar to the one available in pyftsubset, which takes either "woff" or "woff2".
I also made the handling of the argument parser errors more consistent.


anthrotype:
(I removed the commit about GetoptError handling, it's better to open a separate pull request for that one)

PR #356, opened 2015-09-09
[feaLib] make LookupList.Lookup into a list of Lookup objects, not subtables

anthrotype:
the `feaLib.builder.Builder.makeTable` is missing one level in the structure of OpenType layout tables.

`table.LookupList.Lookup` in GSUB is being built as a list of lookup *subtables* (e.g. `LigatureSubst`, `AlternateSubst`, etc.). Instead, it should be a list of `Lookup` objects, each having a `SubTable` attribute containing the list of subtables, as well as `LookupType`, `LookupFlag`, `SubTableCount`, etc.

https://github.com/behdad/fonttools/blob/master/Lib/fontTools/ttLib/tables/otData.py#L88

@brawer this PR is just to show the issue with code, feel free to close this and fix it the way you like.
Thanks for the good work!

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -299,9 +299,15 @@ class AlternateSubstBuilder(LookupBuilder):
                 self.alternates == other.alternates)
 
     def build(self):
-        lookup = otTables.AlternateSubst()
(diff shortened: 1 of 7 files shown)


brawer:
:+1:

PR #366, opened 2015-09-10
[README] reintroduce a "History" paragraph

anthrotype:
for some reasons, this was removed in https://github.com/behdad/fonttools/pull/359

/cc @davelab6 @miguelsousa @benkiel


davelab6:
I recommend that markdown files have one line per sentence, it makes cleaner diffs and doesn't change visible output

PR #368, opened 2015-09-11
[otTables] Use XML snippets for testing fromXML() methods

brawer:
This should make the unit tests more readable, and it also prevents
failures where a changed fromXML() implementation fails to ignore
interspersed whitespace. This has recently caused some developer
nuisance; see https://github.com/behdad/fonttools/pull/367.

Instead of having a custom parser implementation, it would be nicer
to call the actual XMLReader. However, XMLReader has a deeply built-in
assumption that it is processing an entire TTX file. Changing this
assumption would certainly be possible, but it would need a re-write
of the XMLReader code; having a custom parser just for unit tests
seems less involved.

### DIFF
--- a/Lib/fontTools/misc/testTools.py
+++ b/Lib/fontTools/misc/testTools.py
@@ -1,0 +1,42 @@
+"""Helpers for writing unit tests."""
+
+from __future__ import print_function, division, absolute_import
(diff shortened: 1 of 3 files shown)


anthrotype:
this is very handy, thanks Sascha!

PR #371, opened 2015-09-14
[Travis] also test against newly released Python 3.5.0

anthrotype:
Python 3.5.0 was released a few days ago, and Travis already added support for it.
Let's see if it works...


anthrotype:
it works!

PR #374, opened 2015-09-18
Documentation tweaks

davelab6:
I updated the documentation a little more, merging contents from the recently removed documentation.html file and the install.txt file (deleted here) into the README, and hand drafted a changelog file for the last 2 releases, moved into a more typical filename.


davelab6:
Thanks for the review, I agree with all comments - pushed an update.

anthrotype:
is this ready to be merged? LGTM.

behdad:
NEWS and README look better to me, but not going to block landing this.

behdad:
> is this ready to be merged? LGTM.

Please proceed as you see fit.

anthrotype:
I was thinking that it might be time to tag a new 3.1 release and push that to PyPI.
Or do we want to wait for, say, feaLib to be completed first?

behdad:
> Or do we want to wait for, say, feaLib to be completed first?

Sascha and I are working on that this week.  So, by next week we should have much better story there; let's wait a week.

anthrotype:
cool! looking forward to that :)

README.md
```diff
@@ -21,7 +34,11 @@ cd fonttools ;
 python setup.py install ;
 ```
 
-Full build and installation instructions are in [Doc/install.txt](https://github.com/behdad/fonttools/blob/master/Doc/install.txt) 
+This will install all the modules in the right places. 
+On Unix it also installs the command line tools.
```
anthrotype: I removed the "Windows" folder as it was outdated (cf. https://github.com/behdad/fonttools/pull/423). We now rely on setuptools' `console_scripts` feature to create the platform-specific executable scripts. This is done automatically by the setup.py script upon install, so there's no need to differentiate Unix vs Windows in the installation instructions.

@davelab6  Could you please remove lines 40-41, and change line 37-38 so that it simply says something like:

```diff
-This will install all the modules in the right places. 
-On Unix it also installs the command line tools. 
+This will install all the modules and command line tools in the right places.
```

davelab6:
> NEWS and README look better to me, but not going to block landing this.

I've moved CHANGELOG.txt to NEWS, but README.md needs the extension to render as markdown on https://github.com/behdad/fonttools/ 

LICENSE.txt however can drop the extension so I've done that also, hope its ok

davelab6:
> I was thinking that it might be time to tag a new 3.1 release and push that to PyPI.

I think this can be merged before the 3.1 tag?

anthrotype:
the problem now is you have an additional merge commit 41f65f73af5a1d4e6417a9f37b776f54b2a05428
would it be possible for you to rebase and remove that one?
After that I'll merge it.

davelab6:
I think I did it wrong...?

anthrotype:
I'll cherry-pick and merge this manually. Next time, it's better you don't pull requests from your fork's master branch, but make a branch first so that your master stays clean.

davelab6:
Ah right that's it. Yep, noted. Thanks!

khaledhosny:
`git pull -r` should have resolved the merge commits, BTW (or even `git rebase origin/master`, though I’m not very sure about this one).

PR #378, opened 2015-09-24
[ttProgram] if can't disassemble bytecode, warn and dump toXML as it is

anthrotype:
When dumping glyf table to XML with `disassembleInstructions`, if an exception is raised while decompiling one of its glyph programs, the whole process aborts.

With this patch, the disassembling will continue despite the exception, and the faulty glyph's bytecode will be dumped as hex string instead of assembly, preceded by the full exception traceback as an XML comment.

Fixes https://github.com/behdad/fonttools/issues/347

### DIFF
--- a/Lib/fontTools/ttLib/tables/ttProgram.py
+++ b/Lib/fontTools/ttLib/tables/ttProgram.py
@@ -225,31 +225,46 @@ class Program(object):
 
 	def toXML(self, writer, ttFont):
 		if not hasattr (ttFont, "disassembleInstructions") or ttFont.disassembleInstructions:
-			assembly = self.getAssembly()
-			writer.begintag("assembly")
-			writer.newline()
-			i = 0
-			nInstr = len(assembly)
-			while i < nInstr:
-				instr = assembly[i]
-				writer.write(instr)
+			try:
+				assembly = self.getAssembly()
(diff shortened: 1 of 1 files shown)


anthrotype:
@cary-clark
Here's for example how glyphID 492 from kochi-gothic.ttf would look like after this:

```xml
      <instructions><bytecode>
          <!-- An exception occurred during the decompilation of glyph program:
               
                 File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/ttLib/tables/ttProgram.py", line 229, in toXML
                   assembly = self.getAssembly()
                 File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/ttLib/tables/ttProgram.py", line 223, in getAssembly
                   self._disassemble(preserve=preserve)
                 File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/ttLib/tables/ttProgram.py", line 432, in _disassemble
                   value = bytecode[i]
               IndexError: array index out of range -->
          33002100 1e00e1c0 40804180 42804480
          58806080 4080e1c0 2b3031 
        </bytecode></instructions>
```

cary-clark:
Great improvement. Thank you.

PR #382, opened 2015-10-01
[name] add setName method to table__n_a_m_e

anthrotype:
The method takes a unicode `string`, `nameID`, `platformID`, `platEncID` and `langID`.
If the given name record already exists, its string is modified in place.
If it doesn't exist, a new record is created with the specified IDs and string.

As suggest by @brawer in https://github.com/anthrotype/fonttools/commit/380104c2288200aec1a7173162a3b3cfb29fc8f2#commitcomment-13534159

### DIFF
--- a/Lib/fontTools/ttLib/tables/_n_a_m_e.py
+++ b/Lib/fontTools/ttLib/tables/_n_a_m_e.py
@@ -111,6 +111,23 @@ class table__n_a_m_e(DefaultTable.DefaultTable):
 		else:
 			return None
 
+	def setName(self, string, nameID, platformID, platEncID, langID):
+		if not hasattr(self, 'names'):
+			self.names = []
+		namerecord = self.getName(nameID, platformID, platEncID, langID)
+		exists = False if namerecord is None else True
+		if not exists:
+			namerecord = NameRecord()
+			namerecord.nameID = nameID
+			namerecord.platformID = platformID
+			namerecord.platEncID = platEncID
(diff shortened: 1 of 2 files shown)


anthrotype:
Hmm, this is interesting..
The Travis build completed successfully on Python 3, but failed on Python 2.7 and pypy with a mysterious `UnicodeEncodeError` when trying to convert a name record containing some Chinese to `str`.

As you know, the `NameRecord` class has a convenient `__str__` method which calls `toUnicode()`, else returns the raw `.string` when it fails.  

Now, in my `_n_a_m_e_test`, the following lines...

```python
table.setName("寬鬆", 276, 1, 2, 0x13)
str(table.getName(276, 1, 2, 0x13))
```

... would throw the following exception:

`UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-1: ordinal not in range(128)`

Note that it's not a `UnicodeDecodeError` as one would expect in this case.
Besides, If I call `isUnicode()` directly -- or even `namerecord.__str__()` --, then the mac traditional chinese string in the example is decoded just fine!

So it looks like python2's built-in `str` function is complaining about the fact it's receiving a unicode string instead of a bytestring, hence it tries to re-encode it using the default `ascii` encoding -- and fails.

Is this `str(namerecord)` approach being used anywhere else in the code?
While we find a workaround, it's better that we use `toUnicode()` instead, as it works on both py2 and 3.

PR #383, opened 2015-10-02
Add pen which draws to TrueType glyphs

jamesgk:


behdad:
lgtm other than the comment above.  Can you also add the cffGlyphPen?

anthrotype:
It looks good to me. Maybe later we can also add some tests.
Thank you James! :+1:

anthrotype:
@jamesgk Can we merge this now or you need to add more changes?
It looks fine to me.

jamesgk:
Yea, I think this is fine. We'll do the cffGlyphPen in another PR.

anthrotype:
Thanks a lot for your work!

behdad:
> Handle None here? See docstring comments for qCurveTo in basePen. In particular this part:
> 
> ↦       ↦       The last argument (normally the on-curve point) may be None. 
> ↦       ↦       This is to support contours that have NO on-curve points (a rarely 
> ↦       ↦       seen feature of TrueType outlines).

Please address this and add tests.

anthrotype:
@behdad isn't this OK? 
https://github.com/jamesgk/fonttools/commit/b916c5bb935da2d1a55fa0706cb372ec3e54436e

behdad:
> @behdad isn't this OK?
> jamesgk@b916c5b

Oh sorry I didn't see that.  Thanks!

We should add tests for this pen.

anthrotype:
yes, we should.

PR #392, opened 2015-10-21
[psLib] revert to using StringIO.StringIO; temporarily drop support for Python3

anthrotype:
This is only a temporary fix to revert a change which broke support for parsing Type1 fonts in Robofont. It turns out the whole t1Lib and psLib are not working with python3 because of other issues (as usual, unicode vs bytes), so they will need more than simply renaming StringIO to BytesIO...


anthrotype:
(Travis failed because I too hastily deleted the remote branch so it could not fetch it.. Never mind, there weren't any additional tests to run -- but we should eventually make them...)

PR #394, opened 2015-10-22
make t1Lib and psLib compatible with python 3

anthrotype:
I made the `psLib.PSTokenizer` compatible with Python3 by making it independent from Python2-only `StringIO.StringIO`.

I added two test PostScript Type1 fonts and some unittest for testing the basic reading/writing functions. It's only .pfa and .pfb formats, since LWFN is a Mac resource fork and git doesn't support that (we shall deal with those separately, one day).

To keep the testdata and unit tests closer to the t1Lib module, I moved the latter into its own sub-package, but without changing the interface.

Finally, I added a new `dohex` keyword argument to the `saveAs` method of `T1Font` class, so that it can encode the encrypted part as an hex string when writing PFA fonts.

About that, I noticed an error in the docstring, where it says that the 'dohex' argument of the global function `t1Lib.write` only applies to LWFN and PFB. It is actually the case that it only applies to the category 'OTHER' (ie. 'PFA'). So I fixed that as well.

The rest is py23 house-cleaning.

Let me know what you think.
Thanks

C.


behdad:
lgtm!

PR #396, opened 2015-10-23
Add pen which draws to Type 2 charstrings

jamesgk:


anthrotype:
super! :+1: 
thanks a lot, James.
Now we just need to write unit tests for both the TrueType and CFF pens.

anthrotype:
Oh, I just realised this is the same @typesupply's T2CharStrings pen from ufo2fdk:
https://github.com/typesupply/ufo2fdk/blob/master/Lib/ufo2fdk/pens/t2CharStringPen.py
I must assume the move was consensual.
Thanks Tal! :)

Lib/fontTools/pens/t2CharStringPen.py
```diff
@@ -0,0 +1,131 @@
+# Copyright (c) 2009 Type Supply LLC
+# Author: Tal Lemming
```
adrientetar: There are no double-m's.

Lib/fontTools/pens/t2CharStringPen.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+from __future__ import print_function, division, absolute_import
```
adrientetar: As you import `fontTools.misc.py23` now I think it does the future imports by itself.

anthrotype: no, you always need to repeat from `__future__ import ...`

anthrotype:
@jamesgk LGTM overall, though it'd be also nice to have some unit tests for this and the `TTGlyphPen` .
For example, one could use the pen to "re-draw" a few glyphs as are returned by the `TTFont.getGlyphSet()` method, and then assert that the original glyph and the one returned by the pen are identical.
There are test fonts in `ttLib/testdata`, maybe we could use those.

jamesgk:
OK, want to open a new PR for tests or just add them to this one?

anthrotype:
I'd say we merge this one as is, and then we open a new PR to test both the CFF and TT pen.
Would you like to write the tests?

jamesgk:
Sure thing!

anthrotype:
:+1:

Issue #395, opened 2015-10-22
Can't subset Google-style color fonts


rsheeter:
The subsetter currently always drops CBDT/CLBC (for Google Color fontshttp://google-opensource.blogspot.com/2013/05/open-standard-color-font-fun-for.html). If this is changed (based on discussion w/Behdad) as follows:

subset.py

```
# CBDT will inherit it
@_add_method(ttLib.getTableClass('EBDT'))
def subset_glyphs(self, s):
  self.strikeData = [_dict_subset(strike, s.glyphs) for strike in self.strikeData]
  return True

# add CBLC to _no_subset_tables_default
# remove 'CBLC', 'CBDT' from _drop_tables_default
```

And we then try to subset Noto Color Emoji, say with args `NotoColorEmoji-Regular.ttf --gids=2 --verbose` it fails, saying .notdef is missing:

```
Text: ''
Unicodes: []
Glyphs: []
Gids: [2]
maxp pruned
cmap pruned
name pruned
post pruned
GSUB pruned
Added .notdef to subset
Closing glyph list over 'GSUB': 2 glyphs before
Glyph names: ['.notdef', 'zero']
Traceback (most recent call last):
  File "subset.py", line 2746, in <module>
    main()
  File "subset.py", line 2715, in main
    subsetter.subset(font)
  File "subset.py", line 2496, in subset
    self._closure_glyphs(font)
  File "subset.py", line 2424, in _closure_glyphs
    self.log.glyphs(self.glyphs, font=font)
  File "subset.py", line 2536, in glyphs
    self("Glyph IDs:    ", sorted(reverseGlyphMap[g] for g in glyphs))
  File "subset.py", line 2536, in <genexpr>
    self("Glyph IDs:    ", sorted(reverseGlyphMap[g] for g in glyphs))
KeyError: '.notdef'
```

Perhaps we don't care about notdef and try with `NotoColorEmoji-Regular.ttf --gids=2 --verbose --no-notdef_glyph`. Sadly still unhappy:

```
Text: ''
Unicodes: []
Glyphs: []
Gids: [2]
maxp pruned
cmap pruned
name pruned
post pruned
GSUB pruned
Closing glyph list over 'GSUB': 1 glyphs before
Glyph names: ['zero']
Glyph IDs:     [2]
Closed glyph list over 'GSUB': 1 glyphs after
Glyph names: ['zero']
Glyph IDs:     [2]
Retaining 1 glyphs: 
head subsetting not needed
[...]

PR #397, opened 2015-10-23
Experimental support for Google-style color bitmap fonts

rsheeter:
Copying sample code manually into #395 is kind of annoying, creating PR instead.


Lib/fontTools/subset.py
```diff
@@ -1254,6 +1254,23 @@ def collect_features(self):
     return _uniq_sort(sum((s.Script.collect_features()
                            for s in self.ScriptRecord), []))
 
+# CBLC will inherit it
+@_add_method(ttLib.getTableClass('EBLC'))
+def subset_glyphs(self, s):
+  for strike in self.strikes:
+    for indexSubTable in strike.indexSubTables:
+      indexSubTable.names = [n for n in indexSubTable.names if n not in s.glyphs]
```
anthrotype: @rsheeter isn't there a contradiction here? `s.glyphs` is meant to be the list of glyph names which one wants to *retain*. So here you would want to keep all the names which are in `s.glyphs`, and not those which are not...

```diff
-indexSubTable.names = [n for n in indexSubTable.names if n not in s.glyphs]
+indexSubTable.names = [n for n in indexSubTable.names if n in s.glyphs]
```

Lib/fontTools/subset.py
```diff
@@ -1254,6 +1254,23 @@ def collect_features(self):
...
+    strike.indexSubTables = [i for i in strike.indexSubTables if i.names]
+  self.strikes = [s for s in self.strikes if s.indexSubTables]
+
+  return True
+
+# CBDC will inherit it
+@_add_method(ttLib.getTableClass('EBDT'))
+def subset_glyphs(self, s):
+  self.strikeData = [_dict_subset(strike, s.glyphs) for strike in self.strikeData]
```
anthrotype: this raises `KeyError` whenever a glyph name is in the list of glyphs to retain `s.glyphs` but is not also among the keys of the `strike` dictionary.

For example, in the example color fonts from:
https://github.com/googlei18n/color-emoji/tree/master/specification/examples

the glyph '.notdef' is not contained in the CBDT strikeData, but is contained in the post table, so the `_dict_subset` function fails with KeyError.

I think you should subset the CBDT table's strikeData like that:

```python
self.strikeData = [{g: strike[g] for g in s.glyphs if g in strike}
                   for strike in self.strikeData]
```

anthrotype:
@rsheeter don't know why github is saying the your pull request branch has merge conflicts (i tried merging locally and it worked..)
could you maybe try rebasing on behdad/master?

anthrotype:
I guess it's because the subset.py module has been renamed to subset/__init__.py..
Also, before we merge this PR, it'd be nice if we add some unit tests, based on the ones @miguelsousa wrote in https://github.com/behdad/fonttools/pull/411.

rsheeter:
Rebased & added a test.

anthrotype:
excellent, thanks! :+1:

PR #400, opened 2015-10-26
add mac resource fork reader (py23 compatible); drop mac classic support

anthrotype:
I added a new `fontTools.misc.macRes` module with a `ResourceReader` class which takes in a file object or path, reads data from a Mac resource file and builds a dictionary of resources keyed by type.

On HFS+ filesystem, I try to read from a file's `..namedfork/rsrc`, and if that fails I then read the data fork. Reading data-fork resources works on Windows and Linux as well -- as well as on Python 3.

I dropped the `Carbon.Res` module and other old stuff which has been deprecated and is no longer used in Python 3.

In `macCreatorType`, I added support for `xattr`, a Python wrapper for extended filesystem attributes which -- unlike `MacOS` module -- works with Python 3 as well. It comes pre-installed on OS X's Python, and can be easily installed via `pip install xattr`, if required. 

For reading 'sfnt' resources from suitcase or DFont, I've kept the old `ttLib.macUtils` interface, at least for now -- until we define a TTCollection interface.

The tests only cover the generic ResourceReader, as I didn't know how to create custom `.suit` or `.dfont` fonts... but we can add these later.

@brawer, would you like to have a look when you have some time?
Of course, I'm greatly indebted to your Java implementation of a Mac resource reader ;)

Cheers!

C.

Related issues:
https://github.com/behdad/fonttools/issues/229#issuecomment-150890828
https://github.com/behdad/fonttools/issues/228
https://github.com/behdad/fonttools/issues/221


brawer:
Oooh! :+1: 

Just one little thing: Try decoding resource names as mac-roman. On a few Macs, the system encoding was different, but those were quite rare.

anthrotype:
Ok, thanks! Shall I do `tostr` or `tounicode`? I guess the former is fine for this purpose.

anthrotype:
:+1:

PR #401, opened 2015-11-06
TTGlyphPen tests

jamesgk:
This adds tests with the existing TTX test data, and some simple unit tests. It also fixes issues with the pen itself! Tests are nice.

### DIFF
--- a/Lib/fontTools/pens/transformPen.py
+++ b/Lib/fontTools/pens/transformPen.py
@@ -51,6 +51,9 @@ class TransformPen(AbstractPen):
 	def closePath(self):
 		self._outPen.closePath()
 
+	def endPath(self):
+		self._outPen.endPath()
+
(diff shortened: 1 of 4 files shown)


anthrotype:
Excellent!! :+1: 

just one thing:
https://github.com/jamesgk/fonttools/blob/pen-tests/Lib/fontTools/pens/ttGlyphPen.py#L114

```diff
-glyph.program.fromBytecode("")
+glyph.program.fromBytecode(b"")
```

anthrotype:
All looks good to me, except for that `TypeError` spotted by Travis when initialising `array.array` with an empty `str` -- it should be `bytes` for Python3.

Thanks!

anthrotype:
Can I merge now or do you have other commits to push?

jamesgk:
Yea, this is done.

PR #402, opened 2015-11-07
[subset] Support subsetting MS color tables

khaledhosny:
Add support for subsetting COLR table.

The CPAL table does not need subsetting, but unused palettes should be
pruned, this is not implemented however as it depends on COLR table,
which in turn will not be updated after pruning CPAL.


khaledhosny:
One question, should `COLR` and `CPAL` be kept in the dropped by default tables list or removed from it?

khaledhosny:
I rebased the PR and added a test as well.

khaledhosny:
I also removed `COLR` and `CPAL` from the list of tables dropped by default since we can now subset them and if they are present in the font then they should be kept unless we are asked to do otherwise.

anthrotype:
Thanks you!

PR #407, opened 2015-11-13
[subset] Preserve the .notdef glyph width when removing its outlines

miguelsousa:
Seems like a lot of steps just to preserve the glyph's width, but I couldn't figure out a simpler way. The CFF table is a PITA to wrangle.


miguelsousa:
I'm going to close this PR and open a new one that will setup `subset` for unit testing, and will also have a simpler solution for preserving the .notdef's width.

PR #411, opened 2015-11-14
[subset] unittest & .notdef advance width fix

miguelsousa:
@behdad looks good?


anthrotype:
Thanks Miguel!
I like the way you organised the unit tests for the subset module. I don't know the cffLib very well, but by looking at the test output you seem to achieve what you intended (i.e. keeping the width of .notdef when stripping its outline).
So the patch looks good to me.

miguelsousa:
:thumbsup:
Yeah, the `CFF ` table is tricky. For example, [this line](https://github.com/behdad/fonttools/pull/411/files#diff-38da6036d96836b0edf8f3688b602f5fR60) may suggest that the glyph has no advance width as there's no value preceding `endchar`, but that's because the glyph's advance width matches [`defaultWidthX`](https://github.com/behdad/fonttools/pull/411/files#diff-38da6036d96836b0edf8f3688b602f5fR50) so the value is omitted.

behdad:
Thanks Miguel!

PR #417, opened 2015-12-01
Fix setup.py to install fontTools.subset

khaledhosny:
Broken since 29d7edf76f9ca888d120f2018e08e2b337ba7cf8.


anthrotype:
Ouch. Thanks!

PR #420, opened 2015-12-04
feaLib: misnamed function in builder_test.py

moyogo:
There are two functions called `test_alternateSubst()` in feaLib/builder_test.py.
The second `test_alternateSubst()` should be called `test_multipleSubst()`.


anthrotype:
Ping @brawer

anthrotype:
Because of the redefinition, one of the test cases is being skipped.
I'm merging this.

brawer:
:+1:

Issue #419, opened 2015-12-04
[feaLib] Support value records format C


brawer:
Currently, feaLib can parse feature files with ValueRecords in format C — in addition to the usual four values, these have XPlaDevice, YPlaDevice, XAdvDevice, and YAdvDevice. However, when it builds GPOS tables, feaLib currently ignores these four device values. The reason is that I have not been able to figure out how exactly otTables can be used to construct the needed Device tables from code.

See the TODO comment in builder.py that contains the string “construct a Device table”. For testing, un-comment the lines after this TODO, and then  `sh run-tests.sh feaLib/builder`. There are two format C ValueRecords in `feaLib/testdata/GPOS_1.fea` which will trigger the code path to construct a GPOS table with these ValueRecords.

anthrotype:
it look like makeotf doesn't support ValueRecords in format C either. I tried to compile the testdata/GPOS_1.fea and got a syntax error. Also the AFDKO feature file spec says it is "currently not implemented". I have never seen one of these.

Anyway, I believe the `XPlaDevice`, `YPlaDevice`, `XAdvDevice `and `YAdvDevice` fields of this format C ValueRecord are equivalent to a `Device` table. The latter is defined inside `otData.py` as having a `StartSize`, `EndSize`, `DeltaFormat` and a `DeltaValue`. The builder needs to set each of these attributes according to the values parsed from the feature file.

We just need understand how to translate between these two.

(btw, thank you for the great work you are doing with feaLib!)

brawer:
Thank you, @moyogo!

PR #421, opened 2015-12-05
[feaLib] make Device Table from device tuples

moyogo:
See #419.


brawer:
Oooh, thank you!

anthrotype:
:+1:

PR #423, opened 2015-12-07
remove old 'Windows' folder as no longer required

anthrotype:
Since we now use setuptools to create platform-specific console scripts, (including *.exe for Windows) we no longer need Win-only instructions nor extra files.

Thanks @twardoch for his contribution.


twardoch:
Thanks for doing this :) I long forgot how it worked :)

PR #428, opened 2015-12-08
ttx: pad tags with space if length is less than 4

anthrotype:
so that you don't need to do `-t "CFF "` anymore.

Fixes https://github.com/behdad/fonttools/issues/265


davelab6:
<3

adrientetar:
You could use str.ljust(4) for this I think.

anthrotype:
I'll use that, thanks Adrien!

PR #431, opened 2015-12-09
use faster, container-based Travis infrastructure

### LINKED ISSUE #429: travis: switch to container-based infrastructure (sudo: false)
currently we use `sudo` to install gcc 4.8 because Brotli used to require C++11 features in order to compile. That is no longer the case, since brotli can now be compiled with old compilers that don't support C++11.

I have opened a PR on https://github.com/google/brotli/pull/278 to disable the option in brotli's `setup.py`. Once that is merged, we will be able to compile brotli with the gcc version pre-installed on Travis, and no longer need `sudo apt-get` it.

We can then add the line `sudo: false` to our `.travis.yml`, and remove the `before_install` part for gcc-4.8.

Our Travis builds will hopefully be faster.

anthrotype:
Fixes #429

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: python
 python:
   - "2.7"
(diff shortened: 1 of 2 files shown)


anthrotype:
total build time went down from 7 to 3 minutes! :)

moyogo:
Great!

PR #434, opened 2015-12-11
README.md: add Appveyor's status badge

anthrotype:


behdad:
:+1: 

For HarfBuzz, we instructed Travis recently to build on Linux and Mac.  Maybe you can do that here as well?

anthrotype:
OK, I'll do that.

PR #435, opened 2015-12-12
Travis multi-os setup

anthrotype:
This is to enable testing on OS X Python with Travis, alongside Linux.

Unfortunately the Python runtime is not yet available on Travis for OS X environment (cf. https://github.com/travis-ci/travis-ci/issues/2312).

Therefore, I had to use [pyenv](https://github.com/yyuu/pyenv) to downlaod and install the various Pythons (3.3, 3.4, 3.5 and pypy) which are not already installed on OS X. For 2.7 I test both the default system Python 2.7.5 that comes with OS X 10.9 (the default Travis OS X machine), and the Python 2.7.10 which is pre-installed on El Capitan.

For setting up the virtual environment, downloading the requirements and running the tests I use [tox](https://tox.readthedocs.org/en/latest/). This integrates nicely with [pytest](https://pytest.org/), which in turn can discover and run any doctests or unittests, besides its own kind of tests.

The nice thing about using tox in combination with pytest is that, if one has all the various Pythons installed locally on one's machine, then one can simply type `tox` and have the test suite run on all of them, each in its own isolated virtual environment. That's what the Python cool kids are now using, so we can't miss out!

BTW, thanks to setting this up, I found an issue with `struct` in the old Python 2.7.5 (Mavericks' default), which raises when format strings are unicodes (e.g. when one does `from __future__ import unicode_literals`). I fixed that in the trak table module by simply not importing unicode_literals, as it's not really needed there.

One more reasons to test as many python environments as we can :)

Oh, the build passes on my fork: https://travis-ci.org/anthrotype/fonttools/builds/96480740

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,12 @@ matrix:
       env: TOXENV=pypy
     - language: generic
       os: osx
+      # the default 'osx' on Travis is 10.9 Mavericks, with Python 2.7.5
+      env: TOXENV=py27
+    - language: generic
+      os: osx
+      # this is to test on 10.11 El Capitan which comes with Python 2.7.10
+      osx_image: osx10.11
       env: TOXENV=py27
     - language: generic
       os: osx
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ install:
   - ./.travis/install.sh
 
 script:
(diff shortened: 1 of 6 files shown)


behdad:
:100:

PR #450, opened 2016-01-07
boundsPen: ignore single points aka. anchors

adrientetar:
This brings the BoundsPen inline with that of robofab and uses only outlines for the calculation of bounds.

See typemytype/booleanOperations#20.

### DIFF
--- a/Lib/fontTools/pens/boundsPen.py
+++ b/Lib/fontTools/pens/boundsPen.py
@@ -9,8 +9,8 @@ __all__ = ["BoundsPen", "ControlBoundsPen"]
 
 
 class ControlBoundsPen(BasePen):
-
-	"""Pen to calculate the "control bounds" of a shape. This is the
+	"""
+	Pen to calculate the "control bounds" of a shape. This is the
 	bounding box of all control points, so may be larger than the
 	actual bounding box if there are curves that don't have points
 	on their extremes.
@@ -23,19 +23,28 @@ class ControlBoundsPen(BasePen):
 	def __init__(self, glyphSet):
 		BasePen.__init__(self, glyphSet)
 		self.bounds = None
+		self._start = None
 
 	def _moveTo(self, pt):
+		self._start = pt
+
+	def _addMoveTo(self):
+		if self._start is None:
+			return
 		bounds = self.bounds
 		if bounds:
-			self.bounds = updateBounds(bounds, pt)
+			self.bounds = updateBounds(bounds, self._start)
 		else:
-			x, y = pt
+			x, y = self._start
 			self.bounds = (x, y, x, y)
+		self._start = None
 
 	def _lineTo(self, pt):
+		self._addMoveTo()
 		self.bounds = updateBounds(self.bounds, pt)
 
 	def _curveToOne(self, bcp1, bcp2, pt):
+		self._addMoveTo()
 		bounds = self.bounds
 		bounds = updateBounds(bounds, bcp1)
 		bounds = updateBounds(bounds, bcp2)
@@ -43,6 +52,7 @@ class ControlBoundsPen(BasePen):
 		self.bounds = bounds
 
 	def _qCurveToOne(self, bcp, pt):
+		self._addMoveTo()
 		bounds = self.bounds
 		bounds = updateBounds(bounds, bcp)
 		bounds = updateBounds(bounds, pt)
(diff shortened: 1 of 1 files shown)


behdad:
Another, simpler, approach is to modify updateBounds instead using an optional argument, to ignore rectangles with zero height and width.

anthrotype:
hmm.. shouldn't it be optional to ignore single-point contours?

anthrotype:
Adrien's patch makes ignoring anchors the default behaviour and there's no optional argument, like robofab's boundsPen.

In my patch, I kept the current behavior and added an optional "ignoreSinglePoints":
https://github.com/anthrotype/fonttools/commit/dc2ef2e341e4ab868801b4b3ff5a867114f182e4

anthrotype:
I haven't pulled a request yet because I wasn't sure whether fontTools.pens is the right place for this...

anthrotype:
> modify updateBounds... to ignore rectangles with zero height and width

sorry, I don't understand..
`updateBounds` takes a rectangle and a point and must returns a new, possibly modified, rectangle. What do you mean by "ignore rectangles with zero height and width"?

adrientetar:
My question so far has been "Why would you not ignore anchors when calculating the bounds of a glyph?". 
The robofab pen says it's "temporarily(?)" making a pen that does ignore anchors. Is there anyone out there who needs the pen to factor in anchors? If not, we should fix it now.

> Another, simpler, approach is to modify updateBounds instead using an optional argument, to ignore rectangles with zero height and width.

Rectangles always start by being zero height and width since it's built incrementally from each point in the contour. I think the right elision approach is indeed the robofab one: store the startPoint instead of processing it and do factor it when an actual line/curve segment follows.

behdad:
> modify updateBounds... to ignore rectangles with zero height and width
> 
> sorry, I don't understand..
> updateBounds takes a rectangle and a point and must returns a new, possibly modified, rectangle. What do you mean by "ignore rectangles with zero height and width"?

Oops.  My bad.  I thought it merges two rectangles.

anthrotype:
> "Why would you not ignore anchors when calculating the bounds of a glyph?"

I don't know.. I guess a single-point contour in a TrueType glyf might be interpreted as an open (invisible) contour, rather than an "anchor".

moyogo:
> My question so far has been "Why would you not ignore anchors when calculating the bounds of a
> glyph?". 

In the glyf table or in UFO3, single points are not anchors.
In UFO2, single points are either anchors or single points, depending on the interpretation whether they are named or not.

moyogo:
BTW, these single points could be useful in TTF. A GPOS anchor can be attached to a single point that can be hinted with TT instructions.

adrientetar:
> In the glyf table or in UFO3, single points are not anchors.

Afaict in the pen protocol, we have no way to tell anchors and single-points apart. It's what we're working with here and I don't think that it changed with ufo3.

> I don't know.. I guess a single-point contour in a TrueType glyf might be interpreted as an open (invisible) contour, rather than an "anchor".

Yes. Probably. But even if we have that, is it worth including it in the bounds? 
Between including anchors which aren't outlines and excluding points which won't even display, I think the choice is straight-forward. 
Ultimately if there's a problem of expressivity it should be adressed independently of what's being done in this PR. I wouldn't argue that current behavior is less wrong than the one proposed in this PR.

> BTW, these single points could be useful in TTF. A GPOS anchor can be attached to a single point that can be hinted with TT instructions.

Right, I know anything's possible with the power of TT instructions. Is such a technique actually used in some fonts? I'd be surprised that people go ahead using single points like this in production, given that some rasterizers may behave strangely when it comes to rendering of bounds calculation as we see here.

behdad:
> Right, I know anything's possible with the power of TT instructions. Is such a technique actually used in some fonts? I'd be surprised that people go ahead using single points like this in production, given that some rasterizers may behave strangely when it comes to rendering or bounds calculation as we see here.

We've definitely seen this in fonts in the wild.  Some Noto fonts probably include it as well.

anthrotype:
so, what's the consensus then? Should we keep the current default behavior and allow to optionally "ignoreSinglePoints", or change the behavior to always ignore them, or simply do nothing (and let clients subclass BoundsPen if they like)?
I personally don't have a strong opinion, and I'm more confused than before.

behdad:
I'm fine if we change this in a backwards-compatible way (ie, default to previous behavior).

adrientetar:
Alright, I just looked at defcon's source again and it seems that in the ufo3 branch glyph.draw(pen) doesn't "draw" anchors while it did in ufo2. 
This behavior seems to bring the (UFO-owned) pointPen inline with segmentPen in terms of what it's expressing, though the issue remains for ufo2.

This was done in commit https://github.com/typesupply/defcon/commit/73a677b3d3a5f228f0e056a7182989a93454fad5#diff-0d4ef48dc0f6e887190b3bc5a98187b4.

adrientetar:
ufoLib doesn't use the anchors hack anymore so as far as I'm concerned, we're good here.

PR #451, opened 2016-01-07
add "ignoreSinglePoints" argument to bounds pen

anthrotype:
see discussion:
https://github.com/behdad/fonttools/pull/450
https://github.com/typemytype/booleanOperations/pull/20

cc @adrientetar @behdad

### DIFF
--- a/Lib/fontTools/pens/boundsPen.py
+++ b/Lib/fontTools/pens/boundsPen.py
@@ -17,25 +17,39 @@ class ControlBoundsPen(BasePen):
 
 	When the shape has been drawn, the bounds are available as the
 	'bounds' attribute of the pen object. It's a 4-tuple:
-		(xMin, yMin, xMax, yMax)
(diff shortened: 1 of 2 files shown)


adrientetar:
See #450, the ufo3 tools (amongst which ufoLib) don't draw anchors in the pen anymore, so we're good when using ufoLib.

anthrotype:
I'm not sure I understand, Adrien. What do you mean "we are good"? Don't we want to make the fonttools bounds pens behave like the robofab ones in certain circumstances (e.g. in booleanOperations' BooleanContour)?

adrientetar:
Right, defcon doesn't draw anchors like this anymore in the ufo3 branch but booleanOperations still does. Hmm... I'm not sure we should alter fontTools to cater for what's essentially a hack in ufo2 tools.

adrientetar:
I'll ask Tal about this.

PR #454, opened 2016-01-10
Remove obsolete code - 'xml.parsers.expat' is part of Python's stdlib

amitdo:
The 'xml.parsers.expat' module has been a part of Python's Standard Library since version 2.0. Link: https://docs.python.org/2/library/pyexpat.html#module-xml.parsers.expat . FontTools requires Python 2.7, or Python 3.3 or later. Checking for the existence of this module is not necessary anymore.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,12 +13,6 @@ except ImportError:
 else:
 	distutils_scripts = []
 
-try:
-	import xml.parsers.expat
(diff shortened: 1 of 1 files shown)


anthrotype:
Yeah, I've been wanting to remove that obsolete check too.
Thanks!

PR #469, opened 2016-01-19
Added otlLib

paarandika:
Without otlLib in the setup file it gives the error 
```python
ImportError: No module named otlLib
```
when using defcon library.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -71,6 +71,7 @@ setup(
 			"fontTools.ttLib",
 			"fontTools.t1Lib",
 			"fontTools.subset",
+			"fontTools.otlLib",
 			"fontTools.ttLib.tables",
 		],
 		py_modules = ['sstruct', 'xmlWriter'],


behdad:
Thanks.  It was rather by design that we have not added it yet, as it's extremely premature and subject to change.  But as long as you keep that in mind, I don't mind others using it.  Note that we'll be changing API, like, today, and again tomorrow, etc for a while.  You can track that here:
https://github.com/behdad/fonttools/issues/468

paarandika:
Thanks for the heads up.

adrientetar:
> It was rather by design that we have not added it yet

If other fontTools modules are calling that one, you need it in the setup script whatsoever.

behdad:
> If other fontTools modules are calling that one, you need it in the setup script whatsoever.

Ah, ok, good point.  Thanks.

PR #470, opened 2016-01-19
_n_a_m_e: add toStr method to NameRecord

anthrotype:
as discussed in https://github.com/typesupply/compositor/pull/10


anthrotype:
I also changed `__str__` to call `toStr` instead of `toUnicode` (with `errors="backslashreplace"` as a safety net).

behdad:
> I also changed __str__ to call toStr instead of toUnicode (with errors="backslashreplace" as a safety net).

Good one!

davelab6:
cc @felipesanches

PR #474, opened 2016-01-21
[feaLib] Accept empty statements in parser

jamesgk:
This is a minor change, but I'm not sure if this is the best place for it.


jamesgk:
Added

brawer:
Thank you!

PR #480, opened 2016-01-23
Add support for subsetting MATH table

khaledhosny:
Support subsetting MATH table, including adding glyph from math variant records to the final subset.
CC @fred-wang.

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -1620,6 +1620,87 @@ def subset_glyphs(self, s):
 def prune_post_subset(self, options):
     return True
 
+@_add_method(otTables.MathGlyphConstruction)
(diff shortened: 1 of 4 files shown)


anthrotype:
Thanks Khaled! :+1: 

Your patch looks good to me, though I don't know enough about MATH table.
I'm only a little bit concerned about the size of the "TestMATH-Regular.ttx" file (4.7 MB). I wonder if a smaller font could be generated for the sake of the tests?

behdad:
Thanks Khaled.  LGTM!  Cosimo, feel free to merge when the test font issue is resolved.  Thanks.

khaledhosny:
I found a thinko in the code that I missed when removing other instances of it, and subsetted the test font with the new pyftsubset :) so it is 288K now.

PR #483, opened 2016-01-24
implement logging

anthrotype:
This patch addresses issue https://github.com/behdad/fonttools/issues/284

Basically, it replaces all the relevant uses of the `print` or `warnings.warn` functions with equivalent calls to the Python's built-in `logging` package.

For more info on Python's logging, you can read the official documentation here:
https://docs.python.org/3.5/library/logging.html
https://docs.python.org/3.5/library/logging.config.html
https://docs.python.org/3.5/howto/logging-cookbook.html

I think I'll wait a bit before merging this, as it affects almost all the modules in the library and would be nice to hear what others think.

The Python logging system can seem quite complex at the beginning (at least it did for me). I've tried to keep this simple and the most conservative possible.

The biggest change from the point of view of clients of the library is that one has to opt-in, rather than opt-out.

That is, no messages, not even of severity WARNING or ERROR, will be printed to stdout or stderr unless the user configures the logging system: which means, to the very least, adding a handler like `StreamHandler` to the "root" logger (ie. the parent of all loggers), which is what `logging.basicConfig()` does.

To make things easier I also added a `configLogger` function -- accessible directly `from fontTools import configLogger` -- which scripts can call to set up the top-level library logger with predefined options (they can be overridden via keyword args).

The individual commits contain more info on each change.
Thanks.

### DIFF
--- a/Lib/fontTools/__init__.py
+++ b/Lib/fontTools/__init__.py
@@ -7,4 +7,9 @@ import logging
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
 
+# clients may call this to configure logging with a predefined handler and format
+from fontTools.misc.loggingTools import configLogger
+
 version = "3.0"
+
+__all__ = ["version", "log", "configLogger"]
--- a/Lib/fontTools/__init__.py
+++ b/Lib/fontTools/__init__.py
@@ -1,4 +1,10 @@
 from __future__ import print_function, division, absolute_import
 from fontTools.misc.py23 import *
+import logging
+
+# add a do-nothing handler to the libary's top-level logger, to avoid
+# "no handlers could be found" error if client doesn't configure logging
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
 
 version = "3.0"
--- a/Lib/fontTools/__init__.py
+++ b/Lib/fontTools/__init__.py
@@ -1,14 +1,15 @@
 from __future__ import print_function, division, absolute_import
 from fontTools.misc.py23 import *
 import logging
+from fontTools.misc.loggingTools import Logger, configLogger
 
-# add a do-nothing handler to the libary's top-level logger, to avoid
-# "no handlers could be found" error if client doesn't configure logging
+# set the logging.Logger class to one which supports the "last resort" handler,
+# to be used when the client doesn't explicitly configure logging.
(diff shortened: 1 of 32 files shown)


anthrotype:
So, I would like to merge this if no one has objections or comments...
We can always fix things up later. Can I proceed?

anthrotype:
I'm sorry for the insistence, but if this huge patch doesn't get merged soon-ish, it risks to generate conflicts with changes elsewhere.. That means I'll need to fix and rebase it forever -- or even worse I might tinker again with it :-P
So, did anybody have a look by any chance? 
I'm open to discuss changes or provide more details on the implementation.
 
ping @behdad @brawer @moyogo

moyogo:
LGTM

anthrotype:
I changed my mind...

I re-read the logging documentation once again in search for guidance as to the best practices when implementing logging for libraries such as fontTools.

Initially I had settled for the old pre-Python 3.2 behavior. That is, when a logging event occurred and no handlers were found, then either the events would be silently dropped (if `logging.raiseExceptions` is
set to False), or it would throw a single "No handlers could be found" warning
(this was the default). The pre-3.2 docs recommend in this case to attach a `NullHandler` to the top-level library logger, if the library developers don't want the above warning message to be presented to the user who doesn't configure logging.

After some thought, I think now it's better to follow the current Python 3 behavior for when a client of the library doesn't provide any explicit logging configuration.

The new default behavior in this circumstance is that events are output using a "handler of last resort". The latter works like logging.StreamHandler which writes messages without formatting to the current value of `sys.stderr` (thus respecting any redirections which may be in effect).

I believe this is better than forcing everyone to update their own scripts and deliberately configure logging.

I applied the change in the last two commits. Now everyone should be happy and I can finally merge.

behdad:
I've got to say, the logging infrastructure blows my mind away whenever I try to understand it...  Thanks for taking care of this.  Will eventually study and learn.

In the mean time, let's see what @garretrieger has to say about the changes to subset module, as he uses a custom logger with the subsetter.

anthrotype:
Yeah, I know what you mean.

Anyway, I'm happy to assist him or anyone who might have troubles with the logging configuration.

The new system actually enables greater flexibility on the user's side than it was allowed before. You won't need any custom logger anymore, you just need to understand how logging is organised in the library you are working with, and then configure it using all the facilities that the logging and logging.config modules provide.

You can then make it do anything you like.

PR #484, opened 2016-01-25
[ttx] minor changes

miguelsousa:


anthrotype:
Yes, I also noticed that issue with ttx option parser -- fixed it the same way as you did:

https://github.com/anthrotype/fonttools/commit/ac6ac113dd7c5672bf7ac370f87b3b9f857a9662#diff-fce6b83de86f2e83411202adc745ffb2R310

I just tried to merge this PR locally with my logging patch and there were no merge conflicts.
So I'll merge this now.
Thanks!

PR #489, opened 2016-01-26
replace codecs.open with py23.open

anthrotype:


anthrotype:
I also imported `unicode_literals` in mtiLib and make it open text files with utf-8 encoding. So, the mtiLib.Tokenizer should now expect to consistently receive unicodes.
FeaLib does the same, I think.

PR #493, opened 2016-01-31
add --with-zopfli option to pyftsubset and ttx

### LINKED ISSUE #491: Use Zopfli compression for WOFF files
Creating a WOFF1 font with ttx/fonttools results in a larger file than when you create a TTF file and then compress it with tools like [sfnt2woff-zopfli](https://github.com/bramstein/sfnt2woff-zopfli) or [ttf2woff](http://wizard.ae.krakow.pl/~jb/ttf2woff/), which use Zopfli compression. It would be nice to directly output an optimal compressed font.

anthrotype:
Fixes #491.

The `zopfli` extension module which is required to use the new option can be installed with this pip command:
```
pip install -v git+https://github.com/anthrotype/py-zopfli#egg=zopfli
```

The official one currently available on PyPI uses an outdated copy of Zopfli source code and is not compatible with Python 3 nor with Windows.

While working on this, I noticed that pyftsubset's `canonical_order` default value was still set to `False`, whereas it should have been set to `None`.
The meaning of the related `reorderTables` argument of `TTFont.save` method in fact changed a while ago with commit 6ba67ab.
When `reorderTables` is set to False, the input font's original table order is "kept": which still means performing some sort of reordering. If one wishes to avoid any reordering (i.e. use the order determined by inter-table dependencies), then `reorderTables` needs to be set to `None`.

With this patch, `--no-canonical-order` is equal to passing `reorderTables=None`, and `--canonical-order` equals `reorderTables=True` (`reorderTables=False` is not accessible via command line options, as it's not relevant for the subsetter).

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -112,6 +112,11 @@ Output options:
       Specify flavor of output font file. May be 'woff' or 'woff2'.
       Note that WOFF2 requires the Brotli Python extension, available
       at https://github.com/google/brotli
+  --with-zopfli
+      Use the Google Zopfli algorithm to compress WOFF. The output is 3-8 %
+      smaller than pure zlib, but the compression speed is much slower.
+      The Zopfli Python bindings are available at:
+      https://github.com/anthrotype/py-zopfli
 
 Glyph set expansion:
   These options control how additional glyphs are added to the subset.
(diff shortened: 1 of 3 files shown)


behdad:
Two comments:

  * Don't like using a global module-level variable as an option...

  * Perhaps we can reuse flavor for this?  Eg "woff:zopfli" for example?

behdad:
Or a flavor-options...  Woff2 also has a couple different options for speed versus compression.

anthrotype:
> Don't like using a global module-level variable as an option

I wasn't sure about that either. The reason I did that was to avoid having to add another argument to the TTFont (and/or SFNTWriter) constructor. Since zlib.compress function was used in two places inside `sfnt` module (in the main writer's save method, for compressing metadata, and in the WOFFDirectoryEntry.encodeData method), I thought I could switch between zlib and zopfli algorithm by setting a global variable (the directory entry doesn't know about the writer instance).

about reusing `--flavor` or adding a `--flavor-options`, I think I prefer the latter. I'll see what I can do.

PR #498, opened 2016-02-01
[O_S_2f_2] skip updateFirstAndLastCharIndex if cmap table class is DefaultTable

anthrotype:
Fixes https://github.com/behdad/fonttools/issues/497


anthrotype:
I think this is good enough for now. Please revert/open new issue if you think it needs more work.
Thanks.

behdad:
Thanks.  I pushed a slight simplification.

PR #511, opened 2016-02-04
Create renaming-glyphs.py

### LINKED ISSUE #149: [snippets] Renaming glyphs
For TrueType, something like this works:

``` python
    font = fontTools.ttLib.TTFont(inpath)
    glyphnames = font.getGlyphOrder()
        # Do the renaming in glyphnames...
    font = fontTools.ttLib.TTFont(inpath) # load again
    font.setGlyphOrder(glyphnames)
    post = font['post']
    font.save(outpath)
```

For cff, we need to set font['CFF '].cff.topDictIndex[0].charset, but also to remap font['CFF '].cff.topDictIndex[0].CharStrings as in https://github.com/behdad/fonttools/issues/147.

davelab6:
Closes #149

### DIFF
--- a/Snippets/renaming-glyphs.py
+++ b/Snippets/renaming-glyphs.py
@@ -1,0 +1,20 @@
+# To rename glyphs in a TrueType flavour file, something like this works:
+
+font = fontTools.ttLib.TTFont(inpath)
+glyphnames = font.getGlyphOrder()
+# Do the renaming in glyphnames...
+font = fontTools.ttLib.TTFont(inpath) # load again
+font.setGlyphOrder(glyphnames)
+post = font['post']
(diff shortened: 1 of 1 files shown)


miguelsousa:
@anthrotype merge?

anthrotype:
No. It's missing imports and doesn't even run. I suggest you look at how ufo2ft does the glyph renaming here
https://github.com/googlei18n/ufo2ft/blob/2d42d89b41538acdd89e66faa6c7c475bbea52ff/Lib/ufo2ft/postProcessor.py#L28

behdad:
Either finish or close.

khaledhosny:
One year later… closing then (the issue is still open, though).

PR #521, opened 2016-02-10
Replace one last call to debug() in mtiLib

jamesgk:
I think this was mistakenly left out of
b62f5509524a49101ab6f211f1c3b63458f52f5e


anthrotype:
Oops. Thanks for spotting it!

PR #529, opened 2016-02-19
Add Snippets/post-format.py

davelab6:
I wrote this for myself just now, figured it would best live here :)

### DIFF
--- a/Snippets/post-format.py
+++ b/Snippets/post-format.py
@@ -1,8 +1,9 @@
-#! /usr/bin/env python
+#! /usr/bin/env python2
 #
-# Sample script to print post table version, based on cmap-format.py
+# Sample script to print post table version
 #
 # TODO: Extend convert an outfile.{o,t}tf with version 1, 2, 3, 4
+# TODO: Extend to print all table versions
 #
 # To print the versions of all files under the present directory,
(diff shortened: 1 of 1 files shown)


behdad:
I'm more and more being convinced that I need to work out pyftshell or something.  Then you can do things like:
```
  pyftshell -e 'print(fontfile, font["post"].format)' *.ttf
```
adding scripts like this is really a futile exercise.

twardoch:
That'd be cool. It sounds a bit like 
https://amoffat.github.io/sh/
which I may start using instead of the subprocess crap in my tools.

davelab6:
That would rock! That would make it much easier to develop nototools/fontbakery/etc :)

davelab6:
I played around with trying this as a python 1-liner and had no luck, so yes, a `pyftshell` that took care of common sharp edges for that kind of thing would be really great :)

    for font in `find . -name \*.\*tf -print`; do \
      python -c "from fontTools.ttLib import TTFont; print(str(TTFont('"$font"')['post'].formatType))" | sort | cut -d\  -f1 | uniq -c ; \
	done;

I simplified the script; if you want to reject it, no problem, I'll put it elsewhere

PR #544, opened 2016-03-11
test for NotImplemented in __ne__; add __eq__ and __ne__ to ttProgram.Program

anthrotype:
`_g_l_y_f.Glyph` objects have handy `__eq__` and `__ne__` methods. However their `program` attribute did not, so it was using `object`'s default `id(self) == id(other)`, which returns False even when the content of two Program instances is the same.
So I added that.

While doing this I noticed that there are other classes in fonttools which similarly implement `__eq__` and `__ne__`, but they have a problem in that, `__ne__` is defined as `not self.__eq__(other)`, but when the `type(self) != type(other)`, then `__eq__` will (correctly) return `NotImplemented`, and thus `__ne__` will return `not NotImplemeted`, which is always False no matter what.

This can lead to illogic results like:

```python
>>> from fontTools.ttLib.tables.DefaultTable import DefaultTable
>>> t = DefaultTable('test')
>>> t == 0
False
>>> t != 0
False
```

The latter of course should return True.
So I changed all `__ne__` to first check if `__eq__` returned `NotImplemeted`, and if so return that as well. This will delegate the comparison to the superclass (`object`) -- which, in the example above, will return True.

### DIFF
--- a/Lib/fontTools/pens/ttGlyphPen_test.py
+++ b/Lib/fontTools/pens/ttGlyphPen_test.py
@@ -68,7 +68,6 @@ class TTGlyphPenTest(unittest.TestCase):
         pen.endPath()
         endPathGlyph = pen.glyph()
 
-        endPathGlyph.program = closePathGlyph.program
         self.assertEqual(closePathGlyph, endPathGlyph)
 
     def test_glyph_errorOnUnendedContour(self):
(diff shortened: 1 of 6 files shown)


twardoch:
```
>>> t == 0
False
>>> t != 0
False
```
That is awesome :)

anthrotype:
but that is even more awesome:
```python
>>> t == 0
False
>>> not (t == 0)
True
```

twardoch:
Thanks, I just wondered what that would yield :)

PR #545, opened 2016-03-12
Fix handling of default lookupflag

khaledhosny:
According to the spec:
> The lookupflag attribute defaults to 0 at the start of a feature
> block.
>
> The lookupflag attribute stays the same until explicitly changed, until
> a lookup reference statement is encountered that changes it, until the
> script is changed, or until the end of the feature.

This is an attempt to fix this by resetting the lookupflag at the start
and end of feature/lookup blacks. I’m not sure if resetting it in lookup
blocks is correct (my reading of the spec suggests it is not), but one
needs to test this against makeotf and see how it behaves here.

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -338,6 +338,8 @@ class Builder(object):
         self.language_systems = self.get_default_language_systems_()
         self.cur_lookup_ = None
         self.cur_feature_name_ = name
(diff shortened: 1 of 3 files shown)


brawer:
Thank you! Checked with makeotf v2.0.90; it produces the exact same output as the test case in your change.

brawer:
Filed https://github.com/adobe-type-tools/afdko/issues/109 for the spec clarification.

PR #546, opened 2016-03-14
[feaLib] Support name table

khaledhosny:
Needs some tests, but I though to get feedback first since this is my first try with feaLib code.

### DIFF
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -2,6 +2,7 @@ from __future__ import print_function, division, absolute_import
 from __future__ import unicode_literals
 from fontTools.feaLib.error import FeatureLibError
 from fontTools.feaLib.lexer import Lexer, IncludingLexer
+from fontTools.misc.py23 import *
 import fontTools.feaLib.ast as ast
 import os
 import re
@@ -670,6 +671,7 @@ class Parser(object):
         handler = {
             "GDEF": self.parse_table_GDEF_,
             "head": self.parse_table_head_,
+            "name": self.parse_table_name_,
         }.get(name)
         if handler:
             handler(table)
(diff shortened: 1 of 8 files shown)


Lib/fontTools/feaLib/parser.py
```diff
@@ -712,6 +714,77 @@ def parse_table_head_(self, table):
...
+                              platEncID, langID, string)
+
+    def unescape_mac_name_string(self, string):
+        def unescape(match):
+            n = match.group(0)[1:]
+            # Can't figure a way that works for both Python 2 and 3
+            if bytes == str:
+                c = chr(int(n, 16)).decode('mac_roman')
+            else:
+                c = bytes([int(n, 16)]).decode('mac_roman')
```
adrientetar: I think you want to use bytechr() in fontTools.misc.py23.

khaledhosny: Thanks, I looked into py23 but I obviously missed this.

khaledhosny:
I’ve added a couple of test files and squashed all the commits together.

Lib/fontTools/feaLib/parser.py
```diff
@@ -712,6 +714,77 @@ def parse_table_head_(self, table):
...
+                                      self.cur_token_location_)
+
+    def parse_nameid_(self):
+        assert self.cur_token_ == "nameid", self.cur_token_
+        location, nameID = self.cur_token_location_, self.expect_number_()
+        if nameID > 255:
+            raise FeatureLibError("Expected name id < 255",
+                                  self.cur_token_location_)
+        if 1 <= nameID <= 6:
+            # TODO: issue a warning
```
anthrotype: I suggest you import `logging` and create a `log` instance at the top of the parser module:

```python
import logging
```

and then here you call `log.warning(...)` to send an appropriate warning message.

anthrotype:
LGTM! Thanks Khaled!

/cc @brawer

brawer:
:+1:

PR #547, opened 2016-03-15
feaLib: support string input

### LINKED ISSUE #443: [feaLib] allow string as input
feaLib should have an entry-point that takes a string of features as input and works with that.

See jamesgk/ufo2ft#9. cc @brawer

adrientetar:
Untested. Closes #443.

r? @brawer

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -8,15 +8,19 @@ from fontTools.ttLib.tables import otBase, otTables
 import itertools
 
 
-def addOpenTypeFeatures(featurefile_path, font):
-    builder = Builder(featurefile_path, font)
+def addOpenTypeFeatures(font, featurefile_path=None, featurefile_data=None):
+    # we may have a path but no data, or data but no path
+    assert featurefile_path or featurefile_data
+    # pack this until we yield it the lexer
+    featurefile_path_data = (featurefile_path, featurefile_data)
+    builder = Builder(font, featurefile_path_data)
     builder.build()
 
 
 class Builder(object):
-    def __init__(self, featurefile_path, font):
-        self.featurefile_path = featurefile_path
+    def __init__(self, font, featurefile_path_data):
         self.font = font
+        self.featurefile_path_data = featurefile_path_data
         self.glyphMap = font.getReverseGlyphMap()
         self.default_language_systems_ = set()
         self.script_ = None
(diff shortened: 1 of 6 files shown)


brawer:
Looks good, but it [breaks the Travis build](https://travis-ci.org/behdad/fonttools/builds/116202716); mind fixing that?

Lib/fontTools/feaLib/builder.py
```diff
@@ -8,15 +8,19 @@
 import itertools
 
 
-def addOpenTypeFeatures(featurefile_path, font):
-    builder = Builder(featurefile_path, font)
+def addOpenTypeFeatures(font, featurefile_path=None, featurefile_data=None):
```
anthrotype: I suggest to rename the keyword arguments simply "path" and "data" as it's shorter and it's clear what it is about

Lib/fontTools/feaLib/builder.py
```diff
@@ -8,15 +8,19 @@
...
+    # we may have a path but no data, or data but no path
+    assert featurefile_path or featurefile_data
```
anthrotype: Since one of the two is required it is better to raise an exception with some usage info massage rather than an assertion

brawer:
Ping?

adrientetar:
Sorry I have many things to work on these days. Should be fixed now.

brawer:
Thank you!

Lib/fontTools/feaLib/builder.py
anthrotype: wouldn't `feapath` and `feadata` be clear enough as keyword argument names here?
`featurefile_data` seems to me a bit too long/verbose...
/cc @adrientetar @brawer

adrientetar: fea_path should be fine I guess

anthrotype: FWIW, I'm ok with `fea_path` and `fea_data` if you prefer having the underscore.

anthrotype: Ok, i'll change that.

Adrien, I'm also wondering, why did you choose to pack (fea_path, fea_data) into a tuple and pass it down to Builder and Parser as a tuple until the Lexer unpacks it, instead of as two separate arguments?

adrientetar: Because they're not used independently, so I stuck to one variable.

anthrotype: I guess you're right.

PR #548, opened 2016-03-17
[feaLib] Fix typo in testdata/spec9e.*

mashabow:
Feature File Spec itself was fixed in https://github.com/adobe-type-tools/afdko/pull/110


brawer:
Thank you!

PR #550, opened 2016-03-18
[feaLib] Support stylistic set featureNames and size feature

khaledhosny:
Does not handle featureNames for cvXX features, but it shouldn’t be hard
for someone to extend the code to support them if inclined to do so.

### DIFF
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -726,16 +726,7 @@ class Parser(object):
                 raise FeatureLibError("Expected nameid",
                                       self.cur_token_location_)
 
-    def parse_nameid_(self):
-        assert self.cur_token_ == "nameid", self.cur_token_
-        location, nameID = self.cur_token_location_, self.expect_number_()
-        if nameID > 255:
-            raise FeatureLibError("Expected name id < 255",
-                                  self.cur_token_location_)
-        if 1 <= nameID <= 6:
-            # TODO: issue a warning
-            return None
-
+    def parse_name_(self):
         platEncID = None
         langID = None
         if self.next_token_type_ == Lexer.NUMBER:
@@ -764,8 +755,21 @@ class Parser(object):
         elif platformID == 3 and platEncID == 1:
             string = self.unescape_windows_name_string(string)
 
-        return ast.NameRecord(location, nameID, platformID,
-                              platEncID, langID, string)
+        return platformID, platEncID, langID, string
+
+    def parse_nameid_(self):
+        assert self.cur_token_ == "nameid", self.cur_token_
+        location, nameID = self.cur_token_location_, self.expect_number_()
+        if nameID > 255:
+            raise FeatureLibError("Expected name id < 255",
+                                  self.cur_token_location_)
+        if 1 <= nameID <= 6:
+            # TODO: issue a warning
+            return None
+
+        platformID, platEncID, langID, string = self.parse_name_()
+        return ast.NameRecord(location, nameID, platformID, platEncID,
+                              langID, string)
 
     def unescape_mac_name_string(self, string):
         def unescape(match):
@@ -881,6 +885,9 @@ class Parser(object):
         location = self.cur_token_location_
         tag = self.expect_tag_()
         vertical = (tag in {"vkrn", "vpal", "vhal", "valt"})
+        stylisticset = None
+        if tag in ["ss%02d" % i for i in range(1, 20+1)]:
+            stylisticset = tag
 
         use_extension = False
         if self.next_token_ == "useExtension":
@@ -888,7 +895,7 @@ class Parser(object):
             use_extension = True
 
         block = ast.FeatureBlock(location, tag, use_extension)
-        self.parse_block_(block, vertical)
+        self.parse_block_(block, vertical, stylisticset)
         return block
 
     def parse_feature_reference_(self):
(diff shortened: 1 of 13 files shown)


Lib/fontTools/feaLib/parser.py
```diff
@@ -764,8 +755,21 @@ def parse_nameid_(self):
...
+        return platformID, platEncID, langID, string
+
+    def parse_nameid_(self):
+        assert self.cur_token_ == "nameid", self.cur_token_
+        location, nameID = self.cur_token_location_, self.expect_number_()
+        if nameID > 255:
+            raise FeatureLibError("Expected name id < 255",
+                                  self.cur_token_location_)
+        if 1 <= nameID <= 6:
+            # TODO: issue a warning
```
anthrotype: what does makeotf do when one attempts to override one of these nameIDs?

anthrotype:
LGTM! 
Thanks you Khaled :+1: 

I'll wait for Sascha to merge it.

Lib/fontTools/feaLib/parser.py
khaledhosny: My understanding is that makeotf creates the name table from scratch, so no overriding is possible (its code just records the last taken one and increments it for each use).

anthrotype: from http://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#9.e

> Note that ids 1-6 (Family, Subfamily, Unique, Full, Version, and FontName) are reserved by the implementation and cannot be overridden; doing so will elicit a warning message and the record will be ignored.

let's emit this warning then, no?

khaledhosny: Agreed, I even tried to use the logging module here the first time I wrote this code, I just wasn’t sure what the format of the warning should be, so I left it in the hope that someone else will take care of it :)

anthrotype: well, I can take care of it if you prefer.

khaledhosny:
OK, I think this is ready to be merged now (it took some iterations, either the spec is vague or I’m too tired :) )

khaledhosny:
Hmm, I guess it is not ready yet. I keep understanding things differently every time I read the spec!

khaledhosny:
Opened adobe-type-tools/afdko#112, for spec clarification. I think my current implementation is fine, but not sure it 100% matches makeotf.

Lib/fontTools/feaLib/parser.py
```diff
@@ -1062,6 +1128,15 @@ def expect_float_(self):
         raise FeatureLibError("Expected a floating-point number",
                               self.cur_token_location_)
 
+    def expect_decipoint_(self):
+        if self.next_token_type_ == Lexer.FLOAT:
+            return self.expect_float_() * 10
+        elif self.next_token_type_ is Lexer.NUMBER:
+            return self.expect_number_()
+        else:
+            raise FeatureLibError("Expected a fixed or floating point number",
```
anthrotype: What do you mean by "fixed" here? Shouldn't it be just "integer"?

```diff
-"Expected a fixed or floating point number"
+"Expected an integer or floating point number"
```

khaledhosny: A thinko :)

khaledhosny:
FWIW, I have support for BASE table locally and I’ll make a PR for it once this one is merged to avoid merge conflicts.

Lib/fontTools/feaLib/parser.py
```diff
@@ -663,6 +663,30 @@ def parse_subtable_(self):
...
         return ast.SubtableStatement(location)
 
+    def parse_size_parameters_(self):
+        assert self.is_cur_keyword_("parameters")
+        location = self.cur_token_location_
+        DesignSize = self.expect_decipoint_()
+        SubfamilyID = 0
+        RangeStart = 0
+        RangeEnd = 0
+        if self.next_token_type_ is Lexer.NUMBER:
```
anthrotype: I think there's still a difference between your current code's behaviour and that of makeotf.
If I change the "feaLib/testdata/size2.fea" test case to only set the first parameter (DesignSize), like that:

```diff
feature size {
+  parameters 10.0 0;
-  parameters 10.0;
} size;
```

then feaLib sets the other three values (SubfamilyID, RangeStart and RangeEnd) to 0, whereas makeotf raises a syntax error:
```
syntax error at ";" missing { FONTREV NUM } [features 30]
```

Similarly, if I set `parameters 10.0 0 0;`, without the last RangeEnd value, makeotf raises another fatal error, which makes clear what it's expecting:

```
makeotfexe [FATAL] <Test-Regular> size feature must have 4 or 2 parameters if sub family code is zero!
```

Do you mind changing this to match makeotf?
Thanks

Lib/fontTools/feaLib/parser.py
```diff
@@ -1062,6 +1128,15 @@ def expect_float_(self):
         raise FeatureLibError("Expected a floating-point number",
                               self.cur_token_location_)
 
+    def expect_decipoint_(self):
+        if self.next_token_type_ == Lexer.FLOAT:
+            return self.expect_float_() * 10
```
anthrotype: I believe this is wrong, at least according to my reading of the [feature_file_syntax_spec#8.b](http://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#8.b), as well as the comments by @readroberts in https://github.com/adobe-type-tools/afdko/issues/112.

When there is an integer, it is interpreted as a "decipoint" (1/10 of a point); alternatively, this

> may also be specified more directly as decimal point values... For example, "8.0" and "80" will both result in the same value being stored in the font.

In your code, it's the opposite. When there's a float, you multiply it by 10, whereas when it's an integer you take it as is.
I believe you need to divide the integer by 10, and leave the float as is.

Lib/fontTools/feaLib/testdata/size.fea
```diff
@@ -0,0 +1,3 @@
+feature size {
+  parameters 10.0 0 0 0;
```
anthrotype: with makeotf, `parameters 10.0` etc. compiles to ```<DesignSize value="10.0"/>``` etc.
The makeotf output is the same for  `parameters 100` (i.e. using "decipoints" notation).

With your patch, feaLib compiles the same `parameters 10.0` to `<DesignSize value="100.0"/>` (cf. "size.ttx" file); and it compiles `parameters 100` to `<DesignSize value="100"/>`.

See https://github.com/behdad/fonttools/pull/550/files#r56748809

Lib/fontTools/feaLib/parser.py
khaledhosny: OK, I see what is going here. The value stored in the binary is a decipoint, but FontTools converts to/from float when writing/reading the value, so setting 10 here will result in 100 being stored in the binary table. I’ll fix the code.

Lib/fontTools/feaLib/parser.py
khaledhosny: OK, I think it should be fixed now (no nice error message, though).

anthrotype:
Ok, I verified the ssXX featureNames and size feature now match the ones produced by makeotf. I shall merge this as soon as the Appveyor build completes. @brawer can always fix up things later if needed.
Thanks Khaled!

khaledhosny:
Thanks @anthrotype :)

PR #554, opened 2016-03-20
Disable optimisations when decompiling TrueType instructions

schriftgestalt:


behdad:
LGTM.  I'm not on my laptop right now.  Feel free to merge.

PR #556, opened 2016-03-21
[feaLib] take in a path or a file-like object

anthrotype:
as discussed in https://github.com/behdad/fonttools/issues/443#issuecomment-199252850 and following comments


brawer:
Looks good to me.

PR #557, opened 2016-03-22
[feaLib] Support OS/2 table

khaledhosny:


anthrotype:
I'm trying to understand where that apostrophe in `<achVendID value="'ADBE"/>` is coming from, which is making the python 2 CI tests fail.

By the way, I noticed that in the `BuilderTest.expect_ttx` function, the `fromfile` and `tofile` arguments to `difflib.unified_diff` are inverted (cc @brawer). They should be like this:

```diff
diff --git a/Lib/fontTools/feaLib/builder_test.py b/Lib/fontTools/feaLib/builder_test.py
index 6c37909..39ae5b7 100644
--- a/Lib/fontTools/feaLib/builder_test.py
+++ b/Lib/fontTools/feaLib/builder_test.py
@@ -109,7 +109,7 @@ class BuilderTest(unittest.TestCase):
         expected = self.read_ttx(expected_ttx)
         if actual != expected:
             for line in difflib.unified_diff(
-                    expected, actual, fromfile=path, tofile=expected_ttx):
+                    expected, actual, fromfile=expected_ttx, tofile=path):
                 sys.stdout.write(line)
             self.fail("TTX output is different from expected")
```

Lib/fontTools/feaLib/parser.py
```diff
@@ -838,6 +839,35 @@ def parse_table_BASE_(self, table):
             elif self.cur_token_ == ";":
                 continue
 
+    def parse_table_OS_2_(self, table):
+        statements = table.statements
+        numbers = ("FSType", "TypoAscender", "TypoDescender", "TypoLineGap",
+                   "winAscent", "winDescent", "XHeight", "CapHeight",
+                   "WeightClass", "WidthClass", "LowerOpSize", "UpperOpSize")
+        ranges = ("UnicodeRange", "CodePageRange")
+        os2 = {}
```
anthrotype: local variable `os2` is assigned but never used

anthrotype:
Ah, that is where it's coming from: https://github.com/behdad/fonttools/blob/master/Lib/fontTools/ttLib/tables/O_S_2f_2.py#L174

in python 2.7 this happens:
```python
>>> achVendID = u"ADBE"
>>> repr(achVendID)
"u'ADBE'"
>>> repr(value)[1:-1]
"'ADBE"
```

I wonder why we don't simply do `str(value)` here, but instead do a `repr` and then strip the quotes off?

```diff
elif name == "achVendID":
-    writer.simpletag(name, value=repr(value)[1:-1])    
+    writer.simpletag(name, value=str(value))
```

anthrotype:
Oh, now I see. `achVendID` is a 4 byte CHAR array and therefore may contain null bytes `b"\x00"`, but in fromXML we need to use `eval` (via `saveEval`) which raises TypeError if given a string null bytes... So we need to escape the whole thing.

OK, the fix to make the OS/2 tests pass should be the one below. It avoids the little `u` in the python2 repr when the achVendID string is a unicode string instead of a byte string.

@khaledhosny mind applying that please?

```diff
diff --git a/Lib/fontTools/ttLib/tables/O_S_2f_2.py b/Lib/fontTools/ttLib/tables/O_S_2f_2.py
index cbba816..307be56 100644
--- a/Lib/fontTools/ttLib/tables/O_S_2f_2.py
+++ b/Lib/fontTools/ttLib/tables/O_S_2f_2.py
@@ -171,7 +171,7 @@ class table_O_S_2f_2(DefaultTable.DefaultTable):
 			elif name in ("fsType", "fsSelection"):
 				writer.simpletag(name, value=num2binary(value, 16))
 			elif name == "achVendID":
-				writer.simpletag(name, value=repr(value)[1:-1])
+				writer.simpletag(name, value=repr(tostr(value))[1:-1])
 			else:
 				writer.simpletag(name, value=value)
 			writer.newline()
```

Lib/fontTools/feaLib/parser.py
```diff
@@ -838,6 +839,35 @@ def parse_table_BASE_(self, table):
...
+        while self.next_token_ != "}":
```
anthrotype: in the unlikely event that the table definition is incomplete and lacks the ending brace, doesn't this while block goes into infinite loop? This would apply not just here, but also elsewhere in the parser module, every time we do `while self.next_token_ != ...`.
I wonder if there is a way to prevent this from happening. Maybe checking if `self.next_token_` is not None as well (that should be the case when there's nothing else to parse, e.g. at the end of the file)?

```python
while (self.next_token_ and self.next_token_ != "}"):
```

anthrotype:
apart from the comments above, this looks good to me.
Thanks Khaled! :)

brawer:
Same here — thank you, Khaled!

khaledhosny:
I fixed the apostrophe issue before seeing the comments here by using safeEval just like the code OS/2 class is doing.

Lib/fontTools/feaLib/parser.py
khaledhosny: I don’t know, but that is not the only place that does this. So I think we need to check this and if it is actually an issue fix all of them at once.

Lib/fontTools/feaLib/parser.py
khaledhosny: Fixed.

PR #563, opened 2016-03-31
Support subsetting of variation fonts

brawer:


anthrotype:
:+1:

Lib/fontTools/subset/__init__.py
```diff
@@ -2318,14 +2334,14 @@ def subset_glyphs(self, s):
...
-def prune_pre_subset(self, options):
+def prune_pre_subset(self, font, options):
     # Drop all signatures since they will be invalid
     self.usNumSigs = 0
     self.signatureRecords = []
     return True
 
 @_add_method(ttLib.getTableClass('maxp'))
-def prune_pre_subset(self, options):
+def prune_pre_subset(self, font, options):
```
behdad: FWIW, I would have preferred the font to stay out of the arguments.  Instead, we should have a round that collects all nameIDs that stay, and use that.  We should do that later on.

PR #571, opened 2016-04-11
run coverage.py on Travis and upload report to coveralls

anthrotype:
I really recommend that we use [coverage.py](http://coverage.readthedocs.org/en/coverage-4.0.3) to measure our test suite coverage. It's really cool.
We even have a badge on the homepage, so we'd better use it! :)
However that link leads nowhere... https://coveralls.io/r/behdad/fonttools
Perhaps @behdad created it in the past but then decided to close it?

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,28 @@
+[run]
+# measure 'branch' coverage in addition to 'statement' coverage
+# See: http://coverage.readthedocs.org/en/coverage-4.0.3/branch.html#branch
+branch = True
+
+# list of directories to measure
+source =
+    Lib/fontTools
+
+[report]
+# Regexes for lines to exclude from consideration
+exclude_lines =
+    # keywords to use in inline comments to skip coverage
+    pragma: no cover
+
+    # don't complain if tests don't hit defensive assertion code
+    raise AssertionError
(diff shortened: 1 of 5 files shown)


behdad:
That was my goal.  Not sure where I dropped the ball.  Anyway, enabling now.

anthrotype:
Thank you! 🎉

behdad:
First numbers in: 71%.  Not bad at all!

I promise to add tests to mtiLib soon!

anthrotype:
@behdad perhaps it would be better to disable the email notification from coveralls.io on every single push? It's kind of noisy...

behdad:
It's set to only email if coverage changed by +/- 3% only.

On Mon, Apr 11, 2016 at 2:08 AM, Cosimo Lupo <notifications@github.com>
wrote:

> @behdad <https://github.com/behdad> perhaps it would be better to disable
> the email notification from coveralls.io on every single push? It's kind
> of noisy...
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/behdad/fonttools/pull/571#issuecomment-208242582>
>



-- 
behdad
http://behdad.org/

PR #573, opened 2016-04-13
[SVG] Better handling of empty <colorPalettes> element

miguelsousa:
Before, when compiling from XML, only `<colorPalettes></colorPalettes>` would be accepted.
Now, all of these pass compilation as well,
```xml
<colorPalettes>    </colorPalettes>
```
```xml
<colorPalettes />
```
```xml
<colorPalettes>
</colorPalettes>
```


RoelN:
What if it's absent? See https://github.com/behdad/fonttools/issues/531

miguelsousa:
@RoelN thanks for pointing that out. I'll see what I can do.

miguelsousa:
OK, it now compiles and decompiles without the `<colorPalettes>` element.

PR #579, opened 2016-04-17
fealib: allow newlines in string literals (and skip over them)

### LINKED ISSUE #578: [feaLib] Lexer for double quote strings should not stop at newline?
I have fonts and fea files with a name table where some nameid strings have newlines in the string. According to the spec 

> A string is composed of 1-byte ASCII characters enclosed by ASCII double quote characters ("). Newlines embedded within the string are removed from the character sequence to be stored.

Looking at a ttx dump these strings have `&#13;` inserted for the newlines so ttx seems to have no issue, but loading the fea file with feaLib fails because in lexer.py there is `self.scan_until_('"\r\n')`
If I remove  `\r\n` it loads fine, but then test_string fails in lexer_test.py because it doesn't raise an Exception at the embedded newline. Maybe I'm missing something, but I don't see a reason for the `\r\n` in the double quote lexer.

anthrotype:
should fix #578 

/cc @brawer

### DIFF
--- a/Lib/fontTools/feaLib/lexer.py
+++ b/Lib/fontTools/feaLib/lexer.py
@@ -142,10 +142,12 @@ class Lexer(object):
             return (Lexer.SYMBOL, cur_char, location)
         if cur_char == '"':
             self.pos_ += 1
-            self.scan_until_('"\r\n')
+            self.scan_until_('"')
             if self.pos_ < self.text_length_ and self.text_[self.pos_] == '"':
                 self.pos_ += 1
-                return (Lexer.STRING, text[start + 1:self.pos_ - 1], location)
+                # strip newlines embedded within a string
+                string = re.sub("[\r\n]", "", text[start + 1:self.pos_ - 1])
+                return (Lexer.STRING, string, location)
             else:
                 raise FeatureLibError("Expected '\"' to terminate string",
                                       location)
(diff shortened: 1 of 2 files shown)


Lib/fontTools/feaLib/lexer.py
```diff
@@ -142,13 +142,24 @@ def next_(self):
...
+                                          location)
+                cur_char = self.text_[self.pos_]
                 self.pos_ += 1
-                return (Lexer.STRING, text[start + 1:self.pos_ - 1], location)
-            else:
-                raise FeatureLibError("Expected '\"' to terminate string",
-                                      location)
+                if cur_char == '"':
+                    strings.append(text[prev_pos:self.pos_ - 1])
+                    return (Lexer.STRING, "".join(strings), location)
```
brawer: Wouldn’t it be easier to do self.scan_util('"') and then replace \r and \n by '' ?

anthrotype: I can have a look tomorrow. If you want, you can go on and fix it yourself, no prob ;)

anthrotype:
@brawer I simplified the code a bit, like you suggested, and added a test case. It should be good to go now. Feel free to add to it in case I missed something.

brawer:
👍

Issue #536, opened 2016-03-03
file created too early in TTFont.save()


justvanrossum:
If an exception occurs during compilation, an empty file remains. This isn't so bad per se, but when overwriting the original file this destroys the original, which is rather nasty. It can probably be fixed by moving the file creation block to after the SFNTWriter is done.

behdad:
Or maybe create the output in a tmp file and only rename it if compilation was successful, and remove it otherwise.

In general the case of output and input filenames being equal is dangerous, so I wouldn't mind if we disallow that.

behdad:
> But how to best modify in place?

What do you mean?

madig:
> > But how to best modify in place?
> 
> What do you mean?

I often want to do an op on a font without creating a new file, is that discouraged?

behdad:
> I often want to do an op on a font without creating a new file, is that discouraged?

That's fine. Just that internally it creates a new file then rename it onto the old file name. In the operating system it creates a new file, such that if the process was still reading the old file (the source), it can keep doing that uninterrupted.

PR #581, opened 2016-04-17
[WIP] TTFont: Use mmap if possible

behdad:
As discussed here:
https://github.com/behdad/fonttools/issues/580#issuecomment-211110222

Before:
$ python -m timeit 'from fontTools.ttLib import TTFont; TTFont("sazanami-gothic.ttf")'
10 loops, best of 3: 66.9 msec per loop

After:
$ python -m timeit 'from fontTools.ttLib import TTFont; TTFont("sazanami-gothic.ttf")'
10000 loops, best of 3: 110 usec per loop

That's a 600x speedup!

Fixes https://github.com/behdad/fonttools/issues/482

HOWEVER, it reintroduces https://github.com/behdad/fonttools/issues/302
Or worse, we'll crash when overwriting:

$ cp Lobster.ttf t.ttf
$ ./ttx -o ./t.ttf ./t.ttf
Dumping "./t.ttf" to "./t.ttf"...
Dumping 'GlyphOrder' table...
Bus error (core dumped)

IMO we should fix this by changing both XML and font output routines to,
instead of calling open, use os.tmpfile(), create output in a new file
and then rename it to the final destination.  It has the benefit of not
leaving a half-written output file behind if an exception occurs.  The
tempfile.NamedTemporaryFile also comes handy.

While checking those out, tempfile.SpooledTemporaryFile also comes handy,
when it's available, to replace BytesIO in the following part of TTFont.save():

```
# write to a temporary stream to allow saving to unseekable streams
tmp = BytesIO()
```

Looks like we need to start misc.fileTools to abstract the details away.

### DIFF
--- a/Lib/fontTools/ttLib/__init__.py
+++ b/Lib/fontTools/ttLib/__init__.py
@@ -47,12 +47,18 @@ from fontTools.misc.loggingTools import deprecateArgument, deprecateFunction
 import os
 import sys
 import logging
+try:
+	import mmap
+	class _mmap(mmap.mmap):
+		# Subclass native type just so we can assign a new attribute ('name') to instances.
+		pass
+except ImportError:
+	pass
 
 
 log = logging.getLogger(__name__)
 
 class TTLibError(Exception): pass
-
 
 class TTFont(object):
 
@@ -169,8 +175,13 @@ class TTFont(object):
 			# assume "file" is a readable file object
 			closeStream = False
 		if not self.lazy:
-			# read input file in memory and wrap a stream around it to allow overwriting
-			tmp = BytesIO(file.read())
+			# Try mmap()ing it first:
+			try:
+				# TODO: Port to Windows version of mmap's module
+				tmp = _mmap(file.fileno(), 0, flags=mmap.MAP_PRIVATE, prot=mmap.PROT_READ)
+			except mmap.error:
+				# Read into memory and wrap a stream around it to allow overwriting
+				tmp = BytesIO(file.read())
 			if hasattr(file, 'name'):
 				# save reference to input file name
 				tmp.name = file.name
--- a/Lib/fontTools/ttLib/__init__.py
+++ b/Lib/fontTools/ttLib/__init__.py
@@ -44,6 +44,7 @@ Dumping 'prep' table...
 from __future__ import print_function, division, absolute_import
 from fontTools.misc.py23 import *
 from fontTools.misc.loggingTools import deprecateArgument, deprecateFunction
+import io
 import os
 import sys
 import logging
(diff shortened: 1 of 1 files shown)


behdad:
This is not ready to be merged; uploaded for discussion.

That's enough playing with this issue instead of getting started on filing taxes...  @anthrotype Feel free to finish it!

moyogo:
Wow! That’s impressive.

justvanrossum:
I've never used the mmap module, but your timing example can't be right: mmap reads lazily, so your before and after test pretty much compare reading the entire file into memory vs. not reading the entire file into memory. A fair comparison would be to parse all glyph data after creating the TTFont().

justvanrossum:
Also: it's a little (read: very) scary that mmap can cause a core dump...

justvanrossum:
Ah ok, I think I was misunderstanding what you're trying to do. But I think the whole exercise is a little backwards, and your proposal to use temp files (so we can overwrite loaded fonts) is probably much better. (It would also solve #536.)

behdad:
Right, the goal of this exercise was to make the font loading be lean again, like it used to be when we didn't read the whole file.  @justvanrossum you are right that when we do the temp file, then we can go back to not reading the font file at all, like before.  That said, mmap() is still a more efficient way than reading the file, so I suggest we keep this as well.  I know this is not common in Python, but since there is a mmap module we might as well use it, as it saves a copy.

justvanrossum:
Not at all convinced of using mmap in this case. How much speed can be gained? Let's measure that. I doubt data copying and file io is contributing much to any lack of performance. mmap is less portable, and can cause core dumps (that's actually pretty disgusting all by itself). Remember, even a huge TTF is a small file by today's standards.

anthrotype:
I haven't used mmap either, and would like to read a bit more about it.
As far as I understand, it uses the operating system's virtual memory to access the data on the filesystem directly. It feels a bit risky. Instead of making sys calls like open, read etc., it puts the file data into memory and allows one to directly manipulate that in memory.
I quite like the fact that it works the same as both a bytearray and a file object though. However, I see it has different interfaces for Unix and Windows. And It feels quite low-level in general... 😕

justvanrossum:
Exactly. Btw. FontTools has never been about performance. It's about being flexible and reliable in accessng and manipulating font data. Sure, if things can be rewritten in a way that improves performance, great, but it should not make the code more complicated. Might as well rewrite the whole thing in C++ then, and that's not why we're here.

behdad:
I don't insist in this going on.  Definitely not.  But want to share my perspective as well:

- HarfBuzz and FreeType both mmap() the font file and use it.  If they didn't, every app on an Android device would have consumed many megabytes more memory.

- I understand FontTools was not originally design for the kind of work that is latency-critical.  However, Google Fonts *is* using it to subset fonts on demand, and when I added the subset module two years ago, I significantly sped up many parts, most in loading.

I do want to measure this more as well.

justvanrossum:
If the use of mmap can be abstract and reliable (no core dumps!) and portability isn't at stake, and brings us measurable advantages, then by all means, go ahead.

anthrotype:
looks like mmap does not work with standard I/O streams:
http://stackoverflow.com/questions/6510819/using-mmap-with-popen

(we haven't implemented it yet, but I think it's nice if fonttools scripts could read/write from stdin/stdout -- binary data, not just text).

behdad:
> (we haven't implemented it yet, but I think it's nice if fonttools scripts could read/write from stdin/stdout -- binary data, not just text).

That already works in my implementation.   It falls back to the code that was there before.

brawer:
Can this be merged?

brawer:
Ugh, I’ve pressed the “Update branch” button in GitHub, assuming this would rebase to master; instead, it merged master *into* the branch and now the branch is messed up.

Anyhow, what should be done with this pull request? Personally I don’t see any disadvantages with trying to mmap (and falling back to normal I/O in case of failure), as the pull request already does. So I think we should merge it. My only concern is what happens on Windows. @anthrotype, do you know what it would take to make this work on Windows? See the TODO in the code.

behdad:
> Anyhow, what should be done with this pull request if it actually worked? Personally I don’t see disadvantages with trying to mmap (and falling back to normal file I/O in case of failure), like the pull request already does. So I think we should merge it. My only concern is what happens on Windows. @anthrotype, do you know what it would take to make this work on Windows? See the TODO in the code.

Windows is not a concern.  We either implement using mmap on Windows, or it will just fallback to non-mmap.

This cannot be merged yet however, as we need to clean up all places where we write to files to write to a temp file and rename it.  Right now, with the mmap branch, if source and destination of ttx are the same file, we crash.  That's not acceptable.

behdad:
I still think this is a good idea, but closing since others don't :).

PR #582, opened 2016-04-18
[varLib] Unit test for loading .designspace files

brawer:
I wrote this just to understand what the code does, but it might be useful as a test case.


behdad:
Thanks.  Yes, we should integrate this.

behdad:
We can move the designspace loader into it's own submodule.

PR #584, opened 2016-04-21
[SVG] correct __init__ method

miguelsousa:
Something simple like this was failing:
```python
from fontTools import ttLib
fontFilePath = 'font-with-SVG-table.ttf'
font = ttLib.TTFont(fontFilePath)
print font['SVG ']
```

I took a clue from the [CFF table code](https://github.com/behdad/fonttools/blob/master/Lib/fontTools/ttLib/tables/C_F_F_.py#L10).


anthrotype:
Thanks Miguel!

PR #586, opened 2016-04-22
[sbix] correct __init__ method

miguelsousa:
Similar to #584.
In this case, making an **sbix** table from scratch required this `_s_b_i_x.table__s_b_i_x('sbix')`, which is redundant. With this change it's only necessary to do this `_s_b_i_x.table__s_b_i_x()`.


behdad:
Thanks Miguel.  We should convert all tables.

miguelsousa:
@behdad you mean every table should have an **__init__** method like this?
```python
def __init__(self, tag=None):
	DefaultTable.DefaultTable.__init__(self, tag)
```

behdad:
No.  But any table that has an `__init__` method, should work this way.

justvanrossum:
Uhm, this is why there's the newTable() function in ttLib. You should never have to import those ugly submodule names directly.

miguelsousa:
@behdad #587
@justvanrossum wasn't aware of that. Thanks!

PR #589, opened 2016-04-25
glyph class as list and tuple instead of set and frozenset

### LINKED ISSUE #504: [feaLib] Glyph classes should be lists, not sets
The following is currently rejected by feaLib. The root cause is that glyph classes are currently represented as sets; should be lists instead.

Input:
```
feature test {
    sub [a b c d] by [T E S T];
} test;
```

Expected output:
```
        <SingleSubst index="0">
          <Substitution in="a" out="T"/>
          <Substitution in="b" out="E"/>
          <Substitution in="c" out="S"/>
          <Substitution in="d" out="T"/>
        </SingleSubst>
```

Observed:
```
bug.fea:2:5: Expected a glyph class with 4 elements after "by", but found a glyph class with 3 elements
```

moyogo:
Fixes #504

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -34,7 +34,7 @@ class GlyphName(Expression):
         self.glyph = glyph
 
     def glyphSet(self):
-        return frozenset((self.glyph,))
+        return (self.glyph,)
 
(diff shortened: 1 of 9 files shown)


brawer:
👍

PR #594, opened 2016-04-28
varLib: use cElementTree if available to parse designspace

anthrotype:
The C implementation of ElementTree is included in CPython since 2.5.
It's got the same API and is up to 20 times faster than the Python implementation.
We'd better use it.


anthrotype:
go home Travis, you're drunk.

PR #597, opened 2016-04-29
symfont: set dummify=False to fix "NameError: global name 'P' is not defined"

### LINKED ISSUE #595: Error running symfont: global name 'P' is not defined
Running into an error with the symfont.py in Snippets.  Regardless of the font file or glyph arguments, it dies (AFAICT) right after it enters _glyph.draw(transformer) the first time.  This is on Python 2.7.10, on Ubuntu 15.10.  Though I've now tried it both in a virtual environment and using the normal system Python, on two different machines, with identical results.

Traceback:

> upem 1000
>
>glyph: e
>Traceback (most recent call last):
>  File "symfont.py", line 248, in <module>
>    main(sys.argv[1:])
>  File "symfont.py", line 244, in main
>    test(font.getGlyphSet(), font['head'].unitsPerEm, glyphs)
>  File "symfont.py", line 234, in test
>    print ("%s: %g" % (item, getattr(stats, item)))
>  File "symfont.py", line 151, in _penAttr
>    self._glyph.draw(transformer)
>  File "/home/nate/code/virtualenv/sympy/fonttools/Snippets/fontTools/ttLib/__init__.py", line 748, in draw
>    glyph.draw(pen, glyfTable, offset)
>  File "/home/nate/code/virtualenv/sympy/fonttools/Snippets/fontTools/ttLib/tables/_g_l_y_f.py", line 978, in draw
>    pen.qCurveTo(*contour[:nextOnCurve])
>  File "/home/nate/code/virtualenv/sympy/fonttools/Snippets/fontTools/pens/transformPen.py", line 42, in qCurveTo
>    self._outPen.qCurveTo(*points)
>  File "/home/nate/code/virtualenv/sympy/fonttools/Snippets/fontTools/pens/basePen.py", line 278, in qCurveTo
>    _qCurveToOne(pt1, pt2)
>  File "symfont.py", line 94, in _qCurveToOne
>    self._segment(p0,p1,p2)
>  File "symfont.py", line 83, in _segment
>    self.value += self._funcs[len(P) - 1](P)
>  File "<string>", line 1, in <lambda>
>NameError: global name 'P' is not defined

I'm pretty confident this isn't something totally obvious, since I could reproduce it in a virtualenv, but I don't know.  Any suggestions?

anthrotype:
This should fix #595.

I'm not familiar with `sympy`. However, I noticed that if I call `lambdify` with argument `dummify=False`, then symfont works.

From http://docs.sympy.org/dev/modules/utilities/lambdify.html#sympy.utilities.lambdify.lambdify

> The default behavior is to substitute all arguments in the provided expression with dummy symbols... Call the function with dummify=False if dummy substitution is unwanted (and 'args' is not a string).

@behdad could you have a look?

Thanks @n8willis for trying out the symfont module and reporting the issue.

### DIFF
--- a/Snippets/symfont.py
+++ b/Snippets/symfont.py
@@ -49,7 +49,7 @@ def green(f, Bezier=BezierCurve[n]):
 	return sp.integrate(f2 * sp.diff(Bezier[0], t), (t, 0, 1))
 
 def lambdify(f):
-	return sp.lambdify(Psymbol, f)
+	return sp.lambdify(Psymbol, f, dummify=False)
 
 class BezierFuncs(object):
 


n8willis:
Confirming that this fixes #595 .  Thanks!

behdad:
Weird.  My sp.lambdify docstring also mentions dummify, but when I try it I get TypeError: lambdify() got an unexpected keyword argument 'dummify'

n8willis:
The sympy version might make a difference, of course; this:
https://github.com/sympy/sympy/issues/8536
suggests the dummify feature existed but was a bit broken prior to 0.7.6.  It could've broken again.  My system is using 1.0 via pip.

PR #598, opened 2016-04-29
Make sure component coordinates are integers

khaledhosny:
If they are not integers compiling the glyf table will fail.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -1070,11 +1070,13 @@ class GlyphComponent(object):
 				data = data + struct.pack(">HH", self.firstPt, self.secondPt)
 				flags = flags | ARG_1_AND_2_ARE_WORDS
 		else:
+			x = int(round(self.x))
+			y = int(round(self.y))
 			flags = flags | ARGS_ARE_XY_VALUES
-			if (-128 <= self.x <= 127) and (-128 <= self.y <= 127):
-				data = data + struct.pack(">bb", self.x, self.y)
+			if (-128 <= x <= 127) and (-128 <= y <= 127):
+				data = data + struct.pack(">bb", x, y)
 			else:
-				data = data + struct.pack(">hh", self.x, self.y)
+				data = data + struct.pack(">hh", x, y)
 				flags = flags | ARG_1_AND_2_ARE_WORDS
 
 		if hasattr(self, "transform"):


anthrotype:
wait, but doesn't the `Glyph.compileCoordinates` method rounds float coordinates before packing them?

https://github.com/behdad/fonttools/blob/master/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L587-L590

anthrotype:
the auto-rounding of floats was explicitly removed from the `ttGlyphPen` in https://github.com/behdad/fonttools/commit/f02e771527ed080eb5571be78e7217b6ce999871

khaledhosny:
Hmm, I’m hitting it in https://github.com/behdad/fonttools/blob/master/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L1077, I don’t see any code rounding component coordinates.

anthrotype:
Oh the components!
In that case, I think the rounding should also take place in the `_g_l_y_f` table module, like with it is with coordinates.

anthrotype:
maybe we could add a method called `roundOffset()` to the `GlyphComponent` object (which returns `int(round(x)), int(round(y))`), and use that in `GlyphComponent.compile` just before packing up the struct?
That's basically what we already do for `GlyphCoordinates`.

Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -95,7 +95,8 @@ def glyph(self, componentFlags=0x4):
             component.glyphName = glyphName
             if transformation[:4] != (1, 0, 0, 1):
                 component.transform = (transformation[:2], transformation[2:4])
-            component.x, component.y = transformation[4:]
+            component.x = int(transformation[4])
+            component.y = int(transformation[5])
```
anthrotype: instead of truncating the float with `int(f)`, I'd rather do `int(round(f))`, no?

justvanrossum: Absolutely!

khaledhosny:
Implemented both of @anthrotype’s suggestions.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -1070,11 +1070,12 @@ def compile(self, more, haveInstructions, glyfTable):
 				data = data + struct.pack(">HH", self.firstPt, self.secondPt)
 				flags = flags | ARG_1_AND_2_ARE_WORDS
 		else:
+			x, y = self.roundOffset()
```
anthrotype: I would think that `self.roundOffset()` would simply round the `self.x` and `self.y` in place without returning a value.
If the method does return a value, I would maybe call it `self.getRoundOffsets()`.
BTW, the `GlyphCoordinates.toInt()` method does not have a return value, for example.

But these are details anyway. Thanks @khaledhosny!

khaledhosny: I just thought may be we want to keep the original coordinates for other operations e.g. when saving to XML, and having the saved file be different whether one saved before or after compiling the table seems unpredictable IMO.

anthrotype: I see. But again that is what is happening already with coordinates. We may think to revise that too, not sure... /cc @behdad @justvanrossum

behdad: Unfortunately in many places we modify the font when writing it.  That's a bigger problem.

For this particular case I'd say it's simple enough that a method is not needed at al.

anthrotype:
LGTM

PR #600, opened 2016-05-04
py23: always use the iterator version of (x)range

anthrotype:
On python 2, both `range` and `xrange` will be aliases of `xrange`. On Python 3, both `range` and `xrange` will be aliases of `range`.

So in both 2 and 3, either range or xrange will behave like the Python 3 version of range, i.e. as iterators.

comments?

### DIFF
--- a/Lib/fontTools/misc/py23.py
+++ b/Lib/fontTools/misc/py23.py
@@ -6,7 +6,7 @@ import sys
 
 __all__ = ['basestring', 'unicode', 'unichr', 'byteord', 'bytechr', 'BytesIO',
 		'StringIO', 'UnicodeIO', 'strjoin', 'bytesjoin', 'tobytes', 'tostr',
-		'tounicode', 'Tag', 'open']
+		'tounicode', 'Tag', 'open', 'range', 'xrange']
 
 
 try:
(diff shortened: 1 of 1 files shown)


moyogo:
LGTM

justvanrossum:
Not sure why you provide xrange for Py3? I don't think people should be encouraged to use that at all.

anthrotype:
I don't know.. I left it there just in case. But I can remove it if you want.

justvanrossum:
I think the idea is to port as many Py3 things to Py2 so it's easy to keep the code base compatible with both. The other way round seems wrong and unnecessary to me. You're creating more possible noise this way... Maybe provide a dummy xrange() that raises an exception, so that Py2 people quickly realize they shouldn't use it in fonttools?

behdad:
I agree we want to set xrange to be the same across both versions, to reduce variation.  It can, indeed, be set to None or NotImplemented or any other thing that would not work on py2

anthrotype:
the way it is now, it uses iterator in any case, for either range or xrange, on both 2 and 3.
yes, we can drop xrange if we want and have it raise instead.

anthrotype:
how about this? a796b03

I keep the `xrange` name defined in `py23` module, but I make it raise a `Py23Error` (subclass of `NotImplementedError`) whenever one attempts to use it in code.

behdad:
lgtm

PR #603, opened 2016-05-11
Broadcast timing parts as well as final formatted message

rsheeter:
The change to use Python logging makes Google Fonts custom Logger for timing subsetting break. We can easily capture logging messages via a Handler but having only the final formatted message rather than the parts is inconvenient.

This seemed like the smallest change that would make server collection of aggregate timing easy again. We could alternately try to let Timer generally rely more on logger to do the formatting and thus avoid formatting messages that won't be written if the log isn't enabled at level.


anthrotype:
Do you mind adding a little test case? (if it's not too much trouble)

rsheeter:
Sure thing.

anthrotype:
Thank you! 👍

PR #606, opened 2016-05-15
subset: must 'draw' charstrings to set 'width' attribute before calling 'drop_hints'

anthrotype:
It seems that the horizontal advance of hinted charstrings gets lost when the CFF hinting is stripped with '--no-hinting' option, as it was noted by @miguelsousa in this comment:

https://github.com/behdad/fonttools/issues/343#issuecomment-156234918

The reason is `T2CharString` objects only get assigned a 'witdh' attribute after their 'draw' method is called.

The subsetter's `drop_hints` function attempts to insert the width back at the beginning of the de-hinted charstring's program, but can do that only if the charstring does have a 'width' attribute:

https://github.com/behdad/fonttools/blob/master/Lib/fontTools/subset/__init__.py#L1908

Hence, we must 'draw' the charstring (with a `NullPen`) before stripping the hints.

(I will add a quick test tomorrow.)

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -7,7 +7,7 @@ from fontTools.misc.py23 import *
 from fontTools import ttLib
 from fontTools.ttLib.tables import otTables
 from fontTools.misc import psCharStrings
-from fontTools.pens.boundsPen import BoundsPen
+from fontTools.pens.basePen import NullPen
 from fontTools.misc.loggingTools import Timer
 import sys
 import struct
@@ -1847,8 +1847,8 @@ def prune_pre_subset(self, font, options):
                 private = font.Private
             dfltWdX = private.defaultWidthX
             nmnlWdX = private.nominalWidthX
-            pen = BoundsPen(None)
-            c.draw(pen) # this will set the charstring's width
+            pen = NullPen()
+            c.draw(pen)  # this will set the charstring's width
             if c.width != dfltWdX:
                 c.program = [c.width - nmnlWdX, 'endchar']
             else:
@@ -2207,7 +2207,9 @@ def prune_post_subset(self, options):
                 subrs = getattr(c.private, "Subrs", [])
                 decompiler = _DehintingT2Decompiler(css, subrs, c.globalSubrs)
                 decompiler.execute(c)
+            pen = NullPen()
             for charstring in css:
+                charstring.draw(pen)  # this will set the charstring's width
                 charstring.drop_hints()
             del css
 
--- a/Lib/fontTools/feaLib/builder_test.py
+++ b/Lib/fontTools/feaLib/builder_test.py
@@ -110,7 +110,7 @@ class BuilderTest(unittest.TestCase):
         expected = self.read_ttx(expected_ttx)
         if actual != expected:
             for line in difflib.unified_diff(
-                    expected, actual, fromfile=path, tofile=expected_ttx):
+                    expected, actual, fromfile=expected_ttx, tofile=path):
                 sys.stdout.write(line)
             self.fail("TTX output is different from expected")
 
--- a/Lib/fontTools/subset/subset_test.py
+++ b/Lib/fontTools/subset/subset_test.py
@@ -59,7 +59,7 @@ class SubsetTest(unittest.TestCase):
         expected = self.read_ttx(expected_ttx)
         if actual != expected:
             for line in difflib.unified_diff(
-                    expected, actual, fromfile=path, tofile=expected_ttx):
+                    expected, actual, fromfile=expected_ttx, tofile=path):
                 sys.stdout.write(line)
             self.fail("TTX output is different from expected")
 
--- a/Lib/fontTools/subset/subset_test.py
+++ b/Lib/fontTools/subset/subset_test.py
@@ -194,6 +194,44 @@ class SubsetTest(unittest.TestCase):
         # unknown tables are kept if --passthrough-tables option is passed
         self.assertTrue(unknown_tag in subsetfont)
 
+    def test_drop_hints_CFF(self):
+        _, fontpath = self.compile_font(self.getpath("TestOTF-Regular.ttx"), ".otf")
+        subsetpath = self.temp_path(".otf")
+        subset.main([fontpath, "--no-hinting", "--notdef-outline",
+                     "--output-file=%s" % subsetpath, "*"])
+        subsetfont = TTFont(subsetpath)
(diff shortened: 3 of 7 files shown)


anthrotype:
I should confess, I'm not 100% sure this is the right thing to do in all cases, so I'd appreciate comments.
Thanks

/cc @behdad

anthrotype:
I added some unit tests for the `--no-hinting` and `--desubroutinize` options. Without commit e03a538d5b98862e2c7a412c99965711b829a0bc (which sets the charstring's `width` attribute), the CFF-related tests fail -- as expected.

I also added a new `dehint.py` script to the Snippets folder, which shows how to use the `subset` module to remove the hinting data from CFF or TT-flavored OpenType fonts.

I guess eventually we want the CFF table to "know" how to drop hints or desubroutinize itself, without needing to be monkey-patched by the subset module. After all, the Glyph objects in the glyf table can "trim" the TrueType instructions.

anthrotype:
For CFF fonts, the `dehint.py` snippet drops the hints and desubroutinizes at the same time.
I did that because, according to https://github.com/behdad/fonttools/issues/144:

> There's a bug in the hintmask stripping that leaves residual bytes, which makes OTS reject the font. The new --desubroutinize option works around that bug.

However, I was not able to reproduce the issue... I will try tomorrow.

behdad:
I don't recommend accessing the internals of the subsetter.  That's too fragile.  Instead, we should add more preset options to subsetter, or some other scheme.  The rest looks good to me.  Thanks.

Lib/fontTools/subset/__init__.py
```diff
@@ -2191,7 +2191,9 @@ def prune_post_subset(self, options):
                 subrs = getattr(c.private, "Subrs", [])
                 decompiler = _DehintingT2Decompiler(css, subrs, c.globalSubrs)
                 decompiler.execute(c)
+            pen = NullPen()
             for charstring in css:
+                charstring.draw(pen)  # this will set the charstring's width
```
behdad: Actually, upon further inspection, doesn't the decompiler.execute(c) set the width?

behdad: It doesn't, but will set the width on decompiler.  We can check and copy there and avoid an extra draw().  WDYT?

anthrotype:
I'd like to revive this PR. I rebased it on master, and I removed the `dehint.py` snippet that was accessing the subsetter's internals.

@behdad I am not quite sure what you meant in your [previous comment](https://github.com/fonttools/fonttools/pull/606#r63813953):

> doesn't the decompiler.execute(c) set the width? [...]
> It doesn't, but will set the width on decompiler. We can check and copy there and avoid an extra draw().

I'm not familiar with the details `psCharStrings` module.
I can see the `SimpleT2Decompiler` does not decompile the charstring's width, whereas the `T2OutlineExtractor` does do that in its `popallWidth` method.

Did you mean we should make the `subset._DehintingT2Decompiler` do the same as what the `T2OutlineExtractor` does (ie. pop and set the width attribute), so we spare the extra `draw`?

I have this patch on the current PR branch, but I want to clarify this with you first before I commit:

```diff
diff --git a/Lib/fontTools/subset/__init__.py b/Lib/fontTools/subset/__init__.py
index 0f1340a..fa755a0 100644
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -1981,11 +1981,34 @@ class _DehintingT2Decompiler(psCharStrings.SimpleT2Decompiler):
             self.has_hintmask = False
         pass
 
-    def __init__(self, css, localSubrs, globalSubrs):
+    def __init__(self, css, localSubrs, globalSubrs, nominalWidthX,
+                 defaultWidthX):
         self._css = css
         psCharStrings.SimpleT2Decompiler.__init__(self,
                                                   localSubrs,
                                                   globalSubrs)
+        self.nominalWidthX = nominalWidthX
+        self.defaultWidthX = defaultWidthX
+
+    def reset(self):
+        psCharStrings.SimpleT2Decompiler.reset(self)
+        self.gotWidth = 0
+        self.width = 0
+
+    def popallWidth(self, evenOdd=0):
+        args = self.popall()
+        if not self.gotWidth:
+            if evenOdd ^ (len(args) % 2):
+                self.width = self.nominalWidthX + args[0]
+                args = args[1:]
+            else:
+                self.width = self.defaultWidthX
+            self.gotWidth = 1
+        return args
+
+    def countHints(self):
+        args = self.popallWidth()
+        self.hintCount = self.hintCount + len(args) // 2
 
     def execute(self, charString):
         old_hints = charString._hints if hasattr(charString, '_hints') else None
@@ -1993,6 +2016,9 @@ class _DehintingT2Decompiler(psCharStrings.SimpleT2Decompiler):
 
         psCharStrings.SimpleT2Decompiler.execute(self, charString)
 
+        # set the width, so we can insert it back when we 'drop_hints'
+        charString.width = self.width
+
         hints = charString._hints
 
         if hints.has_hint or hints.has_hintmask:
@@ -2011,6 +2037,18 @@ class _DehintingT2Decompiler(psCharStrings.SimpleT2Decompiler):
         if old_hints:
             assert hints.__dict__ == old_hints.__dict__
 
+    def op_rmoveto(self, index):
+        self.popallWidth()
+
+    def op_hmoveto(self, index):
+        self.popallWidth(1)
+
+    def op_vmoveto(self, index):
+        self.popallWidth(1)
+
+    def op_endchar(self, index):
+        self.popallWidth()
+
     def op_callsubr(self, index):
         subr = self.localSubrs[self.operandStack[-1]+self.localBias]
         psCharStrings.SimpleT2Decompiler.op_callsubr(self, index)
@@ -2205,11 +2243,11 @@ def prune_post_subset(self, options):
                 c,sel = cs.getItemAndSelector(g)
                 c.decompile()
                 subrs = getattr(c.private, "Subrs", [])
-                decompiler = _DehintingT2Decompiler(css, subrs, c.globalSubrs)
+                decompiler = _DehintingT2Decompiler(css, subrs, c.globalSubrs,
+                                                    c.private.nominalWidthX,
+                                                    c.private.defaultWidthX)
                 decompiler.execute(c)
-            pen = NullPen()
             for charstring in css:
-                charstring.draw(pen)  # this will set the charstring's width
                 charstring.drop_hints()
             del css
```

anthrotype:
The issue of the CFF advance widths being dropped with `--dehint` option is quite bad, so I shall merge this PR as is for now, with the extra call to `draw()`. We can optimize it later.

PR #608, opened 2016-05-19
Check for file attribute before deleting it

khaledhosny:
I don’t know why this is happening, but I’m getting attribute error when
I build https://github.com/khaledhosny/mada with compreffor
optimisations on.

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -2251,7 +2251,8 @@ def prune_post_subset(self, options):
                 local_subrs = subrs
 
             subrs.items = [subrs.items[i] for i in subrs._used]
-            del subrs.file
+            if hasattr(subrs, 'file'):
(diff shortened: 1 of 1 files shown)


adrientetar:
Was it with Python 2 or 3?

khaledhosny:
Python 3.

behdad:
This might happen if CFF is generated in Python, not by reading a file...
No?

On Thu, May 19, 2016 at 3:28 PM, Khaled Hosny <notifications@github.com>
wrote:

> Python 3.
>
> —
> You are receiving this because you modified the open/close state.
> Reply to this email directly or view it on GitHub
> <https://github.com/behdad/fonttools/pull/608#issuecomment-220463268>
>



-- 
behdad
http://behdad.org/

khaledhosny:
Yes, generated from Python here and not read from file.

PR #614, opened 2016-06-01
[feaLib]: fix unescape_mac_name_string() and unescape_windows_name_string()

moyogo:


anthrotype:
good catch! thanks :)

PR #618, opened 2016-06-09
Area and Perimeter pens

anthrotype:
see https://github.com/behdad/fonttools/issues/617#issuecomment-225029592

### DIFF
--- a/Lib/fontTools/pens/areaPen.py
+++ b/Lib/fontTools/pens/areaPen.py
@@ -53,3 +53,8 @@ class AreaPen(BasePen):
         p0 = self._getCurrentPoint()
         self.value += cubic_curve_area(p0, p1, p2, p3)
         self.value += polygon_area(p0, p3)
+
+    def _qCurveToOne(self, pt1, pt2):
+        p0 = self._getCurrentPoint()
+        self.value += quadratic_curve_area(p0, p1, p2)
+        self.value += polygon_area(p0, p2)
--- a/Lib/fontTools/pens/areaPen.py
+++ b/Lib/fontTools/pens/areaPen.py
@@ -1,0 +1,57 @@
+"""Calculate the area of a glyph."""
+
+import math
+from fontTools.pens.basePen import BasePen
+from collections import namedtuple
+
+
+P = namedtuple('P', ('x', 'y'))
+
+
+def distance(p0, p1):
+    return math.hypot(p0.x - p1.x, p0.y - p1.y)
+
+
+def interpolate(p0, p1, t):
+    return P(p0.x * (1 - t) + p1.x * t, p0.y * (1 - t) + p1.y * t)
+
+
+def polygon_area(p0, p1):
+    return (p1.x - p0.x) * (p1.y + p0.y) * 0.5
+
+
+def quadratic_curve_area(p0, p1, p2):
+    new_p2 = interpolate(p2, p1, 2.0 / 3)
+    new_p1 = interpolate(p0, p1, 2.0 / 3)
+    return cubic_curve_area(p0, new_p1, new_p2, p2)
+
+
+def cubic_curve_area(p0, p1, p2, p3):
+    x0, y0 = p0.x, p0.y
+    x1, y1 = p1.x - x0, p1.y - y0
+    x2, y2 = p2.x - x0, p2.y - y0
+    x3, y3 = p3.x - x0, p3.y - y0
+    return (
+        x1 * (   -   y2 -   y3) +
+        x2 * (y1        - 2*y3) +
+        x3 * (y1 + 2*y2       )
+    ) * 0.15
+
+
+class AreaPen(BasePen):
+
+    def __init__(self, glyphset):
+        BasePen.__init__(self, glyphset)
+        self.value = 0
+
+    def _moveTo(self, p0):
+        pass
+
+    def _lineTo(self, p1):
+        p0 = self._getCurrentPoint()
+        self.value += polygon_area(P(*p0), P(*p1))
+
+    def _curveToOne(self, p1, p2, p3):
+        p0 = self._getCurrentPoint()
+        self.value += cubic_curve_area(P(*p0), P(*p1), P(*p2), P(*p3))
+        self.value += polygon_area(P(*p0), P(*p3))
--- a/Lib/fontTools/pens/areaPen.py
+++ b/Lib/fontTools/pens/areaPen.py
@@ -1,7 +1,9 @@
 """Calculate the area of a glyph."""
 
-import math
+from __future__ import print_function, division, absolute_import
+from fontTools.misc.py23 import *
 from fontTools.pens.basePen import BasePen
+import math
 
 
 def distance(p0, p1):
--- a/Lib/fontTools/pens/areaPen.py
+++ b/Lib/fontTools/pens/areaPen.py
@@ -3,11 +3,6 @@
 from __future__ import print_function, division, absolute_import
 from fontTools.misc.py23 import *
 from fontTools.pens.basePen import BasePen
-import math
-
-
-def distance(p0, p1):
-    return math.hypot(p0[0] - p1[0], p0[1] - p1[1])
 
 
 def interpolate(p0, p1, t):
--- a/Lib/fontTools/pens/areaPen.py
+++ b/Lib/fontTools/pens/areaPen.py
@@ -6,50 +6,50 @@ from fontTools.pens.basePen import BasePen
 
 
 def interpolate(p0, p1, t):
-    return (p0[0] * (1 - t) + p1[0] * t, p0[1] * (1 - t) + p1[1] * t)
+	return (p0[0] * (1 - t) + p1[0] * t, p0[1] * (1 - t) + p1[1] * t)
 
 
 def polygon_area(p0, p1):
-    return (p1[0] - p0[0]) * (p1[1] + p0[1]) * 0.5
+	return (p1[0] - p0[0]) * (p1[1] + p0[1]) * 0.5
 
 
 def quadratic_curve_area(p0, p1, p2):
-    new_p2 = interpolate(p2, p1, 2.0 / 3)
-    new_p1 = interpolate(p0, p1, 2.0 / 3)
-    return cubic_curve_area(p0, new_p1, new_p2, p2)
+	new_p2 = interpolate(p2, p1, 2.0 / 3)
+	new_p1 = interpolate(p0, p1, 2.0 / 3)
+	return cubic_curve_area(p0, new_p1, new_p2, p2)
 
 
 def cubic_curve_area(p0, p1, p2, p3):
-    x0, y0 = p0[0], p0[1]
-    x1, y1 = p1[0] - x0, p1[1] - y0
-    x2, y2 = p2[0] - x0, p2[1] - y0
-    x3, y3 = p3[0] - x0, p3[1] - y0
-    return (
-        x1 * (   -   y2 -   y3) +
-        x2 * (y1        - 2*y3) +
-        x3 * (y1 + 2*y2       )
-    ) * 0.15
+	x0, y0 = p0[0], p0[1]
+	x1, y1 = p1[0] - x0, p1[1] - y0
+	x2, y2 = p2[0] - x0, p2[1] - y0
+	x3, y3 = p3[0] - x0, p3[1] - y0
+	return (
+		x1 * (   -   y2 -   y3) +
+		x2 * (y1        - 2*y3) +
+		x3 * (y1 + 2*y2       )
+	) * 0.15
 
 
 class AreaPen(BasePen):
 
-    def __init__(self, glyphset):
-        BasePen.__init__(self, glyphset)
-        self.value = 0
+	def __init__(self, glyphset):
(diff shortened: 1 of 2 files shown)


Lib/fontTools/pens/areaPen.py
```diff
@@ -0,0 +1,57 @@
...
+
+def interpolate(p0, p1, t):
+    return P(p0.x * (1 - t) + p1.x * t, p0.y * (1 - t) + p1.y * t)
+
+
+def polygon_area(p0, p1):
+    return (p1.x - p0.x) * (p1.y + p0.y) * 0.5
+
+
+def quadratic_curve_area(p0, p1, p2):
```
behdad: This seems to be unused.  Can you hook it up to the pen?

Also, I can give you direct code calculating this, instead of up-ordering.

behdad:
WDYT about lifting PerimeterPen from symfont as well?

Lib/fontTools/pens/areaPen.py
anthrotype: > Can you hook it up to the pen?

sure

> I can give you direct code calculating this

that would be nice

anthrotype:
> WDYT about lifting PerimeterPen from symfont as well?

yeah, why not. shall I make a separate perimeterPen.py module?

anthrotype:
and I guess `distance` and `interpolate` functions should be moved to `bezierTools.py`.

anthrotype:
BTW, that `cubic_curve_area` function is just magic. I'm curious how/where did you find that? :stuck_out_tongue_winking_eye:

anthrotype:
we AppVeyor exit early if there are newer queued builds for a pull request, to avoid it taking ages to complete. A little annoyance is that we still all get an email notifications saying the build failed :-/
Please ignore that.

behdad:
> BTW, that cubic_curve_area function is just magic. I'm curious how/where did you find that? 😜

Originally, when @jamesgk joined our team, I asked him to solve the integral and write it down; I them subtracted `polygon_area()` from it and put it in a nice matrix format.  But with symfont, it can be derived:

```
In [11]: green(f=1, Bezier=BezierCurve[0])
Out[11]: 0

In [12]: green(f=1, Bezier=BezierCurve[1])
Out[12]: -P[0][0]*P[0][1]/2 - P[0][0]*P[1][1]/2 + P[0][1]*P[1][0]/2 + P[1][0]*P[1][1]/2

In [13]: green(f=1, Bezier=BezierCurve[2])
Out[13]: -P[0][0]*P[0][1]/2 - P[0][0]*P[1][1]/3 - P[0][0]*P[2][1]/6 + P[0][1]*P[1][0]/3 + P[0][1]*P[2][0]/6 - P[1][0]*P[2][1]/3 + P[1][1]*P[2][0]/3 + P[2][0]*P[2][1]/2

In [14]: green(f=1, Bezier=BezierCurve[3])
Out[14]: -P[0][0]*P[0][1]/2 - 3*P[0][0]*P[1][1]/10 - 3*P[0][0]*P[2][1]/20 - P[0][0]*P[3][1]/20 + 3*P[0][1]*P[1][0]/10 + 3*P[0][1]*P[2][0]/20 + P[0][1]*P[3][0]/20 - 3*P[1][0]*P[2][1]/20 - 3*P[1][0]*P[3][1]/20 + 3*P[1][1]*P[2][0]/20 + 3*P[1][1]*P[3][0]/20 - 3*P[2][0]*P[3][1]/10 + 3*P[2][1]*P[3][0]/10 + P[3][0]*P[3][1]/2
```

The first item is not interesting (degree-zero bezier).  The second item gives the area-under-curve of a first-degree bezier, ie. a line segment.  That's the same as the `polygon_area()` function.  The next two are area-under-curve of a quadratic and cubic bezier.  If you subtract the first-degree area from them, gives you the curve area as defined in `cubic_curve_area()`, and indeed what should go into `quadratic_curve_area()`; some manual massaging makes it look better, but that's it.

```
In [15]: green(f=1, Bezier=BezierCurve[3]) - green(f=1, Bezier=BezierCurve[1])
Out[15]: P[0][0]*P[1][1]/5 - 3*P[0][0]*P[2][1]/20 - P[0][0]*P[3][1]/20 - P[0][1]*P[1][0]/5 + 3*P[0][1]*P[2][0]/20 + P[0][1]*P[3][0]/20 - P[1][0]*P[1][1]/2 - 3*P[1][0]*P[2][1]/20 - 3*P[1][0]*P[3][1]/20 + 3*P[1][1]*P[2][0]/20 + 3*P[1][1]*P[3][0]/20 - 3*P[2][0]*P[3][1]/10 + 3*P[2][1]*P[3][0]/10 + P[3][0]*P[3][1]/2

In [16]: green(f=1, Bezier=BezierCurve[2]) - green(f=1, Bezier=BezierCurve[1])
Out[16]: P[0][0]*P[1][1]/6 - P[0][0]*P[2][1]/6 - P[0][1]*P[1][0]/6 + P[0][1]*P[2][0]/6 - P[1][0]*P[1][1]/2 - P[1][0]*P[2][1]/3 + P[1][1]*P[2][0]/3 + P[2][0]*P[2][1]/2
```

miguelsousa:
The AreaPen pen has a bug somewhere. While testing it with [this glyph](https://github.com/behdad/fonttools/issues/617#issuecomment-224823457) the value/area returned for the outer contour is either `-83368.15` or `-104561.35`, depending on its direction. I can say that the first value is wrong, because if I move the starting point to another place the returned values are now `±104561.35`.

anthrotype:
We need to write some tests.

Have you tried maybe using the AreaPen on a fonttools' TTGlyph object instead of a robofab/defcon glyph object? That would require you to compile the font to .otf  or .ttf and then load the TTFont, get its glyph set, etc.

I ask because the respective `draw` methods might be doing slightly different things...

miguelsousa:
I haven't tried that, but what I can see is that the area calculation doesn't account the last segment if it's a line. Does that help?

anthrotype:
hm, that's looks like the SegmentToPointPen adapter doing his things.

anthrotype:
or actually, the `PointToSegmentPen`, sorry. I assume you are using defcon, so the `glyph.draw` method actually calls its `drawPoints` method through the point to segment adapter pen. Try with the native fonttools glyphs, which only have `draw` for segment pens, and see if you also get unexpected results.

mashabow:
`AreaPen` gives negative value for counterclockwise contours; is this intentional?

```python
In [1]: from fontTools.pens.areaPen import AreaPen

In [2]: pen = AreaPen(None)
In [3]: pen.moveTo((0,0))
In [4]: pen.lineTo((100,0))
In [5]: pen.lineTo((50,100))
In [6]: pen.lineTo((0,0))
In [7]: pen.value
Out[7]: -5000.0

In [8]: pen = AreaPen(None)
In [9]: pen.moveTo((0,0))
In [10]: pen.lineTo((100,0))
In [11]: pen.curveTo((100,50), (0,50), (0,0))
In [12]: pen.value
Out[12]: -3000.0
```

anthrotype:
> is this intentional?

yes. Among the uses of this area pen is detecting the winding direction of a contour.
see discussion here: https://github.com/behdad/fonttools/issues/617

mashabow:
@anthrotype Uh, I mean it is more natural for singed area that positive is CCW and negative is CW, not negative is CCW and positive is CW. What do you think?

anthrotype:
math is fun: https://www.mathsisfun.com/geometry/area-irregular-polygons.html

miguelsousa:
@anthrotype the problem happens when using just fontTools. Here's a test case https://gist.github.com/miguelsousa/a67497ab2c9dbd11d8e9bb98b16a0a2f

And here's a unittest for the pen https://gist.github.com/miguelsousa/2a3537bf22d0d739382a996385ed605a

miguelsousa:
Perhaps the pen needs to handle `closePath`?

behdad:
Fixed the sign in symfont.

behdad:
Fixed closePath() issue in symfont.

anthrotype:
Thanks Behdad, as well as @mashabow for clarifying the convention about signed area and orientation, I shall fix the AreaPen in this PR accordingly.

https://en.m.wikipedia.org/wiki/Curve_orientation

And thanks Miguel for the tests. Maybe you could   a pull request with those once this one is merged?

miguelsousa:
> send a pull request with those once this one is merged?

Yes

anthrotype:
I applied the sign and closePath fixes, and derived the quadratic area function.
PTAL, thanks.

Lib/fontTools/pens/areaPen.py
```diff
@@ -0,0 +1,61 @@
...
+
+	def _qCurveToOne(self, p1, p2):
+		p0 = self._getCurrentPoint()
+		self.value += quadratic_curve_area(p0, p1, p2)
+		self.value += polygon_area(p0, p2)
+
+	def _closePath(self):
+		p0 = self._getCurrentPoint()
+		if p0 != self.__startPoint:
+			self._lineTo(self.__startPoint)
```
miguelsousa: What about this instead (for the last line)?
```python
self.value += polygon_area(p0, self.__startPoint)
```

Lib/fontTools/pens/areaPen.py
```diff
@@ -0,0 +1,61 @@
+"""Calculate the area of a glyph."""
+
+from __future__ import print_function, division, absolute_import
+from fontTools.misc.py23 import *
+from fontTools.pens.basePen import BasePen
+
+
+def interpolate(p0, p1, t):
```
miguelsousa: No longer used. OK to remove?

anthrotype:
I just applied @miguelsousa's suggestions.
The test gist now passes, provided the signs are inverted to account for the "CCW is positive, CW is negative" rule.

miguelsousa:
PR #619

jamesgk:
Dang, I should have seen this before! I made a similar pen for area calculation a while back and put it in nototools: https://github.com/googlei18n/nototools/blob/master/nototools/glyph_area_pen.py

anthrotype:
oh nooo! :scream: 
you mean, *I* should have seen that before.. It actually took me a while to figure out how to derive the quadratic formula from the sympy/symfont polynomial. 
Well, at least they match. ;)
Let me know if you want to change anything.

anthrotype:
> Originally, when @jamesgk joined our team, I asked him to solve the integral and write it down...

by the way, thanks for that, James :+1:

behdad:
I like @jamesgk's code better :D.

Another thing: do we really want to put each new pen in its own file?

behdad:
> BTW, that cubic_curve_area function is just magic

Here is, BTW, how to derive that:   By definition of integrals, "area under curve" is `$\int{y.dx}$`.  Typically we use that formula when y is a function of x, but it works even if that is not the case, and it does the right thing (subtracting when the curve goes backwards).  Since both y and x are a function of t from 0 to 1, we can rewrite that as `$\int_0^1 y(t) dx/dt dt$`.  From there, just plug in y(t), the bezier formula, and dx/dt, the derivative of the bezier formula.  You get a fifth degree polynomial in t inside the integral, which results in a sixth degree polynomial solution.  Since the integral bounds are 0 and 1, all those high-order powers of t result in 0 and 1 respectively, and we are left with a plain formula that has at most one y and one x point coordinate as coefficients in each term.  From there, you can factor and clean it up.

anthrotype:
> I like @jamesgk's code better :D.

@behdad well, I simply copy-pasted yours ;-)
But it's true, James' pen has the area functions  inlined, and has some nice docstrings.
Fell free to integrate the changes.

> Another thing: do we really want to put each new pen in its own file?

It doesn't harm to keep them separate. We could always import them in the top-level fontTools.pens module, and add them to an __all__ list of exported pen classes if that's more convenient.
But I don't mind if you wish to group them elsewhere.

justvanrossum:
Please don't import stuff in the top-level fontTools.pens module: that way you can never just import one thing without pulling in the rest.

anthrotype:
roger that

behdad:
> Please don't import stuff in the top-level fontTools.pens module: that way you can never just import one thing without pulling in the rest.

Is that true?  Right now, we do:
```
from fontTools.pens.areaPen import AreaPen
```
if there are things in toplevel fontTools.pens, are they even processed when the above runs?

And, even if we added them all in the toplevel, one can still do:
```
from fontTools.pens import AreaPen
```
and the rest won't be pulled in.  Or by pulled in you mean that they needs to be loaded by the interpretter?

FWIW I wasn't suggesting putting all in one place, but possibly putting similar or closely related ones into one module.  The problem is, loading modules is costly as well, so having one module per each 100 lines of code does slow down startup time more than loading a few longer modules.

anthrotype:
> if there are things in toplevel fontTools.pens, are they even processed when the above runs?

the top-level statements in both `fontTools.__init__` module and `fontTools.pens` module are executed, but the names are not imported in the current namespace unless, you one does `import *`

justvanrossum:
Indeed, you can't import a submodule without the parent __init__.py being executed.

> Or by pulled in you mean that they needs to be loaded by the interpretter?

Yes I meant that "unrequested" modules would then still be loaded, which is not nice.

PR #623, opened 2016-06-21
[feaLib] Fix exclude_dflt handling

jamesgk:
@brawer
Should fix https://github.com/behdad/fonttools/issues/613


jamesgk:
This is somewhat incomplete; we may want to fail when encountering a language statement when no script is set, as makeotf does.

brawer:
Thank you!

PR #624, opened 2016-06-22
Rejct language statements when the script is DFLT or undefined

brawer:
According to the OpenType Feature File Specification, section 4.b.ii,
the DFLT script cannot have any language systems other then "dflt".
Adobe's makeotf rejects such input as well.

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -615,6 +615,12 @@ class Builder(object):
                 "within \"feature %s\"" % self.cur_feature_name_, location)
         self.cur_lookup_ = None
 
+        # OpenType Feature File Specification, section 4.b.ii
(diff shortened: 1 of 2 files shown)


khaledhosny:
Any reason this have not been merged? (I know that the usefulness of this restriction is debatable, but I think we either match the spec here or get it changed).

behdad:
Note that there's been discussion on the OpenType mailing list to go in the reverse direction, ie. change spec to allow langsys in DFLT script.  There are a bunch of scenarios where this is needed.  I don't mind if we don't land this.  Or maybe just show a warning?

jamesgk:
Oh, just noticed this was still open. I actually submitted a duplicate PR a while ago, https://github.com/behdad/fonttools/pull/625, not realizing this had been opened. So at this point I think this should be closed.

PR #625, opened 2016-06-22
[feaLib] Don't set language when script is unset

jamesgk:
@brawer
Following up on https://github.com/behdad/fonttools/pull/623#issuecomment-227644699


brawer:
What do you think of https://github.com/behdad/fonttools/pull/624?

jamesgk:
Yea, I didn't see #624 until I'd already opened this PR. Anyways, I think it's also important to reset the script to DFLT with each new feature as this change does (in the tests I updated, the script was carried over from the last feature so they weren't throwing errors even when a language was set without a script).

PR #627, opened 2016-06-27
Add unload method to area pen

jamesgk:
I think this is a worthwhile convenience function for the common case
of calculating the area of many glyphs in succession.

### DIFF
--- a/Lib/fontTools/pens/areaPen.py
+++ b/Lib/fontTools/pens/areaPen.py
@@ -11,6 +11,13 @@ class AreaPen(BasePen):
 		BasePen.__init__(self, glyphset)
 		self.value = 0
 
+	def unload(self):
+		"""Return and reset the calculated area."""
+
+		area = self.value
+		self.value = 0
+		return area
+
 	def _moveTo(self, p0):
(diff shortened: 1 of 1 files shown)


behdad:
I'm not a huge fan.  I'd be more interested in adding a reset() if anything.  I personally think having users set value to zero manually is simpler and better, at least until we figure out a more uniform way.  The reason I called the value 'value' and not 'area' is because I have this vague idea about having a class of pens that calculate a value, and all put it in pen.value.  That might not be super useful, but for now let's not add new API.

anthrotype:
What if we make 'value' a property getter which returns the value and also resets it? Just sayin'

anthrotype:
Actually that would be confusing, nevermind.

behdad:
> What if we make 'value' a property getter which returns the value and also resets it? Just sayin'

The "also resets it" is a big NO based on the "minimize surprise" principle.

adrientetar:
> What if we make 'value' a property getter which returns the value and also resets it? Just sayin'

A pop() function would make more sense, FWIW. Anyway this is bikeshed.

PR #629, opened 2016-06-30
Add cmap handling to mtiLib

jamesgk:
r? @behdad


behdad:
Thanks.  Maybe, instead of passing container into the parsers, instantiate the container in the parsers and return?  Feel free to merge.

behdad:
lgtm

jamesgk:
I can't merge, I don't have write access to fonttools.

PR #632, opened 2016-07-06
Fix xml-roundtripping of CFF fonts (name-keyed & cid-keyed)

miguelsousa:
Something like this was failing,
```python

font = TTFont()
font.importXML('myFont.ttx')
font.saveXML('myNewFont.ttx')
```

### DIFF
--- a/Lib/fontTools/cffLib.py
+++ b/Lib/fontTools/cffLib.py
@@ -547,7 +547,7 @@ class CharStrings(object):
 		else:
 			if hasattr(self, 'fdArray'):
 				if hasattr(self, 'fdSelect'):
-					sel = self.fdSelect[index]  # index is not defined at this point. Read R. ?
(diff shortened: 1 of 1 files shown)


anthrotype:
Thanks Miguel. I wonder, does the inline comment have anything to do with the proposed patch?
> `# index is not defined at this point. Read R. ?`

miguelsousa:
@anthrotype no, but the second commit addresses it. Now a CID font xml-roundtrips, whereas it didn't before. I've updated the title of the PR accordingly.

anthrotype:
Thank you. I don't know enough about CFF to assess this, so I'll cc @behdad or @readroberts

readroberts:
This patch is correct and safe. The FDSelect index table is present in a CFF  if and only if the FDArray index table is present.

anthrotype:
Thank you Read!

PR #636, opened 2016-07-17
[subset] Recalculate OS/2.xAvgCharWidth

khaledhosny:


anthrotype:
It looks good to me. I wonder if a method like that could be placed in the OS/2 table class itself, so it is available not just for the subsetter.
@behdad @justvanrossum

davelab6:
Hope does this differ from what Felipe did for fontbakery?

anthrotype:
@davelab6 could you maybe put a link to Felipe's patch so we can compare?

khaledhosny:
I don’t mind even if we changed FontTools to just recalculate xAvgCharWidth when it compiles the font, I can’t think of any reason why one one need a different value than what would be calculated here.

davelab6:
OK, I think this patch is a good idea but the code is a bit too simple. Please see https://github.com/googlefonts/fontbakery/commit/c70df187719ca8984b79b536d3e0f4867ac286eb by @felipesanches and https://github.com/googlefonts/fontbakery/issues/781#issuecomment-208589366 for an explanation from @jtanx

khaledhosny:
Do we really need different handling of OS/2 < 3 versions? What is xAvgCharWidth used for anyway?

davelab6:
Spec is spec ;)

khaledhosny:
Is it? I just checked both Firefox and Chrome and they don’t seem to interpret the value differently for different to OS/2 table versions. Not to mention that the old none-sense in the spec was removed back in 2002, if people are so inclined we can change the value only for fonts with version OS/2 >= 3 and leave fonts with old table version alone in their misery.

anthrotype:
the weighting factors in the `os2ver2` spec are just provided as an "example", and they are accompanied by a warning "when comparing fonts that use different frequency of use values for font mapping".

The simpler, unweighted algorithm proposed by Khaled should work fine for practical purposes. I think it's fine to merge it as is for the subsetter. We can maybe add it later to the OS/2 module after we figure out how to pass such options to TTFont without overloading its constructor with additional kwargs.

anthrotype:
I added a `--recalc-average-width` command line option. It's on by default, but you can turn it off by passing `--no-recalc-average-width`.
Thanks Khaled.

behdad:
> I added a --recalc-average-width command line option. It's on by default

Can you please make it off by default?  That's subsetter's current defaults: fastest.  Note that recalc-bounding-box etc are also off by default.

I'm also not sure how useful this is in the subsetter.  Khaled, why do you need this?

khaledhosny:
I don’t have an actual need for this, I was just checking why fontbakery is complaining about the value in my fonts, and as it happens I’m using the subsetter to drop Latin characters I don’t need (easier than having to parse and subset feature files).

moyogo:
This comment https://github.com/adobe-fonts/source-han-sans/issues/154#issuecomment-233897031 might be relevant
> It is a known bug in GDI, when PANOSE's proportion set to monospaced, it will ignore the entire hmtx and force use OS/2.xAvgCharWidth, at least for TTF.

PR #637, opened 2016-07-19
ttLib: add _TTGlyph.height and _TTGlyph.tsb if vmtx is in font

moyogo:
This is going to be needed by [extractor](https://github.com/typesupply/extractor) for example.


anthrotype:
LGTM

moyogo:
Hmm, nevermind, one can access vmtx or VORG instead.

anthrotype:
as you prefer ;)

Issue #641, opened 2016-07-27
[feaLib] can't get include(...) to work


justvanrossum:
I somehow can't get feaLib to handle an `include()`. It gives me this exception:

`fontTools.feaLib.error.FeatureLibError: myfeatures.fea:3:22: Expected feature, languagesystem, lookup, markClass, table, or glyph class definition`

Am I missing something obvious? Run the `addfeatures.py` script in the attached folder.

[feaLibIncludeTest.zip](https://github.com/behdad/fonttools/files/385432/feaLibIncludeTest.zip)

anthrotype:
It's the semicolon.

https://github.com/behdad/fonttools/commit/7eed24725f33c76d033e40888d865e13d47622c2

/cc @brawer

anthrotype:
I found a fix, I'll send a PR shortly. Can I use your test feature files in feaLib tests?

justvanrossum:
Sure!

PR #642, opened 2016-07-27
[feaLib] ignore stray semicolons

anthrotype:
This addresses issue #641.

in https://github.com/behdad/fonttools/commit/7eed24725f33c76d033e40888d865e13d47622c2, Sascha allowed the `include` statement to not be followed by a semicolon.
However, the symbol is not eaten up by the Lexer, but is passed on to the Parser, which barks at it because it does not expect a stray SYMBOL ";" at the top level of a feature file.

I tested with makeotf, and confirmed that it ignores any extra semicolons which appear in a feature file. I guess we can do the same.

### DIFF
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -49,6 +49,8 @@ class Parser(object):
             elif self.is_cur_keyword_("valueRecordDef"):
                 statements.append(
                     self.parse_valuerecord_definition_(vertical=False))
+            elif self.cur_token_type_ is Lexer.SYMBOL and self.cur_token_ == ";":
+                continue
             else:
                 raise FeatureLibError(
                     "Expected feature, languagesystem, lookup, markClass, "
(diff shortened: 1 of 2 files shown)


anthrotype:
kind of, yeah.

Inside blocks, the semicolon is already ignored in a similar way:
https://github.com/behdad/fonttools/blob/master/Lib/fontTools/feaLib/parser.py#L1114-L1115

only in top-level main parser loop it isn't, hence the error in #641

justvanrossum:
I'm sure I'm notpicking (and I do care most about the bug being fixed, so thank you!) but your wording sounds backwards to me: the ; is (as far as I can tell) part of the statement, and the special case is to allow it to be skipped, not to allow it to be present :) (Regardless of how it's implemented.)

anthrotype:
I see what you mean. That was in fact the original implementation (semicolon as part of the the `include` statement), which was more in line with how the spec defines it:
http://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#3

Also see the discussion here:

https://github.com/behdad/fonttools/issues/552

https://github.com/adobe-type-tools/afdko/issues/114

brawer:
Thank you! makeotf also accepts a feature file with the content `;;;`, so we should too.

PR #646, opened 2016-08-06
requirements.txt: download Brotli 0.4.0 pre-compiled wheel from Github Releases

anthrotype:
This should speed up the Travis and Appveyor builds, as we don't need to compile Brotli from source, at least on OSX and Windows.
Linux will still use the .tag.gz source distribution.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
-git+https://github.com/google/brotli@v0.3.0#egg=Brotli
+--find-links https://github.com/google/brotli/releases
+brotli==0.4.0


anthrotype:
elapsed time for Travis went from 11 to 6 minutes, Appveyor from 5 to 4 minutes.

anthrotype:
Maybe we could even get rid of Python 3.3... I don't think anybody is still using that.

moyogo:
> Maybe we could even get rid of Python 3.3... I don't think anybody is still using that.

SGTM

behdad:
SGTM.

PR #653, opened 2016-08-28
[CPAL] XML enhancements

miguelsousa:
Just an improvement and an addition to the XML output of the CPAL table.


anthrotype:
LGTM, but can't merge from my phone as Appveyor has failing status (it simply timed out).

PR #655, opened 2016-08-29
Fixed encoding error in subset tool in Windows

JasonAtGithub:
Fixed encoding error in subset tool in Windows which system default codepage isn't 65001.
For Example: 
When windows default codepage is 936, the command like 'pyftsubset myfont.ttf --text-file=mytextfile.txt' will throw an exception.

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -2902,7 +2902,7 @@ def main(args=None):
             text += g[7:]
             continue
         if g.startswith('--text-file='):
-            text += open(g[12:]).read().replace('\n', '')
+            text += open(g[12:], encoding='utf-8').read().replace('\n', '')
             continue
         if g.startswith('--unicodes='):
             if g[11:] == '*':


Lib/fontTools/subset/__init__.py
```diff
@@ -2902,7 +2902,7 @@ def main(args=None):
             text += g[7:]
             continue
         if g.startswith('--text-file='):
-            text += open(g[12:]).read().replace('\n', '')
+            text += open(g[12:], encoding='utf-8').read().replace('\n', '')
```
brawer: This probably won’t work with Python 2.7. Try codecs.open().

JasonAtGithub: It works with python 2.7.10, I have tried myself.
The open method is not io.open().

brawer: Ah, I forgot that we have a [backport of python3 open()](https://github.com/behdad/fonttools/blob/master/Lib/fontTools/misc/py23.py#L178) in fonttools.

anthrotype:
Thanks! It looks good to me. Maybe we could also mention in the relevant `--help` section that the file must be UTF-8 encoded?

JasonAtGithub:
@anthrotype The `--help` section mention `--text` must be UTF-8 string, and `--text-file is the same as --text, but input from a file`, so we need do nothing.

anthrotype:
Excellent!

PR #659, opened 2016-09-01
varlib/builder.py:

readroberts:
varlib/builder.py:
Fix mis-match in structure names with otData.py: VarRegion
Remove outdated field: self.Reserved
Handle item == None

varlib/__init__.py
Use postscript name entry from designspace file
Add Windows name ID entries as well as Mac name ID entries.
Add logic for making CFFV table rather than gvar when source master font has CFF instead of glyf. Although CFFV table is deprecated as an OpenType top level table, the CFFV table is is still used in the CFF23 table.

ttlib/tables/__init__.py
Add CFF2 and CFFV. Latter is still useful as an intermediate step.

ttlib/tables/_f_v_a_r.py
Support instance postscript name ID. Change name od nameid to subfamilyNameID to distinguish from ps name id.

ttlib/tables/otBase.py
Support long offsets.

ttlib/tables/otConverters.py
Support long computed counts.

ttlib/tables/otData.py
Add structures needed for GSUB FeatureVariations subtable.

cffLib.py
Changes to support CFF2. Note that this is work in progress: the current version does not remove all the CFF data structures that have been removed from the OT CFF2 spec.

misc/psCharStrings.py
Changes to support CFF2.

### DIFF
--- a/Lib/fontTools/ttLib/tables/otData.py
+++ b/Lib/fontTools/ttLib/tables/otData.py
@@ -837,6 +837,56 @@ otData = [
 	]),
 
 	#
+	# variations
+	#
+	('VarAxis', [
+		('uint16', 'AxisIndex', None, None, ''),
+		('F2Dot14', 'StartCoord', None, None, ''),
+		('F2Dot14', 'PeakCoord', None, None, ''),
+		('F2Dot14', 'EndCoord', None, None, ''),
+	]),
+
+	('VarTuple', [
+		('uint16', 'VarAxisCount', None, None, ''),
+		('struct', 'VarAxis', 'VarAxisCount', 0, ''),
+	]),
+
+	('VarTupleList', [
+		('uint16', 'VarTupleCount', None, None, ''),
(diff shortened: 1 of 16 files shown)


Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -443,10 +444,45 @@
 	#
 
 	('GSUB', [
-		('Version', 'Version', None, None, 'Version of the GSUB table-initially set to 0x00010000'),
+		('Version', 'Version', None, None, 'Version of the GSUB table-initially set to 0x00010000, or 0x00010000 if supporting FeatureVariations'),
```
behdad: You meant 0x00010001?

behdad:
I merged some of this already.  Working on the rest.

readroberts:
Have newer updates: will submit new pull request in a bit.

behdad:
Wait a few minutes.  Let me push what I have out.

behdad:
Pushed out what I've got.  I'm merging FeatureVariations now.  Should be another ten minutes or so.

PR #660, opened 2016-09-05
Updated cfflib.py and psCharStrings.py to support current CFF2 format.

readroberts:
Updated OT feels to support GSUB feature Variations table.

### DIFF
--- a/Lib/fontTools/ttLib/tables/otData.py
+++ b/Lib/fontTools/ttLib/tables/otData.py
@@ -6,6 +6,7 @@ otData = [
 	#
 	# common
 	#
+	# type, name, repeat, aux, descr 
 
 	('LookupOrder', []),
 
@@ -134,7 +135,7 @@ otData = [
 		('uint16', 'StartSize', None, None, 'Smallest size to correct-in ppem'),
 		('uint16', 'EndSize', None, None, 'Largest size to correct-in ppem'),
 		('uint16', 'DeltaFormat', None, None, 'Format of DeltaValue array data: 1, 2, or 3'),
-		('DeltaValue', 'DeltaValue', '', 'DeltaFormat in (1,2,3)', 'Array of compressed data'),
+		('DeltaValue', 'DeltaValue', '', 0, 'Array of compressed data'),
 	]),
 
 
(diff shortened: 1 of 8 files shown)


Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -202,17 +202,78 @@ def preWrite(self, font):
...
-		for item in self.Item:
-			for i in range(count - 1, numShorts - 1, -1):
-				if not (-128 <= item[i] <= 127):
-					numShorts = i + 1
-					break
-			if numShorts == count:
-				break
+		if hasattr(self, "Item") and self.Item:
+			for item in self.Item:
+				if item:
```
behdad: This doesn't make sense to me.  If the item is None, we fail later.  No?  Why do you need this?

Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -202,17 +202,78 @@ def preWrite(self, font):
...
+						if not (-128 <= item[i] <= 127):
+							numShorts = i + 1
+							break
+					if numShorts == count:
+						break
 
 		rawTable['NumShorts'] = numShorts
 		return rawTable
 
+class GSUB(BaseTable):
```
behdad: This is also not needed.  There's a cleaner way to support this.  I'm pushing that out in a few minutse.

behdad:
Ok, please pull again.  I've now pushed FeatureVariations out.  It's slightly different from yours.  Please rebase and send the CFF stuff.  Thanks :)

behdad:
Note.  You now have to set the GSUB table version yourself for it to pick up FeatureVariations.  Set to 0x00010001.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -134,7 +135,7 @@
 		('uint16', 'StartSize', None, None, 'Smallest size to correct-in ppem'),
 		('uint16', 'EndSize', None, None, 'Largest size to correct-in ppem'),
 		('uint16', 'DeltaFormat', None, None, 'Format of DeltaValue array data: 1, 2, or 3'),
-		('DeltaValue', 'DeltaValue', '', 'DeltaFormat in (1,2,3)', 'Array of compressed data'),
+		('DeltaValue', 'DeltaValue', '', 0, 'Array of compressed data'),
```
behdad: The 'DeltaFormat in (1,2,3)' is needed for GPOS variations.

PR #662, opened 2016-09-07
Opentype gx

readroberts:
Add entries for CFF2 support: C_F_F__2.py, and reference in ttlib.tables.__init__.py

### DIFF
--- a/Lib/fontTools/cffLib.py
+++ b/Lib/fontTools/cffLib.py
@@ -5,6 +5,12 @@ from fontTools.misc.py23 import *
 from fontTools.misc import sstruct
 from fontTools.misc import psCharStrings
 from fontTools.misc.textTools import safeEval
+from fontTools.ttLib import TTFont
+from fontTools.ttLib.tables.otBase  import BaseTTXConverter
+from fontTools.ttLib.tables.otBase  import BaseTable
+from fontTools.ttLib.tables.otBase  import OTTableWriter
+from fontTools.ttLib.tables.otBase  import OTTableReader
+from fontTools.ttLib.tables import otTables as ot
 import struct
 import logging
 
@@ -27,20 +33,32 @@ class CFFFontSet(object):
 		pass
 
 	def decompile(self, file, otFont):
+		self.otFont = otFont
 		sstruct.unpack(cffHeaderFormat, file.read(4), self)
-		assert self.major == 1 and self.minor == 0, \
+		assert ((self.major == 1 and self.minor == 0) or (self.major == 2 and self.minor == 0)), \
 				"unknown CFF format: %d.%d" % (self.major, self.minor)
 
 		file.seek(self.hdrSize)
-		self.fontNames = list(Index(file))
-		self.topDictIndex = TopDictIndex(file)
-		self.strings = IndexedStrings(file)
-		self.GlobalSubrs = GlobalSubrsIndex(file)
-		self.topDictIndex.strings = self.strings
-		self.topDictIndex.GlobalSubrs = self.GlobalSubrs
+		isCFF2 = self.major == 2
+		if self.major == 1:
+			self.fontNames = list(Index(file))
+			self.topDictIndex = TopDictIndex(file)
+			self.strings = IndexedStrings(file)
+			self.GlobalSubrs = GlobalSubrsIndex(file)
+			self.topDictIndex.strings = self.strings
+			self.topDictIndex.GlobalSubrs = self.GlobalSubrs
+		else: # major is 2.
+			self.fontNames = ["CFF2Font"]
+			cff2GetGlyphOrder = otFont.getGlyphOrder
+			self.topDictIndex = TopDictIndex(file, cff2GetGlyphOrder)
+			self.strings = IndexedStrings(file)
+			self.GlobalSubrs = GlobalSubrsIndex(file)
+			self.topDictIndex.strings = self.strings
+			self.topDictIndex.GlobalSubrs = self.GlobalSubrs
 
 	def __len__(self):
 		return len(self.fontNames)
+		
 
 	def keys(self):
 		return list(self.fontNames)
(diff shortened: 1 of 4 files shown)


behdad:
Thanks Read.

Can you rebase, instead of merge, on top of latest branch?  Ie. we shouldn't see a "Merge pull request" in your pull-request.

My biggest concern is that this is a big change to CFF code, instead of small incremental commits.  If something regresses, it will be a nightmare to debug.

If it perhaps possible that after you have your final code, then you go back and break it down into smaller chunks?  I can help with that as well, like the other changes.  But I would need test fonts.

Thanks.

Lib/fontTools/cffLib.py
```diff
@@ -27,20 +33,32 @@ def __init__(self):
 		pass
 
 	def decompile(self, file, otFont):
+		self.otFont = otFont
 		sstruct.unpack(cffHeaderFormat, file.read(4), self)
-		assert self.major == 1 and self.minor == 0, \
+		assert ((self.major == 1 and self.minor == 0) or (self.major == 2 and self.minor == 0)), \
```
behdad: Use "self.major in (1, 2)"?

Lib/fontTools/cffLib.py
```diff
@@ -27,20 +33,32 @@ def __init__(self):
...
 				"unknown CFF format: %d.%d" % (self.major, self.minor)
 
 		file.seek(self.hdrSize)
-		self.fontNames = list(Index(file))
-		self.topDictIndex = TopDictIndex(file)
-		self.strings = IndexedStrings(file)
-		self.GlobalSubrs = GlobalSubrsIndex(file)
-		self.topDictIndex.strings = self.strings
-		self.topDictIndex.GlobalSubrs = self.GlobalSubrs
+		isCFF2 = self.major == 2
```
behdad: This is unused.

Lib/fontTools/cffLib.py
```diff
@@ -93,21 +127,35 @@ def toXML(self, xmlWriter, progress=None):
...
 		if name == "CFFFont":
-			if not hasattr(self, "fontNames"):
-				self.fontNames = []
-				self.topDictIndex = TopDictIndex()
-			fontName = attrs["name"]
-			topDict = TopDict(GlobalSubrs=self.GlobalSubrs)
-			topDict.charset = None  # gets filled in later
-			self.fontNames.append(fontName)
+			isCFF2 = (self.major == 2)
+			if self.major == 1:
```
behdad: Maybe remove isCFF2 and always just check self.major directly?  It's very inconsistent right now.

behdad: Maybe remove isCFF2 and always just check self.major directly?  It's very inconsistent right now.

Lib/fontTools/cffLib.py
```diff
@@ -186,50 +243,65 @@ def getItems(self, items, strings):
...
 	def getOffsets(self):
-		pos = 1
-		offsets = [pos]
-		for item in self.items:
-			if hasattr(item, "getDataLength"):
-				pos = pos + item.getDataLength()
-			else:
-				pos = pos + len(item)
-			offsets.append(pos)
+		if self.items:
```
behdad: Why do you need these "if self.items" checks?  Why weren't they needed before?

Lib/fontTools/cffLib.py
```diff
@@ -239,10 +311,10 @@ def getItems(self, items, strings):
 			out.append(item.getCompiler(strings, self))
 		return out
 
-	def getChildren(self, strings):
+	def getChildren(self, strings, isCFF2):
```
behdad: Set isCFF2 to default to False whenever you are changing API please.

readroberts:
Hi Behdad;

My apologies: I started to set up the pull request for the cffLib.py and pscharstrings.py , but did not intend to finish it, as I am not quite ready for the cffLib.py and pscharstrings.py to be merged yet with your branch. I found another issue I want to work on.  I see that I must have committed it after all.

It will be hard to divide up the CFF changes too much, many of the  changes make a  different CFF structure There would be no tools other than fontTools which can use any of the intermediate states.   That said, I could reasonably commit in several steps: Support CFF 2 version

-          Omit TopDict fields

-          Omit charset, and write empty string index

-          Support blend and vsindex

-          Support internal VarStore data

I have already promised Sascha test data and unit tests. If you prefer, I can hold off on the check in until I have those done, which will probably be early next week. At the moment, I am the only user of the CFF2-capable the cffLib.py and pscharstrings.py moduels, so I don’t think there is any hurry.

I could use advice on how to commit several steps. I see that if I make a commit on my fork, and then issue a pull request, and then make another commit on my fork and issue another pull request, that the second pull request contains not only the changed files from the second commit, but also the changed files from the first commit. A bit of web searching suggests the only way to avoid this is to commit each of several sets of changes to a different branch on my fork, and then issue a pull request for each branch.  If you know a better way, let me know.


-          Read

behdad:
> My apologies: I started to set up the pull request for the cffLib.py and pscharstrings.py , but did not intend to finish it, as I am not quite ready for the cffLib.py and pscharstrings.py to be merged yet with your branch.

Sure, I can wait.  It's good to take a look at what you are doing from time to time though. :)

> I found another issue I want to work on. I see that I must have committed it after all. It will be hard to divide up the CFF changes too much, many of the changes make a different CFF structure There would be no tools other than fontTools which can use any of the intermediate states.

Right.  I'm mostly interested in intermediate states not breaking CFF.  That helps narrow down changes if a regression gets in.

> That said, I could reasonably commit in several steps:

Sounds about right.

> At the moment, I am the only user of the CFF2-capable the cffLib.py and 
> pscharstrings.py moduels, so I don’t think there is any hurry.

Right.

> I could use advice on how to commit several steps. I see that if I make a commit
> on my fork, and then issue a pull request, and then make another commit on my fork 
> and issue another pull request, that the second pull request contains not only the
> changed files from the second commit, but also the changed files from the first commit.
> A bit of web searching suggests the only way to avoid this is to commit each of several
> sets of changes to a different branch on my fork, and then issue a pull request for each 
> branch. If you know a better way, let me know.

Humm.  You create one branch; commit whatever you want to it, push, and make Pull-Request.  You can change things in your branch and push again (including forced pushes) and the pull-request updates automatically to pick up new changes.  You only need to create one pull request, ever.

PR #665, opened 2016-09-12
[py23] add round2 and round3 functions

anthrotype:
This is to address https://github.com/behdad/fonttools/issues/664.

It adds two new functions to the `py23` module: namely, `round2` and `round3`.
The former behaves like Python 2's built-in `round` function, wheres the latter like Python 3's.

I also added some of the tests from the CPython 2.7 test suite.

Please review, thanks.

/cc @moyogo

### DIFF
--- a/Lib/fontTools/misc/py23.py
+++ b/Lib/fontTools/misc/py23.py
@@ -7,7 +7,7 @@ import sys
 __all__ = ['basestring', 'unicode', 'unichr', 'byteord', 'bytechr', 'BytesIO',
 		'StringIO', 'UnicodeIO', 'strjoin', 'bytesjoin', 'tobytes', 'tostr',
 		'tounicode', 'Tag', 'open', 'range', 'xrange', 'round2', 'round3',
-		'Py23Error']
+		'round', 'Py23Error']
 
 
 class Py23Error(NotImplementedError):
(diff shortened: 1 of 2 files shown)


anthrotype:
Do we also want to shadow the built-in `round` with `round3` for both pythons?

```python
round = round3
```

moyogo:
Great! Thanks!

> Do we also want to shadow the built-in round with round3 for both pythons?

Yes.

anthrotype:
ok, I added `round = round3` but for PY2 only, for PY3 I use the built-in round.

damn, I need to catch a flight!

anthrotype:
Since when landscape.io check has become required for merging pull requests? it takes ages, and I'm not sure it's worth it...

benkiel:
FYI, fontParts has been updated to use/match the fontTools behavior here.

anthrotype:
Ok, I'll merge this. Watch out if you `import *` from py23 module. The `round` function is gonna behave like Python 3 from now.
(That's why I never `import *`)
Also, these are the changes that would deserve a version bump. But that's coming very soon.

benkiel:
@anthrotype Thanks! It's much more explicit than than — it imports round3 and uses it in the normalizer. Also, looking at things, fontMath should likely get updated for this behavior also. There are just a couple of spots where it rounds.

jamesgk:
So, I just discovered this because currently about 25% of fontmake's total runtime is spent in round3 (using python 2). Specifically, when it's used in the t2CharStringPen module. Not sure at the moment how we can get around that. cc @behdad

behdad:
Right.. Since you can reproduce this, can you optimize it?  It occurs to me that if we avoid rounding if number is already int, and avoiding the power if ndigits is zero, plus other straightforward optimizations, might make this go away.

behdad:
I also think we can remove the checks for `__round__` and `__index__`.

behdad:
Maybe file an issue first.

PR #666, opened 2016-09-14
voltLib: parsing VOLT files

moyogo:
This add fontTools.voltLib which much like feaLib parses OTL definitions but from VOLT files.
I haven’t had time to implement building yet, so merging can let others help.

### DIFF
--- a/Lib/fontTools/voltLib/ast.py
+++ b/Lib/fontTools/voltLib/ast.py
@@ -91,3 +91,9 @@ class AnchorDefinition(ast.Statement):
         self.component = component
         self.locked = locked
         self.pos = pos
+
+class SettingDefinition(ast.Statement):
+    def __init__(self, location, name, value):
+        ast.Statement.__init__(self, location)
+        self.name = name
+        self.value = value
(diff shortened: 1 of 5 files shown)


Lib/fontTools/voltLib/ast.py
```diff
@@ -0,0 +1,232 @@
+from __future__ import print_function, division, absolute_import
+from __future__ import unicode_literals
+import fontTools.feaLib.ast as ast
```
brawer: This dependency seems a bit weird; could voltLib define its own syntax trees?

Lib/fontTools/voltLib/parser.py
```diff
@@ -0,0 +1,633 @@
+from __future__ import (
+    print_function, division, absolute_import, unicode_literals)
+from collections import OrderedDict
+import fontTools.voltLib.ast as ast
+import fontTools.feaLib.parser as parser
```
brawer: Why does voltLib use the feaLib parser? The dependency on feaLib.parser.SymbolTable seems strange — doesn’t voltLib needs its own SymbolTable?

anthrotype:
Awesome! Thank you Denis 👍

moyogo:
@brawer oh there’s another commit right after you merged where I also remove the import from feaLib.parser. I cherry-picked it to master.

brawer:
Thank you!

behdad:
Evil!

PR #675, opened 2016-09-27
update Makefile; use pytest-runner; etc.

anthrotype:
Follow up from https://github.com/fonttools/fonttools/commit/064932e2e6557852d311f901e3c798cca8e34174

Tests can now be run from a local checkout with the canonical `python setup.py test`. You can also use the `run-tests.sh` script or `make check`, so everyone is happy. :)

We use the pytest-runner plugin to bootstrap pytest when running `setup.py test`, without having to pip install it beforehand.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,22 @@
 all:
-	./setup.py bdist
+	./setup.py build
 
 dist:
-	./setup.py sdist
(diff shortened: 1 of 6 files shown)


behdad:
Yay, thanks!

Issue #540, opened 2016-03-10
tableVersion in 'vhea' is confusing


mashabow:
I got a dump of 'vhea' table from `SourceHanSans-Regular.otf` like this:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<ttFont sfntVersion="OTTO" ttLibVersion="3.0">

  <vhea>
    <tableVersion value="1.0625"/>
...
```
It says the version is 1.0625, while this font has a 'vhea' table with version 1.1 (`0x00011000`). 

"Version Numbers" section in [the OpenType spec](https://www.microsoft.com/typography/otspec/otff.htm "The OpenType Font File") states as below.

> When a Fixed number is used as a version, the upper 16 bits comprise a major version number, and the lower 16 bits a minor. Tables with non-zero minor version numbers always specify the literal value of the version number since the normal representation of Fixed numbers is not necessarily followed. For example, the version number of 'maxp' table version 0.5 is 0x00005000, and that of 'vhea' table version 1.1 is 0x00011000.

behdad:
Ouch!

Note that OpenType is not consistent about this. For example, GDEF table version 1.2 uses 0x00010002:

https://www.microsoft.com/typography/otspec/gdef.htm

Maybe we should remove the float conversion and go back to pure hex.

behdad:
'head' table is also affected; though that one has only one version.

behdad:
In the opentype-gx branch I changed Version of GSUB/GPOS/... to always be hex.  The ones reported in this issue still remain broken.
https://github.com/behdad/fonttools/commit/56c0a784d2da52c2ec125c0e1392e51d24b38f5c

anthrotype:
Fixed in #711

PR #677, opened 2016-09-27
Merge Opentype Variation Fonts (previously known as OpenType GX)

behdad:


Lib/fontTools/ttLib/tables/C_F_F_test.py
```diff
@@ -0,0 +1,384 @@
...
+
+        cffData = deHexStr(cffHexData)
+        CFF2Table = font['CFF '] = newTable('CFF ')
+        CFF2Table.decompile(cffData, font)
+        writer = XMLWriter(BytesIO())
+        font['CFF '].toXML(writer, font)
+        xml = writer.file.getvalue().decode("utf-8")
+        # normalize spacing and new-lines, so we can edit the XML without being very careful.
+        xml = re.sub(r"  +", "    ", xml)
+        xml = re.sub(r"[\r\n][\r\n]+", "\r", xml)
```
anthrotype: why substituting it with "\r" here, instead of "\n"?

Lib/fontTools/ttLib/tables/C_F_F_test.py
```diff
@@ -0,0 +1,384 @@
...
+from fontTools.misc.py23 import *
+from fontTools.misc.testTools import parseXML
+from fontTools.misc.textTools import deHexStr, hexStr
+from fontTools.misc.xmlWriter import XMLWriter
+from fontTools.ttLib import TTFont, newTable
+import re
+import unittest
+
+
+cffXML = """
```
anthrotype: In all the other test modules, we usually keep the TTX files in a `testdata` folder rather than embedded as strings in the test module itself, especially when they are long.

Lib/fontTools/ttLib/tables/C_F_F_test.py
```diff
@@ -0,0 +1,384 @@
...
+class CFFTableTest(unittest.TestCase):
+    def test_toXML(self):
+        global cffXML
+        # make font with post table.
+        font = MakeFont()
+
+        cffData = deHexStr(cffHexData)
+        CFF2Table = font['CFF '] = newTable('CFF ')
+        CFF2Table.decompile(cffData, font)
+        writer = XMLWriter(BytesIO())
```
anthrotype: XMLWriter can also write directly on a `io.StringIO()` -- in `py23` we call that `UnicodeIO`. If you use that, you don't need to decode the result.

Lib/fontTools/ttLib/tables/C_F_F_test.py
```diff
@@ -0,0 +1,384 @@
...
+
+def parseTableXML(otTable, xmlText, font):
+    xmlList = parseXML(xmlText)
+    for entry in xmlList:
+        if len(entry) != 3: # skip comments and newlines.
+            continue
+        name, attrs, content = entry
+        otTable.fromXML(name, attrs, content, font)
+
+def MakeFont():
```
anthrotype: what is the purpose of this? `font = MakeFont()` here would be the same as `font = TTFont()`.

behdad:
Thanks Cosimo.  I'll try to fix these tomorrow.  Or feel free to fix,
please!

Lib/fontTools/ttLib/tables/C_F_F_test.py
anthrotype: Alternatively here you could use `getXML()` function from `fontTools.misc.testTools`, which returns a string stripped of os-specific newlines.
It is used in `feaLib/builder_test.py`, for example.

anthrotype:
Wow, this PR is huge! Thanks Behdad and Read!
We can fix the small things tomorrow.

anthrotype:
I guess this PR deserves a minor version bump to 3.2.0.
Anything else that needs to be done before that?
Also, could anybody sum up the list of changes in a few sentences that we can add to the release note and `NEWS` file?

anthrotype:
I see you changed the format of the GSUB/GPOS/GDEF/MATH/BASE/JSTF/... version with https://github.com/fonttools/fonttools/commit/402d72669281b2ad3bf4fe18e82a0e02336ff5cd

If this is an API break, then I think we need to bump the MAJOR version, to 4.0.0.
And since we changed these, we might also want to fix the rest of the tables' versions like vhea, head, and such https://github.com/fonttools/fonttools/issues/540

WDYT?

brawer:
How about “Support OpenType 1.8” as one-line summary?

2016-09-28 20:02 GMT+02:00 Cosimo Lupo <notifications@github.com>:

Issue #686, opened 2016-10-02
Cannot change woff compression level with sfnt.ZLIB_COMPRESSION_LEVEL


severinmeyer:
Currently, setting `sfnt.ZLIB_COMPRESSION_LEVEL` to a custom value does not have any effect.

The culprit: [Lib/fontTools/ttLib/sfnt.py#L496](https://github.com/fonttools/fonttools/blob/1eba7b383c785924a33c30f6e38cebd571819a5f/Lib/fontTools/ttLib/sfnt.py#L496)

The problem is that `zlibCompressionLevel` is accessed as an instance variable, but it is actually declared as a class variable at [Lib/fontTools/ttLib/sfnt.py#L481](https://github.com/fonttools/fonttools/blob/1eba7b383c785924a33c30f6e38cebd571819a5f/Lib/fontTools/ttLib/sfnt.py#L481). As a result, it is instantiated once when the module is loaded, and any later change of `sfnt.ZLIB_COMPRESSION_LEVEL` is not propagated.

Possible solution: Ditch the variable and use `ZLIB_COMPRESSION_LEVEL` directly:
```python
if not self.uncompressed:
    compressedData = compress(data, ZLIB_COMPRESSION_LEVEL)
```

anthrotype:
Right. Or you could also set the `WOFFDirectoryEntry.zlibCompressionLevel` class attribute to the value of you choice?

brawer:
@severinmeyer, do you want to send a pull request?

anthrotype:
The thing is, that `zlibCompressionLevel` class attribute was already there when I added Zoplfi support with https://github.com/fonttools/fonttools/commit/7fafe1f9a71f78b371e9b8670c157810c2e60211.
It was set to a constant value, 6, the same as the global variable ZLIB_COMPRESSION_LEVEL that I added.
So the previous as well as current way to set a different zlib compression level is to set that attribute on the WOFF directory entry class. If we remove that, and use the global module variable, then existing code that uses the other approach will break.
I suggest we keep it like that and the user adjust the compression level via the class attribute.

severinmeyer:
That makes sense.

However, it still seems problematic to use **self.**zlibCompressionLevel, because it doesn't exist and falls back to the class variable. Which is why I assumed this is a mistake in the first place.

[...]

Issue #495, opened 2016-02-01
define an API for options


anthrotype:
we need to find a better way to define optional parameters (e.g. "recalculate bounding boxes", "compress this font to WOFF 1.0 using zopfli",  etc.) without needing to change the TTFont constructor's signature every time we define new ones.

Suggestions?

anthrotype:
I was thinking to define a bare `class Options(object): pass` in the `ttLib` module, and initialise that as an `options` attribute when constructing a `TTFont` instance.

I would also add a `**kwargs` argument to the `TTFont.__init__` and use its values to populate `self.options`.

I would also add to this `self.options` all the values of the existing arguments of `TTFont.__init__`, which we currently set as attributes of the TTFont instance itself. For example, I would make, e.g., `self.lazy` into `self.options.lazy` (and use a property decorator to make sure that scripts getting `self.lazy` will get the value of `self.options.lazy`).

I would also add a `**kwargs` argument to `TTFont.save`, specifying the export-time options (e.g. use zopfli), which I would pass on to the `SFNTWriter` instance.  Unlike the load-time options defined in the TTFont constructor, these export-time options would not be remembered by the TTFont instance (e.g. like the current `reorderTables` option).

Comments?

behdad:
Done now I believe?

PR #687, opened 2016-10-02
sfnt.py: drop `WOFFDirectoryEntry.zlibCompressionLevel`

anthrotype:
... use global `ZLIB_COMPRESSION_LEVEL` instead

Fixes https://github.com/fonttools/fonttools/issues/686

cc @severinmeyer

### DIFF
--- a/Lib/fontTools/ttLib/sfnt.py
+++ b/Lib/fontTools/ttLib/sfnt.py
@@ -478,7 +478,17 @@ class WOFFDirectoryEntry(DirectoryEntry):
 
 	format = woffDirectoryEntryFormat
 	formatSize = woffDirectoryEntrySize
-	zlibCompressionLevel = ZLIB_COMPRESSION_LEVEL
+
+	def __init__(self):
+		super(WOFFDirectoryEntry, self).__init__()
+		# With fonttools<=3.1.2, the only way to set a different zlib
+		# compression level for WOFF directory entries was to set the class
+		# attribute 'zlibCompressionLevel'. This is now replaced by a globally
+		# defined `ZLIB_COMPRESSION_LEVEL`, which is also applied when
+		# compressing the metadata. For backward compatibility, we still
+		# use the class attribute if it was already set.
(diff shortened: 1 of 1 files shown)


Lib/fontTools/ttLib/sfnt.py
```diff
@@ -478,7 +478,17 @@ class WOFFDirectoryEntry(DirectoryEntry):
...
+	def __init__(self):
+		super(WOFFDirectoryEntry, self).__init__()
+		# With fonttools<=3.1.2, the only way to set a different zlib
+		# compression level for WOFF directory entries was to set the class
+		# attribute 'zlibCompressionLevel'. This is now replaced by a globally
+		# defined `ZLIB_COMPRESSION_LEVEL`, which is also applied when
+		# compressing the metadata. For backward compatibility, we still
+		# use the class attribute if it was already set.
+		if not hasattr(WOFFDirectoryEntry, 'zlibCompressionLevel'):
+			self.zlibCompressionLevel = ZLIB_COMPRESSION_LEVEL
```
adrientetar: Should this be WOFFDirectoryEntry.zlibCompressionLevel, right now the hasattr will always return False.

anthrotype: indeed, that's what I wanted to achieve. I wanted to set the instance's `zlibCompressionLevel` to the global `ZLIB_COMPRESSION_LEVEL`; and only use the class attribute `zlibCompressionLevel` for existing code that might still use the previous API, whereby changing the compression level was defined at the class level.
Now I defined it at module scope, so it affects both the directory entries and the metadata.

anthrotype: the plan is to actually get rid of these globals, and use a better API for options that doesn't require adding a new keyword argument to TTFont constructor all the time: 
https://github.com/fonttools/fonttools/issues/495

severinmeyer: Based on some test runs it works as intended in both cases, i.e. setting the compression level with the class variable or the global variable.

Issue #683, opened 2016-09-30
[varLib] Need to support Platform 1 and 3 in name table


robmck-ms:
varLib should generate a variable font with both platform 1 and 3 names. Previously, it did only platform 1. Recently, it switched to only platform 3. These fonts wouldn't work on OSX (though the older ones did): FontBook said that the instances were all duplicates of each other, and all instances other than Regular looked like the light. If I simply add platform 1 entries to the name table for the nameids referenced from var, the font works fantastically.

(already emailed Behdad on this).

anthrotype:
Thanks Rob. For reference, this is the commit in question:
https://github.com/fonttools/fonttools/commit/24346aaa88fbd78299b86caa3411e86ecc3ec001

anthrotype:
@robmck-ms I wonder if we can use platformID=0 (Unicode) instead of platformID=1 (Macintosh). Would that work on OSX?

robmck-ms:
@anthrotype Having platformID=0 makes more sense to me as you wouldn't need both 1 & 3. I tried it, though, with the same symptoms as with a font with platformID=3 only.

anthrotype:
yeah, I also tried Unicode platformID but didn't work. So I guess we are stuck with both MacRoman and Windows. :-/

robmck-ms:
I *think* it was El Capitan (10.11), but I'm not 100% certain. That sounds about right for what I was running at the time, as I didn't switch to Sierra right away. Unfortunately, none of the email I can find on the subject mentioned the OS version.

anthrotype:
Thanks. IIRC, El Capitan was released before OpenType variations became officially supported (Sierra 10.12 I think was the first), so maybe that limitation of platformID=1 was the legacy of the old TrueType GX code.
In whic case, perhaps we could stop adding Mac names in fvar by default, and only use Windows (unicode) names.

/cc @justvanrossum

anthrotype:
after  #3688, varLib no longer adds MacRoman names for variations-related tables (fvar, STAT) unless the name table of the default master ttf already contained some. fontmake/ufo2ft don't normally create those (unless sources have low-level name record definitions with explicit platformID).

PR #688, opened 2016-10-04
Fix varlib names

anthrotype:
This should fix https://github.com/fonttools/fonttools/issues/683

varLib now adds both Macintosh (platformID=1) and Windows (platformID=3) name records for axes names and instance styles, to work around an issue when loading OpenType Variable fonts on macOS.

I also fixed another str/bytes issue related to the `NameRecord.setName` method. This now accepts as input string both unicode or bytes, with the latter assumed to be already encoded with the specified (platformID, platEncID, langID) triplet.

Note that this is how the `string` attribute of `NameRecord` worked already: it takes either a unicode string (in which case, it attempts to encode it with the correct platform encoding upon compiling), or it takes a pre-encoded bytes string which is treated literally. It's the client's responsibility to either pass unicodes (and let the library encode them), or to take care of passing in the correct bytes.

### DIFF
--- a/Lib/fontTools/misc/loggingTools.py
+++ b/Lib/fontTools/misc/loggingTools.py
@@ -429,6 +429,41 @@ class ChannelsFilter(logging.Filter):
 		return False
 
 
+class CapturingLogHandler(logging.Handler):
+	def __init__(self, logger, level):
+		self.records = []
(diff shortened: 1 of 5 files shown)


anthrotype:
I'm gonna go and merge this. Feel free to open a new issue.

PR #690, opened 2016-10-04
Allow single-point open contours in area pen

jamesgk:
Sometimes an anchor is drawn as a moveTo followed by an endPath, for
that case it makes sense just to ignore and count the area as zero.

r? @anthrotype


behdad:
Care adding test?  lgtm otherwise.

PR #691, opened 2016-10-07
varLib usability stuff

jamesgk:
I know this is all unstable, but in the interest of getting something working earlier I wanted to make the interpolate_layout and designspace modules more usable.

r? @behdad @anthrotype


Lib/fontTools/varLib/designspace.py
```diff
@@ -43,6 +43,17 @@ def _load(et):
 
 	return masters, instances
 
+def _loadMaps(location_lists, key):
+	maps = []
+	for location_list in location_lists:
+		location_map = {}
+		for loc in location_list:
+			abspath = os.path.normpath(os.path.join(
```
anthrotype: `NameError: global name 'os' is not defined`

anthrotype: `NameError: global name 'designspace_path' is not defined`

Lib/fontTools/varLib/designspace.py
```diff
@@ -51,3 +62,15 @@ def load(filename):
...
+def loadMaps(filename, key):
+	"""Load designspace from a file name or object into dictionaries mapping
+	some font attribute (given as key, e.g. 'filename') to location data.
+	"""
+	data = load(filename)
+	return _loadMaps(data, key)
+
+def loadsMaps(string, key):
+	"""Load designspace mappings from a string."""
+	data = loads(filename)
```
anthrotype: `NameError: global name 'filename' is not defined`

jamesgk:
Regarding the loadMaps changes, I felt it was more convenient to pull data from a designspace as a map like `{'MyFont-Regular': {'weight': 100, 'width': 100}, ...}` instead of the list that's returned from the load/s functions. But we can definitely drop that stuff here and just put it in fontmake instead (I already have code doing so which is why this is utterly untested, sorry about that).

PR #695, opened 2016-10-08
Remove extra_path distribution kwarg

htgoebel:
This is of no use, it is undocumented and bound to be removed.

See https://bugs.python.org/issue901727 and https://bugs.python.org/issue27919.


anthrotype:
Thanks for the PR.
I think that was a relic of an old setup.py. We don't need to use that any more.

Issue #697, opened 2016-10-09
Version is out of sync with git tags


miguelsousa:
```python
from fontTools.version import __version__ as ftversion
print(ftversion)
```
I'm sync'ed up to the latest commit and the code above will print `3.0` instead of `3.1.2`. Is this expected?

Prior to a502694d1421b09d78654657089633f0e9253e1e I was getting the version this way,
```python
from fontTools import version as ftversion
print(ftversion)
```
which no longer works as before.
What's the best way for a script to know what version of FontTools is installed?

anthrotype:
Duplicate of https://github.com/fonttools/fonttools/issues/680

If you installed FontTools from a git checkout in editable mode, after you do `git pull`, you have to re-run `pip install -e .` or do `python setup.py egg_info` in order to get the updated version string.

The latter is generated upon creating the source distribution from the git metadata.

anthrotype:
the reason I made a separate `version.py` module is to avoid the over-head of importing `pkg_resources` every time one imports `fontTools`. 
Only when one really wants to know the fontTools version, one does

```python
from fontTools.version import __version__
print(__version__)
```

Regular users who install from the source/wheel distributions (which are created at tagged releases) just have the plain version string, such as "3.1.2".

Developers who install --editable from git are expected to refresh the package metadata from time to time, using `python setup.py egg_info`, or re-running `pip install -e .` (which in turn runs the `egg_info` command).

I hope this is clear now.

justvanrossum:
Ok, I just checked, but this is an issue when fontTools gets embedded into another app. Example at hand: RoboFont. It includes fontTools via py2app. Now, `fontTools.version.__version__` is unknown.

I somehow really prefer to have an old fashioned hard coded (possibly generated) version number in the source code.

Cc @typemytype

anthrotype:
I see. There's an alternative way, which is to have setuptools_scm automatically replace `version.py` containing the current version as generated from the git metadata.
Then the module would contain a static string literal, instead of calling pkg_resources to get the string from the (possibly missing) package metadata.
I'll do that then.

justvanrossum:
Btw. also not happy about how https://github.com/fonttools/fonttools/commit/a502694d1421b09d78654657089633f0e9253e1e breaks backwards compatibility.

anthrotype:
yes, the above mentioned approach would also fix that issue.
I can have 
```python
from fontTools.version import __version__ as version
```
[...]

PR #698, opened 2016-10-10
Write version.py file with static string

anthrotype:
The file `Lib/fontTools/version.py` is no longer checked in the repository (it's .gitignored); instead, it's generated by setuptools_scm upon installation, or by directly running the `egg_info` command. It will just contain a static string, instead of using using `pkg_resources` to dynamically load the version from the package metadata -- since the package metadata may not be present, for example, when embedding fonttools (eg. py2app and Drawbot).

Fixes https://github.com/fonttools/fonttools/issues/697

/cc @justvanrossum @miguelsousa


justvanrossum:
What is so bad about having version.py in the repo? This still relies on doing some setup command (that I've already forgotten) to get the correct value there. I just want to do git pull and be all set. That is what embedding app developers also are likely doing. There's a good chance now that a wrong version number will show in those cases.

anthrotype:
And I put back `from fontTools import version` so we restore backward compatibility.

anthrotype:
> I just want to do git pull and be all set. 

I'm afraid software deployment is not as easy as git pull.

anthrotype:
Those who wants to embed fonttools source code just have to unzip the sdist tarball, and they will have a nice and clean version.py.

anthrotype:
I shall also update the README with info for other developers as to how to generate the `version.py` from a fresh git clone.

justvanrossum:
@typemytype also "deploys" straight from git (and I'm sure many others). I don't understand what's so wrong about that, and why we can't have a version number in the code.

anthrotype:
If we had version number in the code, then we would have to change that number every time we change the code. Because if we don't change it, then that version number quickly becomes meaningless.

The current setup allows you to do whatever you prefer, that is:

- install from a released (source or binary) distribution, as available on PyPI or Github Releases (this should be the "supported" way for end users);
- install straight from the remote git repository's HEAD with `pip install git+https://github.com/fonttools/fonttools`,  with/without `--editable`.
- install from a local git checkout with `pip install .` or `pip install --editable .`
- install from anywhere (dist, remote/local repo) to a different target directory, with `pip install --target <some_dir> .` (this could be the suggested method for those who wish to vendor fonttools?)
- import from local git checkout without actually installing, by doing `PYTHONPATH=Lib python -c "import fontTools"` or using a `.pth` file -- provided you first generate the package metadata with `python setup.py egg_info`, at least once after cloning (but ideally every time after pulling to keep those in sync with the code).

Sure, we shall make all this clear in the documentation, but overall it's not that hard, and it comes with a lot of benefits.
That's just my opinion, of course, and I'm open to changes.

justvanrossum:
That number should only change when an actual release is made. I've seen (and used) the convention that a "+" was added to the version string to indicate that it was arbitrarily "newer" than the release. That's simple and useful.

The current setup allows many things, but not a guaranteed correct version number after doing a git pull.

anthrotype:
Re-installing, or regenerating the `egg_info`, will guarantee a correct version number after doing git pull.

Git-pulling new code bearing the same version as the old one is a recipe for confusion.

One could always set up a git alias such as `git pull-egg-info` that pulls and updates the package metadata in one go.

justvanrossum:
The kind of confusion I'm worried about it this: people will forget to update the egg info upon git pull, and then there is a *wrong* version number. That *will* happen and scares the shit out of me.

anthrotype:
The other kind of confusion I'm worried about is people conflating their development and production environments. In the former packages are editable, versions are loose and subject to change; whereas in the latter packages are installed somewhere (system, user, virtualenv, etc.), the code is frozen, and versions should be pinned down precisely to being able to reproduce exactly the same output, and debug any issues that may arise.

justvanrossum:
I give up. I will reluctantly try to make something so I will always run `python setup.py egg_info` after I do `git pull`. Good luck trying to tell everyone else to do the same :)

moyogo:
@justvanrossum I think if you had waited a couple of minutes @anthrotype might have given up first ;-)

justvanrossum:
Well if you want that to happen, perhaps contribute to the argument!

moyogo:
Hehe. I see both points in the argument. I guess we’ll see if people always forget to install after pulling.

miguelsousa:
@anthrotype is it possible to have a hardcoded version number that gets updated with every git tag deploy and serves as a fallback, alongside a more precise versioning mechanism (for cutting edge developers) that must be ran explicitly?

I share the same concerns as @justvanrossum, and as you can see from the issues we've opened we've both been tricked by the current solution. I think you've clearly explained its benefits, but I feel that it lacks a good fallback, for the case when someone does not take any of the routes you've outlined.

For me, having a (fallback) version number that's in sync with the git tags is all I need in 99% of the times. The remaining 1% is when I'm trying to develop or debug something, and at that time I'll sure make myself remember to update the egg.

anthrotype:
> is it possible to have a hardcoded version number that gets updated with every git tag deploy and serves as a fallback

I'm afraid it's not possible, because to change the code you have to make a commit, and if you make a commit, well.. you technically end up changing the version number as well.
The whole point of setuptools_scm is to not having to manually hardcode the version string, not having to a do "bump version to 1.1.1"  commits, etc. besides the tagging that one already does though git.

justvanrossum:
Well, then that "whole point" is severely flawed. I don't install this way on a day-to-day basis. This may be easy or "clean" from a release managers point of view, for me as a developer/user it does not work. And with me many others.

justvanrossum:
The exact git state is nice to have in a version string, but really, that "+" system I described earlier is plenty good. You need exact release info at _actual releases_, you need nothing more than "this is close to release X.Y.X" in between, as at that stage you're dealing with git anyway.

So a release could be done this way:
- bump version string in version.py
- commit version.py
- tag release
- add "+" to version string in version.py
- commit version.py

That set of actions can be automated at the release end and no confusion will ever happen for end users, no matter how the install or don't install.

benkiel:
I also try just to use .pth files in my installs, as I find it's easier for me to keep track of. @anthrotype I understand that you can't do a post commit hook, as then you're committing again and then will never not be committing a file, but is there no way to do a pre-commit hook to bump a version.py file? Didn't SVN allow for such things? I'm not 100% up on Git, so my question may be dumb, but worth exploring if so.

anthrotype:
I don't want to argue any further. I understand this is problem I'm looking for a solution that makes everyone happy. I think I've found one, but now it's late so I'll work on it tomorrow.
Thank you for all your feedback and suggestions, and sorry for the inconvenience this might have caused you! :)

benkiel:
Thank you @anthrotype; I know that you've been trying very hard to get proper version numbers in font tools to make debugging/comparisons/etc possible, and this must feel like "Oh man, more about version numbers", but I really appreciate you looking into a solution that can work for other use cases. Thank you.

typemytype:
what @benkiel said! Thanks @anthrotype!

miguelsousa:
@anthrotype the solution may be as "simple" as updating a file with every release to use the git tag instead of [`unknown`](https://github.com/fonttools/fonttools/commit/a502694d1421b09d78654657089633f0e9253e1e#diff-ebf1ccd4de641f298f037f932720260aR4) or [`0.0.0`](https://github.com/fonttools/fonttools/pull/698/commits/7223536d7d6236a3ce72aca460a77e19b52f6f21#diff-ec456c0261da7510fd8490f0d1f0d092R13), and leaving everything else the same.

I really appreciate all the work you put into fontTools! Sorry we're so picky 🙏

PR #700, opened 2016-10-12
version

anthrotype:
Following the discussion in https://github.com/fonttools/fonttools/pull/698, I changed the way the version string is computed to accommodate all the possible use cases.

In this PR, I drop `setuptools_scm` and replace it with a similar tool called [`versioneer`](https://github.com/warner/python-versioneer), which like the former asks the version-control system for version information, but it does that dynamically *at import time*, rather than only at setup/installation time.

This means that, if you are a developer and installed the package `--editable`, or use `*.pth` files or `PYTHONPATH` variable, any change in the code (local edits, git pull, etc.) will be reflected in the `fontTools.version` string. Thus, no more need to recreate the package metadata. Pretty cool, uh? :) Of course, this presumes your are working from a git checkout.

The tool also supports storing the calculated version string in all distribution packages (sdist, wheel, the `build/` folder when doing `setup.py build`), and even so-called git archives (e.g. those produced by Github when clicking 'download'), so that the information can be retrieved also when the full repository or git command are not available.

If for any reason all of these methods fail, versioneer will signal an error, and by default it would report a version of `0+unknown`.

However -- we do not want this to happen so, as @miguelsousa among others suggested, we will have a hard-coded fall-back version string which refers to the latest known tag.

And to make it easier updating this static version string as well as ensure it's in sync with the tag, we shall use another nice tool called [`bumpversion`](https://github.com/peritus/bumpversion).

This is a command-line tool that updates all matching version strings in the source code by the correct increment following the semver method (ie. major, minor, patch).

Say we want to increment the third and last digit, we can simply do:

```shell
$ bumpversion patch
```

And `bumpversion` will replace the `__version__` value in `fontTools/__init__.py` with the new version, commit the change with a "Bump version: 3.1.2  → 3.1.3" message, and _also_ create an annotated git tag with the same number and message, all in a single command.
Same thing with `bumpversion minor` or `bumpversion major`.

I believe this should cover any possible scenario and make everyone happy.

Let me just point out I am also among those "developers/users" (to [quote Just](https://github.com/fonttools/fonttools/pull/698#issuecomment-252704772)), and don't like playing the "release manager". I'm doing this exactly because I wish `non-developers/users` could also enjoy the product of our work, without having to know about things like "git", "site-packages", "egg_info", and what not! ;)

Thanks


anthrotype:
The `Lib/fontTools/_version.py` and `versioneer.py` files are autogenerated by the `versioneer` tool and need be checked in the repository for the tool to work (they are in the public domain, CC license).
In built/released packages there is not `versioneer.py` and the `_version.py` just contains the computed version strings, no extra code.

miguelsousa:
Brilliant! 👏
I think you addressed all the issues. If I find any problems with this setup I'll sure let you know. Well done Cosimo!

anthrotype:
I'm now working on updating the changelog for version 3.2.0. Stay tuned.

justvanrossum:
Awesome, Cosimo, that was a lot of work, and overall sounds like the "right thing". Thanks!

benkiel:
This is great, thank you so much @anthrotype!

PR #706, opened 2016-10-12
[_h_m_t_x] Read advanceWidth as unsigned short (uint16)

### LINKED ISSUE #645: [hmtx] advanceWidth should be USHORT, why does fonttools reads it as SHORT?
It seems intentional, but why?
https://github.com/behdad/fonttools/commit/035bc7da2318f0dcacecba2e1e24c0a4bf3cf80c#diff-f1a37102608ecea22d52a8cb91aba526R21

Also see: https://github.com/googlei18n/ufo2ft/pull/60

anthrotype:
Print a warning on decompile if advance width (or height for vmtx) exceeds 0x7FFF (32767).

Upon compile, `struct.pack` will raise if the value is negative as it is out of range for the unsigned "H" format.

Fixes #645 

---

Apparently both FontLab Studio (both 5 and VI) and Robofont allow these negative advances to occur in a font. UFO spec either does not enforce the value to be positive. I wonder why this is the case. Glyphs like that are definitely broken, as you can see from the mega-huge width they take when loaded in FreeType, MS and Apple rasterizers.

/cc @twardoch @typemytype @typesupply

### DIFF
--- a/Lib/fontTools/ttLib/tables/_h_m_t_x.py
+++ b/Lib/fontTools/ttLib/tables/_h_m_t_x.py
@@ -53,11 +53,11 @@ class table__h_m_t_x(DefaultTable.DefaultTable):
 					"Glyph %r has a huge advance %s (%d); is it intentional or "
 					"an (invalid) negative value?", glyphName, self.advanceName,
 					advanceWidth)
-			self.metrics[glyphName] = [advanceWidth, lsb]
+			self.metrics[glyphName] = (advanceWidth, lsb)
 		lastAdvance = metrics[-2]
 		for i in range(numberOfSideBearings):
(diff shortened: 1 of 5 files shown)


anthrotype:
OK, I don't mind changing both to `tuple()`.

anthrotype:
The only thing I'm worried is if anybody was relying on it being a list and was updating either the width or the sb in place, as in `hmtx['a'][0] = 500`...

justvanrossum:
True, although we could say that is poor style. I just checked my own code and `ufo2fdk` and `ufo2ft` and none of these do that. I'd be happy to take the risk.

anthrotype:
@justvanrossum ok, I changed hmtx/vmtx metrics values to be tuples instead of lists.

PR #707, opened 2016-10-14
[Travis] try to speed up osx builds a bit

anthrotype:
Travis CI has become too slow, especially for the osx environment.

It's not just for the recent outage -- https://www.traviscistatus.com/ --, but I think it's also because of its increasing popularity, and the fact that it's the only one free providing OS X environments.
Even Appveyor, which does one build at the time, is now faster than Travis!

So I think we need to clean up our Travis configuration and only run the most essential ones.

We don't need to test two Python 2.7 bugfix releases on OSX; one (whatever is the current Travis default) is enough.

Also, PyPy is quite niche and we can get away with only testing it on the Linux platform.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,12 +18,6 @@ matrix:
       env: TOXENV=pypy
     - language: generic
       os: osx
-      # the default 'osx' on Travis is 10.9 Mavericks, with Python 2.7.5
-      env: TOXENV=py27
-    - language: generic
-      os: osx
-      # this is to test on 10.11 El Capitan which comes with Python 2.7.10
-      osx_image: xcode7.3
       env: TOXENV=py27
     - language: generic
       os: osx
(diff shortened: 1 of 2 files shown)


khaledhosny:
I’d even go far and only test just one Python 2.x and 3.x on macOS for the same reasons.

anthrotype:
@khaledhosny so you mean getting rid of 3.4 on OS X? Maybe..

To fonttools core devs: one trick that could speed up the Travis builds is pushing to a branch on one's fork instead of on the upstream fonttoos/fonttools repo. So that we we open a PR, only the PR branch builds will run (one for Appveyor and one for Travis).

When we push directly to the upstream, the builds are duplicated: two for each CI, one on the branch alone, the other on the result of merging the PR branch into master (so a total of 4 builds).

We skip the queue basically, and build the PR directly instead of the branch and then the PR. I hope I didn't confuse you even more..

PR #708, opened 2016-10-14
Add otConverters.NameID

brawer:
When writing XML, this produces a comment with the English name.
If the name ID is missing from the name table, it logs a warning.

### DIFF
--- a/Lib/fontTools/ttLib/tables/otConverters_test.py
+++ b/Lib/fontTools/ttLib/tables/otConverters_test.py
@@ -1,8 +1,10 @@
 from __future__ import print_function, division, absolute_import
 from fontTools.misc.py23 import *
-from fontTools.misc.testTools import FakeFont
+from fontTools.misc.loggingTools import CapturingLogHandler
+from fontTools.misc.testTools import FakeFont, makeXMLWriter
 from fontTools.misc.textTools import deHexStr
 import fontTools.ttLib.tables.otConverters as otConverters
+from fontTools.ttLib import newTable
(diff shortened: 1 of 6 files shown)


Lib/fontTools/ttLib/tables/otConverters_test.py
```diff
@@ -28,5 +30,49 @@ def test_write(self):
...
+        writer = makeXMLWriter()
+        self.converter.xmlWrite(writer, self.makeFont(), 666,
+                                "Entity", [("attrib", "val")])
+        xml = writer.file.getvalue().decode("utf-8")
+        self.assertEqual(
+            xml,
+            '<Entity attrib="val"'
+            ' value="666"/>  <!-- missing from name table -->')
+        otConverters.log.removeHandler(handler)
+        self.assertEqual(logStream.getvalue().strip(),
```
anthrotype: hey Sascha, your way of intercepting log messages works perfectly fine. But just to let you know, in the subset and hmtx test modules we use a `CapturingLogHandler` context manager for a similar purpose (I think it was added by @rsheeter). It makes these sorts of tests more compact and expressive.

E.g.:
https://github.com/fonttools/fonttools/blob/2300b9c16846ec588f191b9cad89e2e403ef2c13/Lib/fontTools/ttLib/tables/_h_m_t_x_test.py#L78
https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/subset/subset_test.py#L161

brawer: Thank you! Done.

Issue #144, opened 2014-08-13
[subset] Removing CFF hints can make font illegal


behdad:
There's a bug in the hintmask stripping that leaves residual bytes, which makes OTS reject the font.  Ouch.

behdad:
The new --desubroutinize option works around that bug.

behdad:
Closing as duplicate of https://github.com/behdad/fonttools/issues/343

miguelsousa:
@anthrotype can this note be removed?
```
      XXX Note: Currently there is a known bug in 'CFF ' hint stripping that
      might make the font unusable as a webfont as they will be rejected by
      OpenType Sanitizer used in common browsers. For more information see:
      https://github.com/behdad/fonttools/issues/144
      The --desubroutinize options works around that bug.
```
https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/subset/__init__.py#L179-L183

anthrotype:
I believe so. @behdad?

behdad:
Yes!

miguelsousa:
Done https://github.com/fonttools/fonttools/commit/e69960cce06aa4e23bb5c01bfebbd270928b8421

PR #709, opened 2016-10-15
[subset] [WIP] second attempt at fixing --no-hinting with CFF

anthrotype:
After merging PR #606, we spotted an issue so we decided to revert while investigating.

PR #606 was meant to fix an issue with the subsetter's `--no-hinting` option, which makes it strip all the advance widths from CFF charstrings.
I tried to fix this by 'drawing' the charstring with a NullPen, which would force setting its `width` attribute.

However I discovered that, with the above mentioned patch, the `--no-hinting` option would now throw an error with some fonts -- but not with the (admittedly too simple) 'TestOTF-Regular.otf' which is why I didn't catch this while running the automated tests.

Basically some charstrings have their `private` attribute set to None, hence have no `nominalWidthX` or `defaultWidthX` and the `draw()` method can't compute their width.

Now, if _both_ `--no-hinting` and `--desubroutinize` options are passed, the subsetter completes without error.
I suspect it might be related to https://github.com/fonttools/fonttools/issues/144
But I am not sure how to solve this, so I defer to someone more experienced wih CFF and T2CharStrings.

Here's the traceback

```
$ python -m fontTools.subset "/Users/cosimolupo/Library/Fonts/AGaramondPro-Regular.otf" --no-hinting --notdef-outline '*'
  [...]
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2967, in main
    subsetter.subset(font)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2783, in subset
    self._prune_post_subset(font)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2766, in _prune_post_subset
    retain = table.prune_post_subset(self.options)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2212, in prune_post_subset
    charstring.draw(pen)  # this will set the charstring's width
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/misc/psCharStrings.py", line 282, in draw
    self.private.nominalWidthX, self.private.defaultWidthX)
AttributeError: 'NoneType' object has no attribute 'nominalWidthX'
```

Again, apologies for the revert commit.

### DIFF
--- a/Lib/fontTools/misc/psCharStrings.py
+++ b/Lib/fontTools/misc/psCharStrings.py
@@ -642,19 +642,57 @@ class SimpleT2Decompiler(object):
 	def op_roll(self, index):
 		raise NotImplementedError
 
-class T2OutlineExtractor(SimpleT2Decompiler):
 
-	def __init__(self, pen, localSubrs, globalSubrs, nominalWidthX, defaultWidthX):
+class T2WidthExtractor(SimpleT2Decompiler):
+
+	def __init__(self, localSubrs, globalSubrs, nominalWidthX, defaultWidthX):
 		SimpleT2Decompiler.__init__(self, localSubrs, globalSubrs)
-		self.pen = pen
 		self.nominalWidthX = nominalWidthX
 		self.defaultWidthX = defaultWidthX
 
 	def reset(self):
 		SimpleT2Decompiler.reset(self)
-		self.hints = []
 		self.gotWidth = 0
 		self.width = 0
+
+	def popallWidth(self, evenOdd=0):
+		args = self.popall()
+		if not self.gotWidth:
+			if evenOdd ^ (len(args) % 2):
+				self.width = self.nominalWidthX + args[0]
+				args = args[1:]
+			else:
+				self.width = self.defaultWidthX
+			self.gotWidth = 1
+		return args
+
+	def countHints(self):
+		args = self.popallWidth()
+		self.hintCount = self.hintCount + len(args) // 2
+
+	def op_rmoveto(self, index):
+		self.popallWidth()
+
+	def op_hmoveto(self, index):
+		self.popallWidth(1)
+
+	def op_vmoveto(self, index):
+		self.popallWidth(1)
+
+	def op_endchar(self, index):
+		self.popallWidth()
+
+
+class T2OutlineExtractor(T2WidthExtractor):
+
+	def __init__(self, pen, localSubrs, globalSubrs, nominalWidthX, defaultWidthX):
+		T2WidthExtractor.__init__(
+			self, localSubrs, globalSubrs, nominalWidthX, defaultWidthX)
+		self.pen = pen
+
+	def reset(self):
+		T2WidthExtractor.reset(self)
+		self.hints = []  # XXX this attribute seems to be unused
 		self.currentPoint = (0, 0)
 		self.sawMoveTo = 0
 
@@ -688,21 +726,6 @@ class T2OutlineExtractor(SimpleT2Decompiler):
 		# In T2 there are no open paths, so always do a closePath when
 		# finishing a sub path.
 		self.closePath()
-
-	def popallWidth(self, evenOdd=0):
-		args = self.popall()
-		if not self.gotWidth:
-			if evenOdd ^ (len(args) % 2):
-				self.width = self.nominalWidthX + args[0]
-				args = args[1:]
-			else:
-				self.width = self.defaultWidthX
-			self.gotWidth = 1
-		return args
-
-	def countHints(self):
-		args = self.popallWidth()
-		self.hintCount = self.hintCount + len(args) // 2
 
 	#
 	# hint operators
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -7,7 +7,7 @@ from fontTools.misc.py23 import *
 from fontTools import ttLib
 from fontTools.ttLib.tables import otTables
 from fontTools.misc import psCharStrings
-from fontTools.pens.boundsPen import BoundsPen
+from fontTools.pens.basePen import NullPen
 from fontTools.misc.loggingTools import Timer
 import sys
 import struct
@@ -1851,8 +1851,8 @@ def prune_pre_subset(self, font, options):
                 private = font.Private
             dfltWdX = private.defaultWidthX
             nmnlWdX = private.nominalWidthX
-            pen = BoundsPen(None)
-            c.draw(pen) # this will set the charstring's width
+            pen = NullPen()
+            c.draw(pen)  # this will set the charstring's width
             if c.width != dfltWdX:
                 c.program = [c.width - nmnlWdX, 'endchar']
             else:
@@ -2211,7 +2211,9 @@ def prune_post_subset(self, options):
                 subrs = getattr(c.private, "Subrs", [])
                 decompiler = _DehintingT2Decompiler(css, subrs, c.globalSubrs)
                 decompiler.execute(c)
+            pen = NullPen()
             for charstring in css:
+                charstring.draw(pen)  # this will set the charstring's width
                 charstring.drop_hints()
             del css
 
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -1964,7 +1964,7 @@ class _MarkingT2Decompiler(psCharStrings.SimpleT2Decompiler):
         self.globalSubrs._used.add(self.operandStack[-1]+self.globalBias)
         psCharStrings.SimpleT2Decompiler.op_callgsubr(self, index)
 
-class _DehintingT2Decompiler(psCharStrings.SimpleT2Decompiler):
+class _DehintingT2Decompiler(psCharStrings.T2WidthExtractor):
 
     class Hints(object):
         def __init__(self):
@@ -1985,17 +1985,16 @@ class _DehintingT2Decompiler(psCharStrings.SimpleT2Decompiler):
             self.has_hintmask = False
         pass
 
-    def __init__(self, css, localSubrs, globalSubrs):
+    def __init__(self, css, localSubrs, globalSubrs, nominalWidthX, defaultWidthX):
         self._css = css
-        psCharStrings.SimpleT2Decompiler.__init__(self,
-                                                  localSubrs,
-                                                  globalSubrs)
+        psCharStrings.T2WidthExtractor.__init__(
+            self, localSubrs, globalSubrs, nominalWidthX, defaultWidthX)
 
     def execute(self, charString):
         old_hints = charString._hints if hasattr(charString, '_hints') else None
         charString._hints = self.Hints()
 
-        psCharStrings.SimpleT2Decompiler.execute(self, charString)
+        psCharStrings.T2WidthExtractor.execute(self, charString)
 
         hints = charString._hints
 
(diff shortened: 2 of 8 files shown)


behdad:
What's the status of this?

anthrotype:
@behdad status is: the issue with `--no-hinting` stripping CFF advance widths still remains; and I'm not sure how to fix this, as I explained [above](https://github.com/fonttools/fonttools/pull/709#issue-183203335)

anthrotype:
just rebased on upstream master and fixed the merge conflict.

anthrotype:
it would be nice to have this fixed before the next 3.5.0 release.

To replicate the issue, check out this PR branch, then take for example Adobe Caslon Regular and run the subsetter with `--no-hinting` option (but without `--desubroutinize`):

```shell
$ pyftsubset ~/Library/Fonts/ACaslonPro-Regular.otf --no-hinting '*'
  File "/usr/local/var/pyenv/versions/fonttools360/bin/pyftsubset", line 11, in <module>
    load_entry_point('fonttools', 'console_scripts', 'pyftsubset')()
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/misc/loggingTools.py", line 372, in wrapper
    return func(*args, **kwds)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2973, in main
    subsetter.subset(font)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2789, in subset
    self._prune_post_subset(font)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2772, in _prune_post_subset
    retain = table.prune_post_subset(self.options)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2216, in prune_post_subset
    charstring.draw(pen)  # this will set the charstring's width
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/misc/psCharStrings.py", line 282, in draw
    self.private.nominalWidthX, self.private.defaultWidthX)
AttributeError: 'NoneType' object has no attribute 'nominalWidthX'
```

Note that if you also pass `--desubroutinize` option, it seems to work fine.

behdad:
Humm.  I don't have Adobe Caslon Regular.  But, nominalWidthX and defaultWidthX both default to zero.  Can you adjust the code to use zero if private is not available perhaps?

anthrotype:
Thanks. I'll try and let you know how it goes.
(btw, I just sent you a copy via email)

behdad:
> $ pyftsubset ~/Library/Fonts/ACaslonPro-Regular.otf --no-hinting '*'

This works for me with the font you sent.

anthrotype:
I tried applying the following patch to fallback to 0 when `self.private` has no `nominalWidthX` or `defaultWidthX` attributes, because it's `None`:
```diff
diff --git a/Lib/fontTools/misc/psCharStrings.py b/Lib/fontTools/misc/psCharStrings.py
index a75b433..fc5e3a4 100644
--- a/Lib/fontTools/misc/psCharStrings.py
+++ b/Lib/fontTools/misc/psCharStrings.py
@@ -279,7 +279,8 @@ class T2CharString(ByteCodeBase):
        def draw(self, pen):
                subrs = getattr(self.private, "Subrs", [])
                extractor = T2OutlineExtractor(pen, subrs, self.globalSubrs,
-                               self.private.nominalWidthX, self.private.defaultWidthX)
+                               getattr(self.private, "nominalWidthX", 0),
+                               getattr(self.private, "defaultWidthX", 0))
                extractor.execute(self)
                self.width = extractor.width
```

However, now I get a different error, using the same file I sent you:

```shell
$ pyftsubset "/Users/cosimolupo/Downloads/ACaslonPro-Regular.otf" --no-hinting '*'
  File "/usr/local/var/pyenv/versions/fonttools360/bin/pyftsubset", line 11, in <module>
    load_entry_point('fonttools', 'console_scripts', 'pyftsubset')()
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/misc/loggingTools.py", line 372, in wrapper
    return func(*args, **kwds)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2973, in main
    subsetter.subset(font)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2789, in subset
    self._prune_post_subset(font)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2772, in _prune_post_subset
    retain = table.prune_post_subset(self.options)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2216, in prune_post_subset
    charstring.draw(pen)  # this will set the charstring's width
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/misc/psCharStrings.py", line 284, in draw
    extractor.execute(self)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/misc/psCharStrings.py", line 530, in execute
    rv = handler(index)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/misc/psCharStrings.py", line 575, in op_callgsubr
    subr = self.globalSubrs[subrIndex+self.globalBias]
```

anthrotype:
I'm sorry, CFF is a bit over my head. I'll leave it to you :(

behdad:
> I'm sorry, CFF is a bit over my head. I'll leave it to you :(

As I said, the subset command that you said reproduces the issue, works for me.  It's extremely broken if private is None.  So let's dig deeper into why you are seeing this.

behdad:
For example, load the font and make sure that before subsetting private is NOT none.  Then see where it's disappearing.

I have had seen unusual things happening in that part of the subsetter.  Garret reported something broken with some version of Noto Sans JP, and I could reproduce it and it was a very confused state of renumbering subroutines.  While not the same, it does suggest that there are things badly broken there, so don't trust it.

anthrotype:
> As I said, the subset command that you said reproduces the issue, works for me. It's extremely broken if private is None.

so you say it's working on your side _without_ applying any patch at all, on current master?
This is super weird...

behdad:
> so you say it's working on your side without applying any patch at all, on current master?
> This is super weird...

Yes.

anthrotype:
Behdad, when you say it works from current master, you mean the command completes without errors. However the generated CFF charsrrings are stripped of all advance widths.
Which is the issue this PR is trying to address.

You have to check out my PR branch to see the error when doing --no-hinting with Adobe Caslon. Master does not have the patch as we reverted #606 because of the issue with private.

I hope this is clear now.

behdad:
> Behdad, when you say it works from current master, you mean the command completes without errors. However the generated CFF charsrrings are stripped of all advance widths.
> Which is the issue this PR is trying to address.
> 
> You have to check out my PR branch to see the error when doing --no-hinting with Adobe Caslon. Master does not have the patch as we reverted #606 because of the issue with private.

I see.  Thanks for the clarification.  I'll check.

anthrotype:
This subroutinized and hinted OTF file contains a single glyph and reproduces the issue:
[Lobster_B_subroutinized_and_hinted.otf.zip](https://github.com/fonttools/fonttools/files/705200/Lobster_B_subroutinized_and_hinted.otf.zip)

Steps to reproduce:

1) checkout my PR branch `cff-dehint-2`
2) unzip the attached file
3) run the following command:

```
$ fonttools subset Lobster_B_subroutinized_and_hinted.otf --no-hinting '*'
```

You should get this traceback:

```
  File "/usr/local/var/pyenv/versions/fonttools2/bin/fonttools", line 11, in <module>
    load_entry_point('fonttools', 'console_scripts', 'fonttools')()
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/__main__.py", line 29, in main
    runpy.run_module(mod, run_name='__main__')
  File "/usr/local/var/pyenv/versions/2.7.10/lib/python2.7/runpy.py", line 180, in run_module
    fname, loader, pkg_name)
  File "/usr/local/var/pyenv/versions/2.7.10/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/subset/__main__.py", line 7, in <module>
    sys.exit(main())
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/misc/loggingTools.py", line 372, in wrapper
    return func(*args, **kwds)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2976, in main
    subsetter.subset(font)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2790, in subset
    self._prune_post_subset(font)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2773, in _prune_post_subset
    retain = table.prune_post_subset(self.options)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2216, in prune_post_subset
    charstring.draw(pen)  # this will set the charstring's width
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/misc/psCharStrings.py", line 282, in draw
    self.private.nominalWidthX, self.private.defaultWidthX)
AttributeError: 'NoneType' object has no attribute 'nominalWidthX'
```

anthrotype:
I updated the tests and expected test data in 23e3107 (I'm now using a subset of Lobster).

If I use the NullPen to draw the charstrings before stripping the hints (in order to set the width attribute), then the `test_no_hinting_CFF` fails with `AttributeError: 'nominalWidthX'`:

https://travis-ci.org/fonttools/fonttools/jobs/191919899

If you run the same tests against current `master` (revert `0eca9b0`), you will see they also reproduce the original issue -- the charstrings' advance widths are being stripped along with the hints. (OSX reads the widths from the CFF charstrings instead of hmtx...)

In the Lobster.subset.ttx, the AttributeError issue only arises with the charstrings "B" and "B.salt". These both call the subroutine index=2 (`-105 callsubr`).

Now, it seems that this subroutine contains one of those "tricky" hintmasks, as @behdad describes them in an inline comment:

>  A hintmask right after hints and a few numbers will act as an implicit vstemhm.

I have no clue how to fix this, so I'll tag v3.5.0 without this patch.

anthrotype:
I rebased the PR on master to accommodate the changes in the test folders layout.

The CI failure for commit 722328d clearly reveals the bug in the current implementation of `--no-hinting`, which also strips the charstrings' advance widths:

https://travis-ci.org/fonttools/fonttools/jobs/192764045#L1305

The following commit (6185a9f), which is meant to fix the previous bug, also fails but for a different reason, i.e. the above mentioned `AttributeError`:

https://travis-ci.org/fonttools/fonttools/jobs/192765664#L1310

anthrotype:
This patch makes the test pass! 🎉 

```diff
diff --git a/Lib/fontTools/subset/__init__.py b/Lib/fontTools/subset/__init__.py
index b05bd48..b6e9ccc 100644
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -2048,7 +2048,7 @@ class _DehintingT2Decompiler(psCharStrings.SimpleT2Decompiler):
         cs = self.callingStack[-1]
         hints = cs._hints
         hints.has_hintmask = True
-        if hints.status != 2 and hints.has_hint:
+        if hints.status != 2:
             # Check from last_check, see if we may be an implicit vstem
             for i in range(hints.last_checked, index - 1):
                 if isinstance(cs.program[i], str):
@@ -2056,6 +2056,7 @@ class _DehintingT2Decompiler(psCharStrings.SimpleT2Decompiler):
                     break
             if hints.status != 2:
                 # We are an implicit vstem
+                hints.has_hint = True
                 hints.last_hint = index + 1
                 hints.status = 0
         hints.last_checked = index + 1
@@ -2205,15 +2206,17 @@ def prune_post_subset(self, options):
             #     we have seen any non-hint operators so far and do the right
             #     thing, recursively... Good luck understanding that :(
             css = set()
+            pen = NullPen()
             for g in font.charset:
                 c,sel = cs.getItemAndSelector(g)
                 c.decompile()
+                # this will set the charstring's width
+                c.draw(pen)
                 subrs = getattr(c.private, "Subrs", [])
                 decompiler = _DehintingT2Decompiler(css, subrs, c.globalSubrs)
                 decompiler.execute(c)
-            pen = NullPen()
+                c.draw(pen)  # this will set the charstring's width
             for charstring in css:
-                charstring.draw(pen)  # this will set the charstring's width
                 charstring.drop_hints()
             del css
```

however... it's stll not clear to me why this works! 😁 
Well, I'll investigate tomorrow -- too tired of Type2 charstrings.

anthrotype:
@behdad could you have a quick look at this?

I think I've identified the root of the issue, with the hintmask operator acting as an implicit vstem but from inside a subroutine.

The patch above applied on this branch seems to fix things, that is I finally seem to be able to strip the hinting and keep the charsrrings width, without also needing to desubroutinize at the same time to work around it.

behdad:
Thanks. Will check tomorrow.

anthrotype:
OK, the tests now pass!

The subsetter now correctly keeps the advance widths when doing `--no-hinting` (dcaf27d), and doesn't leave behind vstemhm hints when `hintmask` occurs inside a subroutine (2b76efb).

I'll release 3.6.0 now without this one to give @behdad some time to review it.
We can always release 3.6.x later.

brawer:
@anthrotype and @behdad, what to do with this pull request?

PR #714, opened 2016-10-20
[ttx/xmlWriter] allow to override OS-specific line endings 

anthrotype:
When TTX files are under version control and are being modified across multiple platforms, it's good to be able to _not_ use the default `os.linesep`, currently used by `XMLWriter`, but instead specify one particular line ending for all different platforms (e.g. most likely, LF).

This PR adds a `newlinestr` keyword argument to `XMLWriter` constructor, as well as to `TTFont.saveXML` method.
It also adds a `--newline` option to `ttx` console script.

The default does not change (ie. we keep using the native line endings).

However, we may wish to revise this in the future and always write Unix-style line feeds, since the XML specification says XML processor have to normalize them as such anyway. See https://www.w3.org/TR/REC-xml/#sec-line-ends

BTW, lxml always writes `"\n"` as line endings when `pretty_print`-ing XML, even on Windows.

### DIFF
--- a/Lib/fontTools/misc/testTools.py
+++ b/Lib/fontTools/misc/testTools.py
@@ -61,9 +61,8 @@ class TestXMLReader_(object):
 
 
 def makeXMLWriter():
-    writer = XMLWriter(BytesIO())
     # don't write OS-specific new lines
-    writer.newlinestr = writer.totype('')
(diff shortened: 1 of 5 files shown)


justvanrossum:
Looks good to me. I wouldn't be opposed to make the default `'\n'`.

benkiel:
Agree to making the default '\n'

PR #720, opened 2016-10-24
Feaplus

mhosken:
This adds two capabilities to feaLib
- Ignore a missing tagname after a block, except for anonymous blocks
- Add support for anon py { } py; type blocks with limited reparsing capability.

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -75,6 +75,9 @@ class AnonymousBlock(Statement):
         Statement.__init__(self, location)
         self.tag, self.content = tag, content
 
+    def build(self, builder):
+        builder.process_anon_block(self.content, self.tag)
+
 
 class Block(Statement):
     def __init__(self, location):
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -77,7 +77,8 @@ class Builder(object):
         self.python = {}
 
     def build(self):
-        self.parseTree = Parser(self.file).parse()
+        self.parser = Parser(self.file)
+        self.parseTree = self.parser.parse()
         self.parseTree.build(self)
         self.build_feature_aalt_()
         self.build_head()
@@ -1000,6 +1001,11 @@ class Builder(object):
                     'builder' : self
                 }
             exec content in self.python
+
+# Methods to provide capabilities to anon py blocks on builder
+    def parse_string(self, data) :
+        doc = self.parser.parse_string(data)
+        doc.build(self)
 
 
 def makeOpenTypeAnchor(anchor):
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -1000,7 +1000,7 @@ class Builder(object):
                     # 'parser' : self.parser,
                     'builder' : self
                 }
-            exec content in self.python
+            exec (content, self.python)
 
 # Methods to provide capabilities to anon py blocks on builder
     def parse_string(self, data) :
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -73,6 +73,8 @@ class Builder(object):
         self.hhea_ = {}
         # for table 'vhea'
         self.vhea_ = {}
+        # for anon py blocks
+        self.python = {}
 
     def build(self):
         self.parseTree = Parser(self.file).parse()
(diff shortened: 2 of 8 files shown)


anthrotype:
Hello @mhosken, thanks for your pull request.
I'm not fully convinced about these changes.

1) about allowing missing labels after a block, I don't see that anywhere in the current Feature File specification. Was it discussed elsewhere maybe?

2) about executing arbitrary python code from a feature file with `exec`, I don't think that would be safe... What's the use case you have in mind? Isn't `anon` meant to contain data rather than code?

adrientetar:
You should probably use a preprocessor if you want to generate features, e.g. [feaPyFoFum](https://github.com/typesupply/feaPyFoFum).

mhosken:
Both of these issues were discussed with @brawer and with Adobe at a recent ATypI. In more detail:

1. There is no need for the final block name in all cases except anon blocks. The only purpose of the final block name is as a check that it is the same as the block name. But the real test of correct block names is between declaration and reference. This is a redundant check and just annoying for those writing fea. About the only useful use for it is for very large (> 1 page) blocks to help a reader keep track. Hence making the need for it optional. If it is there, it must match.

2. If I were content with generating all my fea from code then I would use one of these fea generators. But I want to have the python interact with the fea that a font designer has written or generated. This requires access to both the builder and the parser at build time. The current patch doesn't do all that yet and I have questions as to the best way to do that (how do we expose the various symbol tables in the parser that should be accessible?) While an anon block was intended for other data, there is nothing to stop us using it for code in this way and I think it is a natural solution to the problem.

    As to the security issue. It's real. This code calls exec without doing thorough checking of the code it is passing. But given any compilation process will write to an arbitrary file, there is nothing to stop someone overwriting some key file in someone's home directory with, in effect, garbage. So in that respect, it is just as secure as the environment in which it is run.

brawer:
Regarding making the final block names optional: Sorry, but we really shouldn’t unilaterally deviate from the specification. To change the feature file syntax, can you make a proposal with Adobe? File an [issue on AFDKO](https://github.com/adobe-type-tools/afdko/issues) and discuss it there; if and when there’s agreement to change the spec, fonttools will definitely follow.

Regarding the interaction with Python, I wonder if there might be a better (less intrusive) way for doing what you want. For example, if feaLib allowed you to register a Python callback that gets invoked whenever an `anon` block has been parsed, you could do whatever you want inside your callback, including evaluating the block’s text. But this logic would live in your own code, separate from the core library. Would this work for you?

anthrotype:
I'd be curious to see more concrete examples of such macro type capabilities applied to fea.

I also like the idea of registering a callback that gets invoked when `anon` is parsed, so the responsibility of executing arbitrary code is on the client's side.

adrientetar:
> This is a redundant check and just annoying for those writing fea.

I do not disagree with this particular proposal, but note autocompletion can do it for you (this is what TruFont does).

mhosken:
I've raised the issue with afdko. We'll see if there is any response.

An alternative is to put the python support into a builder subclass. I would suggest that I don't hide the python support away in our tools in the interest of sharing. But that doesn't require clients to use the subclass (i.e. ship fealib with both classes).

At the very least we need the anon block handler in the builder and for the builder to hold a reference to the parser.

justvanrossum:
Have you actually looked at [feaPyFoFum](https://github.com/typesupply/feaPyFoFum)? As far as I can see it does what you describe: have "python interact with the fea that a font designer has written or generated".

mhosken:
@justvanrossum yes I have. Looking carefully at feaPyFoFum reveals it provides no capability to access any information created in the rest of the fea in the file. It can access the font. But there is no way to, for example:

. Add rules to an existing lookup or query a lookup to see whether it exists or what is in it

. Query, or change a fea class

. Interact with (query or change) the anchors defined in fea

The only way to provide such capability is to integrate such capability into a fully fledged fea syntactic parser and builder, such as feaLib. And when you do integrate it, it becomes something relatively simple.

LettError:
Relatively simple? Go for it!

behdad:
I'm strictly against running code from .fea files.  That's a security nightmare.

khaledhosny:
Sometime the feature file syntax can get into the way and a bit of code can be handy.

Case in the point: you can’t do `sub @class1 by @class2 glyph1;` with feature syntax, forcing one to not use the classes here and keep code in sync manually. But if I can insert some code for such special cases that have access to the parsed feature data I can just read the classes at build time and write the substitutions automatically. The other option is to fix the feature syntax, but this is takes longer and might not practical if your needs are special.

So I personally wold love to see this, but we can make sure such code is never executed unless explicitly requested.

davelab6:
> I'm strictly against running code from .fea files. That's a security nightmare.

Isn't @mhosken 's proposal a preprocessor, though?

anthrotype:
This was superseded by #794, can we close it?

brawer:
Indeed, superseded by #794.

Issue #699, opened 2016-10-11
varLib `_merge_OTL` assumes presence of GPOS table


djrrb:
Is there a reason that a GPOS table is required to generate a GX font using `varLib/__init__.py`? I was trying out a monospaced font and hitting an error. Removing the `"GPOS"` string from the tables argument of `merge_tables()` seemed to fix it, but I wasn't sure if you had reasons for hardcoding that table in.

anthrotype:
passing an empty list to `merge_tables` is like skipping it altogether.
I guess the whole `_merge_OTL` function should be skipped in there's no GPOS -- but I'm not 100% sure. /cc @behdad

PR #721, opened 2016-10-24
Add CFF2 table support to fonttools

readroberts:
Includes simple test CFF2 data.


Lib/fontTools/varLib/__init__.py
```diff
@@ -313,17 +313,18 @@ def _merge_OTL(font, model, master_fonts, axisTags, base_idx):
...
-	try:
-		GDEF = font['GDEF'].table
-		assert GDEF.Version <= 0x00010002
-	except KeyError:
-		font['GDEF']= newTable('GDEF')
-		GDEFTable = font["GDEF"] = newTable('GDEF')
-		GDEF = GDEFTable.table = ot.GDEF()
-	GDEF.Version = 0x00010003
-	GDEF.VarStore = store
+	if 'GPOS' in font:
```
anthrotype: Is this `if 'GPOS' in font:` meant to address issue https://github.com/fonttools/fonttools/issues/699? 
if so, maybe it should be part of a separate commit/pull request.

Lib/fontTools/ttLib/tables/C_F_F_.py
```diff
@@ -3,7 +3,6 @@
 from fontTools import cffLib
 from . import DefaultTable
 
-
```
anthrotype: why removing the blank line? Top-level classes/functions definitions should be separated by two blank lines in PEP8 style guide.

Lib/fontTools/misc/psCharStrings.py
```diff
@@ -1113,7 +1194,7 @@ def __init__(self, strings):
 		self.dict = {}
 
 	def getDict(self):
-		assert len(self.stack) == 0, "non-empty stack"
+		#assert len(self.stack) == 0, "non-empty stack"
```
anthrotype: any reasons for commenting this out? should the line be removed?

Lib/fontTools/cffLib.py
```diff
@@ -27,21 +32,38 @@ def __init__(self):
...
-				"unknown CFF format: %d.%d" % (self.major, self.minor)
-
-		file.seek(self.hdrSize)
-		self.fontNames = list(Index(file))
-		self.topDictIndex = TopDictIndex(file)
-		self.strings = IndexedStrings(file)
-		self.GlobalSubrs = GlobalSubrsIndex(file)
-		self.topDictIndex.strings = self.strings
-		self.topDictIndex.GlobalSubrs = self.GlobalSubrs
+		global isCFF2
```
anthrotype: is this use of a global variable `isCFF2` strictly necessary?

Lib/fontTools/misc/psCharStrings.py
```diff
@@ -164,6 +170,14 @@ def buildOperatorDict(operatorList):
...
 ]
 
+CFF2Operators = [entry for entry in t2Operators if ((isinstance(entry[0], int) or
+											(entry[0][1] < 34)) and (not entry[0] in [11,14])) ]
+CFF2Operators.extend(
+[
+	(15,		'vsindex'),
+	(16,		'blend'),
+])
+CFF2Operators.sort()
```
anthrotype: this fails in Python 3 as you are comparing `int` and `tuple`.
https://ci.appveyor.com/project/fonttools/fonttools/build/1.0.123/job/i4k20he0fk551xat#L72

Lib/fontTools/varLib/__init__.py
readroberts: Yes, it is the same issue as #699. I am still somewhat new to github. Once I have submitted a pull request, can I back out specific changes?

Lib/fontTools/ttLib/tables/C_F_F_.py
readroberts: My error, then.

Lib/fontTools/misc/psCharStrings.py
readroberts: The line should be removed.  I will do this.

Lib/fontTools/cffLib.py
readroberts: More accurately, I would need to establish as parent-child relationship that does not now exist throughout the entire tree of structures. In some cases this can be done with adding a new argument to carry the context, in others the subclasses would need to be able to access parent classes.

Lib/fontTools/cffLib.py
```diff
@@ -1293,13 +1726,25 @@ def xmlRead(self, name, attrs, content, parent):
...
+]
+
+topDictOperators2 = [
+#	opcode		name			argument type	default	converter
+	((12, 7),	'FontMatrix',		'array',	[0.001, 0, 0, 0.001, 0, 0],	None),
+	((12, 37),	'FDSelect',		'number',	None,	FDSelectConverter()),
+	((12, 36),	'FDArray',		'number',	None,	FDArrayConverter()),
+	(17,		'CharStrings',		'number',	None,	CharStringsConverter()),
+	(24,		'VarStore',		'number',	None,	VarStoreConverter()),
+	(25,		'maxstack',		'number',	None,	None),
```
mashabow: According to [the spec](https://www.microsoft.com/typography/otspec/cff2.htm#section7), `vstore` instead of `VarStore`, and default value of `maxstack` is 193.

Lib/fontTools/cffLib.py
```diff
@@ -1340,6 +1785,24 @@ def xmlRead(self, name, attrs, content, parent):
...
+	(9,		'FamilyOtherBlues',	'delta',	None,	None),
+	((12, 9),	'BlueScale',		'number',	0.039625, None),
+	((12, 10),	'BlueShift',		'number',	7,	None),
+	((12, 11),	'BlueFuzz',		'number',	1,	None),
+	(10,		'StdHW',		'number',	None,	None),
+	(11,		'StdVW',		'number',	None,	None),
+	((12, 12),	'StemSnapH',		'delta',	None,	None),
+	((12, 13),	'StemSnapV',		'delta',	None,	None),
+	(19,		'Subrs',		'number',	None,	SubrsConverter()),
+]
```
mashabow: `LanguageGroup` and `ExpansionFactor` are missing.

Lib/fontTools/cffLib.py
readroberts: I have thought about how to do this. If this code needs to be thread safe and re-entrant, I can certainly make the changes, but would rather do so after the current logic has proven stable,

readroberts:
Right, I saw that the Python 3 integration test failed soon after I made the pull request. I fixed the particular issue you noted, and am working on some other issues with use of the map function. Should have this checked in today.

- Read

readroberts:
Closing so that I can re do PR for main branch of fonttools/fonttools.

Issue #457, opened 2016-01-11
[feaLib] backslash-prefixed glyph names not supported


adrientetar:
Given the following feature file:

```
@group = [\f \i];

feature liga {
    sub @group by f_i;
} liga;
```

Traceback:

```
Traceback (most recent call last):
  File "C:\Users\Admin\Downloads\trufont\Lib\defconQt\scriptingWindow.py", line 40, in runScript
    exec(code, global_vars)
  File "<string>", line 8, in <module>
  File "C:\Python34\lib\site-packages\defcon\objects\base.py", line 337, in getRepresentation
    representation = factory["factory"](self, **kwargs)
  File "C:\Users\Admin\Downloads\trufont\Lib\defconQt\representationFactories\openTypeFactory.py", line 5, in TTFontFactory
    otf = compileOTF(font)
  File "C:\Python34\lib\site-packages\ufo2ft-0.1-py3.4.egg\ufo2ft\__init__.py", line 29, in compileOTF
    featureCompilerClass, kernWriter, markWriter)
  File "C:\Python34\lib\site-packages\ufo2ft-0.1-py3.4.egg\ufo2ft\__init__.py", line 18, in _compile
    featureCompiler.compile()
  File "C:\Python34\lib\site-packages\ufo2ft-0.1-py3.4.egg\ufo2ft\makeotfParts.py", line 26, in compile
    self.setupFile_features()
  File "C:\Python34\lib\site-packages\ufo2ft-0.1-py3.4.egg\ufo2ft\makeotfParts.py", line 59, in setupFile_features
    autoFeatures["kern"] = self.writeFeatures_kern()
  File "C:\Python34\lib\site-packages\ufo2ft-0.1-py3.4.egg\ufo2ft\makeotfParts.py", line 82, in writeFeatures_kern
    writer = self.kernWriter(self.font)
  File "C:\Python34\lib\site-packages\ufo2ft-0.1-py3.4.egg\ufo2ft\kernFeatureWriter.py", line 30, in __init__
    parser.parseFeatures(self, font.features.text)
  File "C:\Python34\lib\site-packages\feaTools\parser.py", line 645, in parseFeatures
    _parseUnknown(writer, text)
  File "C:\Python34\lib\site-packages\feaTools\parser.py", line 300, in _parseUnknown
    _parseUnknown(writer, precedingText)
  File "C:\Python34\lib\site-packages\feaTools\parser.py", line 408, in _parseUnknown
[...]

PR #732, opened 2016-11-02
[feaLib] support backslash-escaped glyphs in FEA

anthrotype:
Backslash-prefixed glyph name can be used in a Feature file to distinguish them from identically-named keywords.

From section "2.f.i. Glyph name" of Adobe's Feature File Specification:

> An initial backslash serves to differentiate a glyph name from an identical keyword in the feature file language. For example, a glyph named "table" must be specified in the feature file as: \table

Thus, when we parse a glyph name that begins with a backslash, we need to ignore the first character.

Note that makeotf rejects feature files with glyph names that start with or contain backslashes escaped with another backslash.

```
feature liga {
    sub \\glyphWithBackslash by A;
} liga;
```

This yields:

    makeotfexe [FATAL] <Backslash-Regular> invalid token (text was "\") [features 2]

Fixes https://github.com/fonttools/fonttools/issues/457

/cc @brawer @adrientetar

### DIFF
--- a/Lib/fontTools/feaLib/builder_test.py
+++ b/Lib/fontTools/feaLib/builder_test.py
@@ -39,7 +39,7 @@ def makeTTFont():
         a_n_d T_h T_h.swash germandbls ydieresis yacute breve
         grave acute dieresis macron circumflex cedilla umlaut ogonek caron
         damma hamza sukun kasratan lam_meem_jeem noon.final noon.initial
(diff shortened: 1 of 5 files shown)


anthrotype:
similarly, `feaLib.lexer` only recognizes `\` at the start of a character name token, but not in its continuation.

So basically the backslash is a special reserved character and cannot/should not appear in real glyph names (ie. the `post` or `CFF ` table ones).

PR #742, opened 2016-11-21
Fix "Private Use (plane 15)" range

mashabow:
https://twitter.com/ken_lunde/status/796564061206913024
https://www.microsoft.com/typography/otspec/errata.htm


justvanrossum:
Thanks!

Issue #753, opened 2016-11-30
py23.round3 is slow


jamesgk:
Following from https://github.com/fonttools/fonttools/pull/665

The round3 function in py23 (used prominently in t2CharStringPen) is slow. Would be nice to optimize it somehow. From @behdad:

> It occurs to me that if we avoid rounding if number is already int, and avoiding the power if ndigits is zero, plus other straightforward optimizations, might make this go away.

> I also think we can remove the checks for `__round__` and `__index__`.

jamesgk:
Oh man, adding

```diff
--- a/Lib/fontTools/misc/py23.py
+++ b/Lib/fontTools/misc/py23.py
@@ -310,6 +310,9 @@ def round3(number, ndigits=None):
        Derived from python-future:
        https://github.com/PythonCharmers/python-future/blob/master/src/future/builtins/newround.py
        """
+       if isinstance(number, int) or int(number) == number:
+               return number
+
        return_int = False
        if ndigits is None:
                return_int = True
```

seems to make a huge difference indeed.

behdad:
That's what I thought.  Do you need the first check?  The second might do just fine.

behdad:
Though, you probably get YUUUGE speedup for masters, but not much for interpolated instances.

jamesgk:
Well I have been using it on instances, but it turns out mutatormath rounds coordinates during interpolation by default and fontmath seems to have its own implementation of this: https://github.com/typesupply/fontMath/blob/master/Lib/fontMath/mathFunctions.py#L55.

anthrotype:
See https://github.com/typesupply/fontMath/pull/36

anthrotype:
I think we squeezed what we could from round3.

If you are still concerned, you could try this backport of the built-in Python 3 `round`; it's a C extension module for Python 2.7, with code taken straight off Python 3.5 `floatobject.c`.

https://github.com/anthrotype/round3

jamesgk:
Nah, this has been significantly improved. Thank you!

PR #754, opened 2016-11-30
[py23] optimize round3 function

anthrotype:
See https://github.com/fonttools/fonttools/issues/753

In addition to the optimizations suggested by Behdad and James, I also made so that round2 is only defined in Python 3, and round3 only in Python 2, with the other ones just pointing to the respective built-ins.

I also dropped support for the second argument, ndigits, having a `__index__` attribute, which is a bit niche, I think.

/cc @behdad @jamesgk

### DIFF
--- a/Lib/fontTools/misc/py23.py
+++ b/Lib/fontTools/misc/py23.py
@@ -261,90 +261,94 @@ def xrange(*args, **kwargs):
 
 import decimal as _decimal
 
+if PY3:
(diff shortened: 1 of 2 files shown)


anthrotype:
and I also dropped the feature whereby `round3` delegates to the `__round__` method of the number (that was cool though...)

jamesgk:
Best way to find out is to benchmark it and see!

PR #755, opened 2016-12-01
[py23] further optimizie round3 function

anthrotype:
in normal circumstances, we get a huge speed improvement if we use the built-in Python 2 `round()` function for those values which we know will produce the same results as our slow implementation of Python 3 rounding (that uses the pure-python `decimal` module).

The latter will only be used when a number is half-way between the two multiples of 10 to the power `ndigits`, and the multiple towards zero is even.

When I check if the float number is "half-way" I can't simply take the `% 1 == .5`, because floating point math is funny. For that I use `math.isclose()` function (added in Python 3.5), for which I add a pure-python backport.

Checking that a float `isclose` is still faster than rounding it using the slow `Decimal` method.

@jamesgk @behdad

### DIFF
--- a/Lib/fontTools/misc/py23.py
+++ b/Lib/fontTools/misc/py23.py
@@ -259,7 +259,48 @@ def xrange(*args, **kwargs):
 	raise Py23Error("'xrange' is not defined. Use 'range' instead.")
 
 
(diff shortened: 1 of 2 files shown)


behdad:
lgtm

PR #757, opened 2016-12-02
[py23] fix inconsistency of built-in round on Python 3.5 if second argument is None

anthrotype:
This round thing is killing me..

There's a problem with the built-in `round()` function on Python 3.5.

If one passes `None` as a second argument, and the first argument is a float, everything is ok 

```python
>>> round(1.0, None)
1
```

But if the first argument is an integer and you pass None as second argument, it raises:

```python
>>> round(1, None)
TypeError: 'NoneType' object cannot be interpreted as an integer
```

Now, this happens to be a known issue (https://bugs.python.org/issue27936) which has been fixed in Python 3.6.

Here, on Python3 below 3.6, I have to wrap the built-in round() so that when it gets None as second argument, it behaves as if one did not provide the second argument at all.

See also https://github.com/typesupply/fontMath/pull/37#issuecomment-264435328

### DIFF
--- a/Lib/fontTools/misc/py23.py
+++ b/Lib/fontTools/misc/py23.py
@@ -360,7 +360,7 @@ else:
 		Implementation of Python 3 built-in round() function.
 
 		Rounds a number to a given precision in decimal digits (default
-		0 digits). This returns an int when called with one argument,
+		0 digits). This returns an int when ndigits is omitted or is None,
 		otherwise the same type as the number.
 
 		Values are rounded to the closest multiple of 10 to the power minus
--- a/Lib/fontTools/misc/py23.py
+++ b/Lib/fontTools/misc/py23.py
@@ -336,8 +336,19 @@ if PY3:
 
 			return float(d)
 
-	# in Python 3, 'round3' is an alias to the built-in 'round'
-	round = round3 = round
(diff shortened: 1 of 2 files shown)


adrientetar:
Is it common to call round with None?

anthrotype:
well, fontMath does.
And with our own pure-python `round3` (used on Python 2 only) we do allow to pass ndigits=None, so it would be an inconsistency if on Python2 we allow round3 having a second None argument, whereas on Python3 we don't (or we do, but only if the first arg is a float..)

Lib/fontTools/misc/py23.py
```diff
@@ -336,8 +336,19 @@ def round2(number, ndigits=None):
 
 			return float(d)
 
-	# in Python 3, 'round3' is an alias to the built-in 'round'
-	round = round3 = round
+	if sys.version_info[:2] >= (3, 6):
+		# in Python 3.6, 'round3' is an alias to the built-in 'round'
+		round = round3 = round
+	else:
```
adrientetar: Looking at https://bugs.python.org/issue27936 it seems the issue only affects 3.5 (right?) so you could do `if sys.version_info[:2] == (3, 5)`

PR #760, opened 2016-12-05
Snippet for namelist generation

RimeOCRLIB:
#! /usr/bin/env python

#get font namelist

from fontTools.ttLib.tables import _c_m_a_p

def writeText(text,path):
    with open(path, "w") as text_file:
        text_file.write(unicode(text))

def main():
    font=TTFont("C:\path\yourFont.ttf")
    cmapTable = font["cmap"]
    cmap=cmapTable.getcmap(0,3)
    text=""
    for code  in cmap.cmap:
        #print (hex(code)+" "+cmap.cmap[code])
        text+=hex(code)+" "+cmap.cmap[code]+"\n"
        
    path="C:\path\yourFont.nam"
    writeText(text,path)
    print ("done")

if __name__ == "__main__":
    main()


anthrotype:
@RimeOCRLIB sorry but copy/pasting your code as text in the pull request message is not the way go, I'm afraid. Please have a read at git and github documentation before contributing again.

Issue #696, opened 2016-10-08
Remove top-level sstruct and xmlWriter?


anthrotype:
Now that `extra_path=FontTools` is gone from our setup.py (see https://github.com/fonttools/fonttools/pull/695), I would like to use this opportunity to also drop support for the stand-alone `sstruct` and `xmlWriter` modules that we still install to site-packages alongside the main `fontTools` package.

These were at some point moved by Behdad to their current location, inside fontTools.misc package, but they were subsequently restored to their original position (as aliases) after some users started complaining about sudden ImportErrors.

I think we can now safely drop those, as compatibility with the "sourceforge" fonttools is no longer so important.

Anybody who still relies on importing from those instead of the fontTools.misc.\* counterparts, could keep using version 3.1.2. But future versions would no longer have them.

WDYT?

/cc @justvanrossum @LettError @typemytype @typesupply @benkiel

justvanrossum:
That would have been fine if DeprecationWarnings had been issued for a while. I suggest to do that anyway, and wait a release or two with removing them.

anthrotype:
Ok, let's do that.

justvanrossum:
I think Adobe used these module as well, I remember @frankrolf mentioning it back then.

anthrotype:
It should be easy to replace this:

```python
from xmlWriter import XMLWriter
```
with this:
```python
from fontTools.misc.xmlWriter import XMLWriter
```

The reason I want remove them is that, without the containing "FontTools" folder which was previously created by the deprecated `extra_path` setup keyword, now we are kind of polluting the  site-packages with extra modules which are outside of the `fontTools` namespace.

It is conceivable for another unrelated project to use the names `xmlWriter` or even `sstruct`. With `extra_path`, one would mask another in the sys.path; without the `extra_path`, one risks overwriting the other...

Anyway, I'll just add the deprecation warning for now.

typemytype:
I dont understand what the issue is... Did you encounter other packages using a different `sstruct` and `xmlWriter`?
And if so, it would be nice to send PR to all of these: https://github.com/search?p=1&q=%22from+xmlWriter+import+XMLWriter%22+language%3APython&ref=searchresults&type=Code&utf8=✓ <https://github.com/search?p=1&q=%22from+xmlWriter+import+XMLWriter%22+language:Python&ref=searchresults&type=Code&utf8=%E2%9C%93>

(github seems not to be able to search for an exact match...)



gr Frederik
www.typemytype.com

[...]

PR #761, opened 2016-12-07
Remove py_modules sstruct and xmlWriter from setup.py for issue#696

pnemade:
With the new 3.3.0 release, I found that when you do "python setup.py build", you will see 

```
$ python2 setup.py build 
running build
running build_py
file Lib/sstruct.py (for module sstruct) not found
file Lib/xmlWriter.py (for module xmlWriter) not found
.....
```

This can be fixed by removing following line from setup.py
``` 
 py_modules=['sstruct', 'xmlWriter'],
```

I think this is also related with #696

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,6 @@ setup(
 	long_description=long_description,
 	package_dir={'': 'Lib'},
 	packages=find_packages("Lib"),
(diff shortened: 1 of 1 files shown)


anthrotype:
Ouch! Thank you

PR #763, opened 2016-12-07
[varLib] Build weight axis according to spec

jamesgk:
With this the weight axis values are on a range from 100-900, matching the OS/2 weight class values.

I'm pretty sure the math is right here, but someone should double check. Also not 100% sure about the weight-class-guessing method -- is this hardcoded dictionary OK? (And if so, can we put it somewhere easily imported into other libraries, since we use something similar in glyphsLib and nototools.)

Should fix https://github.com/googlei18n/fontmake/issues/167. cc @brawer

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -70,7 +70,8 @@ def _add_fvar(font, axes, instances, axis_map):
 	font['fvar'] = fvar = newTable('fvar')
 	nameTable = font['name']
 
-	weight_re = re.compile('(%s)' % '|'.join(CLASS_VALS.keys()))
+	weight_names = sorted(CLASS_VALS.keys(), key=lambda s: -len(s))
+	weight_re = re.compile('(%s)' % '|'.join(weight_names))
 	wght_class_axis = {'wght': (100, 400, 900)}
 	wght_design_axis = None
 	avar_wght_mappings = {}
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -31,8 +31,21 @@ from fontTools.ttLib.tables import otBase as otBase
 from fontTools.varLib import designspace, models, builder
 from fontTools.varLib.merger import merge_tables, Merger
 import warnings
+import re
 import os.path
 from argparse import ArgumentParser
+
+CLASS_VALS = {
+	'Thin': 100,
+	'ExtraLight': 200,
+	'Light': 300,
+	'Regular': 400,
+	'Medium': 500,
+	'SemiBold': 600,
+	'Bold': 700,
+	'ExtraBold': 800,
+	'Black': 900,
+}
 
 #
 # Creation routines
@@ -57,10 +70,19 @@ def _add_fvar(font, axes, instances, axis_map):
 	font['fvar'] = fvar = newTable('fvar')
 	nameTable = font['name']
 
+	weight_re = re.compile('(%s)' % '|'.join(CLASS_VALS.keys()))
+	wght_class_axis = {'wght': (100, 400, 900)}
+	wght_design_axis = None
+	avar_wght_mappings = {}
+
 	for iden in sorted(axes.keys(), key=lambda k: axis_map[k][0]):
 		axis = Axis()
 		axis.axisTag = Tag(axis_map[iden][0])
-		axis.minValue, axis.defaultValue, axis.maxValue = axes[iden]
+		axis_vals = axes[iden]
+		if axis.axisTag == 'wght':
+			wght_design_axis = {'wght': axis_vals}
+			axis_vals = wght_class_axis['wght']
+		axis.minValue, axis.defaultValue, axis.maxValue = axis_vals
 		axisName = tounicode(axis_map[iden][1])
 		axis.axisNameID = nameTable.addName(axisName)
 		fvar.axes.append(axis)
@@ -69,14 +91,36 @@ def _add_fvar(font, axes, instances, axis_map):
 		coordinates = instance['location']
 		name = tounicode(instance['stylename'])
 		psname = instance.get('postscriptfontname')
+		weight_match = weight_re.search(name)
+		weight = weight_match.group(0) if weight_match else 'Regular'
 
 		inst = NamedInstance()
 		inst.subfamilyNameID = nameTable.addName(name)
 		if psname is not None:
 			psname = tounicode(psname)
 			inst.postscriptNameID = nameTable.addName(psname)
-		inst.coordinates = {axis_map[k][0]:v for k,v in coordinates.items()}
+		inst.coordinates = {}
+		for iden, val in coordinates.items():
+			tag = axis_map[iden][0]
+			if tag == 'wght':
+				design_loc = models.normalizeLocation(
+					{'wght': val}, wght_design_axis)
+				val = CLASS_VALS[weight]
+				class_loc = models.normalizeLocation(
+					{'wght': val}, wght_class_axis)
+				avar_wght_mappings[class_loc['wght']] = design_loc['wght']
+			inst.coordinates[tag] = val
 		fvar.instances.append(inst)
+
+	if avar_wght_mappings:
+		assert 'avar' not in font
+		for k in (-1.0, 0.0, 1.0):
+			if k not in avar_wght_mappings:
+				avar_wght_mappings[k] = k
+			assert avar_wght_mappings[k] == k
+		font['avar'] = avar = newTable('avar')
+		avar.segments = {
+			'wght': avar_wght_mappings, 'wdth': {-1: -1, 0: 0, 1: 1}}
 
 	return fvar
 


Lib/fontTools/varLib/__init__.py
```diff
@@ -31,9 +31,22 @@
 from fontTools.varLib import designspace, models, builder
 from fontTools.varLib.merger import merge_tables, Merger
 import warnings
+import re
 import os.path
 from argparse import ArgumentParser
 
+CLASS_VALS = {
+	'Thin': 100,
```
brawer: Could varLib perhaps take the weight and width values from the [OS/2 table](https://www.microsoft.com/typography/otspec/os2.htm) in the merged fonts? This would be less fragile than parsing camel-cased English names.

brawer: cc @LettError for designspace question

jamesgk: Ah! All that axes/map stuff is relatively new in mutatormath, thanks for bringing it to my attention. I think we can definitely use it here. As for the custom axis, that's this old issue: https://github.com/googlei18n/fontmake/issues/144.

twardoch:
Note that according to the spec, `fvar.wght` values may go from 1 to 1000, not 100 to 900. 

*Values must be in the range 1 to 1000. Values can be interpreted in direct comparison to values for usWeightClass (in the 'OS/2' table), or the CSS font-weight property — see the note below.*
*In a variable font that implements 'wght' variations, the value in the usWeightClass field of the 'OS/2' table must match the default 'wght' value specified in the 'fvar' table.*
— https://www.microsoft.com/typography/otspec/fvar.htm

twardoch:
+ @schriftgestalt 
Currently, Glyphs specifies masters like this: 
```
(
        {
        custom = Text;
        customValue = 12;
        weight = Light;
        weightValue = 28;
        width = Condensed;
        widthValue = 70;
    },
        {
        custom = Text;
        customValue = 12;
        weight = Regular;
        weightValue = 94;
        width = Condensed;
        widthValue = 70;
    },
        {
        custom = Text;
        customValue = 12;
        weight = Bold;
        weightValue = 194;
        width = Condensed;
        widthValue = 70;
    }
)
```

while it specifies instances like this: 

```
(
        {
        interpolationCustom = 12;
        interpolationWeight = 28;
        interpolationWidth = 70;
        name = "Light Condensed Text";
        weightClass = Light;
        widthClass = Condensed;
    },
        {
        interpolationCustom = 12;
        interpolationWeight = 94;
        interpolationWidth = 70;
        name = "Condensed Text";
        widthClass = Condensed;
    },
        {
        interpolationCustom = 12;
        interpolationWeight = 194;
        interpolationWidth = 70;
        name = "Bold Condensed Text";
        weightClass = Bold;
        widthClass = Condensed;
    }
)
```

**If** we have instances corresponding to masters, then we can get an instance, get the mapping from `weightClass` (`Bold` = `700`, `400` if it's omitted) to `interpolationWeight` and then match the corresponding master `weightValue`. In this example: 

* `weightClass` = `Light` (300) -> Master `weightValue` = `28`
* `weightClass` = `Regular` (400) -> Master `weightValue` = `94`
* `weightClass` = `Bold` (700) -> Master `weightValue` = `194`

The Glyphs Master `weightValue` entries (`28`, `94`, `194`) should actually be converted into the **normalized coordinates** and should never be used for **user coordinates**, so they should become `-1`, `0`, `1`. Currently, the only sensible way to infer appropriate **user coordinates** is to use the `weightClass` value in instances (if there are instances matching the masters). 

This would be sufficient to build an `fvar.wght` axis that is scaled from `300` through `400` all the way to `700`, and build an `avar` table maps the weights. If there are more predefined instances, with more `weightClass` settings, then those should be used to add breakpoints in the `avar` table. Especially with `fvar.wght` axis, we'll pretty much always need an `avar` table, because the user coordinate space for `fvar.wght` (masters and predefined instances) will most likely go [`100`, `200`, `300`, ..., `900`], in equal "100" steps (controlled by `weightClass`), while the mapping to Master `weightValue` would be different. 

An alternative logic could be that you only track the Glyphs masters definition. There, you have a parameter `weight` which can take one of five entries: [`Light`, `SemiLight`, `Regular`, `SemiBold`, `Bold`], and provides mapping to the `weightValue` settings. But note that very often, the Master with `master.weight = Light` has a corresponding instance with `instance.weightClass = Thin` (and not `Light`). 

Overall, in the current Glyphs version, I’d always mandate that all masters have corresponding instances, because it’s only from them where we can reliably pick the user coordinates for `wght`. 

As a fallback, if there are no corresponding instances, I’d assume that the `master.weight` values map to `fvar.wght` user coordinates so: 
```
{ 
  Light = 100, 
  SemiLight = 200, 
  Regular = 400, 
  SemiBold = 700, 
  Bold = 900
} 
```

Best,
Adam

schriftgestalt:
> An alternative logic could be that you only track the Glyphs masters definition.

That is not a good idea. Those values do not represent anything meaning full.

> If we have instances corresponding to masters, then we can get an instance, get the mapping from weightClass

One could also try to reverse interpolate the weight class for the masters even if the instances do not exactly match the masters.

> The Glyphs Master weightValue entries (28, 94, 194) should actually be converted into the normalized coordinates and should never be used for user coordinates,

Why? I understand that fvar axis can have any values?

twardoch:
According to the spec, `fvar.wght` values may go from 1 to 1000,  but the spec suggests that for `wght`, they “can be interpreted in direct comparison to usWeightClass / font-weight”. While this is not a strong requirement, I’d say it definitely makes sense **especially** if you also plan to ship (or used to ship) instance-based fonts that used usWeightClass. 

Best practice for `wght` is that the user coordinates range from 1 to 1000, or more practically, from 100 to 900+, that the central master has the user coordinate 400, in the predefined instances the Regular has the `fvar.wght` user coordinate 400 and the Bold has the `fvar.wght` user coordinate 700, and other predefined instances use `fvar.wght` user coordinates that correspond to usWeightClass or CSS font-weight. 

Quote: 
*Values must be in the range 1 to 1000. Values can be interpreted in direct comparison to values for usWeightClass (in the 'OS/2' table), or the CSS font-weight property — see the note below.*
*In a variable font that implements 'wght' variations, the value in the usWeightClass field of the 'OS/2' table must match the default 'wght' value specified in the 'fvar' table.*

— https://www.microsoft.com/typography/otspec/fvar.htm

twardoch:
In other words: 

1. I may have three Glyphs masters, with `weightValue` = [`Light` = `185`, `Regular` = `382`, `Bold` = `926`]
2. Those coordinates should be best converted into normalized `wght` coordinates [`-1`, `0`, `1`]
3. Those coordinates should also be converted into user `fvar.wght` coordinates [`100`, `400`, `900`]. 
4. Further, Glyphs instances with various `interpolationWeight` values should be best converted into predefined `fvar.wght` instances with user coordinates ideally matching the usWeightClasses [`100`, `200`, `300`, `400`, `500`, `600`, `700`, `800`, `900`], and possibly some more in-between. 
5. Finally, there should be an `avar` table that maps user coordinates into normalized coordinates using all the breakpoints from the predefined instances. 

In effect, I need two conversions: 
1. From “spec-conventional” `fvar.wght` **equidistant** user scale [`100`, `200`, `300`, `400`, `500`, `600`, `700`, `800`, `900`] into Glyphs `weightValue` / `interpolationWeight` numbers, which might be e.g. [**`185`**, `241`, `311`, **`382`**, `463`, `566`, `685`, `809`, **`926`**] (Bold means master Glyphs `weightValue` coordinates)
2. From the Glyphs scale into the `fvar.wght` normalized scale [`-1`...`0`...`1`]. The `avar` table needs to be a cumulation of those two conversions!

twardoch:
There are many reasons why the `fvar.wght` user coordinates are best if they are spec-conventional, i.e. correspond to the equidistant usWeightClass / font-weight scale: 
1. Variation-enabled web browsers will use the `fvar.wght` user scale directly, i.e. if I specify `font-weight: light;` in CSS, the browser will request the variation instance with `wght=300`. 
2. Variable fonts that are a replacement of previously shipping static fonts should have predefined instances with `fvar.wght` using the conventional scale where a variable predefined instance with `wght=300` is a replacement for a static instance with `usWeightClass=300`. 
3. When fonts are changed in a project, apps could re-use `wght` values specified for one font with another font. This would be especially important for things like font fallback. 

The other predefined axes in `fvar`, namely `wdth` and `opsz` have “reality-based” user coordinate scales. For `wdth`, it’s “percentage of normal width” with a default to 100, which is pretty much the same as in Glyphs. For `opsz`, it’s optical size expressed in typographic points, which is what you’d usually use in your design as well. 

But `wght` needs special treatment because in Glyphs (or any other type design app), your `interpolationWeight` or `weightValue` parameters actually control the interpolation, so they’re “reality-based” is some way (especially if you have more than two masters), but the conventional `wght` user coordinate scale is equidistant.

jamesgk:
So I think we're all in agreement that the fvar weight axis should go 1-1000, and that we need an avar mapping this axis to the design axis.

@brawer to confirm, is your thinking that all the weight interpolation values in the designspace exported from glyphsLib should have values from 1-1000, and the axis element contains a map from that to the original interpolation values?

My concern is that it may not always be possible to reverse-interpolate the weight class of a master when it doesn't have a corresponding instance. I'm also not sure if MutatorMath actually uses this axis element for interpolation yet (but I have't tried).

jamesgk:
Alternatively we can map original design values to weight class values, but I think that's semantically incorrect and kind of a hacky abuse of the axis tag. Probably the easiest implementation though.

jamesgk:
OK, I think the issue of "reverse-interpolating" weight classes isn't actually an issue because you can just choose arbitrary weight classes for masters and make sure they map to the proper interpolation value. So ignore my last two posts.

However: I have another concern. Take this example:
https://github.com/google/roboto/blob/everything-in-source-interp/src/v2/RobotoUpright.designspace.
In this designspace, we have a Bold instance with interpolation parameters (173,173) and a Condensed Bold instance with parameters (175,175). Both should have weight class 700. If we use the user-facing weight classes as interpolation parameters in the designspace, there's no way to differentiate these two Bolds and the intended interpolation can't take place.

davelab6:
I'm curious, what is the status of this? :)

brawer:
See https://github.com/googlei18n/fontmake/issues/167 for current state. (tl;dr: designspace spec has been updated/clarified; glyphsLib emits new designspace format; varLib still needs to be updated).

PR #764, opened 2016-12-10
[feaLib][test] No need to create temporary file

khaledhosny:
Loading features from file objects have been supported for a while
already.

### DIFF
--- a/Lib/fontTools/feaLib/builder_test.py
+++ b/Lib/fontTools/feaLib/builder_test.py
@@ -1,7 +1,8 @@
 from __future__ import print_function, division, absolute_import
 from __future__ import unicode_literals
 from fontTools.misc.py23 import *
-from fontTools.feaLib.builder import Builder, addOpenTypeFeatures
+from fontTools.feaLib.builder import Builder, addOpenTypeFeatures, \
+        addOpenTypeFeaturesFromString
 from fontTools.feaLib.builder import LigatureSubstBuilder
 from fontTools.feaLib.error import FeatureLibError
 from fontTools.ttLib import TTFont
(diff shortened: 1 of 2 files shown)


Lib/fontTools/feaLib/builder_test.py
```diff
@@ -117,11 +118,8 @@ def expect_ttx(self, font, expected_ttx):
             self.fail("TTX output is different from expected")
 
     def build(self, featureFile):
-        path = self.temp_path(suffix=".fea")
-        with open(path, "w", encoding="utf-8") as outfile:
-            outfile.write(featureFile)
         font = makeTTFont()
-        addOpenTypeFeatures(font, path)
+        addOpenTypeFeaturesFromString(font, featureFile)
```
anthrotype: Perhaps it might be worth having a separate test for parsing/building from a filesystem path input?

anthrotype:
As you can see, the STAT tests are failing since @behdad added that extra field to the table header.
It's sill not clear to me whether we should revert the change or adjust the expected test results...
/cc @brawer @moyogo

Lib/fontTools/feaLib/parser_test.py
```diff
@@ -1263,22 +1260,9 @@ def test_empty_statement_ignored(self):
...
-
     def parse(self, text):
-        if not self.tempdir:
-            self.tempdir = tempfile.mkdtemp()
-        self.num_tempfiles += 1
-        path = os.path.join(self.tempdir, "tmp%d.fea" % self.num_tempfiles)
-        with open(path, "w", encoding="utf-8") as outfile:
-            outfile.write(text)
-        return Parser(path).parse()
+        featurefile = UnicodeIO(tounicode(text))
```
anthrotype: We are importing unicode_literals from __future__, so I think tounicode() should not be needed here.

Lib/fontTools/feaLib/builder_test.py
khaledhosny: If I’m reading the code correctly, there is already one few lines below.

anthrotype: Oh ok! I was reading from my phone :)

Lib/fontTools/feaLib/parser_test.py
khaledhosny: Right.

Issue #767, opened 2016-12-13
versions


typemytype:
DrawBot and RoboFont users complain about a pop up prompting to install git when launching the latests version of the apps.

I suspect `_version.py` is requesting it, this feels a bit over engineered  :)

justvanrossum:
The prompt is actually about installing the OSX dev tools. I'll hope to reproduce it tomorrow, and will try to determine whether it is indeed caused by _version.py.

anthrotype:
Hold on. You have to install fonttools in a regular way, not just editable or using pth before you package up your py2app, else the _version.py will be the dynamic one that is meant for git source control, instead of the static one that's generated upon install.

anthrotype:
You first need to do `pip install -v .` from the fonttools source directory. Then you can run your `setup.py py2app` so that py2app will copy the fonttools package that was installed in your python environment, instead of grabbing it straight from the git repository in your PYTHONPATH.

Another option is to do `python setup.py build` and have py2app copy the generated `build/lib/fontTools` package.

Please let me know if that works for you.

justvanrossum:
And we're back to square one with the version discussion :(

anthrotype:
please do not reopen this discussion now.
The correct way to install a python package is... to install by having the installer generate the package metadata.
As I said, It is easy for other application developers bundle fonttools.
They could even simply unzip the source distribution, or the `.whl` package (you can rename it `.zip` and extract the content).

I just opened up the Robofont.app package and, indeed, the `Contents/Resources/lib/python2.7/fontTools/_version.py` is the wrong one.

justvanrossum:
I understand you don't like to reopen the discussion, but to me this is another proof that the version system is brittle. It tries to do too much too smartly. I don't like how it enforces one way of "correct" deployment, or rather, how it weirdly fails when deployment wasn't done "correctly".

anthrotype:
Ok, let me recap.

versioneer computes the package version from the git metadata, so it only works when it is inside a git repository.

When you create a distribution, or run setup.py install, pip install etc., versioneer generates a static _version.py module containing just strings, so git is not required.

py2app works by looking at all the package imported in your python code and copy them inside the .app bundle. If you PYTHONPATH contains the fontTools git repository, py2app copies its content inside your app, but without the .git folder, hence breaking things.

[...]

PR #768, opened 2016-12-14
don't compute git version at runtime; bump version with custom 'release' command

anthrotype:
Following the discussion at #767 and the overall negative feedback about the versioneer.py approach for computing the package version string at import time through git, I decided to reverted those changes and go back to a static version string for the setup.py and fontTools.__init__.py files.

I'm still using the `bumpversion` tool to automate the process, and also created a custom setuptools command (`python setup.py release`) to bootstrap bumpversion and apply the version scheme.

Below is the docstring explaining how this works.

```
        """
	Tag a new release with a single command, using the 'bumpversion' tool
	to update all the version strings in the source code.
	The version scheme conforms to 'SemVer' and PEP 440 specifications.

	Firstly, the pre-release '.devN' suffix is dropped to signal that this is
	a stable release. If '--major' or '--minor' options are passed, the
	the first or second 'semver' digit is also incremented. Major is usually
	for backward-incompatible API changes, while minor is used when adding
	new backward-compatible functionalities. No options imply 'patch' or bug-fix
	release.

	A new header is also added to the changelog file ("NEWS"), containing the
	new version string and the current 'YYYY-MM-DD' date.

	All changes are committed, and an annotated git tag is generated. With the
	--sign option, the tag is GPG-signed with the user's default key.

	Finally, the 'patch' part of the version string is bumped again, and a
	pre-release suffix '.dev0' is appended to mark the opening of a new
	development cycle.

	Links:
	- http://semver.org/
	- https://www.python.org/dev/peps/pep-0440/
	- https://github.com/peritus/bumpversion
	"""
```

I would basically be the only user, but encourage others to have a look and give feedback if they like.

Thank you, and apologies for my obstinacy with this not-very-cool-but-somebody-has-to-do-it stuff.


anthrotype:
(of course, I don't want to be the only user... that's why I tried to make it simple for others to use as well)

justvanrossum:
Thanks so much for your continuous dedication and efforts! I'm very happy with this solution.

anthrotype:
Cool, I'm happy you're happy! I'll use this shortly to tag a 3.3.1 release. Let's see if actually works.. ;)

anthrotype:
yes, it worked!

These are the commands I run:

```shell
python setup.py release --sign
git push origin $(git describe --abbrev=0 --tags)
```

I didn't have the --sign option on by default, because it depends on having git properly configured for using a GPG signing key.
E.g. see https://help.github.com/articles/telling-git-about-your-gpg-key/

The second command simply pushes the current tag to the `origin` remote, which is fonttools/fonttools on my machine.

anthrotype:
Actually the commands above only push the tag, but not the commits..
There should be an extra git push in between: 

```
python setup.py release --sign
git push origin master
git push origin $(git describe --abbrev=0 --tags)
```

I wish I could do `git push --follow-tags`, but that would also push along all my private tags (which we did for internal deployment).

PR #774, opened 2016-12-19
Map AGL glyph names to Unicode

brawer:
Implements the mapping algorithm from the [Adobe Glyph List specification](
https://github.com/adobe-type-tools/agl-specification#2-the-mapping).

Currently, the implementation only recognizes glyph names from the
Adobe Glyph List for New Fonts (AGLFN), not the legacy AGL which is a
superset. If there is interest, it would be easy to support the legacy AGL
as well, preferrably in a follow-up change.

https://github.com/googlei18n/glyphsLib/issues/88#issuecomment-267408215

### DIFF
--- a/Lib/fontTools/agl.py
+++ b/Lib/fontTools/agl.py
@@ -1,8 +1,12 @@
+# -*- coding: utf-8 -*-
 # The table below is taken from
 # http://www.adobe.com/devnet/opentype/archives/aglfn.txt
 
-from __future__ import print_function, division, absolute_import
+from __future__ import (print_function, division, absolute_import,
+                        unicode_literals)
 from fontTools.misc.py23 import *
+import re
+
 
 _aglText = """\
 # -----------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Lib/fontTools/agl.py
```diff
@@ -1,8 +1,12 @@
+# -*- coding: utf-8 -*-
 # The table below is taken from
 # http://www.adobe.com/devnet/opentype/archives/aglfn.txt
 
-from __future__ import print_function, division, absolute_import
+from __future__ import (print_function, division, absolute_import,
+                        unicode_literals)
```
justvanrossum: Is `unicode_literals` really necessary here? It breaks some of my doctests under Python 2.7. With this change, `TTFont().getGlyphNames()` can now return a mix of `str` and `unicode`.

anthrotype: yes, it's not good to mix the two. 
I think it's best to keep the `unicode_literals`, but make sure the native `str` type is used in the `AGL2UV` and `UV2AGL` dictionaries.

See d8c4058

justvanrossum: Thanks!

Issue #772, opened 2016-12-19
[feaLib] Generate Adobe feature files from feaLib AST


brawer:
Emit Adobe feature files from the feaLib AST (Abstract Syntax Tree). @mhosken would like to work on it.

anthrotype:
That would be great, thanks!

adrientetar:
Does the ft2fea tool in #479 use the feaLib AST?

anthrotype:
do we keep this open now that #776 has been merged?

@mhosken in his https://github.com/fonttools/fonttools/pull/776#issue-196626145 mentioned two features which are not implemented yet:

> Ranges are expanded in the parser and not contracted in fea generation. This might be worked around by keeping a side copy of the original glyph list specification including ranges.

> nameid attributes are often defaulted and expanded in the parser. These are not defaulted out in the fea generation.

anthrotype:
Ok, both of the above mentioned issues are fixed now, thanks to Martin. 🎉 
Feel free to open new specific issues.

PR #776, opened 2016-12-20
Add fea generation support

mhosken:
This PR enables fonttools to generate fea syntax from a fea ast, as per issue #772. I notice that #479 is not ast based and simply generates text as it goes.

One calls asFea() on the output from a Parser().parse() and is returned a string containing the generated fea text.

At this point in the PR's history not all the tests are giving identical results. This is due to the following features not yet having been implemented:

- Ranges are expanded in the parser and not contracted in fea generation. This might be worked around by keeping a side copy of the original glyph list specification including ranges.
- nameid attributes are often defaulted and expanded in the parser. These are not defaulted out in the fea generation.

In order to get the tests to pass, many of them have been adjusted in layout and some keywords regularised. These need to be reviewed to ensure that the tests are still testing what the test was originally designed to test. A special comment can also be used to acknowledge different output from fea2fea for a test. E.g. feaLib/testdata/size.fea

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -383,7 +383,7 @@ class AttachStatement(Statement):
         return "Attach {} {};".format(self.glyphs.asFea(), " ".join([str(c) for c in self.contourPoints]));
 
 
-class ChainContextStatement(Statement) :
+class ChainContextPosStatement(Statement):
     def __init__(self, location, prefix, glyphs, suffix, lookups):
         Statement.__init__(self, location)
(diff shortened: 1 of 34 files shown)


brawer:
Thank you, looks good to me! There’s a minor typo which I’ll fix myself.

anthrotype:
wow this is huge! thanks a lot Martin! 💯 
I'm look forward to use this.

twardoch:
cool!

anthrotype:
I was wondering if instead of, or in addition to, `asFea()` methods, we could have `__str__` methods for each `ast` object, so one could do `str(something)` or `print(something)`.

Issue #689, opened 2016-10-04
support MVAR table


robmck-ms:
It would be helpful if fonttools supported the MVAR table from the OpenType 1.8 spec:
https://www.microsoft.com/typography/otspec/MVAR.htm

anthrotype:
Behdad pushed b4d53811e904bd8315639b364a9d9747bb08350c yesterday.
It would be nice to have some unittest before tagging a release with MVAR support.

behdad:
Here's the MVAR from a Selawik test Rob sent me before (I had to fix the offset length), in case someone wants to make it into a test:

```
  <MVAR>
    <Version value="0x00010000"/>
    <!-- AxisCount=1 -->
    <ValueRecordSize value="8"/>
    <!-- ValueRecordCount=7 -->
    <VarStore Format="1">
      <Format value="1"/>
      <VarRegionList>
        <!-- RegionAxisCount=1 -->
        <!-- RegionCount=1 -->
        <Region index="0">
          <VarRegionAxis index="0">
            <StartCoord value="0.0"/>
            <PeakCoord value="1.0"/>
            <EndCoord value="1.0"/>
          </VarRegionAxis>
        </Region>
      </VarRegionList>
      <!-- VarDataCount=1 -->
      <VarData index="0">
        <!-- ItemCount=4 -->
        <NumShorts value="1"/>
        <!-- VarRegionCount=1 -->
        <VarRegionIndex index="0" value="0"/>
        <Item index="0" value="[-200]"/>
        <Item index="1" value="[-50]"/>
        <Item index="2" value="[100]"/>
        <Item index="3" value="[200]"/>
      </VarData>
    </VarStore>
    <ValueRecord index="0">
      <ValueTag value="hasc"/>
      <VarIdx value="3"/>
    </ValueRecord>
    <ValueRecord index="1">
      <ValueTag value="hcla"/>
      <VarIdx value="3"/>
    </ValueRecord>
    <ValueRecord index="2">
      <ValueTag value="hcld"/>
      <VarIdx value="3"/>
    </ValueRecord>
    <ValueRecord index="3">
      <ValueTag value="hdsc"/>
      <VarIdx value="0"/>
    </ValueRecord>
    <ValueRecord index="4">
      <ValueTag value="hlgp"/>
      <VarIdx value="2"/>
    </ValueRecord>
    <ValueRecord index="5">
      <ValueTag value="sbyo"/>
      <VarIdx value="1"/>
    </ValueRecord>
[...]

PR #781, opened 2016-12-20
Add MVAR tests

moyogo:
Based on a Selawik test MVAR table from https://github.com/fonttools/fonttools/issues/689#issuecomment-268146882.

### DIFF
--- a/Lib/fontTools/ttLib/tables/M_V_A_R_test.py
+++ b/Lib/fontTools/ttLib/tables/M_V_A_R_test.py
@@ -1,0 +1,123 @@
+from __future__ import print_function, division, absolute_import
+from fontTools.misc.py23 import *
+from fontTools.misc.testTools import FakeFont, getXML, parseXML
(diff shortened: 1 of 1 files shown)


Lib/fontTools/ttLib/tables/M_V_A_R_test.py
```diff
@@ -0,0 +1,135 @@
...
+        mvar = newTable('MVAR')
+        font = TTFont()
+        self.maxDiff = None
+        for name, attrs, content in parseXML(MVAR_XML):
+            mvar.fromXML(name, attrs, content, font=font)
+        # Ignore AxisCount
+        # see https://github.com/fonttools/fonttools/issues/689
+        data = MVAR_DATA
+        data = data[:5] + b'\x00' + data[6:]
+        self.assertEqual(hexStr(mvar.compile(font)), hexStr(data))
```
moyogo: Thanks!

PR #784, opened 2016-12-21
[testTools] have getXML() return a list of lines

anthrotype:
When writing unit tests for XML data, it is more convenient to compare list of lines, instead of a single string without newlines.

This helps identifying which lines in the diff printed on the console don't match the expected result.

So this makes `getXML()` returns a list of lines.

Similarly, the complementary function `parseXML()` now also accepts a list of
strings, as well as a single string, to allow reusing the same data for to/fromXML round-trip tests.

/cc @brawer @moyogo


anthrotype:
fixed, sorry about that.

PR #790, opened 2016-12-27
In cmap decompile skip gid0

behdad:


Lib/fontTools/ttLib/tables/_c_m_a_p.py
```diff
@@ -15,6 +15,19 @@
 log = logging.getLogger(__name__)
 
 
+def _make_map(font, chars, gids):
+	assert len(chars) == len(gids)
+	cmap = {}
+	lenCmap = len(gids)
+	glyphOrder = font.getGlyphOrder()
+	try:
+		names = list(map(operator.getitem, [glyphOrder]*lenCmap, gids ))
```
justvanrossum: Why not
 `names = [glyphOrder[gid] for gid in gids]`
?

Lib/fontTools/ttLib/tables/_c_m_a_p.py
```diff
@@ -15,6 +15,19 @@
...
+	except IndexError:
+		getGlyphName = font.getGlyphName
+		names = list(map(getGlyphName, gids ))
```
justvanrossum: `names = [font.getGlyphName(gid) for gid in gids]`
?

Lib/fontTools/ttLib/tables/_c_m_a_p.py
```diff
@@ -15,6 +15,19 @@
...
+	list(map(operator.setitem, [cmap]*lenCmap, chars, names))
```
justvanrossum: Building a list and just discarding is wasteful. Since nothing happens with `cmap` in the meantime, the following does the same, but (to me) more readable, and def. more Pythonic:
`cmap = {char: name for char, name in zip(chars, names)}`

Lib/fontTools/ttLib/tables/_c_m_a_p.py
justvanrossum: Ah nevermind, I'm stupid, this has been rewritten in a subsequent commit.

behdad:
Yeah, let's not discuss who wrote those lines ;).

Thanks.

Issue #766, opened 2016-12-13
different ClassDef values between python 2 and 3


anthrotype:
It turns out that fonts built with python 2 and 3 may end up having different GPOS tables, because of the non-deterministic way `ClassDef` class definition tables are encoded in feaLib/otlLib.

The following patch:
https://github.com/fonttools/fonttools/compare/issue-766?expand=1

.. produces different test outputs, in a seemingly random fashion:
https://travis-ci.org/fonttools/fonttools/jobs/183647027

I think the problem lies in this line, where a list of sets of (glyph names) strings is sorted by the length of the sets:
https://github.com/fonttools/fonttools/blob/f81e1411b3fb1ca402065fbd9a3c53907d205b58/Lib/fontTools/otlLib/builder.py#L622

If two sets have the same length the order of the set list is not guaranteed to be the same all the time, as they might be hashed differently by different implementations, sometimes even by the same one at different times.

We need to find a better way to sort the ClassDef.

/cc @brawer

behdad:
Fixed, but didn't test.

anthrotype:
I spoke too fast in https://github.com/fonttools/fonttools/commit/7d1ddb237ec2f4c6d4982ff076fd15c1d47771f2#commitcomment-20329688

The test for ClassDef sets with same length is still failing in a random fashion:
https://travis-ci.org/fonttools/fonttools/jobs/187721425
https://ci.appveyor.com/project/fonttools/fonttools/build/1.0.329/job/u2vg1dn24w2jwvt7

Shall I mark it as `xfail` until we find a better fix?

anthrotype:
I tried to restart the Travis build (which was apparently passing), and now is failing again, and on the same commit:

https://travis-ci.org/fonttools/fonttools/builds/187819971

So we need to find another sorting method for ClassDef sets.

justvanrossum:
I don't completely follow everything here, but it seems that in the ClassDefBuilder canAdd() and add() methods the glyphs collection is converted to a tuple without doing any sorting. I think that can mean that the order of the glyphs in that tuple can vary across implementations.

anthrotype:
I tried to sort the tuple in 9e76d16, which seems to fix this issue, but now two other test cases fail: `test_FeatureFile_GPOS_2b` and `test_FeatureFile_spec6b_ii`:

https://travis-ci.org/fonttools/fonttools/jobs/187889444#L1216-L1281

I'm not sure how these are related.

[...]

PR #793, opened 2016-12-31
[otlLib] only sort ClassDefBuilder input glyphs if they are unsorted

anthrotype:
This fixes broken tests after 9e76d16, and should finally close https://github.com/fonttools/fonttools/issues/766

/cc @justvanrossum @behdad @brawer

### DIFF
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -589,7 +589,10 @@ class ClassDefBuilder(object):
         self.useClass0_ = useClass0
 
     def canAdd(self, glyphs):
-        glyphs = tuple(glyphs)
+        if isinstance(glyphs, (set, frozenset)):
+            glyphs = tuple(sorted(glyphs))
+        elif not isinstance(glyphs, tuple):
+            glyphs = tuple(glyphs)
         if glyphs in self.classes_:
             return True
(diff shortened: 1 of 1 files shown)


Lib/fontTools/otlLib/builder.py
```diff
@@ -589,7 +589,10 @@ def __init__(self, useClass0):
         self.useClass0_ = useClass0
 
     def canAdd(self, glyphs):
-        glyphs = tuple(glyphs)
+        if isinstance(glyphs, (set, frozenset)):
+            glyphs = tuple(sorted(glyphs))
+        elif not isinstance(glyphs, tuple):
```
behdad: Since tuples are immutable, if x is already a tuple, then "tuple(x)" returns x itself.  So you don't need to specialcase it.  You can simply write:
```
+        if isinstance(glyphs, (set, frozenset)):
+            glyphs = tuple(sorted(glyphs))
+        glyphs = tuple(glyphs)

Comment (anthrotype) [inline Lib/fontTools/otlLib/builder.py]: you're right

```
In [1]: t = (0, 1)

In [2]: id(t)
Out[2]: 4384694152

In [3]: id(tuple(t))
Out[3]: 4384694152
```

Thanks, I'll do that.

Issue #769, opened 2016-12-15
[T2CharStringPen] Rounding coordinates?


mashabow:
Currently, [T2CharStringPen](https://github.com/fonttools/fonttools/blob/fcd697a328b439165b3e9d04904eb73f065b6838/Lib/fontTools/pens/t2CharStringPen.py#L78) always rounds glyph coordinates and casts to int, but it does not seem a good idea. I think we should remove this behavior, or make it optional.

cf. TTGlyphPen (#526)

anthrotype:
Yes. It was like that in the original ufo2fdk package by @typesupply.

I agree we should make the rounding optional, but am not sure if we should also change the current default, as others might be relying on that behaviour.

(Although that's not a big deal if we communicate it in the changelog and bump the version).

anthrotype:
@behdad shall I make rounding optional in T2CharStringPen, or keep rounding on by default for backward compatibility, and make non-rounding optional?

behdad:
> @behdad shall I make rounding optional in T2CharStringPen, or keep rounding on by default for backward compatibility, and make non-rounding optional?

This is relevant in the context of:

https://github.com/googlei18n/ufo2ft/issues/24

While having option to not round is useful in theory, my gut feeling is that a tolerance parameter is more useful.  Integers take 1 or 2 bytes for most common numbers to encode.  A float takes five.  So we should not blindly round-all or not-round-all numbers.  Something like a tolerance allows for a more granular encoding.   tolerance=0 results in no rounding.  tolerance>=.5 means rounding.  Numbers in between result in some rounding and not others.

The reason I didn't implement this before is that I'm not sure if that's a useful, or best, heuristic.  For example, if a control-point vector is large, say, has a length > 100 units, then rounding never makes a visible difference.  Whereas if its length is really short, then rounding might be more visible, with very thin fonts.  That's why I wanted to get Read's test font (which he sent me) and then experiment and find a better heuristic.  The theory I just sketched suggests that rounding or not should also be dependent on the absolute length of the vector or number at hand, not just the distance to closest integer.

Anyway, that's why I have not fixed that other issue yet, because I think some in-depth research is needed to come up with good defaults that don't bloat fonts but improve accuracy for most common cases.

LettError:
It might be useful to consider an option for rounding some categories of points like corners, anchors, components, vertical and horizontal lines, but allow more resolution for points where it matters like angled tangent points, short control points and so on. But I don't know if that is an option in t2.

[...]

PR #804, opened 2017-01-12
[t2CharStringPen] add "roundTolerance" option

anthrotype:
I went for the easy route, and simply added a `roundTolerance` argument to the `T2CharStringPen` as @behdad was discussing in #769.

I personally think that doing more clever things would be a bit beyond the scope of fontTools.
But I'm happy to be convinced otherwise ;)

PTAL, thanks

### DIFF
--- a/Lib/fontTools/pens/t2CharStringPen.py
+++ b/Lib/fontTools/pens/t2CharStringPen.py
@@ -68,16 +68,16 @@ class RelativeCoordinatePen(BasePen):
 
 
 def makeRoundFuncs(tolerance):
-    assert 0 <= tolerance <= 1
+    assert 0 <= tolerance <= 0.5
 
     def _round(number):
         if tolerance == 0:
             return number  # no-op
         rounded = round(number)
-        # return rounded integer if the tolerance is >= 0.5, or if the
-        # absolute difference between the original float and the rounded
-        # integer is within the tolerance
-        if tolerance >= .5 or fabs(rounded - number) <= tolerance:
+        # return rounded integer if the tolerance is 0.5, or if the absolute
+        # difference between the original float and the rounded integer is
+        # within the tolerance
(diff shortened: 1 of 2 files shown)


justvanrossum:
I think the default should remain rounding behavior and that the new behavior should be explicitly turned on.

anthrotype:
Yeah, initially I also thought it'd better to keep the current default, but then got carried away :)
I'll restore the default rounding.

Shall I assert the tolerance is comprised between 0 and 0.5? Values between 0.5 and 1.0 don't make much sense..

behdad:
No, please don't assert tolerance.  I like to be able to say "UPEM * .005" for example and that should work.  Assertions are for other things: for when the input is wrong.  Which is not the case here.

anthrotype:
I'm confused. I'm using the built in round() to convert float to int. The maximum tolerance (which Indefine as the max absolute difference between the rounded and unrounded values) is 0.5, and couldn't be greater.. Am I'm missing something?

Lib/fontTools/pens/t2CharStringPen.py
```diff
@@ -77,7 +76,7 @@ def _round(number):
         # return rounded integer if the tolerance is 0.5, or if the absolute
         # difference between the original float and the rounded integer is
         # within the tolerance
-        if tolerance == .5 or fabs(rounded - number) <= tolerance:
+        if tolerance == .5 or abs(rounded - number) <= tolerance:
```
behdad: Just make this "tolerance >= .5"

behdad:
> the max absolute difference between the rounded and unrounded values) is 0.5

Correct.  Which means rounding is an acceptable solution for any user-specified tolerance >= .5.  A tolerance of 1.5 is NOT by any means an error.

anthrotype:
Ok, I got it. Thanks

Issue #740, opened 2016-11-11
ヒラギノ角ゴシック W4.ttc (Hiragino Sans W4) decompilation fails


jenskutilek:
I tried to convert "ヒラギノ角ゴシック W4.ttc" version 11.0d7e4 (macOS Sierra system font), with this error:

```bash
$ ttx -y 0 hiragino.ttc 
Dumping "hiragino.ttc" to "hiragino.ttx"...
Dumping 'GlyphOrder' table...
Dumping 'head' table...
Dumping 'hhea' table...
Dumping 'maxp' table...
Dumping 'OS/2' table...
Dumping 'name' table...
Dumping 'cmap' table...
Dumping 'post' table...
Dumping 'CFF ' table...
ERROR: Unhandled exception has occurred
Traceback (most recent call last):
  File "/Library/Python/2.7/site-packages/fonttools-3.2.0+2.gdf967e7-py2.7.egg/fontTools/ttx.py", line 398, in main
    process(jobs, options)
  File "/Library/Python/2.7/site-packages/fonttools-3.2.0+2.gdf967e7-py2.7.egg/fontTools/ttx.py", line 372, in process
    action(input, output, options)
  File "/Library/Python/2.7/site-packages/fonttools-3.2.0+2.gdf967e7-py2.7.egg/fontTools/misc/loggingTools.py", line 372, in wrapper
    return func(*args, **kwds)
  File "/Library/Python/2.7/site-packages/fonttools-3.2.0+2.gdf967e7-py2.7.egg/fontTools/ttx.py", line 272, in ttDump
    newlinestr=options.newlinestr)
  File "/Library/Python/2.7/site-packages/fonttools-3.2.0+2.gdf967e7-py2.7.egg/fontTools/ttLib/__init__.py", line 308, in saveXML
    self._tableToXML(tableWriter, tag, progress)
  File "/Library/Python/2.7/site-packages/fonttools-3.2.0+2.gdf967e7-py2.7.egg/fontTools/ttLib/__init__.py", line 344, in _tableToXML
    table.toXML(writer, self, progress)
  File "/Library/Python/2.7/site-packages/fonttools-3.2.0+2.gdf967e7-py2.7.egg/fontTools/ttLib/tables/C_F_F_.py", line 42, in toXML
    self.cff.toXML(writer, progress)
[...]

PR #744, opened 2016-11-26
Fix for #740.

justvanrossum:
This fixes #740.

Commits https://github.com/fonttools/fonttools/commit/3063def35b7ad733bf584b45452f9ef8f6efbab4 and https://github.com/fonttools/fonttools/commit/5b47971 introduced a separate fontDictOperators list for FontDict, only listing those TopDict key/value pairs that are actually used in the FontDict context. It provided a fallback that TTX files containing such "useless" key/value pairs would not be rejected.

However, the code still rejected binary fonts that contained such values, even though it didn't before, and yes, such fonts exist. Also: such fonts are not broken per spec, they just contain some fields that otherwise no one ever looks at, so it's a little harsh to reject them.

This patch removes most of the special FontDict code, and uses everything from TopDict, *except* the order attribute: it sets that to a list of the relevant keys for the FontDict. The effect of this is that "useless" key/value pairs are ignored, not just upon reading XML, but also upon decompilation and compilation of binary fonts. It improves on the previous XML reading behavior in that it no longer silently ignores key typos in the TTX input.

Ideally, we would *output* everything that is actually in the FontDict to TTX, and only ignore the values when compiling, but I didn't find a clean solution for that, so I decided to just fix the issue.

PR #805, opened 2017-01-12
Add CFF2 support.

readroberts:
This adds the changes to cffLib.py and psCharstrings.py to add support for CFF2, and a test. I have not yet removed the global data references, but will do so eventually.


Lib/fontTools/ttLib/tables/C_F_F__2.py
```diff
@@ -0,0 +1,47 @@
...
+	def compile(self, otFont):
+		f = BytesIO()
+		self.cff.compile(f, otFont)
+		return f.getvalue()
+
+	def haveGlyphNames(self):
+		if hasattr(self.cff[self.cff.fontNames[0]], "ROS"):
+			return False  # CID-keyed font
+		else:
+			return True
```
mashabow: In CFF2, `haveGlyphNames()` should always return `False`.

anthrotype:
Thank you Read.
I still believe that having a global `isCFF2` variable modify the behavior of a class is not the best way to go. I think it'd be cleaner if we factor all the common behavior in abstract base classes and make concrete sub-classes that add format-specific behavior.
I was thinking we could maybe have a `cff2` branch here on upstream fonttools/fonttools where we work together on that, and merge to master once ready?

justvanrossum:
That global `isCFF2` variable is indeed not good. I think fonttools is otherwise thread-safe, and we should keep it that way.

readroberts:
Keeping a branch with CFF2 separate until we fix the issues is fine by me. I agree the global variable should go away. Left to myself, though, I would get to it after I get a full workflow going for building CFF2 fonts,  which is another month or two away.

Lib/fontTools/ttLib/tables/C_F_F__2.py
anthrotype: > In CFF2, haveGlyphNames() should always return False.

That's also my understanding of the spec where it says "The CFF2 format does not contain glyph names".

@readroberts any reason why you left it like that?
Also it appears from `git grep "haveGlyphNames"` that this function is only defined here but not used anywhere...

behdad:
We might as well think of the CFF2 support as WIP and merge, and fix as we go...  Branch is nice, but always more work.

readroberts:
I would argue that a  better name for  'haveGlyphNames' would have been "isNotCID-Keyed", and hence 'haveGlyphNames()' should always return True for  CFF2, as CID support is removed in CFF2. For CFF2 tables, glyph names are no longer  stored in the CFF2 table, but but glyph names for the CFF2 glyphs still exist - they may be defined in the 'post' table, or inferred from the 'cmap' table, as is done by the 'ttx' tool. I don't know why this function is in C_F_F_.py - I did not write that code - but without knowing where it is used, I thought it best to keep it in C_F_F__2.py. All that said, I have no reason to be attached to the existence of this function, as it is not called by any of the AFDKO tools. We can always remove it, and see if anyone complains.

anthrotype:
according to git blame, @justvanrossum added that method to the CFF table in 2002:
https://github.com/fonttools/fonttools/commit/a2a75b348da46073cdd1dea50915384eabfff7c9#diff-1da091ac5ef1be92c6660bd402a26729

if it's not used anywhere else, and it does not make sense to always return True for CFF2, I think we could remove it.

Lib/fontTools/cffLib.py
```diff
@@ -1484,37 +2030,17 @@ def getChildren(self, strings):
 		return children
 
 
-class FontDictCompiler(TopDictCompiler):
+class FontDictCompiler(DictCompiler):
```
anthrotype: Here it seems like you reverted the patch #744:

https://github.com/fonttools/fonttools/pull/744/files#diff-5c4ae0351e8d367dbe256f28eba7db01R1487

@justvanrossum added that to fix issue #740.

@readroberts Did you do that intentionally?
Or perhaps you based your CFF2 patch on an outdated master?

justvanrossum: Aargh :(

readroberts:
I based it on an outdated master; this was not intentional. Will fix this when I am back in the office tomorrow.

anthrotype:
> I based it on an outdated master; this was not intentional. Will fix this when I am back in the office tomorrow.

OK, thanks. Can you also move the files "C_F_F__2_test.py" and "testdata/C_F_F__2.{bin,ttx}" to "Tests/ttLib/tables" and "Test/ttLib/tables/data" respectively?
Cf. #811

readroberts:
I will also move the test files.

anthrotype:
I manually merged `readroberts/master` branch into a new `cff2` branch, here on upstream fonttools/fonttools:

https://github.com/fonttools/fonttools/tree/cff2

Let's keep working together on the new branch.

Issue #791, opened 2016-12-28
Import tests from AOTS


behdad:
Similar to the mtiLib test suite, we can import tests from AOTS and roundtrip them to Python to binary & XML and back and compare...

Other test suites that might have interesting stuff: WOFF and WOFF2.

behdad:
I'm working on this.  The test data is a few megabytes.  I'm going to put it in ttLib/testdata/aots.  But I think it's time to move test dat out of Lib/ now?  I don't know if this has any bad implications.  Are they installed? The fact that they have to be listed in manifest confuses me.  If they don't have any extra burden on end users' install, then maybe it's fine the way it is.

behdad:
@anthrotype  ---^

anthrotype:
yes, please, let's move the tests outside the package and no longer install them.
It makes things easier for me.
I can help you with this tomorrow.

behdad:
Thanks Cosimo.  Is it ok if I add them the way we have right now and be done with it, then you do whatever you like?

anthrotype:
yes, thanks!

brawer:
Closing this bug because the work seems done... please re-open if I’m wrong.

PR #811, opened 2017-01-17
move tests to external "Tests/" folder

anthrotype:
as discussed in https://github.com/fonttools/fonttools/issues/791#issuecomment-271667697, we decided it's better to move the test modules to an external folder and no longer install them alongside the fontTools package.
They simply had become too many, and we now plan to add more test data.

Running the test suite works just like before. You have several options:

- `python setup.py test`, which downloads pytest (as a temporary .egg) and runs it with the specified interpreter against the source code in Lib/fontTools (that's also what the `./run-test.sh` script uses; I don't use that, but @behdad may do)

- run the `pytest` command yourself (no arguments will run all tests in "Tests/"); in which case, you need to ensure `fontTools` is in the PYTHONPATH (`pip install -e .` does the job).

- run the `tox` command, which (unless you give a different environment list) will create virtual environments for python 2.7, 3.5 and 3.6, make an sdist and install it in each of them, run the test suite against the installed package, and finally create a HTML coverage report in `hmtlcov` folder

Note that the embedded doctests (">>> ") which are in Lib/fontTools are no longer automatically discovered and run by pytest. This is because we are now running the tests from "Test/" folder instead of "Lib/fontTools".
I think the doctests are still useful for documentation purposes, but for unit testing we should use proper unittest or pytest tests modules.
I have converted all the doctests to pytest tests, so the coverage should not decrease.

BTW, I named the folder with an uppercase T for consistency with "Lib" and "Doc".


anthrotype:
The reason why coverage dropped from 73 to 65 % is because previously pytest was discovering tests in Lib/fontTools so it had to import everything in there, even modules which we did not have explicit tests for. For coverage.py, even simply importing a module means you "cover" the top-level module-scope statements and definitions (ie. the ones which are run when importing it).

Hence the previous 73% was bloated, while the current 65 % is more a accurate estimate:

https://codecov.io/gh/fonttools/fonttools/compare/c54d7548e90c06e35b827d3f6e1913d8f9eca181...89b688cecf53b1135e506297488c4c8d83475ebf/changes

behdad:
> The reason why coverage dropped from 73 to 65 % is because previously pytest was discovering tests in Lib/fontTools so it had to import everything in there, even modules which we did not have explicit tests for. For coverage.py, even simply importing a module means you "cover" the top-level module-scope statements and definitions (ie. the ones which are run when importing it).
> 
> Hence the previous 73% was bloated, while the current 65 % is more a accurate estimate:

But, just testing that importing modules works is useful testing ;).

anthrotype:
Ok, then. I'll have pytest also discover doctests from Lib/fontTools.

anthrotype:
doctests are re-enabled with 223c39f77e437ce06a2bd921ff04e9999293a451

but I run them in a separate pytest run outside coverage, so we don't feel like cheating and actually write unit tests ;)

jamesgk:
Is there any way now to e.g. import makeTTFont from feaLib/builder_test.py? I found this useful in other unit tests. Could that function be moved back into the installed code, if necessary?

anthrotype:
No, you can't import it any more.
Sure, we could move it to `fontTools.misc.testTools`. But it's a very simple function which one can easily redefine in one's tests.
Note that the testTools module already have a `FakeFont` and a `MockFont`, used in different test modules, with partially overlapping functionalities.

jamesgk:
Yea, I ended up just redefining it. So probably not worth worrying about.

PR #813, opened 2017-01-18
convert README to reStructuredText for use with PyPI

anthrotype:
The Python Package Index does not support Markdown, whereas Github is happy with either Markdown or reStructuredText.

They are very similar to each other, and `pandoc` makes it easy to convert from one to the other. E.g.:

```sh
$ pandoc --from=markdown --to=rst --output README.rst README.md
```


anthrotype:
@justvanrossum I've always wondered what those extra "Copyright" listed at the bottom are there for...
Is the "Stichting Mathematisch Centrum" the copyright for Python itself?
I've never seen this for other third-party python packages. Do we really need them here?
Or could we just keep the the top one?
And what about the date... 😕 

```markdown
### Copyrights

Copyright (c) 1999-2004 Just van Rossum, LettError (just@letterror.com)  
See [LICENSE](LICENSE) for the full license.

Copyright (c) 2000 BeOpen.com. 
All Rights Reserved.

Copyright (c) 1995-2001 Corporation for National Research Initiatives. 
All Rights Reserved.

Copyright (c) 1991-1995 Stichting Mathematisch Centrum, Amsterdam. 
All Rights Reserved.
```

justvanrossum:
Oh that must have been for the TTX app that bundled Python. The three latter copyrights are old Python copyrights. Should totally be removed. But yeah, what to do, it's no longer purely my thing, so what can we do that is not too complicated?

anthrotype:
> what to do, it's no longer purely my thing, so what can we do that is not too complicated?

do we actually need something at all? I'm no legal expert. Maybe @davelab6 knows more.

anthrotype:
I also converted the NEWS file to reStructuredText, and have the setup.py `long_description` be a concatenation of README plus NEWS.
This way on the next PyPI release we shall have a nice project page which also includes the changelog.

PR #819, opened 2017-01-23
Indent XML output for TT assembly instructions

jenskutilek:
Indentation for TTX XMLF is 2 spaces. I chose 4 spaces for indentation of instructions to make it more noticeable. I hope it’s not a problem, because assembly code lines are usually very short.


anthrotype:
Thanks!
A little test case in https://github.com/fonttools/fonttools/blob/master/Tests/ttLib/tables/ttProgram_test.py would also be nice ;)

Lib/fontTools/ttLib/tables/ttProgram.py
```diff
@@ -248,9 +250,13 @@ def toXML(self, writer, ttFont):
 				writer.begintag("assembly")
 				writer.newline()
 				i = 0
+				indent = 0
 				nInstr = len(assembly)
 				while i < nInstr:
 					instr = assembly[i]
+					if _unindentRE.match(instr):
+						indent -= 1
+					writer.write("  " + "    " * indent)
```
anthrotype: @jenskutilek why here are you doing " " + " " * indent and not just " " * indent?
I presume it's because the default indentation for the XMLWriter is two spaces so you are padding it to four...

I think it would be better to stick to whatever XMLWriter.indentwhite value is, i.e. currently two spaces by default, and maybe add a new command-line option for ttx that passes a different indentation setting down to the XMLWriter instance (which already takes a indentwhite=INDENT keyword argument in its constructor).

This way the indentation for the whole xml file would be consistent, i.e. either two spaces everywhere, or four, or tabs, etc. instead of mixing two/four like in here.

WDYT?

jenskutilek:
I’ll have a go at writing a test case.

BTW here’s a [Sublime Text syntax file](https://github.com/jenskutilek/TextMate-Bundles/blob/master/XML-TTX.sublime-syntax) for TTX, I have just added syntax matches for assembly code.

bildschirmfoto 2017-01-23 um 18 52 04

anthrotype:
super cool! ... well, as cool as hinting could be ;)

Lib/fontTools/ttLib/tables/ttProgram.py
jenskutilek: I agree, better stick to what’s already there. Will have a look at it tomorrow.

anthrotype: ah, since you are at it, could you also fix the inconsistency in the indentation levels for the glyph instructions?

```xml
<instructions><assembly>
    NPUSHB[ ]	/* 24 values pushed */
    26 72 28 16 17 175 28 1 38 28 17 17 28 38 3 3 20 74 13 3 33 74 3 9
    ...
    IUP[0]	/* InterpolateUntPts */
    IUP[1]	/* InterpolateUntPts */
  </assembly></instructions>
```

I think you just need to properly add a `newline()` before and after dumping the glyph program. Not sure it was not already like that:

```diff
diff --git a/Lib/fontTools/ttLib/tables/_g_l_y_f.py b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
index a346813..437a646 100644
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -375,7 +375,9 @@ class Glyph(object):
                                writer.newline()
                        if self.numberOfContours:
                                writer.begintag("instructions")
+                               writer.newline()
                                self.program.toXML(writer, ttFont)
+                               writer.newline()
                                writer.endtag("instructions")
                                writer.newline()
```

anthrotype:
Thanks a lot adding a test! :)

Before we can merge this, I still have a few comments.

In 07e4f24, you fixed the indentation of assembly and bytecode XML elements inside the instructions element, as I had suggested. However, now I realize that glyphs with empty instructions/assembly will end up having four lines instead of two, which is arguably a bit of a waste:

```diff
-      <instructions><assembly>
-        </assembly></instructions>
+      <instructions>
+        <assembly>
+        </assembly>
+      </instructions>
```

Maybe this was the reason why `ttProgram.Program.toXML` method did not end with a `writer.newline()` call, unlike the rest of the `toXML` methods inside `ttLib`. (cc @justvanrossum?)

Now I wonder if we could _not_ write empty assembly elements, like this:

```diff
-      <instructions><assembly>
-        </assembly></instructions>
+      <instructions>
+      </instructions>
```

The instructions element must be there, even if empty, because it's the presence of a 'program' attribute that defines whether a composite should have the `WE_HAVE_INSTRUCTIONS` flag set.

In any case, whether we drop empty assembly elements or not, I think the `writer.newline()` which follows `self.program.toXML(writer, ttFont)` should be in the placed inside the `toXML` method instead of its caller, i.e. after each `writer.endtag`, for consistency with the rest of the codebase.

```diff
diff --git a/Lib/fontTools/ttLib/tables/_g_l_y_f.py b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
index 735c546..88b7208 100644
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -358,7 +358,6 @@ class Glyph(object):
                                writer.begintag("instructions")
                                writer.newline()
                                self.program.toXML(writer, ttFont)
-                               writer.newline()
                                writer.endtag("instructions")
                                writer.newline()
                else:
@@ -379,7 +378,6 @@ class Glyph(object):
                                writer.begintag("instructions")
                                writer.newline()
                                self.program.toXML(writer, ttFont)
-                               writer.newline()
                                writer.endtag("instructions")
                                writer.newline()
 
diff --git a/Lib/fontTools/ttLib/tables/ttProgram.py b/Lib/fontTools/ttLib/tables/ttProgram.py
index d189b70..8ffb5ca 100644
--- a/Lib/fontTools/ttLib/tables/ttProgram.py
+++ b/Lib/fontTools/ttLib/tables/ttProgram.py
@@ -246,6 +246,7 @@ class Program(object):
                                writer.newline()
                                writer.dumphex(self.getBytecode())
                                writer.endtag("bytecode")
+                               writer.newline()
                        else:
                                writer.begintag("assembly")
                                writer.newline()
@@ -279,11 +280,13 @@ class Program(object):
                                        if _indentRE.match(instr):
                                                indent += 1
                                writer.endtag("assembly")
+                               writer.newline()
                else:
                        writer.begintag("bytecode")
                        writer.newline()
                        writer.dumphex(self.getBytecode())
                        writer.endtag("bytecode")
+                       writer.newline()
 
        def fromXML(self, name, attrs, content, ttFont):
                if name == "assembly":
```

anthrotype:
hey @jenskutilek, did you see this?
https://github.com/jenskutilek/fonttools/issues/1

jenskutilek:
I’ve seen it now ;)

Tests/ttLib/tables/ttProgram_test.py
```diff
@@ -101,19 +100,19 @@ def test_roundtrip(self):
         assert BYTECODE == p.getBytecode()
 
     def test_xml_indentation(self):
-        with open(TTPROGRAM_TTX, 'rb') as f:
-            ttProgramXML = f.read().splitlines()
+        with open(TTPROGRAM_TTX, 'r') as f:
```
anthrotype: it's good practice to always use `encoding='utf-8'` when reading text files which we know are encoded as UTF-8, since on Windows `sys.getdefaultencoding()` does not returns 'UTF-8', but some narrow codepage. Though in this case it should be fine since the `ttProgram.ttx` only contains ascii.

jenskutilek: Sorry, I accidentally lost the `encoding = 'utf-8'` in a merge conflict.

justvanrossum:
> Maybe this was the reason why `ttProgram.Program.toXML` method did not end with a `writer.newline()` call, unlike the rest of the `toXML` methods inside `ttLib`. (cc @justvanrossum?)

Yeah, I think that's correct. If you find a way to make this nicer in all situations, go ahead, but it feels a bit like a change for change's sake.

anthrotype:
@jenskutilek no, no, that was not the issue why Appveor fails, please revert c043cfb

the problem is XMLWriter defaults to the os-specific line endings; instead you should pass `newlinestr="\n"` to `XMLWriter` in your test.

anthrotype:
```diff
-        writer = XMLWriter(buf)
+        writer = XMLWriter(buf, newlinestr='\n')
```

jenskutilek:
Are you sure? When I revert the newline at the end of file, the test fails again, even with `XMLWriter(buf, newlinestr='\n')`

anthrotype:
> If you find a way to make this nicer in all situations, go ahead, but it feels a bit like a change for change's sake.

The thing is, once we indented the FDEF and IF instructions, the inconsistent indentation of the assembly/bytecode elements came to the fore.

The current solution looks nice to me:

- A glyph with instructions have nicely indented <assembly>, and within those, indented FDEF/IF,etc,
- A contour-only glyph without instructions has an empty instructions element.
- A composite glyph without instructions, has no instructions element.
- An empty glyph, with no contours nor component data, has no sub-elements.

This solution is also backward compatible, in the sense that ttx files which do contain empty assembly/bytecode elements will work the same.

anthrotype:
> Are you sure?

I'm sure we do need a newline at the end of the ttProgram.ttx, as that's what we do in here:
https://github.com/fonttools/fonttools/pull/819/commits/499bc960f9f7e9c7428e938b99821b26bfc83bb8

jenskutilek:
I was confused ... my commit message wasn’t what actually changed. I think everything is correct now

anthrotype:
no, we _need_ a newline at the end of the ttProgram.ttx. You reverted it twice.
Please remove this commit: ff9c046

jenskutilek:
But the test passes locally now ... apparently I'm still confused.

anthrotype:
the appveyor tests still fail because you didn't add the newline at the end of the ttProgram.ttx:
https://ci.appveyor.com/project/fonttools/fonttools/build/1.0.447/job/96wa7x1eir37q85p#L1152

As you can see, it is missing:
https://github.com/fonttools/fonttools/pull/819/files#diff-6187713984923e56fc3892dc9a6a8b74R1627

please add that

anthrotype:
There you go, it's passing now!
Sorry for all the troubles.

next time, maybe you could create a branch instead of submitting a PR directly from your fork's master, so you can squash commits, force-push, etc. which shouldn't happen on master.

jenskutilek:
OK, thanks for your help! I will add a branch next time, this just seemed only a small change in the beginning ;)

anthrotype:
Sorry, there's one thing which I'm still not super happy about...
It's when a instructions element does not contain any subelements, in which case it could be reduced to a simple tag with the same meaning:

```diff
-      <instructions>
-      </instructions>
+      <instructions/>
```

I'll try to see if I can find a solution for that after lunch.

jenskutilek:
What's confusing me is that now when I run `ttProgram_test.py` locally on macOS 10.12 with the system Python 2.7.10, it fails … but the AppVeyor test passes.

anthrotype:
You screwed your local repo master branch, si I'd suggest you delete and re-clone it for your remote.

anthrotype:
It appears that both ElementTree and lxml serialize empty elements using the special empty `<element/>` notation, but both allow to turn this off.
ElementTree `tostring()` function has `short_empty_elements=True` argument, while `lxml.ElementTree` class has a `write_c14n` method that uses the Canonical XML formatting, which among other things disallows the special handling of short elements.

https://en.wikipedia.org/wiki/Canonical_XML

> empty elements are encoded as start/end pairs, not using the special empty-element syntax

Although it wouldn't be too difficult to add such a feature to our XMLWriter, perhaps for the sake of simplicity we could stick to what we currently have.

justvanrossum:
One can generate 'short' elements with XMLWriter with its `simpletag()` method, just not automatically.

anthrotype:
@jenskutilek I sent you a few more cosmetic changes to https://github.com/jenskutilek/fonttools/pull/3

Could you please also write a small changelog entry at the top of `NEWS`? It's just a short phrase summarizing any user-facing changes for this pull request (#819).

Thank you for contribution!

anthrotype:
> One can generate 'short' elements with XMLWriter with its simpletag() method, just not automatically.

sure, I know. Would you like to do that for `<instructions></instructions>`? It'd be just a simple check `if self.program:` do begintag and endtag, else do simpletag.

justvanrossum:
Sure, go for it!

behdad:
I just envy how much time you guys have on your hands ;).

anthrotype:
@jenskutilek excellent, thank you!

miguelsousa:
> I just envy how much time you guys have on your hands ;)

And I really appreciate Cosimo's patience and attention to detail 👏

PR #822, opened 2017-01-26
[WIP] Add test fonts from AOTS and start of a driver, copied from mtiLib

behdad:
I don't have time to finish this, so posting such that someone else can.  The fonts and expected output are in there.  The test driver is copied from mtiLib; needs to be adapted.

The tests are similar to what's already in there: compare expected XML to XML-from-binary to XML-from-XML.  We can also try compiling from XML and back.  Might also decide to do a deep-compare of object systems. Might be good to make the various tests separate, you know, tests within the framework.

Anyway, I think this is a good generic infrastructure to add tests for tables without writing code.


anthrotype:
Thanks Behdad!
I started looking at this today, and I like the approach.

There are still a few xml2xml tests failing (e.g. `classdef1_font4.otf-GSUB` and `classdef2_font4.otf-GSUB` raise `AttributeError: classDefs`).

Since we are using pytest as test runner, I thought we could take advantage of its powerful test parametrization features, so I converted the tables_test.py module to a pytest-style test, and used `pytest_generate_tests` (see http://doc.pytest.org/en/latest/parametrize.html#basic-pytest-generate-tests-example)

I'll continue tomorrow.

Tests/ttLib/tables/data/aots/base.ttx.CFF
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<ttFont ttLibVersion="3.5">
+<ttFont sfntVersion="OTTO" ttLibVersion="3.5">
```
behdad: How come my ttx run didn't produce sfntVersion? Sounds like a bug somewhere.

anthrotype: you probably used the "split tables" option `ttx -s font.ttf` which only writes the sftnVersion in the index ttx file.

Tests/ttLib/tables/tables_test.py
```diff
@@ -1,327 +1,298 @@
...
+ttLibVersion_RE = re.compile(r' ttLibVersion=".*"')
 
-    def setUp(self):
-        pass
 
-    def tearDown(self):
-        pass
+def getpath(testfile):
+    path = os.path.dirname(__file__)
+    return os.path.join(path, "data", testfile)
```
behdad: You also need to split testfile on '/' and rejoin it.  My code did that in `generate_mti_file_test`. Not the best place, I agree.

behdad:
Thanks Cosimo!  Feel free to port mtiLib test driver to the same infrastructure.  Maybe that, this, and feaLib ones, can all be merged into one infrastructure.

Tests/ttLib/tables/tables_test.py
anthrotype: it doesn't look like it's needed. Maybe the `os` module is the right thing when joining the paths.
At least Appveyor didn't complain about it (you can see the number of failing tests is the same as on Travis):

https://ci.appveyor.com/project/fonttools/fonttools/build/1.0.459/job/knqaovpde5a65g37

I think `os.path.normpath` would be the best way to do that.

anthrotype:
I rebased the branch and fixed the rest of the failing tests.
Everything should be good now. Waiting for the bots to complete, then I think we can merge this.

behdad:
Thanks.

PR #827, opened 2017-01-30
[varLib.merger] set initializer for reduce() to handle empty sequences

anthrotype:
When reduce() receives an empty sequence, it raises TypeError, unless it is also given a third 'initializer' argument

ValueFormat values should default to 0, so we shall use that as initializer.

Also, the reduce() built-in is no longer available on Python 3.
It's still accessible for both py2 and py3 from functools.

Should fix https://github.com/googlei18n/fontmake/issues/241


brawer:
Looks good, thank you!

Issue #559, opened 2016-03-27
[feaLib] Handle dash in glyph names


behdad:
I know this has been discussed before, but it's still a pain, that glyphs2ufo chained to ufo2ft can fail.  In my example:

```
fontTools.feaLib.error.FeatureLibError: features.fea:12:5179: Bad range: "LL" and "middleWelsh" should have the same length
```

I like to suggest two changes to feaLib, both of which will ideally be pushed to AFDKO / Fea spec as well:
- A dash, when following a backslash, can be part of a glyph name to mean a single 'dash' and as such work around the current limitation that dashes cannot be part of glyph names.  In fact, I'd go as far as suggesting that a backslash would escape any punctuation to be part of the glyph name,
- When the above-mentioned error occurs, ie. when a dash range fails to parse, then try parsing the dashed sequence as a glyph name (only if there were no spaces involved).  In fact, in most cases, either a range or a glyph name is valid, not both.  It's ok to fail if there is ambiguity.

Implementing one or both of these makes the glyphs2ufo+ufo2ft experience much better.  Right now, fontmake replaces dashes with underscores before calling the above tools.  That's too much to ask each client to do.

@schriftgestalt @jamesgk

khaledhosny:
Feature files already use the backslash to escape whole glyph names to avoid conflict with reserved words as well as for CIDs, how would that work here?

schriftgestalt:
I don't think we need new markup. 

In my makeotf fork, I added method that accepts hyphens as part of glyph names and in case if can’t find a glyph to a name and the name has a hyphen, it treats it as a range. This works very well. You suggested a similar solution that first checks for range and then for glyph name. That can work well, too. It depends on the parser what method fits better. 

Adding backslash is not a good idea, very designer unfriendly ;)

twardoch:
@schriftgestalt I remember you told me the MakeOTF implementation of ranges was flawed in some other aspects. Can you remond us what was the problem there?

schriftgestalt:
There is no flaw. The spec and implementation are not matching up.

from the spec:
> For non-CID fonts, the ordering is independent of the ordering of glyphs in the font. 'firstGlyph' and 'lastGlyph' must be the same length and can differ ...

The implementation in makeOTF does the opposite. It does use the GID of the first and second glyph and uses all GIDs in between.

schriftgestalt:
[...]

PR #835, opened 2017-02-11
Support glyph names with dashes

brawer:
The OpenType Feature File Syntax has been changed to support dashes:
https://github.com/adobe-type-tools/afdko/issues/152

Resolves https://github.com/fonttools/fonttools/issues/559.
Needed for https://github.com/googlei18n/fontmake/issues/249.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -1,3 +1,6 @@
+- [feaLib] Glyph names can have dashes, as per new AFDKO syntax v1.20 (#559).
+- [feaLib] feaLib.Parser now needs the font's glyph map for parsing.
+
 3.6.3 (released 2017-02-06)
 ---------------------------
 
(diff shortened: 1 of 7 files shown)


NEWS.rst
```diff
@@ -1,3 +1,10 @@
+3.6.4 (not yet released)
```
anthrotype: Thanks Sascha!

Could you please remove the custom header in the changelog? Just add your new entries without leaving a blank line at the top. The version string, timestamp, dash titling and newlines are automatically generated by the release script. See inside the setup.py, the `release` command class.

brawer: Thanks for the explanation! Done.

behdad:
We should update fontmake / glyphsLib to not replace dashes with underscores now.

brawer:
Yes, https://github.com/googlei18n/fontmake/issues/249.

Issue #833, opened 2017-02-08
In GSUB alternate substitutions, order can be changed on round-trip.


readroberts:
On my system, with the main branch of fonttools, lists of GSUB alternate glyphs gets sorted alphabetically by glyph name when writing GSUB to XML. When the TTX dump is compiled back to a file, the alternate substitution glyph order is then in alphabetic order. This is a problem as alternate substitutions are selected by index in the alternate list, not by GID or glyph name, and the change in order will change which glyph is selected. 

Example from ArnoPro-Bold aalt feature:.
Original font: sub ampersand from [ u0026.a u0026.db u0026.dc u0026.da]
After roundtrip: sub ampersand from [ u0026.a u0026.da u0026.db u0026.dc];

justvanrossum:
Seems to have been caused by https://github.com/fonttools/fonttools/commit/f38cc9f821dde8a2f44fa00c88c425a151472fdd

anthrotype:
I think we should fix this before we can tag a new release. 
@brawer are you already on it? Otherwise no prob, I can reassign to me.

brawer:
Almost there. It also messed up the ordering of alternates in MTI feature files; fixing that too.

brawer:
OK, done. @anthrotype, mind cutting a new release? I think the replication of makeotf behavior for `aalt` can be done later.

anthrotype:
Sure, I'll do it shortly. Thanks for your work!

PR #837, opened 2017-02-11
Preserve ordering of glyph alternates when round-tripping through TTX

brawer:
Also fixes a bug where glyph alternates in MTI feature files were
wrongly sorted by glyph name. After this change, the output is using
the same ordering as in the input MTI feature file.

Fixes https://github.com/fonttools/fonttools/issues/833.


anthrotype:
Thanks Sascha!

Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -536,7 +536,7 @@ def preWrite(self, font):
 		cov.glyphs = [ item[1] for item in items]
 		alternates = []
 		setList = [ item[-1] for item in items]
-		for  set in setList:
+		for set in setList:
```
khaledhosny: Since `set` is a built-in class, wouldn’t it be safer/more readable to use a different variable name?

justvanrossum: Absolutely. I think this code predates the `set()` builtin, though... Feel free to improve!

PR #838, opened 2017-02-12
[feaLib] Make `include()` work like makeotf

miguelsousa:
The way **makeotf** handles `include()` is to literally replace the statement with the contents of the file that is being referenced, therefore the paths need to be relative to the first file instead of relative to each other. I've adjusted the test files to cover other file location examples. The test results remain the same.


anthrotype:
Hi Miguel, thanks for your pull request.
I understand this is how makeotf, i.e. the reference implementation, works, but I'm 100 % convinced this is right.

As you know, the OpenType Feature File specification simply says:

http://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#3
> If the file name were relative, then it would be appended to the directory of the including feature file.

The spec however does not say thta "the paths need to be relative to the first file instead of relative to each other".

I wonder whether the spec should be modified to clarify this?

I personally think the way feaLib implements the include statement makes more sense, as it's similar to how the C preprocessor handles relative paths like `#include "../myheader.h"`, ie. relative to the current source file.

/cc @brawer @readroberts

anthrotype:
I see you moved `Tests/feaLib/data/include0.fea → Tests/feaLib/include0.fea`.
We usually try to keep test data files inside the respective `data` subfolders.

miguelsousa:
I don't oppose changing the current behavior, but it has been working like this for as long as I can remember, and we have a fair number of projects that rely on it. Other developers may depend on it too.

miguelsousa:
@anthrotype yes, I did move `include0.fea` because I wanted to have an example where the included file is one level above the base file, and I don't see a way to do so without shuffling a few other files. So I opted for the simplest change. I understand that it puts the file outside of the `data` folder, but it seems like an edge case that we can live with.

miguelsousa:
Here's a simple example to illustrate the situation that the PR is trying to address.
Consider the files arranged like this
```
└── DIR1
    ├── file1.fea
    ├── file2.fea
    └── DIR2
        └── base.fea
```
where `base.fea` includes `file1.fea` which in turn includes `file2.fea`.

---
For both **makeotf** and **feaLib**, the contents of **`base.fea`** are
```
include(../file1.fea);
```
and the contents of **`file2.fea`** are
```
@CLASS = [a b c];
```
---
But when it comes to the contents of **`file1.fea`**, **makeotf** requires
```
include(../file2.fea);
```
whereas **feaLib** needs it to be
```
include(file2.fea);
```

brawer:
Agree with @anthrotype that the makeotf behavior is funky, but I’d say supporting existing sources (and compatibility with makeotf) is more important. For the location of the test files, I’ll move them into the data directory in a follow-up change. Will also file a bug on the feature file spec, since it would be nice if the spec reflected reality.

brawer:
It might be good to clarify the spec, since the path resolution is quite unusual compared to other languages. Filed https://github.com/adobe-type-tools/afdko/issues/164.

miguelsousa:
Thanks Sasha.

anthrotype:
Thank you Miguel and Sascha.

PR #840, opened 2017-02-13
Change no GSUB assert to warning.

zjusbo:
In most cases the dups are space, comma, period or other non language related simple symbols. It should be reasonable to degrade the assert to a warning and print the duplicated glyph names.

### DIFF
--- a/Lib/fontTools/merge.py
+++ b/Lib/fontTools/merge.py
@@ -505,7 +505,9 @@ def merge(self, m, tables):
 	assert len(tables) == len(m.duplicateGlyphsPerFont)
 	for i,(table,dups) in enumerate(zip(tables, m.duplicateGlyphsPerFont)):
 		if not dups: continue
-		assert (table is not None and table is not NotImplemented), "Have duplicates to resolve for font %d but no GSUB" % (i + 1)
+		if table is None or table is NotImplemented:
+			log.warn("Have duplicates %s to resolve for font %d but no GSUB" % (dups, (i + 1)))
+			continue
 		lookupMap = {id(v):v for v in table.table.LookupList.Lookup}
 		featureMap = {id(v):v for v in table.table.FeatureList.FeatureRecord}
 		synthFeature = None
@@ -923,7 +925,6 @@ __all__ = [
 @timer("make one with everything (TOTAL TIME)")
 def main(args=None):
 	from fontTools import configLogger
-
 	if args is None:
 		args = sys.argv[1:]
 
--- a/Lib/fontTools/varLib/merger.py
+++ b/Lib/fontTools/varLib/merger.py
@@ -3,6 +3,7 @@ Merge OpenType Layout tables (GDEF / GPOS / GSUB).
 """
 from __future__ import print_function, division, absolute_import
 from fontTools.misc.py23 import *
+from fontTools.misc import classifyTools
 from fontTools.ttLib.tables import otTables as ot
 from fontTools.ttLib.tables import otBase as otBase
 from fontTools.ttLib.tables.DefaultTable import DefaultTable
@@ -296,9 +297,71 @@ def _PairPosFormat1_merge(self, lst, merger):
 
 	merger.mergeLists(self.PairSet, padded)
 
+def _ClassDef_invert(self):
+	if not self or not self.classDefs:
+		return []
+
+	classDefs = self.classDefs
+	m = max(classDefs.values())
+
+	ret = []
+	for _ in range(m + 1):
+		ret.append(set())
+
+	for k,v in classDefs.items():
+		ret[v].add(k)
+
+	return ret
+
+def _ClassDef_merge_classify(lst, ignoreClass0=False):
+	self = ot.ClassDef()
+	self.classDefs = classDefs = {}
+
+	classifier = classifyTools.Classifier(sort=False)
+	for l in lst:
+		sets = _ClassDef_invert(l)
+		if ignoreClass0:
+			sets = sets[1:]
+		classifier.update(sets)
+	classes = classifier.getClasses()
+
+	if ignoreClass0:
+		classes.insert(0, set())
+
+	for i,classSet in enumerate(classes):
+		for g in classSet:
+			classDefs[g] = i
+
+	return self, classes
+
 def _PairPosFormat2_merge(self, lst, merger):
-	# Everything must match; we don't support smart merge yet.
-	merger.mergeObjects(self, lst)
+	merger.mergeObjects(self, lst,
+			    exclude=('ClassDef2', 'Class2Count',
+				     'Class1Record'))
+
+	self.ClassDef2, classes = _ClassDef_merge_classify([l.ClassDef2 for l in lst], ignoreClass0=True)
+	self.Class2Count = len(classes)
+	class1Recordses = []
+	for l in lst:
+		classDef2 = l.ClassDef2.classDefs
+		class1Records = []
+		for rec1old in l.Class1Record:
+			oldClass2Records = rec1old.Class2Record
+			rec1new = ot.Class1Record()
+			class2Records = rec1new.Class2Record = []
+			for classSet in classes:
+				if not classSet: # class=0
+					rec2 = oldClass2Records[0]
+				else:
+					exemplarGlyph = next(iter(classSet))
+					klass = classDef2.get(exemplarGlyph, 0)
+					rec2 = oldClass2Records[klass]
+				class2Records.append(rec2)
+			class1Records.append(rec1new)
+		class1Recordses.append(class1Records)
+
+	self.Class1Record = list(class1Recordses[0]) # TODO move merger to be selfless
+	merger.mergeLists(self.Class1Record, class1Recordses)
 
 @AligningMerger.merger(ot.PairPos)
 def merge(merger, self, lst):


brawer:
Could you test if the conflicting glyphs have the same contours and advance width? If the conflicting glyphs are identical, don’t even warn.

zjusbo:
@brawer outlines and advance width are tested before warning. PTAL. IIRC, we have to discard duplicated glyphs if one unicode maps to multiple glyphs in different fonts in merging? We can definitely assign the duplicated glyphs with a different name, but essentially only one glyph can be used in uni -> glyph mapping?

zjusbo:
@anthrotype Thanks for letting me know. Probably it's because my eclipse automatically convert tab to 2 spaces but my sublime doesn't. And I mixed used these two editors together when editing the file. It seems I can't approve your change request now. I changed the space into tabs. PTAL.

Lib/fontTools/merge.py
```diff
@@ -564,6 +572,50 @@ def merge(self, m, tables):
 def mapLookups(self, lookupMap):
 	pass
 
+
+# Gets unicode and font index from the composite gid
```
anthrotype: I like you added comments describing function parameters, return value, etc. In Python these "docstrings" are usually placed below a function definition wrapped with `"""` quotes.

Lib/fontTools/merge.py
```diff
@@ -923,7 +976,6 @@ def _postMerge(self, font):
 @timer("make one with everything (TOTAL TIME)")
 def main(args=None):
 	from fontTools import configLogger
-
```
anthrotype: I see no reason to delete these empty these lines.

Lib/fontTools/merge.py
```diff
@@ -505,7 +505,15 @@ def merge(self, m, tables):
 	assert len(tables) == len(m.duplicateGlyphsPerFont)
 	for i,(table,dups) in enumerate(zip(tables, m.duplicateGlyphsPerFont)):
 		if not dups: continue
-		assert (table is not None and table is not NotImplemented), "Have duplicates to resolve for font %d but no GSUB" % (i + 1)
+		if table is None or table is NotImplemented:
+			# Checks whether the dups are equivalent or not.
+			# Discard gid if its shape is not equal to that of oldgid.
+			for oldgid, gid in dups.iteritems():
```
anthrotype: iteritems() no longer exists on python3, please just use `items()` for py2.py3 compatibility.

Lib/fontTools/merge.py
```diff
@@ -564,6 +572,50 @@ def merge(self, m, tables):
...
+# @param {list<TTFont>} fonts
+# @param {string} gid_0 - e.g. uni0000#1
+# @param {string} gid_1 - e.g. space#2
+# @return {bool}
+def isGlyphSame(fonts, gid_0, gid_1):
+	# Checks outline
+	index_0 = getGlyphNameAndFontIndex(gid_0)[1]
+	index_1 = getGlyphNameAndFontIndex(gid_1)[1]
+	
+	assert fonts[index_0].has_key('glyf') and fonts[index_1].has_key('glyf')
```
anthrotype: just a nit: I'd use `'glyf' in fonts[index_0]` instead of `has_key`. Even if both works, since `TTFont` is a user-defined class and not a built-in `dict` which no longer has `has_key` on python3, the latter stinks too much of legacy python. I wonder if we should fire a deprecation warning and eventually get rid of it.

Lib/fontTools/merge.py
```diff
@@ -572,24 +572,38 @@ def merge(self, m, tables):
...
-# Two glyphs are considered as equal iff:
-# 1. Outlines are equal.
-# 2. Advance width are equal.
-# @param {list<TTFont>} fonts
-# @param {string} gid_0 - e.g. uni0000#1
-# @param {string} gid_1 - e.g. space#2
-# @return {bool}
+
 def isGlyphSame(fonts, gid_0, gid_1):
+	"""Checks if the given glyphs specified by gid are euqal or not.
```
anthrotype: typo `euqual` -> `equal`

Lib/fontTools/merge.py
zjusbo: I was thinking to make code more compact, but it seems these empty lines separate these the logic blocks. I'm adding them back.

Lib/fontTools/merge.py
zjusbo: in is definitely more pythonic. I'm changing it.

Lib/fontTools/merge.py
zjusbo: Thanks for letting me know. Changed.

Lib/fontTools/merge.py
zjusbo: I see. Changed to use docstring. PTAL.

Lib/fontTools/merge.py
```diff
@@ -564,6 +572,64 @@ def merge(self, m, tables):
...
+
+	Returns:
+		A tuple with two elements. The first element is a string representing the 
+		glyph name. The second element is an integer representing the font index in 
+		command line args.
+	
+	Example:
+		'uni0000', 3 = getGlyphNameAndFontIndex('uni0000#3')
+	"""
+	assert '#' in gid and gid.split('#')[-1].isdigit() and len(gid.split('#')) == 2, 'incorrect gid format'
```
anthrotype: why you assert `len(gid.split('#')) == 2`? I think `#` is a valid character in a glyph name so it could already be present. I think it's enough to split at the right-most '#', with `rsplit('#', 1)`.

zjusbo: Oh I thought '#' wasn't a valid character in glyph name. Changed to use rsplit('#', 1)

brawer:
Can you you add tests? https://github.com/fonttools/fonttools/blob/master/Tests/merge_test.py Perhaps browse also at some other tests in fonttools, so see how the testing is done in the codebase.

It might be good to test what happens when the input fonts have identical glyphs (eg. comma or period) that participate in GSUB or GPOS lookups. (@behdad, should we really open this can of worms? I guess at some point, it’ll be easier to merge sources manually; algorithmic subsetting is probably easier than algorithmic merging).

zjusbo:
I was stuck at mocking GSUB table in adding test. Still figuring out how to create a mock GSUB table and add some data to it. I'm pining Behdad to get some tips on it.

zjusbo:
The GSUB table object is complicated and was not mocked in our test cases before. I tried to create a simple GSUB table in code manually but found it's time consuming and hard to maintain. I'm planing to import the GSUB table from font file directly. 

Following is my plan:

1. Add two test font files containing GSUB table to repository
2. Construct GSUB object from font files and merge them.
3. Compare the result with expected result.

behdad:
This does NOT look ok to me.

Comparing compiled glyf entries is wrong when 1. hinting is involved, and 2. component glyphs are involved.  Bo, as I responded before, you need to use a recording pen and compare the recordings for equality. The recording pen needs to decompose components.  I'm working on a simple way to do that.  This can wait a few more days.

Lib/fontTools/merge.py
```diff
@@ -566,6 +569,30 @@ def merge(self, m, tables):
...
 	pass
 
+# Check if the given glyphs specified by gid are euqal or not.
+# Two glyphs are considered as equal iff:
+# 1. Outlines are equal.
+# 2. Advance width are equal.
+def isGlyphSame(fonts, gid_0, gid_1):
+	# Checks outline
+  assert '#' in gid_0 and '#' in gid_1
+  index_0 = int(gid_0.split('#')[-1])
```
behdad: This is too much of a hack. We should find another way to wire down the fonts to this function if needed.

Lib/fontTools/merge.py
```diff
@@ -566,6 +569,30 @@ def merge(self, m, tables):
...
+  index_1 = int(gid_1.split('#')[-1])
+  
+  assert fonts[index_0].has_key('glyf') and fonts[index_1].has_key('glyf')
+  glyfTable_0 = fonts[index_0]['glyf']
+  glyfTable_1 = fonts[index_1]['glyf']
+  data_0 = glyfTable_0[gid_0].compile(glyfTable_0)
```
behdad: This doesn't work if the fonts have different hinting.  We still want to merge that case.
Worse, if the glyphs are composites and both refer to the same glyph indices in their original font, they do not necessarily look the same but this test will pass, and vice versa.

zjusbo:
@brawer Test cases are added. It turns out to be more complicated than I have originally thought. 
1. Learned from @behdad that GSUB table is special. It is built dynamically and I should check otTable and otData to explore the structure of it.
2. GSUB table is complex. I was thinking whether to import the GUSB table from a small notofont directly or build up it manually. Finally I chose the second approach to build up it manually since the GSUB table in noto-fonts is always big and contains many rules. It would be impractical (or hard) to check whether the result is what we expects.
3. IsGlyphSame is not tested yet because:
   1. The implementation and expected functionality is subject to change. As Behdad mentioned, the current implementation is of too much hack. 
   2. Testing requires a font object containing, glyf, vmtx, hmtx tables. Our current test sources in data/ folder does not work in this case. Probably importing a true font file as our test source sounds plausible.

Tests/merge_test.py
```diff
@@ -1,12 +1,260 @@
...
+		self.merger.fonts = []
+		self.merger.fontfiles = ['font1', 'font2']
+		fontTools.merge.isGlyphSame = lambda fonts, ogid, gid : False
+
+		self.mergedTable.merge(self.merger, [self.table1, self.table2])
+
+		table = self.mergedTable.table
+		self.assertEqual(table, self.table1.table)
+		# Checks warning
+		self.assertNotEqual(self.stderr.getvalue().strip(), '')
```
anthrotype: @zjusbo instead of redirecting `stderr`, you could use the `CapturingLogHandler` inside `fontTools.misc.loggingTools`. Do `git grep CapturingLogHandler` to see how this is used in other test modules.

anthrotype: If you are only interested in the log message text, the `setup.py` has a simpler context manager, `capture_logger`, which captures a logger output with a StringIO stream. Maybe that could be moved to `loggingTools` module if you think it could be useful elsewhere. With the `CapturingLogHandler` one has more fine grained access to all the log message attributes.

zjusbo: Thanks Cosimo. Changed to use CapturingLogHandler.

Tests/merge_test.py
```diff
@@ -233,8 +233,7 @@ def test_GSUB_merge_with_different_dups_and_no_second_GSUB(self):
 
 		table = self.mergedTable.table
 		self.assertEqual(table, self.table1.table)
-		self.assertTrue(len(captor.records) > 0)
-
+		self.assertTrue(len([r for r in captor.records if r.msg != '']) > 0)
```
anthrotype: what I mean is, assert that `len(["is dropped and replaced by" in r.msg for r in captor.records]) > 0`

zjusbo: Doing the warning message matching seems focus too much on the implementation detail not the behavior of the function. Say, the test case will fail if someone else just changes the words in warning message to fix typo or make it more user friendly. In this case I think we should not let the test case fails. Just check there is warning message shown to the user is good enough. Ideas?

anthrotype: @zjusbo sorry, I'm drunk.. forget the last message

This is what I actually meant:

```python
self.assertTrue(len([r for r in captor.records if "is dropped and replaced by" in r.msg]) > 0)
```

anthrotype: or `self.assertTrue(any("is dropped and replaced by" in r.msg for r in captor.records))`

zjusbo:
As a side note: the merge_noto.py in nototools https://github.com/googlei18n/nototools/blob/master/nototools/merge_noto.py
add a default gsub table to solve this problem.

brawer:
@behdad and @anthrotype, have another look? Bo’s PR now uses a recording pen as requested.

Lib/fontTools/merge.py
```diff
@@ -780,6 +849,7 @@ def merge(self, fontfiles):
 		# TODO Is it necessary to reload font?  I think it is.  At least
 		# it's safer, in case tables were loaded to provide glyph names.
 		fonts = [ttLib.TTFont(fontfile) for fontfile in fontfiles]
+		self.fonts = copy.deepcopy(fonts)
```
anthrotype: where is this `self.fonts` attribute used?

Tests/merge_test.py
```diff
@@ -1,12 +1,247 @@
...
+		self.assertEqual(table.FeatureList.FeatureCount, 3)
+		self.assertIn(self.table1.table.FeatureList.FeatureRecord[0], table.FeatureList.FeatureRecord)
+		self.assertIn(self.table2.table.FeatureList.FeatureRecord[0], table.FeatureList.FeatureRecord)
+		self.assertEqual(table.FeatureList.FeatureRecord[-1].FeatureTag, 'locl')
+
+	def test_GSUB_merge_with_same_dups_and_no_second_GSUB(self):
+		self.table2 = NotImplemented
+		self.merger.duplicateGlyphsPerFont = [{}, {'uni0032#0': 'uni0032#1'}]
+		self.merger.fonts = []
+		fontTools.merge.isGlyphSame = lambda fonts, ogid, gid : True
```
anthrotype: hm, this will persistently set the `merge` module's `isGlyphSame` attribute to the lambda function..
Use a `try: ... finally: ...` or a `@contextlib.contextmanager` to restore the original value at the end.

Lib/fontTools/merge.py
behdad: deepcopy'ing fonts is not a supported operation.  Why is this needed?

behdad:
> @behdad and @anthrotype, have another look? Bo’s PR now uses a recording pen as requested.

Does it? Which commit?

Can we close this and get a clean pull request (with no merge commits)?

zjusbo:
> @behdad and @anthrotype, have another look? Bo’s PR now uses a recording pen as requested.
Does it? Which commit?
Can we close this and get a clean pull request (with no merge commits)?
From this commit, the code uses recordingPen to compare the glyphs.
[add test case for GSUB emrge](https://github.com/fonttools/fonttools/pull/840/commits/1df61b4338f002dfd253e1722d5ece28e439cabd )

Sure. I'm closing it and create a clean pull request.

Issue #780, opened 2016-12-20
fea writer should escape non-ASCII characters in nameid strings


anthrotype:
as discussed in https://github.com/fonttools/fonttools/pull/778#discussion_r93225716

Cf. http://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#9.e

E.g. `nameid 9 "Joachim Müller-Lancé";` should become `nameid 9 "Joachim M\00fcller-Lanc\00e9";`

twardoch:
Trivia: for Unicode > ANSI tests, you can use `Jovica Veljović` ;)

twardoch:
I’d welcome it if `feaLib` diverged from http://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#9.e and also accepted pure-Unicode strings inside the `nameid` statements (i.e. that `feaLib` accepted UTF-8 `.fea` files). Perhaps we could “invent” some magic comment that could appear at the beginning of the `.fea`: 

```
#!org.fonttools.fealib
# -*- coding: utf-8 -*-
```

or something. It might be useful to agree on a simple set of “shebang-style” declarations at the beginning of `.fea` which could denote certain extensions to the format. I.e. without the shebang, all processors (makeOTF, feaLib, fontforge) would be expected to adhere to the common syntax, but (1) the user could enforce using one over the other in a higher tool, (2) things like https://github.com/typesupply/feaPyFoFum/ could be supported this way. 

To keep things a bit Pythonic-Robofabby, we could perhaps use the reverse-domain-style shebang strings: `#!com.adobe.afdko` for **makeOTF**, `#!org.fonttools.fealib` for **feaLib**, `#!org.fontforge` for **fontforge**. 

Maybe it shouldn’t be a "shebang` (`#!`) but something else (`#?` or so).

brawer:
Good idea, although I think it would be in the interest of everyone if we didn’t fragment the format. I’ve suggested a simplified version of your proposal to Adobe, hopefully they’ll add it to the spec. See https://github.com/adobe-type-tools/afdko/issues/165.

brawer:
Closing bug, since fonttools now implements the current spec.

PR #843, opened 2017-02-14
[feaLib]Escape nameid strings when writing feature files

brawer:
https://github.com/fonttools/fonttools/issues/780


anthrotype:
thank you Sascha!

PR #845, opened 2017-02-16
[subset] Set the correct glyph width for blank .notdef

mashabow:
`font.FDSelect[i]` returns a GID, not an index of `font.FDArray`.

Before this commit `.notdef` might have wrong glyph width in CharStrings, if its `fdSelectIndex` is not 0 in the input font.


anthrotype:
thanks. Would be nice to also have a test for that :)

mashabow:
OK, added!

anthrotype:
Thank you!

Issue #633, opened 2016-07-08
[feaLib] Class-to-class kerning zeroed out


jamesgk:
I've noticed that for class-to-class kerning rules, if one rule has a value of zero some other ones get cleared out. For example this feature source:

```
@public.kern1.K = [K X];
@public.kern2.O = [C O];
@public.kern2.V = [V W];

feature kern {
    pos @public.kern1.K @public.kern2.O -20;
    pos @public.kern1.K @public.kern2.V 0;
} kern;
```

produces this output: 

```
$ hb-shape out.otf K
[K=0+619]
$ hb-shape out.otf KO
[K=0+619|O=1+781]
```

with this GPOS:

``` xml
  <GPOS>
    <Version value="1.0"/>
    <ScriptList>
      <!-- ScriptCount=1 -->
      <ScriptRecord index="0">
        <ScriptTag value="DFLT"/>
        <Script>
          <DefaultLangSys>
            <ReqFeatureIndex value="65535"/>
            <!-- FeatureCount=1 -->
            <FeatureIndex index="0" value="0"/>
          </DefaultLangSys>
          <!-- LangSysCount=0 -->
        </Script>
      </ScriptRecord>
    </ScriptList>
    <FeatureList>
      <!-- FeatureCount=1 -->
      <FeatureRecord index="0">
        <FeatureTag value="kern"/>
        <Feature>
          <!-- LookupCount=1 -->
          <LookupListIndex index="0" value="0"/>
        </Feature>
      </FeatureRecord>
    </FeatureList>
    <LookupList>
      <!-- LookupCount=1 -->
      <Lookup index="0">
        <!-- LookupType=2 -->
        <LookupFlag value="0"/>
        <!-- SubTableCount=2 -->
        <PairPos index="0" Format="2">
          <Coverage Format="1">
            <Glyph value="K"/>
            <Glyph value="X"/>
          </Coverage>
          <ValueFormat1 value="0"/>
          <ValueFormat2 value="0"/>
          <ClassDef1 Format="2">
          </ClassDef1>
          <ClassDef2 Format="1">
            <ClassDef glyph="V" class="1"/>
            <ClassDef glyph="W" class="1"/>
          </ClassDef2>
          <!-- Class1Count=1 -->
          <!-- Class2Count=2 -->
          <Class1Record index="0">
            <Class2Record index="0">
            </Class2Record>
            <Class2Record index="1">
            </Class2Record>
          </Class1Record>
        </PairPos>
[...]

PR #850, opened 2017-02-16
[feaLib] Compile zero values to different formats based on context

brawer:
If a zero value appears in a SinglePos statement, feaLib continues to
produce ValueRecords of format 0. But if a zero value appears in a
PairPos statement inside a horizontal compilation context, feaLib now
produces ValueRecords of format 4. Likewise, if a zero value appears
in a PairPos statement inside a vertical compilation context, feaLib
now produces ValueRecords of format 8.

The OpenType Feature Syntax specification is completely silent about this,
but the new behavior matches that of makeotf.

https://github.com/fonttools/fonttools/issues/633

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -1012,7 +1012,7 @@ _VALUEREC_ATTRS = {
 }
 
 
-def makeOpenTypeValueRecord(v):
+def makeOpenTypeValueRecord(v, pairPosContext):
     """ast.ValueRecord --> (otBase.ValueRecord, int ValueFormat)"""
     if v is None:
         return None, 0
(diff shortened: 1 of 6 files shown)


behdad:
> But if a zero value appears in a
> PairPos statement inside a horizontal compilation context, feaLib now
> produces ValueRecords of format 4. Likewise, if a zero value appears
> in a PairPos statement inside a vertical compilation context, feaLib
> now produces ValueRecords of format 8.

This is adhoc and prune to fail.  There's nothing special about 4 or 8 in those contexts except for a sugar syntax in fea file format.

brawer:
Agree.

brawer:
See here for current behavior (both makeotf and fonttools now produce the same output): https://github.com/adobe-type-tools/afdko/issues/168

PR #854, opened 2017-02-21
explicitly require 'head' table in TTFont.save()

anthrotype:
it's implicitly required anyway, e.g. see ttx's `ttCompile`, maxp's `recalc`, OS/2 `compile`, etc.

/cc @justvanrossum

### DIFF
--- a/Lib/fontTools/ttLib/__init__.py
+++ b/Lib/fontTools/ttLib/__init__.py
@@ -203,8 +203,11 @@ class TTFont(object):
 			# assume "file" is a writable file object
 			closeStream = False
 
-		if self.recalcTimestamp and 'head' in self:
-			self['head']  # make sure 'head' is loaded so the recalculation is actually done
+		if 'head' not in self:
+			raise TTLibError("missing required table: head")
+		elif self.recalcTimestamp and not self.isLoaded('head'):
+			# make sure 'head' is loaded so the recalculation is actually done
+			self['head']
 
 		tags = list(self.keys())
 		if "GlyphOrder" in tags:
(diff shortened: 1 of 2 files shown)


justvanrossum:
There were some test cases without `head`, so those started failing if I didn't check for the presence of `head`.

anthrotype:
Yes, the woff2_test was expecting a TTLibError when head is missing, I'm raising the same error in TTFont.save now.

Lib/fontTools/ttLib/__init__.py
```diff
@@ -203,8 +203,11 @@ def save(self, file, reorderTables=True):
 			# assume "file" is a writable file object
 			closeStream = False
 
-		if self.recalcTimestamp and 'head' in self:
-			self['head']  # make sure 'head' is loaded so the recalculation is actually done
+		if 'head' not in self:
+			raise TTLibError("missing required table: head")
```
behdad: No, please don't do this. It makes it hard to build partial fonts.

anthrotype: It will simply crash later on. Other tables make the assumption head will be there...

anthrotype: I wasn't sure either, TBH. This was easier than going through all the places where head is implicitly required and fix them so they continue without it. Maybe that's the right approach..

behdad:
> It will simply crash later on. Other tables make the assumption head will be there...

What if my font doesn't have those "other tables" (yet)?

> I wasn't sure either, TBH. This was easier than going through all the places where head is implicitly
> required and fix them so they continue without it. Maybe that's the right approach..

What was even easier would have been to not touch it at all. If it fails later, so be it. Only if recalc timestamp was requested and head table is present it should be loaded. The rest of the logic should remain untouched.

anthrotype:
Ok, I'm going to revert it.

PR #857, opened 2017-02-23
varLib. Use designspace <axes> element to define the axisMap and order.

readroberts:
Here is a method to use the designspace <axes> element  to define the axis map, and preserve the axis order. If the <axes> element does not exist, still uses the standard_map, but with the order preserved.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -29,6 +29,7 @@ from fontTools.ttLib.tables._g_v_a_r import TupleVariation
 from fontTools.ttLib.tables import otTables as ot
 from fontTools.varLib import builder, designspace, models
 from fontTools.varLib.merger import VariationMerger
+import collections
 import warnings
 import os.path
 import logging
@@ -59,7 +60,9 @@ def _add_fvar(font, axes, instances, axis_map):
 	font['fvar'] = fvar = newTable('fvar')
 	nameTable = font['name']
 
-	for iden in sorted(axes.keys(), key=lambda k: axis_map[k][0]):
+	for iden in axis_map.keys():
+		if not axes.has_key(iden):
+			continue
 		axis = Axis()
 		axis.axisTag = Tag(axis_map[iden][0])
 		axis.minValue, axis.defaultValue, axis.maxValue = axes[iden]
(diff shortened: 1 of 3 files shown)


Lib/fontTools/varLib/__init__.py
```diff
@@ -59,7 +60,9 @@ def _add_fvar(font, axes, instances, axis_map):
 	font['fvar'] = fvar = newTable('fvar')
 	nameTable = font['name']
 
-	for iden in sorted(axes.keys(), key=lambda k: axis_map[k][0]):
+	for iden in axis_map.keys():
+		if not axes.has_key(iden):
```
anthrotype: You need to use `in` instead of `has_key`, as that no longer exists in py3

Lib/fontTools/varLib/__init__.py
```diff
@@ -29,6 +29,7 @@
 from fontTools.ttLib.tables import otTables as ot
 from fontTools.varLib import builder, designspace, models
 from fontTools.varLib.merger import VariationMerger
+import collections
```
miguelsousa: This is not used downstream

Lib/fontTools/varLib/designspace.py
```diff
@@ -1,5 +1,6 @@
 """Rudimentary support for loading MutatorMath .designspace files."""
 from __future__ import print_function, division, absolute_import
+import collections
```
miguelsousa: Wouldn't it be better to do `from collections import OrderedDict` ?

anthrotype: I don't think it would make any difference.

The designspace_test.py also needs to be modified (that's why CI are failing).

anthrotype: I guess if you use `OrderedDict` more than once, you can spare an extra attribute lookup if you import `from`, as Miguel suggested.

anthrotype:
@behdad @brawer could we merge this as is, and then do any fix-ups?

behdad:
> @behdad @brawer could we merge this as is, and then do any fix-ups?

behdad:
> @behdad @brawer could we merge this as is, and then do any fix-ups?

Yes please.

anthrotype:
@behdad ok then you can merge it. The CI will still fail as the tests needs to be adjusted.

miguelsousa:
@anthrotype I can take care of the fixes

anthrotype:
Thanks Read and Miguel.

PR #858, opened 2017-02-24
[varLib] Build avar; build fvar with spec-compliant value ranges

brawer:
This is a work in progress which hasn't been tested. Also, it assumes
that the interpolation masters have correct usWidthClass and usWeightClass
in their OS/2 table, which is not the case for the current Noto Sans fonts.
Part of the problem is in glyphsLib, and part of it seems in the font
sources; both can be fixed.

https://github.com/googlei18n/fontmake/issues/167

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -29,6 +29,7 @@ from fontTools.ttLib.tables._g_v_a_r import TupleVariation
 from fontTools.ttLib.tables import otTables as ot
 from fontTools.varLib import builder, designspace, models
 from fontTools.varLib.merger import VariationMerger
+import collections
 import warnings
 import os.path
 import logging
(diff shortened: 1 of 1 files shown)


brawer:
@behdad, mind having a look at this pull request? It’s very much a work in progress, trying to generate spec-compliant `fvar` tables. Also, this produces `avar` tables, but I’m not sure yet if that approach actually works; this definitely needs to be tested. For the time being, I’ve hardcoded the instance-specific interpolation parameters for [NotoSansArabic-MM.glyphs](https://github.com/googlei18n/noto-source/raw/master/src/NotoSansArabic-MM.glyphs), which of course can’t be done in the final version. Note that for NotoSansArabic, current glyphsLib generates masters that have the wrong width and weight class in their `OS/2` table (and also the wrong name). For example, it produces a “Bold” master which is actually Black, but there’s no way to tell from the master. I need to find out whether this is because of a glyphsLib bug (there’s a few places in glyphsLib that look suspicious), or because of a bug in the design sources, or both.

brawer:
Turns out that the “Bold” master being Black is intentional; see [“Names and weight values of masters”](https://forum.glyphsapp.com/t/names-and-weight-values-of-masters/1245) in the Glyphs forum. On masters, the names and weight/width values are purely cosmetic and one cannot infer anything from them.

Personally, I find this extremely confusing. For example, when converting a Glyphs file to UFO, one would clearly expect “Myfont-Bold.ufo” to contain the Bold weight. But it doesn‘t. Instead, it contains Black, renamed to Bold (and marked as `OS/2.usWeightClass` 700) simply because Glyphs doesn’t have a UI widget for marking a master as Black. Looking at the Glyphs forum, I’m not the first person who got seriously confused by this.

In the context of this pull request, this means varLib must absolutely _not_ look at the `OS/2` table of the interpolation masters. Will change the code.

anthrotype:
Oops.. It looks like there are some merge conflicts here after merging #857 and #864.

brawer:
@behdad, so what do you think? Specifically:

* regarding `fvar`, should varLib take the min/max/default value for each axis from the interpolation masters (as in this draft pull request), or should they rather go into the `axes` element in the designspace file? The latter would probably be cleaner.

* regarding `avar`, is this the right approach of building the mapping table? If so, how to pass the mapping to varLib — via a side channel (as in this draft pull request), or via the designspace file? The latter would probably be cleaner, but should the mapping from font user space (eg. 0–1000 for weight) to interpolation space (eg. stem width 12–96 for weight) happen in varLib, or in glyphsLib where the designspace is being constructed? In other words, should the `mapping` element in the designspace file be in font user space, or in interpolation space, or in the –1.0..+1.0 range? Tentatively I’d say “interpolation space”, but you understand this stuff better than me.

Only after making the initial draft for this change, I found out that the interpolation masters in Glyphs don’t have any meaningful properties beyond the letterforms. Perhaps that’s an argument for computing all the information inside glyphsLib, since it would contain the “mess” to the place where it’s caused.

Looking forward to your feedback...

brawer:
Instead of this approach, we’ll emit `<axes>` as per the new, clarified designspace document format. See [specification](https://github.com/LettError/designSpaceDocument).

PR #861, opened 2017-02-24
TupleVariation: round deltas before encoding

moyogo:
Python 3 was raising 'struct.error: required argument is not an integer' and Python 2 was truncating when deltas are floats


anthrotype:
good catch!
I'd be nice to add a small test to `TupleVariation_test.py` as well.

anthrotype:
LGTM

Issue #859, opened 2017-02-24
pyftmerge: gasp table is optional, but pyftmerge uses 'equal' logic to merge it


zjusbo:
**Steps to reproduce:**
1. merge a font with gasp and a font without gasp, you will get the error attached at the end of this post.

**Expected result:**
Two fonts are merged successfully.

**Reason:**
[gasp](https://www.microsoft.com/typography/otspec/otff.htm#otttables) is optional, but pyftmerge uses logic 'equal' when merging it. 'equal' requires all fonts contain such table tag.

ttLib.getTableClass('gasp').mergeMap = {
	'tableTag': equal,
	'version': max,
	'gaspRange': first, # FIXME? Appears irreconcilable
}

**Solution:**

1. Change equal to 'firstImplemented'
2. Create 'firstImplemented' logic that returns the first non NotImplemented element in the list.

**Error stack**
(dev) $ pyftmerge noto-fonts/unhinted/NotoSans-Regular.ttf NotoSansCham-Regular.ttf 
Traceback (most recent call last):
  File "/Users/bosong/i18n/fonttools/dev/bin/pyftmerge", line 11, in <module>
    load_entry_point('fonttools', 'console_scripts', 'pyftmerge')()
  File "/Users/bosong/i18n/fonttools/Lib/fontTools/misc/loggingTools.py", line 372, in wrapper
    return func(*args, **kwds)
  File "/Users/bosong/i18n/fonttools/Lib/fontTools/merge.py", line 1011, in main
    font = merger.merge(args)
  File "/Users/bosong/i18n/fonttools/Lib/fontTools/merge.py", line 881, in merge
    table = clazz(tag).merge(self, tables)
  File "/Users/bosong/i18n/fonttools/Lib/fontTools/merge.py", line 159, in merge
    return m.mergeObjects(self, self.mergeMap, tables)
  File "/Users/bosong/i18n/fonttools/Lib/fontTools/merge.py", line 926, in mergeObjects
    value = mergeLogic(getattr(table, key, NotImplemented) for table in tables)
  File "/Users/bosong/i18n/fonttools/Lib/fontTools/merge.py", line 48, in equal
    assert all(item == first for item in t), "Expected all items to be equal: %s" % lst
AssertionError: Expected all items to be equal: **['gasp', NotImplemented]**

anthrotype:
[...]

PR #862, opened 2017-02-24
Change gasp merge logic to 'first'

zjusbo:
Reference #859 
Still open to other ideas.

### DIFF
--- a/Lib/fontTools/merge.py
+++ b/Lib/fontTools/merge.py
@@ -349,7 +349,7 @@ def merge(self, m, tables):
 ttLib.getTableClass('prep').mergeMap = lambda self, lst: first(lst)
 ttLib.getTableClass('fpgm').mergeMap = lambda self, lst: first(lst)
 ttLib.getTableClass('cvt ').mergeMap = lambda self, lst: first(lst)
-ttLib.getTableClass('gasp').mergeMap = lambda self, lst: first(lst)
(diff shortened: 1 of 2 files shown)


zjusbo:
The test fails due to https://github.com/fonttools/fonttools/pull/857

behdad:
I'd say `onlyExisting(equal)` is better.

zjusbo:
I guess you are suggesting:

ttLib.getTableClass('gasp').mergeMap = {
	'tableTag': onlyExisting(equal),
	'version': onlyExisting(max), # max(NotImplemented, 1) returns NotImplemented
	'gaspRange': onlyExisting(first), # FIXME? Appears irreconcilable
}

The difference between these two options happens in this case

fontA['gasp'] = NotImplemented
fontB['gasp'] = a gasp table that is designed for glyphs in fontB

**pyftmerge fontA, fontB**

Your option returns the gasp in fontB while current option drops gasp and returns nothing. 

I think the root problem is: does a gasp table designed for fontB probably hurt the legibility of fontA?

If so, probably dropping the gasp in fontB is a better choice? Actually, default rules to decide how to render the glyphs on grayscale devices are used if the gasp table is absent. Using gasp in fontB implicitly drops the default rules in fontA.

behdad:
Actually, since we retain hinting from first don't only, this also has to be first.  Lgtm

PR #864, opened 2017-02-26
Followup to PR #857

miguelsousa:


Lib/fontTools/varLib/designspace.py
```diff
@@ -66,11 +77,14 @@ def _load(et):
...
 		instances.append(_loadItem(et))
 
-	return masters, instances, axisMap
+	return axes, masters, instances
 
 def load(filename):
-	"""Load designspace from a file name or object.  Returns two items:
-	list of masters (aka sources) and list of instances."""
+	"""Load designspace from a file name or object.
+	   Returns three items:
```
anthrotype: Thanks Miguel. Talking with @behdad yesterday, we were thinking of having the load and _load function return a dictionary instead of a tuple of two, now three (and tomorrow who knows how many) items. Could you also  make that change here?

miguelsousa: Yes, good idea. So you want it to return something like this
```python
{'axes': ..., 'masters': ..., 'instances': ...}
```
yes?

anthrotype: Yes!

anthrotype: Mind also adding a changelog entry about support for axes element and the new return value for designspace load? Thank you.

miguelsousa: done!

behdad:
So, if file doesn't have axes element, shouldn't we return None? Same with masters and instances.

miguelsousa:
@behdad good point. So return None if the element is not in the file, and an empty list if parsing the element yielded no results?

anthrotype:
Instead of None, I would prefer that the key/value pair is not set in the returned dict if one element is not present in the designspace.

miguelsousa:
Just realized that wrapping `axis_map.keys()` with `list()` would have fix it for py3 🤦
```python
for iden in sorted(axes.keys(), key=lambda i: list(axis_map.keys()).index(i)):
```
I still think that this is a better solution because it clearly shows that we're iterating using the contents of `axes`.

anthrotype:
well, it's the same.. you said `axes` has same keys or a subset of `axis_map`, and `axis_map` is already an ordereddict.

PR #865, opened 2017-02-27
[varLib/designspace] fixups

anthrotype:
mostly cosmetic, two changes deserve mention:

- if any of `axes`, `instances` or `sources` element is not present, the dictionary returned from `designspace.load` will not contain the respective key/value pair; an empty list means the element is defined but empty;

- what was previously called `masters` is now called `sources` for consistency with the designspace document specification.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -37,6 +37,10 @@ from pprint import pformat
 
 log = logging.getLogger("fontTools.varLib")
 
+
+class VarLibError(Exception):
+	pass
+
 #
 # Creation routines
 #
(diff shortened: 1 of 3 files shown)


Lib/fontTools/varLib/__init__.py
```diff
@@ -265,9 +269,10 @@ def build(designspace_filename, master_finder=lambda s:s, axisMap=None):
 	"""
 
 	ds = designspace.load(designspace_filename)
-	axes = ds['axes']
-	masters = ds['masters']
-	instances = ds['instances']
+	axes = ds['axes'] if 'axes' in ds else []
+	if 'sources' not in ds or not ds['sources']:
+		raise VarLibError("no 'sources' defined in .designspace")
+	instances = ds['instances'] if 'instances' in ds else []
```
miguelsousa: @anthrotype You forgot to define the 'masters' variable

anthrotype: Ouch! thanks Miguel. I guess it's time to add some tests to varLib. :)

miguelsousa: also, `varLib.interpolate_layout` needs some of these changes

miguelsousa: BTW, I'm putting together some test cases that will exercise `fontTools.varLib.build()` and `fontTools.varLib.interpolate_layout.interpolate_layout()`. I think these will help a lot validating this and upcoming changes.

anthrotype: That'd be great, thanks a lot!

miguelsousa:
I was getting ready to submit a PR with some of these changes. At this point the only thing salvageable is probably the test case I had to go along with them.
Cosmo, do you have anything else to commit? I just want to make sure we're not doing the same thing.

anthrotype:
Ok, I'm done now!

PR #867, opened 2017-03-01
[tests] Change `hhea` and `vhea` table version from float to hex to mute fontTools warnings

miguelsousa:
Some housekeeping as I just noticed these warnings.


miguelsousa:
@anthrotype OK to merge? It's a pretty trivial change.

miguelsousa:
Thanks Just. Thanks Cosimo.

PR #869, opened 2017-03-01
[varLib.builder] Fix roundtripping XML

anthrotype:
PTAL

### DIFF
--- a/Lib/fontTools/varLib/builder.py
+++ b/Lib/fontTools/varLib/builder.py
@@ -6,7 +6,7 @@ from fontTools.ttLib.tables import otTables as ot
 
 def buildVarRegionAxis(axisSupport):
 	self = ot.VarRegionAxis()
-	self.StartCoord, self.PeakCoord, self.EndCoord = axisSupport
+	self.StartCoord, self.PeakCoord, self.EndCoord = [float(v) for v in axisSupport]
 	return self
 
 def buildVarRegion(support, axisTags):
--- a/Lib/fontTools/varLib/builder.py
+++ b/Lib/fontTools/varLib/builder.py
@@ -57,7 +57,8 @@ def calculateNumShorts(self, optimize=True):
 			items[i] = _reorderItem(items[i], narrows)
 		self.NumShorts = count - len(narrows)
 	else:
-		self.NumShorts = 1 + max(set(range(count)) - narrows)
+		wides = set(range(count)) - narrows
+		self.NumShorts = 1+max(wides) if wides else 0
 	return self
 
 def buildVarData(varRegionIndices, items, optimize=True):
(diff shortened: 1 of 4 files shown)


miguelsousa:
The changes to `varLib_test.py` and `BuildTest.ttx` look good.

anthrotype:
thanks Miguel. Let's see what @behdad says about the changes in `varLib.builder` and `otTables`.

Lib/fontTools/varLib/builder.py
```diff
@@ -68,6 +68,16 @@ def buildVarData(varRegionIndices, items, optimize=True):
...
+	numShorts = 0
+	for item in records:
+		assert len(item) == regionCount, ("Item length mismatch", len(item), regionCount)
+		for i in range(regionCount - 1, numShorts - 1, -1):
+			if not (-128 <= item[i] <= 127):
+				numShorts = i + 1
+				break
+		if numShorts == regionCount:
+			break
+	self.NumShorts = numShorts
```
behdad: This is wrong.  This block should go *after* call to optimizeVarData(). In fact, maybe move code from that function here, and only make the loop that calls _reorderItem() be dependent on "if optimize".

Rest looks great. Thanks!

anthrotype: but `optimizeVarData` is reused later in `OnlineVarStoreBuilder.finish` method..

behdad: > but optimizeVarData is reused later in OnlineVarStoreBuilder.finish method..

You need to set NumShorts there too! So, maybe rename optimizeVarData to calculateNumShorts(), and have an option to optimize or not.

anthrotype: like this? 3f8eab4

anthrotype:
I wonder if `NumShorts` should be commented out in the TTX dump..

behdad:
> I wonder if NumShorts should be commented out in the TTX dump..

No. It's significant part of encoding. I have had made it automatic initially, but then decided it shouldn't be. We don't expect people to edit the VarData XML by hand anyway.

Lib/fontTools/varLib/builder.py
```diff
@@ -51,11 +50,12 @@ def optimizeVarData(self):
 				break
 		if not narrows:
 			break
-
-	self.VarRegionIndex = _reorderItem(self.VarRegionIndex, narrows)
-	for i in range(self.ItemCount):
-		items[i] = _reorderItem(items[i], narrows)
-
+	self.NumShorts = count - len(narrows)
```
behdad: This line is correct if optimize is True. Otherwise, NumShorts should be set to `1+max(set(range(count)) - narrows)`.

behdad: Or you can also track a `wides` set, and use `1+max(wides)`

anthrotype: ops, too late. Already pushed 1883e64

anthrotype:
@behdad this good now?

PR #871, opened 2017-03-03
varLib.interpolate_layout test

miguelsousa:
This is meant to be the beginning of stress-testing the support for GSUB and GPOS merging. Currently it only contains one basic GSUB test. I'd like for these changes to get reviewed and merged before introducing GPOS tests that make merging fail.


anthrotype:
LGTM

PR #876, opened 2017-03-06
[varLib tests] Test for interpolatable.py

miguelsousa:
@anthrotype this test requires an additional package. Is this how to set it up?

### DIFF
--- a/Tests/varLib/interpolatable_test.py
+++ b/Tests/varLib/interpolatable_test.py
@@ -1,0 +1,78 @@
+from __future__ import print_function, division, absolute_import
+from fontTools.misc.py23 import *
+from fontTools.ttLib import TTFont
+from fontTools.varLib.interpolatable import main as interpolatable_main
+import os
+import shutil
+import sys
+import tempfile
+import unittest
+
+
+class InterpolatableTest(unittest.TestCase):
+    def __init__(self, methodName):
+        unittest.TestCase.__init__(self, methodName)
+        # Python 3 renamed assertRaisesRegexp to assertRaisesRegex,
+        # and fires deprecation warnings if a program uses the old name.
+        if not hasattr(self, "assertRaisesRegex"):
+            self.assertRaisesRegex = self.assertRaisesRegexp
+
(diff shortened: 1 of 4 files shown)


Tests/varLib/interpolatable_test.py
```diff
@@ -0,0 +1,78 @@
...
+        suffix = '.ttf'
+        ttx_dir = self.get_test_input('master_ttx_interpolatable_ttf')
+
+        self.temp_dir()
+        ttx_paths = self.get_file_list(ttx_dir, '.ttx', 'TestFamily2-')
+        for path in ttx_paths:
+            self.compile_font(path, suffix, self.tempdir)
+
+        ttf_paths = self.get_file_list(self.tempdir, suffix)
+        self.assertIsNone(interpolatable_main(ttf_paths))
```
anthrotype: but.. what is this testing? Simply that `main` completes?
you know, I think it's a bit premature to add tests for this module, it's still very much a work in progress, Behdad may wish to complete it before we sanction the current API with tests like these.

miguelsousa: It's testing that it completed without errors. 
OK, I'll stash it for later.

behdad:
> but.. what is this testing? Simply that main completes?

That would be useful test still.

> you know, I think it's a bit premature to add tests for this module, it's still very much a work in progress, Behdad may wish to complete it before we sanction the current API with tests like these.

That's definitely true, but so is all of varLib!

behdad:
> before we sanction the current API with tests like these.

FWIW I never intended for `varLib.interpolate_layout` to remain a module. It was just where I put the code and one day fontmake started using it. I understand working in branch and what not, but when something takes a year to finish on and off, branches are not that helpful.

miguelsousa:
Cosimo, can you take what I submitted and make the necessary changes? Thank you.

anthrotype:
ok, some tests

anthrotype:
ok, having some tests is always better than having none :)

PR #877, opened 2017-03-07
[requires.io] dependency update on master branch

anthrotype:


anthrotype:
hm., don't know where this is coming from, I don't remember having authorized requires.io to do that.. One update bot is enough.

Issue #829, opened 2017-02-02
Add ability for parser to handle some comments


mhosken:
When processing fea files, particularly when generating fea, it would be good to be able to process some comments. Some applications include processing information in comments and it would be good to be able to see them. When writing a fea to fea type processor, it would be good to pass comments through.

Arbitrary handling of comments is hard, and therefore I propose that we only support comments in certain locations (between statements).

I am happy to do the work.

anthrotype:
That would be nice, thanks for working on this!

brawer:
Adding comments to the syntax tree sounds like a good idea. Thanks for doing it.

PR #879, opened 2017-03-08
Add support for comments to parser

mhosken:
This addresses issue #829. All comments end up on a line of their own and there are plenty of places where comments are still not allowed. But I hope this does support all the sensible places where people might put comments. The only ones of interest that may get missed are:

- Inside classes
- Within a rule
- Within lists of values for some tables

I.e. comments are only supported between statements.

### DIFF
--- a/Lib/fontTools/feaLib/lexer.py
+++ b/Lib/fontTools/feaLib/lexer.py
@@ -5,6 +5,7 @@ from fontTools.feaLib.error import FeatureLibError
 import re
 import os
 
+
 class Lexer(object):
     NUMBER = "NUMBER"
     FLOAT = "FLOAT"
--- a/Lib/fontTools/feaLib/lexer.py
+++ b/Lib/fontTools/feaLib/lexer.py
@@ -5,7 +5,6 @@ from fontTools.feaLib.error import FeatureLibError
 import re
 import os
 
-
 class Lexer(object):
     NUMBER = "NUMBER"
     FLOAT = "FLOAT"
@@ -33,6 +32,9 @@ class Lexer(object):
     MODE_NORMAL_ = "NORMAL"
     MODE_FILENAME_ = "FILENAME"
 
+    _OMITS = {NEWLINE}
+    _OMITC = {NEWLINE, COMMENT}
+
     def __init__(self, text, filename):
         self.filename_ = filename
         self.line_ = 1
@@ -45,13 +47,13 @@ class Lexer(object):
     def __iter__(self):
         return self
 
-    def next(self):  # Python 2
-        return self.__next__()
+    def next(self, comments = False):  # Python 2
+        return self.__next__(comments)
 
-    def __next__(self):  # Python 3
+    def __next__(self, comments = False):  # Python 3
         while True:
             token_type, token, location = self.next_()
-            if token_type not in {Lexer.COMMENT, Lexer.NEWLINE}:
+            if token_type not in (Lexer._OMITS if comments else Lexer._OMITC):
                 return (token_type, token, location)
 
     def location_(self):
@@ -193,14 +195,14 @@ class IncludingLexer(object):
     def __iter__(self):
         return self
 
-    def next(self):  # Python 2
-        return self.__next__()
+    def next(self, comments = False):  # Python 2
+        return self.__next__(comments)
 
-    def __next__(self):  # Python 3
+    def __next__(self, comments = False):  # Python 3
         while self.lexers_:
             lexer = self.lexers_[-1]
             try:
-                token_type, token, location = lexer.next()
+                token_type, token, location = lexer.next(comments)
             except StopIteration:
                 self.lexers_.pop()
                 continue
--- a/Lib/fontTools/feaLib/lexer.py
+++ b/Lib/fontTools/feaLib/lexer.py
@@ -32,8 +32,8 @@ class Lexer(object):
     MODE_NORMAL_ = "NORMAL"
     MODE_FILENAME_ = "FILENAME"
 
-    _OMITS = {NEWLINE}
-    _OMITC = {NEWLINE, COMMENT}
+    _OMITMINIMUM = {NEWLINE}
+    _OMITCOMMENTS = {NEWLINE, COMMENT}
 
     def __init__(self, text, filename):
         self.filename_ = filename
@@ -47,13 +47,13 @@ class Lexer(object):
     def __iter__(self):
         return self
 
-    def next(self, comments = False):  # Python 2
+    def next(self, comments=False):  # Python 2
         return self.__next__(comments)
 
-    def __next__(self, comments = False):  # Python 3
+    def __next__(self, comments=False):  # Python 3
         while True:
             token_type, token, location = self.next_()
-            if token_type not in (Lexer._OMITS if comments else Lexer._OMITC):
+            if token_type not in (Lexer._OMITMINIMUM if comments else Lexer._OMITCOMMENTS):
                 return (token_type, token, location)
 
     def location_(self):
@@ -195,10 +195,10 @@ class IncludingLexer(object):
     def __iter__(self):
         return self
 
-    def next(self, comments = False):  # Python 2
+    def next(self, comments=False):  # Python 2
         return self.__next__(comments)
 
-    def __next__(self, comments = False):  # Python 3
+    def __next__(self, comments=False):  # Python 3
         while self.lexers_:
             lexer = self.lexers_[-1]
             try:
--- a/Lib/fontTools/feaLib/lexer.py
+++ b/Lib/fontTools/feaLib/lexer.py
@@ -32,9 +32,6 @@ class Lexer(object):
     MODE_NORMAL_ = "NORMAL"
     MODE_FILENAME_ = "FILENAME"
 
-    _OMITMINIMUM = {NEWLINE}
-    _OMITCOMMENTS = {NEWLINE, COMMENT}
-
     def __init__(self, text, filename):
         self.filename_ = filename
         self.line_ = 1
@@ -47,13 +44,13 @@ class Lexer(object):
     def __iter__(self):
         return self
 
-    def next(self, comments=False):  # Python 2
-        return self.__next__(comments)
+    def next(self):  # Python 2
+        return self.__next__()
 
-    def __next__(self, comments=False):  # Python 3
+    def __next__(self):  # Python 3
         while True:
             token_type, token, location = self.next_()
-            if token_type not in (Lexer._OMITMINIMUM if comments else Lexer._OMITCOMMENTS):
+            if token_type != Lexer.NEWLINE:
                 return (token_type, token, location)
 
     def location_(self):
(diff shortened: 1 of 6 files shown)


Lib/fontTools/feaLib/parser.py
```diff
@@ -1339,14 +1363,24 @@ def expect_string_(self):
...
-            (self.next_token_type_, self.next_token_,
-             self.next_token_location_) = self.lexer_.next()
-        except StopIteration:
-            self.next_token_type_, self.next_token_ = (None, None)
+    def advance_lexer_(self, comments = False):
+        if not self.ignore_comments and comments and len(self.cur_comments_):
+            self.cur_token_type_ = Lexer.COMMENT
+            self.cur_token_, self.cur_token_location_ = self.cur_comments_.pop(0)
+            return
+        else
```
brawer: Missing colon (see test failures)

Lib/fontTools/feaLib/lexer.py
```diff
@@ -33,6 +32,9 @@ class Lexer(object):
     MODE_NORMAL_ = "NORMAL"
     MODE_FILENAME_ = "FILENAME"
 
+    _OMITS = {NEWLINE}
```
brawer: The name is a bit cryptic.

Lib/fontTools/feaLib/lexer.py
```diff
@@ -45,13 +47,13 @@ def __init__(self, text, filename):
...
-        return self.__next__()
+    def next(self, comments = False):  # Python 2
+        return self.__next__(comments)
 
-    def __next__(self):  # Python 3
+    def __next__(self, comments = False):  # Python 3
         while True:
             token_type, token, location = self.next_()
-            if token_type not in {Lexer.COMMENT, Lexer.NEWLINE}:
+            if token_type not in (Lexer._OMITS if comments else Lexer._OMITC):
```
brawer: This might be a bit easier to read? Then you don’t need such cryptic constants.

```python
if token_type == COMMENT and not comments:
    continue
if token_type == Lexer.NEWLINE:
    continue
```

Or perhaps:

```python
ignore = ((token_type == Lexer.NEWLINE) or
          (token_type == Lexer.COMMENT and not comments))
if not ignore:
    return (token_type, token, location)
```

Lib/fontTools/feaLib/parser.py
mhosken: python colons. Syntactic sugar that should be optional, especially designed to frustrate just me! :)

Lib/fontTools/feaLib/lexer.py
```diff
@@ -193,14 +195,14 @@ def __init__(self, featurefile):
     def __iter__(self):
         return self
 
-    def next(self):  # Python 2
-        return self.__next__()
+    def next(self, comments = False):  # Python 2
```
brawer: Style nitpick from [PEP8](https://www.python.org/dev/peps/pep-0008/#other-recommendations): “Don’t use spaces around the = sign when used to indicate a keyword argument or a default parameter value”.

Tests/feaLib/builder_test.py
```diff
@@ -154,6 +156,7 @@ def check_fea2fea_file(self, name, base=None, parser=Parser):
                 sys.stderr.write(line+"\n")
             self.fail("Fea2Fea output is different from expected. "
                       "Generated:\n{}\n".format("\n".join(actual)))
+        parser.ignore_comments = temp
```
anthrotype: a `try: ... finally: ...`, or a [context manager](https://docs.python.org/3.6/library/contextlib.html#contextlib.contextmanager) could be used here to temporarily patch the `parser` module and ensure it is restored to the original state, whether or not exceptions are raised within the block.

Lib/fontTools/feaLib/lexer.py
```diff
@@ -5,7 +5,6 @@
 import re
 import os
 
-
```
anthrotype: no need to remove this empty line

Lib/fontTools/feaLib/lexer.py
mhosken: I chose the route I did for speed. But I'm happy to go slow and pretty if you like.

Tests/feaLib/builder_test.py
mhosken: No because then you are introducing a redundant exception capture, which would require re-raising any exceptions raised in the code inside. Just go with a temporary variable.

Lib/fontTools/feaLib/lexer.py
mhosken: Correct, and there is no need for it. Is is important to have 3 empty lines there? Is that part of your coding standard.

anthrotype: two empty lines should be fine.

Lib/fontTools/feaLib/lexer.py
mhosken: wow yes very nit picky. I hadn't realise just how nit picky for PEP8 this code base is. Perhaps we need a PEP8 reviewer in the continuous integration testing. I think there is one around somewhere.

Tests/feaLib/builder_test.py
anthrotype: a `try: ... finally: ...` block will _not_ capture exceptions but let them propagate; it will only ensure that `finally:` statement is always called whether exceptions are raised or not. If one wishes to handle an exception, one needs to use `except:`.

the temporary variable is risky because if anything goes wrong (the test errors or fails) it will change the global state of the module.

Lib/fontTools/feaLib/lexer.py
anthrotype: one of the reason we didn't add an automatic PEP8 checker (e.g. flake8) to the CI is because unfortunately half of the code base will not pass the test... :)
But we try to ensure that new code which is added follows it as far as possible, without being overly strict. The spaces around the equal sign look a bit odd to me as well.

Lib/fontTools/feaLib/lexer.py
```diff
@@ -45,13 +47,13 @@ def __init__(self, text, filename):
     def __iter__(self):
         return self
 
-    def next(self):  # Python 2
-        return self.__next__()
+    def next(self, comments=False):  # Python 2
```
anthrotype: an `iterator.__next__()` method cannot not take any extra argument (besides `self`):
https://docs.python.org/3.6/library/stdtypes.html#iterator.__next__

mhosken: I have the following text in my rendering of that page:


iterator.__next__()

    Return the next item from the container. If there are no further items, raise the StopIteration exception. This method corresponds to the tp_iternext slot of the type structure for Python objects in the Python/C API.

No mention of extra parameters, which would obviously have to be named parameters. It also doesn't mention that in normal cases you never call the __next__ method directly.

I really don't want to go down the road of constantly switching the state of the Lexer. Hmm, in the case of the parser, if I did that, then I would turn on passing comments, and perhaps not have the option to turn it off. Is that preferred?

anthrotype: > No mention of extra parameters

sorry, I miss-typed `cannot not`, I meant `cannot`: no extra parameters are allowed in the iterator API.

anthrotype: > in normal cases you never call the next method directly

the `__next__()` method is called upon by the `next()` built-in function
https://docs.python.org/3/library/functions.html#next

Lib/fontTools/feaLib/parser.py
```diff
@@ -787,9 +792,11 @@ def parse_table_(self):
 
     def parse_table_GDEF_(self, table):
         statements = table.statements
-        while self.next_token_ != "}":
-            self.advance_lexer_()
-            if self.is_cur_keyword_("Attach"):
+        while self.next_token_ != "}" or (not self.ignore_comments and len(self.cur_comments_)):
```
brawer: Actually, does the parser really need to have a special mode for ignoring comments? If it always recognized comments, and always added them to the syntax tree, the API would be simpler. One less option to worry about.

mhosken: OK Let's do that then. In answer to @anthrotype, yes the number of fixed position parameters to any function is, well, fixed. But you can have any number of named parameters. There is nothing special about next() in that respect.

anthrotype: >  you can have any number of named parameters

only that you won't be able to pass those to the `next()` built-in function, which only takes the iterator object.

Lib/fontTools/feaLib/lexer.py
```diff
@@ -195,14 +192,14 @@ def __init__(self, featurefile):
...
+    def next(self):  # Python 2
+        return self.__next__()
 
-    def __next__(self, comments=False):  # Python 3
+    def __next__(self):  # Python 3
         while self.lexers_:
             lexer = self.lexers_[-1]
             try:
-                token_type, token, location = lexer.next(comments)
+                token_type, token, location = lexer.next()
```
anthrotype: Could we change all occurrence of `lexer.next()` with `next(lexer)`?
I know, it works in both cases,  but using the latter idiom would make it appear less legacy python 2.

/cc @brawer 

```
$ git grep "\.next()"
Lib/fontTools/feaLib/lexer.py:202:                token_type, token, location = lexer.next()
Lib/fontTools/feaLib/lexer.py:207:                fname_type, fname_token, fname_location = lexer.next()
Lib/fontTools/feaLib/lexer.py:210:                #semi_type, semi_token, semi_location = lexer.next()
Lib/fontTools/feaLib/parser.py:1379:                 self.next_token_location_) = self.lexer_.next()
Lib/fontTools/voltLib/parser.py:598:             self.next_token_location_) = self.lexer_.next()
```

Lib/fontTools/feaLib/parser.py
```diff
@@ -10,13 +10,15 @@
 import re
 
 
+
```
anthrotype: sorry to be such a pain.. but maybe you could set your text editor to not add/remove empty lines at random?

Lib/fontTools/feaLib/parser.py
```diff
@@ -10,13 +10,15 @@
...
 log = logging.getLogger(__name__)
 
 
 class Parser(object):
 
     extensions = {}
     ast = ast
+    ignore_comments=True
```
anthrotype: when the equal sign is used in assignments, it usually is surrounded by spaces
(to be clear, I don't enjoy playing the pep8-er).

Lib/fontTools/feaLib/parser.py
mhosken: I put the line back because you asked me. My editor does what I ask it :) So do you want that line in there or not?

anthrotype: no, I asked you to not remove line 8 of lexer.py; but you kept the removal, and instead added a new empty line in `parser.py`...
Anyway, not worth arguing about these minor things :)
I simply don't want random whitespace changes to distract from the content of the patch.

anthrotype:
changes LGTM, thanks!

mhosken:
> no, I asked you to not remove line 8 of lexer.py; but you kept the removal, and instead added a new empty line in `parser.py`...
> Anyway, not worth arguing about these minor things :)
> I simply don't want random whitespace changes to distract from the content of the patch.

Please bear in mind that you and I have spent waaaay more timing sending emails back and forth about this one minor whitespace niggle than either of us would have spent if we had ignored the issue and read through it. There is no need to make the perfect patch given there are 3 capable people reading it. Yes if it's really ugly, then fine. But for one redundant change?

All changes to changes have costs.

anthrotype:
The Travis macOS build is taking more than 3 hours to complete...
I'll merge this, since it's gonna pass anyway.

Lib/fontTools/feaLib/parser.py
brawer: >> Actually, does the parser really need to have a special mode for ignoring comments? 

> OK Let's do that then.

Done in https://github.com/fonttools/fonttools/pull/882.

PR #882, opened 2017-03-09
[feaLib] Simplify parser API by removing the `ignore_comments` option

brawer:
https://github.com/fonttools/fonttools/pull/879#discussion_r104947968
https://github.com/fonttools/fonttools/issues/829


mhosken:
I'm OK with always parsing in comments into the AST. This assumes that everyone is OK with always having comments in the AST. It makes it harder to parse the AST for builder actions, but perhaps that isn't a value.

PR #881, opened 2017-03-09
[varLib.interpolate_layout] Tests for GPOS LookupType 2 class pairs

miguelsousa:
Merging of class kerning pairs seems to be broken.
```python
Traceback (most recent call last):
  File "interpolate_layout_test.py", line 425, in test_varlib_interpolate_layout_GPOS_only_LookupType_2_class_pairs_same_val_ttf
    instfont = interpolate_layout(ds_path, {'weight': 500}, finder)
  File "fonttools/Lib/fontTools/varLib/interpolate_layout.py", line 85, in interpolate_layout
    merger.mergeTables(font, master_fonts, axes_dict, base_idx, ['GPOS'])
  File "fonttools/Lib/fontTools/varLib/merger.py", line 107, in mergeTables
    self.mergeThings(font[tag], [m[tag] for m in master_ttfs])
  File "fonttools/Lib/fontTools/varLib/merger.py", line 93, in mergeThings
    self.mergeObjects(out, lst)
  File "fonttools/Lib/fontTools/varLib/merger.py", line 70, in mergeObjects
    mergerFunc(self, value, values)
  File "fonttools/Lib/fontTools/varLib/merger.py", line 93, in mergeThings
    self.mergeObjects(out, lst)
  File "fonttools/Lib/fontTools/varLib/merger.py", line 70, in mergeObjects
    mergerFunc(self, value, values)
  File "fonttools/Lib/fontTools/varLib/merger.py", line 93, in mergeThings
    self.mergeObjects(out, lst)
  File "fonttools/Lib/fontTools/varLib/merger.py", line 70, in mergeObjects
    mergerFunc(self, value, values)
  File "fonttools/Lib/fontTools/varLib/merger.py", line 95, in mergeThings
    self.mergeLists(out, lst)
  File "fonttools/Lib/fontTools/varLib/merger.py", line 80, in mergeLists
    self.mergeThings(value, values)
  File "fonttools/Lib/fontTools/varLib/merger.py", line 91, in mergeThings
    mergerFunc(self, out, lst)
  File "fonttools/Lib/fontTools/varLib/merger.py", line 531, in merge
    merger.mergeLists(self.SubTable, subtables)
  File "fonttools/Lib/fontTools/varLib/merger.py", line 80, in mergeLists
    self.mergeThings(value, values)
  File "fonttools/Lib/fontTools/varLib/merger.py", line 91, in mergeThings
    mergerFunc(self, out, lst)
[...]

PR #885, opened 2017-03-10
[CFF2] Support ‘maxstack’

miguelsousa:
A late patch from @readroberts.

### DIFF
--- a/Lib/fontTools/cffLib.py
+++ b/Lib/fontTools/cffLib.py
@@ -26,6 +26,9 @@ cffHeaderFormat = """
 	hdrSize: B
 """
 
+defaultMaxStack = 193
+maxStackLimit = 513
+
 class CFFFontSet(object):
 
 	def __init__(self):
@@ -414,6 +417,7 @@ class IndexedStringsCompiler(IndexCompiler):
 
 
 class TopDictIndexCompiler(IndexCompiler):
+	maxOpStack = defaultMaxStack
 
 	def getItems(self, items, strings):
 		out = []
@@ -497,6 +501,13 @@ class SubrsCompiler(GlobalSubrsCompiler):
 		self.parent.rawDict["Subrs"] = offset
 
 class CharStringsCompiler(GlobalSubrsCompiler):
+	def getItems(self, items, strings):
+		out = []
+		for cs in items:
+			cs.compile()
+			out.append(cs.bytecode)
+		return out
+
 	def setPos(self, pos, endPos):
 		self.parent.rawDict["CharStrings"] = pos
 
@@ -1696,6 +1707,7 @@ class ROSConverter(SimpleConverter):
 
 topDictOperators = [
 #	opcode		name			argument type	default	converter
+	(25,		'maxstack',		'number',	None,	None),
 	((12, 30),	'ROS',	('SID', 'SID', 'number'),	None,	ROSConverter()),
 	((12, 20),	'SyntheticBase',	'number',	None,	None),
 	(0,		'version',		'SID',		None,	None),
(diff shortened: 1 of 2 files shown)


Lib/fontTools/cffLib.py
```diff
@@ -414,6 +417,7 @@ def getItems(self, items, strings):
 
 
 class TopDictIndexCompiler(IndexCompiler):
+	maxOpStack = 193
```
anthrotype: @miguelsousa where is this attribute used? why is it not not simply `maxOpStack = defaultMaxStack`?

Lib/fontTools/cffLib.py
```diff
@@ -1731,17 +1743,16 @@ def xmlRead(self, name, attrs, content, parent):
 	((12, 36),	'FDArray',		'number',	None,	FDArrayConverter()),
 	(17,		'CharStrings',		'number',	None,	CharStringsConverter()),
 	(24,		'VarStore',		'number',	None,	VarStoreConverter()),
-	(25,		'maxstack',		'number',	None,	None),
```
anthrotype: any specific reason why you moved this `maxstack` operator at the top of this list? just curious

readroberts: Yes.  maxstack needs to be set before any PrivateDict is processed, in order to be able to parse the blend operators correctly. The simplest way to do this is to put maxstack first in the order, although it only needs to be before the FDArray.

Lib/fontTools/cffLib.py
readroberts: It should indeed be maxOpStack = defaultMaxStack.

Lib/fontTools/cffLib.py
anthrotype: I wonder, if this `maxstack` is a CFF2-only operator why is it also listed in `topDictOperators` and not just in `topDictOperatos2`?
I see it's mentioned in CFF2 spec https://www.microsoft.com/typography/otspec/cff2charstr.htm, but don't see any reference to it in the CFF 1.0 spec (Tech Note #5176).

readroberts: This a work-around for a current limitation on my scripts for building OpenType/CFF2 fonts. It is currently convenient to able to convert the default master CFF object to a CFF2 object by doing a bit of field editing and then compiling it. That requires the CFF compiler methods to be able to handle the CFF2 fields. This is messy, and I plan to fix this in the longer run - it is not particularly difficult to do so.

readroberts: More specifically, I plan to work in this in about two weeks, when I will also remove the global data references.

behdad:
> I see it's mentioned in CFF2 spec https://www.microsoft.com/typography/otspec/cff2charstr.htm, but don't see any reference to it in the CFF 1.0 spec (Tech Note #5176).

I should have raised this way before, but I truly don't see why we need a maxstack operator.  I'm sure all implementations could have just used the max maxstack and be fine. Bummer

Lib/fontTools/cffLib.py
miguelsousa: Fixed

anthrotype:
> I should have raised this way before, but I truly don't see why we need a maxstack operator. I'm sure all implementations could have just used the max maxstack and be fine.

Yes, but it's in the official spec by now so I believe we should support it until/unless is officially deprecated.

> I plan to work in this in about two weeks, when I will also remove the global data references.

Thanks @readroberts.

Read, Miguel, you must have been using this updated cffLib for a while internally; do you believe this is stable enough to cut a new fonttools release?

I'd like to push out some varLib bugfixes today so that fontmake use of them. However, I fear that after merging `cff2` branch we may have introduced other regressions in the CFF 1.0 part of the code which we might not be aware of.

I guess it's a bit too late to revert the ``cff2`` merge; so I'm thinking of tagging 3.9.0 and wish that users will spot and report back any issues they may find.

miguelsousa:
@anthrotype it's as stable as it can be for the time being, until @readroberts has time to clean up the code. So I'd say go ahead with a new release. Would be great if the #881 changes were included as well, but it sounds like they will have to wait until they're validated.

Issue #178, opened 2014-09-29
AAT tables support: kern (format > 0), feat, mort, morx, ...)


olivierberten:
Since it is the easiest one, I started with the `feat` table but Apple seems to have added undocumented data between the _Feature name array_ and the _Setting name array_... Has anybody heard about it? The new spec doesn't bring anything new...

behdad:
Supporting mort/morx and more complex kern/kerx tables is a huge task. I don't see a point keeping an issue open for it.

PR #886, opened 2017-03-10
[AAT] Add Char64 converter for 64-byte ASCII character strings

brawer:
Used by Apple Advanced Typography tables, for example `gcid`.
https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6gcid.html

https://github.com/fonttools/fonttools/issues/178

### DIFF
--- a/Lib/fontTools/ttLib/tables/otConverters.py
+++ b/Lib/fontTools/ttLib/tables/otConverters.py
@@ -306,6 +306,37 @@ class Version(BaseConverter):
 		return fl2fi(v, 16)
 
 
+class Char64(SimpleValue):
+	"""An ASCII string with up to 64 characters.
+
+	Unused character positions are filled with 0x00 bytes.
+	Used in Apple AAT fonts in the `gcid` table.
+	"""
+	staticSize = 64
+
+	def read(self, reader, font, tableDict):
+		data = reader.readData(self.staticSize)
+		zeroPos = data.find(b"\0")
+		if zeroPos >= 0:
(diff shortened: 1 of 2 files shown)


Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -306,6 +306,37 @@ def fromFloat(v):
 		return fl2fi(v, 16)
 
 
+class Char64(SimpleValue):
```
behdad: I'd think a more specific name is better, given how specific of a type it is.

Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -306,6 +306,37 @@ def fromFloat(v):
...
+	staticSize = 64
+
+	def read(self, reader, font, tableDict):
+		data = reader.readData(self.staticSize)
+		zeroPos = data.find(b"\0")
+		if zeroPos >= 0:
+			data = data[:zeroPos]
+		s = tounicode(data, encoding="ascii", errors="replace")
+		if s != tounicode(data, encoding="ascii", errors="ignore"):
+			log.warning('replaced non-ASCII characters in "%s"' %
```
behdad: Or just fail? Or just write?

Lib/fontTools/ttLib/tables/otConverters.py
brawer: What name would you suggest?

PR #887, opened 2017-03-14
[varLib.merger] Fix interpolating PairPos Format 2

anthrotype:
This is the follow-up of another PR #881 opened by @miguelsousa a few days ago.

Miguel added some tests for `varLib.interpolate_layout`, specifically for GPOS PairPos Format 2 subtables, but these were failing for a number of issues in `varLib.merger` module.

1) `IndexError` was being raised when aligning the first classes in PariPos Format 2, and one of the `ClassDef1` was empty (aad0d46)

2) An extra empty PairPos Format 1 was being inserted all the time, even if the fonts to merge/interpolate didn't have any Format 1 PairPos subtables (d83c2fb)

3) multiple reference to the same mutable `Class2Record` instance were being used and inadvertently modified (e6125b3)

4) the `Format` attribute of `ClassDef1` and `ClassDef2` was only set upon compilation (in the `otTables.ClassDef.preWrite` method); dumping to XML immediately after `varLib.interpolate_layout` would produce a different output than saving it first, and then dumping to XML (23511fd302283a68774308c03fa93ae0411987a9)

### DIFF
--- a/Lib/fontTools/ttLib/tables/otTables.py
+++ b/Lib/fontTools/ttLib/tables/otTables.py
@@ -418,19 +418,19 @@ class ClassDef(FormatSwitchingBaseTable):
 			assert 0, "unknown format: %s" % self.Format
 		self.classDefs = classDefs
 
-	def preWrite(self, font):
+	def _getClassRanges(self, font):
 		classDefs = getattr(self, "classDefs", None)
 		if classDefs is None:
-			classDefs = self.classDefs = {}
-		format = 2
-		rawTable = {"ClassRangeRecord": []}
+			self.classDefs = {}
+			return
 		getGlyphID = font.getGlyphID
 		items = []
 		for glyphName, cls in classDefs.items():
-			if not cls: continue
+			if not cls:
+				continue
 			items.append((getGlyphID(glyphName), glyphName, cls))
-		items.sort()
 		if items:
+			items.sort()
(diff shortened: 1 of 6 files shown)


miguelsousa:
I've attempted to build a VF of Source Sans and it fails GPOS merging with the error below. I'm going to open a separate issue for it.

```python
  File "fontTools/varLib/merger.py", line 377, in _PairPosFormat2_merge
    assert l == lookup.SubTable[-1]
AssertionError: ('PairPos', '[1]', 'Lookup', '[10]', 'list', '.Lookup', 'LookupList', '.LookupList', 'GPOS', '.table', 'table_G_P_O_S_')
```

anthrotype:
@miguelsousa it looks like it's a known limitation; there is the comment saying:

```
	# Currently, if the coverage of PairPosFormat2 subtables are different,
	# we do NOT bother walking down the subtable list when filling in new
	# rows for alignment.  As such, this is only correct if current subtable
	# is the last subtable in the lookup.  Ensure that.
	# TODO: Remove this requirement
```

miguelsousa:
Bummer. I'm going to open the issue anyway and provide a test case, to help address this limitation.

behdad:
I'll merge and do followup fixes.

behdad:
Pushed followups.  This should be good for now.

PR #889, opened 2017-03-15
[AAT] Support the `gcid` table for Glyph-to-CID mapping

brawer:
https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6gcid.html
https://github.com/fonttools/fonttools/issues/178


brawer:
@behdad, how to extend otConverters with a computed UInt32 for the total table size?

brawer:
> how to extend otConverters with a computed UInt32 for the total table size

Figured it out. Rebased to current master, and did some cleanup. Now merging into master.

Issue #890, opened 2017-03-15
[pyftmerge] Duplicated DFLT tag in GSUB and GPOS table


zjusbo:
Steps to reproduce:
1. merge two fonts both contain DFLT tag in GSUB table. For example, noto-fonts/unhinted/NotoSans-Regular.ttf and noto-fonts/unhinted/NotoSansLimbu-Regular.ttf
pyftmerge NotoSans-Regular.ttf NotoSansLimbu-Regular.ttf
2. dump the GSUB table of the merged font
ttx -t GSUB merged.ttf
3. Check the script tags in GSUB in the result

Expected result:

Only one DFLT tag in it and values in DFLT are merged into it.

Actual result:

Multiple DFLT tags. 

DFLT tag in GPOS table has the same problem.

<GSUB>
    <Version value="0x00010000"/>
    <ScriptList>
      
      <ScriptRecord index="0">
        **<ScriptTag value="DFLT"/>**
        <Script>
          <DefaultLangSys>
            <ReqFeatureIndex value="65535"/>
            
            <FeatureIndex index="0" value="3"/>
          </DefaultLangSys>
          
        </Script>
      </ScriptRecord>
      <ScriptRecord index="1">
        **<ScriptTag value="DFLT"/>**
        <Script>
          <DefaultLangSys>
            <ReqFeatureIndex value="65535"/>
            
            <FeatureIndex index="0" value="0"/>
            <FeatureIndex index="1" value="1"/>

@behdad @anthrotype I know multiple DFLT in GSUB, GPOS is not a decent thing. Does it affects the functionality of the font? If so, I can take over it if it's worth spend time solving. 

One of our client is complaining about this issue and asked if it could be solved.

PR #891, opened 2017-03-16
update merge logic for ScriptList. Merge DFLT correctly.

zjusbo:
This PR fixes the bug in #890 

Our current merge logic does not merge DFLT correctly. Instead it creates multiple DFLT tags in the merged font.

### DIFF
--- a/Lib/fontTools/merge.py
+++ b/Lib/fontTools/merge.py
@@ -413,10 +413,33 @@ def merge(self, m, tables):
 	self.numSubTables = len(self.tables)
 	return self
 
+def mergeScriptRecord(lst):
+	"""Merge ScriptRecord. Merge all DFLT records into one.
+	"""
+	# TODO: merge all other duplicated tags?
+	lst = sumLists(lst)
+	records = []
+	DFLTRecord = None
+	for record in lst:
+		if record.ScriptTag == 'DFLT':
+			script = record.Script
+			assert hasattr(script, 'DefaultLangSys') and script.DefaultLangSys, 'DefaultLangSys must present in DFLT script'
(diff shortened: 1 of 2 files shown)


behdad:
This is wrong.

behdad:
You cannot just append the features of multiple language systems into one.  You need to merge duplicate features, which you *can* do by just appending their lookups (plus any sort requirement) BUT that slows down the font.  Properly merging lookups is a MUCH bigger problem than a small patch.

Please make sure you fully understand the semantics of the involved tables before proposing changes.  Thanks.

zjusbo:
Thanks for the feedback Behdad. I'll be more mindful next time.

So how serious is this problem, given the merged font contains multiple DFLT tags in GSUB and GPOS? Our current testing tool didn't detect this problem. And should this be considered as a stop-and-not-ship problem?

We are currently merging scores of non CJK fonts for our clients to use.

zjusbo:
Plus, do you mind if I add comments to the ScriptList merge logic to remind developers that current logic leads to multiple DFLT tables and should be fixed later?

behdad:
> So how serious is this problem, given the merged font contains multiple DFLT tags in GSUB and GPOS?

Very serious. Kerning / ligatures do not work.

> Our current testing tool didn't detect this problem.

Our testing is close to non-existent in terms of breath and depth.

> And should this be considered as a stop-and-not-ship problem?

Yes.

behdad:
> Plus, do you mind if I add comments to the ScriptList merge logic to remind developers that current logic leads to multiple DFLT tables and should be fixed later?

I added a note.

PR #894, opened 2017-03-23
Decomposing pen

anthrotype:
Following @behdad's comment:
https://github.com/fonttools/fonttools/pull/880#issuecomment-288558012

> RecordingPen records component references. We want to compare outlines, not component references (which might be the same while components themselves are different.) We need a version of RecordingPen that does that.

Instead of inheriting from `BasePen`, I separated the `addComponent` method from BasePen into a distinct `DecomposingPen` class, and use that as a base class for both `BasePen` and the new `ContourRecordingPen`.

Inheriting from `BasePen` directly would have meant also splitting up curveTo and qCurveTo, and converting quadratics to cubics, which we don't want to do for a recording pen.

### DIFF
--- a/Lib/fontTools/pens/basePen.py
+++ b/Lib/fontTools/pens/basePen.py
@@ -141,7 +141,34 @@ class NullPen(object):
 		pass
 
 
-class BasePen(AbstractPen):
+class DecomposingPen(AbstractPen):
+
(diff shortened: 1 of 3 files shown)


anthrotype:
closing in favour of #895

PR #895, opened 2017-03-23
Decomposing pen, using composition instead of inheritance

anthrotype:
actually, I changed my mind.. I believe it's better to have a filter pen that wraps another pen (e.g. a recording pen) for decomposing components.

Both the new filterPen and decomposingPen modules can be useful for other stuff, e.g. ufo2ft filters (https://github.com/googlei18n/ufo2ft/issues/120)

So this supersedes #894

### DIFF
--- a/Lib/fontTools/pens/decomposingPen.py
+++ b/Lib/fontTools/pens/decomposingPen.py
@@ -1,0 +1,22 @@
+from __future__ import print_function, division, absolute_import
+from fontTools.misc.py23 import *
+from fontTools.pens.filterPen import FilterPen
+from fontTools.pens.transformPen import TransformPen
+
+
+class DecomposingPen(FilterPen):
+
+    """A filter pen that decomposes components before drawing them with
+    another pen.
+    """
+
+    def __init__(self, outPen, glyphSet):
+        super(DecomposingPen, self).__init__(outPen)
+        self.glyphSet = glyphSet
+
+    def addComponent(self, glyphName, transformation):
+        """Transform the points of the base glyph and draw it onto self.
+        """
+        glyph = self.glyphSet[glyphName]
+        tPen = TransformPen(self, transformation)
+        glyph.draw(tPen)
--- a/Lib/fontTools/pens/filterPen.py
+++ b/Lib/fontTools/pens/filterPen.py
@@ -1,0 +1,37 @@
+from __future__ import print_function, division, absolute_import
+from fontTools.misc.py23 import *
+from fontTools.pens.basePen import AbstractPen
+
+
+class FilterPen(AbstractPen):
+
+    """ Base class for pens that apply some transformation to the coordinates
+    they receive and pass them to another pen.
+
+    You can override any of its methods. The default implementation does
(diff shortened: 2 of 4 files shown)


anthrotype:
E.g., if @zjusbo wants a RecordingPen that decomposes the components for https://github.com/fonttools/fonttools/pull/880, one can concatenate the two pens like I do in this test:
https://github.com/fonttools/fonttools/pull/895/commits/3c725f4d3486ec6517e6c06b94cef893f686a384

behdad:
Thanks. I've been thinking about moving in this direction (mixins) for a while. Will review.

anthrotype:
I used mixin approach in the other PR #894, but then I was getting lost in the intricacies of MRO stuff, so I thought it'd be simpler here to concatenate the pens.

If you like the idea of using a mixin class for the addComponent method, I can revisit this.

anthrotype:
perhaps "nest the pens" describes it better than "concatenate the pens".

zjusbo:
Thanks for the case. I'll look into it.

anthrotype:
@behdad how about we merge this? I'd like to use the decomposing pen in ufo2ft.

Lib/fontTools/pens/filterPen.py
```diff
@@ -0,0 +1,37 @@
...
+class FilterPen(AbstractPen):
+
+    """ Base class for pens that apply some transformation to the coordinates
+    they receive and pass them to another pen.
+
+    You can override any of its methods. The default implementation does
+    nothing, but passes the commands unmodified to the other pen.
+    """
+
+    def __init__(self, outPen, *args, **kwargs):
```
behdad: I don't think it should accept extra args...

Lib/fontTools/pens/decomposingPen.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class DecomposingPen(FilterPen):
+
+    """A filter pen that decomposes components before drawing them with
+    another pen.
+    """
+
+    def __init__(self, outPen, glyphSet):
+        super(DecomposingPen, self).__init__(outPen)
```
behdad: I think I like calling out "FilterPen" more than super here...

adrientetar: Nice thing about super is you don't have to hardcode the parent class

behdad: Sure, you don't have to hardcode, when you don't care.  In this case, it is assumed that FilterPen is called, ie. the call arguments are that of FilterPen, not any arbitrary parent.

anthrotype:
going to split this PR in two, one for the filterPen and another for the decomposing one

Lib/fontTools/pens/decomposingPen.py
anthrotype: > In this case, it is assumed that FilterPen is called, ie. the call arguments are that of FilterPen, not any arbitrary parent.

True. But one may also wish to call the `__init__` of a subcass of FilterPen, no?

PR #903, opened 2017-03-30
Snippets/README.md new link

davelab6:
thanks for sharing, @typefounding :)

### DIFF
--- a/Snippets/README.md
+++ b/Snippets/README.md
@@ -8,3 +8,4 @@ If you do and think your work is useful for others, please add a link to it here
 * https://github.com/twardoch/ttfdiet
 * https://github.com/googlei18n/nototools
 * https://github.com/googlefonts/fontbakery
+* https://github.com/Typefounding/setUseTypoMetricsFalse


anthrotype:
@benkiel I wonder why would one want to set UseTypoMetrics bit to false nowadays... That OS/2 flag was supposed to put an end to the old WinAscent/WinDescent mess. Using the OS/2 Typo vertical metrics for default line spacing should be the way to go for modern Windows apps.

benkiel:
@anthrotype Except that Word just started to respect that bit — but old documents opened in new Word with same font now can have different line spacing depending on how one sets the vertical metrics — and it seems that Word still uses the hhea metrics when printing, but not for on screen display, so it's a right mess.

khaledhosny:
Ideally hhea metrics should match OS/2 typo metrics, is there any reason for them to be different (just curious)?

benkiel:
@khaledhosny Yes, in an ideal world — however, that world doesn't happen inside of Word. If your accents are above the the hhea metrics, they'll be clipped off (not shown) in Word. This happens very often, so the way that many folks set metrics is hhea ascent/decent & win ascent/descent are set to the bounding box max/min of a family. No clipping. hhea linegap is set to 0. Typo metrics are set to be more rational, and sum to the UPM. Typo line gap is set to the difference of the sum of hhea (when I say sum, I mean the sum of the abs of each) minus the sum of the typo values.

Doing this makes things work well in Word and in design applications. Yes, this isn't 100% to the spec, but try telling customers that it's Word's problem — they just want their fonts to work.

khaledhosny:
That is weird, is this perhaps Word for Mac?

benkiel:
@khaledhosny largely Word for Mac, but not always.

justvanrossum:
Btw. @benkiel, why should `sTypoAscender` and `sTypoDescender` add to the UPM? I hear that all the time, but the MS spec doesn't say that.

benkiel:
@justvanrossum They don't have to — my apologies for implying that this was a must. It is a MS recommendation that they do (https://www.microsoft.com/typography/otspec/recom.htm#tad), but it's not a requirement. Making it sum means that you won't have overlap in things that calculate baseline to baseline with just the typoAscender + typoDescender.

Anyway, vertical metrics are no fun — and we may now have written too much for a pull of a link for something I wrote to solve an issue with setting one bit if you've set vertical metrics up one way!

PR #910, opened 2017-04-07
pens: add a wx pen

adrientetar:
r? @anthrotype


adrientetar:
Seems some of the buildbots have trouble fetching from a fork...

miguelsousa:
@adrientetar would it valuable to add tests for this new pen? There are already a few pens in the collection that are lacking tests

adrientetar:
I could add a test but unless we're compiling and installing wx on the buildbots, it's not very useful. Besides it's just 1:1 function dispatch, and the code won't change.

PR #913, opened 2017-04-10
[TSI*] Fix computation of the textLength of VTT private tables entries

anthrotype:
This patch rewrites the TSI1 table `decompile` method to work around an issue which I encountered while decompiling tables generated with the latest VTT 6.20.
Many thanks @be5invis for the hints found in his otfcc compiler ;)

Besides this also adds:
- a `match` method to the `loggingTools.CapturingLogHandler` (9528717);
- a backport for python 3's [`SimpleNamespace` class](https://docs.python.org/3/library/types.html#types.SimpleNamespace) (a844249)

both can be handy e.g. when writing tests.

### DIFF
--- a/Lib/fontTools/misc/loggingTools.py
+++ b/Lib/fontTools/misc/loggingTools.py
@@ -463,7 +463,7 @@ class CapturingLogHandler(logging.Handler):
 	def createLock(self):
 		self.lock = None
 
-	def match(self, regexp):
+	def assertRegex(self, regexp):
 		import re
 		pattern = re.compile(regexp)
 		for r in self.records:
--- a/Lib/fontTools/misc/loggingTools.py
+++ b/Lib/fontTools/misc/loggingTools.py
@@ -463,6 +463,13 @@ class CapturingLogHandler(logging.Handler):
 	def createLock(self):
 		self.lock = None
 
+	def match(self, regexp):
+		import re
+		pattern = re.compile(regexp)
+		for r in self.records:
+			if pattern.search(r.msg):
+				return True
+		assert 0, "Pattern '%s' not found in logger records" % regexp
 
 def deprecateArgument(name, msg, category=UserWarning):
 	""" Raise a warning about deprecated function argument 'name'. """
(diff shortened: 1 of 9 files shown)


Lib/fontTools/misc/loggingTools.py
```diff
@@ -463,6 +463,13 @@ def emit(self, record):
 	def createLock(self):
 		self.lock = None
 
+	def match(self, regexp):
```
brawer: Maybe call the method "assertMatch", to make clear that it will raise an error if there’s no match?

Lib/fontTools/ttLib/tables/T_S_I__0.py
```diff
@@ -3,7 +3,7 @@
 from . import DefaultTable
 import struct
 
-tsi0Format = '>HHl'
+tsi0Format = '>HHL'
```
brawer: Is there a somewhere a spec for the table format? If so, it might be good to add its URL here (couldn’t find anything on the web, but maybe I’m using the wrong search engine). Otherwise, mind adding a comment that explains the purpose of the table?

Lib/fontTools/misc/loggingTools.py
anthrotype: I wanted to mimic the `match` method of pytest ExceptionInfo object:
https://docs.pytest.org/en/latest/builtin.html#_pytest._code.ExceptionInfo.match

I was thinking of returning `False` instead of `assert 0` and let the user choose to do `assert captor.match(...)` when writing pytest tests, or `self.assertTrue(captor.match(...))` when writing unittest tests. But then one would not get the error message, "Pattern '%s' not found...".

Ok, I'll go for `assertMatch`.

Lib/fontTools/ttLib/tables/T_S_I__0.py
anthrotype: Unfortunately no, the VTT private tables' spec is not publicly available. I could only obtain some of this information indirectly via informal conversations.
I'll try to add more comments to make clear what the purpose of these tables is.

PR #915, opened 2017-04-11
[loggingTools] added LogMixin class

anthrotype:
This is useful to quickly add logging functionality to classes, and to reduce boilerplate.

It adds a `log` property to the class inheriting from it, which uses `logging.getLogger` to get a logging.Logger (sigleton) object named after `<module>.<class>` of `self`.

### DIFF
--- a/Lib/fontTools/misc/loggingTools.py
+++ b/Lib/fontTools/misc/loggingTools.py
@@ -471,6 +471,18 @@ class CapturingLogHandler(logging.Handler):
 				return True
 		assert 0, "Pattern '%s' not found in logger records" % regexp
 
+
+class LogMixin(object):
+	""" Mixin class that adds a 'log' property getter returning an instance
+	of logging.Logger class, which is named after self <module>.<class> name.
+	"""
+
+	@property
+	def log(self):
+		name = ".".join([self.__class__.__module__, self.__class__.__name__])
+		return logging.getLogger(name)
+
+
 def deprecateArgument(name, msg, category=UserWarning):
 	""" Raise a warning about deprecated function argument 'name'. """
 	warnings.warn(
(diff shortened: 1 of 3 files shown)


anthrotype:
markdown doesn't like angle brackets. The name (or channel) of the Logger is set to the inheriting class's `__module__ + "." + __class__`. E.g., if the class `MyClass` is defined in the module `mypackage.mymodule`, then the logger name for all the instances of that class is `mypackage.mymodule.MyClass`.

behdad:
So how do you see this used?  Any class that likes to have informational logging adds it and calls self.log....?

brawer:
Interesting trick! But it adds yet another layer of complexity that developers need to understand. Personally I’m someone who prefers boring but straightforward code — admittedly, that’s because I’m just too stupid to understand the smart stuff... On the other hand, each smart trick makes it harder for new developers to join the project.

anthrotype:
Exactly yes, you just add that mixin class to your multiple inheritance list an you get a logger for free already with the correct name.

anthrotype:
For example, `table_T_S_I__3` is a subclass of `table_T_S_I__1`, so its `self.log` would emit from `fontTools.ttLibl.tables.T_S_I__3.table_T_S_I__3` whenever it's called from the former, or `....table_T_S_I__1` when called from the latter.

anthrotype:
@brawer I share your concern about avoiding "smart trick makes it harder for new developers to join the project". However, mixins are relatively straightforward once you understand how multiple inheritance and the method resolution order works in Python. There's plenty of resources online that describe how it works.
Basically, the classes listed from left to right describe a bottom-up class hierarchy, i.e. a class that comes before another is interpreted as a subclass of the next; so when one looks up a method or property from an instance, the chain of super-classes is walked until one is found, else AttributeError.
I added a docstring explaining how to use the LogMixin.

The tests are failing for unrelated reason, going to fix that afterwards.

PR #918, opened 2017-04-12
[pens] add filterPen module

anthrotype:
The base `FilterPen` class is simply a pen that delegates the drawing to another one, which is passed as argument in the `__init__` method.

The existing `TransformPen` is an example of this class of pens.

The default implementation of `FilterPen` simply passes through the drawing commands to the sub-pen, without modifications.

One can subclass FilterPen and override any of its methods in order to manipulate the contours or components, before sending them to the other pen.

### DIFF
--- a/Lib/fontTools/pens/filterPen.py
+++ b/Lib/fontTools/pens/filterPen.py
@@ -1,0 +1,74 @@
+from __future__ import print_function, division, absolute_import
+from fontTools.misc.py23 import *
+from fontTools.pens.basePen import AbstractPen
+
+
+class FilterPen(AbstractPen):
+
+    """ Base class for pens that apply some transformation to the coordinates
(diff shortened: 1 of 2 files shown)


anthrotype:
I'm going to merge the FilterPen class, as it should be uncontroversial. It does not change anything, simply adds a new class of pens that already implicitly existed, and makes writing new filter pens easier, as subclasses only need to override the methods that they wish to behave differently than a simple pass-through filter.

I will keep the other PR about the DecomposingPenMixin (#919) open for review/discussion.

anthrotype:
Maybe one thing we could decide about the FilterPen api is whether the `_outPen` should be stay a semi-private attribute, or it should be made public, maybe the name simplified as just `pen`.
I'm thinking of `pen.pen.pen.pen`...

PR #919, opened 2017-04-12
[pens] add DecomposingPen and DecomposingRecordingPen

anthrotype:
This moves the `addComponent` method of BasePen into a separate `DecomposingPen` that can be mixed in with a RecordingPen to make a `DecomposingRecordingPen` class that records components decomposed, as simple contours.

### DIFF
--- a/Lib/fontTools/pens/basePen.py
+++ b/Lib/fontTools/pens/basePen.py
@@ -38,6 +38,7 @@ sequence of length 2 will do.
 
 from __future__ import print_function, division, absolute_import
 from fontTools.misc.py23 import *
+from fontTools.misc.loggingTools import LogMixin
 
 __all__ =  ["AbstractPen", "NullPen", "BasePen",
 			"decomposeSuperBezierSegment", "decomposeQuadraticSegment"]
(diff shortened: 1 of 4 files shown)


Lib/fontTools/pens/basePen.py
```diff
@@ -141,16 +142,52 @@ def addComponent(self, glyphName, transformation):
...
+
+	Subclasses can set the 'skipMissingComponents' attribute to True, if they
+	want to ignore the 'KeyError' exception normally raised when a base glyph
+	is missing from the glyphSet, and replace it by a warning message.
+	"""
+
+	skipMissingComponents = False
+
+	def __init__(self, glyphSet):
+		super(DecomposingPenMixin, self).__init__()
```
anthrotype: the call to `super` here is ok if the `__init__` method of the next class in the inheritance line does not take any required arguments (as is the case of the `RecordingPen`).
However, it breaks if the super class constructor requires any arguments... Maybe that's ok.

anthrotype: Say I want to make a class that inherits from both `FilterPen` (#918), as it controls another output pen, and from the `DecomposingPenMixin`, because I want it to decompose the components as well.

If I place the `FilterPen` class after the `DecomposingPenMixin` class in the __mro__, the `super()` call in the latter's `__init__` method would in turn try to call the `FilterPen.__init__` method, but this would fail as it's missing the required `outPen` argument.

If I put `FilterPen` before `DecomposingPenMixin`, it would not call the `addComponent` method that decomposes components, but the one from `FilterPen` would be used instead...

The only way I can think of making that work is cheat and set the private attributes myself without calling the superclasses' `__init__`, but this breaks encapsulation:

```python
class DecomposingFilterPen(DecomposingPenMixin, FilterPen):
    def __init__(self, outPen, glyphSet):
        self._outPen = outPen
        self.glyphSet = glyphSet
```

I'm stuck.

anthrotype:
@justvanrossum thanks for taking the time to review the other #894 PR.
Unfortunately I had already deleted that branch, so now I manually merged that patch into this one.
Please take a look.

Issue #921, opened 2017-04-12
[name] Support adding entries for multiple languages


behdad:
Name table object has addName(). Would be good to support adding name for multiple languages at the same time.  Eg. if string is a dict (like `{'en':'Weight', 'fa-IR':'وزن'}).  I need this in fvar generation.

brawer:
A while a go I’ve added support for multilingual names to OpenType.js, exposing BCP47 language codes in the API. I didn’t bother with `name` format 1 (since nobody seems to implement this), but I did support `ltag`. https://github.com/nodebox/opentype.js/blob/master/src/tables/name.js

behdad:
@brawer can you work on this?  Initially maybe just map to Windows-platform language codes?

PR #928, opened 2017-04-18
[name] Add helper for building multi-lingual names

brawer:
Fixes https://github.com/fonttools/fonttools/issues/921.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_n_a_m_e.py
+++ b/Lib/fontTools/ttLib/tables/_n_a_m_e.py
@@ -1,8 +1,11 @@
+# -*- coding: utf-8 -*-
 from __future__ import print_function, division, absolute_import
+from __future__ import unicode_literals
 from fontTools.misc.py23 import *
 from fontTools.misc import sstruct
 from fontTools.misc.textTools import safeEval
 from fontTools.misc.encodingTools import getEncoding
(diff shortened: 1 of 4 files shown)


Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -145,13 +149,49 @@ def setName(self, string, nameID, platformID, platEncID, langID):
...
+		The nameID is assigned in the range between 'minNameID' and 32767 (inclusive),
+		following the last nameID in the name table.
+		"""
+		names = getattr(self, 'names', [])
+		nameID = 1 + max([n.nameID for n in names] + [minNameID - 1])
+		if nameID > 32767:
+			raise ValueError("nameID must be less than 32768")
+		return nameID
+
+	def addMultilingualName(self, names, ttFont, nameID=None):
```
behdad: I think we still want to take platforms to add.  Ie. for Noto fonts, and most other uses, I believe we don't want to add Mac platform names, as Mac happily parses and uses Windows names.

behdad: Can you make ttFont default to None, in which case ltag won't be used? And add documentation as to the exact semantics.  Thanks!

Issue #931, opened 2017-04-19
name.addMultilingualName() issues


behdad:
Multiple issues:
- When storing name with Mac platform, we should check whether the name is representable in the encoding chosen by macScript.  I personally like to forgo the Mac platform names completely.

- When encoding using ltag: ltag table spec says encode name under "Unicode platform".  Unicode platform is number 0, but code seems to use platform 2 (deprecated "ISO"):
```
self.names.append(makeName(name, nameID, 2, 4, ltag.addTag(lang)))
```
For encoding ID, we're using 4.  Wonder if 6 is more correct?

Anyway, because of this, right now, unknown language tags with non-ASCII text don't work because getEncoding() returns ASCII for those.  It also means that this codepath is untested.

behdad:
Thanks Sascha!

PR #934, opened 2017-04-20
[name] Fix issues with name.addMultilingualName()

brawer:
Fixes https://github.com/fonttools/fonttools/issues/931.


moyogo:
Should this use format 1 name table when there is no language code for Windows?

brawer:
> Should this use format 1 name table when there is no language code for Windows?

See the TODO comment in the patch. Actually it depends a bit on the fate of `name` format 1 versus `ltag` — I’m having hopes that MS and Apple might reconcile this difference between their respective specs, but obviously I don’t know what the outcome of that discussion will be. The advantage of `ltag` is that it’s implemented by existing fonts, which isn’t the case for `name` v1.

PR #935, opened 2017-04-24
setting up Sphinx-based autodocs

gferreira:


anthrotype:
Thank you Gustavo!

I see you have removed the man page for ttx, was that intentional? I'd be ok to drop that, but maybe others may like to keep it, or have it generated from the Sphinx-based docs somehow (I don't know how).

Could you add a `sphinx>=1.5.5` line to the file `dev-requirements.txt`? (I see 1.5.5 is the currnt latest version). It may look obvious that one needs Sphinx to build a sphinx-based documentation, but it's fine to have it there, if if only as a reminder.

I see that all the docstrings would need to be reformatted as reStructuredText, so they will render correctly by Sphinx. But we can do that gradually, as we add more documentation.

Also, it'd be have the docs build script run on Travis CI, but we can do that later as well.

anthrotype:
Another minor suggestion. I know that Sphinx comes with a setuptools integration feature which lets you build the docs with `python setup.py build_sphinx` instead of via Makefile.
I haven't used it, but I wonder if that could be better than implicitly requiring GNU make on UNIX, and running a batch script on Windows, as you did here.
See http://www.sphinx-doc.org/en/master/setuptools.html

I also noticed a lots of WARNING and ERROR messages while running `make html`, some have to do with the incorrect formatting, others a bit more worrying say `SEVERE: Duplicate ID...`, which affect modules with the same name `parser.py` but under different sub-packages (e.g. `feaLib.parser`, `voltLib.parser`...).
Anyway, that's a good start, I wouldn't consider these issues as blocking.
Let me know when you think the current PR is ready to be merged.

gferreira:
hello Cosimo, thanks for the review. I’ve updated `dev-requirements.txt`, and restored the ttx man page. Please feel free to make any other change you think is necessary.

I would like to focus on reviewing the docstrings to improve the output in Sphinx. Have a look at [fontParts](https://github.com/robofab-developers/fontParts/), I think it’s a good example.

anthrotype:
thanks! Another thing we need to do is updating the MANIFEST.in file so that all the source files in Doc folder are included in the source distribution.

Issue #403, opened 2015-11-07
Add CFF operator specializer


behdad:
There's starting code Sam wrote last year:

https://github.com/samfishman/fonttools/commit/6998f15c4480862fb7f3c2cb314f94f0b243890e

It can be improved, but mostly it need tests to be added and integrated.

anthrotype:
Things have improved with https://github.com/fonttools/fonttools/pull/940.

I was rereading the path operators chapter in the [Type2 spec](http://wwwimages.adobe.com/content/dam/Adobe/en/devnet/font/pdfs/5177.Type2.pdf
), and noticed that it's not just the `rcurveline` and `rlinecurve` operators that combine multiple segments, as Miguel noted, but also all the `*lineto` and `*curveto` ones can take "one or more" line or Bezier segments depending on the number of arguments on the stack.
With Miguel's patch, we currently add one `*lineto` or `*curveto` operator for each segment, even when we the same operator could draw multiple segments of the same kind.
So there's still room for optimization.
If we are happy with this (I am), I could tag a new release.

behdad:
FWIW here's what AFDKO does:
https://github.com/khaledhosny/psautohint/blob/0e920ed195569eeae4b562893afa6c2d75a55d74/python/BezTools.py#L391

I'll comment extensively later.

miguelsousa:
I've hooked up **psOperatorSpecializer.py** to the latest version of `t2CharStringPen` by doing this,
```python
from fontTools.misc.psOperatorSpecializer import simplifyCharstring
#...
program = simplifyCharstring(self._program) + ["endchar"]
```
and the only difference I saw was that the new charstrings use `rcurveline`.
The specializer does not implement `rlinecurve` and does not merge operators that can be combined.
So I think that at this point we're better off enhancing `t2CharStringPen` further.

behdad:
I think we want to revert https://github.com/fonttools/fonttools/pull/940 and go with a separate specializer than the pen calls in the end, like Miguel did in the previous comments.  Main reason for this is such that we can optimize existing programs as well.

I have a design of a specializer (and a "de"specializer to go with it) which avoids the if-tree hell.  I like to give that a try.  The more tests we have the better...

behdad:
Oh, and let's make cffLib a proper module directory and move all PS-related code from misc/ under cffLib and introduce more submodules...

anthrotype:
sgtm

behdad:
[...]

PR #940, opened 2017-04-28
[t2CharStringPen] Add support for more path operators

miguelsousa:
These changes will reduce the CFF table a bit.

### DIFF
--- a/Lib/fontTools/pens/t2CharStringPen.py
+++ b/Lib/fontTools/pens/t2CharStringPen.py
@@ -113,7 +113,12 @@ class T2CharStringPen(RelativeCoordinatePen):
     def _relativeMoveTo(self, pt):
         pt = self.roundPoint(pt)
         x, y = pt
-        self._heldMove = [x, y, "rmoveto"]
+        if x == 0:
+            self._heldMove = [y, "vmoveto"]
+        elif y == 0:
+            self._heldMove = [x, "hmoveto"]
+        else:
+            self._heldMove = [x, y, "rmoveto"]
 
     def _storeHeldMove(self):
         if self._heldMove is not None:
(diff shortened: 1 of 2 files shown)


anthrotype:
Thanks Miguel! Do we support all the type2 drawing operators with this?

miguelsousa:
No, not all of them. Some operators such as `rcurveline` are a combination of two or more segments. Supporting those would make the pen unnecessarily complex.
I think the pen now supports all the Type2 drawing operators that make sense in the context of a pen. The remaining operators, as well as optimizations (such as combining consecutive moveTos and curveTos), can be implemented during subroutinization.

anthrotype:
I see. Well, this is definitely an improvement so we shall merge it. However, I think it still would be good to eventually support also the operators that combine multiple segments in this pen. We would just have to collect the current path's coordinates and "flush" them into the `self._program` only when `_closePath` or `_endPath` are called.

anthrotype:
The line 119 is not covered by the tests:
https://codecov.io/gh/fonttools/fonttools/commit/70c695f8f69e614a097a54fbf8483392da9fab95

This is in the `T2CharStringPen._relativeMoveTo` method, when dy == 0 and the `hmoveto` operator is used.

miguelsousa:
> We would just have to collect the current path's coordinates and "flush" them into the `self._program` only when `_closePath` or `_endPath` are called.

Yeah, that may be a way to do it, now that you mentioned it. I shall try it at a later date. But first I'll probably attempt to hook up the psOperatorSpecializer that Behdad referenced at #403 to the pen.

behdad:
Thanks Miguel.  I checked your code.  It's definitely a good start.  It would be trivial to combine subsequent operators of the same kind.  Maybe I do that.

I think we should remove the RelativeCoordinatePen() and simplify the code a bit first.  It's more complex than it has to be.

miguelsousa:
Yeah, I never quite understood the benefits of the RelativeCoordinatePen, so I left it alone.

PR #943, opened 2017-05-02
Varlib.mutator interpolation

jenskutilek:
I have implemented the interpolation of points with implicit deltas when generating an instance with varLib.mutator.

On the build side, there is only one optimization yet, to store only one delta for the whole contour if possible, and only if the resulting compiled code would be shorter that explicit deltas.

I’d be happy if anybody would like to take a look at the code.

My test data is not yet part of the repo, I should add some tests before this could be merged.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -247,7 +247,7 @@ def _compressedDelta(delta, control):
 			# TODO: Check for point deltas that have the same explicit coordinates
 			# as if they were simply interpolated ("IUP" feature). These could be omitted.
 			# For now, just add all delta values of this contour verbatim.
-			compressed_delta.extend([d for d in contour_delta])
+			compressed_delta.extend(contour_delta)
 	return compressed_delta
 
 
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -203,6 +203,54 @@ def _SetCoordinates(font, glyphName, coord):
 	font["hmtx"].metrics[glyphName] = horizontalAdvanceWidth, leftSideBearing
 
 
+def _compressedDelta(delta, control):
+	# Try to optimize the deltas for each contour.
+	# E.g. if a whole contour is moved, only one delta needs to be stored.
+	# The result is not always smaller than storing the deltas verbatim,
+	# because it requires that the point index for each of the remaining
+	# deltas is stored.
+	contour_deltas = []
+	i = 0
+	if type(control[0]) == str:
+		# Composite glyph: control contains the component names
+		# Apparently, probably depending on the component flags, the delta
+		# record already contains the sidebearing points in some cases, but
+		# not in others.
+		contour_deltas.extend([[d] for d in delta[:len(control)]])
+		if len(delta) > len(control):
+			# Phantom points are included in the delta record, add them to the contour deltas.
+			# Phantom points are not interpolated, they can each be optimized separately.
+			contour_deltas.extend([[d] for d in delta[-4:]])
+		assert len(contour_deltas) == len(delta)
+	else:
+		# Simple glyph: control contains the list of contour end points and flags
+		for j in control[0]:
+			# Store each contour in a sublist of contour_deltas
+			contour_deltas.append(delta[i:j + 1])
+			i = j + 1
+		# Phantom points are not interpolated, they can each be optimized separately.
+		contour_deltas.extend([[d] for d in delta[-4:]])
+	compressed_delta = []
+	for contour_delta in contour_deltas:
+		delta_set = set(contour_delta)
+		if len(delta_set) == 1:
+			# The whole contour is moved, just store one delta value for the start point.
+			first_delta = list(delta_set)[0]
+			if first_delta == (0, 0):
+				# The contour is not actually moved, store None for all points
+				compressed_delta.extend([None for _ in range(len(contour_delta))])
+			else:
+				# Store a delta for the first point, and None for all others.
+				compressed_delta.append(first_delta)
+				compressed_delta.extend([None for _ in range(len(contour_delta) - 1)])
+		else:
+			# TODO: Check for point deltas that have the same explicit coordinates
+			# as if they were simply interpolated ("IUP" feature). These could be omitted.
+			# For now, just add all delta values of this contour verbatim.
+			compressed_delta.extend([d for d in contour_delta])
+	return compressed_delta
+
+
 def _add_gvar(font, model, master_ttfs, tolerance=.5):
 
 	log.info("Generating gvar")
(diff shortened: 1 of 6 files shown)


PeterCon:
Note that there is an issue with the current OT spec gvar chapter with regard to the algorithm given for calculating inferred deltas -- Behdad identified a divide-by-zero scenario that wasn't covered. While discussing the appropriate revision, Laurence Penney provided a test font (GVAR-9 in [unicode-org/text-rendering-tests](https://github.com/unicode-org/text-rendering-tests)) that revealed some subtleties in the GX logic on the Mac. So, I think we have a correct understanding now. This will be reflected in the next OT spec update. Here's pseudo-code from the draft that summarizes the logic.

In the following, precedingPoint and followingPoint refer to adjacent points on the contour, before and after in contour point order, nearest to the target point for which a delta is being inferred, that have explicit deltas. This assumes it has been determined that some but not all points on the contour have explicit deltas. This logic assumes one grid direction, X or Y; it gets applied to both directions independently.

    if precedingPoint.coord = followingPoint.coord
    {
        if precedingPoint.delta = followingPoint.delta
            targetPoint.delta = precedingPoint.delta
        else
            targetPoint.delta = 0
    }
    else  /* precedingPoint.coord <> followingPoint.coord */
    {
        if targetPoint.coord <= min(precedingPoint.coord, followingPoint.coord)
        {
            if precedingPoint.coord < followingPoint.coord
                targetPoint.delta = precedingPoint.delta
            else  /* followingPoint.coord < precedingPoint.coord */
                targetPoint.delta = followingPoint.delta
        }
        else if targetPoint.coord >= max(precedingPoint.coord, followingPoint.coord)
        {
            if precedingPoint.coord > followingPoint.coord, then
                targetPoint.delta = precedingPoint.delta
            else  /* followingPoint.coord > precedingPoint.coord */
                targetPoint.delta = followingPoint.delta
        }
        else  /* target point coordinate is between adjacent point coordinates */
        {
            /* target point delta is derived from the adjacent point deltas
               using linear interpolation */
        }
    }

jenskutilek:
Thank you! I have looked at how the FreeType library calculates the interpolated points.

I’ll compare the pseudo code with my version tomorrow.

behdad:
Thanks Jens!  I'll take a look soon.

jenskutilek:
I have adapted my code to use the sample pseudo code.

The results were the same for my test cases, but staying close to sample code is better anyway.

jenskutilek:
Not sure why the tests fail ... the LSB seems to be not calculated correctly in some cases (but not in all build environments!?)

anthrotype:
The computation of the xMin value of the "q" glyph, used in `varLib._SetCoordinates` to compute its hmtx lsb value, seems to not be deterministic, as it changes as the PYTHONHASHSEED changes.

When running the tests with tox, the PYTHONHASHSEED is set every time to a different (random) integer (it's printed in of tox console output); this way the Python 3 hash randomization feature is kept under control, and the tests failures/successes are deterministic.

For example, this specific value for the hash seed makes it trigger the failure we see on the CI:
```
PYTHONHASHSEED='2825304365' pytest Tests/varLib/mutator_test.py
```

To fix the issue, you should make sure that computation of the coordinates does not depend on the implicit order of unsorted collections like dict or set, etc.

Lib/fontTools/varLib/__init__.py
```diff
@@ -203,6 +203,54 @@ def _SetCoordinates(font, glyphName, coord):
...
+			contour_deltas.append(delta[i:j + 1])
+			i = j + 1
+		# Phantom points are not interpolated, they can each be optimized separately.
+		contour_deltas.extend([[d] for d in delta[-4:]])
+	compressed_delta = []
+	for contour_delta in contour_deltas:
+		delta_set = set(contour_delta)
+		if len(delta_set) == 1:
+			# The whole contour is moved, just store one delta value for the start point.
+			first_delta = list(delta_set)[0]
```
anthrotype: hm, this line looks suspicious `list(delta_set)[0]`, as sets are unsorted (it's order is implementation specific and on python 3 uses a randomized seed), thus the `first_delta` will be different at every run.

anthrotype: ok well, in this case, `first_delta` is always the same as the `delta_set` length is 1.. sorry.
Anyway, you should look for things like that, and make sure you don't rely on any predefined order for sets, dict keys or items, etc.

Lib/fontTools/varLib/__init__.py
```diff
@@ -203,6 +203,54 @@ def _SetCoordinates(font, glyphName, coord):
...
+				compressed_delta.extend([None for _ in range(len(contour_delta))])
+			else:
+				# Store a delta for the first point, and None for all others.
+				compressed_delta.append(first_delta)
+				compressed_delta.extend([None for _ in range(len(contour_delta) - 1)])
+		else:
+			# TODO: Check for point deltas that have the same explicit coordinates
+			# as if they were simply interpolated ("IUP" feature). These could be omitted.
+			# For now, just add all delta values of this contour verbatim.
+			compressed_delta.extend([d for d in contour_delta])
```
anthrotype: Ah, I guess it's this line where you are making a list from a set: `[d for d in contour_delta]`.

anthrotype: forget what I said, `contour_deltas` is not a set, but a list. Sorry, I should properly read the code before commenting :)

justvanrossum: But then again, what does `[d for d in contour_delta]` add to just feeding `contour_delta` to `.extend()`?

anthrotype: Nothing indeed

Lib/fontTools/varLib/mutator.py
```diff
@@ -40,12 +40,14 @@ def main(args=None):
 
 	gvar = varfont['gvar']
 	for glyphname,variations in gvar.variations.items():
```
jenskutilek: This will cause problems for composite glyphs. If a composite glyph is processed before its source glyph, the bounds of the composite glyph may be calculated incorrectly.

Lib/fontTools/varLib/__init__.py
jenskutilek: Following your pointers, I think I found the cause: https://github.com/fonttools/fonttools/pull/943/files#diff-c65e1b93720832bb19270ab9620ac461L42

`for glyphname,variations in gvar.variations.items():`

If a composite glyph is processed before its base glyph, the composite glyph bounds may be calculated incorrectly because the deltas haven’t been applied to the base glyph yet.

jenskutilek: Nested composites could complicate this further ...

jenskutilek: I filed this as issue #945 separately.

jenskutilek: Thanks for spotting this, @justvanrossum! I removed the useless list comprehension in 40b092c.

anthrotype:
@jenskutilek  so after #947, if you merge master into the current branch, or rebase it the updated master, the tests should now pass, correct?

anthrotype:
would be nice if you could also squash your commits a bit with interactive rebase, `git rebase -i master`, so that the git log contains less noise. Otherwise, we can squash them all into a single commit as we merge the pull request.

jenskutilek:
Yep, the tests should pass now. I cleaned up the commit history a bit.

anthrotype:
>  I cleaned up the commit history a bit.
 
did you? I still see lots of redundant commits (49) as well as merge commits here and there.

jenskutilek:
I tried to do a rebase, but maybe I messed it up :/

anthrotype:
The git docs explain how to rewrite history with interactive rebase
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History
but don't worry, we'll just squash everything once the PR is merged.
I'll let @behdad or @sascha comment on the actual code, as I'm not familiar enough with the gvar spec.

jenskutilek:
I tried again ...

anthrotype:
Thanks Jens, it worked now!

jenskutilek:
Closing this in favor of #969.

anthrotype:
@jenskutilek thank you for setting the ball rolling :)

PR #947, opened 2017-05-03
Sort glyphs by component depth before processing (#945)

### LINKED ISSUE #945: varLib.mutator: glyph processing order
https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/varLib/mutator.py#L42

`for glyphname,variations in gvar.variations.items():`

If a composite glyph is processed before its base glyph, the composite glyph bounds may be calculated incorrectly because the deltas haven’t been applied to the base glyph yet.

Nested composites could complicate this further ...

jenskutilek:
Fixes #945.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -298,7 +298,7 @@ SCALED_COMPONENT_OFFSET		= 0x0800  # composite designed to have the component of
 UNSCALED_COMPONENT_OFFSET	= 0x1000  # composite designed not to have the component offset scaled (designed for MS)
 
 
-MaxpValues = namedtuple('MaxpValues', ['nPoints', 'nContours', 'maxComponentDepth'])
+CompositeMaxpValues = namedtuple('CompositeMaxpValues', ['nPoints', 'nContours', 'maxComponentDepth'])
 
 
 class Glyph(object):
@@ -440,7 +440,7 @@ class Glyph(object):
 						glyfTable, maxComponentDepth + 1)
 			nPoints = nPoints + nP
 			nContours = nContours + nC
-		return MaxpValues(nPoints, nContours, maxComponentDepth)
+		return CompositeMaxpValues(nPoints, nContours, maxComponentDepth)
 
 	def getMaxpValues(self):
 		assert self.numberOfContours > 0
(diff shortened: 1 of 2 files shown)


anthrotype:
how about using `sorted(..., key=...)` with a `lambda` expression instead?

```diff
diff --git a/Lib/fontTools/varLib/mutator.py b/Lib/fontTools/varLib/mutator.py
index 349425d..132ba25 100644
--- a/Lib/fontTools/varLib/mutator.py
+++ b/Lib/fontTools/varLib/mutator.py
@@ -39,7 +39,16 @@ def main(args=None):
        print("Normalized location:", loc)
 
        gvar = varfont['gvar']
-       for glyphname,variations in gvar.variations.items():
+       glyf = varfont['glyf']
+       # get list of glyph names in gvar sorted by component depth
+       glyphnames = sorted(
+               gvar.variations.keys(),
+               key=lambda name: (
+                       glyf[name].getCompositeMaxpValues(glyf)[2]
+                       if glyf[name].isComposite() else 0,
+                       name))
+       for glyphname in glyphnames:
+               variations = gvar.variations[glyphname]
                coordinates,control = _GetCoordinates(varfont, glyphname)
                for var in variations:
                        scalar = supportScalar(loc, var.axes)
```

also for better legibility, we could make `getCompositeMaxpValues` return a `namedtuple`, so instead of `.getCompositeMaxpValues(glyf)[2]` one could also do `.getCompositeMaxpValues(glyf).maxComponentDepth`.

jenskutilek:
Sure, looks good. I’m just not that familiar with lambda expressions, that’s why I never remember to use them :)

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -297,6 +298,9 @@ def flagEncodeCoords(flag, x, y, xBytes, yBytes):
 UNSCALED_COMPONENT_OFFSET	= 0x1000  # composite designed not to have the component offset scaled (designed for MS)
 
 
+MaxpValues = namedtuple('MaxpValues', ['nPoints', 'nContours', 'maxComponentDepth'])
```
anthrotype: nit: shall we call this `CompositeMaxpValues` instead, for clarity sake?

jenskutilek: Ah yes, I didn’t look closely enough and thought these applied to all glyphs.

PR #959, opened 2017-05-08
Specializer tests

miguelsousa:


Lib/fontTools/cffLib/specializer.py
```diff
@@ -50,7 +50,10 @@ def programToCommands(program):
 				try:
 					width = stack.pop(0)
 				except IndexError:
-					raise ValueError(stack)
+					if ignoreErrors:
+						width = ''
```
anthrotype: @miguelsousa you sure an empty string `''` for the width is ok here? I thought the args list should be numbers only, but maybe i'm missing something else

miguelsousa: Yes, it's ok. The code handles int, float, None and str equally fine.

anthrotype: hm, ok if you say so.. To me it would be clearer if it skipped appending the command in that case. But nevermind

miguelsousa: I understand your concern. I did it this way to not complicate and modify the code downstream.

anthrotype: no problem, thanks

Lib/fontTools/cffLib/specializer.py
```diff
@@ -47,7 +47,10 @@ def programToCommands(program):
 					       'endchar'}:
 			parity = token in {'hmoveto', 'vmoveto'}
 			if (len(stack) % 2) ^ parity:
-				width = stack.pop(0)
+				try:
```
behdad: For width, since it's optional, I'll just change the if line like this:
```
if stack and (len(stack) % 2) ^ parity:
```
Then you wouldn't need the try except.

Lib/fontTools/cffLib/specializer.py
```diff
@@ -97,10 +100,12 @@ def vmoveto(args):
 
 	@staticmethod
 	def rlineto(args):
+		if len(args) == 0: raise ValueError(args)
```
behdad: I prefer `if not args:` but that might be just me.

Lib/fontTools/cffLib/specializer.py
```diff
@@ -25,7 +25,7 @@ def programToString(program):
 	return ' '.join(str(x) for x in program)
 
 
-def programToCommands(program):
+def programToCommands(program, ignoreErrors=True):
```
behdad: I don't understand why you made this change.

Lib/fontTools/cffLib/specializer.py
behdad: I'm not sure we need to modify this part of code at all.

Lib/fontTools/cffLib/specializer.py
```diff
@@ -220,14 +223,15 @@ def generalizeCommands(commands, ignoreErrors=True):
...
+				if len(args):
+					result.append(('', args))
+					result.append(('', [op]))
 			else:
 				raise
 	return result
 
 def generalizeProgram(program, **kwargs):
-	return commandsToProgram(generalizeCommands(programToCommands(program), **kwargs))
+	return commandsToProgram(generalizeCommands(programToCommands(program, **kwargs), **kwargs))
```
behdad: This doesn't work, as generalizeCommands() takes arguments that programToCommands() doesn't take.

Lib/fontTools/cffLib/specializer.py
```diff
@@ -220,14 +223,15 @@ def generalizeCommands(commands, ignoreErrors=True):
 			if ignoreErrors:
 				# Store op as data, such that consumers of commands do not have to
 				# deal with incorrect number of arguments.
-				result.append(('', args))
-				result.append(('', [op]))
+				if len(args):
```
behdad: I don't like this either.  The way I had written the code, it could just passthrough erratic bytecode without changing it.  That was intentional.

behdad:
I pushed a few commits that should enforce minimum number of arguments for all supported operators.  Can you please rework your patches on top of that?  Thanks.

Lib/fontTools/cffLib/specializer.py
miguelsousa: Nice!

Lib/fontTools/cffLib/specializer.py
miguelsousa: I do too, but decided to err on the conservative side because checks like these sometimes contain nuances.

Lib/fontTools/cffLib/specializer.py
miguelsousa: It was needed for the try/except that is now gone.

Lib/fontTools/cffLib/specializer.py
miguelsousa: Yeah, that was one of the changes I wasn't proud of.

Lib/fontTools/cffLib/specializer.py
miguelsousa: > That was intentional.

I'm aware of that. But I'm not sure how some test cases can pass without it. I should have just committed the changes to the tests and let you deal with the failures 🤣

miguelsousa:
Sure will.

miguelsousa:
> Maybe we should make these data driven. [...]

Not quite sure what you have in mind. You'll need to show me a snippet which I'll then expand.
In this new setup, will it still be possible to determine if something failed during the generalizing or specializing phases independently? I think that's an important feature.

I'll make spot changes now, and revamp the specializer_test.py file at a later date.

miguelsousa:
@behdad the new commits are in. Please address the failures. Thanks.

behdad:
> Please address the failures.

I see.  So, you expect ignoreErrors=True to drop the wrong operators.  My thinking model was to leave them alone.  While it's trivial what to do when there are zero arguments, it's less clear when, say, there's an rrcurveto with 4 arguments.  What should we do?  I still think the best we can do upon error is just passthru the input. WDYT?

miguelsousa:
I think it's not that useful to output invalid charstrings for cases that we know are definitely invalid, so my interpretation of ignoreErrors=True is more along the lines of "I'm going to silently ignore the things that are not according to spec".

anthrotype:
what's the status of this PR? Btw, I see the master branch has 5 failing tests.

behdad:
> I think it's not that useful to output invalid charstrings for cases that we know are definitely invalid, so my interpretation of ignoreErrors=True is more along the lines of "I'm going to silently ignore the things that are not according to spec".

It's my philosophy that libraries should NOT silently ignore errors.  Indeed, passing errors untouched allows for much simpler error handling.  Which is, you can build a pipeline and only check for errors at the output stage, or whenever convenient.

As such, I suggest we change the `ignoreErrors` to `errors` and model it after `io.open`.  Ie. the default `errors=None` is equal to `errors='strict'` and will raise exceptions.  `errors='ignore'` will drop erronous operators (only when it's safe.  Ie. `hlineto` with zero operators; it should NOT drop operations when the argument count is wrong otherwise.  Ie. do NOT drop an `rrcurveto` with four arguments.).  Finally, a `errors='passthru'` will do what I had in mind, which is all erratic operations are encoded as is in the final output.

We can then test them separately.

miguelsousa:
I like that approach

miguelsousa:
> Can you or Cosimo implement it please?

I don't feel qualified to implement such functionality, so I'll defer to Cosimo.

behdad:
I'll propose a patch, and then we can fix the tests.

behdad:
Maybe we don't need these...

miguelsousa:
It still seems like a good thing to have, but I think we'll be fine until the day it eventually becomes clear it's really necessary.

PR #968, opened 2017-05-18
Cleanup work in cffLib.py

readroberts:
Remove specialized classes for CFF2: move logic into regular classes. To change a CFF table from CFF to CFF2, you now need only to set cff.major = 2, and move the Private Dict from the TopDict to a FontDict in an FDArray.
Removed global state: use instead a context class.

### DIFF
--- a/Lib/fontTools/cffLib/__init__.py
+++ b/Lib/fontTools/cffLib/__init__.py
@@ -183,7 +183,7 @@ class CFFFontSet(object):
 				fontName = "CFF2Font"
 				self.fontNames.append(fontName)
 				cff2GetGlyphOrder = self.otFont.getGlyphOrder
-				topDict = TopDict2(GlobalSubrs=self.GlobalSubrs, cff2GetGlyphOrder=cff2GetGlyphOrder)
+				topDict = TopDict(GlobalSubrs=self.GlobalSubrs, cff2GetGlyphOrder=cff2GetGlyphOrder)
 				self.topDictIndex = TopDictData(None, cff2GetGlyphOrder, None)
 			self.topDictIndex.append(topDict)
 			for element in content:
@@ -673,7 +673,7 @@ class TopDictData(TopDictIndex):
 		return dataLength
 
 	def produceItem(self, index, data, file, offset, size):
-		top = TopDict2(self.strings, file, offset, self.GlobalSubrs, self.cff2GetGlyphOrder)
+		top = TopDict(self.strings, file, offset, self.GlobalSubrs, self.cff2GetGlyphOrder)
 		top.decompile(data)
 		return top
 
@@ -1105,23 +1105,18 @@ class TableConverter(SimpleConverter):
 
 class PrivateDictConverter(TableConverter):
 	def getClass(self):
-		if isCFF2:
-			return PrivateDict2
-		else:
-			return PrivateDict
+		return PrivateDict
 
 	def read(self, parent, value):
 		size, offset = value
 		file = parent.file
-		if isCFF2:
-			priv = PrivateDict2(parent.strings, file, offset, parent)
-		else:
-			priv = PrivateDict(parent.strings, file, offset, parent)
+		priv = PrivateDict(parent.strings, file, offset, parent)
 		file.seek(offset)
 		data = file.read(size)
 		assert len(data) == size
 		priv.decompile(data)
 		return priv
+
 	def write(self, parent, value):
 		return (0, 0)  # dummy value
 
@@ -2129,7 +2124,6 @@ class BaseDict(object):
 
 class TopDict(BaseDict):
 
-	defaults = buildDefaults(topDictOperators)
 	converters = buildConverters(topDictOperators)
 	compilerClass = TopDictCompiler
 	order = buildOrder(topDictOperators)
@@ -2139,6 +2133,14 @@ class TopDict(BaseDict):
 		BaseDict.__init__(self, strings, file, offset)
 		self.cff2GetGlyphOrder = cff2GetGlyphOrder
 		self.GlobalSubrs = GlobalSubrs
+		if isCFF2:
+			if cff2GetGlyphOrder == None:
+				import pdb
+				pdb.set_trace()
+			self.defaults = buildDefaults(topDictOperators2)
+			self.charset = cff2GetGlyphOrder()
+		else:
+			self.defaults = buildDefaults(topDictOperators)
 
 	def getGlyphOrder(self):
 		return self.charset
@@ -2180,16 +2182,7 @@ class TopDict(BaseDict):
 				progress.increment(0)  # update
 			i = i + 1
 
-class TopDict2(TopDict):
-	defaults = buildDefaults(topDictOperators2)
-
-	def __init__(self, strings=None, file=None, offset=None, GlobalSubrs=None, cff2GetGlyphOrder = None):
-		BaseDict.__init__(self, strings, file, offset)
-		self.cff2GetGlyphOrder = cff2GetGlyphOrder
-		self.charset = cff2GetGlyphOrder()
-		self.GlobalSubrs = GlobalSubrs
-
-class FontDict(TopDict):
+class FontDict(BaseDict):
 	#
 	# Since fonttools used to pass a lot of fields that are not relevant in the FDArray
 	# FontDict, there are 'ttx' files in the wild that contain all these. These got in
@@ -2217,15 +2210,17 @@ class FontDict(TopDict):
 	# - https://github.com/fonttools/fonttools/issues/601
 	# - https://github.com/adobe-type-tools/afdko/issues/137
 	#
-	order = ['FontName', 'FontMatrix', 'Weight', 'Private']
+	converters = buildConverters(topDictOperators)
 	compilerClass = FontDictCompiler
+	order = ['FontName', 'FontMatrix', 'Weight', 'Private']
+	decompilerClass = TopDictDecompiler
 
 	def __init__(self, strings=None, file=None, offset=None, GlobalSubrs=None, topDict = None):
-		super(FontDict, self).__init__(strings, file, offset, GlobalSubrs, None)
+		super(FontDict, self).__init__(strings, file, offset)
 		self.topDict = topDict
+		self.defaults = {}
 
 class PrivateDict(BaseDict):
-	defaults = buildDefaults(privateDictOperators)
 	converters = buildConverters(privateDictOperators)
 	order = buildOrder(privateDictOperators)
 	decompilerClass = PrivateDictDecompiler
@@ -2234,15 +2229,13 @@ class PrivateDict(BaseDict):
 	def __init__(self, strings=None, file=None, offset=None, parent= None):
 		super(PrivateDict, self).__init__(strings, file, offset)
 		self.fontDict = parent
-		self.isCFF2 = False
-
-class PrivateDict2(PrivateDict):
-	defaults = buildDefaults(privateDictOperators2)
-
-	def __init__(self, strings=None, file=None, offset=None, parent= None):
-		super(PrivateDict2, self).__init__(strings, file, offset, parent)
+		self.isCFF2 = isCFF2
 		self.vstore = None
 		self.isCFF2 = True
+		if self.isCFF2:
+			self.defaults = buildDefaults(privateDictOperators2)
+		else:
+			self.defaults = buildDefaults(privateDictOperators)
 
 	def getNumRegions(self, vi = None):
 		# if getNumRegions is being called, we can assume that VarStore exists.
@@ -2255,13 +2248,6 @@ class PrivateDict2(PrivateDict):
 				vi = 0
 		numRegions = self.vstore.getNumRegions(vi)
 		return numRegions
-
-	def decompile(self, data):
-		log.log(DEBUG, "    length %s is %d", self.__class__.__name__, len(data))
-		dec = self.decompilerClass(self.strings, self)
-		dec.decompile(data)
-		self.rawDict = dec.getDict()
-		self.postDecompile()
 
 class IndexedStrings(object):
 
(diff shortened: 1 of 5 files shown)


Lib/fontTools/cffLib/__init__.py
```diff
@@ -164,44 +182,37 @@ def fromXML(self, name, attrs, content, otFont = None):
...
-				subrCharStringClass = psCharStrings.T2CharString
-				if not hasattr(self, "GlobalSubrs"):
-					self.GlobalSubrs = GlobalSubrsIndex()
-			elif isCFF2:
-				subrCharStringClass = psCharStrings.CFF2SubrCharString
-				if not hasattr(self, "GlobalSubrs"):
-					self.GlobalSubrs = GlobalSubrsIndex2()
-			else:
-				assert False, "Unsupported CFF format"
+			subrCharStringClass = psCharStrings.T2CharString
```
adrientetar: why do we always use t2 (so CFF1) charstring now?

readroberts:
We could certainly rename this class to just CharString - which would have the advantage of matching the class name with is output to ttx XML. The only reason not to is that existing code which imports this class will break. However the changes to several class API's in the last month would likely break old code as well, so this weighs less than it usually would. Left to myself, I wouldn't make the change, but if anyone cares deeply, I wouldn't object.

Lib/fontTools/ttLib/tables/C_F_F_.py
```diff
@@ -44,4 +44,4 @@ def toXML(self, writer, otFont, progress=None):
 	def fromXML(self, name, attrs, content, otFont):
 		if not hasattr(self, "cff"):
 			self.cff = cffLib.CFFFontSet()
-		self.cff.fromXML(name, attrs, content)
+		self.cff.fromXML(name, attrs, content, otFont)
```
behdad: Why is this line in this commit?  Is this a bug fix?

Lib/fontTools/ttLib/tables/C_F_F__2.py
```diff
@@ -1,47 +1,8 @@
...
-
-	def toXML(self, writer, otFont, progress=None):
-		self.cff.toXML(writer, progress)
-
-	def fromXML(self, name, attrs, content, otFont):
-		if not hasattr(self, "cff"):
-			self.cff = cffLib.CFFFontSet()
-		self.cff.fromXML(name, attrs, content, otFont)
+class table_C_F_F__2(table_C_F_F_):
+	pass
```
behdad: How is it communicated to cffLib that this is a CFF2 table then?

Lib/fontTools/misc/psCharStrings.py
```diff
@@ -1002,9 +1000,12 @@ def draw(self, pen):
 		self.width = extractor.width
 
 	def check_program(self, program):
-		assert self.program, "illegal CharString: decompiled to empty program"
-		assert self.program[-1] in ("endchar", "return", "callsubr", "callgsubr",
-				"seac"), "illegal CharString"
+		if self.isCFF2:
+			if self.program:
+				assert self.program[-1] not in ("seac",), "illegal CFF2 CharString Termination"
```
behdad: Why not also check for 'endchar' and 'return'?

Is 'seac' allowed in CFF 1?

Lib/fontTools/misc/psCharStrings.py
```diff
@@ -1002,9 +1000,12 @@ def draw(self, pen):
...
+		else:
+			assert self.program, "illegal CharString: decompiled to empty program"
```
behdad: Does this matter?  How is this illegal?!

behdad: and how is this CFF 1.0-specific?

Lib/fontTools/misc/psCharStrings.py
```diff
@@ -1042,6 +1043,10 @@ def compile(self):
 			raise
 		self.setBytecode(bytecode)
 
+		if self.isCFF2:
+			if self.bytecode and (byteord(self.bytecode[-1]) in (11, 14)):
+				self.bytecode = self.bytecode[:-1]
```
behdad: As I mentioned earlier, I don't think we should do this transparently.  We should have methods for converting from CFF1 to CFF2 that take care of these, instead of doing them automatically.  I'm fine if you fix this later though.

behdad: Also, at least add a comment re what 11, 14 are.

Lib/fontTools/cffLib/__init__.py
behdad: We can rename it to CharString when we move it inside cffLib, and keep old module import it for compat.

Lib/fontTools/misc/psCharStrings.py
```diff
@@ -251,23 +251,29 @@ class CharStringCompileError(Exception): pass
...
 		self.operandStack = []
 		self.hintCount = 0
 		self.hintMaskBytes = 0
+		self.numRegions = 0
 
 	def check_program(self, program):
-		assert program, "illegal CharString: decompiled to empty program"
-		assert program[-1] in ("endchar", "return", "callsubr", "callgsubr",
-				"seac"), "illegal CharString"
+		if self.private and self.private.isCFF2:
```
behdad: Why is this block duplicated?  It was also in an earlier commit, which I commented on.  Same issues here.

Lib/fontTools/misc/psCharStrings.py
```diff
@@ -251,23 +251,29 @@ class CharStringCompileError(Exception): pass
 
 class SimpleT2Decompiler(object):
 
-	def __init__(self, localSubrs, globalSubrs):
+	def __init__(self, localSubrs, globalSubrs, private):
 		self.localSubrs = localSubrs
 		self.localBias = calcSubrBias(localSubrs)
 		self.globalSubrs = globalSubrs
 		self.globalBias = calcSubrBias(globalSubrs)
+		self.private = private
```
behdad: So, we pass in private just to lookup isCFF2?  Can't we have isCFF2 passed in directly?

Lib/fontTools/misc/psCharStrings.py
```diff
@@ -1167,10 +1167,11 @@ class DictDecompiler(ByteCodeBase):
 
 	operandEncoding = cffDictOperandEncoding
 
-	def __init__(self, strings):
+	def __init__(self, strings, parent = None):
```
behdad: No whitespace in `parent=None`.

Lib/fontTools/cffLib/__init__.py
```diff
@@ -183,7 +183,7 @@ def fromXML(self, name, attrs, content, otFont = None):
 				fontName = "CFF2Font"
 				self.fontNames.append(fontName)
 				cff2GetGlyphOrder = self.otFont.getGlyphOrder
-				topDict = TopDict(GlobalSubrs=self.GlobalSubrs, cff2GetGlyphOrder=cff2GetGlyphOrder)
+				topDict = TopDict2(GlobalSubrs=self.GlobalSubrs, cff2GetGlyphOrder=cff2GetGlyphOrder)
```
behdad: Why keep a revert commit in the history?

Lib/fontTools/cffLib/__init__.py
```diff
@@ -18,47 +18,66 @@
...
-maxStackLimit = 513
+maxStackLimit = 513 
+# maxstack operator has been deprecated. max stack is now always 513.
+
+class CFFContext(object):
+	def __init__(self, fontSet):
+		self.fontSet = fontSet
+		self.varStore = None
+		
+	def __getattr__(self, name):
```
behdad: Use a property instead?

Lib/fontTools/cffLib/__init__.py
```diff
@@ -18,47 +18,66 @@
...
-
 		if self.major == 1:
 			self.offSize = struct.unpack("B", file.read(1))[0]
 			file.seek(self.hdrSize)
-			isCFF2 = False
-			self.fontNames = list(Index(file))
-			self.topDictIndex = TopDictIndex(file)
-			self.strings = IndexedStrings(file)
-			self.GlobalSubrs = GlobalSubrsIndex(file)
+			self.cffCtx.isCFF2 = False
```
behdad: Isn't this just recursively setting `self.major`?

Lib/fontTools/cffLib/__init__.py
```diff
@@ -18,47 +18,66 @@
...
 	minor:   B
 	hdrSize: B
 """
 
-defaultMaxStack = 193
-maxStackLimit = 513
+maxStackLimit = 513 
+# maxstack operator has been deprecated. max stack is now always 513.
+
+class CFFContext(object):
```
behdad: Instead of a context that carries nothing other than isCFF2, I would have just passed isCFF2, or maybe cffVersion down to all functions that **need it**.  Ie, probably just compile/decompile, and NOT constructors of most objects.  But I'm not sure how feasible that is.  Can be done at a later time too.

adrientetar: I'm thinking you could even go and use metaprogramming to create two different classes for CFF1 and CFF2. They would be different objects, but generated from the same code.

behdad:
> I'm thinking you could even go and use metaprogramming to create two different classes for CFF1 and CFF2. They would be different objects, but generated from the same code.

No.  We don't want separate classes.

Lib/fontTools/ttLib/tables/C_F_F_.py
readroberts: It is there because C_F_F_2.py inherits this method, and needs the otFont argument.

Lib/fontTools/ttLib/tables/C_F_F__2.py
readroberts: You can tell that CFF instance is CFF2 because self.cffCtx.isCFF2 == True, aka self.major == 2. With the refactoring, the CFF2 class now contains the union of the methods and fields from CFF and CFF2. My idea is that differentiation is now made only at compile time, when the decision is made to decide either into the CFF or CFF2 format, based on self.major ==2. I have not yet gone all the way; with this: if a CFF table was decompiled, you need to still manually re-arrange the FDArray and PrivateDICT location before being able to recompile it as CFF2.

Lib/fontTools/misc/psCharStrings.py
readroberts: Noted.

Lib/fontTools/misc/psCharStrings.py
readroberts: I pass in 'private' at this point because in psCharStrings.py::SimpleT2Decompiler also uses private.getNumRegions().

Lib/fontTools/cffLib/__init__.py
readroberts: I chose to pass cffCtx with the one field to allow further refactoring. I am considering passing several other fields that get widely used - topDict,  the current privateDict, and adding a getNumRegions that would call the same function on the current private dict. 

The reason to have cffCtx set in the constructors is that some actions, like snapshotting or using a pen, may behave differently if you want the data to act like CFF or like CFF2.

Lib/fontTools/cffLib/__init__.py
readroberts: Yes. Left over from when sCFF2  was a global variable.

miguelsousa:
@behdad @anthrotype can this be merged?

behdad:
I think so.  @anthrotype can decide and merge.

anthrotype:
I'm not ignoring this, though I'd like to find some time to review it in more depth. However, if you guys want to go ahead and merge, that's fine.

miguelsousa:
@anthrotype OK if I merge this?

anthrotype:
@miguelsousa sure

anthrotype:
Hm.. as I feared, changing the `__init__` of all the cffLib classes broke ufo2ft, and possibly compreffor too. We should address these before we can tag a new fonttools release.

miguelsousa:
@anthrotype does that mean @readroberts needs to make further changes to `fonttools`, or something else?

behdad:
> as I feared, changing the __init__ of all the cffLib classes broke ufo2ft

Why do we need to change those?  Can't we make CFF1 be default?  I agree such breakage is not good.

readroberts:
With the current code, there is a context object that gets passed all the way down the call chain. There is now a place to put any new context that may be needed for any future state specific logic. I thought it was worth breaking the code once in order to not have to do it again, and order to be able to depend on the existence of the context object.  That said, I think it should be possible to re-arrange the broken calls so that the new arguments take a default value, and current code will still work. I will take a look at doing this.

readroberts:
A quick review shows that I can revert the calls to the original arguments, plus a new argument with a default value.  For the classes functions that now reference the context object, when no context object is supplied,  I would now make a new  context object in the class __init__ that is appropriate for CFF1. This approach will not serve the purpose of allowing the single context object to be altered during execution and getting a unified change in behavior from all the CFF classes,  but if the CFF code continues to evolve to having more versions, then I think this will be easier than special casing the tests for the context fields to first test if the context object exists. 

This will look a little uglier, as a class __init__ that can now omit all but one or of several arguments will now have to specify a bunch of None arguments in order to be able to specify the context object, or remember to use keywords.

An example:
Latest version
class GlobalSubrsIndex(Index):
	def __init__(self, cffCtx, file=None, globalSubrs=None, private=None, fdSelect=None, fdArray=None):#   the cffCtx argment is expected to always be present.

A sample call from FontSet.fromXML()
    GlobalSubrsIndex(self.cffCtx)

Backwards compatible version:
class GlobalSubrsIndex(Index):
	def __init__(self, file=None, globalSubrs=None, private=None, fdSelect=None, fdArray=None, cffCtx):  
# To be useful, the cffCtx argument must always be present, but the code allows it to be omitted.
 A sample call from FontSet.fromXML()
   GlobalSubrsIndex(cffCtx = self.cffCtx

Please confirm that it the current commenters feel that that this would be a useful approach.

readroberts:
Whoops, didn't realize that a Python comment causes Markdown emphasis.

anthrotype:
Having a keyword argument for cffCtx that defaults to CFF1 sounds good to me.

FYI, I'll be away for a couple of weeks with limited connectivity, in case you don't hear from me.

miguelsousa:
@readroberts I've updated the markdown of your comment. Please check that everything is still correct.

PR #969, opened 2017-05-18
Varlib iup

behdad:
I ended up redoing this on my own.  It's similar in functionality to Jens's work in https://github.com/fonttools/fonttools/pull/943

I haven't implemented the mutator part and will borrow from his code, but probably won't add those methods in GlyphCoordinates themselves but keep in varLib, and later move all IUP-related code into varLib.iup modules.

Tests should be ported from his branch as well.

I plan to implement full IUP optimization soon.  Maybe I work on that instead of the mutator part first.

WIP.

### DIFF
--- a/Lib/fontTools/ttLib/tables/TupleVariation.py
+++ b/Lib/fontTools/ttLib/tables/TupleVariation.py
@@ -31,8 +31,8 @@ log = logging.getLogger(__name__)
 
 class TupleVariation(object):
 	def __init__(self, axes, coordinates):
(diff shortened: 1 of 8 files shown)


behdad:
cc @jenskutilek

behdad:
I've implemented IUP in mutator as well.  I think this can be merged and makes Jens's PR obsolete.  I'll work on optimal packing of contours in a few days or weeks.  Have to tend to some bugs now.

PR #970, opened 2017-05-19
CFF bounds recalculation

### LINKED ISSUE #958: CFF bounds recalculation
CFF bounds recalculation hasn't been implemented yet; see [_h_h_e_a.py#L83](https://github.com/fonttools/fonttools/blob/30b804003e700677e72522b78334c3a6e89dc82f/Lib/fontTools/ttLib/tables/_h_h_e_a.py#L83), [_v_h_e_a.py#L82](https://github.com/fonttools/fonttools/blob/30b804003e700677e72522b78334c3a6e89dc82f/Lib/fontTools/ttLib/tables/_v_h_e_a.py#L82), and ~~[_m_a_x_p.py#L47](https://github.com/fonttools/fonttools/blob/30b804003e700677e72522b78334c3a6e89dc82f/Lib/fontTools/ttLib/tables/_m_a_x_p.py#L47)~~ (Edited: no need to recalc `maxp` values, while we have to recalc bounds in `head`). In addition, it would be nice to update Top DICT's `FontBBox` too, whose incorrect values may cause hinting distortion.

mashabow:
Fix #958

- [x] CFF
- [x] head
- [x] hhea
- [x] vhea

### DIFF
--- a/Lib/fontTools/cffLib/__init__.py
+++ b/Lib/fontTools/cffLib/__init__.py
@@ -2320,20 +2320,19 @@ class TopDict(BaseDict):
 			i = i + 1
 
 	def recalcFontBBox(self):
-		bounds = None
+		fontBBox = None
 		for charString in self.CharStrings.values():
-			if not hasattr(charString, 'bounds'):
-				charString.recalcBounds()
-			if charString.bounds is not None:
-				if bounds is not None:
-					bounds = unionRect(bounds, charString.bounds)
+			bounds = charString.calcBounds()
+			if bounds is not None:
+				if fontBBox is not None:
+					fontBBox = unionRect(fontBBox, bounds)
 				else:
-					bounds = charString.bounds
+					fontBBox = bounds
 
-		if bounds is None:
+		if fontBBox is None:
 			self.FontBBox = self.defaults['FontBBox'][:]
 		else:
-			self.FontBBox = list(intRect(bounds))
+			self.FontBBox = list(intRect(fontBBox))
 
 
 class FontDict(BaseDict):
--- a/Lib/fontTools/cffLib/__init__.py
+++ b/Lib/fontTools/cffLib/__init__.py
@@ -4,6 +4,7 @@ from __future__ import print_function, division, absolute_import
 from fontTools.misc.py23 import *
 from fontTools.misc import sstruct
 from fontTools.misc import psCharStrings
+from fontTools.misc.arrayTools import unionRect
 from fontTools.misc.textTools import safeEval
 from fontTools.ttLib import TTFont
 from fontTools.ttLib.tables.otBase import OTTableWriter
@@ -101,6 +102,11 @@ class CFFFontSet(object):
 			# use current 'major' value to determine output format
 			assert self.major in (1, 2), "Unknown CFF format"
 			isCFF2 = self.major == 2
+
+		if otFont.recalcBBoxes and not isCFF2:
+			for topDict in self.topDictIndex:
+				topDict.recalcFontBBox()
+
 		if not isCFF2:
 			strings = IndexedStrings()
 		else:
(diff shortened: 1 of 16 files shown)


Lib/fontTools/cffLib/__init__.py
```diff
@@ -2231,6 +2237,20 @@ def decompileAllCharStrings(self, progress):
 				progress.increment(0)  # update
 			i = i + 1
 
+	def recalcFontBBox(self):
+		initialBounds = (float('inf'), float('inf'), float('-inf'), float('-inf'))
+		bounds = initialBounds
```
adrientetar: I feel like this could be None initially then below you would do

if bounds is not None:
    unionRect(..)
else:
    bounds = csBounds

Lib/fontTools/ttLib/tables/_h_h_e_a.py
```diff
@@ -32,56 +34,63 @@ class table__h_h_e_a(DefaultTable.DefaultTable):
...
+			for name in ttFont.getGlyphOrder():
+				cs = topDict.CharStrings[name]
+				if cs.bounds is None:
+					continue
+				boundsWidthDict[name] = math.ceil(cs.bounds[2]) - math.floor(cs.bounds[0])
+
+		advanceWidthMax = 0
+		minLeftSideBearing = float('inf')
+		minRightSideBearing = float('inf')
+		xMaxExtent = -float('inf')
```
adrientetar: Same here, those could be None then below in the for:

~~~
if xMaxExtent is None:
    minLeftSideBearing = lsb
    minRightSideBearing = rsb
    xMaxExtent = extent
else:
    minLeftSideBearing = max(minLeftSideBearing, lsb)
    minRightSideBearing = max(minRightSideBearing, rsb)
    xMaxExtent = max(xMaxExtent, extent)
~~~

I feel it's more pythonic this way.

Lib/fontTools/cffLib/__init__.py
mashabow: Hmm, you mean like this?

```diff
diff --git a/Lib/fontTools/cffLib/__init__.py b/Lib/fontTools/cffLib/__init__.py
index bece702d..51a0d334 100644
--- a/Lib/fontTools/cffLib/__init__.py
+++ b/Lib/fontTools/cffLib/__init__.py
@@ -2238,15 +2238,17 @@ class TopDict(BaseDict):
                        i = i + 1

        def recalcFontBBox(self):
-               initialBounds = (float('inf'), float('inf'), float('-inf'), float('-inf'))
-               bounds = initialBounds
+               bounds = None
                for charString in self.CharStrings.values():
                        if not hasattr(charString, 'bounds'):
                                charString.recalcBounds()
                        if charString.bounds is not None:
-                               bounds = unionRect(bounds, charString.bounds)
+                               if bounds is not None:
+                                       bounds = unionRect(bounds, charString.bounds)
+                               else:
+                                       bounds = charString.bounds

-               if bounds == initialBounds:
+               if bounds is None:
                        self.FontBBox = self.defaults['FontBBox'][:]
                else:
                        self.FontBBox = list(intRect(bounds))
```

adrientetar: yes

mashabow:
Could you please review this PR? @behdad

behdad:
> Could you please review this PR? @behdad

Will do.  Thanks.

We also should do performance tuning later on, for recalcBounds on both TrueType and CFF.  Right now they cause glyph shapes to be decompiled, which then needs to be recompiled...  That's very wasteful to say the least...  For TrueType, should be trivial to fetch them without decompiling at all.

Lib/fontTools/ttLib/tables/_h_h_e_a.py
```diff
@@ -65,6 +65,8 @@ def recalc(self, ttFont):
 			topDict = ttFont['CFF '].cff.topDictIndex[0]
 			for name in ttFont.getGlyphOrder():
 				cs = topDict.CharStrings[name]
+				if not hasattr(cs, 'bounds'):
```
behdad: This caching of bounds means that modifying a charstring and saving again doesn't get the bounds right.  I would go without caching.  WDYT?  I know it will be twice the work at least... I don't have a better solution right now.

behdad:
lgtm in general.  May be worth waiting for CFF2 cleanup branch to be merged and update this to go on top?  Thanks.

Lib/fontTools/ttLib/tables/_h_h_e_a.py
mashabow: I'm not sure how we should cache... I'd like to remove this for the moment.

mashabow:
All right. Thank you!

mashabow:
> May be worth waiting for CFF2 cleanup branch to be merged and update this to go on top?

I rebased on the current master. Besides, now we don't cache charString bounds.

anthrotype:
Later on we shall extend this to also support recalculating CFF2 bounds.

PR #971, opened 2017-05-22
Fix broken ttx tests

mt-caret:
Currently, running tests with the source distribution fails due to missing test font files.
This PR adds the missing files to the MANIFEST.in and fixes the issue.


anthrotype:
thanks!

Issue #448, opened 2016-01-06
[feaLib] Do not de-duplicate lookups


brawer:
Currently, feaLib is merging lookups that have the same content in oder to make the resulting font smaller. However, this space-saving optimization changes the semantics of the resulting feature file: OpenType layout engines do not execute the lookups in order; instead, they sort lookups by numeric ID. Therefore, @behdad recommends to remove this optimization.

anthrotype:
I just got bitten by this. I got unexpected behaviour when applying multiple features (in my case was frac and numr), because the order of the lookups in the compiled GSUB didn't match the order in which lookups had been explicitly defined in the feature file.
I wonder if it would be better to disable such an optimization...
BTW, does makeotf also do this?

khaledhosny:
I think any optimization that changes the order of lookups is dangerous since the order often has important semantics and must be preserved.

behdad:
Plus, the bytes are reused by compiler anyway...

justvanrossum:
@anthrotype can you confirm your problem is solved by removing lines 463 to 469 from feaLib/builder.py?

anthrotype:
yes, that's exactly what I did to fix it, but I wanted to check with others before committing anything. As far as I can tell, there seems to be unanimity to disable the optimization. I'm off at the moment, I can apply the patch tomorrow, or any of you can do it.

justvanrossum:
Yeah, that optimization is definitely wrong.

PR #976, opened 2017-05-30
[feaLib] don't de-duplicate lookups

anthrotype:
keep original lookup order and preserve the feature file's semantics

Fixes https://github.com/fonttools/fonttools/issues/448


anthrotype:
i'm going to also modify the inline comments that say "Exact same content as feature f1; lookup should be shared."

PR #980, opened 2017-06-05
Copy emptied Script and LangSys records between GSUB and GPOS if missing

moyogo:
This replicates VOLT’s and FontForge’s behaviour of adding empty Script and LangSys records if one of GPOS or GSUB is empty. Older versions of Uniscribe need Script and LangSys to be in both GPOS and GSUB for features to be applied in some cases.

See https://github.com/googlei18n/fontmake/issues/258

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -74,8 +74,8 @@ class Builder(object):
         # for table 'vhea'
         self.vhea_ = {}
 
-    def _copy_empty_script_records(self, src, dst):
-        """Copies emptied Script records from one src OT table
+    def _copy_empty_script_langsys_records(self, src, dst):
+        """Copies emptied Script and LangSys records from one src OT table
            to another dst OT table if they are not in the dst OT table.
         """
         dst_script_tags = set(srec.ScriptTag for srec in
@@ -94,6 +94,17 @@ class Builder(object):
             else:
                 dst_srec = [srec for srec in dst.ScriptList.ScriptRecord if
                             srec.ScriptTag == src_srec.ScriptTag][0]
+            dst_langsys_tags = set(lsrec.LangSysTag for lsrec in
+                                   dst_srec.Script.LangSysRecord)
+            for src_lsrec in src_srec.Script.LangSysRecord:
+                if src_lsrec.LangSysTag not in dst_langsys_tags:
+                    dst_lsrec = otTables.LangSysRecord()
+                    dst_lsrec.LangSysTag = src_lsrec.LangSysTag
+                    dst_lsrec.LangSys = otTables.LangSys()
+                    dst_lsrec.LangSys.ReqFeatureIndex = 0xFFFF
+                    dst_lsrec.LangSys.FeatureCount = 0
+                    dst_lsrec.LangSys.FeatureIndex = []
+                    dst_srec.Script.LangSysRecord.append(dst_lsrec)
             dst_srec.Script.LangSysCount = len(dst_srec.Script.LangSysRecord)
             dst.ScriptList.ScriptCount = len(dst.ScriptList.ScriptRecord)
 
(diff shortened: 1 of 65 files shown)


moyogo:
Actually, to be clear, if they are in one OT table but not the other, FontForge copies empty ScriptRecords, whereas VOLT copies empty ScriptRecords and LangSysRecords.

behdad:
I prefer to do the absolute minimum required by Win XP.  What other tools do is of little importance IMO.

anthrotype:
what's the status of this?

moyogo:
We need to find an environment where we can replicate the issue to see what is actually needed. I haven’t had a change to set up those tests yet.

behdad:
Given the status, what's the plan here?

khaledhosny:
What was the original issue? (I got lost in the fontmake issue linked above). I have a working Windows XP VM if that is of an help checking this.

moyogo:
@khaledhosny the original PR description says:
> This replicates VOLT’s and FontForge’s behaviour of adding empty Script and LangSys records if one of GPOS or GSUB is empty. Older versions of Uniscribe need Script and LangSys to be in both GPOS and GSUB for features to be applied in some cases.

But given that FontForge actually only copies empty ScriptRecords, we need to know if that is good enough for older versions of Uniscribe.

khaledhosny:
I understand this, but what is actually broken currently. In other words, how do I test this to verify what is enough?

moyogo:
Take an Arabic or Hebrew font, empty the Language records in one of the tables and see if the Lookups in the other table still work.

khaledhosny:
This comment in FontForge code might be interesting: https://github.com/fontforge/fontforge/blob/cec4a984abb41419bf92fc58e5de0170404f0303/fontforge/lookups.c#L309-L334
> Sergey Malkin from MicroSoft tells me:
> > Each shaping engine in Uniscribe can decide on its requirements for
    layout tables - some of them require both GSUB and GPOS, in some cases
    any table present is enough, or it can work without any table.
> >
> > Sometimes, purpose of the check is to determine if font is supporting
    particular script - if required tables are not there font is just
    rejected by this shaping engine. Sometimes, shaping engine can not just
    reject the font because there are fonts using older shaping technologies
    we still have to support, so it uses some logic when to fallback to
    legacy layout code.
> >
> > In your case this is Hebrew, where both tables are required to use
    OpenType processing. Arabic requires both tables too, Latin requires
    GSUB to execute GPOS. But in general, if you have both tables you should
    be safe with any script to get fully featured OpenType shaping.
>
> In other words, if we have a Hebrew font with just GPOS features they won't work,
and MS will not use the font at all. We must add a GSUB table. In the unlikely
event that we had a hebrew font with only GSUB it would not work either.
>
> So if we want our lookups to have a chance of executing under Uniscribe we
better make sure that both tables have the same script set.
>
> (Sergey says we could optimize a little: A Latin GSUB table will run without
a GPOS, but he says the GPOS won't work without a GSUB.)

khaledhosny:
I tested on Windows XP:
* Missing GSUB or GPOS: font is installed but Arabic text uses fallback font.
* GSUB and GPOS with empty ScriptRecords: font is installed but Arabic text uses fallback font.
* GSUB and GPOS with empty ScriptRecords, FeatureList and LookupList (FontForge output): Font is used for Arabic.
* GSUB and GPOS with empty `arab` ScriptRecord, FeatureList and LookupList (FontForge output): Font is used for Arabic.
* The reverse seems to be true.

So I think the minimum is to do what this PR is doing for non-DFLT scripts, as copying DFLT seems to be superfluous.

khaledhosny:
Or may be copy DFLT as well, just in case.

behdad:
Can we resurrect this? Or do we not care about WinXP anymore?

anthrotype:
> Or do we not care about WinXP anymore?

I'd say we close until someone does care enough to resurrect this.

PR #990, opened 2017-06-22
Add a snippet to remove LangSys from DFLT script

khaledhosny:
Such fonts violate the spec and OTS rejects them, this snippet should
help quickly fixing such fonts.

### DIFF
--- a/Snippets/fix-dflt-langsys.py
+++ b/Snippets/fix-dflt-langsys.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python
+
+import argparse
+import logging
+import os
+import sys
+
+from fontTools.ttLib import TTFont
+
+
+def ProcessTable(table):
+    found = set()
+
+    for rec in table.ScriptList.ScriptRecord:
+        if rec.ScriptTag == "DFLT" and rec.Script.LangSysCount != 0:
+            tags = [r.LangSysTag for r in rec.Script.LangSysRecord]
+            logging.info("Removing %d extraneous LangSys records: %s",
+                         rec.Script.LangSysCount, " ".join(tags))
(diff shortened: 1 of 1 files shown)


Snippets/fix-dflt-langsys.py
```diff
@@ -0,0 +1,87 @@
...
+
+    logformat = "%(levelname)s: %(message)s"
+    if args.silent:
+        logging.basicConfig(format=logformat, level=logging.DEBUG)
+    else:
+        logging.basicConfig(format=logformat, level=logging.INFO)
+
+    ProcessFiles(args.files)
+
+    return 0
```
anthrotype: You don't need to return 0 from main(), as sys.exit(None) achieves the same result.

khaledhosny: Thanks, didn’t know that.

behdad:
> Such fonts violate the spec and OTS rejects them

I still think that's a nonsense part of the spec and OTS should NOT enforce it.  Such fonts definitely work fine in all known implementations.

khaledhosny:
I’m impartial about this, but last time this issue was discussed @jfkthame wanted to keep the the check in OTS.

jfkthame:
There was discussion a little while ago on the OpenType list about modifying the spec to allow this, but I don't recall if a firm conclusion was reached. If/when the keepers of the spec agree to make such a change, I'd be quite happy to see the relevant check dropped from OTS.

PeterConstable:
Please clarify: to allow what?

jfkthame:
To allow the DFLT entry in the ScriptListTable to have a non-zero LangSysCount.

PR #996, opened 2017-07-05
Recalculate ValueFormat when merging SinglePos

### LINKED ISSUE #995: [varLib] cpsp feature is not variable
In the attached demo build folder, the master fonts contain 1 kern pair and 1 cpsp lookup. When building a variable font with varLib, the kerning pair becomes variable, but the cpsp value is static.

The cpsp values in the wdth -1 master has XPlacement = 1, XAdvance = 2, and in the wdth 0 master XPlacement = 40, XAdvance = 80. The variable font only contains the 40/80 values.

More precisely, it looks like there are entries for the 40 and 80 in the variation item store in GDEF (item index 0 and 1), but they are not used in the cpsp lookup (Lookup Index 0).

[Variable Test cpsp.zip](https://github.com/fonttools/fonttools/files/1124294/Variable.Test.cpsp.zip)

jenskutilek:
The ValueFormat of the ValueRecord in a SinglePos must be recalculated after merging, or the reference to the item store will be discarded when the font is compiled. Fixes #995.

### DIFF
--- a/Lib/fontTools/varLib/merger.py
+++ b/Lib/fontTools/varLib/merger.py
@@ -202,6 +202,7 @@ def merge(merger, self, lst):
 	if all(v.Format == 1 for v in lst) and all(self.Coverage.glyphs == v.Coverage.glyphs for v in lst):
 		self.Value = otBase.ValueRecord(valueFormat)
 		merger.mergeThings(self.Value, [v.Value for v in lst])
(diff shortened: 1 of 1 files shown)


behdad:
lgtm. Thanks.

behdad:
We should try getting a test for this in.

PR #1007, opened 2017-07-19
[cffLib/psCharStrings] remove cffCtx, pass down isCFF2

anthrotype:
As discussed in https://github.com/fonttools/fonttools/pull/968#issuecomment-309920007 and in private conversation with @behdad, here is an (admittedly huge) patch that should break the current deadlock, and allow us to keep backward compatibility with the old cffLib interface (so fontmake or robofont don't break).

* Removed `CFFContext`
* Added `isCFF2` argument to CFFFontSet.decompile/compile, used from respective ttLib classes
* Index classes get a `isCFF2` argument in constructor (used for decompiling); must be True/False if `file` argument is not None; it is stored as self._isCFF2 to support lazy loading
* Removed `TopDictData` class; reuse same `TopDictIndexCompiler` for both CFF and CFF2
* `CFFWriter` and all `*Compiler` classes get an `isCFF2` argument; defaults to the parent compiler's `isCFF2` attribute
* Removed `size` argument from `produceItem` method as unused and useless (`len(data)` is the same)
* psCharStrings: removed useless ByteCodeBase class
* A reference to the TopDict's VarStoreData is passed down to all the FontDicts' PrivateDict, so it can be used to get the number of regions while decompiling blend and vsindex operators

I know I should have broken this into smaller patches, but... 😬 

/cc @readroberts

### DIFF
--- a/Lib/fontTools/cffLib/__init__.py
+++ b/Lib/fontTools/cffLib/__init__.py
@@ -27,57 +27,41 @@ maxStackLimit = 513
 # maxstack operator has been deprecated. max stack is now always 513.
 
 
-class CFFContext(object):
-
-	def __init__(self, fontSet):
-		self.fontSet = fontSet
-		self.varStore = None
-
-	@property
-	def isCFF2(self):
-		return self.fontSet.major == 2
-
-	@isCFF2.setter
-	def isCFF2(self, value):
-		if value:
-			self.fontSet.major = 2
-		else:
-			self.fontSet.major = 1
-
-
 class CFFFontSet(object):
 
-	def __init__(self):
-		self.cffCtx = CFFContext(self)
-
-	def decompile(self, file, otFont):
+	def decompile(self, file, otFont, isCFF2=None):
 		self.otFont = otFont
 		sstruct.unpack(cffHeaderFormat, file.read(3), self)
-		assert self.major in (1, 2), (
-			"unknown CFF format: %d.%d" % (self.major, self.minor))
-		if self.major == 1:
+		if isCFF2 is not None:
+			# called from ttLib: assert 'major' as read from file matches the
+			# expected version
+			expected_major = (2 if isCFF2 else 1)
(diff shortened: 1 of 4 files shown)


readroberts:
I was concerned after a conversation with Behdad today that a vsindex defined in a PrivateDict would not be inherited as the default vsindex by the charstrings which reference the PrivateDict, but I see that is not the case. Looks good to me.

behdad:
Read,

FWIW, one reason we are removing the context is because it didn't work as you thought it does.  Since some structs are loaded lazily, and the decompilation depends on isCFF2 bit of the context, you cannot load CFF, change the major version, and save as CFF2.  That will confuse the lazy decompiler to think it should load CFF2 objects instead of CFF!  Hence why we removed the context and always pass down isCFF2 during compile and decompile, separately.

Lib/fontTools/cffLib/__init__.py
```diff
@@ -2310,6 +2304,7 @@ def decompileAllCharStrings(self, progress):
 				progress.increment(0)  # update
 			i = i + 1
```
miguelsousa: `i += 1`

miguelsousa: Or better yet, replace L2296-2297 with,
```python
for i, charString in enumerate(self.CharStrings.values()):
```
and remove L2305.

anthrotype: @miguelsousa I think i'll leave that old piece of code like that for now. I'm actually tempted to deprecate the progress bar interface altogether. BTW, do you know if anybody is using that? /cc @justvanrossum

justvanrossum: I am not, and I doubt anyone is. If there's a safe way to deprecate it then I'm fine with that.

miguelsousa: I'm not using it either

Issue #1001, opened 2017-07-13
[varLib] Reactivate shared tuples?


jenskutilek:
Since font variations have been [deactivated in WebKit](https://bugs.webkit.org/show_bug.cgi?id=173618) and the Safari Technology Preview on macOS 10.12, should we reactivate the shared tuple mechanism?

https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/ttLib/tables/TupleVariation.py#L517

Maybe the bug is already fixed, because the [current report for CoreText](https://rawgit.com/unicode-org/text-rendering-tests/master/reports/CoreText.html#GVAR-1) shows the test as OK.

behdad:
Good question..

behdad:
Sascha, would you be able to test?

jenskutilek:
A word of caution ... I tried to reactivate the code, but it can result in invalid gvar data.

What happens is that the size of the resulting data using private points is always compared with the compiled data using shared points for all points of a glyph. However, if any of the tuple variations don't use all glyph points (e.g. because they rely on the IUP mechanism), and the shared points version would save space, the ‘incomplete’ tuple is compiled with shared points, and thus loses its point indices. The compiled font will be invalid, and on decompilation, a traceback is thrown.

You can see this for yourself if you uncomment the line mentioned earlier and try to ttx the test file [Mutator_IUP.ttx](https://github.com/fonttools/fonttools/blob/master/Tests/varLib/data/master_ttx_varfont_ttf/Mutator_IUP.ttx).

jenskutilek:
#1009 reactivates the shared points logic and should also fix the bug described in the previous comment.

anthrotype:
@jenskutilek anything more to do about this issue, or we can close?

jenskutilek:
Oops, I missed this. We can close it.

PR #1009, opened 2017-07-20
[varLib] Reactivate shared points logic, bugfixes

jenskutilek:
Related to issue https://github.com/fonttools/fonttools/issues/1001, this reactivates the sharing of point indices in TupleVariations if doing so saves file size.

It also fixes the bug leading to invalid gvar data described in the issue.

### DIFF
--- a/Lib/fontTools/ttLib/tables/TupleVariation.py
+++ b/Lib/fontTools/ttLib/tables/TupleVariation.py
@@ -512,11 +512,6 @@ def compileTupleVariationStore(variations, pointCount,
 			axisTags, sharedTupleIndices, sharedPoints=None)
 		sharedTuple, sharedData, usesSharedPoints = v.compile(
 			axisTags, sharedTupleIndices, sharedPoints=allPoints)
-		# TODO: Apple macOS 10.9.5 (maybe also earlier) up to 10.12 had a bug
-		# that broke variations if the `gvar` table contains shared tuples.
-		# Apple will likely fix this in macOS 10.13. But for the time being,
-		# we never emit shared points although the result would be more compact.
-		# https://rawgit.com/unicode-org/text-rendering-tests/master/reports/CoreText.html#GVAR-1
 		if (len(sharedTuple) + len(sharedData)) < (len(privateTuple) + len(privateData)):
 			tuples.append(sharedTuple)
 			data.append(sharedData)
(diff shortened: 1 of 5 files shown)


anthrotype:
@jenskutilek you know that you don't need to wait for the CI, but can run the test locally before pushing?
Just make sure you have both python2.7 and python3.6 on your $PATH, then `pip install tox` and run the full test suite with just `tox`.

jenskutilek:
Thanks! I didn’t know how to run them all at once. I ran each test on its own and missed some other ones that failed ;)

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -514,8 +515,7 @@ def compileTupleVariationStore(variations, pointCount,
 		# Apple will likely fix this in macOS 10.13. But for the time being,
 		# we never emit shared points although the result would be more compact.
```
behdad: Shouldn't you also update the comment here?

behdad:
I've reviewed this and looks fine, though it doesn't look like the code finds the optimal set of points to share, right?  Isn't it always sharing the "all points"?  I thought the "all points" case doesn't even need to be encoded?  I need to go read the spec again.

behdad:
Checked the spec again.  I think we really want to find the point set that is referenced most and make that shared...

behdad:
I can take this over if there's no rush.  Otherwise, feel free to give it a try.  The IUP optimizer code then also needs to be updated to take into account the point sharing for maximal optimization.

jenskutilek:
You’re right, as it is, this checks only if all or no points should be shared. Sharing all points still saves some space in the file, because the number of points has to be stored only once, not for each variation. At least in my current test project (10 masters), that shaved off a few kB.

If determining the optimal point set to share is connected to the IUP optimization, that could really become a hard optimization problem. Sometimes a few points unneccessary for IUP could be added in a variation, so the point set gets shareable ...

One quick optimization could be to check not for all points, but for the points actually used. If the IUP optimization happened to select the same point set for all variations, it could save some space. Not sure how likely this is to happen just by accident.

jenskutilek:
I think this is low in my priorities right now, so if you could take over, it would be great. No need to hurry.

behdad:
> One quick optimization could be to check not for all points, but for the points actually used. If the IUP optimization happened to select the same point set for all variations, it could save some space. Not sure how likely this is to happen just by accident.

Yes, that was what I had in mind.  Ie. optimize each deltaset IUP separately, then see what point set saves most if shared.

behdad:
Jens, what's the status of this?

jenskutilek:
In my opinion, it could be merged as is. I don’t have time to work on the mentioned optimizations at the moment, but the current status would save some file size.

jenskutilek:
A small improvement on this is here: https://github.com/fonttools/fonttools/pull/1090

Issue #1011, opened 2017-07-21
[varLib] Clarification regarding avar segment maps


jenskutilek:
The [avar spec](https://www.microsoft.com/typography/otspec/avar.htm) says:

> Axis value maps can be provided for any axis, but are required only if the normalization mapping for an axis is being modified. If the segment map for a given axis has any value maps, then it must include at least three value maps: -1 to -1, 0 to 0, and 1 to 1. These value mappings are essential to the design of the variation mechanisms and are required even if no additional maps are specified for a given axis.

 1. Currently, `varLib._add_fvar_avar` doesn’t enforce this. Should it?

 2. When an axis doesn’t cover the full range from -1 to 1, it makes no sense to include the entries as described.

For example, take a wdth axis from Condensed to Normal. The mappings may be:

```
-1.0 to -1.0
-0.5 to -0.3
 0.0 to  0.0
```

What would an additional 1 to 1 mapping, as required by the spec, do? There is no master at 1.0 for this axis. If something has to be added to the segment map, it should be `1.0 to 0.0`, making the range from 0 to 1 effectively flat. If that is the intent of the spec, the wording should be changed.

PeterConstable:
Yes, varLib should enforce this to ensure fonts are valid. This is an old requirement, originating in [GX](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6avar.html):

"A segment must always have at least three correspondence pairs: minus one mapping to minus one, zero mapping to zero, and one mapping to one."

Your example considers a width axis from Condensed to Normal, but those values are defined only in "user" space, not normalized space. The 'avar' table is a transformation entirely within normalized space which is always defined as having axes going from -1 to +1. An 'fvar' table may happen to define an axis in which default and min or default and max are the same, in which case nothing changes between -1 and 0, or between 0 and +1. But that is a consideration in the mapping from "user" space to normalized space that is independent of the 'avar' transformation that happens entirely within normalized space.

jenskutilek:
I see ... in that case the normalization done in varLib may be buggy. If I understand it correctly, for my example the normalization should look like this:

```
User space   Design space    Normalized
 50       ->  50             -1.0 -> -1.0
 75       ->  85             -0.5 ->  0.4
100       -> 100              0.0 ->  1.0 (default)
100       -> 100              1.0 ->  1.0
```

Is that correct?

PeterConstable:
[...]

PR #1012, opened 2017-07-21
[varLib] avar required value maps

jenskutilek:
This adds the required value maps describes in #1011. Before merging, we must decide about the question in the issue regarding the spec intent.

It also changes the behaviour regarding empty avar segment maps:

If segment maps are present for any axis, DirectWrite currently requires to have explicit maps for all axes, even if they just contain the default entries. As can be verified in Chrome Canary on Windows Creators Update, the font is not loaded if any axis has an empty avar entry.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -97,7 +97,7 @@ def _add_fvar_avar(font, axes, instances):
 			interesting = True
 		if not axis.map:
 			# Deduce a default segment map for the axis
-			items = sorted(list(set([(axis.minimum, axis.minimum), (axis.default, axis.default), (axis.maximum, axis.maximum)])))
+			items = [(v, v) for v in sorted(list(set((axis.minimum, axis.default, axis.maximum))))]
 		else:
 			items = sorted(axis.map.items())
 		keys   = [item[0] for item in items]
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -96,6 +96,11 @@ def _add_fvar_avar(font, axes, instances):
 		interesting = True
 
 		items = sorted(axis.map.items())
+		if not axis.map:
+			# Deduce a default segment map for the axis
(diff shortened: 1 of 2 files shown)


Lib/fontTools/varLib/__init__.py
```diff
@@ -92,11 +92,15 @@ def _add_fvar_avar(font, axes, instances):
...
-		interesting = True
-
-		items = sorted(axis.map.items())
-		keys   = [item[0] for item in items]
+			interesting |= False
+		else:
+			interesting = True
+		if not axis.map:
+			# Deduce a default segment map for the axis
+			items = [(v, v) for v in sorted(list(set((axis.minimum, axis.default, axis.maximum))))]
```
anthrotype: `sorted` already returns a list (copy), no need to do `sorted(list(set(...)))`

jenskutilek: Thanks!

PeterConstable:
The DirectWrite behaviour mentioned -- requiring explicit maps for all axes -- is a known bug in DirectWrite (the OT spec does not require explicit maps for all axes) and should not influence any decisions.

jenskutilek:
So I guess most of this PR can be disregarded. I made a new PR: #1014

Lib/fontTools/varLib/__init__.py
```diff
@@ -92,10 +92,9 @@ def _add_fvar_avar(font, axes, instances):
 	for axis in axes.values():
 		curve = avar.segments[axis.tag] = {}
 		if not axis.map or all(k==v for k,v in axis.map.items()):
-			continue
-		interesting = True
-
-		items = sorted(axis.map.items())
+			interesting |= False
```
behdad: This line doesn't make sense.

jenskutilek: It was supposed to make sure segment entries are being added even if not different from the default mapping after the first neccessary segment map has been found. It was a workaround for the DirectWrite bug.

I see now it's buggy …

PR #1014, opened 2017-07-24
[varLib] Initialize the avar segment map with required default entries

jenskutilek:
All that remains of PR #1012

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -94,7 +94,7 @@ def _add_fvar_avar(font, axes, instances):
 		if not axis.map or all(k==v for k,v in axis.map.items()):
 			continue
 		interesting = True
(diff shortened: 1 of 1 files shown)


anthrotype:
mind also adding a test?

anthrotype:
In `Tests/varLib/varLib_test.py` we don't have a test with a designspace containing axis map elements, so that branch is not covered.
https://codecov.io/gh/fonttools/fonttools/src/2a67bddaeb59cab6d0ca48c88030a84759b2fdcc/Lib/fontTools/varLib/__init__.py#L94...120

jenskutilek:
Would it be preferable to add completely new test data for this, or to integrate into one of the existing data files?

anthrotype:
New specific data would be preferable

anthrotype:
I mean, you can reuse the test master fonts, but a new designspace with axes map elements would be nic.

PR #1016, opened 2017-07-25
ttx.py Clarify help text

davelab6:
Confusion reported at https://groups.google.com/forum/#!topic/fonttools/7ocZWHbsCLw

### DIFF
--- a/Lib/fontTools/ttx.py
+++ b/Lib/fontTools/ttx.py
@@ -4,9 +4,11 @@ usage: ttx [options] inputfile1 [... inputfileN]
     TTX -- From OpenType To XML And Back
 
     If an input file is a TrueType or OpenType font file, it will be
-       dumped to an TTX file (an XML-based text format).
-    If an input file is a TTX file, it will be compiled to a TrueType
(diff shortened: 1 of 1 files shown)


Lib/fontTools/ttx.py
```diff
@@ -4,9 +4,11 @@
     TTX -- From OpenType To XML And Back
 
     If an input file is a TrueType or OpenType font file, it will be
-       dumped to an TTX file (an XML-based text format).
-    If an input file is a TTX file, it will be compiled to a TrueType
-       or OpenType font file.
+       decompiled to an TTX file (an XML-based text format).
```
khaledhosny: While at it, you may as well fix the “**an** TTX” typo.

anthrotype:
oh, I thought this was the README.rst, not the ttx --help...
I would like to remove the mention to the `otf2ttf.py` snippet actually. It's currently just a quick hack, doesn't deserve to share the same help message of the venerable ttx...
Also, snippets are not installed, one can only find them if one clones the repository.
People who are interested can find them anyway.

PR #1019, opened 2017-07-26
Varlib iup optimal

behdad:
I think this is ready to be merged.  Test is failing because expected output needs to be updated.  I like to hear how it performs for people.  It's enabled by default in fontmake.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -210,11 +210,11 @@ def _all_interpolatable_in_between(deltas, coords, j, i, tolerance):
 
 	assert len(deltas) == len(interp)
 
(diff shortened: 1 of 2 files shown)


behdad:
cc @robmck-ms

jenskutilek:
Well, I’m back from vacation :) I have used this code in my current project, and it works well. The savings in this particular font are not as big as I imagined, though. That may depend on the design and the range spanned by the masters.

In this project, the weight ranges from Hairline to Black with an intermediate Regular master. The file size savings are only ~10 kB on a total TTF size of 280 kB.

But on a font with two optical size masters that were quite close in design, the automatic optimization saved as much as I could with a manual optimization, so it seems to work very well.

For general evaluation, we'd need to look at larger sets of test projects.

behdad:
Thanks @jenskutilek.  Glad to hear that it's working for you.

PR #1025, opened 2017-08-14
Add suport for AAT lookups

brawer:
Although there is more work to be done before fonttools can handle `morx` tables,
I’d like to merge the current state of the AAT work into master so it does not bitrot.

### DIFF
--- a/Lib/fontTools/misc/testTools.py
+++ b/Lib/fontTools/misc/testTools.py
@@ -37,6 +37,7 @@ def parseXML(xmlSnippet):
 class FakeFont:
     def __init__(self, glyphs):
         self.glyphOrder_ = glyphs
+        self.reverseGlyphOrderDict_ = {g:i for i,g in enumerate(glyphs)}
         self.lazy = False
         self.tables = {}
 
@@ -50,7 +51,7 @@ class FakeFont:
         return self.tables.get(tag, default)
 
     def getGlyphID(self, name):
-        return self.glyphOrder_.index(name)
+        return self.reverseGlyphOrderDict_[name]
 
     def getGlyphName(self, glyphID):
         if glyphID < len(self.glyphOrder_):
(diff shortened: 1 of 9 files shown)


adrientetar:
Curious why you're investing into implementing AAT at this point, aren't almost all fonts OpenType or Type1 nowadays?

twardoch:
The state tables model in AAT is said to be about 3x faster in processing than OpenType Layout, especially for complex scripts. It's kind of like the difference between a compiled (AAT) and an interpreted (OTL) language. 

Glyph layout based on the basis of a single font is a finite business, so it could be expressed using a static "cache", which morx largely is, but OpenType uses a dynamic model and queries all the active lookups on every glyph in a run. Of course a platform can implement a custom cache on their own, but that wouldn't be portable. 

On the other hand, AAT (esp. morx) already can serve for this, and has the benefit of being supported on the Apple platforms Mayberry natively, so there is an independent implementation which makes testing easier. 

AFAIU, OTL tables can be losslessly converted into morx state tables (though it may require automated addition of some glyphs). If such an automatic module was developed, it could give a significant speed boost for the processing. 

I could imagine makers of platforms like Android or Chrome welcoming such a development.

khaledhosny:
Even if one does not go full AAT model, some of its tables seem to be interesting. For example `opbd` table seems to be more useful from implementation perspective (doing line breaking and justification) than the corresponding GPOS tables that are more cumbersome to use (no wonder that they are virtually unused).

behdad:
> AFAIU, OTL tables can be losslessly converted into morx state tables (though it may require automated addition of some glyphs). If such an automatic module was developed, it could give a significant speed boost for the processing.

This statement is only partially true.  It's prohibitively hard to convert anything using the Indic shaper to such AAT fonts.  For Universal Shaping Engine, it might be possible, but would require encoding the Unicode data part of USE in the font, which negates the point of USE.  For Arabic, Hangul, Thai, and Latin/simple scripts this is possible, but then it's of questionable import, since code still has to be prepared for super-slow fonts and as such needs to cache shaping results the way everyone currently is doing.

PR #1026, opened 2017-08-16
[varLib] add tests for avar; always include default maps

anthrotype:
The patch looks huge because I also included the UFO sources used to generate the master interpolatable ttx used in the tests.
To review the actual code changes it's better to check the single commits' diff and messages, excluding the last big one that contains the UFOs.

Besides the avar tests, the only library change follows a comment by @jenskutilek (19c4b37#commitcomment-23458151) about making sure that all the axes (even those which don't have any `<map>` elements defined in a designspace) end up with a non-empty segment maps array containing all the three default maps {-1.0: -1.0, 0.0: 0.0, 1.0: 1.0}.

The latest OT spec says that axis value maps "are required only if the normalization mapping for an axis is being modified. "  However, when default maps are missing for an axis' segment, both CoreText and DirectWrite display no modification to that axis coordinates, hence the current workaround.

### DIFF
--- a/Lib/fontTools/varLib/models.py
+++ b/Lib/fontTools/varLib/models.py
@@ -182,7 +182,7 @@ class VariationModel(object):
 			axis = next(iter(loc))
 			value = loc[axis]
 			if axis not in axisPoints:
-				axisPoints[axis] = {0}
+				axisPoints[axis] = {0.}
 			assert value not in axisPoints[axis]
 			axisPoints[axis].add(value)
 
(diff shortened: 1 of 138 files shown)


Lib/fontTools/varLib/models.py
```diff
@@ -182,7 +182,7 @@ def getMasterLocationsSortKeyFunc(locations, axisOrder=[]):
 			axis = next(iter(loc))
 			value = loc[axis]
 			if axis not in axisPoints:
-				axisPoints[axis] = {0}
+				axisPoints[axis] = {0.}
```
behdad: Should the XML writer be changed to always write float?  Not saying that this change is wrong, but might be better to also review the XML bits.

anthrotype: makes sense. I did that in a1c2cf113785fb1c482c5b3f22c7adb4643627e3

Issue #1031, opened 2017-08-19
[AAT] Cannot decompile AppleChancery


brawer:
Current fonttools crashes when decompiling the `opbd` table of the Apple Chancery font (version 8.0d1e1 of 2013-02-06, shipping on macOS 10.12.6). Likewise on some other AAT fonts. The problem only seems to occur with fonts whose `opbd` table is stored in format 4.

PR #1032, opened 2017-08-19
[AAT] Fix decompilation of AppleChancery and other AAT fonts

brawer:
Fixes https://github.com/fonttools/fonttools/issues/1031.


behdad:
Care to explain?

brawer:
seek() takes an absolute offset, whereas the argument to getSubReader() is relative to the current offset.

behdad:
Right.  Thanks.  Doesn't really need the github URL in the code IMO...

Issue #153, opened 2014-08-22
"Supersplit" mode to split _g_l_y_f.ttx files into 1-file-per-TTGlyph?


davelab6:
The 'split' mode of TTX, with one table per file, is useful for version control systems, and where a single TTX file can be megabytes of text. 

However, the `Font._g_l_y_f.ttx` file can still be huge. 

In contrast, the `UFO` and `SFDir` formats have one file per glyph.

So I suggest a 'supersplit' mode where there is one `<TTGlyph>` element per file. 

The files could be named in part with the [UFOv3 glyphname-to-filename convention](http://unifiedfontobject.org/versions/ufo3/conventions/#common-user-name-to-file-name-algorithm), eg, `Font._g_l_y_f.A_.ttx`

RoelN:
Bumping this.

In the mean time, what is the way to extract PNG images from am sbix table? I can iterate over the strikes (`for strike in font['sbix'].strikes:`) but I'm a bit clueless on how to proceed.

What's the best way to find out these things without asking? There aren't any "let play with the sbix table" docs out there for TTX/FontTools, right?

Jolg42:
@RoelN I did that not so long ago, here is the script https://gist.github.com/Jolg42/f22676fdfba6dbfe4589aa30f0326fb4

anthrotype:
@RoelN I'm afraid we don't have much documentation for the library modules yet, but feel free to ask anytime!

The `ttx` command has `-z` option to specify "a bitmap data export option for `EBDT`" (and `CBDT`), so one can pass `-z extfile` and save the bitmaps to external files.
However, this doesn't work with `sbix` table yet.

If somebody wants to give it a try, it'd be glad to review a PR.

RoelN:
@Jolg42 Fantastic, thanks! That's very helpful.

@anthrotype Thanks for the invitation to ask. I'll play around with @Jolg42's script and hope to make some time for a PR in the near future!

anthrotype:
I guess a script like @Jolg42's could be added to the fonttools/Snippets folder, if you like. I see it's using python2-only idioms, so one should make it py2.py3 compatible first. PR anyone?

Jolg42:
@RoelN You're welcome!
@anthrotype 👍 No problem with that, but I am not an expert in Python so I can't help for making it compatible...

RoelN:
I don't mind giving that a shot. I also have some other small example scripts I could dust off!

anthrotype:
[...]

PR #1035, opened 2017-08-23
Add ability to split glyphs to 1 glyph per ttx file.  Issue #153

bamidei:
I needed 1 glyph per file in order to use this to manage fonts under source control without large files.  It addresses the issue mentioned in issue #153.


Lib/fontTools/misc/xmlReader.py
```diff
@@ -35,6 +36,7 @@ def __init__(self, fileOrPath, ttFont, progress=None, quiet=None):
 			self.quiet = quiet
 		self.root = None
 		self.contentStack = []
+		self.contentOnly = contentOnly
```
behdad: Did you notice `XMLReader.read(..., rootless=True)`?  That should do the same, no?

bamidei: It's not really the same.  

From reviewing the code, I gather that rootless option allows for skipping the <ttFont> root tag along with the version reading.  The contentOnly is intended to skip the parsing of the table tag and instead allow parsing of a table's content  from a subfile directly into an in-progress table.  This fits the situation we have with splitting a <glyf> table across many subfiles.  Each of the subfiles will need to parse it's content and contribute the content to the in-progress <glyf> table.  This situation is currently detected when the xmlReader encounters a level-2 node which has a src.  It knows that it's already inside of a table and is being asked to parse the table's content from a subfile.  This is contrasted to when it reads an entire table from a subfile when it is at a level-1 node and detects a src attribute.

An example of the output needing to be parsed is below:

Example <Fabrikat-Black._g_l_y_f> fil structure:

``` 
<ttFont ttLibVersion="3.15">
  <glyf>
    <!-- The xMin, yMin, xMax and yMax values
         will be recalculated by the compiler. -->
    <TTGlyph src="Fabrikat-Black._g_l_y_f._2e_n_o_t_d_e_f.ttx"/>
    <TTGlyph name=".null"/><!-- contains no outline data -->
    <TTGlyph ...>
  </glyf>
</ttFont>

```

Then, an example of one of the files which would need to be parsed with contentOnly:
 <Fabrikat-Black._g_l_y_f._2e_n_o_t_d_e_f.ttx>

```
<ttFont ttLibVersion="3.15">

  <TTGlyph name=".notdef" xMin="65" yMin="-60" xMax="495" yMax="770">
    <contour>
      <pt x="65" y="770" on="1"/>
      <pt x="95" y="770" on="1"/>
      <pt x="95" y="-60" on="1"/>
      <pt x="65" y="-60" on="1"/>
    </contour>
   <contour.../>
  </TTGlyph>
</ttFont>

```

anthrotype:
@bamidei thanks for your contribution!
I haven't fully reviewed the code yet, but at quick glance it looks in the right direction.
It would be nice if you could add some unit tests as well.

bamidei:
@anthrotype Sure.  I haven't looked at the unit test framework for this project yet, but I'll take a look when I next get a bit of time and add some appropriate unit tests for the additional functionality.

bamidei:
I added a test for the XMLReader to ensure that the contentonly continues to be supported.  I think that this is the most important thing to test with the modifications that I made.  the XMLReader is the component which is reused for all parsing.  The other changes that I made were specific to the glyf table.  I didn't see many tests in that area existing.

@anthrotype Please let me know if you have the opportunity to complete review of this submission and if there is anything else you'd like me to add in order to prepare for merging.

Thank you

bamidei:
@behdad @anthrotype Any chance we could move forward with this?  Anything left that I should do?

Thank you.

behdad:
I'm fine with general approach and implementation.  @anthrotype can do a full review and merge.  Thanks.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -126,17 +130,34 @@ def toXML(self, writer, ttFont, progress=None):
...
-				glyph.toXML(writer, ttFont)
-				writer.endtag('TTGlyph')
-				writer.newline()
+				if splitGlyphs:
+					path, ext = os.path.splitext(tablePath)
+					fileNameTemplate = path + ".%s" + ext
+					glyphPath = fileNameTemplate % nameToIdentifier(glyphName)
+					glyphWriter = xmlWriter.XMLWriter(glyphPath, idlefunc=idlefunc,
+				      newlinestr=newlinestr)
+					glyphWriter.begintag("ttFont", ttLibVersion=ttLibVersion)
```
anthrotype: I wonder, wouldn't it be better to have "glyf" as the root tag of the single TTGlyph ttx file rather than a generic "ttFont"?
Other tables could receive a similar treatment, and it'd good to keep a reference to the container element in the split file.

Lib/fontTools/ttLib/__init__.py
```diff
@@ -299,8 +300,11 @@ def saveXML(self, fileOrPath, progress=None, quiet=None,
 			if progress:
 				progress.set(i)
 			tag = tables[i]
-			if splitTables:
+			if splitTables or splitGlyphs:
```
anthrotype: when splitGlyphs is True here you're calling tagToIdentifier for all the tables, even though only glyf currently support the extra argument

Lib/fontTools/ttLib/__init__.py
```diff
@@ -324,7 +329,8 @@ def saveXML(self, fileOrPath, progress=None, quiet=None,
 		if not hasattr(fileOrPath, "write") and fileOrPath != "-":
 			writer.close()
 
-	def _tableToXML(self, writer, tag, progress, quiet=None):
+	def _tableToXML(self, writer, tag, progress, tablePath, ttLibVersion=None,
```
anthrotype: i think tablePath should be an optional keyword argument `tablePath=None`, then below you assert it's not None when tag is "glyf"

Lib/fontTools/ttLib/__init__.py
```diff
@@ -862,8 +870,8 @@ def _escapechar(c):
 		return hex(byteord(c))[2:]
 
 
-def tagToIdentifier(tag):
-	"""Convert a table tag to a valid (but UGLY) python identifier,
+def nameToIdentifier(name):
+	"""Convert a name to a valid (but UGLY) python identifier,
```
anthrotype: this docstring is too verbose, it's a straight copy paste from tagToIdentifier. Let's just point to the other one and mention the differences or what's the purpose of this one. Or simply remove the docstring.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -110,7 +112,9 @@ def compile(self, ttFont):
 			ttFont['maxp'].numGlyphs = len(self.glyphs)
 		return data
 
-	def toXML(self, writer, ttFont, progress=None):
+	def toXML(self, writer, ttFont, progress=None, tablePath=None, 
+            ttLibVersion=None, idlefunc=None, newlinestr=None, splitGlyphs=None):
```
anthrotype: actually, can't we take all these extra values from the current writer instance?
- tablePath is writer.file.name
- ttLibVersion we can compute again from fonttools.version
- idlefunc is writer.idlefunc
- newlinestr is writer.newlinestr

I wouldn't like the change the API for tables' `toXML` methods. They are all predictably the same.
But I see that you need to tell the glyf table that you want to `splitGlyphs` so I see no way around adding a extra optional keyword argument `splitGlyphs=False` (not `None`)

Lib/fontTools/ttLib/__init__.py
anthrotype: scrap that, we can do without passing tablePath and the other extra arguments for the writer, see my other comment https://github.com/fonttools/fonttools/pull/1035#discussion_r149935588

Lib/fontTools/ttLib/__init__.py
justvanrossum: For this use case, it is of no importance that the result is a valid Python identifier. `.notdef` might as well translate to `2e_n_o_t_d_e_f`, without the leading underscore.

Lib/fontTools/ttLib/__init__.py
```diff
@@ -346,7 +352,9 @@ def _tableToXML(self, writer, tag, progress, quiet=None):
 			attrs['raw'] = True
 		writer.begintag(xmlTag, **attrs)
 		writer.newline()
-		if tag in ("glyf", "CFF "):
+		if tag in ("glyf"):
```
justvanrossum: How about
`if tag == "glyf":`

Lib/fontTools/ttLib/__init__.py
```diff
@@ -346,7 +352,9 @@ def _tableToXML(self, writer, tag, progress, quiet=None):
...
+			table.toXML(writer, self, progress, tablePath, ttLibVersion, idlefunc, newlinestr, splitGlyphs)
+		elif tag in ("CFF "):
```
justvanrossum: `tag == "CFF "`

Lib/fontTools/ttLib/__init__.py
ragingstorm: Right.  I can change the condition to if splitTables or (splitGlyphs && (tag == 'glyp')) so that the tagToIdentifier is only called when needed.

Lib/fontTools/ttLib/__init__.py
ragingstorm: I'll update the docstring as @anthrotype  suggests; @justvanrossum you are correct that the requirements aren't as strict for this case - however since we already have logic that creates identifiers which could be used, I didn't want to introduce another, separate, generator.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
ragingstorm: k. great.  Yes, I agree and I'll remove the unnecessary extra parameters and extract the values from the writer when they are needed inside of the glyp writing function.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
ragingstorm: How strongly do you feel about this suggestion?  My thought is that it is preferable to have every file generated by the ttx tools be normalized with a ttFont root tag with version information so we can verify / act if the file isn't generated by ttx tools or the 'right' version of ttx.  When a font is split between many different files, it is possible for different files to be generated with different versions of the tool (i.e. the files are disconnected and can be read/written separately).  

Additionally, it is also nice to be able to share the existing xmlReader which always expects the ttFont to be the root node with version.  

What would the use case for keeping a reference to the container element be?  The tables which are split do not have a reference back to the containing ttf font b/c they don't have a need for it.  The glyph's split from the glyp table have the similar situation - they don't have any need for information in the containing table.  The pattern which I attempted to follow seems to be that the readers responsibility is to read information and make available to their owner who puts the information in context (i.e. font -> table, glyp->glyph).

I can invest in making the change as you suggest, but it will be a bit larger change and I'm not sure that I understand the motivation.

Lib/fontTools/ttLib/__init__.py
ragingstorm: BTW - splitTables and splitGlyphs work well together - it is actually the use case which we always use.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
anthrotype: Maybe we could keep both the root `ttFont` element (with the ttLibVersion attribute) and the second-level container element (e.g. `glyf`).
The use case I see is mostly clarity for the reader, to better the hierarchy of the data. The TTGlyph element does not "belong" to ttFont directly, but to the glyf table, which in turns is contained in the ttFont.
If it's too much trouble to you adding this, you can leave it there and I'll pick it up once I find the time.
Thanks for you work!

bamidei: I think that this is good.  I made the necessary changes.  Now both ttFont and glyf are present in the generated glyf file.  I've updated the xmlReader test so that it expects this new format and passes.

Lib/fontTools/ttLib/__init__.py
bamidei: Sure. I did this.

Lib/fontTools/ttLib/__init__.py
bamidei: Updated this as well.

bamidei:
@anthrotype I've implemented all of the requested changes.  Assuming the build build/tests pass (they did locally), hopefully this is getting close to being able to merge.  Please let me know of anything else.

Lib/fontTools/ttLib/__init__.py
```diff
@@ -299,8 +300,11 @@ def saveXML(self, fileOrPath, progress=None, quiet=None,
 			if progress:
 				progress.set(i)
 			tag = tables[i]
-			if splitTables:
+			if splitTables or (splitGlyphs and (tag == 'glyf')):
```
anthrotype: the parentheses around `(tag == 'glyf')` are redundant

bamidei: ...but it makes order of operation explicit which can be nice for readability.

Lib/fontTools/ttx.py
```diff
@@ -38,6 +38,9 @@
        to the individual table dumps. This file can be used as
        input to ttx, as long as the table files are in the
        same directory.
+    -g Split glyf table: Save the glyf data into separate TTX files
+        per glyph and write a small TTX for the glyph table which
```
anthrotype: the glyf table, not the "glyph table"

Lib/fontTools/ttx.py
```diff
@@ -255,6 +261,7 @@ def ttDump(input, output, options):
 			tables=options.onlyTables,
 			skipTables=options.skipTables,
 			splitTables=options.splitTables,
+      splitGlyphs=options.splitGlyphs,
```
anthrotype: indentation looks wrong here compared to previous and next lines

Tests/misc/data/xmlSubFile.xml
```diff
@@ -0,0 +1,13 @@
...
+<ttFont ttLibVersion="3.15">
+
+  <TTGlyph name=".notdef" xMin="65" yMin="-60" xMax="495" yMax="770">
+    <contour>
+      <pt x="65" y="770" on="1"/>
+      <pt x="95" y="770" on="1"/>
+      <pt x="95" y="-60" on="1"/>
+      <pt x="65" y="-60" on="1"/>
+    </contour>
+   <contour.../>
```
anthrotype: looks like this test file "xmlSubFile.xml" is not used. if so, they we should remove it

Lib/fontTools/ttLib/__init__.py
anthrotype: well, it's obvious (to me at least) the precedence of `==` operator over `and`.

anthrotype:
tests failed for unrelated reasons (i'll get to fix that later).

Thanks again @bamidei :+1:

bamidei:
@anthrotype Thank you!

PR #1046, opened 2017-08-31
[AAT] Write AATLookups to XML in glyph order

brawer:
Before this change, the AAT lookups were alphabetically sorted by glyph name,
but it seems more natural to write XML in the usual glyph order. No changes to
the binary format (which already was in glyph order); this only affects how
XML output is produced.


behdad:
I *think* other tables write sorted by glyph name. The idea was to separate glyph order of the font from the XML.

brawer:
Ah, thanks for telling! Reverted.

PR #1049, opened 2017-09-06
[AAT] Support `cidg` table with CID-to-glyph mapping

brawer:


anthrotype:
You are on a roll lately! 👍

PR #1051, opened 2017-09-08
add parser for SVG paths supporting the pen protocol

anthrotype:
I was prompted by a conversation on the fonttools mailing list about ways to convert SVG paths to UFOs so I decided to write this small wrapper around an existing SVG path parser https://github.com/regebro/svg.path

I included a copy of the original svg.path parser here to avoid an external dependency, and also allow to modify the code so it directly uses a pen object instead of returning a list of Path objects.

The license is MIT so it's permissible and is compatible with fonttools BSD license.
There's also another fork of svg.path called [svgpathtools](https://github.com/mathandy/svgpathtools/blob/master/svgpathtools/parser.py) which has more feature besides parsing but is out of scope here.

There's a new module `fontTools.misc.svgPath` which exports an `SVGPath` class (for lack of better names). This takes a path to an SVG file containing path elements with `"d"` attributes and has a `draw` method that takes a FontTools segment pen object. This can be used to draw an SVG outline on any glyph object that supports the pen protocol (e.g. fontTools, robofab, defcon, fontParts, etc.).

One application would be a script that converts between SVG outlines and UFO GLIF files, using ufoLib.

Let me know what you think.


behdad:
I'd say make it fontTools.svgLib. We can also add other routines there later, for, for example, building SVG+OpenType font from bitmap color fonts or regular outline fonts, or the other way around, etc...

PR #1054, opened 2017-09-14
Graphite table support

mhosken:
This PR adds support for the Graphite font tables: Feat, Glat, Gloc, Silf, Sill providing ttx XML support as well.


brawer:
Nice! Do you plan adding unit tests? Something like [this](https://github.com/fonttools/fonttools/blob/master/Tests/ttLib/tables/_l_c_a_r_test.py), for example. Tests help a lot when maintaining a large codebase.

anthrotype:
Thank you Martin for working on this! 👍 
Yes, some simple tests would be nice to have.

one minor suggestion I was thinking of, that `fontTools/ttLib/tables/grUtils.py` module could named more explicitly `graphiteUtils` maybe?

Lib/fontTools/ttLib/tables/G__l_o_c.py
```diff
@@ -0,0 +1,77 @@
...
+from fontTools.misc import sstruct
+from fontTools.misc.textTools import safeEval
+from itertools import *
+from functools import partial
+from . import DefaultTable
+import array
+import struct
+import operator
+import warnings
+from _ast import Num
```
anthrotype: this import seems unused

behdad:
Are the specs for these tables available online?
Can these tables be implemented using the otData machinery?  I prefer not adding more hand-coded table parsing if we can avoid it.

mhosken:
The specs are available as part of the Graphite source tree in doc/GTF.txt. I did consider trying to use the otData machinery but it would have involved so many extensions to the core supporting types and a huge growth in what are already big files that I felt it better to use the old approach. I did try to use it for the Feat table, but gave up after it refused to accept any table with a version > 1! The otData machinery is very tightly tied together and hard to open up and as such is really only good for tables that are part of the OT machinery and that can make best use of the monolithic structure.

Now, if I were starting again on the Graphite binary format, I would probably go for something much closer in style to the OT structures to make use of such machinery. But even then I don't know how close I would get.

Lib/fontTools/ttLib/tables/G__l_o_c.py
mhosken: Cleared out the redundant imports there. I'll use graphiteUtils if you use opentypeUtils :) gr || ot just as graphite || opentype.

mhosken:
Test are in.

mhosken:
I'm done fiddling subject to fixing up review comments.

PR #1063, opened 2017-10-02
Respect SOURCE_DATE_EPOCH for time stamp

khaledhosny:
For reproducible builds, check the presence of SOURCE_DATE_EPOCH
environment variable and use it for the time stamp. This affects the
head.modified (and head.created in merge.py).

See https://reproducible-builds.org/specs/source-date-epoch/

### DIFF
--- a/Lib/fontTools/misc/timeTools.py
+++ b/Lib/fontTools/misc/timeTools.py
@@ -3,6 +3,7 @@
 
 from __future__ import print_function, division, absolute_import
 from fontTools.misc.py23 import *
+import os
 import time
 import calendar
 
@@ -47,6 +48,10 @@ def timestampFromString(value):
 	return calendar.timegm(time.strptime(value)) - epoch_diff
 
 def timestampNow():
+	# https://reproducible-builds.org/specs/source-date-epoch/
+	source_date_epoch = os.environ.get("SOURCE_DATE_EPOCH")
+	if source_date_epoch is not None:
+		return int(source_date_epoch) - epoch_diff
 	return int(time.time() - epoch_diff)
 
 def timestampSinceEpoch(value):
--- a/Tests/misc/timeTools_test.py
+++ b/Tests/misc/timeTools_test.py
@@ -1,9 +1,25 @@
 from __future__ import print_function, division, absolute_import
 from fontTools.misc.py23 import *
-from fontTools.misc.timeTools import asctime
+from fontTools.misc.timeTools import asctime, timestampNow, epoch_diff
+import os
 import time
 
 
 def test_asctime():
     assert isinstance(asctime(), basestring)
     assert asctime(time.gmtime(0)) == 'Thu Jan  1 00:00:00 1970'
+
+def test_source_date_epoch():
+    os.environ["SOURCE_DATE_EPOCH"] = "150687315"
+    assert timestampNow() + epoch_diff == 150687315
+
+    # Check that malformed value fail, any better way?
+    os.environ["SOURCE_DATE_EPOCH"] = "ABCDEFGHI"
+    try:
+        timestampNow()
+        assert False
+    except ValueError:
+        assert True
+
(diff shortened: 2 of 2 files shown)


Lib/fontTools/misc/timeTools.py
```diff
@@ -47,7 +48,12 @@ def timestampFromString(value):
 	return calendar.timegm(time.strptime(value)) - epoch_diff
 
 def timestampNow():
-	return int(time.time() - epoch_diff)
+	# https://reproducible-builds.org/specs/source-date-epoch/
+	source_date_epoch = os.environ.get("SOURCE_DATE_EPOCH")
+	if source_date_epoch is not None:
+		return int(int(source_date_epoch) - epoch_diff)
```
anthrotype: i believe `epoch_diff` is already an int, because that's what `calendar.timegm()` returns, so I think you don't need to call `int` twice.

anthrotype:
For the records, this is the linked ufo2ft discussion https://github.com/googlei18n/ufo2ft/issues/161

anthrotype:
Should we try to catch a possible `ValueError` exception, in case the SOURCE_DATE_EPOCH is malformed? I would fall back to the current time, and log a warning.

Lib/fontTools/misc/timeTools.py
```diff
@@ -47,7 +48,12 @@ def timestampFromString(value):
...
+	else:
```
adrientetar: else isn't needed when you have a return

Lib/fontTools/misc/timeTools.py
khaledhosny: The existing code did an `int` cast as well, since I couldn’t tell why I chose the safe said and did the same.

khaledhosny:
> Should we try to catch a possible ValueError exception, in case the SOURCE_DATE_EPOCH is malformed? I would fall back to the current time, and log a warning.

I did that, but on a second thought I’d rather let FontTools crash here. If someone set a malformed `SOURCE_DATE_EPOCH` (intentionally or accidentally) I’d rather let him know that loudly rather than silently ignore it.

Lib/fontTools/misc/timeTools.py
anthrotype: the reason the original code was doing that is because `time.time()` returns a float, and subtracting a int from a float yields a float.

khaledhosny: I just found this out 😄

anthrotype:
we could also add a basic test in `Tests/misc/timeTools_test.py`.
That test module is using pytest-style tests, so it's just a matter of have a function that starts with `test_*` and has a bunch of `assert` statements.

anthrotype:
I just noticed that the SOURCE_DATE_EPOCH spec says "If the value is malformed, the build process SHOULD exit with a non-zero error code." So crashing is good.

behdad:
Nice.

Tests/misc/timeTools_test.py
```diff
@@ -1,9 +1,25 @@
...
 
 def test_asctime():
     assert isinstance(asctime(), basestring)
     assert asctime(time.gmtime(0)) == 'Thu Jan  1 00:00:00 1970'
+
+def test_source_date_epoch():
+    os.environ["SOURCE_DATE_EPOCH"] = "150687315"
+    assert timestampNow() + epoch_diff == 150687315
+
+    # Check that malformed value fail, any better way?
```
anthrotype: @khaledhosny you can use `pytest.raises()`, as I did here https://github.com/fonttools/fonttools/commit/2e4b1d05df2d720bc1f4ba03b6dacb9171116ee1

More info https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

khaledhosny: Thanks!

PR #1067, opened 2017-10-06
Add interpolation of cvt/cvar table

jenskutilek:


behdad:
lgtm

jenskutilek:
I implemented the building of cvar tables from master cvt tables as well. The output is not optimized yet, it always writes out all cvts even if the delta is 0.

Merging cvt tables from the master TTFs may be risky. I'm not sure if there are any tools that produce hinted TTFs that may be safely variablized. VTT, which can hint variable TTFs, works on the variable fonts directly, not on the individual masters.

I basically only implemented this to make the tests for the mutator part of cvar/cvt handling pass.

My code only checks that all CVT tables have the same number of values. Is this okay for now, putting the burden on the user to make sure the results make sense?

jenskutilek:
Added a simple optimization. If all deltas in a cvar tuple are 0, the tuple can be omitted. This saves file size in any case.

Lib/fontTools/varLib/__init__.py
```diff
@@ -502,6 +503,30 @@ def _add_gvar(font, model, master_ttfs, tolerance=0.5, optimize=True):
...
+	num_cvts0 = len(allCVTs[0])
+
+	if (any(len(c) != num_cvts0 for c in allCVTs)):
+		log.warning("Master has incompatible cvt table, not building cvar table.")
+		del font["cvar"]
+		return
+	deltas = model.getDeltas(allCVTs)
+	supports = model.supports
+	for i,(delta,support) in enumerate(zip(deltas[1:], supports[1:])):
+		delta = [int(round(d)) for d in delta]
```
anthrotype: just `round` here is fine, don't need to cast to int as we are using the python3's round which always return an int when used with one argument

Lib/fontTools/ttLib/tables/_c_v_t.py
```diff
@@ -47,3 +48,34 @@ def __setitem__(self, index, value):
...
+		self.values = list(values)
+
+	def __getitem__(self, index):
+		return self.values[index]
+
+	def __len__(self):
+		return len(self.values)
+
+	def __repr__(self):
+		return u"CVTValues(%s)" % self.values
```
anthrotype: `__repr__` usually return `str` objects, not `unicode`. It's ascii only anyway

behdad:
I think we should assert that the contents of fpgm and prep tables as well as the hinting instructions in all master ttfs are the same.

Lib/fontTools/ttLib/tables/_c_v_t.py
behdad: Fixed.

jenskutilek:
I agree in principle, though maybe someone might want to apply glyph instructions from one master ttf to the whole variable font. But in that case a separate script could distribute the appropriate data to the other master ttfs before building the variable font.

behdad:
Is there are cvt's, I'm fine allowing fpm, prep, and glyph instructions only existing in base master. So, either just in base, or same in all masters.

jenskutilek:
I’ll have a go at it.

jenskutilek:
In progress: #1069

PR #1069, opened 2017-10-10
Merging TT hinting

jenskutilek:
In progress …

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -508,8 +508,9 @@ def _remove_TTHinting(font):
 	for tag in ("cvar", "cvt ", "fpgm", "prep"):
 		if tag in font:
 			del font[tag]
-	for attr in ("maxZones", "maxTwilightPoints", "maxStorage", "maxFunctionDefs", "maxInstructionDefs", "maxStackElements", "maxSizeOfInstructions"):
+	for attr in ("maxTwilightPoints", "maxStorage", "maxFunctionDefs", "maxInstructionDefs", "maxStackElements", "maxSizeOfInstructions"):
 		setattr(font["maxp"], attr, 0)
+	font["maxp"].maxZones = 1
 	glyf = font["glyf"]
 	for name in glyf.keys():
 		glyph = glyf[name]
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -538,24 +538,23 @@ def _merge_TTHinting(font, model, master_ttfs, tolerance=0.5):
 
 	# glyf table
 
-	#glyf = font["glyf"]
-	#for name in glyf.keys():
-	#	all_pgms = [
-	#		m["glyf"][name].program.getBytecode()
-	#		for m in master_ttfs
-	#		if hasattr(m["glyf"][name], "program")
-	#	]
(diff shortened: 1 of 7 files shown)


Lib/fontTools/varLib/__init__.py
```diff
@@ -503,22 +504,88 @@ def _add_gvar(font, model, master_ttfs, tolerance=0.5, optimize=True):
...
-	log.info("Generating cvar")
+	log.info("Merging TT hinting")
 	assert "cvar" not in font
+
+	# Check that the existing hinting is compatible
+
+	# fpgm and prep table
+
+	for tag in ("fpgm", "prep"):
+		all_pgms = [m[tag].program.getBytecode() for m in master_ttfs if tag in m]
```
anthrotype: ttProgam.Program class has `__eq__` method so in theory you should be able to compare two program instances as such, without needing to manually do getBytecode and then compare the bytes. you tried?

Lib/fontTools/varLib/__init__.py
```diff
@@ -503,22 +504,89 @@ def _add_gvar(font, model, master_ttfs, tolerance=0.5, optimize=True):
...
+		if tag in font:
+			del font[tag]
+	for attr in ("maxTwilightPoints", "maxStorage", "maxFunctionDefs", "maxInstructionDefs", "maxStackElements", "maxSizeOfInstructions"):
+		setattr(font["maxp"], attr, 0)
+	font["maxp"].maxZones = 1
+	glyf = font["glyf"]
+	for name in glyf.keys():
+		glyph = glyf[name]
+		if hasattr(glyph, "program"):
+			glyph.program = Program()
```
anthrotype: the `_g_l_y_f.Glyph` objects have a `removeHinting` method

robmck-ms:
Just wanting to doublecheck so I understand: Is the long-term intent here to compute copy the fpgm & prep tables and glyph programs forward (assuming they're identical in all masters), and then to compute a cvar from the master cvt tables? (i.e. the only things that really get merged are the cvt tables, and not do fancy things like consolidate delta hints under a GETVARIATION conditional, etc...).

behdad:
> Just wanting to doublecheck so I understand: Is the long-term intent here to compute copy the fpgm & prep tables and glyph programs forward (assuming they're identical in all masters), and then to compute a cvar from the master cvt tables? (i.e. the only things that really get merged are the cvt tables, and not do fancy things like consolidate delta hints under a GETVARIATION conditional, etc...).

The latter, while not impossible, would be very hard to get right and probably would generate huge hinting bytecode.  So I don't see us working on it anytime soon.  We'd be better off getting ttfautohint support varfonts than to merge bytecode.

jenskutilek:
Rob, you describe it correctly, no fancy things for now :) I just wanted to be able to generate an instance from a hinted VF with correct cvt, and add this instance as a new master to the VF and have the cvts correctly merged. That’s better than what varLib does now, which is just copying fpgm, prep, cvt, and glyf instructions from the base master verbatim when building a VF.

I doubt that automatically generating GETVARIATION instructions would have results that work without intervention in a VF-savvy hinting editor. I imagine deltas present in the master fonts would have to apply not only for each master in the VF, but for ranges of variations which have to be very carefully checked for the desired results.

If this branch is working correctly so far (I believe it is), I would leave it at this and merge it, if no-one has any objections.

behdad:
lgtm

PR #1071, opened 2017-10-11
add ReverseContourPen

anthrotype:
A filter pen to reverse the contour winding direction, similar to ufoLib's `ReverseContourPointPen`, but using segment pen API, so we don't need to use the segment/point pen converters.


behdad:
lgtm!

PR #1073, opened 2017-10-14
Add SVGPen

khaledhosny:
Copied and slightly adapted from:
https://github.com/typesupply/ufo2svg/blob/master/Lib/ufo2svg/svgPathPen.py

### DIFF
--- a/Lib/fontTools/pens/svgPen.py
+++ b/Lib/fontTools/pens/svgPen.py
@@ -2,47 +2,8 @@ from __future__ import print_function, division, absolute_import
 from fontTools.misc.py23 import *
 from fontTools.pens.basePen import BasePen
 
-# Number to String Conversion
-
-def valueToString(v):
-    """
-    >>> valueToString(0)
-    '0'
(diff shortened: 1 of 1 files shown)


behdad:
Nice!

anthrotype:
We shall add some round trip tests later

benkiel:
Would it be better to add these sorts of things to https://github.com/robofab-developers/fontPens so that they can be more easily shared?

anthrotype:
@benkiel well, fonttools is an install requirement of fontPens, and most other font-related python tools, so as far as "can be more easily shared", putting stuff in fonttools gains the most in terms of share-ability. 

I think fontPens is a good place for the robofab-like "point" pens (it's in fact a copy of the old robofab pens package), while fontTools.pens can keep its "segment" pens, at least for time being.

justvanrossum:
@anthrotype, I understand what you mean, but we're seeing a fairly arbitrary divide between `fontTools.pens` and `fontPens` now.

Issue #1074, opened 2017-10-15
Glyph side-bearing attributes


m4rc1e:
It would be nice to have these attributes in the glyf table or any other suitable table.

```
>>> font['glyf']['a'].lsb
50

>>> font['glyf']['a'].rsb
50
```

I've needed these attributes on a few occasions. My usual approach is:

```
lsb = font['glyf']['a'].xMin
rsb = font['hmtx'].metrics['a'][0] - font['glyf']['a'].xMax
```

I understand that glyph side bearings may be a concept of font editors and not any font spec.

More than happy to implement this and submit a pr if people approve.

khaledhosny:
The left side bearing and glyph advance are stored in the `hmtx` table, the right side bearing is the difference between x max and glyph advance.

khaledhosny:
```python
adv = font['hmtx'].metrics['a'][0]
lsb = font['hmtx'].metrics['a'][1]
rsb = adv - font['glyf']['a'].xMax
```

m4rc1e:
@khaledhosny Thanks, I pretty much the same thing in my example.

anthrotype:
For higher level stuff that's shared among glyf/CFF glyphs there is the ttLib._TTGlyph class. That already has both advance width and lsb. You could add a property that returns the rsb as well.

anthrotype:
That's the objects that you get when you do TTFont.getGlyphSet()

m4rc1e:
@anthrotype I see :-)

```
>>> a = font.getGlyphSet()['a']
>>> dir(a)
[..., 'draw', 'height', 'lsb', 'tsb', 'width']
```

Mind if I implement rsb?

anthrotype:
If don't mind if you do, thanks :)

anthrotype:
For consistency then you should also add a `bsb` (bottom side bearing) property if font has vmtx table.

Although I'm not super concinced of the necessity of these extra attributes at this level.. I think they are a bit fictitious, in the sense that they are not "there" in the font, but only derived from other data. It has to be clear to the user that there is no "right side bearing" value in the sfnt tables.

m4rc1e:
Closing due to discussion in #1075

PR #1075, opened 2017-10-15
[ttLib] add rsb attribute to _TTGlyph

m4rc1e:
As discussed in #1074. 

More than happy to add a test if needed.

### DIFF
--- a/Lib/fontTools/ttLib/__init__.py
+++ b/Lib/fontTools/ttLib/__init__.py
@@ -750,6 +750,10 @@ class _TTGlyph(object):
 		self._glyphset = glyphset
 		self._glyph = glyph
 		self.width, self.lsb = horizontalMetrics
+		if hasattr(self._glyph, "xMax"):
+			self.rsb = self.width - self._glyph.xMax 
+		else:
+			self.rsb = 0
 		if verticalMetrics:
 			self.height, self.tsb = verticalMetrics
 		else:


anthrotype:
well, it's good that we have some tests already
See https://travis-ci.org/fonttools/fonttools/jobs/288179180#L2939

sometime a glyf Glyph does not have xMax (empty glyphs). 
replace `self._glyph.xMax` with `self.glyph._xMax if hasattr(self._glyph, "xMax") else 0` or similar.

m4rc1e:
@anthrotype thanks, I've forced pushed the fix

anthrotype:
Hm, I wonder... When the xMax is 0 because the glyph has no contours nor components, its right side bearing is also equal to 0, or is it equal to the advance width?

Also, I think this won't work with CFF charstrings, does it? The latter don't have xMax or other bbox attributes

justvanrossum:
I'm fairly sure this implementation of `rsb` is not correct when `lsb != xMin`.

justvanrossum:
I don't think CFF charstrings have an `.lsb` attr even...

m4rc1e:
@justvanrossum Good point. I'll take a look now.

justvanrossum:
Like @anthrotype, I'm not super convinced of this addition. It's a higher level bit of info, which suggests it should also work with charstrings, which it cannot.

m4rc1e:
@justvanrossum @anthrotype Thank you for both raising your valid concerns. Yeah, side-bearings seem to be a concept of font editors.

Closing.

justvanrossum:
Actually, I was wrong, CFF-based fonts still must have an `hmtx` table, so the glyph objects do have an `lsb` attribute. That said, the _main_ reason that the glyph object carries the `lsb` around is so it can calculate an x offset for drawing a TT glyph to a pen.

PR #1079, opened 2017-10-20
Finish varLib.mutator

behdad:
WIP. Adds MVAR and avar.  To be done: GSUB/GPOS/GDEF.


behdad:
The GPOS part is not currently tested by our tests.  Otherwise, I think this can be merged.  Thanks.
@jenskutilek would be nice if you can test as well.

anthrotype:
I see you had a productive flight ;)

behdad:
Oh why did we have VarStore in GDEF in the test but no GPOS entries referring to them?

jenskutilek:
Nice work, thank you!

I found one problem in the instance `GPOS`, a lot of values are set to 0. E.g. the value of the `cpsp` feature lookup. This is in a font built of two masters with identical lookups for `cpsp` and `kern`. Some pair positioning values also come out as 0. But other pair positioning values come out of the interpolation correct.

There are also some weird things caused by TrueType hinting, components are moved around where they shouldn’t be. I will have to investigate this.

A couple of small things that would be nice-to-have:

* Do we care about the name table entries used for axes and instances? They could be removed, though checking if they are referenced from any other place (e.g. feature params ...) may be complicated.
* Building a synthetic style name based on the instance parameters would be nice for desktop use, but doesn't matter for web.
* When interpolating a font with a `slnt` or `ital` axis, maybe put the resulting angle in `post.italicAngle` (but which of both if they are different?).

I’ll do some more tests with a different font.

jenskutilek:
Some more findings:

* The `GDEF` table is retained in the instance font even if it’s empty.
* `OS/2.xAvgCharWidth` seems to be not recalculated
* `GPOS` valueFormat flags should be recalculated; e.g. in most cases the flag bits for `X_PLACEMENT_DEVICE `and `X_ADVANCE_DEVICE` will need to be cleared in the  instance.

The TrueType hinting problem is not easily fixed: The instructions in a composite glyph autohinted by VTT reset the components horizontal position (like in the ‘i’ in the image below). If the italic font is hinted in VTT on its own, the glyph hints are completely different. The hinting person would need to put in the italic hints in the default master in VTT to make this work.

bildschirmfoto 2017-10-23 um 18 04 55

I think that’s it for now.

jenskutilek:
Hm, on second thought, the TT hinting bug looks like a bug in VTT. It is also visible when an instance font is not generated, but when the variable font is installed in Windows and the variation instances are previewed in the WIndows font viewer.

behdad:
> Nice work, thank you!

Thanks.

> I found one problem in the instance GPOS, a lot of values are set to 0. E.g. the value of the cpsp feature lookup. This is in a font built of two masters with identical lookups for cpsp and kern. Some pair positioning values also come out as 0. But other pair positioning values come out of the interpolation correct.

Humm.  Can you email me the font?

>     Do we care about the name table entries used for axes and instances? They could be removed, though checking if they are referenced from any other place (e.g. feature params ...) may be complicated.

Pruning unused name table entries has been on my list.  I'll get started on that some time.

>     Building a synthetic style name based on the instance parameters would be nice for desktop use, but doesn't matter for web.

Right.  That also ties into https://github.com/LettError/designSpaceDocument/issues/8#issuecomment-334924289

>     When interpolating a font with a slnt axis, maybe put the resulting angle in post.italicAngle.
> 
> I’ll do some more tests with a different font.
> 
> Edit: I previously mentioned the ital axis in the last bullet point, but it is not mapped to the italic angle.

Right.  Yes, we can set from slant axis.

> Some more findings:
> 
>     The GDEF table is retained in the instance font even if it’s empty.

Will fix.

>     OS/2.xAvgCharWidth seems to be not recalculated

We never recalculate that one.  We discussed having an MVAR entry for that item but everyone agreed that value is useless...  If you thin kit should vary, then proposing to add it to MVAR is the right way.  cc @PeterCon

>     GPOS valueFormat flags should be recalculated; e.g. in most cases the flag bits for X_PLACEMENT_DEVICEand X_ADVANCE_DEVICE will need to be cleared in the instance.

Humm.  The code is there to do that.  Have to investigate why it's not happening.  Can you share font?

behdad:
BTW, did you notice that `slnt` is not linear and needs avar?  Ie. if you have a slanted master at -15, then without avar, what you get at slnt=-7.5 is not 7.5 degrees slanted, but `atan(tan(15*pi/180.)/2.)*180/pi = 7.630740212430058`.

jenskutilek:
Interesting, I hadn't noticed. I made a small script that will generate mappings to correct the differences: [avar-slant.py](https://gist.github.com/jenskutilek/f328f2ca250e8b9ebad0fb398a5a64c1)

behdad:
> Interesting, I hadn't noticed. I made a small script that will generate mappings to correct the differences: avar-slant.py in the format needed for a .designspace file.

Would be nice to put something like that somewhere others can use.  Maybe we can have a `<mapping auto="true" />` in designspace. cc @LettError

LettError:
Where does `<mapping auto="true">` go? `axis`? What does it do?

justvanrossum:
I feel it is overengineering to want to do this automatically. Yes, the angle is not interpolated linearly, but the error is only about 1 degree if the max slant is 30 degrees, and only about 0.13 degrees with a slant of 15. The type designer can decide if that's a problem or not, and fix it like Jens did if it is.

LettError:
So it would only be for the slant axis? Hmm.

jenskutilek:
Are there any other axes that would always need an automatic mapping? Not in the currently registered axes, I think.

jenskutilek:
At 15° maximum angle, a difference of 0.13° is just something over 1 unit difference over the height of a capital letter (700 units on 1000 upm).

LettError:
If it is such a general transformation, it should be added to the specification?

Issue #1055, opened 2017-09-15
[AAT] morx table decompilation fails for Ayuthaya font


jenskutilek:
Current version of Ayuthaya.ttf (macOS 10.12.6) throws an error:

```bash
  File "/Users/jens/Library/Python/2.7/lib/python/site-packages/fontTools/ttLib/__init__.py", line 426, in __getitem__
    table.decompile(data, self)
  File "/Users/jens/Library/Python/2.7/lib/python/site-packages/fontTools/ttLib/tables/otBase.py", line 42, in decompile
    self.table.decompile(reader, font)
  File "/Users/jens/Library/Python/2.7/lib/python/site-packages/fontTools/ttLib/tables/otBase.py", line 622, in decompile
    table[conv.name] = conv.readArray(reader, font, table, countValue)
  File "/Users/jens/Library/Python/2.7/lib/python/site-packages/fontTools/ttLib/tables/otConverters.py", line 145, in readArray
    l.append(self.read(reader, font, tableDict))
  File "/Users/jens/Library/Python/2.7/lib/python/site-packages/fontTools/ttLib/tables/otConverters.py", line 458, in read
    table.decompile(reader, font)
  File "/Users/jens/Library/Python/2.7/lib/python/site-packages/fontTools/ttLib/tables/otBase.py", line 622, in decompile
    table[conv.name] = conv.readArray(reader, font, table, countValue)
  File "/Users/jens/Library/Python/2.7/lib/python/site-packages/fontTools/ttLib/tables/otConverters.py", line 145, in readArray
    l.append(self.read(reader, font, tableDict))
  File "/Users/jens/Library/Python/2.7/lib/python/site-packages/fontTools/ttLib/tables/otConverters.py", line 458, in read
    table.decompile(reader, font)
  File "/Users/jens/Library/Python/2.7/lib/python/site-packages/fontTools/ttLib/tables/otBase.py", line 612, in decompile
    table["MorphType"])
  File "/Users/jens/Library/Python/2.7/lib/python/site-packages/fontTools/ttLib/tables/otConverters.py", line 534, in getConverter
    tableClass = self.lookupTypes[tableType][lookupType]
KeyError: 2
```

behdad:
@brawer

olivierberten:
converting ttx back to ttf fails with 
AssertionError: ('AATStateTable', 'LigatureMorph', 'MorxSubtable[2]', 'MorxChain[0]')

behdad:
[...]

PR #1082, opened 2017-10-25
[morx] Implement morx ligatures subtable

brawer:
This implements the AAT ligatures subtable (type 2). There is still one last subtable of `morx` that remain to be implemented afterwards. But the the `morx-ligatures` branch can probably be merged.

https://github.com/fonttools/fonttools/issues/1055


anthrotype:
This is great, thanks Sascha!
I can't assess the details of the implementation as I'm not familiar enough with AAT. If you think this can be merged, let's do it. I'd like to tag a new release today. We can always fix bugs later.

brawer:
It’s certainly an improvement over the current state. At some point I’d like to clean up the entire AAT handling (as you’ll see, it’s quite messy), but that can be done later. Merged this PR.

Issue #993, opened 2017-06-25
TypeError: Struct() argument 1 must be string, not unicode


banxian:
I have ttx via pip install fonttools, and use it to convert SourceHanSans.
but it give me an exception like this:

Dumping "SourceHanSansCN-Normal.otf" to "SourceHanSansCN-Normal.ttx"...
Dumping 'GlyphOrder' table...
Dumping 'head' table...
Dumping 'hhea' table...
Dumping 'maxp' table...
Dumping 'OS/2' table...
ERROR: Unhandled exception has occurred
Traceback (most recent call last):
  File "f:\python27\lib\site-packages\fontTools\ttx.py", line 387, in main
    process(jobs, options)
  File "f:\python27\lib\site-packages\fontTools\ttx.py", line 361, in process
    action(input, output, options)
  File "f:\python27\lib\site-packages\fontTools\misc\loggingTools.py", line 372, in wrapper
    return func(*args, **kwds)
  File "f:\python27\lib\site-packages\fontTools\ttx.py", line 258, in ttDump
    newlinestr=options.newlinestr)
  File "f:\python27\lib\site-packages\fontTools\ttLib\__init__.py", line 311, in saveXML
    self._tableToXML(tableWriter, tag, progress)
  File "f:\python27\lib\site-packages\fontTools\ttLib\__init__.py", line 329, in _tableToXML
    table = self[tag]
  File "f:\python27\lib\site-packages\fontTools\ttLib\__init__.py", line 419, in __getitem__
    tableClass = getTableClass(tag)
  File "f:\python27\lib\site-packages\fontTools\ttLib\__init__.py", line 825, in getTableClass
    module = getTableModule(tag)
  File "f:\python27\lib\site-packages\fontTools\ttLib\__init__.py", line 807, in getTableModule
    __import__("fontTools.ttLib.tables." + pyTag)
  File "f:\python27\lib\site-packages\fontTools\ttLib\tables\_n_a_m_e.py", line 26, in <module>
    nameRecordSize = sstruct.calcsize(nameRecordFormat)
  File "f:\python27\lib\site-packages\fontTools\misc\sstruct.py", line 108, in calcsize
    return struct.calcsize(formatstring)
TypeError: Struct() argument 1 must be string, not unicode
(Hit any key to exit)

python version is 2.7.3, os is windows 7 x64 sp1.

behdad:
Humm.  I suspect e8530360b, but why does it not fail in our continuous testing?

[...]

PR #1083, opened 2017-10-25
#993 - fix - TypeError: Struct() argument 1 must be string, not unicode

nef0s:


anthrotype:
thanks but this was already fixed with 9d08673c2fa142b9c578808a9a217b8229a7a7af and 3a97f425b850bd0cca482491ec6fd0327c75842f

anthrotype:
a new fonttools release that includes the fix will be published shortly on PyPI.

PR #1090, opened 2017-11-02
[varLib] Try a set of used points instead of all points

jenskutilek:
… when testing whether to share points between tuples.

The old code tried to share either all points of a glyph or None. The new version uses the used points instead of all points. This shaved 108 bytes off a 500 bytes cvar table where always the same 2 of 133 CVTs are varied.

### DIFF
--- a/Lib/fontTools/ttLib/tables/TupleVariation.py
+++ b/Lib/fontTools/ttLib/tables/TupleVariation.py
@@ -502,8 +502,7 @@ def compileTupleVariationStore(variations, pointCount,
 	# For the time being, we try two variants and then pick the better one:
 	# (a) each tuple supplies its own private set of points;
 	# (b) all tuples refer to a shared set of points, which consists of
-	#     "every control point in the glyph".
-	allPoints = set(range(pointCount))
+	#     "every control point in the glyph that has explicit deltas".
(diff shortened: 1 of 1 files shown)


anthrotype:
@jenskutilek does the preceding inline comment still apply to the new code, or does it also need updating?

anthrotype:
bonus point if there was an explicit test for this situation in TupleVariation_test.py

jenskutilek:
I've changed details in the comment, but left the bulk of it there. The finding of the shared points may be better optimized in the future. I'll add a test next week.

anthrotype:
thanks! I see all lines from that function are covered anyway. 
https://codecov.io/gh/fonttools/fonttools/src/84b423fd8deba327df2eb505392121a6e7b7a8ec/Lib/fontTools/ttLib/tables/TupleVariation.py#L506...534
Feel free to merge as is and perhaps add an explicit test later

jenskutilek:
I'm experiencing problems with a font that uses shared points in cvar, it will not display correctly in Chrome Canary on Windows. Investigating ...

jenskutilek:
Explicit tests for shared points/private points in cvar have been added in #1113.

PR #1091, opened 2017-11-02
[CFF2] Fixes for writing PrivateDict BlueValues and Subrs to XML.

readroberts:


Lib/fontTools/misc/psCharStrings.py
```diff
@@ -1102,6 +1102,15 @@ def toXML(self, xmlWriter):
 					args = []
 				else:
 					args.append(token)
+			if args:
```
anthrotype: I find a bit odd that a method from a super-class checks if `self` is an instance of one of the sub-classes..
Maybe the `if args: ...` branch could be refactored into a method `_checkLastStackItems` or a similar, and the `T2Charstring` default implementation would assert, whereas the `CFF2Subr` subclass would override it and write the extra arguments. Or maybe it's ok.

Lib/fontTools/misc/psCharStrings.py
```diff
@@ -1288,7 +1296,13 @@ def arg_delta(self, name):
...
 			current = 0
 			for v in valueList:
-				current = current + v
+				try:
+					current = current + v
+				except:
+					print(valueList)
+					import pdb
+					pdb.set_trace()
+					print("Hello")
```
anthrotype: Hello! :)

anthrotype:
> This update changes the ttx format so that in PrivateDict BlueValues, the default font values are written as absolute coordinates, preserving continuity with how CFF tables are written, but the region values are written as the deltas from the VariationStore delta list.

so basically this would invalidate all the CFF2 ttx files out there, right?
Shouldn't we try to detect this and somehow upgrade to the new notation?

readroberts:
You are correct this would invalidate all CFF2 ttx files out there However, any previous versions were already incorrect for any fonts with regions not at the end points of axes, and I don't think anyone has made shippable CFF2 fonts. My feeling is that this cost is acceptable at this time. However, if the consensus is that there should be a way to distinguish the new format, I am open to to doing so. Any suggestions? I am not aware of a current  ttx mechanism for doing this.

anthrotype:
> I don't think anyone has made shippable CFF2 fonts

If you guys says so (we certainly haven't).
I'm ok to break it, since this is all very new. At least, we could try to emit some warning when parsing a CFF2 table from a TTX file whose ttLibVersion is equal or less than 3.18. The problem is the XMLWriter does write a ttLibVersion, but the XMLReader throws it away... But we could fix that in an other PR

readroberts:
Adobe has not made shippable CFF2 fonts (the one road block left is that we  are still working on blend-compatible hinting). TheCFF2  variable fonts that are shipping now are clearly identified as proof of concept, largely because they are un-hinted.  I agree that the process is new, and  suggest that adding complexity to the XML format in order to avoid short-term problems is not worth living with for the next several decades. Emitting a warning if a CFF2 table is compiled to/decompiled from a font with a fonttools module with a ttLibVersion less that some value is a very good idea. I think the value needs to be greater than 3.18 - that is already tagged as a release.

Lib/fontTools/misc/psCharStrings.py
readroberts: Yes, this was inelegant. I thought a bit about minimum call overhead, and just added a class field 'isCFF2' that is False for T2Charstring and True for CFFCharstring. I chose 'isCFF2', as this is used a number of other class. If you prefer another, let me know, or just change it directly - I have no attachment to any particular solution.

PR #1092, opened 2017-11-03
[ttLib] Add a convenience method to the cmap table to return the best available cmap

justvanrossum:
Often I just want to quickly access the best available cmap from a font, without being interested (or having to worry about) exactly which cmap subtables are in the font. I've used this functionality for while now in my own code as a separate function, and it's been so handy that I think it should just be in fonttools.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_c_m_a_p.py
+++ b/Lib/fontTools/ttLib/tables/_c_m_a_p.py
@@ -40,7 +40,9 @@ class table__c_m_a_p(DefaultTable.DefaultTable):
 
 	def getBestCmap(self, cmapPreferences=((3, 10), (3, 1), (0, 3))):
 		"""Return the 'best' unicode cmap dictionary available in the font.
-		By default it will search for the following (platformID, platEncID) pairs:
+
+		By default it will search for the following (platformID, platEncID)
+		pairs:
 			(3, 10), (3, 1), (0, 3)
 		But this can be customized via the cmapPreferences argument.
 		"""
--- a/Lib/fontTools/ttLib/tables/_c_m_a_p.py
+++ b/Lib/fontTools/ttLib/tables/_c_m_a_p.py
@@ -39,7 +39,8 @@ class table__c_m_a_p(DefaultTable.DefaultTable):
 		return None # not found
 
 	def getBestCmap(self, cmapPreferences=((3, 10), (0, 4), (3, 1), (0, 3))):
-		"""Return the 'best' unicode cmap dictionary available in the font.
+		"""Return the 'best' unicode cmap dictionary available in the font,
+		or None, if no unicode cmap subtable is available.
 
 		By default it will search for the following (platformID, platEncID)
 		pairs:
@@ -50,7 +51,7 @@ class table__c_m_a_p(DefaultTable.DefaultTable):
 			cmapSubtable = self.getcmap(platformID, platEncID)
 			if cmapSubtable is not None:
 				return cmapSubtable.cmap
-		raise ValueError("None of the requested cmap subtables were found")
+		return None  # None of the requested cmap subtables were found
 
 	def buildReversed(self):
 		"""Returns a reverse cmap such as {'one':{0x31}, 'A':{0x41,0x391}}.
--- a/Lib/fontTools/ttLib/tables/_c_m_a_p.py
+++ b/Lib/fontTools/ttLib/tables/_c_m_a_p.py
@@ -38,12 +38,12 @@ class table__c_m_a_p(DefaultTable.DefaultTable):
 				return subtable
 		return None # not found
 
-	def getBestCmap(self, cmapPreferences=((3, 10), (3, 1), (0, 3))):
+	def getBestCmap(self, cmapPreferences=((3, 10), (0, 4), (3, 1), (0, 3))):
 		"""Return the 'best' unicode cmap dictionary available in the font.
 
 		By default it will search for the following (platformID, platEncID)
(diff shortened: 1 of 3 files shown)


justvanrossum:
Btw. it could also be a method on `TTFont`, I use it often in conjunction with `TTFont.getGlyphSet()`.

Lib/fontTools/ttLib/tables/_c_m_a_p.py
```diff
@@ -38,6 +38,20 @@ def getcmap(self, platformID, platEncID):
 				return subtable
 		return None # not found
 
+	def getBestCmap(self, cmapPreferences=((3, 10), (3, 1), (0, 3))):
```
anthrotype: just for symmetry, maybe you could also add plaftormID=0 (Unicode), platEncID=4 (non-BMP, Unicode full repertoire), since you also added (3, 10)

Lib/fontTools/ttLib/tables/_c_m_a_p.py
```diff
@@ -38,6 +38,20 @@ def getcmap(self, platformID, platEncID):
...
+		By default it will search for the following (platformID, platEncID)
+		pairs:
+			(3, 10), (3, 1), (0, 3)
+		But this can be customized via the cmapPreferences argument.
+		"""
+		for platformID, platEncID in cmapPreferences:
+			cmapSubtable = self.getcmap(platformID, platEncID)
+			if cmapSubtable is not None:
+				return cmapSubtable.cmap
+		raise ValueError("None of the requested cmap subtables were found")
```
anthrotype: why raise instead of simply returning None? Perhaps is better to leave it to the caller to decide what to do in case the requested cmap is missing.

anthrotype:
> it could also be a method on TTFont

sounds good to me

justvanrossum:
> >  it could also be a method on TTFont
>
> sounds good to me

- if I go that route, shall I just commit to this pull request?
- can the test code stay in test__c_m_a_p?

Lib/fontTools/ttLib/tables/_c_m_a_p.py
justvanrossum: Hm, I think I prefer an exception, because that gives the better error message if the caller _doesn't_ check the result. And there's always try/except :)

Lib/fontTools/ttLib/tables/_c_m_a_p.py
justvanrossum: In which order should it be in the default list?

anthrotype:
you could have a same named `getBestCmap` method in both the cmap class and TTFont class, and the TTFont's method just calls the former.

Lib/fontTools/ttLib/tables/_c_m_a_p.py
anthrotype: I don't want to insist too much on this, if you prefer the exception. But I think that a ValueError may be too heavy handed. This method is more about "finding" a best cmap than "getting", because you don't know in advance whether there will be one. That's why I thought a None result is more meaningful; it's obvious that it means "I coudn't find any best cmap" and your exception message does not add any more information than "None".
And if one then doesn't check for None and attempts to call cmap methods on NoneType, well Python will surely raise a TypeError, which speaks by itself. And the docstring can make it clear that the result may be None. Anyway, up to you.

Lib/fontTools/ttLib/tables/_c_m_a_p.py
anthrotype: i'd say `(3, 10), (0, 4), (3, 1), (0, 3)`

anthrotype:
if you're going to keep the exception, then I'd rather have TTLibError than a catch-all ValueError (but I still prefer None)

anthrotype:
... although TTLibError may also be misleading because it would seem to imply that a font without Unicode cmaps is somehow "faulty".

Lib/fontTools/ttLib/tables/_c_m_a_p.py
justvanrossum: Hm, I see your point, let me think. (So far, it seems very exceptional if no such cmap is available, hence the exception. But yeah, if a fallback is needed, None is easier to detect.)

Lib/fontTools/ttLib/tables/_c_m_a_p.py
```diff
@@ -38,6 +38,21 @@ def getcmap(self, platformID, platEncID):
...
+		By default it will search for the following (platformID, platEncID)
+		pairs:
+			(3, 10), (0, 4), (3, 1), (0, 3)
+		But this can be customized via the cmapPreferences argument.
+		"""
+		for platformID, platEncID in cmapPreferences:
+			cmapSubtable = self.getcmap(platformID, platEncID)
+			if cmapSubtable is not None:
+				return cmapSubtable.cmap
+		return None  # None of the requested cmap subtables were found
```
anthrotype: even without this line `return None` the function would have returned None. But explicit is better than implicit! :)

justvanrossum: I know, and yes, exactly! I always write `return None` if it has an explicit meaning other than "we're falling out of the method".

justvanrossum:
Thanks for the feedback, Cosimo!

behdad:
Can we make this match https://github.com/behdad/harfbuzz/blob/master/src/hb-ot-font.cc#L390

Also, would be good to make it return a base subtable and a Format14 subtable, if any.

PR #1093, opened 2017-11-03
Fix small graphite round trip bugs

mhosken:
This patch fixes a few minor but significant bugs in the Graphite table support to ensure clean round tripping via ttx.

### DIFF
--- a/Lib/fontTools/ttLib/tables/S__i_l_f.py
+++ b/Lib/fontTools/ttLib/tables/S__i_l_f.py
@@ -727,9 +727,9 @@ class Pass(object):
                 oConstraints[i] = oConstraints[i+1]
         self.ruleConstraints = [(data[s:e] if (e-s > 1) else "") for (s,e) in zip(oConstraints, oConstraints[1:])]
         data = data[oConstraints[-1]:]
-        for i in range(len(oActions)-2,-1,-1):
-            if oActions[i] == 0:
-                oActions[i] = oActions[i+1]
+#        for i in range(len(oActions)-2,-1,-1):
+#            if oActions[i] == 0:
+#                oActions[i] = oActions[i+1]
         self.actions = [(data[s:e] if (e-s > 1) else "") for (s,e) in zip(oActions, oActions[1:])]
         data = data[oActions[-1]:]
         # not using debug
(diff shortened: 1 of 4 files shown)


Lib/fontTools/ttLib/tables/S__i_l_f.py
```diff
@@ -727,18 +727,18 @@ def decompile(self, data, ttFont, version=2.0):
                 oConstraints[i] = oConstraints[i+1]
         self.ruleConstraints = [(data[s:e] if (e-s > 1) else "") for (s,e) in zip(oConstraints, oConstraints[1:])]
         data = data[oConstraints[-1]:]
-        for i in range(len(oActions)-2,-1,-1):
-            if oActions[i] == 0:
-                oActions[i] = oActions[i+1]
+#        for i in range(len(oActions)-2,-1,-1):
+#            if oActions[i] == 0:
+#                oActions[i] = oActions[i+1]
```
anthrotype: why you commented this out instead of simply deleting the lines?

anthrotype:
the test data also needs to be adjusted as you see from the bots failure.

anthrotype:
we should probably handle the case when TTX file was generated with a previous version of fonttools and does not have that new "default" attribute

anthrotype:
for py2.py3 compatiblity here you should use `byteord` instead of plain `ord`:
https://travis-ci.org/fonttools/fonttools/jobs/296902189#L2937

on python 3 indexing a bytes string yields an int, not a character.

for this other test failure, you need to convert the iterator returned by `map` function to a list if you want to get it's `len()`
https://travis-ci.org/fonttools/fonttools/jobs/296902189#L2992-L2993

anthrotype:
`byteord` is our compatibility wrapper defined in [`fontTools.misc.py23`](https://github.com/fonttools/fonttools/blob/b9bca702f6efb06e9c3c77f86bbb824d84b7b810/Lib/fontTools/misc/py23.py#L104). If you did `from fontTools.misc.py23 import *` at the top, then it is already in your namespace, otherwise you can import explicitly `from fontTools.misc.py23 import byteord`

mhosken:
Thanks for the tip on byteord. Would probably have reinvented the wheel, badly, wasted time and resulted in not good code.

Issue #314, opened 2015-07-02
ignore length of the last 'kern' subtable"


davelab6:
I noticed this warning with ttx'ing Open Sans Bold; the other styles in Open Sans don't have this issue:

``` sh
$ git clone https://github.com/google/fonts.git ; 
$ ttx fonts/apache/opensans/OpenSans-Bold.ttf ;
Dumping "OpenSans-Bold.ttf" to "OpenSans-Bold.ttx"...
Dumping 'GlyphOrder' table...
Dumping 'head' table...
Dumping 'hhea' table...
Dumping 'maxp' table...
Dumping 'OS/2' table...
Dumping 'hmtx' table...
Dumping 'cmap' table...
Dumping 'fpgm' table...
Dumping 'prep' table...
Dumping 'cvt ' table...
Dumping 'loca' table...
Dumping 'glyf' table...
/Library/Python/2.7/site-packages/fonttools-2.5-py2.7.egg/fontTools/ttLib/tables/_k_e_r_n.py:121: UserWarning: excess data in 'kern' subtable: 46628 bytes
Dumping 'kern' table...
Dumping 'name' table...
Dumping 'post' table...
Dumping 'gasp' table...
Dumping 'GDEF' table...
Dumping 'GPOS' table...
Dumping 'GSUB' table...
Dumping 'DSIG' table...
$ ttx OpenSans-Bold.ttx ;
$ ls -l OpenSans-Bold.ttf OpenSans-Bold#1.ttf ;
-rw-r--r--  1 user  group   158,720 Jul  1 20:53 OpenSans-Bold#1.ttf
-rw-r--r--  1 user  group   224,592 Jan 15 16:28 OpenSans-Bold.ttf
```

Since the data doesn't round trip, I guess its a bug in Open Sans, but before reporting this to the developer, I thought I'd file an issue to better understand this error :)

behdad:
Either bug in font, or in our tools. :)  Please report to developer.

davelab6:
Ruh roh:
```
$ ot-sanitise OpenSans-Bold.ttf 
ERROR: kern: Too large subtable., table discarded
```

davelab6:
Is it conceivable that a canny fontTools script could heal this? :)

khaledhosny:
If the OTS error is to be trusted, then the suntable is larger than what it takes to store 65536 kerning pairs (the largest possible number), which probably indicates a faulty translation of `GPOS` pair positioning to old style `kern` table.

[...]

PR #1094, opened 2017-11-03
fix compilation of AAT kern tables

### LINKED ISSUE #1089: Apple kern tables are wrongly compiled
The Apple kern table (with uint32 version) has different subtable format than the OpenType one.  We ignore this when compiling kern subtables.
https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6kern.html
https://www.microsoft.com/typography/otspec/kern.htm

anthrotype:
- When compiling kern subtables for version=1.0 kern tables (AAT) the subtable header was written incorrectly: there is no `version` there, the length is a uint32 and there's an additional uint16 for `tupleIndex` (used for GX)
- Use the `coverage` low byte to select subtable "format", instead of the `version` field, only present in OT kern subtable header.
- `KernTable_format_0` now takes an `apple=False` argument, used to choose the different headers and whether to read/write `tupleIndex`.
- A new `tupleIndex` attribute is written out to TTX for Apple-style kern subtables. Old TTX files which lack that attribute will still be read (with a warning) and will default to 0 when recompiled or dumped with current fonttools.

Fixes #1089

### DIFF
--- a/Lib/fontTools/ttLib/tables/_k_e_r_n.py
+++ b/Lib/fontTools/ttLib/tables/_k_e_r_n.py
@@ -17,7 +17,7 @@ class table__k_e_r_n(DefaultTable.DefaultTable):
 
 	def getkern(self, format):
 		for subtable in self.kernTables:
-			if subtable.version == format:
+			if subtable.format == format:
 				return subtable
 		return None  # not found
 
@@ -33,21 +33,23 @@ class table__k_e_r_n(DefaultTable.DefaultTable):
 		else:
 			self.version = version
 			data = data[4:]
-		tablesIndex = []
 		self.kernTables = []
 		for i in range(nTables):
 			if self.version == 1.0:
 				# Apple
-				length, coverage, tupleIndex = struct.unpack(">lHH", data[:8])
-				version = coverage & 0xff
+				length, coverage = struct.unpack(">LH", data[:6])
 			else:
-				version, length = struct.unpack(">HH", data[:4])
-			length = int(length)
-			if version not in kern_classes:
-				subtable = KernTable_format_unkown(version)
+				# in OpenType spec the "version" field refers to the common
+				# subtable header; the actual subtable format is stored in
+				# the last 8 mask bits of "coverage" field.
+				# Since this "version" is always 0 (and is not present in the
+				# later AAT extensions), we simply ignore it here
+				_, length, coverage = struct.unpack(">HHH", data[:6])
+			subtableFormat = coverage & 0xff
+			if subtableFormat not in kern_classes:
+				subtable = KernTable_format_unkown(subtableFormat)
 			else:
-				subtable = kern_classes[version]()
-			subtable.apple = apple
+				subtable = kern_classes[subtableFormat](apple)
 			subtable.decompile(data[:length], ttFont)
 			self.kernTables.append(subtable)
 			data = data[length:]
@@ -59,7 +61,7 @@ class table__k_e_r_n(DefaultTable.DefaultTable):
 			nTables = 0
 		if self.version == 1.0:
 			# AAT Apple's "new" format.
-			data = struct.pack(">ll", fl2fi(self.version, 16), nTables)
+			data = struct.pack(">LL", fl2fi(self.version, 16), nTables)
 		else:
 			data = struct.pack(">HH", self.version, nTables)
 		if hasattr(self, "kernTables"):
@@ -85,26 +87,41 @@ class table__k_e_r_n(DefaultTable.DefaultTable):
 		if format not in kern_classes:
 			subtable = KernTable_format_unkown(format)
 		else:
-			subtable = kern_classes[format]()
+			apple = self.version == 1.0
+			subtable = kern_classes[format](apple)
 		self.kernTables.append(subtable)
 		subtable.fromXML(name, attrs, content, ttFont)
 
 
 class KernTable_format_0(object):
 
+	# 'version' is kept for backward compatibility
+	version = format = 0
+
+	def __init__(self, apple=False):
+		self.apple = apple
+
 	def decompile(self, data, ttFont):
-		version, length, coverage = (0,0,0)
 		if not self.apple:
 			version, length, coverage = struct.unpack(">HHH", data[:6])
+			if version != 0:
+				from fontTools.ttLib import TTLibError
+				raise TTLibError(
+					"unsupported kern subtable version: %d" % version)
+			tupleIndex = None
+			# Should we also assert length == len(data)?
 			data = data[6:]
 		else:
-			version, length, coverage = struct.unpack(">LHH", data[:8])
+			length, coverage, tupleIndex = struct.unpack(">LHH", data[:8])
 			data = data[8:]
-		self.version, self.coverage = int(version), int(coverage)
+		assert self.format == coverage & 0xFF, "unsupported format"
+		self.coverage = coverage
+		self.tupleIndex = tupleIndex
 
 		self.kernTable = kernTable = {}
 
-		nPairs, searchRange, entrySelector, rangeShift = struct.unpack(">HHHH", data[:8])
+		nPairs, searchRange, entrySelector, rangeShift = struct.unpack(
+			">HHHH", data[:8])
 		data = data[8:]
 
 		nPairs = min(nPairs, len(data) // 6)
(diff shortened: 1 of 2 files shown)


Lib/fontTools/ttLib/tables/_k_e_r_n.py
```diff
@@ -33,21 +33,23 @@ def decompile(self, data, ttFont):
 		else:
 			self.version = version
 			data = data[4:]
-		tablesIndex = []
```
anthrotype: this `tableIndex` was unused so I removed it

Lib/fontTools/ttLib/tables/_k_e_r_n.py
```diff
@@ -59,7 +61,7 @@ def compile(self, ttFont):
 			nTables = 0
 		if self.version == 1.0:
 			# AAT Apple's "new" format.
-			data = struct.pack(">ll", fl2fi(self.version, 16), nTables)
+			data = struct.pack(">LL", fl2fi(self.version, 16), nTables)
```
anthrotype: the Apple spec has unsigned long integers (uint32) for version and nTables, not signed

Lib/fontTools/ttLib/tables/_k_e_r_n.py
```diff
@@ -85,26 +87,41 @@ def fromXML(self, name, attrs, content, ttFont):
...
+			if version != 0:
+				from fontTools.ttLib import TTLibError
+				raise TTLibError(
+					"unsupported kern subtable version: %d" % version)
+			tupleIndex = None
+			# Should we also assert length == len(data)?
 			data = data[6:]
 		else:
-			version, length, coverage = struct.unpack(">LHH", data[:8])
+			length, coverage, tupleIndex = struct.unpack(">LHH", data[:8])
```
anthrotype: we were also decompiling `coverage` incorrectly here for apple kern subtables: the `tupleIndex` was being read in as "coverage" attribute, whereas the actual `coverage` (in second position, where `length` used to be) was being thrown away.

Lib/fontTools/ttLib/tables/_k_e_r_n.py
```diff
@@ -115,50 +132,95 @@ def decompile(self, data, ttFont):
...
-				# Slower, but will not throw an IndexError on an invalid glyph id.
-				kernTable[(ttFont.getGlyphName(left), ttFont.getGlyphName(right))] = value
-		if len(data) > 6 * nPairs + 4: # Ignore up to 4 bytes excess
-			log.warning("excess data in 'kern' subtable: %d bytes", len(data) - 6 * nPairs)
+				# Slower, but will not throw an IndexError on an invalid
+				# glyph id.
+				kernTable[(
+					ttFont.getGlyphName(left),
+					ttFont.getGlyphName(right))] = value
+		if len(data) > 6 * nPairs + 4:  # Ignore up to 4 bytes excess
```
anthrotype: perhaps part of the responsibility for not noticing this bug before is that we are silently letting pass up to 4 extra bytes in the kern subtable. Why don't we always warn when the data length doesn't match the stated number of pairs or is not multiple of 6?

Lib/fontTools/ttLib/tables/_k_e_r_n.py
```diff
@@ -115,50 +132,95 @@ def decompile(self, data, ttFont):
...
-			kernTable = sorted((getGlyphID(left), getGlyphID(right), value) for ((left,right),value) in self.kernTable.items())
+			kernTable = sorted(
+				(getGlyphID(left), getGlyphID(right), value)
+				for ((left, right), value) in self.kernTable.items())
 
 		for left, right, value in kernTable:
 			data = data + struct.pack(">HHh", left, right, value)
-		return struct.pack(">HHH", self.version, len(data) + 6, self.coverage) + data
+		# ensure mask bits 8-15 (subtable format) are set to 0
+		self.coverage &= ~0xFF
```
anthrotype: here I follow the pattern found elsewhere in ttLib whereby any necessary fix-ups or modifications occur at compile time.

Lib/fontTools/ttLib/tables/_k_e_r_n.py
```diff
@@ -115,50 +132,95 @@ def decompile(self, data, ttFont):
...
-		self.version = safeEval(attrs["format"])
+		subtableFormat = safeEval(attrs["format"])
+		if self.apple:
+			if "tupleIndex" in attrs:
+				self.tupleIndex = safeEval(attrs["tupleIndex"])
+			else:
+				# previous fontTools versions didn't export tupleIndex
+				log.warning(
+					"Apple kern subtable is missing 'tupleIndex' attribute")
+				self.tupleIndex = None
```
anthrotype: here I decided to set tupleIndex to None when it's missing from a kern TTX file; if I had set it to 0, it wouldn't be possible to distinguish from a kern TTX file with a missing tupleIndex attribute or one with tupleIndex=0.
The value still defaults to 0 when the imported font is compiled or dumped again to XML.

behdad:
Kinda relevant https://github.com/fonttools/fonttools/issues/314

anthrotype:
https://github.com/fonttools/fonttools/issues/314#issuecomment-244432181
> we can detect [kern subtable exceeding maximum length] and possibly split into multiple subtables.

I could do that later in another PR, as this is about fixing AAT kern decompilation/compilation.

Lib/fontTools/ttLib/tables/_k_e_r_n.py
anthrotype: so, Calibri has 4 bytes in excess.. https://github.com/fonttools/fonttools/commit/5cd0a55635062f382ff86ccea6e42bbf9b8b030a
ok then  :man_shrugging:

justvanrossum: Well, I still see Calibri's kern table shrink a _lot_ after roundtripping with ttx. I mentioned it in https://github.com/fonttools/fonttools/commit/5cd0a55635062f382ff86ccea6e42bbf9b8b030a but I don't think anyone followed up on that. Check `ttx -l` before and after roundtripping. But then again, I'm not entirely sure where that Calibri came from.

justvanrossum: I can send you the version I have if you're curious.

anthrotype: Not particularly. I'd be more interested in having the AAT kern compilation fixed. Maybe you could also do a quick review since you wrote the original code? Thanks anyway :)

anthrotype: sure, you can send me that file, thanks

Lib/fontTools/ttLib/tables/_k_e_r_n.py
```diff
@@ -33,21 +33,23 @@ def decompile(self, data, ttFont):
...
-			length = int(length)
-			if version not in kern_classes:
-				subtable = KernTable_format_unkown(version)
+				# in OpenType spec the "version" field refers to the common
+				# subtable header; the actual subtable format is stored in
+				# the last 8 mask bits of "coverage" field.
+				# Since this "version" is always 0 (and is not present in the
+				# later AAT extensions), we simply ignore it here
+				_, length, coverage = struct.unpack(">HHH", data[:6])
+			subtableFormat = coverage & 0xff
```
anthrotype: ooh.... actually also the order of format and coverage bytes is different between OT and AAT kern! In OT the first byte is format and the following one is coverage, whereas in AAT it's the reverse. I got this completely wrong, sorry.. :grimacing: 
Thank @justvanrossum that had me check that CALIBRI.TTF kern. I was surprised when I say a microsoft font with a kern subtable format _1_ (Apple-only). It turns out I was reading wrong, the 1 was the coverage bit (horizontal data), not the format.
Why did they make it so complicated :confounded:

anthrotype:
About that CALIBRI.TTF font mentioned by @justvanrossum, it also seems to use the trick of having a huge single kern subtable with the USHORT length truncated (like @twardoch explained in  https://github.com/fonttools/fonttools/issues/314#issuecomment-118116527).
I wonder if that kerning works at all.
But again, this PR is about fixing compilation of valid AAT kern tables, rather than rescuing broken OT kern tables.

behdad:
> we can detect [kern subtable exceeding maximum length] and possibly split into multiple subtables.
> 
> I could do that later in another PR, as this is about fixing AAT kern decompilation/compilation.

Note that splitting doesn't help as Windows will ignore them all.  Looks like we should ignore num-items and use table length instead.  But then that's not what this PR is about.  Just wanted to make sure you don't go down that route.

Lib/fontTools/ttLib/tables/_k_e_r_n.py
behdad: Yeah it's as if someone took the existing format and said let's change everything about it!  Allocate higher byte for coverage, allocate bits MSB-to-LSB, and allocate a bit for horizontal, instead of vertical :)).

Lib/fontTools/ttLib/tables/_k_e_r_n.py
```diff
@@ -17,7 +19,7 @@ class table__k_e_r_n(DefaultTable.DefaultTable):
 
 	def getkern(self, format):
```
behdad: The name of this function is so wrong!  Shall we remove it or what?

Lib/fontTools/ttLib/tables/_k_e_r_n.py
```diff
@@ -85,80 +89,142 @@ def fromXML(self, name, attrs, content, ttFont):
...
 		self.kernTables.append(subtable)
 		subtable.fromXML(name, attrs, content, ttFont)
 
 
 class KernTable_format_0(object):
 
+	# 'version' is kept for backward compatibility
+	version = format = 0
+
+	def __init__(self, apple=False):
```
behdad: The apple is really an input to decompile() and compile(), not a property of the table itself.  What do you think about removing it and moving it there? (like isCFF2 if you remember that discussion.)

Also, the way I implemented this in HarfBuzz, I separated the Apple/MS subtable wrapper abstraction from the subtable types themselves.  I think we can do something similar here to simplify the code if you feel like.

Or you merge this, and I do that when implementing format-2.

Here's the HB code BTW:
https://github.com/behdad/harfbuzz/blob/master/src/hb-ot-kern-table.hh

Lib/fontTools/ttLib/tables/_k_e_r_n.py
```diff
@@ -85,80 +89,142 @@ def fromXML(self, name, attrs, content, ttFont):
...
 	def decompile(self, data, ttFont):
-		version, length, coverage = (0,0,0)
 		if not self.apple:
-			version, length, coverage = struct.unpack(">HHH", data[:6])
+			version, length, subtableFormat, coverage = struct.unpack(
+				">HHBB", data[:6])
+			if version != 0:
+				from fontTools.ttLib import TTLibError
+				raise TTLibError(
+					"unsupported kern subtable version: %d" % version)
```
behdad: Someone is going to get really confused by this error some time. :)

Lib/fontTools/ttLib/tables/_k_e_r_n.py
```diff
@@ -85,80 +89,142 @@ def fromXML(self, name, attrs, content, ttFont):
...
+			tupleIndex = None
+			# Should we also assert length == len(data)?
```
behdad: That should always pass, right? If that's the case, sure.  If it would fail on bogus fonts, no.

Lib/fontTools/ttLib/tables/_k_e_r_n.py
behdad: I see Win10 calibri.ttf has 160kb kern table.  I suppose MS started ignoring the binary search header completely and use the subtable length instead.  We should do the same.  I'll fix HarfBuzz as well.

anthrotype:
The description of the kern subtable format 0 on Apple website says that:
https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6kern.html

> The table needs to end with an entry for a left-hand glyph of 0xFFFF, a right-hand glyph of 0xFFFF, and a kerning value of 0.

However, Zapfino.ttf (12.0d1e6; 2016-07-18), which is the only font I was able to find on my macOS Sierra which has a kern version=1.0 and a format 0 subtable does _not_ end with `FFFF FFFF 0000`, but with a normal kern pair `04BA 04BA 0019` (which is `<pair l="fouroldstyle" r="fouroldstyle" v="25"/>`).

Shall we enforce that requirement somehow? Not raise of course, but maybe log a warning?

Lib/fontTools/ttLib/tables/_k_e_r_n.py
behdad: > I see Win10 calibri.ttf has 160kb kern table. I suppose MS started ignoring the binary search header completely and use the subtable length instead. We should do the same. I'll fix HarfBuzz as well.

I keep confusing myself.  Even subtable length is being ignored.  The whole table length is used as subtable length.  Argh!

Lib/fontTools/ttLib/tables/_k_e_r_n.py
anthrotype: I doubt anyone ever used it because it was throwing AttributeError because of missing "version" attribute on anything other than format-0 subtables (the only one implemented). Sure, I can remove it. We can add a better one when we implement other subtable formats.

Lib/fontTools/ttLib/tables/_k_e_r_n.py
anthrotype: I thought about putting "apple" as an argument for decompile/compile, but then I needed that attribute inside fromXML and toXML as well to decide whether to read/write the tupleIndex...
I think I shall leave it like that until we add more formats.

Lib/fontTools/ttLib/tables/_k_e_r_n.py
anthrotype: I doubt anyone will ever see that error. All the version=0 kern tables out there almost certainly have a subtable version=0, whatever that means. We'll see..

Lib/fontTools/ttLib/tables/_k_e_r_n.py
anthrotype: It should, because the wrapper class used the same length to trim the data and pass it down. The assert would simply check that this assumption holds.

behdad:
Let's merge and move on.  We'll fix more later.  I've asked for clarification re calibri.ttf.

Lib/fontTools/ttLib/tables/_k_e_r_n.py
behdad: My bad.  Ignoring subtable length, respecting bsearch header!  This is now implemented in HarfBuzz:
https://github.com/behdad/harfbuzz/commit/82a38d1f7a65537a4ef540af08c489512d6297ac

PR #1096, opened 2017-11-07
Split context

mhosken:
Given a split contextual lookup like:
```
sub a'lookup upper x x A'lookup lower;
```
Fonttools silently does the wrong thing in that it brings the two marked glyphs together resulting in a reordered rule that looks like:
```
sub a'lookup upper A' lookup lower x x;
```
Fontforge silently resolves this by inserting ' in all the intervening glyphs as in:
```
sub a'lookup upper x' x' A'lookup lower;
```
This PR does the same thing. The alternative is that an error should be reported and the user forced to insert the extra marks themselves. I leave that option as an open bug.

### DIFF
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -328,6 +328,9 @@ class Parser(object):
                 self.expect_symbol_("'")
                 hasMarks = marked = True
             if marked:
+                glyphs.extend(suffix)
+                lookups.extend([None] * len(suffix))
+                suffix = []
                 glyphs.append(gc)
             elif glyphs:
                 suffix.append(gc)
--- a/Lib/fontTools/ttLib/tables/F__e_a_t.py
+++ b/Lib/fontTools/ttLib/tables/F__e_a_t.py
@@ -99,7 +99,7 @@ class table_F__e_a_t(DefaultTable.DefaultTable):
             fobj = Feature()
             fobj.flags = int(safeEval(attrs['flags']))
             fobj.label = int(safeEval(attrs['label']))
-            fobj.default = int(safeEval(attrs['default']))
+            fobj.default = int(safeEval(attrs.get('default','0')))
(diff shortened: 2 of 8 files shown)


mhosken:
Blast. This PR can only be applied after #1093 :( Sorry.

anthrotype:
We also need to check what makeotf does with this.
And do you mind also adding a small test case?

mhosken:
Looking  at makeot it reports an error. But here is a comment in the code that implies they might have liked to support multiple runs: "At most one run of marked positions supported, for now". So I think it's open for us.

I realise it's hard to see, but there is already a test case in this PR (click on the last commit to see what I'm really proposing :))

behdad:
I agree we should support this.

brawer:
@readroberts, for the construct `sub a' lookup upper x x A' lookup lower;`, do you prefer the fontforge behavior (silently marking the `x` glyphs) or the makeotf behavior (raising an error)? Current fonttools reorders the `x` towards the end, which is clearly wrong. Personally I’d have a slight preference for raising an error, because compilers aren’t very good at second-guessing user intents. In any case, it would be desirable if all implementations interpreted the spec in the same way. So, curious about your thoughts here.

readroberts:
I also prefer intent to be expressed clearly in the source document, and will leave this as a fatal error in makeotf,  However,  I think it reasonable that an implementation fix the problem on the fly in order to allow the font to compile, if this is accompanied by a warning to the user that the syntax is in fact incorrect, that the fix may not reflect the user's intent. A bonus would be showing the fixed version, which could also help the user understand what the problem was.

anthrotype:
So, both Sascha and Read expressed their preference for raising an error in this case rather than guessing the user's intent.
If I need to choose, I prefer to match makeotf's behavior rather than fontforge's, as the former can be regarded as the reference implementation.

anthrotype:
see #1108

PR #1097, opened 2017-11-07
[py23] Add backports for redirect_stdout/stderr context managers

anthrotype:
I was prompted by another conversation to resuscitate something I had in my stash.

This adds py2.py3-compatible `redirect_stdout` and `redirect_stderr` context managers that can come handy in many situations.

https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stdout


anthrotype:
(i think I'm gonna drop python3.4 support once 3.7 is out)

PR #1101, opened 2017-11-14
Move arc length calculations from pens.perimeterPen to misc.bezierTools

jenskutilek:
As discussed in https://github.com/robofab-developers/fontPens/pull/5.


anthrotype:
and 💯 for adding the doctests!

jenskutilek:
Thanks :) I could recycle the doctests from the fontPens functions ;)

PR #1106, opened 2017-11-17
add fontTools.unicodedata, plus a script to fetch "Scripts.txt" from UCD  

anthrotype:
I need to lookup script names for given unicode characters, and the built-in unicodedata module doesn't provide that.

It's trivial to parse the UCD data files, so I just made a script `MetaTools/buildUCD.py` that fetches, for now only "Scripts.txt" (but we can add more later), parses it and writes a python module (`fontTools.unicodedata.scripts`) containing a list of script ranges and respective script name.

Then a `script` function in `fontTools.unicodedata` returns the script name given a unicode character (eg. works like `unicodedata.category`).

Since this kind of data changes once a year, I don't think it is worth the trouble to run the script that generates the data module automatically every time we build the distribution package.
We just do it manually once a new unicode data version is published.


anthrotype:
I forgot to also mention, this new `fontTools.unicodedata` is named like that because it is meant to be a sort of drop-in replacement for the built-in `unicodedata` module, so that one can do

```python
from fontTools import unicodedata

print(unicodedata.name("a"))  # using the built-in function

print(unicodedata.script("b"))  # using our own implementation
```

it re-exports all the public names from the built-in module, plus will add more like scripts, blocks, etc.

And it first attempts to import from the backport `unicodedata2` (which is usually more up-to-date), otherwise falls back to the built-in module.

anthrotype:
I've implemented the script name lookup using a binary search that narrows down the list of ranges until it finds one matching. It also caches the result so the same character is only searched once.

I could have alternatively created a set of codepoints for each script (there are 141 scripts) using these ranges, and then iterate over them and checking if a codepoint is in the set (which is what we do for the OS/2 unicode ranges
https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/ttLib/tables/O_S_2f_2.py#L481

I haven't tried the other solution nor profiled this, but I thought this would be more efficient.
I could change it if you think of a better solution.

anthrotype:
Finally, last question is, do I have to also include a LICENSE text file for the Unicode Character Database data?
http://unicode.org/copyright.html

Since we are technically not storing the original data files, but simply downloading them in memory and creating a derivative piece of text that contains the same data, I don't know.

I already included the URL and the header of the original data file, which contains its own copyright notice.
Is the LICENCE test also required, do you think?
If so, where shoud I put it?

justvanrossum:
Would it be hard to make the data from Blocks.txt available, too, here? I don't think that's in the builtin unicodedata module, either, but it's very useful.

behdad:
There's some stuff in nototools that you wanted to move here as well?

Lib/fontTools/unicodedata/__init__.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@_memoize
+def script(char):
+    """For the unicode character 'char' return the script name."""
+    code = byteord(char)
+    return _binary_search_range(code, SCRIPT_RANGES, default="Unknown")
+
+
+def _binary_search_range(code, ranges, default=None):
```
justvanrossum: Any reason not to use the `bisect` module?

anthrotype: Honestly I didn’t know about bisect module, i’ll try to use it and see if’s faster, thanks!

anthrotype:
Yeah, after discussing with Sascha I figured that it would be easier to simply rewrite the stuff from nototools unicode_data than going through the process of splitting the project, create new repo under googlei18n, etc.

anthrotype:
Scripts are more useful than blocks, but sure I can add those too

Lib/fontTools/unicodedata/__init__.py
justvanrossum: I don't think it will be faster, or much faster, but it's a well tested implementation and it will make your code shorter, so a win nevertheless :)

justvanrossum:
> Scripts are more useful than blocks, but sure I can add those too
That would be great. (It's just a different table, that has it's own uses.)
The Blocks.txt file uses the same format, so it should be pretty straight-forward, I hope. Let me know if can help in any way.

Tests/unicodedata/scripts_test.py
```diff
@@ -0,0 +1,179 @@
...
+    assert search(23, ranges) is None
+
+
+def test_script():
+    assert unicodedata.script("a") == "Latin"
+
+    assert hasattr(unicodedata.script, "cache")
+    assert unicodedata.script("a") == unicodedata.script.cache["a"]
+
+    # these were randomly sampled, one character per script
```
justvanrossum: I would add some Unknowns, too, right before and after an existing script range.

brawer:
About the license, perhaps change the metatools script so it emits a comment at the beginning of the generated table, with “# License of Unicode data: http://unicode.org/copyright.html#License” or similar.

About bsearch, here’s snippets: https://docs.python.org/2/library/bisect.html#searching-sorted-lists

Tests/unicodedata/scripts_test.py
anthrotype: Indeed, I thought the ranges were continuous. Aren’t they? I’ll check

justvanrossum: Hm, it seems that they are, sorry!

justvanrossum: Actually, I checked wrong. Look here:

```
0376..0377    ; Greek # L&   [2] GREEK CAPITAL LETTER PAMPHYLIAN DIGAMMA..GREEK SMALL LETTER PAMPHYLIAN DIGAMMA
037A          ; Greek # Lm       GREEK YPOGEGRAMMENI
```
There's nothing between `0377` and `037A`.

anthrotype: Ah! I’m away from the laptop so I can’t check. I wonder what my script function returns in between those. It should return “Unknown”. If it doesn’t it’s a bug

anthrotype: But the ranges in the text file are not sorted by codepoint, so it might be that those codepoints are covered by some other script.
In theory all codepoints should be assigned a script...

justvanrossum: Check the generated scripts.py, same gap.

punchcutter: There should be tons of gaps between codepoints within a script range because the block is reserved for that script, but not all codepoints are used yet, if ever. I tried some I know like Chakma 0x11145 which is proposed and almost accepted, but not officially added yet. So right now it returns Unknown as expected.

anthrotype: So it works. Thanks for checking.

anthrotype:
looks like the built-in bisect module has a fast C implementation
https://github.com/python/cpython/blob/3.6/Modules/_bisectmodule.c
i'll switch to use that tomorrow

chrissimpkins:
I don't know if you intend to use this for one off script searches (in which case the following is completely irrelevant) or if this going to be implemented in a large loop (in which case it might be relevant), but out of interest since you are discussing optimizations I mapped your Unicode scripts list of tuples to a dictionary mapped as `key = character integer : value = Unicode script string` (2.1 MB module):

```python
i = 0
uniscript_dict = {}
for line in SCRIPT_RANGES:
    for x in range(SCRIPT_RANGES[i][0], SCRIPT_RANGES[i][1]):
        uniscript_dict[x] = SCRIPT_RANGES[i][2]
    i += 1
print(uniscript_dict)
```

wrote this dictionary and the SCRIPT_RANGES list to separate modules, then tested tested the binary search against dictionary lookups and there is an order of magnitude difference in time when tested with the Py3.6.3 interpreter:

```
Starting timed trials of binary_search_list()
Trial 1:	0.2818267345428467
Trial 2:	0.27074575424194336
Trial 3:	0.2770988941192627
Trial 4:	0.26982998847961426
Trial 5:	0.27674317359924316
Trial 6:	0.2746002674102783
Trial 7:	0.2848508358001709
Trial 8:	0.2847299575805664
Trial 9:	0.2798137664794922
Trial 10:	0.27668190002441406
--------------------------------------------------
Mean for 100000 repetitions: 0.2776921272277832 sec
Mean per repetition: 2.776921272277832e-06 sec

Starting timed trials of search_dict()
Trial 1:	0.04316210746765137
Trial 2:	0.048757314682006836
Trial 3:	0.04414200782775879
Trial 4:	0.04790687561035156
Trial 5:	0.043189048767089844
Trial 6:	0.043663978576660156
Trial 7:	0.04821014404296875
Trial 8:	0.044342756271362305
Trial 9:	0.04677605628967285
Trial 10:	0.045201778411865234
--------------------------------------------------
Mean for 100000 repetitions: 0.04553520679473877 sec
Mean per repetition: 4.553520679473877e-07 sec
```

Test file:

```python
# -*- coding: utf-8 -*-

from __future__ import (
    print_function, division, absolute_import, unicode_literals)

import os
import sys
import time
import gc
from functools import wraps


from scripts_list import SCRIPT_RANGES
from scripts_dict import scriptsdict


try:
    # use unicodedata backport compatible with python2:
    # https://github.com/mikekap/unicodedata2
    from unicodedata2 import *
except ImportError:
    # fall back to built-in unicodedata (possibly outdated)
    from unicodedata import *

def make_dict():
    i = 0
    uniscript_dict = {}
    for line in SCRIPT_RANGES:
        for x in range(SCRIPT_RANGES[i][0], SCRIPT_RANGES[i][1]):
            uniscript_dict[hex(x)] = SCRIPT_RANGES[i][2]
        i += 1
    print(uniscript_dict)


def timer_trials_benchmark(func, repetitions=100000, trials=10):
    @wraps(func)
    def wrapper(*args, **kwargs):
        sys.stdout.write("Starting timed trials of " + func.__name__ + "()")
        sys.stdout.flush()
        result_list = []
        benchmark_list = []
        # disable garbage collection
        gc.collect()
        gc.disable()
        for x in range(trials):
            # test function
            start = time.time()
            for y in range(repetitions):
                func(*args, **kwargs)
            end = time.time()
            result = func(*args, **kwargs)
            result_list.append(end-start)
        gc.enable()  # re-enable garbage collection
        gc.collect()
        print(" ")
        n = 1
        for run in result_list:
            print("Trial " + str(n) + ":\t" + str(run))
            n += 1
        print("-"*50)
        mean = sum(result_list)/len(result_list)
        print("Mean for " + str(repetitions) + " repetitions: " + str(mean) + " sec")
        try:
            import numpy as np
            array = np.array(result_list)
            print( "Standard Deviation: " + str(np.std(array)))
        except ImportError as ie:
            pass
        print("Mean per repetition: " + str(mean/repetitions) + " sec" + os.linesep)
        return result
    return wrapper



@timer_trials_benchmark
def binary_search_list():
    code = byteord('x')
    return _binary_search_range(code, SCRIPT_RANGES, default="Unknown")


def _binary_search_range(code, ranges, default=None):
    left = 0
    right = len(ranges) - 1
    while right >= left:
        mid = (left + right) >> 1
        if code < ranges[mid][0]:
            right = mid - 1
        elif code > ranges[mid][1]:
            left = mid + 1
        else:
            return ranges[mid][2]
    return default

@timer_trials_benchmark
def search_dict():
    code = byteord('x')
    try:
        return scriptsdict[code]
    except Exception:
        return "Unknown"


if __name__ == '__main__':
    binary_search_list()
    search_dict()


```

The large data file import will clearly make a significant difference if you are not dealing with a large number of script searches.  FWIW in case these data are helpful.  If not, I had the opportunity to learn something and apologize for the thread hijack :)

[unilist.tar.gz](https://github.com/fonttools/fonttools/files/1484236/unilist.tar.gz)
[unidict.tar.gz](https://github.com/fonttools/fonttools/files/1484240/unidict.tar.gz)

PeterCon:
Can you clarify this:

>Since we are technically not storing the original data files, but simply downloading them in memory and creating a derivative piece of text that contains the same data, I don't know.

When/how are they getting downloaded?

Unicode does not exactly appreciate it when software pings Unicode servers to retrieve UCD data files. It's definitely preferred for software developers to download once and incorporate data into their software.

chrissimpkins:
@PeterCon Answered in original post by Cosimo:

> Since this kind of data changes once a year, I don't think it is worth the trouble to run the script that generates the data module automatically every time we build the distribution package.
We just do it manually once a new unicode data version is published.

Cosimo is currently storing the Unicode code point ranges mapped to the scripts in this proposed module in the PR: https://github.com/anthrotype/fonttools/blob/2eb59f163fb6bd0998415e94467575647e100e67/Lib/fontTools/unicodedata/scripts.py

anthrotype:
@chrissimpkins I thought about that too, but didn’t want to add tens of megabytes modules (which are also slow to import). Yes of course dict lookup is much faster.
I’ll try to see if I can use pickle to serialise the dict to disk and compare the performance with the bisect method. But thanks for your tests!

@PeterCon as Chris said, i’m using a script (to be run about once a year) that will generate data for python to consume from the online unidata. I said “in memory” because the script does not even save the data file to disk in order to parse it. Sorry for the misunderstanding.

chrissimpkins:
> I thought about that too, but didn’t want to add tens of megabytes modules (which are also slow to import)

Absolutely.  There is a great deal of size reduction that can happen in the dict but still unnecessary if there is not a performance issue at hand that warrants this size/speed balance.  I was interested to see what the lookup difference actually was.  Figured I would share since I had the data available.

anthrotype:
happy to report that using the built-in bisect module is 4-5 times faster than my pure python binary search implementation, as also almost as fast as the dict lookup (pickled). On my machine both method take between 0.6 and 0.7 _micro_seconds to look up a (random) character.

MetaTools/buildUCD.py
```diff
@@ -75,20 +86,65 @@ def parse_range_properties(infile):
...
+    for (cs, ce, cn), (ns, ne, nn) in zip(merged_ranges, merged_ranges[1:]):
+        assert ce+1 == ns
+    assert merged_ranges[-1][1] == MAX_UNICODE
+
+    return merged_ranges
+
 
+def build_scripts(local_ucd=None, output_path=None):
+    """Fetch "Scripts.txt" data file from Unicode official website, parse
+    the script ranges and write them as a list of Python tuples to
```
mashabow: Now it writes two lists instead of a list of tuple

anthrotype: I’ll fix the docstring, thanks for checking

anthrotype:
and to facilitate downstream packagers such as Debian, I also made `MetaTools/buildUCD.py` script accept a `--ucd-path` option to pass a local directory containing UCD data files. /cc @medicalwei

Issue #911, opened 2017-04-08
[varLib] Use designspaceDocument?


brawer:
Should `fontTools.varLib` use @LettError’s [designspaceDocument](https://github.com/LettError/designSpaceDocument) class?

* Advantages: Nice API. Less duplication of code between font libraries, such as glyphsLib (which writes designspace) and varLib (which reads them). Files would always be spec-compliant.
* Disadvantage: Our dependency hell gets worse, for a relatively minor piece.

Thoughts? Perhaps designspaceDocument coud become part of fonttools?

anthrotype:
+1 for using designspaceDocument, +2 for including it in fonttools if Erik steps up to maintain it

davelab6:
Are the dependencies really hellish? It seems to be smooth to me. Curious
what the advantages of inlining it are? 🤔

brawer:
Yes, they are.

LettError:
If this is the way to go, I don't mind maintaining it.

anthrotype:
Thanks Erik! You already have push access, so you can create a new branch here on the fonttools repo where we work together on integrating the new subpackage into the library.
I guess it would go inside fontTools.varLib, replacing the current designspace.py. Or do we want to place it one level above, under fontTools.designspaceDocument?

anthrotype:
Maybe outside of varLib, as fontTools.designspaceDocument, would be better.

behdad:
I like folding it in as well. Will review the code / api.

khaledhosny:
Already done #1313.

PR #1110, opened 2017-11-22
Merge designSpaceDocument

belluzj:
Hello,

This MR import the whole history of https://github.com/LettError/designSpaceDocument (master branch only) into fontTools, as per
* https://github.com/LettError/designSpaceDocument/issues/28
* https://github.com/fonttools/fonttools/issues/911

I have a few questions:
* Is the new module name OK? (`fonttools.designSpaceDocument`)
* Where should I move the documentation? It is in `README.md` and `scripting.md`. Maybe in `Doc/source/designSpaceDocument/` and translated to reStructuredText? I don't know Sphinx and I don't know how documentation is handled in this project or in Python in general, so I will gladly take any advice or links to relevant documentation/best practices!
* Should I move the doctests and test designspace files to the `Tests/designSpaceDocument` folder?
* Should I turn the doctests into `unittest.TestCase`s?
* Should I delete the UFO-specifics parts, as suggested in https://github.com/LettError/designSpaceDocument/issues/28#issue-274587260?
* What else is there to do in this MR?

Thanks in advance!

@LettError @anthrotype


anthrotype:
thanks Jany!

about the module name: I'm fine with either keeping the original name `fontTools.designSpaceDocument`, or change to something like `fontTools.designspaceLib` (for consistency with `ttLib`, `varLib`), or even a compressed `dsLib`. I also wouldn't mind  `fontTools.designspace`.

About the documentation: yes, ideally we would have it converted to reStructuredText and built with Sphinx. I'm also not very familiar with this stuff honestly. What we have in Doc/ folder right now is simply pulling the docstrings (most of which are not properly formatted). I know pandoc can covert between markdown and restructuredtext. Maybe for know that would be enough.

about the tests: since we have all our other tests in a separate folder, we should do the same for this new sub-package. And yes, ideally we should port doctests to unittest (I personally prefer pytest-style tests, which is also our test runner, but I'm ok with both). We can leave some doctests to document the API and ensure examples actually run.

and finally, we should delete the ufoProcessor.py module. Let's make this ufo-agnostic, at least until we have a working fontTools.ufoLib.

adrientetar:
designSpaceLib sgtm.

justvanrossum:
Since we tend to write "designspace" and not "design space", I would prefer `designspaceLib`.

belluzj:
Thanks everyone for the feedback! I have a new question now that I've started working on this:

`designspaceLib` actually imports `mutatorMath`. What should I do with this dependency? I could add it to requirements.txt/setup.py, but I have read about a "dependency hell", so that's why I'm asking :)

The dependency is only required by the `check()` and `checkDefault()` methods, which are not called automatically when just using the `DesignSpaceDocument` class, so maybe I could move the import into those methods and do NOT declare `mutatorMath` as a mandatory dependency of fonttools. However the existing tests that I'm going to migrate are calling those functions and thus need mutatorMath.

What should I do?

LettError:
`checkDefault()` is used when no source is designated as the neutral and we need to find one to be able to move on. Maybe this could move to `ufoProcessor` which already has a mutatorMath dependency. And that was going to be split off anyway;

anthrotype:
> designspaceLib actually imports mutatorMath

It sholdn't need to do that, aside from the ufoProcessor.py which we are getting rid of.
Let's reimplement those function here and not require mutatorMath.

MutatorMath shuld require fontTools, not the other way around.

belluzj:
OK thanks!

LettError:
> MutatorMath shuld require fontTools, not the other way around.
Not sure I agree with that.

anthrotype:
@LettError what I meant by that is that fontTools is a lower-level library, and mutatorMath something that may (or may not) build on top of fontTools. I didn't mean it has to :)

LettError:
MutatorMath only depends on standard modules. FontTools on the other hand.. 
Anyway, I understand your question and I'll prepare a separate ufoProcessor thing. Note if `checkDefault` moves out of `DesignSpaceDocument` it might be necessary to throw a meaningful exception if there is no origin defined in the sources. Or maybe declare the origin explicitly so we don't have to rely on the `copyInfo` flag.

behdad:
I'll try to review this soon.

belluzj:
Same question for dependency on `defcon`: it's only used by the function `getFonts()`, so I can either remove that function and the `fontClass` parameters/attributes, or keep everything but delay the import of `defcon` to when someone actually calls `getFonts()`.

anthrotype:
how important is to keep that `getFonts` function in?
I would prefer to avoid adding any UFO related dependency

anthrotype:
@belluzj for the travis test failure you have to merge origin/master or rebase on it, because Travis CI now ships a new osx image with an updated homebrew, so when we do `brew install python3` we now get python3.6 (previously it was 3.5). The .travis.yml needs to be updated like so 
https://github.com/fonttools/fonttools/commit/b0bf3ce4f5cb82e22fa13ecf975d3bd258cd9afa

belluzj:
It does not look like a very important function, just a convenience helper.

The mutatorMath-dependent function actually looks much more complex and useful and the existing tests depend on it, so I'm considering leaving it here, and stripping away the mutatorMath part. I'll push some code soon.

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -959,11 +982,15 @@ def __init__(self, readerClass=None, writerClass=None, fontClass=None):
...
-            self.fontClass = Font
+        self._fontClass = fontClass
+
+    @property
+    def fontClass(self):
+        if self._fontClass is not None:
+            return self._fontClass
+
+        from defcon.objects.font import Font
+        return Font
```
belluzj: Here I import `defcon` only when someone calls `fontClass` or its only caller, `getFonts()`

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -1129,13 +1156,7 @@ def checkDefault(self):
...
                 # we choose you!
                 flaggedDefaultCandidate = sourceDescriptor
-        masterLocations = [src.location for src in self.sources]
-        mutatorBias = biasFromLocations(masterLocations, preferOrigin=False)
-        c = [src for src in self.sources if src.location==mutatorBias]
-        if c:
-            mutatorDefaultCandidate = c[0]
-        else:
-            mutatorDefaultCandidate = None
+        mutatorDefaultCandidate = self.getMutatorDefaultCandidate()
```
belluzj: Here I moved into a new method the part of `checkDefault()` that actually needs MutatorMath, with the idea that another python project that uses both `fonttools` and `mutatorMath` can override this method to re-enable the implementation.

Tests/designspaceLib/designspace_test.py
```diff
@@ -0,0 +1,870 @@
...
+
+
+@pytest.fixture
+def invalid_designspace():
+    p = "testCheck.designspace"
+    __removeAxesFromDesignSpace(p)
+    yield p
+
+
+@pytest.mark.xfail(reason="The check method requires MutatorMath")
```
belluzj: I disabled this test since it requires the mutatorMath functionality. Should I leave it or remove altogether?

Tests/designspaceLib/designspace_test.py
```diff
@@ -0,0 +1,870 @@
...
+    doc.addAxis(a2)
+    # add an axis that is not part of any location to see if that works
+    a3 = AxisDescriptor()
+    a3.minimum = 333
+    a3.maximum = 666
+    a3.default = 444
+    a3.name = "spooky"
+    a3.tag = "spok"
+    a3.map = [(0.0, 10.0), (401.0, 66.0), (1000.0, 990.0)]
+    #doc.addAxis(a3)    # uncomment this line to test the effects of default axes values
```
belluzj: Maybe here there is the opportunity to add a test with the uncommented line, that asserts about "the effects of default axes values"?

Tests/designspaceLib/designspace_test.py
```diff
@@ -0,0 +1,870 @@
...
+    r1.subs.append(("a", "a.alt"))
+    doc.addRule(r1)
+    # write the document
+    doc.write(testDocPath)
+    assert os.path.exists(testDocPath)
+    # import it again
+    new = DesignSpaceDocument()
+    new.read(testDocPath)
+    new.write(testDocPath2)
+    f1 = open(testDocPath, 'r', encoding='utf-8')
```
belluzj: I had to add `encoding='utf-8'` to make it work on Python 3. Is this the best practice?

Tests/designspaceLib/designspace_test.py
```diff
@@ -0,0 +1,870 @@
...
+    doc.check()
+
+    # FIXME: this test fails, but I don't know why (I don't understand the
+    # importance of the ordering)
+    # E       AssertionError: assert ['width', 'weight', 'spooky'] == ['spooky', 'weight', 'width']
+    # E         At index 0 diff: 'width' != u'spooky'
+    # E         Full diff:
+    # E         - ['width', 'weight', 'spooky']
+    # E         + [u'spooky', u'weight', u'width']
+    # assert doc.getAxisOrder() == ['spooky', 'weight', 'width']
```
belluzj: @LettError: what was your intention here?
1. the order matters and I broke something while moving the code (what is the "rule" for ordering?)
2. the order does not matter and I should add `set()` on both sides

Doc/source/designspaceLib/readme.rst
```diff
@@ -0,0 +1,953 @@
+DesignSpaceDocument
```
belluzj: This is `README.md` converted to reST using Pandoc. I'm sure that it could fit much better in Sphinx's model if it were re-organized by hand (maybe by moving most of it into docstrings?) but maybe it's for another PR?

belluzj:
Link to view only my changes to the original designSpaceDocument: https://github.com/fonttools/fonttools/pull/1110/files/44727fe32a475632edfa8b213419bbf178da99c0..4e8818278b9c2547315b013b993c0b9c07dca743

Tests/designspaceLib/designspace_test.py
LettError: Not sure what you're asking. But `doc.getAxisOrder()` should return the axis names in the order they were added to the document.

Doc/source/designspaceLib/readme.rst
anthrotype: > moving most of it into docstrings?) but maybe it's for another PR?

yes, let's do this another time.

Lib/fontTools/designspaceLib/__init__.py
anthrotype: and who is calling getFonts()? Can't clients re-implement their own getFonts()? I think DesignSpaceDocument can live without a self.fontClass property, at least until we have a working fontTools.ufoLib.Font implementation to use as default.

Tests/designspaceLib/designspace_test.py
anthrotype: > encoding='utf-8'... Is this the best practice?

totally yes

Tests/designspaceLib/designspace_test.py
```diff
@@ -0,0 +1,870 @@
+# coding=utf-8
+
+from __future__ import (print_function, division, absolute_import,
+                        unicode_literals)
+
+import os
+import posixpath
```
anthrotype: I don't see this `posixpath` module used anywhere

Tests/designspaceLib/designspace_test.py
anthrotype: leave it as xfail until we reimplment that piece of code from mutatormath in here

moyogo:
> I'll try to review this soon.

@behdad We can address your concerns after the merge. Merging this would help getting the ball rolling.

belluzj:
I will be happy to make further MRs to address any concerns that might come up upon closer inspection! I will make other MRs anyway to add the `lib` elements.

behdad:
I'm fine merging as long as changes can happen after (which I don't see why they can't).

moyogo:
Thanks Behdad!

PR #1113, opened 2017-11-24
Don't share points in cvar by default

jenskutilek:
* Don't share points in cvar by default, font may fail to render properly in Chrome/WIndows (introduced in #1090)
* Adjust tests to test both shared and private points in cvar

TODO:
* Make `useSharedPoints` accessible to varLib (now only directly calling `cvar.compile()` or `TupleVariation.compileTupleVariationStore()` can take this argument)

### DIFF
--- a/Lib/fontTools/ttLib/tables/TupleVariation.py
+++ b/Lib/fontTools/ttLib/tables/TupleVariation.py
@@ -465,7 +465,8 @@ def compileSharedTuples(axisTags, variations):
 
 
 def compileTupleVariationStore(variations, pointCount,
-                               axisTags, sharedTupleIndices):
+                               axisTags, sharedTupleIndices,
+                               useSharedPoints=True):
 	variations = [v for v in variations if v.hasImpact()]
 	if len(variations) == 0:
 		return (0, b"", b"")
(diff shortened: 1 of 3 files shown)


PeterCon:
Do you have fonts that you think have valid cvar tables using shared points but that do fail to render properly in some context? If so, it would be good to get bugs logged on the renderers.

anthrotype:
> font may fail to render properly in Chrome/WIndows

Got more details? Did you find out why?

jenskutilek:
I’ve reconstructed this issue with a font that I can share. Here ist what it looks like:

bildschirmfoto 2017-11-27 um 16 28 30

Both TTFs are compiled from the same ttx file. The one which looks reverse-contrast has cvar shared ‘points’ (actually they’re CVTs) on, the one which looks normal doesn't use shared points.

[VotoSerif_cvar.zip](https://github.com/fonttools/fonttools/files/1506408/VotoSerif_cvar.zip)

The font where I first noticed it was even more severe, it wouldn't display variations at all (only the default instance was displayed).

This is visible in FontView and in Chrome on Windows, which also uses FreeType for variable fonts afaik. I'll look for where to file a bug on FreeType.

jenskutilek:
This is the FreeType bug report: https://savannah.nongnu.org/bugs/index.php?52532

anthrotype:
Thanks Jens

jenskutilek:
Should I merge this now, and reactivate sharing CVT indices some time after FreeType has been fixed?

anthrotype:
sounds good to me

PR #1118, opened 2017-11-27
[subset] delete empty SubrsIndex after subsetting CFF

### LINKED ISSUE #994: Subsetter and CFF CharStrings
I've been banging my head against a Windows error with some fonts that have been made with the subsetter: turns out that it is removing the CharString entries that are marked raw="1". This is for the <Subrs> entry. I don't know why these are getting dumped as raw — that's the next thing to figure out, but removing them breaks the font in Windows.

anthrotype:
Fixes #994 

According to @benkiel, when a CFF PrivateDict contains an empty SubrsIndex (one with no local subroutines), Windows rejects the font.

If the SubrsIndex is left empty by the subsetting operation, we delete it from the PrivateDict.

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -2407,11 +2407,26 @@ def prune_post_subset(self, options):
             for subr in subrs.items:
                 subr.subset_subroutines (local_subrs, font.GlobalSubrs)
 
+        # Delete local SubrsIndex if empty
+        if hasattr(font, 'FDSelect'):
+            for fd in font.FDArray:
+                _delete_empty_subrs(fd)
+        else:
+            _delete_empty_subrs(font)
+
         # Cleanup
         for subrs in all_subrs:
             del subrs._used, subrs._old_bias, subrs._new_bias
 
     return True
+
+
+def _delete_empty_subrs(font):
+    if hasattr(font.Private, 'Subrs') and not font.Private.Subrs:
+        if 'Subrs' in font.Private.rawDict:
+            del font.Private.rawDict['Subrs']
+        delattr(font.Private, 'Subrs')
+
 
 @_add_method(ttLib.getTableClass('cmap'))
 def closure_glyphs(self, s):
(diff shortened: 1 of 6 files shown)


Lib/fontTools/subset/__init__.py
```diff
@@ -2407,12 +2407,27 @@ def prune_post_subset(self, options):
...
             del subrs._used, subrs._old_bias, subrs._new_bias
 
     return True
 
+
+def _delete_empty_subrs(font):
+    if hasattr(font.Private, 'Subrs') and not font.Private.Subrs:
+        if 'Subrs' in font.Private.rawDict:
+            del font.Private.rawDict['Subrs']
+        delattr(font.Private, 'Subrs')
```
justvanrossum: `del font.Private.Subrs` should do the same.

Lib/fontTools/subset/__init__.py
```diff
@@ -2407,12 +2407,27 @@ def prune_post_subset(self, options):
...
+            _delete_empty_subrs(font)
+
         # Cleanup
         for subrs in all_subrs:
             del subrs._used, subrs._old_bias, subrs._new_bias
 
     return True
 
+
+def _delete_empty_subrs(font):
```
behdad: Maybe better to make this take a Private, not "font" which is sometimes a font, sometimes a fontDict.

Lib/fontTools/subset/__init__.py
```diff
@@ -2407,12 +2407,27 @@ def prune_post_subset(self, options):
...
+    if hasattr(font.Private, 'Subrs') and not font.Private.Subrs:
+        if 'Subrs' in font.Private.rawDict:
+            del font.Private.rawDict['Subrs']
```
behdad: Why is this one needed?

anthrotype: otherwise when compiling, the raw stuff gets written out

anthrotype: I copied from a similar snippet in cffLib

anthrotype:
thanks @justvanrossum and @behdad for the feedback. I applied the requested changes

Lib/fontTools/subset/__init__.py
behdad: Yeah I see in the code.  If we were maintaining this properly we would have added a `__delattr__`. But not now.

PR #1123, opened 2017-12-08
Split the mkmk lookup into subtables

belluzj:
Hello,

@moyogo and I found that generating the `mkmk` feature as 1 lookup with several subtables (1 subtable/mark class) fixed problems with Thai mark-to-mark positioning in InDesign, when the same font would work fine in browsers. This PR goes in pair with https://github.com/googlei18n/ufo2ft/pull/189

This PR changes the `MarkToMarkPosBuilder` to output several subtables, one per class.

We haven't tested yet the impact on mark-to-mark positioning for other scripts, and we haven't really assessed the filesize implications of this.

We didn't use "subtable;" statements in our lookup in the feature file because according to the spec they are only allowed in kerning (https://github.com/googlei18n/ufo2ft/pull/189#issuecomment-350298207)

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -9,6 +9,7 @@ from fontTools.otlLib import builder as otl
 from fontTools.ttLib import newTable, getTableModule
 from fontTools.ttLib.tables import otBase, otTables
 import itertools
+from collections import defaultdict
 
 
 def addOpenTypeFeatures(font, featurefile):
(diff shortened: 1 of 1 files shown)


khaledhosny:
Do we still need this? The mentioned googlei18n/ufo2ft#189 PR above is closed.

moyogo:
I think this can be closed as feaLib should respect subtable statement instead. Once that is done we can see if ufo2ft’s markFeatureWriter needs to be fixed.

PR #1132, opened 2017-12-19
Update for UFO naming of glyph files and make -g ttx parameter force -s

bamidei:
This new PR implements requested changes for the naming of the glyph file names - we now use the UFO naming convention.  Additionally, the '-g' parameter which indicates glyphs should be split to individual files now also implies '-s' or splitting of font tables to individual files.


Lib/fontTools/ttLib/__init__.py
```diff
@@ -879,8 +879,8 @@ def _escapechar(c):
 		return hex(byteord(c))[2:]
 
 
-def nameToIdentifier(name):
```
bamidei: NOTE: these changes revert the tagToIdentifier to previous implementation b/c we no longer need the nameToIdentifier function exposed.  After this commit, there should be no changes to this function since prior to the split_glyphs branch.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -139,9 +142,8 @@ def toXML(self, writer, ttFont, progress=None, splitGlyphs=False):
 			glyph = self[glyphName]
 			if glyph.numberOfContours:
 				if splitGlyphs:
-					path, ext = os.path.splitext(writer.file.name)
-					fileNameTemplate = path + ".%s" + ext
-					glyphPath = fileNameTemplate % nameToIdentifier(glyphName)
+					glyphPath = userNameToFileName(unicode(glyphName, 'utf-8'), prefix=("%s%s"%(path, "_")), suffix=ext)
```
bamidei: It is nice to have an extra '_' prior to the glyph name so that the table .ttx appears in the directory listing as the first file (i.e. font_name._g_l_y_f.ttx) and not mixed with all of the other _g_l_y_f_ files...

anthrotype:
Thanks.

I was thinking maybe we could move that code to a separate module inside fontTools.misc?
We could call it `filenames.py` like the one from ufoLib.

https://github.com/unified-font-object/ufoLib/blob/master/Lib/ufoLib/filenames.py

I haven't checked, are there any differences between the code from ufoLib and that from the ufoNormalizer that you referenced here?

Also, we need to add tests for this before merging, to keep the overall test coverage.

bamidei:
@anthrotype Updated with the items you highlighted - moved to its own file; added some unit tests.

The contents of the two files are almost identical - non-functional differences as far as I can see.  When I moved the code to its own file, I sourced the file that you referenced instead of the one I had initially sourced.

anthrotype:
thanks @bamidei, I'll do some cleanups and merge the split-glyphs branch into master soon

Lib/fontTools/ttLib/tables/_g_l_y_f.py
anthrotype: I am not sure about that `_` separating the `_g_l_y_f` table tag from the glyph name. It's not clear where the table tag ends and where the glyph name begins. A dot could be better in that regard (though there's `.notdef` or `.null` which may look odd with two dots).

You know, I'm thinking that perhaps for the second split level (the first being the individual tables) we group the files in a sub-directory instead of appending a suffix (the glyph name) to the table file name.

So when -g option (splitGlyphs) is used, the glyf table is dumped like that:

- the individual glyphs' ttx files are dumped inside a new sudirectory called FontName._g_l_y_f (or FontName#1._g_l_y_f etc. when dumped more than once to the same destination)
- an index file named FontName._g_l_y_f.ttx contains the list of filenames; it doesn't need to verbosely specify the sub-path, but it can just presume that the glyph files are under a subdirectory called FontName._g_l_y_f.
i.e. instead of the current:

```xml
  <glyf>
    <TTGlyph src="CALIBRI._g_l_y_f_.notdef.ttx"/>
    <TTGlyph src="CALIBRI._g_l_y_f_A_.ttx"/>
```

we could have something like:

```xml
  <glyf>
    <TTGlyph src=".notdef.ttx"/>
    <TTGlyph src="A_.ttx"/>
```

The index file FontName._g_l_y_f.ttx and the subdirectory FontName._g_l_y_f will nicely show up next to each other in the directory's file listing, and the thousands of glyph files will not be mixed up with the rest of the individual tables, but will be logically grouped together.

WDYT?

bamidei: I think that using a directory name for the prefix is a great idea.  I think that we might want to leave the directory name in the index table as well though so that it is explicitly readable without somewhat hidden context information.

i.e.:

`  <glyf>
    <TTGlyph src="Fontname._g_l_y_f/.notdef.ttx"/>
    <TTGlyph src="Fontname._g_l_y_f/A_.ttx"/>
`
This may also keep the code (xmlReader) a bit simpler - again not needing to pull context information - always treating 'src' values the same.

bamidei:
@anthrotype Any thoughts on when this branch will get into master and a release?  I'd like to start using an official release rather than our current branch.

If there is anything that needs to be done to prep for merge, please let me know and I may be able to help out. 

Thank you,
  Bill

anthrotype:
@bamidei hey, sorry for the delay (I was away), and thanks for the reminder. I can't remember right now what the remaining issues were. I'll try to take another look next week. If anyone else would also like to review this `split-glyphs` branch, please do thanks.

bamidei:
@anthrotype np.  Thanks.  Were you able to get a chance to look at this last week?  Anything I can do to help?

anthrotype:
Not yet. Maybe open a pull request from the split-glyphs branch to master

anthrotype:
there are lots of merge conflicts, the branch needs to be rebased on the latest master
https://github.com/fonttools/fonttools/compare/master...split-glyphs

i'd say, @bamidei you can fix the merge conflicts from a branch on your fork, then send a PR to upstream/mastrer from there and will continue from there.

bamidei:
@anthrotype k.  I'll rebase and resolve the new merge conflicts.  Do you have time later this week to review and resolve merge?  I would prefer not to need to do this another time and so would like to coordinate the work for when we'll both have time to focus and finish this integration.

anthrotype:
@bamidei I'll rebase myself and merge it today. Thanks

bamidei:
@anthrotype Thank you!  I really appreciate you doing the rebase and moving the PR forward.

Issue #612, opened 2016-05-28
[feaLib] handling of mixed single and multiple or ligature substitutions


khaledhosny:
If I have a feature like this:

```
feature isol {
 sub a by A;
 sub b by B C;
} isol;
```

feaLib will generate two lookups, a single substitution lookup and a multiple substitution one. But IMO it should have been a single multiple substitution lookup, as the first rule can be seen as a multiple substitution lookup with just one replacement glyph. The same happens when mixing single substitutions with ligature substitutions.

In many of my fonts this results in 10s of lookups which can slow the OpenType layout processing in some implementations. It does not help that the feature syntax does not provide any way to make this explicit, since the three different substitution subtypes use the same syntax.

It gets even worse with named lookups, a lookup like this will not even compile:

```
lookup isol0 {
 sub a by A;
 sub b by B C;
} isol0;
```

FontForge used to handle this gracefully and guess the right lookup type in these situations. I didn’t check with FDK does here.

brawer:
Currently, it seems that feaLib and FDK produce identical output for these snippets. But of course that doesn’t mean they absolutely have to: feaLib could certainly make optimizations which makeotf doesn’t make, as long as everyone conforms to the published specification. However, it would add friction for font designers if a given feature file is accepted by some implementations but rejected by others.

Here’s the output from running makeotf v2.0.90 on the first example:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<ttFont sfntVersion="true" ttLibVersion="3.0">

  <GSUB>
    <Version value="1.0"/>
    <ScriptList>
      <!-- ScriptCount=1 -->
      <ScriptRecord index="0">
        <ScriptTag value="DFLT"/>
        <Script>
          <DefaultLangSys>
            <ReqFeatureIndex value="65535"/>
            <!-- FeatureCount=1 -->
            <FeatureIndex index="0" value="0"/>
          </DefaultLangSys>
          <!-- LangSysCount=0 -->
        </Script>
      </ScriptRecord>
    </ScriptList>
    <FeatureList>
      <!-- FeatureCount=1 -->
      <FeatureRecord index="0">
        <FeatureTag value="isol"/>
        <Feature>
          <!-- LookupCount=2 -->
          <LookupListIndex index="0" value="0"/>
          <LookupListIndex index="1" value="1"/>
        </Feature>
      </FeatureRecord>
    </FeatureList>
    <LookupList>
      <!-- LookupCount=2 -->
      <Lookup index="0">
[...]

PR #1133, opened 2017-12-21
[feaLib] Allow mixed single/multiple substitutions

khaledhosny:
A multiple substitution may have a single destination, in which case it will look just like a single substitution. So if there are both multiple and single substitutions, upgrade all the single ones to multiple substitutions. Previously we would just give an error message in the builder and abort, which means certain valid OpenType lookups can’t be represented by feature files.

This is the same logic implemented by FontForge (actually the explanation above is almost copied verbatim from its source), makeotf does not do this AFAIK but I consider it a bug not a feature.

Fixes https://github.com/fonttools/fonttools/issues/612

### DIFF
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -1295,8 +1295,8 @@ class Parser(object):
             for i, s in enumerate(statements):
                 if isinstance(s, self.ast.SingleSubstStatement):
                     statements[i] = self.ast.MultipleSubstStatement(s.location,
-                        s.prefix, tuple(s.glyphs[0].glyphSet())[0], s.suffix,
-                        tuple([list(r.glyphSet())[0] for r in s.replacements]))
+                        s.prefix, s.glyphs[0].glyphSet()[0], s.suffix,
+                        [r.glyphSet()[0] for r in s.replacements])
 
     def is_cur_keyword_(self, k):
         if self.cur_token_type_ is Lexer.NAME:
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -1276,6 +1276,28 @@ class Parser(object):
                                   self.cur_token_location_)
         self.expect_symbol_(";")
 
+        # A multiple substitution may have a single destination, in which case
+        # it will look just like a single substitution. So if there are both
+        # multiple and single substitutions, upgrade all the single ones to
+        # multiple substitutions.
+
+        # Check if we have a mix of non-contextual singles and multiples.
+        has_single = False
+        has_multiple = False
+        for s in statements:
+            if isinstance(s, self.ast.SingleSubstStatement):
+                has_single = not any([s.prefix, s.suffix, s.forceChain])
+            elif isinstance(s, self.ast.MultipleSubstStatement):
+                has_multiple = not any([s.prefix, s.suffix])
+
+        # Upgrade all single substitutions to multiple substitutions.
+        if has_single and has_multiple:
+            for i, s in enumerate(statements):
+                if isinstance(s, self.ast.SingleSubstStatement):
+                    statements[i] = self.ast.MultipleSubstStatement(s.location,
+                        s.prefix, tuple(s.glyphs[0].glyphSet())[0], s.suffix,
+                        tuple([list(r.glyphSet())[0] for r in s.replacements]))
+
     def is_cur_keyword_(self, k):
         if self.cur_token_type_ is Lexer.NAME:
             if isinstance(k, type("")):  # basestring is gone in Python3
--- a/Tests/feaLib/parser_test.py
+++ b/Tests/feaLib/parser_test.py
@@ -1256,6 +1256,18 @@ class ParserTest(unittest.TestCase):
                         "    sub a' lookup upper x x A' lookup lower;"
                         "} test;")
 
+    def test_substitute_mix_single_multiple(self):
+        doc = self.parse("lookup Look {"
+                         "  sub f_f   by f f;"
+                         "  sub f     by f;"
+                         "  sub f_f_i by f f i;"
+                         "} Look;")
+        statements = doc.statements[0].statements
+        for sub in statements:
+            self.assertIsInstance(sub, ast.MultipleSubstStatement)
+        self.assertEqual(statements[1].glyph, "f")
+        self.assertEqual(statements[1].replacement, ["f"])
+
     def test_substitute_from(self):  # GSUB LookupType 3
         doc = self.parse("feature test {"
                          "  substitute a from [a.1 a.2 a.3];"


khaledhosny:
No tests for now, I’m just trying to see if this approach is OK and will add some tests later.

Lib/fontTools/feaLib/parser.py
```diff
@@ -1276,6 +1276,28 @@ def parse_block_(self, block, vertical, stylisticset=None,
...
+            elif isinstance(s, self.ast.MultipleSubstStatement):
+                has_multiple = not any([s.prefix, s.suffix])
+
+        # Upgrade all single substitutions to multiple substitutions.
+        if has_single and has_multiple:
+            for i, s in enumerate(statements):
+                if isinstance(s, self.ast.SingleSubstStatement):
+                    statements[i] = self.ast.MultipleSubstStatement(s.location,
+                        s.prefix, tuple(s.glyphs[0].glyphSet())[0], s.suffix,
+                        tuple([list(r.glyphSet())[0] for r in s.replacements]))
```
anthrotype: You can pass a generator expression to tuple() constructor, don’t need to make a list first

anthrotype:
Thanks Khaled, looks good to me. We can also propose a change to FEA spec to explicitly allow this
situation if it’s not the case already.

anthrotype:
Oh now I remember we had a long discussion a few months back in #612.
It’s not clear what the conclusion was.

/cc @behdad @readroberts @brawer and others

khaledhosny:
I don’t think the issue is as controversial, a multiple substitution with just single replacement glyph is valid multiple substitution. Not allowing this makes many constructs needlessly hard. Here is an excerpet from a project I’m porting from FontForge to feaLib:
```
lookup break_kashida {
  sub uni0640.7 by uni0640 uni0640 uni0640 uni0640 uni0640 uni0640 uni0640;
  sub uni0640.6 by uni0640 uni0640 uni0640 uni0640 uni0640 uni0640;
  sub uni0640.5 by uni0640 uni0640 uni0640 uni0640 uni0640;
  sub uni0640.4 by uni0640 uni0640 uni0640 uni0640;
  sub uni0640.3 by uni0640 uni0640 uni0640;
  sub uni0640.2 by uni0640 uni0640;
  sub uni0640.1 by uni0640;
} break_kashida;

@curved_kashida = [uni0640.1 uni0640.2 uni0640.3 uni0640.4 uni0640.5 uni0640.6 uni0640.7];
@kashida = [uni0640 @curved_kashida];

feature rlig {
  lookupflag IgnoreMarks;
  sub @kashida @curved_kashida' lookup break_kashida;
} rlig;
```

Without this patch I need to break this into something like:
```
lookup break_kashida1 {;
  sub uni0640.1 by uni0640;
} break_kashida1;

lookup break_kashida2 {
  sub uni0640.7 by uni0640 uni0640 uni0640 uni0640 uni0640 uni0640 uni0640;
  sub uni0640.6 by uni0640 uni0640 uni0640 uni0640 uni0640 uni0640;
  sub uni0640.5 by uni0640 uni0640 uni0640 uni0640 uni0640;
  sub uni0640.4 by uni0640 uni0640 uni0640 uni0640;
  sub uni0640.3 by uni0640 uni0640 uni0640;
  sub uni0640.2 by uni0640 uni0640;
} break_kashida2;

@curved_kashida1 = [uni0640.1];
@curved_kashida2 = [uni0640.2 uni0640.3 uni0640.4 uni0640.5 uni0640.6 uni0640.7];
@kashida1 = [uni0640 @curved_kashida1];
@kashida2 = [uni0640 @curved_kashida2];

feature rlig {
  lookupflag IgnoreMarks;
  sub @kashida1 @curved_kashida1' lookup break_kashida1;
  sub @kashida2 @curved_kashida2' lookup break_kashida2;
} rlig;
```

doubling the number of lookups needlessly (and when you do this 100 of time it can slow some OT layout engines considerably) and making it harder to follow.

anthrotype:
I'm in favour of merging, but I'd like to also hear from those who raised concerns in the other thread.

brawer:
Sounds good to me. I’ve filed https://github.com/adobe-type-tools/afdko/issues/224 to keep @readroberts in the loop.

anthrotype:
Thanks Sascha. Khaled, we can add a test and then merge

benkiel:
I'm unclear why this was merged when there is an outstanding question to @readroberts and @anthrotype's request to hear from those who raised concerns hasn't happened.

behdad:
My stance is unchanged from other issue discussion.  This is an FDK bug we don't want to replicate.  So I'm supportive of Khaled's change.

benkiel:
@behdad I reread that prior discussion again, and at the risk of coming off as a pain, in that discussion, @khaledhosny only offered his opinion that it was a bug: both @readroberts and @brawer didn't confirm or agree that it was a bug. I would argue that at least @readroberts should weigh in about about this change before it's made, and an update to the FEA spec should be made also to make this situation clear. Right now, fea code that will compile with fealib will now not compile in makeOTF — and I think that's a problem. Yes, I do understand that FontForge will also allow this behavior, but will all respect, it's implementation of fea sytanx isn't something that I would consider an always reliable reference.

readroberts:
I will discuss this with Miguel Sousa, once he is back from the Adobe break on Jan 2nd.  I also want to think about this some more. It is true that the FDK does stuff like re-order ligatures, but this is explicitly documented in the spec. We all agree that users should not have to break the examples into separate lookups.  The problem with doing it silently is that you lose the ability to detect when a user specified a mixed lookup type unintentionally. Yes, it useful to have the rules converted when you intend that to happen, but typos are pretty common, too. This is the main reason I would argue for an explicit indicator, either for single lookups or for an entire lookup, that conversion is desired.

khaledhosny:
I feel we are overthinking this. The solution suggested here does not break any existing valid feature files, simple, clear enough and predictable. Typos happen all the time and I don’t see why we need to worry about a `sub f_f by f;` typo more than we would worry about a `sub f_f by f l;` one.

behdad:
> @behdad I reread that prior discussion again, and at the risk of coming off as a pain, in that discussion, @khaledhosny only offered his opinion that it was a bug: both @readroberts and @brawer didn't confirm or agree that it was a bug. I would argue that at least @readroberts should weigh in about about this change before it's made, and an update to the FEA spec should be made also to make this situation clear. Right now, fea code that will compile with fealib will now not compile in makeOTF — and I think that's a problem. Yes, I do understand that FontForge will also allow this behavior, but will all respect, it's implementation of fea sytanx isn't something that I would consider an always reliable reference.

It's ok to disagree. If we were to bound ourselves but every bug in every other piece of software we try to supersede, we'd never be able to innovate.  The benchmark here must be that feature files accepted by makeOTF must always be accepted by feaLib. There's no guarantee re the other way around.  Indeed, feaLib already accepts feature files that makeOTF doesn't, because feaLib does automatic extension subtables and subtable breaks that makeOTF doesn't.

In this case it's clear to me, and Khaled, that the spec should allow this.  There's no uncertainty to the semantics of this case.  I don't accept handwavy unclear "software compat issues" or "font developer must be in control".  I build software that Just Works.

readroberts:
I agree we should go ahead and allow this. My use case about noticing errors that re not meant to be mixed case can be addresses by the implementation, by issuing a warning.

PR #1136, opened 2018-01-01
svgLib: Add support for arcs in SVG path parser

santhoshtr:
Implement the arc to curve points calculation using the original
svg.path library and use the curve points for the pen.

The Arc class and methods to parameterize and get curve points are
copied.

This makes the svg to glif conversion possible for any SVGs.

Follow up for 71e97b3e74f2cc8061dfd3ada6d436d7ff3173d1

### DIFF
--- a/Lib/fontTools/svgLib/path/parser.py
+++ b/Lib/fontTools/svgLib/path/parser.py
@@ -23,7 +23,7 @@ class Arc(object):
 
     def __init__(self, start, radius, rotation, arc, sweep, end):
         """radius is complex, rotation is in degrees,
-           large and sweep are 1 or 0 (True/False also work)"""
+           arc and sweep are 1 or 0 (True/False also work)"""
 
         self.start = start
         self.radius = radius
@@ -317,17 +317,15 @@ def parse_path(pathdef, pen, current_pos=(0, 0)):
                 end += current_pos
 
             svg_arc = Arc(current_pos, radius, rotation, arc, sweep, end)
-            arc_points = []
+            arc_points = [(current_pos.real, current_pos.imag)]
             for x in range(1, 5):
                 # There are infinite points in an arc, but for our context,
                 # define the arc using 5 points(0.2, 0.4, 0.6...)
                 arc_point = svg_arc.point(x*0.2)
                 arc_points.append((arc_point.real, arc_point.imag))
-            pen.qCurveTo(
-                (current_pos.real, current_pos.imag),
-                *arc_points,
-                (end.real, end.imag)
-            )
+            arc_points.append((end.real, end.imag))
+
+            pen.qCurveTo(*arc_points)
             current_pos = end
 
     # no final Z command, it's an open path
--- a/Lib/fontTools/svgLib/path/parser.py
+++ b/Lib/fontTools/svgLib/path/parser.py
@@ -127,9 +127,6 @@ def parse_path(pathdef, pen, current_pos=(0, 0)):
 
     If 'current_pos' (2-float tuple) is provided, the initial moveTo will
     be relative to that instead being absolute.
-
-    Arc segments (commands "A" or "a") are not currently supported, and raise
-    NotImplementedError.
     """
     # In the SVG specs, initial movetos are absolute, even if
     # specified as 'm'. This is the default behavior here as well.
@@ -310,21 +307,19 @@ def parse_path(pathdef, pen, current_pos=(0, 0)):
             end = float(elements.pop()) + float(elements.pop()) * 1j
 
             if not absolute:
-                if end == 0:
-                    # Guard against a situation where arc start and end being same.
-                    # That results division by zero issues in Arc parameterization.
-                    end = 0.00009
                 end += current_pos
 
+            if end == current_pos:
+                # Guard against a situation where arc start and end being same.
+                # That results division by zero issues in Arc parameterization.
+                end += 0.00009
             svg_arc = Arc(current_pos, radius, rotation, arc, sweep, end)
-            arc_points = [(current_pos.real, current_pos.imag)]
-            for x in range(1, 5):
+            arc_points = []
+            for point in [0.2, 0.4, 0.6, 0.8, 1]:
                 # There are infinite points in an arc, but for our context,
-                # define the arc using 5 points(0.2, 0.4, 0.6...)
-                arc_point = svg_arc.point(x*0.2)
+                # define the arc using 5 points.
+                arc_point = svg_arc.point(point)
                 arc_points.append((arc_point.real, arc_point.imag))
-            arc_points.append((end.real, end.imag))
-
             pen.qCurveTo(*arc_points)
             current_pos = end
 
--- a/Lib/fontTools/svgLib/path/parser.py
+++ b/Lib/fontTools/svgLib/path/parser.py
@@ -10,6 +10,7 @@
 from __future__ import (
     print_function, division, absolute_import, unicode_literals)
 from fontTools.misc.py23 import *
+from math import sqrt, cos, sin, acos, degrees, radians
 import re
 
 COMMANDS = set('MmZzLlHhVvCcSsQqTtAa')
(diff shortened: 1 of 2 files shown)


santhoshtr:
Here is a sample svg having arcs.   Snippets/svg2glif.py should be able to convert this svg to glif successfully.
[arc-test.zip](https://github.com/fonttools/fonttools/files/1596118/arc-test.zip)

Tests/svgLib/path/parser_test.py
```diff
@@ -226,6 +226,21 @@
                 ("closePath", ()),
             ]
         ),
+        # absolute A command
+        (
+            "M 100 100 A 150 150 0 1 0 150 -150 z",
+            [
+                ('moveTo', ((100.0, 100.0),)),
+                ('qCurveTo', ((100.0, 100.0),
+                    (217.1758387643403, 139.78681395054122),
```
santhoshtr: Should we round this floats?

Lib/fontTools/svgLib/path/parser.py
```diff
@@ -18,6 +19,98 @@
 COMMAND_RE = re.compile("([MmZzLlHhVvCcSsQqTtAa])")
 FLOAT_RE = re.compile("[-+]?[0-9]*\.?[0-9]+(?:[eE][-+]?[0-9]+)?")
 
+class Arc(object):
+
+    def __init__(self, start, radius, rotation, arc, sweep, end):
+        """radius is complex, rotation is in degrees,
+           large and sweep are 1 or 0 (True/False also work)"""
```
anthrotype: perhaps here you mean "arc" instead of "large"?

Lib/fontTools/svgLib/path/parser.py
```diff
@@ -209,7 +302,33 @@ def parse_path(pathdef, pen, current_pos=(0, 0)):
...
+            arc_points = []
+            for x in range(1, 5):
+                # There are infinite points in an arc, but for our context,
+                # define the arc using 5 points(0.2, 0.4, 0.6...)
+                arc_point = svg_arc.point(x*0.2)
+                arc_points.append((arc_point.real, arc_point.imag))
+            pen.qCurveTo(
+                (current_pos.real, current_pos.imag),
+                *arc_points,
+                (end.real, end.imag)
```
anthrotype: This fails on python < 3.5

`SyntaxError: only named arguments may follow *expression`

https://travis-ci.org/fonttools/fonttools/jobs/323793110#L588

Tests/svgLib/path/parser_test.py
anthrotype: we are asserting equality, so we can't round.
if you want to round, you could try using `pytest.approx`
https://docs.pytest.org/en/latest/builtin.html#pytest.approx

anthrotype: looking at the CI test failures, it looks like we do need to test for almost equality.
Can you try changing line 250 of parser_test.py like this:

```diff
-    assert pen.value == expected
+    assert pen.value == pytest.approx(expected)
```

Tests/svgLib/path/parser_test.py
```diff
@@ -226,6 +226,21 @@
                 ("closePath", ()),
             ]
         ),
+        # absolute A command
+        (
+            "M 100 100 A 150 150 0 1 0 150 -150 z",
+            [
+                ('moveTo', ((100.0, 100.0),)),
+                ('qCurveTo', ((100.0, 100.0),
```
anthrotype: why the duplicate point here?

Lib/fontTools/svgLib/path/parser.py
```diff
@@ -317,17 +317,15 @@ def parse_path(pathdef, pen, current_pos=(0, 0)):
                 end += current_pos
 
             svg_arc = Arc(current_pos, radius, rotation, arc, sweep, end)
-            arc_points = []
+            arc_points = [(current_pos.real, current_pos.imag)]
```
anthrotype: actually, you don't have to pass the current_pos to qCurveTo. That's implied by the preceding moveTo. If you do that, you end up with a duplicate point (see my other comment)

anthrotype:
@santhoshtr can you also make sure all new lines added are covered by tests?

you can look online at https://codecov.io/gh/fonttools/fonttools (but this may be incomplete as some builds failed), or you can check locally by running `tox -e py27-cov` or `tox -e py36-cov`; you'll get a htmcov/index.html where you can browse the test coverage.

Tests/svgLib/path/parser_test.py
```diff
@@ -232,23 +232,28 @@
...
 )
+
 def test_parse_path(pathdef, expected):
     pen = RecordingPen()
     parse_path(pathdef, pen)
 
-    assert pen.value == expected
-
+    assert len(pen.value) == len(expected)
+    for i, (instr, coords) in enumerate(pen.value):
```
anthrotype: i think it's better if you use `zip` to iterate over the two lists instead of enumerate and indexing.
Also you need to assert the tuples of coordinates have the same length.
And it's not "real" and "imag" in the tuples, but x and y floats.
 
```diff
diff --git a/Tests/svgLib/path/parser_test.py b/Tests/svgLib/path/parser_test.py
index f8aa0951..3e001bb7 100644
--- a/Tests/svgLib/path/parser_test.py
+++ b/Tests/svgLib/path/parser_test.py
@@ -249,11 +249,12 @@ def test_parse_path(pathdef, expected):
     parse_path(pathdef, pen)
 
     assert len(pen.value) == len(expected)
-    for i, (instr, coords) in enumerate(pen.value):
-        # instr are 'moveTo', 'qCurveTo', 'closePath' etc.
-        assert instr == expected[i][0]
-        for j, (real, imag) in enumerate(coords):
-            assert pytest.approx((real, imag)) == expected[i][1][j]
+    for (instr, coords), (exp_instr, exp_coords) in zip(pen.value, expected):
+        assert instr == exp_instr
+        assert len(coords) == len(exp_coords)
+        for c, e in zip(coords, exp_coords):
+            assert c == pytest.approx(e)
+
 
 @pytest.mark.parametrize(
     "pathdef1, pathdef2",
```

anthrotype:
Thank you @santhoshtr 
I'll merge as soon as all the CI complete

santhoshtr:
Thanks @anthrotype!

Removing overlaps for the qCurves we generated from arcs is not supported by fontmake yet. Once that also available, I think I have everything in place for my workflow :)

anthrotype:
@santhoshtr actually, on second thought, I think I'll have to revert this...
It's my fault I didn't check the math was correct before merging.
The code you copied from svg.path doesn't compute the (approximate) Bezier control points. It computes an endpoint parametrization of elliptical arcs to allow evaluating points _on the arc_ at a given `t` between 0 and 1.
But then you are using the sampled points to call the `pen.qCurveTo` method, which instead takes a series of off-the-curve control points and a final on-curve point.

Approximating an elliptical arc with a bezier curve is a bit more complicated than that.
@behdad suggested to look at how cairo does that.

I'm sorry I have to revert this for now...

justvanrossum:
Here is some DrawBot code I wrote a long time ago. Maybe some if it is useful.

```python
x = y = 320
width = x * 2
size(width, width)


def blob(x, y, diameter=4):
    radius = diameter / 2
    oval(x - radius, y - radius, diameter, diameter)


NINETY = radians(90)

def arcSegmentToBezier(x, y, radius, start, end):
    start = radians(start)
    end = radians(end)
    half = 0.5 * abs(end - start)

    ax, ay = x + radius * cos(start), y + radius * sin(start)
    bx, by = x + radius * cos(end), y + radius * sin(end)
    startp = start + NINETY  # angle of the first bcp line
    endp = end - NINETY      # angle of the other bcp line

    D = radius * (1 - cos(half))
    X = (4 * D / 3) / sin(half)  # length of the bcp line

    cx, cy = ax + X * cos(startp), ay + X * sin(startp)
    dx, dy = bx + X * cos(endp), by + X * sin(endp)

    return (ax, ay), (cx, cy), (dx, dy), (bx, by)

def arcToBezier(x, y, radius, start, end, clockwise=False):
    if clockwise:
        start, end = end, start
    while end < start:
        end += 360
    points = []
    didMoveTo = False
    angle = start
    while angle < end:
        next = (angle // 90 + 1) * 90
        if next > end:
            next = end
        pt0, pt1, pt2, pt3 = arcSegmentToBezier(x, y,
                                                radius, angle, next)
        if not didMoveTo:
            points.append(pt0)
            didMoveTo = True
        points.extend([pt1, pt2, pt3])
        angle = next

    if clockwise:
        points.reverse()
    return points

def drawPoints(points):
    moveTo(points[0])
    for i in range(1, len(points), 3):
        curveTo(*points[i:i+3])


translate(0.5, 0.5)

tests = []
for r in range(20, 320, 20):
    tests.append((r, random() * 360, random() * 720, randint(0, 1)))

for radius, start, end, clockwise in tests:
    fill(None)
    stroke(0)
    strokeWidth(2)
    newPath()
    arc((x, y), radius, start, end, clockwise)
    drawPath()

    stroke(0.5, 1, 0, 0.4)
    strokeWidth(12)
    newPath()
    points = arcToBezier(x, y, radius, start, end, clockwise)
    drawPoints(points)
    drawPath()
    if clockwise:
        r, g, b = 1, 0, 0
    else:
        r, g, b = 0, 0, 1

    fill(r, g, b)
    stroke(None)
    for px, py in points:
        blob(px, py)
    stroke(r, g, b)
    fill(None)
    strokeWidth(1)
    for i in range(0, len(points) - 1, 3):
        ax, ay = points[i]
        bx, by = points[i + 1]
        line((ax, ay), (bx, by))
        ax, ay = points[i + 2]
        bx, by = points[i + 3]
        line((ax, ay), (bx, by))
```

anthrotype:
Thanks Just!

brawer:
This paper might be relevant: Aleksas Riškus, “Approximation of a Cubic Bezier Curve by Circular Arcs and Vice Versa,” Information Technology and Control, 35(4), 2006 pp. 371-378. [See here for PDF](https://pdfs.semanticscholar.org/1639/0db1a470bd13fe428e0896671a9a5745070a.pdf).

PR #1137, opened 2018-01-03
`*.afm` files are missing from MANIFEST.in

andir:
`*.afm` files are missing from MANIFEST.in causing `make check` to fail on pypi
releases. This PR adds the missing line.


anthrotype:
Thanks for spotting this!

The reason we didn't catch this is that we normally run the tests from the root of the git source tree.
I'll think of a way to catch these sorts of issues automatically.

anthrotype:
And also thanks for working on packaging fonttools for NixOS :+1:

Issue #164, opened 2014-09-23
ast.literal_eval is useless


behdad:
We switched to it as our safeEval, but it can't even evaluate "2*2"!

PR #1139, opened 2018-01-04
textTools: use ast.literal_eval for safeEval

anthrotype:
go back to using `literal_eval`, as nobody does complex expressions in TTX files, and eval is evil.

Fixes https://github.com/fonttools/fonttools/issues/164

### DIFF
--- a/Lib/fontTools/misc/textTools.py
+++ b/Lib/fontTools/misc/textTools.py
@@ -7,14 +7,14 @@ import ast
 import string
 
 
-def safeEval(data):
-	"""A (kindof) safe replacement for eval."""
-	return ast.literal_eval(data)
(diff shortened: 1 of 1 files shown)


Lib/fontTools/misc/textTools.py
```diff
@@ -3,12 +3,13 @@
...
 from fontTools.misc.py23 import *
+import ast
 import string
 
 
-def safeEval(data, eval=eval):
+def safeEval(data):
 	"""A (kindof) safe replacement for eval."""
-	return eval(data, {"__builtins__":{"True":True,"False":False}})
+	return ast.literal_eval(data)
```
adrientetar: safeEval = ast.literal_eval ?

PR #1144, opened 2018-01-09
guessing type1 typ for py2-3

typemytype:
getMacCreatorAndType requires `xattr` and ``MacOS` both not available
for py3

@justvanrossum

### DIFF
--- a/Lib/fontTools/t1Lib/__init__.py
+++ b/Lib/fontTools/t1Lib/__init__.py
@@ -101,14 +101,14 @@ class T1Font(object):
 
 def read(path, onlyHeader=False):
 	"""reads any Type 1 font file, returns raw data"""
-	normpath = path.lower()
-	creator, typ = getMacCreatorAndType(path)
-	if typ == 'LWFN':
+	stat = os.stat(path)
+	# LWFN has no size
+	if stat.st_size == 0:
 		return readLWFN(path, onlyHeader), 'LWFN'
-	if normpath[-4:] == '.pfb':
+	_, ext = os.path.splitext(path)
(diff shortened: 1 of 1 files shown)


anthrotype:
`xattr` is available as a third-party module, works for python 2 and 3: https://pypi.python.org/pypi/xattr/0.9.2

anthrotype:
an empty file with stat.st_size() == 0 does not necessarily have type `LWFN`

anthrotype:
it's true that LWFN being stored in the resource fork has an empty data fork, but the reverse test is too crude. I think you should install the extra dependency if you're interested in getting mac type of a file

anthrotype:
if the problem is the fact the lack of precompiled wheels for xattr and the consequent need for a compiler, we can fix that by proposing the maintainers to produce them.

I can see they have Travis setup that builds and uploads the wheels to PyPI, however they are using pyenv to install the different python versions on Travis. Pyenv compiles CPython from source so it is targeted to the system it is build on (Travis OSX images run El Capitan 10.11). Therefore the wheels generated have thea same macosx deployment target version: e.g. xattr-*-macosx_10_11_x86_64.whl`.

They should rather use the Python.org distribution of CPython that is built on Mac OS 10.6 with a "fat" architecture (both 32 and 64 bit), so the generated wheels are most compatible with any other python distribution and osx version, and not just >= 10.11 and 64-bit.

Also they only have up to python 3.5, so there's no 3.6 wheels yet.

justvanrossum:
xattr builds fine for Python 3.6 on 10.10, so I think that may be good enough. I've sent @typemytype a wheel for him to test. I think we can close this PR.

Issue #1112, opened 2017-11-23
My HarfBuzz-based Unicode script to OpenType script mapping builder


twardoch:
Hi @anthrotype 

Please check out my [`hb_scripts3.py`](https://github.com/twardoch/feaLab/blob/master/Lib/feaLab/hb_scripts3.py) mini-module. 

It’s a mini-module to convert between Unicode and OpenType script tags. It uses the [harfpy](https://github.com/ldo/harfpy) ctypes-based Python 3 bindings for HarfBuzz. Use [my fork](https://github.com/twardoch/harfpy) until my [PR](https://github.com/ldo/harfpy/pull/12) is merged upstream. It uses the Unicode-to-OpenType script tag [conversion logic from HarfBuzz](https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ot-tag.cc). 

Perhaps you might want to move the essential bits into `MetaTools` as a new builder for stuff in `unicodedata`. I would have made it myself but since the HarfBuzz bindings I used are Python 3 only, I didn’t have to mingle directly with the tools inside `MetaTools`. I think it’s OK to have a Python 3-only tool with a HarfBuzz dependancy in `MetaTools` though, right? 

I’ve already written a function in my `hb_scripts3.py` module called `getIsoToOtScriptMap` which takes all the Unicode script tags and makes a dict like this: 

```
OTSCRIPTS = { 
    'Adlm': ['adlm'],
...
    'Arab': ['arab'],
...
    'Beng': ['bng2', 'beng'],
...
    'Cyrl': ['cyrl'],
    'Deva': ['dev2', 'deva'],
...
    'Latn': ['latn'],
...
    'Yiii': ['yi  '],
    'Zanb': ['zanb'],
    'Zinh': ['DFLT'],
    'Zyyy': ['DFLT'],
    'Zzzz': ['DFLT']
}
```

As you notice, the values are lists because of the OT scripts that have multiple implementations. 

Then, you could provide some simple functions in `unicodedata` for looking up the OT scripts based on that OTSCRIPTS mapping — this is why I asked you to use the Unicode script tags (short names) as keys, btw :) 

I have also written a simple function `updateLanguageSystemsInFea` which uses the code to auto-generate `languagesystem` statements in a FEA based on a series of Unicode codepoints (normally, the font’s `cmap` table). Feel free to rewrite it to use the updated `fontTools.unicodedata` thing and then use it in `ufo2ft` as a `languageSystemWriter` or something.

[...]

PR #1150, opened 2018-01-18
unicodedata: add ot_tags_from_script function

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/1112

This implements the same logic found in harbfuzz [hb-ot-tag.cc](https://github.com/harfbuzz/harfbuzz/blob/461a605fdec3361a038d3715adf615353c4f91fa/src/hb-ot-tag.cc#L127) to convert between Unicode (or ISO 15924) script codes to [OpenType script tags](https://www.microsoft.com/typography/otspec/scripttags.htm).

/cc @twardoch

### DIFF
--- a/Lib/fontTools/unicodedata/OTTags.py
+++ b/Lib/fontTools/unicodedata/OTTags.py
@@ -1,0 +1,37 @@
+# Data updated to OpenType 1.8.2 as of January 2018.
+
+# Complete list of OpenType script tags at:
+# https://www.microsoft.com/typography/otspec/scripttags.htm
+
+# Most of the script tags are the same as the ISO 15924 tag but lowercased,
+# so we only have to handle the exceptional cases:
+# - KATAKANA and HIRAGANA both map to 'kana';
+# - spaces at the end are preserved, unlike ISO 15924;
+# - we map special script codes for Inherited, Common and Unknown to DFLT.
+
+DEFAULT_SCRIPT = "DFLT"
+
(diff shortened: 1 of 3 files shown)


anthrotype:
It would be nice if someone could check I didn't miss anything.

One thing I noticed while doing this is that there are some Unicode script codes that seems to be missing from the Microsoft's scripttags.html page. They'll probably too recent and will be added in the next version of the spec. My function simply lower-cases them.

```
Gonm: Masaram Gondi
Hrkt: Katakana Or Hiragana
Nshu: Nushu
Soyo: Soyombo
Zanb: Zanabazar Square
```

anthrotype:
well, actually `Hrkt` becomes `kana`, like `Hira`

anthrotype:
(that was my logical assumption, but perhaps i'm wrong and it will actually become `hrkt`?)

mashabow:
There is no `hrkt` in OT script tags, so you're right.

|From: ISO|To: OT|
|--|--|
|`Hira` (Hiragana)|`kana`|
|`Kana` (Katakana)|`kana`|
|`Hrkt` (Hiragana+Katakana)|`kana`|

anthrotype:
@mashabow thanks for checking. Then there must a bug in HarfBuzz because it's `hb_ot_tags_from_script` function is returning `hrkt` as opentype script tag instead of `kana` for `Hrkt` unicode script code. I'll open a separate issue for that.

Issue #1151, opened 2018-01-19
Bogus nameID comments in FeatureParamsCharacterVariants block


miguelsousa:
[Source Code Pro fonts](https://github.com/adobe-fonts/source-code-pro/tree/release) support a few [Character Variant features](https://www.microsoft.com/typography/otspec/features_ae.htm#cv01-cv99). When dumping their GSUB table, the string stored in `nameID 0` (Copyright) is used in the nameID parameters in **FeatureParamsCharacterVariants** that are meant to be NULL.
```xml
<FeatureParamsCharacterVariants Format="0">
  <Format value="0"/>
  <FeatUILabelNameID value="256"/>  <!-- simple a -->
  <FeatUITooltipTextNameID value="0"/>  <!-- Copyright 2010, 2012 Adobe Systems Incorporated (http://www.adobe.com/), with Reserved Font Name ‘Source’. -->
  <SampleTextNameID value="0"/>  <!-- Copyright 2010, 2012 Adobe Systems Incorporated (http://www.adobe.com/), with Reserved Font Name ‘Source’. -->
  <NumNamedParameters value="0"/>
  <FirstParamUILabelNameID value="0"/>  <!-- Copyright 2010, 2012 Adobe Systems Incorporated (http://www.adobe.com/), with Reserved Font Name ‘Source’. -->
  <!-- CharCount=0 -->
</FeatureParamsCharacterVariants>
```
What's the right way to fix this?

khaledhosny:
FWIW, these comments were introduced in 058850277198b198c4ed825fc4bb0c308809621f. I think somewhere the code needs to distinguish zero that means `nameId` and zero that means `NULL`. May be adding a new subclass of `NameID` that treats zero as NULL and using it in `FeatureParams*`?

anthrotype:
shouldn't they be in a range above 256?
I think we should simply always treat 0 as NULL in `NameID.xmlWrite`. That is we skip printing the comment in the XML dump when the value is 0.

PR #1152, opened 2018-01-19
otConverters: don't write XML comment if NameID value is 0 == NULL

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/1151


anthrotype:
basically we always treat a value of 0 as NULL, and the copyright nameID 0 string will never be printed in the ttx dump. I think that's ok.

PR #1154, opened 2018-01-22
Add option to parse one feature file without resolving included files

belluzj:
* [x] Have an option to not follow includes (get ast.IncludeStatement instead)
* ~~Compute `end_location` as well as location to be able to get the original text of any statement/block~~

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -591,6 +591,22 @@ class IgnoreSubstStatement(Statement):
         return "ignore sub " + ", ".join(contexts) + ";"
 
 
+class IncludeStatement(Statement):
+    def __init__(self, location, filename):
+        super(IncludeStatement, self).__init__(location)
+        self.filename = filename
+
+    def build(self):
+        # TODO: consider lazy-loading the including parser/lexer?
+        raise FeatureLibError(
+            "Building an include statement is not implemented yet. "
+            "Instead, use Parser(..., followIncludes=True) for building.",
+            self.location)
+
+    def asFea(self, indent=""):
+        return indent + "include(%s);" % self.filename
+
+
 class LanguageStatement(Statement):
     def __init__(self, location, language, include_default, required):
         Statement.__init__(self, location)
--- a/Lib/fontTools/designspaceLib/__init__.py
+++ b/Lib/fontTools/designspaceLib/__init__.py
@@ -411,7 +411,7 @@ class BaseDocWriter(object):
         axisElement.attrib['default'] = self.intOrFloat(axisObject.default)
         if axisObject.hidden:
             axisElement.attrib['hidden'] = "1"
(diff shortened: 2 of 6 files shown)


Lib/fontTools/feaLib/ast.py
```diff
@@ -591,6 +591,23 @@ def asFea(self, indent=""):
...
 
+class IncludeStatement(Statement):
+    def __init__(self, location, filename):
+        super(IncludeStatement, self).__init__(location)
+        self.filename = filename
+
+    def build(self):
+        # REVIEW: check that raising makes sense here
+        # REVIEW: is this the correct exception class?
+        raise FeatureLibError(
```
anthrotype: > is this the correct exception class?

it's the only one defined in `fontTools.feaLib.error`

anthrotype: > It does not make sense to build an include statement.
> check that raising makes sense here

I wonder if instead of raising here, we could instead lazily load the including lexer and resolve the include statements when asked to "build" them? not sure if that's possible, just a thought

belluzj: > not sure if that's possible, just a thought

It sounds a bit complex as a solution given that I'm not sure anyone needs that use-case. I think if one wants to get include statements in the AST, they're not interested in building the features, and inversely if someone wants to build the features, there is already a solution for that.

I can rephrase the exception message and/or add a TODO in the code, saying "consider lazy-loading a parser in the future"?

Just FYI, the two use-cases we have in mind that require to not follow include statements are:
* slicing up the feature file into GSFeaturePrefixes and GSFeatures for glyphsLib
* normalizing feature files

khaledhosny: I think the commit message needs a bit more explanation about the use case this is trying to support, right now I don’t even know what is being addressed here.

belluzj: Done. Is it understandable?

belluzj:
I have given up on the the `end_location` thing because it changes too much code. I'll try again later if it turns out to be of interest outside of glyphsLib. The goal of that one was to have `ast.Element.end_location` computed for all AST objects, so that it is really easy to retrieve the original code for say, a feature block: slice the text from `block.location` to `block.end_location`. I have an ugly way to do that in glyphsLib, by saying "the end location of an element is the location of its successor minus a few characters", which works enough for what I do. I wanted to implement it properly here by knowing the `end_location` from the Lexer and propagating upwards in the Parser, but it turns out to be too big a change for me to handle now.

The CI seems to be red because of a flaky designspace test, I'm going to open another PR for it.

belluzj:
OK then, I will add the flaky test fix to this PR and it will be ready to be merged.

Lib/fontTools/feaLib/ast.py
khaledhosny: Clearer now, thanks!

PR #1157, opened 2018-01-24
feaLib.parser: raise meaningful error for substitution of multiple glyphs by multiple glyphs

moyogo:
Raise a meaningful error instead of raising on `assert len(new) == 0, new`.

### DIFF
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -749,6 +749,12 @@ class Parser(object):
             return self.ast.ReverseChainSingleSubstStatement(
                 location, old_prefix, old_suffix, old, new)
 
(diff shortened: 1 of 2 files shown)


Lib/fontTools/feaLib/parser.py
```diff
@@ -749,6 +749,12 @@ def parse_substitute_(self):
...
                 location, old_prefix, old_suffix, old, new)
 
+        if len(old) > 1 and len(new) > 1:
+            raise FeatureLibError(
+                'Direct substitution of multiple glyphs by multiple glyphs '
+                'is not supported',
+                location)
+
         # GSUB lookup type 6: Chaining contextual substitution.
         assert len(new) == 0, new
```
anthrotype: then this assert is redundant, no?

PR #1169, opened 2018-02-06
[feaLib] Support params in character variant features (cvXX)

miguelsousa:
AFAICT this is working. The expected TTX was compared with what makeotf produces.

The test is failing because FEA round tripping produces equivalent instead of exactly matching statements. Not sure what's the best way to address it.

The `cvParameters` block has a series of rules —see the header of *spec8d.fea* file— which the current code does not enforce.

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -1171,7 +1171,7 @@ class CVParametersNameStatement(NameRecord):
         if self.block_name == "ParamUILabelNameID":
             item = "_{}".format(builder.cv_num_named_params_.get(self.nameID, 0))
         builder.add_cv_parameter(self.nameID)
-        self.nameID = '{}.{}{}'.format(self.nameID, self.block_name, item)
+        self.nameID = (self.nameID, self.block_name + item)
         NameRecord.build(self, builder)
 
     def asFea(self, indent=""):
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -1174,6 +1174,13 @@ class CVParametersNameStatement(NameRecord):
         self.nameID = '{}.{}{}'.format(self.nameID, self.block_name, item)
         NameRecord.build(self, builder)
 
+    def asFea(self, indent=""):
+        plat = simplify_name_attributes(self.platformID, self.platEncID,
+                                        self.langID)
+        if plat != "":
+            plat += " "
+        return "name {}\"{}\";".format(plat, self.string)
+
 
 class CharacterStatement(Statement):
     """
@@ -1189,6 +1196,9 @@ class CharacterStatement(Statement):
 
     def build(self, builder):
         builder.add_cv_character(self.character, self.tag)
+
+    def asFea(self, indent=""):
+        return "Character {:#x};".format(self.character)
 
 
 class BaseAxis(Statement):
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -1120,7 +1120,7 @@ class NameRecord(Statement):
 class FeatureNameStatement(NameRecord):
     def build(self, builder):
         NameRecord.build(self, builder)
-        builder.add_featureName(self.location, self.nameID)
+        builder.add_featureName(self.nameID)
 
     def asFea(self, indent=""):
         if self.nameID == "size":
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -1153,6 +1153,14 @@ class SizeParameters(Statement):
         return res + ";"
 
 
+class CVParametersNameStatement(NameRecord):
+    def __init__(self, location, nameID, platformID,
+                 platEncID, langID, string, block_name):
+        NameRecord.__init__(self, location, nameID, platformID,
+                            platEncID, langID, string)
+        self.block_name = block_name
+
+
 class CharacterStatement(Statement):
     """
     Statement used in cvParameters blocks of Character Variant features (cvXX).
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -230,9 +230,15 @@ class FeatureBlock(Block):
 
 
 class NestedBlock(Block):
-    def __init__(self, block_name, location=None):
+    def __init__(self, tag, block_name, location=None):
         Block.__init__(self, location)
+        self.tag = tag
         self.block_name = block_name
+
+    def build(self, builder):
+        Block.build(self, builder)
+        if self.block_name == "ParamUILabelNameID":
+            builder.add_to_cv_num_named_params(self.tag)
 
     def asFea(self, indent=""):
         res = "{}{} {{\n".format(indent, self.block_name)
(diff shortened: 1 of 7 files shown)


Lib/fontTools/feaLib/parser.py
```diff
@@ -1169,10 +1169,15 @@ def parse_feature_block_(self):
         location = self.cur_token_location_
         tag = self.expect_tag_()
         vertical = (tag in {"vkrn", "vpal", "vhal", "valt"})
+
         stylisticset = None
         if tag in ["ss%02d" % i for i in range(1, 20+1)]:
             stylisticset = tag
 
+        cv_feature = None
+        if tag in ["cv%02d" % i for i in range(1, 99+1)]:
```
anthrotype: this `["cv%02d" % i for i in range(1, 99+1)]` (as well as the similar list comprehension two lines above this) should really be class attributes as they are constant.

anthrotype: also, there should some `elif` here, because `tag` can't be all at the same time. If it's `vertical`, then it's not `stylisticset`, if it's `stylisticset` then it's not `cv_features`, etc.

Lib/fontTools/feaLib/ast.py
```diff
@@ -229,14 +229,21 @@ def asFea(self, indent=""):
...
+        self.block_name = block_name
+
+    def build(self, builder):
+        Block.build(self, builder)
+        if self.block_name == "ParamUILabelNameID":
+            builder.add_to_cv_num_named_params(self.tag)
 
     def asFea(self, indent=""):
-        res = indent + "featureNames {\n"
+        res = "{}{} {{\n".format(indent, self.block_name)
```
anthrotype: this is an instance where the new-style string formatting looks less readable than the old `%` style. I guess it comes down to taste. I never got into the `.format()` mini-language.. Anyway, i'm digressing. Whatever works.

Lib/fontTools/feaLib/builder.py
```diff
@@ -298,13 +318,20 @@ def build_name(self):
             table.names = []
         for name in self.names_:
             nameID, platformID, platEncID, langID, string = name
+            # For featureNames block, nameID is 'feature tag'
+            # For cvParameters blocks, nameID is 'feature tag.block name'
```
anthrotype: why do you have to use a dotted string as a dict key? Can't you use a tuple of strings?

Lib/fontTools/feaLib/parser.py
```diff
@@ -1434,6 +1517,13 @@ def expect_decipoint_(self):
             raise FeatureLibError("Expected an integer or floating-point number",
                                   self.cur_token_location_)
 
+    def expect_decimal_or_hexadecimal_(self):
```
anthrotype: how is this `expect_decimal_or_hexadecimal` any different from the existing `expect_number_` (apart from the error message)? Is that what you wanted to change? Then perhaps the lexer needs to be able to distinguish between decimal and hexadecimal numbers.

miguelsousa: The info message is different. Using the one for numbers didn't seem appropriate as it could lead to user confusion.

Lib/fontTools/feaLib/builder.py
miguelsousa: Perhaps. What would be the advantage, in your opinion?

Lib/fontTools/feaLib/ast.py
miguelsousa: yes, I agree. In this instance it was more to make it consistent with code in `asFea` methods of other classes.

Lib/fontTools/feaLib/parser.py
anthrotype: I see.. but then does it mean we are implicitly allowing hexadecimal digits elsewhere, where the spec only mention decimal? that could be problematic. If the spec only allows hexadecimal numbers in some places (like in cv feature) then the lexer needs to be able to distinguish between these.

Lib/fontTools/feaLib/builder.py
anthrotype: because you can (tuples of str are hashable, just like a str is) and because you avoid having to call `.split()` to get to one or the other item

Lib/fontTools/feaLib/ast.py
anthrotype: it's fine

anthrotype:
it's looking good overall. Thanks for working on this! @folengo will be happy ;)

anthrotype:
> The test is failing because FEA round tripping produces equivalent instead of exactly matching statements. Not sure what's the best way to address it.

there's a trick that @mhosken added to work around this kind of things. You have to use a special comment in your test .fea files that starts with `#test-fea2fea:`

Grep for it and see how it's been used in the rest of the test suite.

Lib/fontTools/feaLib/builder.py
miguelsousa: ok, I'll give it a try

Lib/fontTools/feaLib/parser.py
miguelsousa: good point. Should I try to address that in this PR, or sort that out in a later one?

miguelsousa:
> there's a trick that @mhosken added

Didn't know about that. Thanks for the pointer.

Lib/fontTools/feaLib/parser.py
anthrotype: later maybe, thanks

anthrotype: you can add a TODO and fix it later

miguelsousa:
@anthrotype GitHub's "Resolve in browser" feature made me believe that there were just a couple of conflicts. I resolved those and made changes to the NestedBlock calls, but the CI is failing. So there's definitely something else. The best is to back off of 95d2cf9 before making other changes.

Lib/fontTools/feaLib/parser.py
```diff
@@ -46,6 +46,8 @@ def __init__(self, featurefile, glyphNames=(), followIncludes=True,
         lexerClass = IncludingLexer if followIncludes else NonIncludingLexer
         self.lexer_ = lexerClass(featurefile)
         self.advance_lexer_(comments=True)
+        self.SS_FEATURE_TAGS = ["ss%02d" % i for i in range(1, 20+1)]
+        self.CV_FEATURE_TAGS = ["cv%02d" % i for i in range(1, 99+1)]
```
anthrotype: I actually meant class attributes, as they are constant. Not instance attributes initialised in the constructor.

anthrotype:
Thanks, I’ll have a look once I get to the office

Lib/fontTools/feaLib/parser.py
miguelsousa: I originally placed them just below line #18 `ast = ast` and it failed. I must have missed something when I tried that.

miguelsousa:
I've just force-pushed the branch to nuke the last change. Going to leave it alone now.

miguelsousa:
Thanks Cosimo.
BTW, can you have a look at the OT spec https://docs.microsoft.com/en-us/typography/opentype/spec/features_ae#cv01-cv99 ?
I get the impression that a lot of the rules mentioned by the FEA spec wrt `cvParameters` blocks are limitations of the makeotf implementation, but some might not.

anthrotype:
> I get the impression that a lot of the rules mentioned by the FEA spec wrt cvParameters blocks stem from limitations in the makeotf implementation

which ones are you referring to?

anthrotype:
Do you mean this?

https://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html

> If any ParamUILabelNameID entries are specified, the number of ParamUILabelNameID entries must match the number of variants for each base glyph

https://docs.microsoft.com/en-us/typography/opentype/spec/features_ae#cv01-cv99

> The value of numNamedParameters should not exceed the number of alternate glyphs in lookups associated with the feature; note, however, that the number of GIDs in the returned array for a GSUB type 3 lookup should not be assumed to be equal to numNamedParameters: numNamedParameters should not be more than the number of GIDs in the array, but it may be less. If numNamedParameters is zero, then no ‘name’ table strings are associated with feature parameters.

anthrotype:
ok I think this is good to go for now. I'd like to tag another release today. We can change it later if we need.

anthrotype:
And thanks again for working on this, Miguel :+1:

Issue #1171, opened 2018-02-07
Question about morx: xml vs binary


olivierberten:
@brawer First of all, thanks a lot for the morx files parsing. I have a question, though: unlike most other tables, the xml structure for morx is pretty far away from the binary structure which means quite some informations get missing in the process. When round-tripping such fonts, the new morx table gets much smaller while xml is identical so it's pretty hard to figure out what has changed... 
That said, the resulting fonts seems to work as well as the original one so I guess ftxenhancer's implementation is just way suboptimal but I'm far from having thoroughly tested it...

behdad:
Can you share an example?

olivierberten:
http://entiumgay.sourceforge.net/

    Listing table info for "Entiumgay.ttf":
        tag     checksum   length   offset
        ----  ----------  -------  -------
        FFTM  0x5389B45E       28   106924
        GDEF  0x00E40024       40   106952
        GPOS  0xD832DF16       48   106992
        GSUB  0xE4EEDB67      242   107040
        OS/2  0x96686553       96   106828
        cmap  0xDDA201F5      618    71152
        cvt   0x01860747       26    55376
        feat  0x359E0A95       68   107284
        fpgm  0x06599C37      371     1056
        gasp  0x00170008       16   107352
        glyf  0x78A139D0    53468     1908
        head  0xFF378BAE       54    55404
        hhea  0x0C830678       36    55460
        hmtx  0x98032D7C      722      332
        loca  0x8A42C21C      364     1544
        maxp  0x02D6041D       32    71772
        morx  0xF967890C    35024    71804
        name  0xCCE48E4C    15182    55968
        post  0x4B535C4E      471    55496
        prep  0xD4C7B886      114     1428


    Listing table info for "Entiumgay#1.ttf":
        tag     checksum   length   offset
        ----  ----------  -------  -------
        FFTM  0x5389B45E       28    71700
        GDEF  0x002700DA       34    71728
        GPOS  0xD829DEFE       38    71764
        GSUB  0xE45BDADE      168    71804
        OS/2  0x96686553       96      456
        cmap  0xDDA201F5      618     1276
        cvt   0x01860747       26     2384
        feat  0x359E0A95       68    71972
        fpgm  0x06599C37      371     1896
        gasp  0x00170008       16    71684
        glyf  0xAEF11EA7    53308     2776
        head  0x0D3A621D       54      332
        hhea  0x0C830678       36      388
        hmtx  0x98032D7C      722      552
[...]

PR #1174, opened 2018-02-07
[morx] WIP: Implement glyph insertion

behdad:
@brawer Any intention to finish this? Even if not, maybe better merge than leave?  Came to my attention in https://github.com/fonttools/fonttools/issues/1171


brawer:
Done in https://github.com/fonttools/fonttools/pull/1323.

PR #1175, opened 2018-02-12
Add properties to designspace documents

belluzj:
For use in `glyphsLib`, I would like to add 3 things to `designspaceLib`. This PR is first to get approval from the maintainers on the intention, and if it's okay I will add tests and update the documentation/spec.

1. A `lib` element for the whole document and one `lib` element per instance. I use it for the whole document to store font-wide data that (in a multi-master setup) does not really make sense in one UFO or would need to be repeated in all of them. I need it for instances because that is the only way to store more information about an instance, as opposed to sources where I can put stuff in the associated UFO's lib. Was discussed here: https://github.com/LettError/designSpaceDocument/issues/19

2. A `font` attribute for the `SourceDescriptor`, which can be None or an instance of `defcon.Font`. I use that field to do things all in memory: when creating a designspace and some UFOs, I return an instance of `DesignSpaceDocument` with attached `defcon.Font` in each `source.font`, without the need to write any of those to the disk. The user can then decide where to write it, or even to never write it anywhere.

3. A `filename` attribute for `DesignSpaceDocument`. The idea is to provide a suggested name for the document when it is returned from `glyphsLib`, and let the user decide whether to use that name when writing the document somewhere. It's different from the path in the sense that it's not a full path, just a suggested filename. I think this last attribute does not sound as useful as the two others now that I describe it, so maybe it's not a good idea.

What do you think? @LettError 
This PR can also be the occasion for some API changes @behdad


LettError:
1:
I'm still curious to see what type of fields, data, variables you need to store. Rather than just dump it all as a plist, there might be data in it that is useful elsewhere.

2:
`font` attribute for the `SourceDescriptor`: rather than rely blindly on defcon, it should be possible to specify the fontClass somewhere, which makes `DesignSpaceDocument` a better place to handle loading the fonts. It could have a `loadFonts` method that loads all the fonts at all the paths. See [ufoProcessor.DesignSpaceProcessor.loadFonts](https://github.com/LettError/ufoProcessor/blob/8954605b061fdb57ca8e0467a6dd1d42f3862fdf/Lib/ufoProcessor/__init__.py#L398) for an example of specifying fontClass and loading the fonts. Though ufoProcessor stores the fonts in a dict - I don't mind if they're added to the sourceDescriptor.

3:
`filename` attribute: not sure what this is for - for a variable font that is generated from a designspace?

belluzj:
1:
I'm still working on it, so it's not complete, but here are some examples.

```python
>>> from glyphsLib import to_designspace, classes
>>> f = classes.GSFont('tests/downloaded/montserrat/sources/Montserrat.glyphs')
>>> doc = to_designspace(f, minimize_glyphs_diffs=True)
>>> doc.lib
{'GSDimensionPlugin.Dimensions': {'1C84A7F3-C40A-4DC1-B8B4-102D183C348F': {'HV': 20}, '5DA6E103-6A94-47F2-987D-4952DB8EA68E': {'HH': '194', 'HV': '236', 'OH': '191', 'OV': '238', 'nV': '226', 'nd': '190', 'oH': '176', 'oV': '229', 'tH': '165'}, '708134FE-A11E-43C9-84F0-594DA15B6BD1': {'HH': '99', 'HV': '114'}, 'D46AABC6-AB7D-4C0B-AE2B-3079AD121F41': {'HV': 232}, 'UUID0': {'HH': 20, 'HV': 20, 'OH': 20, 'OV': 20, 'nV': 20, 'nd': 15, 'oH': 20, 'oV': 20, 'tH': 20}, 'UUID1': {'HH': 150, 'HV': 190, 'OH': 165, 'OV': 200}}}
# This is custom data from a Glyphs plugin
>>> doc.instances[1].lib
{'com.schriftgestaltung.export': True, 'com.schriftgestaltung.weight': 'ExtraLight', 'com.schriftgestaltung.width': 'Medium (normal)', 'com.schriftgestaltung.intanceInterpolations': {'UUID0': 0.855556, '708134FE-A11E-43C9-84F0-594DA15B6BD1': 0.144444}, 'com.schriftgestaltung.manualInterpolation': False}
# This is instance data that is specific to Glyph's model of instances
# export=True: Glyphs can define instances that won't be generated when building the TTFs (export=False)
# weight, width: Glyphs has several names for the same OS2 value, so I store which name was chosen by the user
# instanceInterpolation: Glyphs can specify the "contribution" of each master to the instance (usually computed automatically but could be set by hand if I understand correctly, and then manualInterpolation=True)
```

Also, like @anthrotype said in LettError/designSpaceDocument#19, Glyphs allows users to associate arbitrary data to instances using `customParameters`, so in order to round-trip we need a place to store arbitrary data. Once this is possible using `designspaceLib`, I will try to change `fontmake` to use only the designspace document to build instances (and stop passing instance data directly from `glyphsLib` to `fontmake`) and then it will be clearer which `lib` fields are worthy of becoming regular, specified fields.

Looking at the example with the key "GSDimensionPlugin.Dimensions", maybe there should be a naming scheme like `com.schriftgestaltung...` to avoid future name clashes between various

2:
I mentioned `defcon.Font` because that is what I use but the code in `designspaceLib` does not mention defcon anymore, so this `font` field could as well hold a `TTFont`. My intent is only to have the field available, and specify a general purpose like "hold a Python object that represents the source".

In `glyphsLib` there is an equivalent of `loadFonts`, and since it is a simple operation, I think it can be up to each user of `designspaceLib` to implement their own. Basically it should go like: check `source.font` for a usable object; else load from `source.path`/`source.filename`.

3:
The general idea would be that a tool that returns a `DesignSpaceDocument` object in-memory can give a hint of how that document could be named if it were to be written to the disk.

In particular in `glyphsLib`, there was code to generate a "nice" name for writing the document to disk, and when I changed the API to return objects in memory by default, I did not want to lose this "nice name", so I wrote it in this field.

If this third thing does not make sense, I'm fine to drop it, I'll find another way :)

behdad:
These sound good to me, specially the font part.

belluzj:
This is ready for review. I have changed the line endings of two rst files to unix (\n) so here is the diff without whitespace changes:

https://github.com/fonttools/fonttools/pull/1175/files?w=1

anthrotype:
sorry I haven't had time to review this yet. I will do probably next week.
Would be nice if @LettError could also have a look.

LettError:
looks good, but have not tested.

anthrotype:
> maybe there should be a naming scheme like com.schriftgestaltung... to avoid future name clashes between various software using the lib.

Definitely, yes. We should document that this new `lib` elements in designspace shold follow a similar "reverse domain naming scheme" that the UFO lib follows.

About the suggested `filename` thing, I'm not sure about its general utility. It if it's only useful in the context of glyphsLib, then this may be a case of using the new `lib` element and define a private glyphsLib key to store this.

moyogo:
If we use keys with `com.schriftgestaltung...` we need to make sure they are approved and used by @schriftgestalt, otherwise we can use `com.googlei18n.glyphsLib...`.

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -874,8 +937,14 @@ def _readSingleInstanceElement(self, instanceElement, makeGlyphs=True, makeKerni
             self.readGlyphElement(glyphElement, instanceObject)
         for infoElement in instanceElement.findall("info"):
             self.readInfoElement(infoElement, instanceObject)
+        for libElement in instanceElement.findall('lib'):
+            self.readLibElement(libElement, instanceObject)
         self.documentObject.instances.append(instanceObject)
 
+    def readLibElement(self, libElement, instanceObject):
+        """Read the lib element for the given instance."""
+        instanceObject.lib = from_plist(libElement.getchildren()[0])
```
anthrotype: `getchildren` is deprecated since 3.2
https://docs.python.org/3.6/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.getchildren

anthrotype: you can simply get the subelements as if element was a list, without the extra call to deprecated getchildren (here and elsewhere in the code)

anthrotype:
> we need to make sure they are approved

sure. Georg can confirm he's OK with it (but that's a glyphsLib issue)

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -5,6 +5,7 @@
 import logging
 import os
 import posixpath
+import plistlib
 import xml.etree.ElementTree as ET
```
anthrotype: we need to try and use the cElementTree module on python2.7 (on python3 it's the default, when available).
Do this instead:

```python
try:
	import xml.etree.cElementTree as ET
except ImportError:
	import xml.etree.ElementTree as ET
```

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -540,6 +597,11 @@ def _addSource(self, sourceObject):
         sourceElement.append(locationElement)
         self.root.findall('.sources')[0].append(sourceElement)
 
+    def _addLib(self, dict):
+        libElement = ET.Element('lib')
+        libElement.append(to_plist(dict))
```
anthrotype: even if it's just in a nested method scope like in here, I like to avoid shadowing built-in types.
makes it harder to read as the syntax highlighting thinks it's a built-in dict type, whereas this is actually a local variable.

schriftgestalt:
I‘m have a closer look ASAP but from reading the thread it seems ok to use ‘com.schriftgestalt... as prefix. The ‘GSDimension...’ stuff might be best put into ‘com.schriftgestaltung.userData’ (or similar) key.

belluzj:
Thanks for the review! I will open a new MR soon to fix all the comments

anthrotype:
@belluzj I just pushed these two fixes to master (696edb6b6db4615bb325f681cfc99ebd8673db19 and 8ca22ab002764df6aae0e381fbc26180fef1f3f1). I'd like to release before the scheduled pyup.io bot starts (around 14:00 GMT). We can always changes things later, as we marked this API as unstable in the documentation anyway.

belluzj:
Ok

schriftgestalt:
I just had a look. In .ufos saved from Glyphs, the `GSDimensionPlugin.Dimensions` stuff is stored in a `com.schriftgestaltung.font.userData` key in the font lib. So all font and master related things can be stored in the .ufo, but instance data needs a place in the designSpace file.

anthrotype:
> instance data needs a place in the designSpace file

yep, that's what this PR was about (adding a `lib` element to designspace)

PR #1180, opened 2018-02-14
[merge] Support merging LangSysRecords

moyogo:


behdad:
Nice. Thanks.

Though, thinking more, it needs further tweaking. If some fonts have LangSys ABC but others don't, we should merge the DefaultLangSys of those other fonts into the merged LangSys ABC. Or assert that all fonts have the same set of LangSyses.

PR #1186, opened 2018-02-21
feaLib: add IncludedFeaNotFound error

anthrotype:
subclass of FeatureLibError, only raised if IOError.errno == ENOENT (i.e. FileNotFoundError)

https://github.com/googlei18n/fontmake/issues/157#issuecomment-367380471

Will be useful in ufo2ft to detect when an included feature file doesn't exist and print a nicer error message explaining that includes must be relative to the UFO itself, and not relative to the embedded features.fea file.

https://github.com/unified-font-object/ufo-spec/issues/55

### DIFF
--- a/Lib/fontTools/feaLib/lexer.py
+++ b/Lib/fontTools/feaLib/lexer.py
@@ -1,7 +1,7 @@
 from __future__ import print_function, division, absolute_import
 from __future__ import unicode_literals
 from fontTools.misc.py23 import *
-from fontTools.feaLib.error import FeatureLibError
+from fontTools.feaLib.error import FeatureLibError, IncludedFeaNotFound
 import re
 import os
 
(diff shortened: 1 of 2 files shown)


Lib/fontTools/feaLib/lexer.py
```diff
@@ -231,7 +231,11 @@ def make_lexer_(file_or_path, location=None):
             try:
                 fileobj = open(filename, "r", encoding="utf-8")
             except IOError as err:
-                raise FeatureLibError(str(err), location)
+                # FileNotFoundError does not exist on Python < 3.3
+                import errno
+                if err.errno == errno.ENOENT:
```
brawer: Does this work on Windows? (Might be good to add a test case that triggers this error, then we’ll know for sure).

anthrotype: yes, it works on Windows too.

we do have a test that asserts that a FeatureLibError (superclass of IncludedFeaNotFound) is raised if we `include blah` (non existent file), and the test passed on Windows, and I can see the line covered in the coverage.

I can make the test more specific if you like.

anthrotype: it works on windows too, see
https://ci.appveyor.com/project/fonttools/fonttools/build/1.0.1535

Issue #1177, opened 2018-02-13
[cffLib] Question about Encoding data specified by TopDict.Encoding


derwind:
I am confused in parsing Encoding data in CFF.

* The problem: I converted a font by update_encoding.py and confirmed the result by inspect_encoding.py, then unexpected result was shown, where these scripts are as follows:

[update_encoding.py]
```python
font = TTFont(sys.argv[1])
topDict = font["CFF "].cff.topDictIndex[0]

encoding = [".notdef"] * 256
encoding[0x20] = "space"
encoding[0x41] = "A"
encoding[0x42] = "B"
encoding[0x43] = "C"

topDict.Encoding = encoding
font.save("out.otf")
```

[inspect_encoding.py]
```python
font = TTFont(sys.argv[1])
topDict = font["CFF "].cff.topDictIndex[0]
print(topDict.Encoding)
```

* Result:
I used Source Sans Pro as a test font and got the result:
```python
['.notdef', '.notdef', 'space', 'A', 'B', 'C', 'D', 'E', 'F', 'G', ...]
```
I expect that 'space' appears at 32nd position in the list because its unicode is 0x20, and it is also strange that 'D', 'E', ... appear because I don't set them to the encoding data.

* My environment:
  * macOS High Sierra 10.13.3
  * Python 2.7.14
  * fontTools.\_\_version\_\_: '3.22.1.dev0'

* Further detail:
I investigated source codes and found some strange behavior.
I inserted `print(file.tell())` just before calling parseEncoding1 in EncodingConverter.read in cffLib/\_\_init\_\_.py and also inserted same code at the top of parseEncoding1.

My log said:
```
27996
27995
```
This result seems to me that the file pointer rewinded for some reason when entering parseEncoding1.
I don't know whether this problem depends on my environment.

behdad:
What are you trying to do?

derwind:
In short I tried to check the implementation of my OTF parser by using fontTools.

To understand the structure of OpenType fonts deeply, I have developed my OTF parser.
I usually check whether my implementations are correct by ttx and fontTools.
The other day I checked the implementaion of my parser's functionality of parsing Encoding data in CFF.

First, I compared the log of my parser and that of ttx. It was as expected.
Next, I compared the log of my parser and that of fontTools, i.e. show_encoding.py mentioned in my previous comment, but it seemed strange.

I investigated the strange behavior and found that the Encoding data changed by the snippet below:

```
font = TTFont(sys.argv[1])
[...]

PR #1187, opened 2018-02-21
[cffLib] save file position against lazy-load of charset

derwind:
This PR is related to #1177.

As is discussed, parseEncoding0 may receive unexpected file position due to lazy-load of 'charset'.
So I save file position before calling parseEncoding0 and restore the position at the beginning of parseEncoding0.

Similar modification is applied to parseCharset0 because of symmetry. Actually it may not be needed since 'strings' which has possibility of lazy-load seems to be always loaded before calling parseCharset0.

### DIFF
--- a/Lib/fontTools/cffLib/__init__.py
+++ b/Lib/fontTools/cffLib/__init__.py
@@ -1180,6 +1180,7 @@ class PrivateDictConverter(TableConverter):
 	def read(self, parent, value):
 		size, offset = value
 		file = parent.file
+		pos = file.tell()
 		isCFF2 = parent._isCFF2
 		try:
 			vstore = parent.vstore
@@ -1191,6 +1192,7 @@ class PrivateDictConverter(TableConverter):
 		data = file.read(size)
 		assert len(data) == size
 		priv.decompile(data)
+		file.seek(pos)
 		return priv
 
 	def write(self, parent, value):
@@ -1204,9 +1206,12 @@ class SubrsConverter(TableConverter):
 
 	def read(self, parent, value):
 		file = parent.file
+		pos = file.tell()
 		isCFF2 = parent._isCFF2
 		file.seek(parent.offset + value)  # Offset(self)
-		return SubrsIndex(file, isCFF2=isCFF2)
+		subrIndex =  SubrsIndex(file, isCFF2=isCFF2)
+		file.seek(pos)
+		return subrIndex
 
 	def write(self, parent, value):
 		return 0  # dummy value
@@ -1216,6 +1221,7 @@ class CharStringsConverter(TableConverter):
 
 	def read(self, parent, value):
 		file = parent.file
+		pos = file.tell()
 		isCFF2 = parent._isCFF2
 		charset = parent.charset
 		globalSubrs = parent.GlobalSubrs
@@ -1232,6 +1238,7 @@ class CharStringsConverter(TableConverter):
 		file.seek(value)  # Offset(0)
 		charStrings = CharStrings(
 			file, charset, globalSubrs, private, fdSelect, fdArray, isCFF2=isCFF2)
+		file.seek(pos)
 		return charStrings
 
 	def write(self, parent, value):
@@ -1264,18 +1271,19 @@ class CharsetConverter(object):
 		if value > 2:
 			numGlyphs = parent.numGlyphs
 			file = parent.file
+			pos = file.tell()
 			file.seek(value)
 			log.log(DEBUG, "loading charset at %s", value)
 			format = readCard8(file)
-			pos = file.tell()
 			if format == 0:
-				charset = parseCharset0(numGlyphs, file, parent.strings, isCID, pos)
+				charset = parseCharset0(numGlyphs, file, parent.strings, isCID)
 			elif format == 1 or format == 2:
-				charset = parseCharset(numGlyphs, file, parent.strings, isCID, format, pos)
+				charset = parseCharset(numGlyphs, file, parent.strings, isCID, format)
 			else:
 				raise NotImplementedError
 			assert len(charset) == numGlyphs
 			log.log(DEBUG, "    charset end at %s", file.tell())
+			file.seek(pos)
 		else:  # offset == 0 -> no charset data.
 			if isCID or "CharStrings" not in parent.rawDict:
 				# We get here only when processing fontDicts from the FDArray of
@@ -1408,8 +1416,7 @@ def packCharset(charset, isCID, strings):
 	return bytesjoin(data)
 
 
-def parseCharset0(numGlyphs, file, strings, isCID, pos):
-	file.seek(pos)
+def parseCharset0(numGlyphs, file, strings, isCID):
 	charset = [".notdef"]
 	if isCID:
 		for i in range(numGlyphs - 1):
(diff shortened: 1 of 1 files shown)


anthrotype:
Thanks for the PR!

That's not exactly what I had in mind though.

The idea was to

> make such behind-the-scene lazy loads always seek the file back where it was...

So, in each of the `*Converter` classes's `read` methods, before they actually do any `file.seek` or `file.read` (either themselves or some other functions which they call), we would `tell` the file's current position, save to a local variable `pos`, then at the end of a converter's `read` method, seek back to the original saved position, so that the callers can call a converter's `read` method in any order and it's guaranteed to leave the file in the position that it found it.

You don't need to change the signatures of parseEncoding or parseCharset to add an extra `pos` argument. Simply do a file.seek(pos) at the end of the `*Converter.read` methods.

derwind:
Thank you for your comments!

I understand and I will update codes again and test them. Indeed it is better if the signatures are unchanged.

derwind:
I modified my codes again. I confirmed the result of reading TopDict.Encoding was as expected. Also I confirmed ttx command dumped same XMLs as those before creating PR by using SourceSansPro-Regular.otf and SourceHanSans-Regular.otf.

anthrotype:
Thanks!

I have another idea (sorry if I keep you editing the code).

What if in the base class for all the Converters we define a `read` method and another one called `_read` which subclasses should implement (which does the actual reading). The `read` method of the superclass calls the _read method, but before and after ensures that the file position doesn't change.

```python
def read(self, ...):
    pos = file.tell()
    try:
        self._read(...)
    finally:
        file.seek(pos)

def _read(self, ...):
    raise NotImplementedError

```

derwind:
@anthrotype Thank you for your advice again!
It is a little difficult for me to decide which is elegant but I implemented your advice.

On of the merits of previous implementation is that one can easily understand the intention when reading `*Converter.read`. But one should pay attention to possible lazy-loads when implementing new Converter.
On the other hand one of the merits of this implementation is that one can easily introduce new Converter without being conscious of lazy-load. But one may feel it difficult to notice the reason why `_read` methods were introduced when he or she sees them for the first time.

This seems a trade-off.

Lib/fontTools/cffLib/__init__.py
```diff
@@ -1660,7 +1673,7 @@ def xmlRead(self, name, attrs, content, parent):
 
 class VarStoreConverter(SimpleConverter):
```
anthrotype: you forgot this one?

Lib/fontTools/cffLib/__init__.py
```diff
@@ -1027,9 +1027,22 @@ def buildConverters(table):
 	return d
 
 
-class SimpleConverter(object):
+class BaseConverter(object):
```
anthrotype: SimpleCoverter is really the BaseConverter, as it's only used as base class and never used as such. so you can define read and _read on SimpleConverter itself.

anthrotype:
another merit of this is that `finally` branch is guaranteed to be called no matter what happens in the `try` block, whereas in the previous one if an exception occurred the file would not rewind itself.

Lib/fontTools/cffLib/__init__.py
derwind: According to your next comment I will update SimpleConverter itself.

Lib/fontTools/cffLib/__init__.py
```diff
@@ -1027,20 +1027,16 @@ def buildConverters(table):
 	return d
 
 
-class BaseConverter(object):
+class SimpleConverter(object):
 
 	def read(self, parent, value):
+		if not hasattr(parent, "file"):
```
anthrotype: good catch

Lib/fontTools/cffLib/__init__.py
```diff
@@ -1027,20 +1027,16 @@ def buildConverters(table):
...
+			return self._read(parent, value)
 		file = parent.file
 		pos = file.tell()
-		try:
```
anthrotype: why did you remove the try... finally... block?

derwind: Now SimpleConverter is the base class from which other Converters are derived, and the default meaningful _read method always exists and it simply returns 'value'. So I removed try finally block.

anthrotype: No, you need to put it back. The function of the finally is to ensure the statement after it will be called at the end whether the block inside the try completed successfully or it raised an exception.

derwind: Uh... try block should exist with considering overridden _read throw any exception ... sorry I will restore the try finally block.

anthrotype: yep. thanks

Lib/fontTools/cffLib/__init__.py
derwind: Thanks!

anthrotype:
one last optimization (but we can do it another time) could be to identify which of the Converter classes take a `parent` argument with a `file` attribute, and which ones don't; and only for the former do the wrapped _read; the latter don't manipulate files so they can implement `read` and be done with it (avoiding the second method call).
Because I imagine that this `parent` thing either always has a `file` or `never`, and that specific converters are always used with the same type of `parent` objects.
Anyway, thanks again. Merging now

derwind:
@anthrotype Thank you for your patientlly reviewing my codes many times!

anthrotype:
> I imagine that this parent thing either always has a file or never,

that's not true, actually. Both Index and BaseDict instances can be either of two state: they may or may not have a `file` attribute, depending on whether they have been initialized from a binary font file or whether they started empty and populated later. So the code is fine like this.

behdad:
I would have guessed we put the seek logic in `__getattr__`. This works too though.

anthrotype:
@behdad damn, you're right! Why didn't I think of that... :duck:

derwind:
It's interesting! I implemented @behdad's idea as below and tested it. It worked fine!

```
$ git diff 4fec016862324092bf7d4643eb694903ffc114ac Lib/fontTools/cffLib/__init__.py
diff --git a/Lib/fontTools/cffLib/__init__.py b/Lib/fontTools/cffLib/__init__.py
index a11b8f6a..b7c4f584 100644
--- a/Lib/fontTools/cffLib/__init__.py
+++ b/Lib/fontTools/cffLib/__init__.py
```
```python
@@ -2234,7 +2234,10 @@ class BaseDict(object):
                if value is None:
                        raise AttributeError(name)
                conv = self.converters[name]
+               pos = self.file.tell() if hasattr(self, "file") else None
                value = conv.read(self, value)
+               if pos is not None:
+                       self.file.seek(pos)
                setattr(self, name, value)
                return value
```

If comparing `__getattr__` to 'a parent' and `*Converter.read` to 'his or her child', previous implementation seems as if the child who cluttered his or her toys put the toys away by his or herself, and newer implementation seems as if the child left his or her toys lying about and the parent put the toys away on behalf of the child.

anthrotype:
Haha! Nice metaphor :)
If that is the only place where the read methods are called from, then it makes more sense to have it that way (i’d still wrap the call to read in a try finally block).
I’ll merge if you send a pull request, thanks

derwind:
> If that is the only place where the read methods are called from

Surely if 'the spoiled child' is only one, 'the parent' can easily take care of 'the child'. But if there are many such children, 'the parent' may fail to take care.
And now I am worried about 'read' call in `FontDictCompiler.__init__` and I don't know when this path is executed and whether conv has 'file' or not.
```python
				default = dictObj.defaults.get(name)
				if default is not None:
					conv = dictObj.converters[name]
					default = conv.read(dictObj, default)
```

Considering this case, previously merged patch seems to me safer because all 'children' are always well-behaved :)

Issue #1188, opened 2018-02-21
[ttFont] xmlWriter import error with saveXML split tables = True requests


chrissimpkins:
Attempt to dump split OT tables with ttx, executed in Python 3 venv with PyPI release of fontTools:

```
$ ttx -s -d regular-ttx Hack-Regular.ttf
Dumping "Hack-Regular.ttf" to "regular-ttx/Hack-Regular.ttx"...
ERROR: Unhandled exception has occurred
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttx.py", line 389, in main
    process(jobs, options)
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttx.py", line 363, in process
    action(input, output, options)
  File "/usr/local/lib/python3.6/site-packages/fontTools/misc/loggingTools.py", line 372, in wrapper
    return func(*args, **kwds)
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttx.py", line 260, in ttDump
    newlinestr=options.newlinestr)
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/ttFont.py", line 221, in saveXML
    self._saveXML(writer, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/fontTools/ttLib/ttFont.py", line 264, in _saveXML
    tableWriter = xmlWriter.XMLWriter(tablePath,
NameError: name 'xmlWriter' is not defined
```

PR for fix incoming...

chrissimpkins:
PR #1189

PR #1192, opened 2018-02-22
[ttx_test] Test for split tables (-s) option

miguelsousa:
#1188

### DIFF
--- a/Tests/ttx/ttx_test.py
+++ b/Tests/ttx/ttx_test.py
@@ -1,5 +1,6 @@
 from __future__ import print_function, division, absolute_import
 from fontTools.misc.py23 import *
+from fontTools.misc.testTools import parseXML
 from fontTools import ttx
 import getopt
 import os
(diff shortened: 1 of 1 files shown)


Tests/ttx/ttx_test.py
```diff
@@ -40,6 +41,13 @@ def temp_font(self, font_path, file_name):
         shutil.copy2(font_path, temppath)
         return temppath
 
+    @staticmethod
+    def read_file(file_path):
```
anthrotype: I don't see much the point of this method, you only use it once..
Anyway, keep it, doesn't matter. You still need to pass `encoding="utf-8"` (on Windows it defaults to some cp1252 if I remember well).
And I prefer the `with open(...) as f:` context manager to calling `f.close()`

miguelsousa: I made it a method because `ttx.py` only has a little over 40% coverage right now, and I expect it will be useful as more tests are added.
I'll make the changes you suggest.

PR #1193, opened 2018-02-22
[cffLib_test] Test setting topDict.Encoding

miguelsousa:
#1177 

Sorry for the boilerplate again.


anthrotype:
there's a fontTools.misc.testTools. Just define a class in there which subclasses `unittest.TestCase` (find a good name for it, I don't know right now), set all the boilerplate methods in it but don't add any `test_` methods of course. Then in your test modules, import it and use it as a base class for your actual test case classes.

anthrotype:
the assertRaisesRegex thing we want it for all our unittest-based tests. The other helper methods, only for some of them that actually use them.
So I'd say, in testTools we `from unittest import TestCase as _TestCase` and then define a generic `class TestCase(_TestCase)`  that has the renamed assertRaisesRegex method. And we change all the tests to subclass that one instead of the built-in so we don't need to repeat that trick.

For this ttx-roundtripping-sort-of-tests, you can define yet another subclass of our TestCase class (always inside testTools) that defines those, and import and use it when you need it, like in here.

anthrotype:
the ci test failures are completely unrelated
https://travis-ci.org/fonttools/fonttools/jobs/344901864#L721

`unicode_literals` is really weird.. why is this triggering only now with your PR whereas it passed on master?

the fix (only old versions of python 2.7.x) would be to ensure that the `struct` format string is a native `str` (`bytes` on python2, `unicode` on python3). Either by removing `unicode_literals` (and then check all the string literals and add `u""` as necessary), or by calling our `tostr(">HHH")` which seems silly but does the right thing.

miguelsousa:
Got it. Working on the changes.

miguelsousa:
> why is this triggering only now with your PR whereas it passed on master?

It must be because I'm special 😁

The changes are in. I opted for the easiest fix on the name table test.

PR #1197, opened 2018-02-25
[Snippets] SHA1 hash checksum list generation and testing script for font binaries and select OpenType table data

chrissimpkins:
We needed a concise (i.e., not ttx dumps + text diffs) way to describe, maintain the state of, and compare compiled OpenType tables across builds. This was driven by the need to provide an approach for Linux packagers to confirm that their font compiles from source do not differ meaningfully (read that as affecting renders) from upstream builds as their build dependency versions across our entire build toolchain (and dependency of dependency of dependency chain...) diverge from our own by nature of the way that most Linux distros package and release software.

This script represents my attempt at a solution to this issue since direct SHA1 hashes of the binary files will always differ when build times differ, as will diffs of full ttx dumps from the font files, unless we force the downstream projects to compile with environment variables that fix these time-dependent values to our own.  We could dump select tables with ttx and use text file diffs but maintaining these data upstream so that downstream projects have access to them requires files that are too large for our project and we only need to evaluate the XML when relevant differences occur.  Hashes of the XML dump from select, relevant OT tables appeared to be a reasonable solution that allows us to maintain the state of these tables in a very concise format as well as permit simple broadly supported checksum comparisons between builds.  I made this script a bit more broad to address other needs but this is how we intend to use it.

This script supports and demonstrates an approach to the following with Python 3 and fontTools library dependencies alone:

- generation of SHA1 hashes from fonts
- generation of ttx style XML dumps of OT tables with `ttLib.TTFont.saveXML` using include + exclude table options to achieve between one to full OT table data dumps from binaries
- generation of SHA1 hashes from the *OpenType XML text data derived from fonts with the `saveXML` method* including:
      - single OT table hash with `--include` option
      - multiple OT tables hash with `--include` or `--exclude` options
      - all OT tables hash by default
- generation of valid checksum files for any of the above hashes that can be used for testing with this script or with other platform-specific tools such as `shasum` (macOS) or `sha1sum` (Linux)
- writes of checksum files to disk (in working directory as `checksum.txt` by default) with optional writes to the stdout stream with `--stdout` flag
- cross platform support for automated testing of SHA1 hashes defined in one or more valid SHA1 checksum files created with this tool or other tools, eliminating the need for different platform dependent executables to perform these checks.  This is an attempt to replicate the `-c` option used in other checksum tools that will work with one approach across all platforms.

The source header contains all available options. I've performed local testing and can verify that the SHA1 hashes are valid for both font binaries and XML OT table dumps, it works interchangeably with `shasum` on macOS in terms of both tools using the other tool's checksum file to validate font and XML files, and that my machine is still running after execution of the script numerous times.  There is no comprehensive test suite at the moment, and this should be considered untested and intended for education as a Snippet if felt to be appropriate in that directory.  If not, I will maintain this in my downstream repository for anyone who might find it to be of use.

### DIFF
--- a/Snippets/checksum.py
+++ b/Snippets/checksum.py
@@ -16,6 +16,9 @@
 #
 # Usage: checksum.py (options) [file path 1]...[file path n]
 #
+#   `file path` should be defined as a path to a font file for all use cases except with use of -c/--check.
+#   With the -c/--check option, use one or more file paths to checksum files
+#
 # Options:
 #   -h, --help          Help
 #   -t, --ttx           Calculate SHA1 hash values from ttx dump of XML (default = font binary)
--- a/Snippets/checksum.py
+++ b/Snippets/checksum.py
@@ -71,7 +71,7 @@ def check_checksum(filepaths):
     check_failed = False
     for path in filepaths:
         if not os.path.exists(path):
-            sys.stderr.write("[checksum.py] ERROR: " + filepath + " is not a valid filepath" + os.linesep)
+            sys.stderr.write("[checksum.py] ERROR: " + path + " is not a valid filepath" + os.linesep)
             sys.exit(1)
 
         with open(path, mode='r') as file:
--- a/Snippets/checksum.py
+++ b/Snippets/checksum.py
@@ -114,12 +114,13 @@ def _read_binary(filepath):
 
 
 if __name__ == '__main__':
-    parser = argparse.ArgumentParser(prog="checksum.py")
+    parser = argparse.ArgumentParser(prog="checksum.py", description="A SHA1 hash checksum list generator and checksum testing script")
     parser.add_argument("-t", "--ttx", help="Calculate from ttx file", action="store_true")
     parser.add_argument("-s", "--stdout", help="Write output to stdout stream", action="store_true")
     parser.add_argument("-n", "--noclean", help="Do not discard *.ttx files used to calculate SHA1 hashes", action="store_true")
     parser.add_argument("-c", "--check", help="Verify checksum values vs. files", action="store_true")
-    parser.add_argument("filepaths", nargs="+", help="One or more file paths to font binary files")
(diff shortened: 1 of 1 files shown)


behdad:
The usage info in the file header is redundant given that --help can generate the same.

chrissimpkins:
Will remove it.  Leave license or remove that too?  It looks like other snippets do not include a license.

chrissimpkins:
removed all header text including license and usage information in https://github.com/fonttools/fonttools/pull/1197/commits/d8e5f5cc37ed9bd6a3228a12b25e3071d3d7ef72

fixed incorrect variable name in stderr message in https://github.com/fonttools/fonttools/pull/1197/commits/db38156b9133ed4d571b83eb9817fd4c057a8bd0

anthrotype:
> The usage info in the file header is redundant

but the description is not redundant. From a brief read I don't understand what this script is meant to do.
Add a brief description in the `argparse.ArgumentParser(description='...')`

anthrotype:
I didn't mean to close it, sorry.

chrissimpkins:
Do you use ttx constants as mock fonts for testing? If so, is it possible to point me to one or more of them? I won't add tests to this branch since this is going to the snippets bin but do need to test it and will do this in a downstream branch.    

Will make the modifications requested to description and push this tonight.

chrissimpkins:
Added description in https://github.com/fonttools/fonttools/pull/1197/commits/53117bcf880d2a466ec0c057cb8cfa4fc23a6d80

anthrotype:
> Do you use ttx constants as mock fonts for testing? If so, is it possible to point me to one or more of them?

anthrotype:
> Do you use ttx constants as mock fonts for testing? If so, is it possible to point me to one or more of them?

i'm not sure what you mean by `ttx constants`. We have a bunch of ttx files somewhere in `Tests/data` that we use for the unit tests.

chrissimpkins:
thanks

chrissimpkins:
just meant string literals for testing

PR #1198, opened 2018-02-26
[varLib] Only recalculate horizontalAdvanceWidth if its not 0

m4rc1e:
Whilst trying to gen instances using varLib.mutator on [MarkaziText](https://github.com/BornaIz/markazitext). I got negative adv warnings for the following glyphs:

```
ERROR: Glyph 'uni0615' has negative advance width
ERROR: Glyph 'uniFBB4' has negative advance width
ERROR: Glyph 'uni0655' has negative advance width
```

These glyphs are Non-Spacing Marks so they should have 0 for their advance width.

When I inspect the glyphs, they have 0 but they're shifted to the right. Even when I center the outlines, the adv width still get recalculated to a negative integer in many cases.

screen shot 2018-02-26 at 10 19 03

This pr will only recalc the horizontalAdvanceWidth if they're not 0 to start with.

I guess I may be missing some edge cases here. I'm more than happy to add them if needs be.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -233,6 +233,7 @@ def _GetCoordinates(font, glyphName):
 # TODO Move to glyf or gvar table proper
 def _SetCoordinates(font, glyphName, coord):
 	glyf = font["glyf"]
+	horizontalAdvanceWidth = font["hmtx"].metrics[glyphName][0]
 	assert glyphName in glyf.glyphs
 	glyph = glyf[glyphName]
 
@@ -261,7 +262,8 @@ def _SetCoordinates(font, glyphName, coord):
 
 	glyph.recalcBounds(glyf)
 
-	horizontalAdvanceWidth = round(rightSideX - leftSideX)
+	if horizontalAdvanceWidth != 0:
+		horizontalAdvanceWidth = round(rightSideX - leftSideX)
 	leftSideBearing = round(glyph.xMin - leftSideX)
 	# XXX Handle vertical
 	font["hmtx"].metrics[glyphName] = horizontalAdvanceWidth, leftSideBearing


Lib/fontTools/varLib/__init__.py
```diff
@@ -261,7 +262,8 @@ def _SetCoordinates(font, glyphName, coord):
 
 	glyph.recalcBounds(glyf)
 
-	horizontalAdvanceWidth = round(rightSideX - leftSideX)
+	if horizontalAdvanceWidth != 0:
+		horizontalAdvanceWidth = round(rightSideX - leftSideX)
```
anthrotype: or `max(round(rightSideX - leftSideX), 0)`

m4rc1e: Far more elegant. Thanks @anthrotype. Shall I make this amend and squash?

anthrotype:
it's a bit weird though that `rightSideX` ends up being smaller than `leftSideX`.
Can you add a print to see what the values are?
I suspect it might be something to do with this issue here:
https://github.com/googlei18n/fontmake/issues/400#issuecomment-366439566

which i've fixed on ufo2ft master but haven't released yet (as there're other unrelated stuff about to be pushed)
https://github.com/googlei18n/ufo2ft/commit/e4f1539811eda8cf6a9ec42ddbba2c53a7d59252

behdad:
Yeah I also like us to better understand the root cause.

anthrotype:
Marc, maybe also try installing ufo2ft from git, and retry to build the interpolatable master ttfs, build the VF, instantiate it with mutator and see if the errors goes away?

anthrotype:
> which i've fixed on ufo2ft master but haven't released yet 

(wrong url, I meant this PR https://github.com/googlei18n/ufo2ft/pull/222)

anthrotype:
on second thought, I don't think it's that. I'll try to debug later.
Marc, how did you build the fonts so I can replicate locally?

m4rc1e:
Print statement before `horizontalAdvanceWidth = round(rightSideX - leftSideX)`
```
name       adv    left    right
uni0615    0      0.0     -1.99987792969
uni0655    0      0.0     -12.999206543
uniFBB4    0      0.0     -7.99951171875
```

Print statement after `horizontalAdvanceWidth = round(rightSideX - leftSideX)`
```
name       adv    left    right
uni0615    -2     0.0     -1.99987792969
uni0655    -13    0.0     -12.999206543
uniFBB4    -8     0.0     -7.99951171875
```

@anthrotype I didn't use fontmake. I genned the fonts with Glyphsapp v2.5b (1116)

---

I've just inspected the source and found they contain adv widths for these Glyphs.

screen shot 2018-02-26 at 11 07 42

When I gen the fonts with Glyphsapp as single ttf instances, It will automatically make the adv width 0.

screen shot 2018-02-26 at 11 08 58


However, If I manually set the adv width to 0 for these glyphs in the source, there isn't a problem anymore. This leads me to believe the fault may be caused by GlyphsApp's approach to removing the adv widths when generating variable fonts.

* On sources with correct adv widths*
Print statement before `horizontalAdvanceWidth = round(rightSideX - leftSideX)`
```
uni0615 0 0.0 0.0
uni0655 0 0.0 0.0
uniFBB4 0 0.0 0.0
```

Looks like it's a problem with the sources. Should we even bother catering for this?

If not, I'll close this.

m4rc1e:
TTX diff of the gvar table for both fonts


left: Font with issue
Right: Adv set to 0 manually.

screen shot 2018-02-26 at 12 19 34

screen shot 2018-02-26 at 12 27 47

Both fonts have the same hmtx table.

anthrotype:
it looks like an issue with glyphsapp to me, if the glyphs in the source files have positive advances but theses don't get zero-ed upon exporting the variable font, wheres as you say they do get zero-ed when exporting the single instance ttf (as they are classified as combining marks). 

/cc @schriftgestalt 

I'm still curious to know whether we incur in the same issue if you make the variable font with fontmake.

still, it'd be better if mutator didn't attempt to write negative advances. Maybe we could detect this case and issue a logging warning but still clam the value to 0.

m4rc1e:
@anthrotype when building the variable font with fontmake on the source with adv widths, I don't get this error.

```
uni0615 0 0 0
uni0655 0 0 0
uniFBB4 0 0 0
```

adv glyphs also get set to 0:

screen shot 2018-02-26 at 12 46 27




I would love to build with fontmake. However, glyphLib needs to  have https://github.com/googlei18n/glyphsLib/pull/306 implemented. This is a separate issue though :-).

> still, it'd be better if mutator didn't attempt to write negative advances. Maybe we could detect this case and issue a logging warning but still clam the value to 0.

I like this approach. Many open source Glyphs projects have this issue. Its so easy to forget that certain marks don't need an adv width.

belluzj:
> I would love to build with fontmake. However, glyphLib needs to have googlei18n/glyphsLib#306 implemented. This is a separate issue though :-)

If you have some time, could you try building with fontmake and this branch of glyphsLib? https://github.com/googlei18n/glyphsLib/pull/244
This branch includes the axes stuff (but I can't guarantee that I have not broken something else).

schriftgestalt:
Thanks for the heads up. I fixes this. Now all nonspacing glyphs have zero width in the glyf and gvar table.

m4rc1e:
@anthrotype Thank you!

PR #1201, opened 2018-02-27
[fealib] set AST element's location=None by default

anthrotype:
I want to change the ufo2ft featureWriters API so that they will manipulate `feaLib.ast.FeatureFile` objects, instead of outputting strings.

Currently, the location is a required positional argument, and it's always the first; in addition, all the constructor arguments are implemented as required positional arguments. 

This makes it cumbersome to create new ast elements from code, as one has to provide a `None` location as first argument, and then one has to pass all the default values for the subsequent positional arguments. The call becomes a very long, with lots of unnecessary `None`s. For example, compare these two (old and proposed) methods to create a new `ValueRecord` object (e.g. for use in a `PairPosStatement` or kerning):

```diff
- value = ast.ValueRecord(None, False,
-                         None, None, -40, None
-                         None, None, None, None)
+ value = ast.ValueRecord(xAdvance=-40)
```

With this PR, I changed all the ast elements' constructors so that the `location` argument becomes optional and defaults to `None`.
Also, we can now make other arguments as kwargs with a default value, now that the first parameter is no longer the location.

This may break other's existing tools that rely on the implicit order of the positional arguments.
I believe it's a necessary break, to move things forward.

cc @mhosken @brawer @behdad

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -849,9 +849,8 @@ class MultipleSubstStatement(Statement):
 
 
 class PairPosStatement(Statement):
-    def __init__(self, enumerated,
-                 glyphs1, valuerecord1, glyphs2, valuerecord2,
-                 location=None):
+    def __init__(self, glyphs1, valuerecord1, glyphs2, valuerecord2,
+                 enumerated=False, location=None):
         Statement.__init__(self, location)
         self.enumerated = enumerated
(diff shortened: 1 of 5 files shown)


behdad:
Nice.

anthrotype:
refactoring is easy when you a have an extensive test suite (thanks Sascha! :+1:)

miguelsousa:
This breaks my open PR 😭 Time to get it in before it gets worse.

PR #1202, opened 2018-02-28
[ttx] add option to dump glyf table's glyphs as individual ttx 

anthrotype:
I rebased @bamidei original PRs (https://github.com/fonttools/fonttools/pull/1035 and https://github.com/fonttools/fonttools/pull/1132) on the current master branch.

With this `ttx` gets a new `-g` option that will dump each glyph from the glyf table as an individual ttx file, in the same directory as the parent ttx file, and with the latter's filename plus an `_` and the glyph name (and the ".ttx" file extension). For simplicity, the `-g` (split glyphs) option also implies the `-s` (spit tables) option, i.e. can't use split glyphs only without also splitting the tables.

The separator between the `_g_l_y_f` table tag and the glyph names in the ttx file name was chosen to be "_" instead of ".",  so that the glyf table's ttx file (containing the index of individual glyph's ttx files) doesn't get lost among hundreds of other ttx files, but appears _after_ all the glyphs ttx files when sorting a folder alphabetically.

The option currently only affects the glyf table, not yet the CFF (PRs are welcome).

Thanks @bamidei for your contribution, and apologies if I've kept your PRs lagging behind for so long.
(@ragingstorm apparently was his old nickname)

Related issue #153

### DIFF
--- a/Lib/fontTools/misc/filenames.py
+++ b/Lib/fontTools/misc/filenames.py
@@ -1,0 +1,207 @@
+"""
+User name to file name conversion.
+This was taken form the UFO 3 spec.
+"""
+from __future__ import unicode_literals
+from fontTools.misc.py23 import basestring, unicode
+
+
+illegalCharacters = "\" * + / : < > ? [ \ ] | \0".split(" ")
+illegalCharacters += [chr(i) for i in range(1, 32)]
+illegalCharacters += [chr(0x7F)]
+reservedFileNames = "CON PRN AUX CLOCK$ NUL A:-Z: COM1".lower().split(" ")
+reservedFileNames += "LPT1 LPT2 LPT3 COM2 COM3 COM4".lower().split(" ")
+maxFileNameLength = 255
+
+
+class NameTranslationError(Exception):
+	pass
+
+# ----------------------
+# User Name to File Name
+# ----------------------
+# This code was taken directly from the ufoNormalizer script in the unified-font-object repositry
+# (https://github.com/unified-font-object/)
+# ...which algorithm was taken directly from the UFO 3 specification
+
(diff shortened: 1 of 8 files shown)


anthrotype:
Ah, and the new `fontTools.misc.filenames` module with the `userNameToFileName` function that converts arbitrary glyph names to compatible filesystem's file names (following UFO3 conventions) is borrowed from ufoLib. Thanks @typesupply

anthrotype:
I'm still actually 51%~49% on the "_" vs "." as the separator between the table tag and the glyph name. E.g.:

FontName._g_l_y_f.U_acute.ttx

vs

FontName._g_l_y_f_U_acute.ttx

I like the "." for consistency reasons, but I see the point raised by @bamidei that with the "_" (when the files are listed alphabetically) it makes it easier to spot the index file `FontName._g_l_y_f.ttx` among the other ones.

:thinking: 

I also thought about somehowe grouping the glyph files in a sub-folder, but that solution opens up other problems (relative vs absolute path, platform-specific path separators, what if a folder with given name already exists, etc.), and I'd like to keep it simple.

And most of the time I think users will use `-d <outputdir>` in combination with `-g` (or `-s`) anyway..

anthrotype:
> FontName._g_l_y_f.U_acute.ttx
> vs
> FontName._g_l_y_f_U_acute.ttx

ok let's put it to a vote: do you guys prefer `_` or `.` to separate `_g_l_y_f` tag from the (ufo3-normalized) glyph name?

benkiel:
`.` would let you split the path easier if you needed to

miguelsousa:
What about using a different separator, like `+` or `-`?

anthrotype:
Hm. I still prefer the `.`. Yeah, I’m going to do that.

Issue #737, opened 2016-11-07
TTX extracts axis values with low precision from fvar


Lorp:
While experimenting with using TTX and OT variations fonts, I was surprised to see that the axis values in fvar (min, max and default, and also settings for fvar's Named Instances) are extracted with low precision. It turns out it's because of the use of fixedToFloat(value, 16). This converts Fixed (16.16) values into decimal floats, using as few digits as possible, so as to still guarantee that converting back to Fixed will be round-trip safe.

I think this choice is the wrong decision for a tool such as TTX, and for some other uses of fontTools — see the use case below.

First, an axample.

Skia contains the following data (hex, then dex) for min, default, max:

wght: 00007AE1, 00010000, 00033333 = 31457, 65536, 209715
wdth: 00009EB7, 00010000, 00014CCC = 40631, 65536, 85196

These values, if extracted with "double" precision, would allow us to set up two extreme corners of Skia using the following CSS:

font-variation-settings: 'wght' 0.47999572753906, 'wdth' 0.61997985839844;
font-variation-settings: 'wght' 3.1999969482422, 'wdth' 1.2999877929688;

However, TTX extracts the following:

   
   <Axis>
     <AxisTag>wght</AxisTag>
     <MinValue>0.48</MinValue>
     <DefaultValue>1.0</DefaultValue>
     <MaxValue>3.2</MaxValue>
     <NameID>257</NameID>
   </Axis>

   
   <Axis>
     <AxisTag>wdth</AxisTag>
     <MinValue>0.61998</MinValue>
     <DefaultValue>1.0</DefaultValue>
     <MaxValue>1.29999</MaxValue>
     <NameID>258</NameID>
   </Axis>

So, we have used a couple fewer bytes, but in the process have introduced some uncertainties.

Imagine the use case of a UI, or an animation, which called for extremely fine control over the variation axis, more precise than the 1/65536 quantization allows. If the min, max and default values are extracted using TTX, the controls would not work correctly around those values. It seems analogous to control points being on a coarse grid, while the curves that they define follow much more precise positions. Thus – for fvar at least – I would advocate for "double"-style precision in the decimal representation of these values in TTX. I would also suggest checking where else fixedToFloat() is used, and verifying if there is any benefit. Note that fixedToFloat() is a slow function compared to the simple floating point division that it replaces.

[...]

PR #1204, opened 2018-03-01
[ttGlyphPen] decompose components if transform overflows F2Dot14

### LINKED ISSUE #1200: [ttGlyphPen] decompose components if transform overflows F2Dot14
https://github.com/googlei18n/ufo2ft/issues/217

The UFO GLIF spec allows any numbers for xScale, xyScale, yxScale, yScale, xOffset, yOffset, however the OpenType glyf spec uses F2Dot14 numbers, which are encoded as a signed 16-bit integer and therefore can only contain values from -32768 (-0x8000, or -2.0) to +32767 included (0x7FFF, or +1.99993896484375...).

We can't let the `struct.error` propagate.

I think we have to handle the case of +2.0 specially, and treat it as if it were 1.99993896484375. By doing that we can support truetype component transforms in the range -2.0 to +2.0 (inclusive), for the sake of simplicity.

Then, we also need to have the ttGlyphPen decompose the components if their transform values are less than -2.0 or they are greater than +2.0 (not greater and equal), as these can't fit in the TrueType glyf table.

moyogo:
Fixes #1200, with little cheat when transform scale is close or equal to 2.

### DIFF
--- a/Lib/fontTools/pens/ttGlyphPen.py
+++ b/Lib/fontTools/pens/ttGlyphPen.py
@@ -86,7 +86,7 @@ class TTGlyphPen(AbstractPen):
         eps = .5 / (1 << 14)
 
         def closeTo2(value):
-            return True if abs(value - 2) <= eps else False
+            return abs(value - 2) <= eps
 
         for i, (glyphName, transformation) in enumerate(self.components):
             if any(s > 2 or s < -2 for s in transformation[:3]):
--- a/Lib/fontTools/pens/ttGlyphPen.py
+++ b/Lib/fontTools/pens/ttGlyphPen.py
@@ -83,21 +83,24 @@ class TTGlyphPen(AbstractPen):
         assert self._isClosed(), "Didn't close last contour."
 
         almost2 = 1.99993896484375  # F2Dot14 0b1.11111111111111
+        eps = .5 / (1 << 14)
+
+        def closeTo2(value):
+            return True if abs(value - 2) <= eps else False
+
         for i, (glyphName, transformation) in enumerate(self.components):
-            if any(s > 2 or s < -2 for s in transformation[0::3]):
+            if any(s > 2 or s < -2 for s in transformation[:3]):
                 # can't have scale >= 2 or scale < -2:
                 tpen = TransformPen(self, transformation)
                 self.glyphSet[glyphName].draw(tpen)
                 self.components.remove((glyphName, transformation))
-            elif any(almost2 < s <= 2 for s in transformation[0::3]):
+            elif any(closeTo2(s) for s in transformation[:3]):
                 # use closest F2Dot14 value to 2 when possible
                 transformation = (
-                    (almost2 if almost2 < transformation[0] <= 2 else
-                     transformation[0]),
-                    transformation[1],
-                    transformation[2],
-                    (almost2 if almost2 < transformation[3] <= 2 else
-                     transformation[3]),
+                    almost2 if closeTo2(transformation[0]) else transformation[0],
+                    almost2 if closeTo2(transformation[1]) else transformation[1],
+                    almost2 if closeTo2(transformation[2]) else transformation[2],
+                    almost2 if closeTo2(transformation[3]) else transformation[3],
                     transformation[4],
                     transformation[5]
                 )
--- a/Lib/fontTools/pens/ttGlyphPen.py
+++ b/Lib/fontTools/pens/ttGlyphPen.py
@@ -82,6 +82,27 @@ class TTGlyphPen(AbstractPen):
     def glyph(self, componentFlags=0x4):
         assert self._isClosed(), "Didn't close last contour."
 
+        almost2 = 1.99993896484375  # F2Dot14 0b1.11111111111111
+        for i, (glyphName, transformation) in enumerate(self.components):
+            if any(s > 2 or s < -2 for s in transformation[0::3]):
+                # can't have scale >= 2 or scale < -2:
+                tpen = TransformPen(self, transformation)
+                self.glyphSet[glyphName].draw(tpen)
+                self.components.remove((glyphName, transformation))
+            elif any(almost2 < s <= 2 for s in transformation[0::3]):
+                # use closest F2Dot14 value to 2 when possible
+                transformation = (
+                    (almost2 if almost2 < transformation[0] <= 2 else
+                     transformation[0]),
+                    transformation[1],
+                    transformation[2],
+                    (almost2 if almost2 < transformation[3] <= 2 else
+                     transformation[3]),
+                    transformation[4],
+                    transformation[5]
+                )
+                self.components[i] = (glyphName, transformation)
+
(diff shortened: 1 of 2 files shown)


justvanrossum:
Why only test for `transformation[0]` and `transformation[3]`? `[1]` and `[2]` have the exact same limitation.

anthrotype:
also, wouldn't it be better to avoid direct comparison of floats with integers, and instead define some small epsilon value for the tolerance i.e. check `abs(value - 2) <= epsilon`

we also have an implementation of `math.isclose` in `fontTools.misc.py23` which could come handy
https://docs.python.org/3/library/math.html#math.isclose

anthrotype:
a good epsilon value could be the one we use for approximating a fixed-point number to the shortest decimal representation in fixedToFloat:

```python
>> precisionBits = 14
>> scale = 1 << precisionBits
>> epsilon = 0.5 / scale
>> epsilon
3.0517578125e-05
```

https://github.com/fonttools/fonttools/blob/c9308d2abce6ba767288eecf9b7a6acd4d941f8d/Lib/fontTools/misc/fixedTools.py#L30

anthrotype:
actually, even simpler would be converting the transform's floats to fixed-point numbers using our `floatToFixed`(value, 14)`, and check if we get +32768 (+0x8000) which would overflow our signed short, in which case we "cheat" and treat it as if we were passed a slightly smaller float 1.99994 (= the result of the `floatToFixed(0x7FFF, 14)`).

Same for the decomposition, if we get a float which when converted to fixed-point number would become `> 0x8000` or `< -0x8000`, then we decompose.

justvanrossum:
I still think that "shortest decimal" thing that `fixedToFloat()` does is a very bad idea :(

I expect `fixedToFloat(0x7FFF, 14)` to result in 1.99993896484375, not 1.99994. That's just bogus.

anthrotype:
oh, I forgot there's also `floatToFixedToFloat` for "quantizing" floats to fixed-point numbers:

https://github.com/fonttools/fonttools/blob/c9308d2abce6ba767288eecf9b7a6acd4d941f8d/Lib/fontTools/varLib/mutator.py#L47-L48

https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/misc/fixedTools.py#L55

anthrotype:
> "shortest decimal" thing that fixedToFloat() does is a very bad idea

I think it's good for pretty-printing fixed-point numbers as floats in the TTX dump without losing precision, and to be able to round-trip between the fixed-point number in the font binary and its float representation in the ttx file.
But it's a separate issue (https://github.com/fonttools/fonttools/issues/737)

Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -83,21 +83,24 @@ def glyph(self, componentFlags=0x4):
         assert self._isClosed(), "Didn't close last contour."
 
         almost2 = 1.99993896484375  # F2Dot14 0b1.11111111111111
+        eps = .5 / (1 << 14)
+
+        def closeTo2(value):
+            return True if abs(value - 2) <= eps else False
```
anthrotype:
```diff
- return True if abs(value - 2) <= eps else False
+ return abs(value - 2) <= eps
```

Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -83,21 +83,24 @@ def glyph(self, componentFlags=0x4):
...
+
         for i, (glyphName, transformation) in enumerate(self.components):
-            if any(s > 2 or s < -2 for s in transformation[0::3]):
+            if any(s > 2 or s < -2 for s in transformation[:3]):
                 # can't have scale >= 2 or scale < -2:
                 tpen = TransformPen(self, transformation)
                 self.glyphSet[glyphName].draw(tpen)
                 self.components.remove((glyphName, transformation))
-            elif any(almost2 < s <= 2 for s in transformation[0::3]):
+            elif any(closeTo2(s) for s in transformation[:3]):
```
anthrotype: wouldn't it be better if you check the closeTo2 stuff as the first `if` branch rather than `elif` one?

Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -83,21 +83,24 @@ def glyph(self, componentFlags=0x4):
...
 
         almost2 = 1.99993896484375  # F2Dot14 0b1.11111111111111
+        eps = .5 / (1 << 14)
+
+        def closeTo2(value):
+            return True if abs(value - 2) <= eps else False
+
         for i, (glyphName, transformation) in enumerate(self.components):
-            if any(s > 2 or s < -2 for s in transformation[0::3]):
+            if any(s > 2 or s < -2 for s in transformation[:3]):
```
anthrotype: you mean `transformation[:4]` here, not `:3`

Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -82,6 +82,30 @@ def addComponent(self, glyphName, transformation):
     def glyph(self, componentFlags=0x4):
         assert self._isClosed(), "Didn't close last contour."
 
+        almost2 = 1.99993896484375  # F2Dot14 0b1.11111111111111
+        eps = .5 / (1 << 14)
+
+        def closeTo2(value):
```
anthrotype: why do we need to define `closeTo2` every time one calls the `.glyph()` method?
you could move it to the global scope and call it `_closeTo2` along with the two other constant values.

Tests/pens/ttGlyphPen_test.py
```diff
@@ -135,6 +135,82 @@ def test_keep_duplicate_end_point(self):
...
+        glyphSet = {}
+        pen = TTGlyphPen(glyphSet)
+
+        pen.moveTo((0, 0))
+        pen.lineTo((0, 1))
+        pen.lineTo((1, 0))
+        pen.closePath()
+        glyphSet[componentName] = _TestGlyph(pen.glyph())
+
+        pen.addComponent(componentName, (1.99999, 0, 0, 1, 0, 0))
```
anthrotype: maybe you could also add an example with `2.00001`, which is almost2 given the epsilon we chose

Lib/fontTools/pens/ttGlyphPen.py
anthrotype: actiually, the two tests (is any float almost 2, and is any float less than -2 or greater than +2) are independent from each other (as in two separate `if` statements following one another, instead of `elif`): in the former case, we want to round a value of +2.0 to `fixedToFloat(0x7FFF, 14)` to try keep the component as such despite the approximation; in the latter case, we want to decompose the whole composite glyph because we can't fit it anymore.

In the case of `(3, 0, 0, 2, 0, 0)`, we really want to decompose, but instead with the current code the first `if` branch matches because there's `any` float that is `_closeTo2`, and the `elif` one is not.

anthrotype:
taken on by #1205

PR #1205, opened 2018-03-01
Fixups on top of PR #1204

### LINKED ISSUE #1200: [ttGlyphPen] decompose components if transform overflows F2Dot14

anthrotype:
this is a follow up from @moyogo's PR #1204 (after a private conversation).

Fixes #1200

### DIFF
--- a/Lib/fontTools/pens/ttGlyphPen.py
+++ b/Lib/fontTools/pens/ttGlyphPen.py
@@ -87,19 +87,11 @@ class TTGlyphPen(AbstractPen):
     def glyph(self, componentFlags=0x4):
         assert self._isClosed(), "Didn't close last contour."
 
-        overflowing = False
-        for i, (glyphName, transformation) in enumerate(self.components):
-            if any(s > 2 or s < -2 for s in transformation[:4]):
-                # can't encode transform values > 2 or < -2 in F2Dot14,
-                # so we must decompose the component
-                overflowing = True
-                break
-            elif any(MAX_F2DOT14 < s <= 2 for s in transformation[:4]):
-                # clamp values ~= +2.0 so we can keep the component
-                clamped = [MAX_F2DOT14 if MAX_F2DOT14 < s <= 2 else s
-                           for s in transformation[:4]]
-                clamped.extend(transformation[4:])
-                self.components[i] = (glyphName, tuple(clamped))
+        # we can't encode transform values > 2 or < -2 in F2Dot14,
+        # so we must decompose the glyph if any transform exceeds these
+        overflowing = any(s > 2 or s < -2
+                          for (glyphName, transformation) in self.components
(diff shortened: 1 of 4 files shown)


Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -91,22 +88,18 @@ def glyph(self, componentFlags=0x4):
         assert self._isClosed(), "Didn't close last contour."
 
         for i, (glyphName, transformation) in enumerate(self.components):
```
anthrotype: ah! we sholdn't modify a list that we are iterating on.. i'll fix it

behdad:
I'm not sure doing this in TTGlyphPen doesn't have consequences. In particular in the varfont workflow. IMO this should be done in ufo2ft, and done to all masters the same way.

anthrotype:
yes, that will need to be done anyway when building interpolatable TTFs. But for the general case when a TTGlyphPen is sent arbitrary components' transforms (without knowing where they come from and what is the purpose), then it will do the right thing and decompose when they exceed the F2Dot14 bounds. It's always better than raise an error.

anthrotype:
unless... we are happy with raising a struct.error here and leave it to the client (in our case, ufo2ft) to deal with it the way they deem appropriate (e.g. clamp to almost 2, decompose if out of range).

behdad:
Make it an option then, detail to True is fine, but disable in interpolatable output.

anthrotype:
> disable in interpolatable output

so we don't have to check twice?

anthrotype:
i'm confused: when you say "default to True" do you mean raise the error by default, or "clamp and decompose" by default? I'd say the latter.

behdad:
Default whatever you want. Just disable decompose in interpolatablr output.

anthrotype:
ok

anthrotype:
ok, I guarded both the decomposition and the clamping-to-2 with an option (called `handleOverflowingTransforms`), defaults to True. Going to merge this once the bots are happy

Issue #1084, opened 2017-10-25
varLib "IUP" delta optimization as an optional feature instead of on by default


andyclymer:
Some variable font implementations seem to have problems drawing glyphs that have "IUP" delta optimizations.

Currently, `_add_gvar()` sets the optimize option to be True as a default, but the API for disabling optimization could be easier to get to if the option was available as an argument of the `build()` method as well.

https://github.com/fonttools/fonttools/blob/3fbc4688eb5ef1057c58ae8bc12a1ba0b59b12d5/Lib/fontTools/varLib/__init__.py#L656

anthrotype:
yes, maybe we could have an `optimize=True` argument to `varLib.build`. Would you like to work on a pull request?
Which implementations are you referring to, if I may ask?

behdad:
Can you share details about which implementations have this problem?  I think I've heard Adobe's does.

brawer:
Any implementations that aren’t yet covered by https://github.com/unicode-org/text-rendering-tests? If we’re talking about CoreText, is this covered by test case [GVAR-9](https://rawgit.com/unicode-org/text-rendering-tests/master/reports/CoreText.html#GVAR-9) or do we need to add another test to the test suite? (Afaik Apple has fixed test case GVAR-9 in macOS 10.13; I just haven’t yet set up a box for reproducing that it’s been fixed).

andyclymer:
I'm specifically observing it in Adobe's implementation. I'm sure it's probably just a bug on their end, but for now it could be nice to have the optimization as an option.

behdad:
Is this communicated to Adobe?
@readroberts @miguelsousa

jenskutilek:
Yes, I sent a bug report with a demo font.

readroberts:
And it is now in the internal Adobe bug database, and has been reproduced. Should be fixed at some point.

andyclymer:
Thanks everyone!

brawer:
@jenskutilek, would you be interested in contributing your test font (just a subset with few failing glyphs) to Unicode’s test suite? https://github.com/unicode-org/text-rendering-tests

jenskutilek:
I opened an issue with the test font: https://github.com/unicode-org/text-rendering-tests/issues/42

behdad:
I'm open to disabling this for now, until Adobe has a new release so we can just tell people "Update XXX".

behdad:
FWIW here's the patch:
```patch
[...]

PR #1208, opened 2018-03-05
Add flag to disable IUP optimization

robmck-ms:
Added a command line flag to varLib, and parameter to varLib.build() to disable IUP optimization. This is handy for building fonts with / without IUP for debugging.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -714,7 +714,7 @@ def load_designspace(designspace_filename):
 	return axes, internal_axis_supports, base_idx, normalized_master_locs, masters, instances
 
 
-def build(designspace_filename, master_finder=lambda s:s, exclude=[]):
+def build(designspace_filename, master_finder=lambda s:s, exclude=[], optimize=True):
 	"""
 	Build variation font from a designspace file.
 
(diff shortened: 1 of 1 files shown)


Lib/fontTools/varLib/__init__.py
```diff
@@ -778,6 +778,7 @@ def main(args=None):
 	parser.add_argument('designspace')
 	parser.add_argument('-o', metavar='OUTPUTFILE', dest='outfile', default=None, help='output file')
 	parser.add_argument('-x', metavar='TAG', dest='exclude', action='append', default=[], help='exclude table')
+	parser.add_argument('--disableIUP', dest='optimize', action='store_false', default=True, help='do not perform IUP optimization')
```
moyogo: `action='store_false'` defaults to True (https://docs.python.org/3/library/argparse.html#action), setting `default=True` is not necessary.

anthrotype: :bike: :house_with_garden:  I don't like the camelCase `--disableIUP` too much. Maybe `--disable-IUP`?

anthrotype:
Thanks @robmck-ms for the PR.

this reminded me that some month ago we sort of agreed on disabling this IUP optimization by default until some Adobe apps fixed some related bugs, but then it seems I forgot to do that:

https://github.com/fonttools/fonttools/issues/1084#issuecomment-344673595

Does anyone know if that is still an issue?

Lib/fontTools/varLib/__init__.py
khaledhosny: `--disable-iup` would even be better, upper case in command line options looks unusual.

moyogo:
@anthrotype this hadn’t been fixed in Adobe apps when we tested a VF a couple of weeks ago.

anthrotype:
aargh. Then should disable it by default and add instead an `--enable-iup`?

moyogo:
Do we really want to switch the default command line argument whenever that happens?
I don’t mind having to use `--disable-iup` for now.

anthrotype:
ok, i'll merge and fix

anthrotype:
https://github.com/fonttools/fonttools/commit/168c7e1cd35a5c008918af522166394da876002f

PR #1211, opened 2018-03-07
Fix bugs in compiling CFF2 when using a font with an FDArray.

readroberts:
See commit comments. Also includes a bug fix in varLib that breaks working with CFF2 fonts.

### DIFF
--- a/Lib/fontTools/ttLib/tables/otTables.py
+++ b/Lib/fontTools/ttLib/tables/otTables.py
@@ -562,7 +562,7 @@ class VarIdxMap(BaseTable):
 
 		glyphOrder = font.getGlyphOrder()
 		mapping = [mapping[g] for g in glyphOrder]
-		while mapping and len(mapping) > 1 and mapping[-2] == mapping[-1]:
+		while len(mapping) > 1 and mapping[-2] == mapping[-1]:
 			del mapping[-1]
 
 		rawTable = { 'mapping': mapping }


Lib/fontTools/cffLib/__init__.py
```diff
@@ -271,8 +271,17 @@ def convertCFFToCFF2(self, otFont):
...
+			fdArray = topDict.FDArray
 			privateOpOrder = buildOrder(privateDictOperators2)
 			for fontDict in fdArray:
+				fontDict.order = fontDict.orderCFF2
+				fontDict._isCFF2 = True
+				for key in fontDict.rawDict.keys():
+					if key not in fontDict.order:
+						del fontDict.rawDict[key]
+						if hasattr(fontDict, key):
+							exec("del fontDict.%s" % (key))
```
miguelsousa: @readroberts `exec()` is bad. Use `delattr()` instead https://docs.python.org/3/library/functions.html#delattr

Lib/fontTools/cffLib/__init__.py
```diff
@@ -256,7 +256,8 @@ def convertCFFToCFF2(self, otFont):
 				charStrings.charStringsIndex.fdArray = fdArray
 			else:
 				charStrings.fdArray = fdArray
-			fontDict = FontDict(isCFF2=True)
+			fontDict = FontDict()
+			fontDict.setCFF2(True)
```
anthrotype: (i preferred the `isCFF=True` in the constructor, but nevermind)

behdad:
I really wanted us to avoid having a isCFF2 state on any object. That should just be the property of read and write methods.

Also, I'd think we should move convertCFFToCFF2 to a separate module.

anthrotype:
yes, I agree with Behdad we should avoid keeping a `isCFF2` state. I just wasn't sure if we're consistent about this in the rest of cffLib.

Lib/fontTools/cffLib/__init__.py
readroberts: Me too, but there is an assert at the start of BaseDict.__init__() which prevents this. I don't understand why the assert exists, and so didn't want to remove it. I can't see that the assert serves a useful check. If you agree, I would like to remove it.

readroberts:
I am also in favor of moving convertCFFToCFF2() out of cffLib.py. I am pretty sure that it is used only by the afdko buildCFF2VF script. It also uses a model which made sense to me when there were only a few differences between CFF and CFF2, but which I now think is confusing: 1) add the master data to a the base CFF font, using blend operators that are illegal in CFF2, 2) convert the necessary structures from CFF to CFF2. I would rather redo this by creating a new CFF2 table, and adding CFF2 data directly. In this case as will, I think it would be better as an external module.

readroberts:
Avoiding an isCFF2 state is appeals to me, too. My original solution to this was subclassing CFF classes to create CFF2 versions. However, this turned out to require either subclassing almost all of the  classes, or using tests to see if parent or child objects were a CFF2 class. The structure looked a lot simpler with the isCFF2 state. However, if someone else wants to refactor cffLib.py so as to avoid this state, I am all in favor.

Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -562,7 +562,7 @@ def preWrite(self, font):
 
 		glyphOrder = font.getGlyphOrder()
 		mapping = [mapping[g] for g in glyphOrder]
-		while mapping and mapping[-2] == mapping[-1]:
+		while mapping and len(mapping) > 1 and mapping[-2] == mapping[-1]:
```
behdad: The check for "mapping" itself is redundant now. So, please rewrite as "while len(mapping) > 1 and ..."

readroberts: Done.

behdad:
> However, if someone else wants to refactor cffLib.py so as to avoid this state, I am all in favor.

I thought @anthrotype already did.. Anyway, yeah, can be done as followup. The problem with keeping isCFF2 was that it was doing the wrong thing if you tried to load a CFF and write a CFF2. All the lazy-loading will be messed up.

anthrotype:
yes, I did https://github.com/fonttools/fonttools/pull/1007

readroberts:
In  PR #1007 by @anthrotype, several classes retained their internal isCFF2 state. The main point was that 'isCFF2' was passed down the compile and decompile chains so that the internal state would get reset. I quote from #1007:
"Index classes get a isCFF2 argument in constructor (used for decompiling); must be True/False if file argument is not None; it is stored as self._isCFF2 to support lazy loading". 

The BaseDict and derived classes also got a "self._isCFF2" field. I don't believe that I have done anything to change this.  My comment was because I thought @behdad  was proposing doing away with the even the self._isCFF2 state that was defined in PR #1007.

Incidentally, I still don't understand the need to require that the decompile  'isCFF2'  and "file' arguments both be either None or not be None (line 552, start of class Index() __init__() method). I certainly have a use for setting 'isCFF2' when file is None - this happens when you are building a CFF2 file from scratch.

anthrotype:
the `self._isCFF2` should be used only for (lazily) decompiling, not when compiling (or that was the intention). When creating a CFF from scratch (without file) it should not be needed.
Do we check that flag also when compiling? If so, then we shouldn't. We should pass isCFF2 from the caller down to the compile chain.

readroberts:
I'd like to move convertCFFToCFF2() out of cffLib.py, and make another pass at suppressing the need to have an isCFF2 status when creating a CFF from scratch. Would you prefer to rebase this and merge, or have me do it, and add the additional changes?

anthrotype:
> ... or have me do it, and add the additional changes?

yes, please. Thanks Read

readroberts:
@behdad @anthrotype @cchapman I'd like to sanity check on some basic goals. At the moment, cffLib does reflect the structural differences between CFF2 and CFF in both the XML output and the classes which hold the decompiled data. This means that CFF-structured data can be compiled only to a CFF table, and CFF2-structured data can be compiled only a CFF2 table. To compile one kind of CFF data to the other, it must first be converted into the structure of the other.  There are a couple significant differences between the two structures:
- no String index in CFF2
- no CID info in CFF2
- topDict held as a single table in CFF2, vs an element in an index in CFF
- Private Dict is referenced from the FDArray index in CFF2, rather than directly from Top Dict in the CFF.

I did this because I wanted to see these structural differences in the XML output. I also didn't see the point of having the data held as a superset of the possible data structures in order to enable  compiling to either CFF or CFF2, because you can't. CFF2 has data that cannot be held in a CFF font: multiple hint dicts (aka FontDicts in the FDArray index), and Variable font stuff. This means that although all CFF data can be compiled as CFF2 data ( although with loss of data - this means throwing away the String table info, and any CID-related info), not all CFF2 data can be compiled to CFF tables. 

Does it seem reasonable to you to maintain this approach, or would you indeed rather see both CFF and CFF2 tables decompiled, and dumped as XML, to a common set data structures? If we go to the latter approach, we would need to accept that some CFF-sh data will either fail to compile to the other, or will have to lose data ( e.g losing CID data when compiling to CFF2, or loosing multiple hint dicts and variable font info when compiling to CFF). We would also need to work out what should be dumped to XML.

anthrotype:
> I did this because I wanted to see these structural differences in the XML output. I also didn't see the point of having the data held as a superset of the possible data structures in order to enable compiling to either CFF or CFF2, because you can't. CFF2 has data that cannot be held in a CFF font [...] although all CFF data can be compiled as CFF2 data [...], not all CFF2 data can be compiled to CFF tables.

These all sound to me like good arguments for maintaining the current approach. I think we should keep the separation (if not make it even more explicit) and have a function that does the necessary conversion between the two.

anthrotype:
rebased and merged with #1271. Opened new issue #1272 to track open TODOs

Issue #1057, opened 2017-09-18
Merge failure for Noto


brawer:
This merge command:

```sh
fonttools merge \
    ../noto-source/master_ttf_interpolatable/NotoSans-Regular.ttf \
    ../noto-source/master_ttf_interpolatable/NotoSansCypriot-Regular.ttf
```

fails with the following error:

```
Traceback (most recent call last):
  File "./fonttools", line 12, in <module>
    sys.exit(main())
  File "/home/sascha/src/fonttools/Lib/fontTools/__main__.py", line 29, in main
    runpy.run_module(mod, run_name='__main__')
  File "/usr/lib/python2.7/runpy.py", line 192, in run_module
    fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/sascha/src/fonttools/Lib/fontTools/merge.py", line 1079, in <module>
    sys.exit(main())
  File "/home/sascha/src/fonttools/Lib/fontTools/misc/loggingTools.py", line 372, in wrapper
    return func(*args, **kwds)
  File "/home/sascha/src/fonttools/Lib/fontTools/merge.py", line 1072, in main
    font = merger.merge(args)
  File "/home/sascha/src/fonttools/Lib/fontTools/merge.py", line 913, in merge
    table = clazz(tag).merge(self, tables)
  File "/home/sascha/src/fonttools/Lib/fontTools/merge.py", line 575, in merge
    assert (table is not None and table is not NotImplemented), "Have duplicates to resolve for font %d but no GSUB" % (i + 1)
AssertionError: Have duplicates to resolve for font 2 but no GSUB
```

The same error happens for quite a few other fonts in Noto. Examples are NotoSansCarian-Regular.ttf or NotoSansDeseret-Regular.ttf, but there's more. See the attached [merge-failures.zip](https://github.com/fonttools/fonttools/files/1311746/merge-failures.zip) for some fonts I'm trying to merge.

This might also be a problem with the Noto sources, but I'm not sure what exactly would be wrong, or how exactly the vendor should be told to change the sources so that the shards can be merged into a single font.

brawer:
[...]

PR #1216, opened 2018-03-15
[merge] warn if font has duplicates to resolve but no GSUB

anthrotype:
#1057


khaledhosny:
@anthrotype what is the status of this?

khaledhosny:
This change seem to have been incorporated at some point already.

PR #1220, opened 2018-03-26
Switching from comma to % so log messages are displayed correctly in a GUI context.

michielkauwatjoe:
Log output was correct on the command line but displayed '%' characters instead of formatting the arguments with a custom logging Handler object.


anthrotype:
no, the current code is correct. We only want to evaluate the string interpolation if the log message is actually handled, that's why we don't do "%" immediately in the string literal, but use the comma.
That's a feature of logging.

You have to use the `LogRecord.getMessage()` method to get the full message string for a LogRecord instance after merging any arguments following the comma.
https://docs.python.org/3.6/library/logging.html#logging.LogRecord.getMessage

What are you trying to do with your custom logging Handler?

michielkauwatjoe:
@anthrotype Right, my bad, I'm new to logging. I'm displaying the various log levels in different colors in a TextField. Thanks!

PR #1230, opened 2018-04-03
Fix GlyphCoordinates overflow

anthrotype:
Early on I pushed https://github.com/fonttools/fonttools/commit/0055f9414ca5094b0c9beb0176ac4998cdbbba70 to fix an OverflowError occurring while computing deltas in VariationModel.getDelta, when the values don't fit a signed short integer.
https://github.com/fonttools/fonttools/commit/0055f9414ca5094b0c9beb0176ac4998cdbbba70

However, I think a better fix is to change GlyphCoordinates so that whenever it is about to modify its inner array, it first checks that the value does not overflow, and if it does, upgrades it to an array of floats (doubles actually).

It might be a bit slower, but I think it's better than checking for overflow errors everywhere we use glyph coordinates.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -1220,6 +1220,9 @@ class GlyphCoordinates(object):
 	def _checkFloat(self, p):
 		if self.isFloat():
 			return p
+		if any(v > 0x7FFF or v < -0x8000 for v in p):
+			self._ensureFloat()
+			return p
 		if any(isinstance(v, float) for v in p):
 			p = [int(v) if int(v) == v else v for v in p]
 			if any(isinstance(v, float) for v in p):
@@ -1259,7 +1262,6 @@ class GlyphCoordinates(object):
 		del self._a[i]
 		del self._a[i]
 
-
 	def __repr__(self):
 		return 'GlyphCoordinates(['+','.join(str(c) for c in self)+'])'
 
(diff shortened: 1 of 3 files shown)


anthrotype:
I don't know, I feel a bit uncomfortable with this, as we kind of lose the advantages of using C array if we have to check every time we update it. I would rather ask for forgiveness than permission, but as you see there are several methods that attempt to modify the internal `self._a` directly. We would have to wrap all of them in `try: except OverflowError:`.
And then there's the `array.extend` method which is fast and avoids a python for loop, but what if a value overflows in the middle of it, we can't recover gracefully so for those we'd have to the loop beforehand to check for possible overflows.
Well, let's keep thinking about it. For now, I'm ok to keep the workaround localised to `VariationModel.getDelta` (https://github.com/fonttools/fonttools/commit/0055f9414ca5094b0c9beb0176ac4998cdbbba70).
Behdad probably knows other places where these can occur.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -1220,6 +1220,9 @@ def _ensureFloat(self):
 	def _checkFloat(self, p):
 		if self.isFloat():
 			return p
+		if any(v > 0x7FFF or v < -0x8000 for v in p):
```
behdad: Merge this with the next if, so we only iterate once, and only if it matches, then check again, which case we have to deal with?  Maybe doesn't matter.

behdad:
Thanks. I was originally thinking about catching exception in the math methods. But this works too.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
anthrotype: the two checks are kind of independent from each other. The first if is to _ensureFloat whenever the value overflows a signed short, the second is to _ensureFloat when any value is actually a float. We want to check the first whether or not the type is float and short-circuit to _ensureFloat, without the extra casting to int and re-check for floats, because when we overflow it's all floats anyway.

PR #1232, opened 2018-04-03
Add context guard to ttfont

andyzickler:
This implements the [`__enter__` and `__exit__` context guard](https://www.python.org/dev/peps/pep-0343/) methods in the `TTFont` class. With this you can do something like
```python
with TTFont('/path/to/font.ttf') as font:
     print(font.getGlyphNames())
```
and have the file auto closed when the block is exited. This seems pretty straight forward, but I'm happy for any feedback.

### DIFF
--- a/Lib/fontTools/ttLib/ttFont.py
+++ b/Lib/fontTools/ttLib/ttFont.py
@@ -141,6 +141,12 @@ class TTFont(object):
 		self.flavor = self.reader.flavor
 		self.flavorData = self.reader.flavorData
 
+	def __enter__(self):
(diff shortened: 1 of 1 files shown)


behdad:
Neat.

anthrotype:
Thanks @andyzickler!
Yes, it’s nice and pythonic.
Although in the default case when lazy argument is None (or explicitly False) and the input file is a path string, we are already closing the file object in the `__init__`, and reading all data in an in-memory bytes stream. But that may change in the future.
Let’s merge this

PR #1242, opened 2018-04-16
[varLib] make STAT 1.2, and reuse fvar nameIDs

anthrotype:
STAT 1.0 is deprecated
See https://github.com/googlei18n/fontmake/issues/417

also we should use the same nameIDs that were defined for the fvar, not add additional ones.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -169,22 +169,26 @@ def _add_avar(font, axes):
 	return avar
 
 def _add_stat(font, axes):
+	# for now we just get the axis tags and nameIDs from the fvar,
+	# so we can reuse the same nameIDs which were defined in there.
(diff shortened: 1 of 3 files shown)


Lib/fontTools/varLib/__init__.py
```diff
@@ -169,22 +169,26 @@ def _add_avar(font, axes):
...
 def _add_stat(font, axes):
+	# for now we just get the axis tags and nameIDs from the fvar,
+	# so we can reuse the same nameIDs which were defined in there.
+	# TODO make use of 'axes' once it adds style attributes info:
+	# https://github.com/LettError/designSpaceDocument/issues/8
 
 	if "STAT" in font:
             return
 
 	nameTable = font['name']
```
behdad: Do we still need the name table?

anthrotype: we don't, I'll remove it

PR #1249, opened 2018-04-25
[varLib] add a --master-finder command line option 

anthrotype:
Allow to customize the location of the interpolatable master ttfs relative to the designspace source filenames.

Takes a template string containing special placeholders that are parametrized with the value of each source file name:
- `{fullname}`: the absolute source file name;
- `{basename}`: the source file name without its directory;
- `{stem}` is the basename without the file extension;
- `{ext}` is the source file extension;
- `{dirname}` is the directory of the absolute source file name.

The default value remains the current one: i.e. `master_ttf_interpolatable/{stem}.ttf`


behdad:
Very nice implementation!

PR #1253, opened 2018-04-28
Designspacelib layer spec

LettError:
This adds `layerName` to designspaceLib sourceDescriptor objects, and reads/writes a `layer` attribute to the source elements.

Note: I can't run the fonttools wide tests so I can't see if the `designspace_test.py` does the right thing.


Doc/source/designspaceLib/scripting.rst
```diff
@@ -111,6 +111,7 @@ outline geometry, kerning and font.info that we want to work with.
 -  The **path** attribute is the absolute path to an existing UFO.
 -  The **name** attribute is a unique name for this source used to keep
    track it.
+-  The **layerName** attribute is the name of the UFO3 layer. Default None for ``foreground``.
```
justvanrossum: Shouldn't that be `public.default`?

anthrotype:
> I can't run the fonttools wide tests so I can't see if the designspace_test.py does the right thing

you should be able to `pip install tox`, and then run the tests with the `tox` command.
You need to have both python2.7 and python3.6 executables in your $PATH, as we configured `tox` to run tests against both of them when no explicit environment is listed on the tox command line.
Or to only run the tests for python2.7, do `tox -e py27-cov`, and `tox -e py36-cov` for python3.6 only.

If you prefer, you can pip install pytest and the other test requirements, and run the pytest command directly, instead of calling them via tox.

This should all be explained in the README, please let me know if anything is not clear.

I'll review the code now, thanks

Tests/designspaceLib/data/test.designspace
```diff
@@ -38,6 +38,12 @@
                 <dimension name="width" xvalue="20" />
             </location>
         </source>
+        <source familyname="MasterFamilyName" filename="masters/masterTest2.ufo" layer="supports" name="master.ufo2" stylename="Supports">
```
anthrotype: the "stylename" attibute in the case of support layers seems a bit redundant, even more so than for normal sources. I wonder if it could be omitted in these cases.

Tests/designspaceLib/designspace_test.py
```diff
@@ -788,98 +801,100 @@ def __removeAxesFromDesignSpace(path):
...
-    yield p
+# @pytest.fixture
+# def invalid_designspace():
+#     p = "testCheck.designspace"
+#     __removeAxesFromDesignSpace(p)
+#     yield p
+
+
+# #@pytest.mark.xfail(reason="The check method requires MutatorMath")
+# def test_check(invalid_designspace, tmpdir):
```
anthrotype: hm, this test was silently broken for a while. It was marked as xfail so we didn't notice. But the reason it was failing was that it was the `invalid_designspace` fixture was attempting to read from an unexisting "testCheck.designspace" file. I don't understand the purpose of the test.
I would prefer we don't leave commented-out code rotting. Either we fix the test or we remove it altogether.

LettError: The test was for missing axes, it should be removed.

Doc/source/designspaceLib/scripting.rst
anthrotype: so shall we rename `foreground` to `public.default`?

Issue #1243, opened 2018-04-17
[designspaceLib] rules element enhancement


justvanrossum:
If we want to use the `<rule>` element to specify feature variations, we need a bit more than the current `<condition>` element to achieve the level of expressiveness that #1240's `addFeatureVariations()` offers.

https://github.com/fonttools/fonttools/tree/master/Doc/source/designspaceLib#50-rules-element

As it stands, the `<rule>` element contains one or more `<condition>` elements, followed by `<sub>` elements. In the context of #1240, this means only "rectangular" (*) regions can be specified. To specify more complex regions we would need to group conditions.

*) "rectangular" if we're talking about a two-dimensional design space. In reality there are N dimensions.

Current syntax:

```xml
<rules>
  <rule name="named.rule.1”>
    <condition minimum="155" maximum="190" name="weight" />
    <condition minimum="50" maximum="100" name="width" />
    <sub name="dollar" byname="dollar.alt”/>
  </rule>
</rules>
```

Proposed new syntax:

```xml
<rules>
  <rule name="named.rule.1”>
    <conditionset>
      <condition minimum="155" maximum="190" name="weight" />
      <condition minimum="50" maximum="100" name="width" />
    </conditionset>
    <conditionset>
      <condition ... />
      <condition ... />
    </conditionset>
    <sub name="dollar" byname="dollar.alt”/>
  </rule>
</rules>
```
The `<condition>` elements within a `<conditionset>` element form an `AND` chain: all conditions must be true to make the `<conditionset>` true. Only one `<conditionset>` needs to be true to trigger the rule, so the set of `<conditionset>` elements together form an `OR` chain. (Apologies if I'm using odd terminology, I hope the intent is clear.)

A `<conditionset>` would correspond to a `Space` in `featureVar.py` terminology, and a set of `<conditionset>` elements would correspond to a `Region`:
https://github.com/fonttools/fonttools/blob/0f1c6b3cf4465a5cf47c959ec5a24e28c3bb352f/Lib/fontTools/varLib/featureVars.py#L94-L102

I'm not sure to what extent the `<rule>` element is already in active use, but I can imagine (keeping) allowing "bare" `<condition>` elements, which together would imply a single `<conditionset>`. Perhaps that's even desirable apart from bw-compat reasons, as it keeps the simple case simple.

LettError:
[...]

Issue #1238, opened 2018-04-14
[designspaceLib] "layer" attribute to sourcedescriptor (edited)


LettError:
Following up on the pre-robothon 2018 talk for `support` layers in designspaces. 

We discussed a couple of steps to enable supporting outlines to specific glyphs. The use case would be a variable font can have `n` masters for most glyphs, but one or more need some extra help and they can have `n+i` additional shapes. If you

* A new `support` attribute in the source element indicates the name of the UFO3 layer the glyph geometry is coming from. (does it need a naming scheme or convention?)
* **The layer does not need to have a complete glyphset, only the glyphs that are present in the layer will be processed.**
* The layer does not need to be in a specific UFO - it can be in the default UFO, but can be any ufo really.
* The `designspace` consumer should be able to handle a varying number of masters in each glyph.

To do:
1. I guess this requires a change to the specification:
 https://github.com/fonttools/fonttools/blob/master/Doc/source/designspaceLib/readme.rst#attributes
2. Add support to `varlib` (what else `cu2qu` ? ) to support a varying number of masters per glyph. 

Kerning is specifically not a part of this proposal because it is difficult to store kerning tables with glyphs that are not present in the font.

anthrotype:
> does it need a naming scheme or convention?

no, I think the support layer name could be abritrary, anything suitable for identifying it among the other layers. What marks it as "support" is the fact that it's been used as such in the designspace.

> Add support to varlib (what else cu2qu ? ) to support a varying number of masters per glyph

yes. Plus, in ufo2ft compileTTF and compileOTF function one should also be able to specify from which layer to get the glyph set (currently it only compiles the glyphs from a font's default layer); and also it should skip generating OTL tables for these "incomplete" support master ttfs.

anthrotype:
For those who were not there while we discussed this, the idea is that one can add additional `<source>` elements which may or may not reference the same `filename` UFO as other sources, and which have an additional `support="some ufo3 layer"` attribute which marks these as special masters containing support outlines only for specific glyphs.
The authoring tools should include these support layers to obtain the list of masters for each glyph (matched between different fonts and layers by their name).

anthrotype:
yes, actually that makes sense. I don't see why only the extra "support" layers should be allowed to have sparse glyph sets, any source (besides the base master) could be like that.
I also think the name "layer" is clearer than "support".

justvanrossum:
> except in the neutral or base master's default layer

[...]

PR #1254, opened 2018-05-01
designspaceLib: This adds conditionsets to the rule element.

LettError:
See https://github.com/fonttools/fonttools/issues/1243

### DIFF
--- a/Doc/source/designspaceLib/readme.rst
+++ b/Doc/source/designspaceLib/readme.rst
@@ -83,21 +83,18 @@ dictionary, ``obj.stylename`` and ``obj.localisedStyleName['en']``.
 Rules
 *****
 
-**The ``rule`` element is experimental.** Some ideas behind how rules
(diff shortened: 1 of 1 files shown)


behdad:
lgtm.

anthrotype:
this new `conditionset` elements are another argument for bumping the designspace format version.
https://github.com/fonttools/fonttools/issues/1238#issuecomment-386243544

Doc/source/designspaceLib/readme.rst
```diff
@@ -312,16 +309,16 @@ RuleDescriptor object
 
 -  ``name``: string. Unique name for this rule. Will be used to
```
anthrotype: here we should mention that the rule "name" is optional: 

https://github.com/fonttools/fonttools/issues/1243#issuecomment-385252995

> The name attribute of the rule element should be optional. It's not needed when compiling a font.

PR #1255, opened 2018-05-03
WIP: implement conditionSets in designspaceLib.RuleDescriptor 

anthrotype:
this is incomplete, I push it here so we can work together with @LettError 

the reader still needs to be updated to support the new conditionset element, while also allow to read previous format (condition elements without a conditionset).

### DIFF
--- a/Doc/source/designspaceLib/readme.rst
+++ b/Doc/source/designspaceLib/readme.rst
@@ -314,6 +314,8 @@ RuleDescriptor object
 -  ``conditionSets``: a list of conditionsets
 -  Each conditionset is a list of conditions.
 -  Each condition is a dict with ``name``, ``minimum`` and ``maximum`` keys.
+-  ``subs``: list of substitutions
+-  Each substitution is stored as tuples of glyphnames, e.g. ("a", "a.alt").
 
 .. code:: python
 
@@ -321,6 +323,7 @@ RuleDescriptor object
     r1.name = "unique.rule.name"
     r1.conditionsSets.append([dict(name="weight", minimum=-10, maximum=10), dict(...)])
     r1.conditionsSets.append([dict(...), dict(...)])
+    r1.subs.append(("a", "a.alt"))
 
 .. _subclassing-descriptors:
 
--- a/Doc/source/designspaceLib/readme.rst
+++ b/Doc/source/designspaceLib/readme.rst
@@ -863,7 +863,7 @@ Attributes
 ----------
 
 -  ``name``: optional, string. A unique name that can be used to
-   identify this rule if it needs to be referenced elsewhere. bThe name
+   identify this rule if it needs to be referenced elsewhere. The name
    is not important for compiling variable fonts.
 
 5.1.1 conditionset element
--- a/Doc/source/designspaceLib/readme.rst
+++ b/Doc/source/designspaceLib/readme.rst
@@ -84,17 +84,12 @@ Rules
 *****
 
 Rules describe designspace areas in which one glyph should be replaced by another.
-A rule has a name and a number of conditionsets. The rule also contains a list of glyphname
-pairs: the glyphs that need to be substituted. For a rule to be triggered
-only one of the conditionsets needs to be true, ``OR``. Within a conditionset all
-conditions need to be true, ``AND``.
+A rule has a name and a number of conditionsets. The rule also contains a list of
+glyphname pairs: the glyphs that need to be substituted. For a rule to be triggered
+**only one** of the conditionsets needs to be true, ``OR``. Within a conditionset 
+**all** conditions need to be true, ``AND``.
 
-Variable fonts
-=======================
-
--  In an variable font the substitution happens at run time: there are
-   no changes in the font, only in the sequence of glyphnames that is
-   rendered.
+The ``sub`` element contains a pair of glyphnames. The ``name`` attribute is the glyph that should be visible when the rule evaluates to **False**. The ``with`` attribute is the glyph that should be visible when the rule evaluates to **True**.
 
 UFO instances
 =============
@@ -547,7 +542,7 @@ Attributes
    to the root path of this document. The path can be at the same level
    as the document or lower.
 -  ``layer``: optional, string. The name of the layer in the source file.
-   If no layer attribute is given assume it is the foreground layer.
+   If no layer attribute is given assume the foreground layer should be used.
 
 .. 31-lib-element:
 
@@ -856,8 +851,11 @@ Example
 -  Defines a named rule.
 -  Each ``rule`` element contains one or more ``conditionset`` elements.
 -  Only one ``conditionset`` needs to be true to trigger the rule.
--  All conditions must be true to make the ``conditionset`` true.
--  For backwards compatibility a ``rule`` can contain ``condition`` elements outside of a conditionset. These are then understood to be part of a single, implied, ``conditionset``.
+-  All conditions in a ``conditionset`` must be true to make the ``conditionset`` true.
+-  For backwards compatibility a ``rule`` can contain ``condition`` elements outside of a conditionset. These are then understood to be part of a single, implied, ``conditionset``. Note: these conditions should be written wrapped in a conditionset.
+-  A rule element needs to contain one or more ``sub`` elements in order to be compiled to a variable font.
+-  Rules without sub elements should be ignored when compiling a font.
+-  For authoring tools it might be necessary to save designspace files without ``sub`` elements just because the work is incomplete.
 
 .. attributes-11:
 
@@ -865,7 +863,8 @@ Attributes
 ----------
 
 -  ``name``: optional, string. A unique name that can be used to
-   identify this rule if it needs to be referenced elsewhere.
+   identify this rule if it needs to be referenced elsewhere.The name
+   is not important for compiling variable fonts.
 
 5.1.1 conditionset element
 =======================
@@ -879,7 +878,7 @@ Attributes
 =======================
 
 -  Child element of ``conditionset``
--  Between the ``minimum`` and ``maximum`` this rule is ``true``.
+-  Between the ``minimum`` and ``maximum`` this rule is ``True``.
 -  If ``minimum`` is not available, assume it is ``axis.minimum``.
 -  If ``maximum`` is not available, assume it is ``axis.maximum``.
 -  The condition must contain at least a minimum or maximum or both.
@@ -900,9 +899,7 @@ Attributes
 =================
 
 -  Child element of ``rule``.
--  Defines which glyphs to replace when the rule is true.
--  This element is optional. It may be useful for editors to know which
-   glyphs can be used to preview the axis.
+-  Defines which glyph to replace when the rule evaluates to **True**.
 
 .. attributes-13:
 
@@ -911,8 +908,8 @@ Attributes
 
 -  ``name``: string, required. The name of the glyph this rule looks
    for.
--  ``byname``: string, required. The name of the glyph it is replaced
-   with.
+-  ``with``: string, required. The name of the glyph it is replaced
+   with. Older versions of designspaceLib used the ``byname`` attribute. 
 
 .. example-7:
 
@@ -928,7 +925,7 @@ contained in a conditionset.
         <rule name="named.rule.1">
             <condition minimum="250" maximum="750" name="weight" />
             <condition minimum="50" maximum="100" name="width" />
-            <sub name="dollar" byname="dollar.alt"/>
+            <sub name="dollar" with="dollar.alt"/>
         </rule>
     </rules>
 
@@ -946,7 +943,7 @@ Example with ``conditionsets``. All conditions in a conditionset must be true.
               <condition ... />
               <condition ... />
             </conditionset>
-            <sub name="dollar" byname="dollar.alt"/>
+            <sub name="dollar" with="dollar.alt"/>
         </rule>
     </rules>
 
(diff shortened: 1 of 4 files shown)


Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -160,45 +160,51 @@ def __init__(self):
...
+
+
+def evaluateConditions(conditions, location):
+    """ Return True if all the conditions matches the given location.
+    If a condition has no minimum, check for < maximum.
+    If a condition has no maximum, check for > minimum.
+    """
+    for cd in conditions:
+        if cd['name'] not in location:
             continue
```
anthrotype: shouldn't this case -- when the requested location does not specify a value for an axis used inside a condition -- be documented? Is it like the location does not represent a single point, but a subspace (rectangle, box, etc.)?

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -434,27 +440,37 @@ def _addRule(self, ruleObject):
...
+                    conditionElement.attrib['maximum'] = self.intOrFloat(cond.get('maximum'))
+                conditionsetElement.append(conditionElement)
+            if len(conditionsetElement):
+                ruleElement.append(conditionsetElement)
+        # XXX shouldn't we require at least one sub element?
+        # if not ruleObject.subs:
+        #     raise DesignSpaceDocument('Invalid empty rule with no "sub" elements')
         for sub in ruleObject.subs:
             # skip empty subs
+            # XXX Hm. What does an empty glyphname string even mean?
```
LettError: I tend to be forgiving for incomplete data because an authoring tool should be able to save whenever it wants. But that might also be motivated by laziness.

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -434,27 +440,37 @@ def _addRule(self, ruleObject):
...
+                if cond.get('minimum') is not None:
+                    conditionElement.attrib['minimum'] = self.intOrFloat(cond.get('minimum'))
+                if cond.get('maximum') is not None:
+                    conditionElement.attrib['maximum'] = self.intOrFloat(cond.get('maximum'))
+                conditionsetElement.append(conditionElement)
+            if len(conditionsetElement):
+                ruleElement.append(conditionsetElement)
+        # XXX shouldn't we require at least one sub element?
+        # if not ruleObject.subs:
+        #     raise DesignSpaceDocument('Invalid empty rule with no "sub" elements')
```
anthrotype: What's the use of a rule that doesn't specify any substitutions? Is it the equivalent of a "ignore sub" in feature files?

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -702,7 +718,7 @@ def readRules(self):
                 ruleObject.conditions.append(cd)
             for subElement in ruleElement.findall('.sub'):
                 a = subElement.attrib['name']
-                b = subElement.attrib['with']
+                b = subElement.attrib['byname']
```
LettError: Where did the `with` come from - it used to be `byname`?

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -434,27 +440,37 @@ def _addRule(self, ruleObject):
...
+                conditionsetElement.append(conditionElement)
+            if len(conditionsetElement):
+                ruleElement.append(conditionsetElement)
+        # XXX shouldn't we require at least one sub element?
+        # if not ruleObject.subs:
+        #     raise DesignSpaceDocument('Invalid empty rule with no "sub" elements')
         for sub in ruleObject.subs:
             # skip empty subs
+            # XXX Hm. What does an empty glyphname string even mean?
             if sub[0] == '' and sub[1] == '':
```
anthrotype: What is the meaning of substituting nothing with something, or something with nothing?
I think we should either remove this `if` statement altogether, or assert that the sub's glyph names are not empty string.

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -160,45 +160,51 @@ def __init__(self):
 class RuleDescriptor(SimpleDescriptor):
     """<!-- optional: list of substitution rules -->
     <rules>
-        <rule name="vertical.bars" enabled="true">
```
LettError: Can we have an optional `enabled` in the rule element?

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -434,27 +440,37 @@ def _addRule(self, ruleObject):
...
+        #     raise DesignSpaceDocument('Invalid empty rule with no "sub" elements')
         for sub in ruleObject.subs:
             # skip empty subs
+            # XXX Hm. What does an empty glyphname string even mean?
             if sub[0] == '' and sub[1] == '':
                 continue
             subElement = ET.Element('sub')
             subElement.attrib['name'] = sub[0]
-            subElement.attrib['with'] = sub[1]
+            subElement.attrib['byname'] = sub[1]
```
anthrotype: note that changing the attribute name from `with` to `byname` will mean that the latter will appear before the `nama` attribute, since ElementTree sorts the attributes alphabetically when serializing XML. The fonttools xmlWriter allows to keep the order, but has a completely different interface...

Lib/fontTools/designspaceLib/__init__.py
anthrotype: no idea.. I found it there

Lib/fontTools/designspaceLib/__init__.py
anthrotype: sure. I removed it from the example because it was not implemented nor documented

Lib/fontTools/designspaceLib/__init__.py
LettError: A location that does not have a value for a specific axis should be assumed to have the default value for that axis. This seemed easier than to assume all locations always have all dimensions.

Lib/fontTools/designspaceLib/__init__.py
LettError: No, it does not mean `ignore sub`. Would that be useful here?

Lib/fontTools/designspaceLib/__init__.py
LettError: Fine, remove it :)

Lib/fontTools/designspaceLib/__init__.py
anthrotype: > should be assumed to have the default value for that axis

in that case we would need access to the axes in this evaluateRule function.
The current code would return True if location does not have explicit values for all the condition axes

Lib/fontTools/designspaceLib/__init__.py
LettError: I'm just not sure when it changed from `with` to `byname`. It can be either.

Lib/fontTools/designspaceLib/__init__.py
justvanrossum: Nah. An empty subs attr _could_ mean: "this is work in progress, I will add substitutions later"

Lib/fontTools/designspaceLib/__init__.py
anthrotype: i'm tempted to revert to "with", so that it's sorted logically `<sub name="a" with="a.alt" />`

LettError: ok

Lib/fontTools/designspaceLib/__init__.py
anthrotype: ok fine, will remove the `# XXX` lines

anthrotype:
i'm off to lunch, feel free to push directly to the branch.
I may have some time to work on this later in the afternoon

LettError:
My code does not go beyond anything described here. I will remove my changes.

LettError:
Have a nice lunch! Not sure I see where to find your changes, but I guess it will sort itself.

anthrotype:
> Not sure I see where to find your changes

This PR branch is here on the upstream repository (not my fork), it's named `designspaceLib-conditionsets`.
Just pull or fetch to sync your local copy, and then `git checkout designspaceLib-conditionsets` to create a local branch tracking the remote one.

anthrotype:
I don't know about sourcetree but, assuming `origin` is the remote name for the fonttools/fonttools repo, you should be able to run these two commands from the Terminal:

```
$ git fetch origin
[...]
$ git checkout designspaceLib-conditionsets
Branch 'designspaceLib-conditionsets' set up to track remote branch 'designspaceLib-conditionsets' from 'origin'.
Switched to a new branch 'designspaceLib-conditionsets'
```

The first command downloads all the objects from the upstream repository. `origin` is the default remote name, but it could also be named differently, e.g. if you originally cloned your own fork and then subsequently added another remote tracking the upstream repo with a different name.

The second command checks out the branch: if it doesn't already exists, it creates one with that name, and makes it track the one with the same name from origin.

Lib/fontTools/designspaceLib/__init__.py
anthrotype: in https://github.com/fonttools/fonttools/blob/master/Doc/source/designspaceLib/readme.rst#513-sub-element, it says about `sub` that

> This element is optional. It may be useful for editors to know which glyphs can be used to preview the axis.

I understand that one may want to write incomplete data, and that a rule without substitutions could simply be ignored when the designspace document is processed by a UFO or varfont builder. However, stressing the fact that `sub` element is "optional" seems counter-intuitive to me.
The sub element is not only "useful for editors to know which glyphs to preview".

justvanrossum: It's possibly a remnant of what @LettError originally thought the `<rule>` element was going to be for? Definitely needs rephrasing.

anthrotype:
FYI I need to cut a release now, so this will go to the next version.

Lib/fontTools/designspaceLib/__init__.py
LettError: Yes, it should be rephrased. Or maybe it is enough to remove that line.

LettError:
Thanks @anthrotype, I found it. I committed some small things to see if it makes a difference.

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -718,7 +717,10 @@ def readRules(self):
                 ruleObject.conditions.append(cd)
             for subElement in ruleElement.findall('.sub'):
                 a = subElement.attrib['name']
-                b = subElement.attrib['byname']
+                if self.toolVersion < 4:
```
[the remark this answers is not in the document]
anthrotype: no problem!
the issue of "byname" vs "with" is a bit complicated, because the spec was saying on thing ("byname") whereas the only implementation (designSpaceDocument == fontTools.designspaceLib) was actually writing something else ("with").
So I changed the implementation to match the spec... So which one was "right" then?

Also, I wonder, are there actual designspaces out there that we wish to support that use either "with" or "byname"? Since the feature was marked as experimental, I think we don't need to be concerned with changing it as we see fit.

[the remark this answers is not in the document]
LettError: So you'd rather not check for `byname`?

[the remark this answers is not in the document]
anthrotype: yeah, that's what I was suggesting. Let's just read/write using `width` and change the spec accordingly.
The change in the designspace document `format` attribute (from an integer to a `MAJOR.MINOR` version string) in itself warrants a version bump to 4.0, since it would break compatibility with readers that assumed format was an integer.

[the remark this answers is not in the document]
anthrotype: I also think format should be stored as a string, it's not really a float.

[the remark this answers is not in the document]
LettError: Ok. I have some changes here. Is there a way I can show you what I have?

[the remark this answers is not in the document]
LettError: I pushed these changes.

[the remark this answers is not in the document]
anthrotype: where did you push them? I see you created a new branch "designspacelib-conditionset-spec" then deleted it. It's ok if you push directly to this PR branch, it's marked as WIP.

Lib/fontTools/designspaceLib/__init__.py
LettError: Here are some changes to the designspaceLib/readme.rst https://github.com/fonttools/fonttools/commit/d26bb392512b54444ca879b0d09c8f15333b78e6
These are the `conditionsets`, `conditions` and `sub` element changes discussed here.

LettError: https://github.com/fonttools/fonttools/commit/d26bb392512b54444ca879b0d09c8f15333b78e6

Doc/source/designspaceLib/readme.rst
```diff
@@ -547,7 +542,7 @@ Attributes
    to the root path of this document. The path can be at the same level
    as the document or lower.
 -  ``layer``: optional, string. The name of the layer in the source file.
-   If no layer attribute is given assume it is the foreground layer.
+   If no layer attribute is given assume the foreground layer should be used.
```
anthrotype: actually, the default layer is not necessarily the "foreground" one, see @typesupply 's reply:
https://github.com/unified-font-object/ufo-spec/issues/38#issuecomment-388103944

so if no layer attribute is given, we should use whatever is the current default layer in the UFO (i.e. the one pointing to the default "glyphs" directory). The use of "foreground" here may be confusing.

PR #1261, opened 2018-05-14
Add ttx tests

chrissimpkins:
This PR increases the test coverage of the source in the ttx module from 48% to 94%.  

It includes tests for the following:

- `ttx.Options` class
- `ttx.ttCompile` function
- `ttx.ttList` function
- `ttx.main` function

After discussion with @anthrotype, I added zopfli as a tox testing dependency (in the tox.ini settings file https://github.com/fonttools/fonttools/commit/cf5b31e9c66751c86084cb6604bbe9e156209717) to test ttx --> woff compiles using zopfli.  The requirements.txt file was not modified.

While the lines in ttCompile are executed, I did not add tests across all combinations of possible requests for the following:

https://github.com/fonttools/fonttools/blob/ceb41ec484b3aed5127962a3e2c2030a56d704e1/Lib/fontTools/ttx.py#L280-L283

Two minor issues were found during the testing:

1) An unhandled ValueError is raised on line https://github.com/fonttools/fonttools/blob/ceb41ec484b3aed5127962a3e2c2030a56d704e1/Lib/fontTools/ttx.py#L181 when a non-integer value is entered on the command line (e.g., typo).  You can replicate it with:

```
$ ttx -y A
```

Propose modification of the block to the following to address it:

```python
            elif option == "-y":
                try:
                    self.fontNumber = int(value)
                except ValueError:
                    raise getopt.GetoptError("%s is not a valid argument for -y.  Please enter an integer value." % (value) )
```

**_#1 Resolved_**: As per https://github.com/fonttools/fonttools/pull/1261#issuecomment-389088705, will not modify the source to address this issue.

2) The `sys.exit(2)` in https://github.com/fonttools/fonttools/blob/ceb41ec484b3aed5127962a3e2c2030a56d704e1/Lib/fontTools/ttx.py#L221 looks like it is unreachable after the GetoptError is raised.

Happy to submit changes to address the above if you agree and are willing to accept fixes for them.

### DIFF
--- a/Tests/ttx/ttx_test.py
+++ b/Tests/ttx/ttx_test.py
@@ -514,13 +514,14 @@ def test_options_woff2_and_zopfli_shouldfail():
 
 def test_ttcompile_otf_compile_default(tmpdir):
     inttx = os.path.join("Tests", "ttx", "data", "TestOTF.ttx")
-    outotf = os.path.join(str(tmpdir), "TestOTF.otf")
+    # outotf = os.path.join(str(tmpdir), "TestOTF.otf")
+    outotf = tmpdir.join("TestOTF.ttx")
     default_options = ttx.Options([], 1)
-    ttx.ttCompile(inttx, outotf, default_options)
+    ttx.ttCompile(inttx, str(outotf), default_options)
     # confirm that font was built
-    assert os.path.isfile(outotf)
+    assert outotf.check(file=True)
     # confirm that it is valid OTF file, can instantiate a TTFont, has expected OpenType tables
-    ttf = TTFont(outotf)
+    ttf = TTFont(str(outotf))
     expected_tables = (
         "head",
         "hhea",
@@ -539,14 +540,14 @@ def test_ttcompile_otf_compile_default(tmpdir):
 
 def test_ttcompile_otf_to_woff_without_zopfli(tmpdir):
     inttx = os.path.join("Tests", "ttx", "data", "TestOTF.ttx")
-    outwoff = os.path.join(str(tmpdir), "TestOTF.woff")
+    outwoff = tmpdir.join("TestOTF.woff")
     options = ttx.Options([], 1)
     options.flavor = "woff"
-    ttx.ttCompile(inttx, outwoff, options)
+    ttx.ttCompile(inttx, str(outwoff), options)
     # confirm that font was built
-    assert os.path.isfile(outwoff)
+    assert outwoff.check(file=True)
     # confirm that it is valid TTF file, can instantiate a TTFont, has expected OpenType tables
-    ttf = TTFont(outwoff)
+    ttf = TTFont(str(outwoff))
     expected_tables = (
         "head",
         "hhea",
@@ -565,15 +566,15 @@ def test_ttcompile_otf_to_woff_without_zopfli(tmpdir):
 
 def test_ttcompile_otf_to_woff_with_zopfli(tmpdir):
     inttx = os.path.join("Tests", "ttx", "data", "TestOTF.ttx")
-    outwoff = os.path.join(str(tmpdir), "TestOTF.woff")
+    outwoff = tmpdir.join("TestOTF.woff")
     options = ttx.Options([], 1)
     options.flavor = "woff"
     options.useZopfli = True
-    ttx.ttCompile(inttx, outwoff, options)
+    ttx.ttCompile(inttx, str(outwoff), options)
     # confirm that font was built
-    assert os.path.isfile(outwoff)
+    assert outwoff.check(file=True)
     # confirm that it is valid TTF file, can instantiate a TTFont, has expected OpenType tables
-    ttf = TTFont(outwoff)
+    ttf = TTFont(str(outwoff))
     expected_tables = (
         "head",
         "hhea",
@@ -592,14 +593,14 @@ def test_ttcompile_otf_to_woff_with_zopfli(tmpdir):
 
 def test_ttcompile_otf_to_woff2(tmpdir):
     inttx = os.path.join("Tests", "ttx", "data", "TestOTF.ttx")
-    outwoff2 = os.path.join(str(tmpdir), "TestTTF.woff2")
+    outwoff2 = tmpdir.join("TestTTF.woff2")
     options = ttx.Options([], 1)
     options.flavor = "woff2"
-    ttx.ttCompile(inttx, outwoff2, options)
+    ttx.ttCompile(inttx, str(outwoff2), options)
     # confirm that font was built
-    assert os.path.isfile(outwoff2)
+    assert outwoff2.check(file=True)
     # confirm that it is valid TTF file, can instantiate a TTFont, has expected OpenType tables
-    ttf = TTFont(outwoff2)
+    ttf = TTFont(str(outwoff2))
     # DSIG should not be included from original ttx as per woff2 spec (https://dev.w3.org/webfonts/WOFF2/spec/)
     assert "DSIG" not in ttf
     expected_tables = (
@@ -619,13 +620,13 @@ def test_ttcompile_otf_to_woff2(tmpdir):
 
 def test_ttcompile_ttf_compile_default(tmpdir):
     inttx = os.path.join("Tests", "ttx", "data", "TestTTF.ttx")
-    outttf = os.path.join(str(tmpdir), "TestTTF.ttf")
+    outttf = tmpdir.join("TestTTF.ttf")
     default_options = ttx.Options([], 1)
-    ttx.ttCompile(inttx, outttf, default_options)
+    ttx.ttCompile(inttx, str(outttf), default_options)
     # confirm that font was built
-    assert os.path.isfile(outttf)
+    assert outttf.check(file=True)
     # confirm that it is valid TTF file, can instantiate a TTFont, has expected OpenType tables
-    ttf = TTFont(outttf)
+    ttf = TTFont(str(outttf))
     expected_tables = (
         "head",
         "hhea",
@@ -649,14 +650,14 @@ def test_ttcompile_ttf_compile_default(tmpdir):
 
 def test_ttcompile_ttf_to_woff_without_zopfli(tmpdir):
     inttx = os.path.join("Tests", "ttx", "data", "TestTTF.ttx")
-    outwoff = os.path.join(str(tmpdir), "TestTTF.woff")
+    outwoff = tmpdir.join("TestTTF.woff")
     options = ttx.Options([], 1)
     options.flavor = "woff"
-    ttx.ttCompile(inttx, outwoff, options)
+    ttx.ttCompile(inttx, str(outwoff), options)
     # confirm that font was built
-    assert os.path.isfile(outwoff)
+    assert outwoff.check(file=True)
     # confirm that it is valid TTF file, can instantiate a TTFont, has expected OpenType tables
-    ttf = TTFont(outwoff)
+    ttf = TTFont(str(outwoff))
     expected_tables = (
         "head",
         "hhea",
@@ -680,15 +681,15 @@ def test_ttcompile_ttf_to_woff_without_zopfli(tmpdir):
 
 def test_ttcompile_ttf_to_woff_with_zopfli(tmpdir):
     inttx = os.path.join("Tests", "ttx", "data", "TestTTF.ttx")
-    outwoff = os.path.join(str(tmpdir), "TestTTF.woff")
+    outwoff = tmpdir.join("TestTTF.woff")
     options = ttx.Options([], 1)
     options.flavor = "woff"
     options.useZopfli = True
-    ttx.ttCompile(inttx, outwoff, options)
+    ttx.ttCompile(inttx, str(outwoff), options)
     # confirm that font was built
-    assert os.path.isfile(outwoff)
+    assert outwoff.check(file=True)
     # confirm that it is valid TTF file, can instantiate a TTFont, has expected OpenType tables
-    ttf = TTFont(outwoff)
+    ttf = TTFont(str(outwoff))
     expected_tables = (
         "head",
         "hhea",
@@ -712,14 +713,14 @@ def test_ttcompile_ttf_to_woff_with_zopfli(tmpdir):
 
 def test_ttcompile_ttf_to_woff2(tmpdir):
     inttx = os.path.join("Tests", "ttx", "data", "TestTTF.ttx")
-    outwoff2 = os.path.join(str(tmpdir), "TestTTF.woff2")
+    outwoff2 = tmpdir.join("TestTTF.woff2")
     options = ttx.Options([], 1)
     options.flavor = "woff2"
-    ttx.ttCompile(inttx, outwoff2, options)
+    ttx.ttCompile(inttx, str(outwoff2), options)
     # confirm that font was built
-    assert os.path.isfile(outwoff2)
+    assert outwoff2.check(file=True)
     # confirm that it is valid TTF file, can instantiate a TTFont, has expected OpenType tables
-    ttf = TTFont(outwoff2)
+    ttf = TTFont(str(outwoff2))
     # DSIG should not be included from original ttx as per woff2 spec (https://dev.w3.org/webfonts/WOFF2/spec/)
     assert "DSIG" not in ttf
     expected_tables = (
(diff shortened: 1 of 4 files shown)


miguelsousa:
For 1., I'd use `try/except ValueError` instead of `if/else`.

AppVeyor is stuck on *ttx_test.py* which suggests that one of the tests is failing on Windows.

chrissimpkins:
Thanks Miguel.  I updated the OP with your suggestion about the try/except source rather than if / else.

I am raising an exception that is handled in this block:

https://github.com/fonttools/fonttools/blob/ceb41ec484b3aed5127962a3e2c2030a56d704e1/Lib/fontTools/ttx.py#L403-L406

I suspect that may be causing the problem.  Will take a look this evening.

chrissimpkins:
@miguelsousa do you have access to Appveyor to cancel that last build?

miguelsousa:
@chrissimpkins I don't

chrissimpkins:
@miguelsousa OK thanks.  Let me see if I can test it locally on a Win VM.  I haven't looked through the Appveyor settings file.  May need to wait out the 60 min per interpreter/architecture timeout to test again through Appveyor tonight.

chrissimpkins:
Windows doesn't seem to like a couple of the mocked Exceptions that are raised with pytest.  I made two of the test functions non-Win only tests in https://github.com/fonttools/fonttools/pull/1261/commits/189e2487fe59ca4b39b7b5a8278b134f2315b53a.  This *should* pass when Appveyor clears the backlog through timeouts tonight...  All tests execute and pass locally on Win 10 x Python 3.6 with these changes.

chrissimpkins:
converted to pytest skipif fixtures for Windows test skips x 2 test functions in https://github.com/fonttools/fonttools/pull/1261/commits/bb4753acd45fdbbafc27b219f1ddd992b7ee553d

Tests/ttx/ttx_test.py
```diff
@@ -263,5 +267,569 @@ def test_guessFileType_invalid_path(self):
...
+
+# ----------------------------
+# ttx.ttCompile function tests
+# ----------------------------
+
+
+def test_ttx_ttcompile_otf_compile_default(tmpdir):
+    inttx = os.path.join("Tests", "ttx", "data", "TestOTF.ttx")
+    outotf = os.path.join(str(tmpdir), "TestOTF.otf")
+    default_options = ttx.Options([("", "")], 1)
```
anthrotype: for default options, just pass an empty list as first argument to Options class

Tests/ttx/ttx_test.py
```diff
@@ -263,5 +267,569 @@ def test_guessFileType_invalid_path(self):
...
+    ttx.ttCompile(inttx, outotf, default_options)
+    # confirm that font was built
+    assert os.path.isfile(outotf) is True
```
anthrotype: remove `is True`

Tests/ttx/ttx_test.py
```diff
@@ -263,5 +267,569 @@ def test_guessFileType_invalid_path(self):
...
+def test_ttx_ttcompile_ttf_compile_default(tmpdir):
+    inttx = os.path.join("Tests", "ttx", "data", "TestTTF.ttx")
+    outttf = os.path.join(str(tmpdir), "TestTTF.ttf")
+    default_options = ttx.Options([("", "")], 1)
+    ttx.ttCompile(inttx, outttf, default_options)
+    # confirm that font was built
+    assert os.path.isfile(outttf) is True
+    # confirm that it is valid TTF file, can instantiate a TTFont, has expected OpenType tables
+    ttf = TTFont(outttf)
+    expected_tables = ("head", "hhea", "maxp", "OS/2", "name", "cmap", "hmtx", "fpgm", "prep", "cvt ", "loca", "glyf", "post", "gasp", "DSIG")
```
anthrotype: these lines are a bit too long, try to stick within 80 characters please

Tests/ttx/ttx_test.py
```diff
@@ -263,5 +267,569 @@ def test_guessFileType_invalid_path(self):
...
+    # confirm that font was built
+    assert os.path.isfile(outttf2) is True
+    # confirm that timestamp is more recent than modified time on ttx file
+    mtime = os.path.getmtime(inttx)
+    epochtime = timestampSinceEpoch(mtime)
+    ttf = TTFont(outttf2)
+    assert ttf['head'].modified > epochtime
+
+
+def test_ttx_ttcompile_otf_timestamp_calcs(tmpdir):
```
anthrotype: this and the previous test only differ by the input ttx file, you can use `@pytest.mark.parametrize` to run the same test function with different inputs, see pytest docs

Tests/ttx/ttx_test.py
```diff
@@ -263,5 +267,569 @@ def test_guessFileType_invalid_path(self):
...
+    assert "OS/2  0x67230FF8        96         0" in out
+
+# -------------------
+# main function tests
+# -------------------
+
+
+def test_ttx_main_default_ttf_dump_to_ttx(tmpdir):
+    inpath = os.path.join("Tests", "ttx", "data", "TestTTF.ttf")
+    outpath = os.path.join(str(tmpdir), "TestTTF.ttx")
```
anthrotype: `tmpdir` fixture returns a py.path.local object; to join it with a sub-path, you can use the division operator, e.g. `tmpdir / "TestTTF.ttx"`.
See the full set of features here: http://py.readthedocs.io/en/latest/path.html

Tests/ttx/ttx_test.py
```diff
@@ -263,5 +267,569 @@ def test_guessFileType_invalid_path(self):
...
+    args = ["-o", outpath, inpath]
+    ttx.main(args)
+    assert os.path.isfile(outpath)
```
anthrotype: http://py.readthedocs.io/en/latest/path.html#checking-path-types

Tests/ttx/ttx_test.py
```diff
@@ -263,5 +267,569 @@ def test_guessFileType_invalid_path(self):
...
+            args = ["-d", "bogusdir", inpath]
+            ttx.main(args)
+
+
+def test_ttx_main_keyboard_interrupt(tmpdir, monkeypatch, capsys):
+    with pytest.raises(SystemExit):
+        inpath = os.path.join("Tests", "ttx", "data", "TestTTF.ttx")
+        outpath = os.path.join(str(tmpdir), "TestTTF.ttf")
+        args = ["-o", outpath, inpath]
+        monkeypatch.setattr(ttx, 'process', (lambda x, y: raise_exception(KeyboardInterrupt)))
```
anthrotype: nice one!

Tests/ttx/ttx_test.py
```diff
@@ -263,5 +267,569 @@ def test_guessFileType_invalid_path(self):
...
+        ttx.main(args)
+
+    out, err = capsys.readouterr()
+    assert "(Cancelled.)" in err
+
+
+@pytest.mark.skipif(sys.platform == "win32",
+                    reason="waitForKeyPress function causes test to hang on Windows platform")
```
anthrotype: which is supposed to be a "feature" :)

tox.ini
```diff
@@ -11,6 +11,7 @@ basepython =
 deps =
     cov: coverage>=4.3
     pytest
+    zopfli
```
anthrotype: put zopfli in the `requirements.txt` and pin to the latest. Brotli is already in there. The requirements.txt is only used here to configure the test suite's environment in a reproducible way.

anthrotype: unlike pytest or coverage which are test-only requirements, zopfli is an (optional) installation requirement, ie. required at runtime if one wants to export woff with it.

anthrotype:
about the `-y ` option throwing ValueError if a non-number is passed in, I don't think we need to bother. The error message and the option help message are clear enough: we expect a number, period.

anthrotype:
You can remove the sys.exit that follows the raise GetOptError, you’re right it will never be reached.

chrissimpkins:
Thanks for taking the time to review it Cosimo.  I know that this PR seems very random and out of the blue for a mature body of source like ttx.  The goal is definitely not to scrutinize source contributed by others.  I have plenty of my own authored drivel that warrants that attention.  I am taking a deep dive in the code here to try to better understand it and, for me, that comes from writing tests.  I figured that it might be beneficial to contribute this back upstream in areas that lack test coverage.  If this is helpful that is great.  If it is not, that's OK too.  I have both the time and an interest in working on tests across a number of modules.  Consider me a resource if there are areas in the project where this is helpful. And please list these areas if you know where they are!  I do not.  I am relying on the coverage report to find problem spots.

Will respond to your review requests above.

 > about the -y option throwing ValueError if a non-number is passed in, I don't think we need to bother. The error message and the option help message are clear enough: we expect a number, period.

Damn users.  RTFD!  :)

> You can remove the sys.exit that follows the raise GetOptError, you’re right it will never be reached.

Sounds good.  Will make this change and include in this PR.

Tests/ttx/ttx_test.py
```diff
@@ -263,5 +267,569 @@ def test_guessFileType_invalid_path(self):
...
+    assert tto.useZopfli is False
+    assert tto.onlyTables == []
+    assert tto.skipTables == []
+    assert tto.fontNumber == -1
+    assert tto.logLevel == logging.INFO
+
+
+def test_ttx_options_class_flag_h(capsys):
+    with pytest.raises(SystemExit):
+        ttx.Options([("-h", None)], 1)
```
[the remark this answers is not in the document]
chrissimpkins: This test fails without the `pytest.raises` because there is a SystemExit raised in the  -h option block.  

```
____________________________ test_ttx_options_class_flag_h ____________________________

capsys = <_pytest.capture.CaptureFixture object at 0x10f6d2c88>

    def test_ttx_options_class_flag_h(capsys):
>       ttx.Options([("-h", None)], 1)

Tests/ttx/ttx_test.py:303:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <fontTools.ttx.Options object at 0x10f6d2358>, rawOptions = [('-h', None)]
numFiles = 1

    def __init__(self, rawOptions, numFiles):
        self.onlyTables = []
        self.skipTables = []
        self.fontNumber = -1
        for option, value in rawOptions:
            # general options
            if option == "-h":
                print(__doc__)
>               sys.exit(0)
E               SystemExit: 0

Lib/fontTools/ttx.py:138: SystemExit
```

Are you suggesting removal of the entire test for the print of the help text to stdout when the `-h` option is parsed?

Tests/ttx/ttx_test.py
```diff
@@ -263,5 +267,569 @@ def test_guessFileType_invalid_path(self):
...
+
+    out, err = capsys.readouterr()
+    assert "TTX -- From OpenType To XML And Back" in out
+
+
+def test_ttx_options_class_flag_version(capsys):
```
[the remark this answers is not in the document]
chrissimpkins: OK, will remove.

Tests/ttx/ttx_test.py
chrissimpkins: OK

tox.ini
chrissimpkins: OK, no problem. Will remove from tox.ini and add to requirements.txt

Tests/ttx/ttx_test.py
anthrotype: It’s part of pytest, you’re already using it!

chrissimpkins: oh... I didn't realize that.  ty :)

chrissimpkins:
> You can remove the sys.exit that follows the raise GetOptError, you’re right it will never be reached.

#1262

Tests/ttx/ttx_test.py
chrissimpkins: @anthrotype 

I modified one of the functions to use the division operator for path joins with the pytest tmpdir fixture as follows:

```python
def test_ttcompile_otf_compile_default(tmpdir):
    inttx = os.path.join("Tests", "ttx", "data", "TestOTF.ttx")
    # outotf = os.path.join(str(tmpdir), "TestOTF.otf")
    outotf = tmpdir/"TestOTF.otf"
    default_options = ttx.Options([], 1)
    ttx.ttCompile(inttx, outotf, default_options)
    # confirm that font was built
    assert outotf.check(file=True)
    # confirm that it is valid OTF file, can instantiate a TTFont, has expected OpenType tables
    ttf = TTFont(outotf)
    expected_tables = (
        "head",
        "hhea",
        "maxp",
        "OS/2",
        "name",
        "cmap",
        "post",
        "CFF ",
        "hmtx",
        "DSIG",
    )
    for table in expected_tables:
        assert table in ttf
```

and I am getting this error raised from ttFont:

```
============================================= FAILURES =============================================
________________________________ test_ttcompile_otf_compile_default ________________________________

tmpdir = local('/private/var/folders/x7/vs8mbwmx1hg1vcb53nlygt0m0000gn/T/pytest-of-chris/pytest-97/test_ttcompile_otf_compile_def0')

    def test_ttcompile_otf_compile_default(tmpdir):
        inttx = os.path.join("Tests", "ttx", "data", "TestOTF.ttx")
        # outotf = os.path.join(str(tmpdir), "TestOTF.otf")
        outotf = tmpdir/"TestOTF.otf"
        default_options = ttx.Options([], 1)
>       ttx.ttCompile(inttx, outotf, default_options)

Tests/ttx/ttx_test.py:520:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
Lib/fontTools/misc/loggingTools.py:372: in wrapper
    return func(*args, **kwds)
Lib/fontTools/ttx.py:290: in ttCompile
    ttf.save(output)
Lib/fontTools/ttLib/ttFont.py:189: in save
    file.write(tmp2.getvalue())
/usr/local/lib/python3.6/site-packages/py/_path/local.py:500: in write
    data = py.builtin._totext(data, sys.getdefaultencoding())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

obj = b'OTTO\x00\n\x00\x80\x00\x03\x00 CFF &\xfc\xc2\x10\x00\x00\x07\xf8\x00\x00\x00\xebDSIG\x00\x00\x00\x01\x00\x00\x08\xfc...x01\xf4\x002\x00\x00\x00\x00\x00\xfa\x00\x00\x00\xfa\x00\x00\x00\xf1\x007\x02\xd3\x007\x00\x00\x00\x01\x00\x00\x00\x00'
encoding = 'utf-8', errors = None

    def _totext(obj, encoding=None, errors=None):
        if isinstance(obj, bytes):
            if errors is None:
>               obj = obj.decode(encoding)
E               UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 7: invalid start byte

/usr/local/lib/python3.6/site-packages/py/_builtin.py:125: UnicodeDecodeError
===================================== short test summary info ======================================
FAIL Tests/ttx/ttx_test.py::test_ttcompile_otf_compile_default
=============================== 1 failed, 78 passed in 2.01 seconds ================================
```

Thoughts?

Tests/ttx/ttx_test.py
chrissimpkins: On hold until we resolve issue in https://github.com/fonttools/fonttools/pull/1261#discussion_r189875115

chrissimpkins:
How would you like me to address this? Leave as originally submitted with the string cast in the path join (works) or modify all of the functions to convert to the string cast in the TTFont instantiation?

anthrotype:
I would prefer if you use py.path.local methods instead of os.path when available, and convert to strings when we need to pass these paths to API that deal with strings.
But I don't want to insist too much. Do what you can or think is best, and let me know when this is ready to review. Thanks

chrissimpkins:
No worries.  Happy to change it.  I need to address this change and modify requirements.txt then will gtg based upon your review.  Should be able to get to this later today and will let you know when ready.

chrissimpkins:
Thanks for all of your help with it @anthrotype.  Appreciate your time!

PR #1262, opened 2018-05-20
Remove unreachable source in ttx module

chrissimpkins:
As discussed in #1261, this removes an unreachable `sys.exit` statement in `ttx.Options`


anthrotype:
you can do it in the same PR #1261

chrissimpkins:
OK

PR #1267, opened 2018-05-27
designspaceLib remove checkDefault, checkAxes

LettError:
* remove checkDefault, checkAxes and any kind of guessing about the axes. The default is the master located at the default value of all axes.
* add findDefault returns the sourceDescriptor for the default. 
* stray conditions are wrapped in a conditionset when reading.
* add some tests.

### DIFF
--- a/Lib/fontTools/designspaceLib/__init__.py
+++ b/Lib/fontTools/designspaceLib/__init__.py
@@ -178,16 +178,14 @@ class RuleDescriptor(SimpleDescriptor):
 
 
 def evaluateRule(rule, location):
-    """ Return True if any of the rule's conditionsets matches the
-    given location.
-    """
+    """ Return True if any of the rule's conditionsets matches the given location."""
     return any(evaluateConditions(c, location) for c in rule.conditionSets)
 
 
 def evaluateConditions(conditions, location):
     """ Return True if all the conditions matches the given location.
-    If a condition has no minimum, check for < maximum.
-    If a condition has no maximum, check for > minimum.
+        If a condition has no minimum, check for < maximum.
+        If a condition has no maximum, check for > minimum.
     """
     for cd in conditions:
         value = location[cd['name']]
@@ -885,49 +883,27 @@ class BaseDocReader(object):
         instanceObject.lib = from_plist(libElement[0])
 
     def readInfoElement(self, infoElement, instanceObject):
-        """ Read the info element.
-
-            ::
-
-                <info/>
-
-                Let's drop support for a different location for the info. Never needed it.
-
-            """
+        """ Read the info element."""
         infoLocation = self.locationFromElement(infoElement)
         instanceObject.info = True
 
     def readKerningElement(self, kerningElement, instanceObject):
-        """ Read the kerning element.
-
-        ::
-
-                Make kerning at the location and with the masters specified at the instance level.
-                <kerning/>
-
-        """
+        """ Read the kerning element."""
         kerningLocation = self.locationFromElement(kerningElement)
         instanceObject.addKerning(kerningLocation)
 
     def readGlyphElement(self, glyphElement, instanceObject):
         """
         Read the glyph element.
-
-        ::
-
             <glyph name="b" unicode="0x62"/>
-
             <glyph name="b"/>
-
             <glyph name="b">
                 <master location="location-token-bbb" source="master-token-aaa2"/>
                 <master glyphname="b.alt1" location="location-token-ccc" source="master-token-aaa3"/>
-
                 <note>
                     This is an instance from an anisotropic interpolation.
                 </note>
             </glyph>
-
         """
         glyphData = {}
         glyphName = glyphElement.attrib.get('name')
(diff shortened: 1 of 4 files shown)


LettError:
Sorry, some tests are failing.

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -677,56 +663,60 @@ def getSourcePaths(self, makeGlyphs=True, makeKerning=True, makeInfo=True):
...
+        # we also need to read any conditions that are outside of a condition set.
         rules = []
         for ruleElement in self.root.findall(".rules/rule"):
             ruleObject = self.ruleDescriptorClass()
             ruleObject.name = ruleElement.attrib.get("name")
+            # read any stray conditions outside a condition set
+            externalConditions = self._readConditionElements(ruleElement)
+            if externalConditions:
+                ruleObject.conditionSets.append(externalConditions)
+                warnings.warn('Found stray rule conditions outside a conditionset. Wrapped them in a new conditionset.')
```
anthrotype: instead of `warnings.warn` (which is more intended for orther developers using a library from their applications), I'd prefer a `self.logger.warning` kind of message here. We use logging elsewhere for messages directed to end users, and when the message is about the data rather than the code itself.

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -677,56 +663,60 @@ def getSourcePaths(self, makeGlyphs=True, makeKerning=True, makeInfo=True):
...
+                # will allow these to be None, assume axis.maximum
+                cd['maximum'] = None
+            cd['name'] = conditionElement.attrib.get("name")
+            # # test for things
+            # if cd.get('minimum') is None and cd.get('maximum') is None:
+            #     if ruleObject.name is not None:
+            #         n = ruleObject.name
+            #     else:
+            #         n = "%d" % len(rules)
+            #     raise DesignSpaceDocumentError("No minimum or maximum defined in rule \"%s\"." % n)
```
anthrotype: why this is commented out? it's kind of invalid, no?

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -870,14 +801,14 @@ def locationFromElement(self, element):
...
         loc = {}
         for dimensionElement in locationElement.findall(".dimension"):
             dimName = dimensionElement.attrib.get("name")
             if self._strictAxisNames and dimName not in self.axisDefaults:
-                # In case the document contains axis definitions,
-                # then we should only read the axes we know about.
-                # However, if the document does not contain axes,
-                # then we need to create them after reading.
+                # In case the document contains no axis definitions,
+                warnings.warn("Location with undefined axis: \"%s\"." % dimName)
```
anthrotype: same here, we should use `self.logger.warning`

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -1222,81 +1135,27 @@ def getAxis(self, name):
...
-        # FIXME: original implementation using MutatorMath
-        # masterLocations = [src.location for src in self.sources]
-        # mutatorBias = biasFromLocations(masterLocations, preferOrigin=False)
-        # for src in self.sources:
-        #     if src.location == mutatorBias:
-        #         return src
+                self.default = sourceDescriptor
+                return sourceDescriptor
+        # failing that, well, fail. We don't want to guess any more. 
+        warnings.warn("Can't find a suitable default location in this document.")
```
anthrotype: shouldn't this be an error then?

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -1222,81 +1135,27 @@ def getAxis(self, name):
...
         return None
 
-    def _prepAxesForBender(self):
+    def _axesAsDict(self):
         """
             Make the axis data we have available in
```
anthrotype: this sentence seems truncated

Tests/designspaceLib/designspace_test.py
```diff
@@ -5,6 +5,7 @@
 
 import os
 import pytest
+import warnings
```
anthrotype: this seems unused in here

Lib/fontTools/designspaceLib/__init__.py
anthrotype: or we may simply document that findDefault may return None, and then we don't need to warn here, the user has to handle the None case.

LettError:
Please remember that `varLib` is not the only consumer of designspaces. There are still a lot of files out there that rely `copyInfo` flags. I will make changes to Superpolator and ufoProcessor but those effects won't be immediate.

anthrotype:
Yes, I understand that. But I thought we said we would use copyInfo only when axes wheren't defined.
How will other designspace consumers deal with a master that says "I'm the default" but the axes elements say "no, she is lying"?

LettError:
Ok, what if `findDefault` only returns the proper _default-on-axes_. ?

anthrotype:
> what if findDefault only returns the proper default-on-axes?

yes, or None if none could be found.

Tests/designspaceLib/designspace_test.py
LettError: Fixed

PR #1276, opened 2018-06-18
WIP [mutator] set usWeightClass, usWidthClass and italicAngle

### LINKED ISSUE #1264: [varLib.mutator] Does not mutate weight/width/italicAngle
Because those are not in MVAR, but in standard axes. Quoting MVAR spec:

> Note: The OS/2.usWeightClass, OS/2.usWidthClass and post.italicAngle values are not supported by variation data in the 'MVAR' table. This is because values for these three fields correspond directly to input axis values for the 'wght', 'wdth' and 'slnt' variation axes. See the discussion of these axes in the OpenType Design-Variation Axis Tag Registry for details on the relationship between these fields and the corresponding design axes.

There's probably some binary flags that we can also tweak, re boldness, italicness, etc of the font.

anthrotype:
Fixes #1264

### DIFF
--- a/Lib/fontTools/varLib/mutator.py
+++ b/Lib/fontTools/varLib/mutator.py
@@ -21,6 +21,19 @@ import logging
 log = logging.getLogger("fontTools.varlib.mutator")
 
 
+OS2_WIDTH_CLASS_VALUES = {
+    50.0: 1,  # Ultra-condensed
+    62.5: 2,  # Extra-condensed
+    75.0: 3,  # Condensed
+    87.5: 4,  # Semi-condensed
+    100.0: 5,  # Medium (normal)
+    112.5: 6,  # Semi-expanded
+    125.0: 7,  # Expanded
+    150.0: 8,  # Extra-expanded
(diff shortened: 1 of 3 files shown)


Lib/fontTools/varLib/mutator.py
```diff
@@ -123,6 +136,21 @@ def instantiateVariableFont(varfont, location, inplace=False):
...
 		]
 
+	if "wght" in location and "OS/2" in varfont:
+		varfont["OS/2"].usWeightClass = otRound(
+			max(1, min(location["wght"], 1000))
+		)
+	if "wdth" in location:
+		wdth = location["wdth"]
+		for percent, widthClass in sorted(OS2_WIDTH_CLASS_VALUES.items()):
+			if wdth <= percent:
```
anthrotype: Not sure if this is correct, or if we should round in a more clever way, testing in the middle of those predefined percent intervals...

anthrotype:
ok, i'm rounding to the closest width class now

PR #1278, opened 2018-06-20
Allow any languages under DFLT script

mhosken:
The standard has been changed to allow any languages under the DFLT script. This patch removes the test and raised error that stops users from using languages other than dflt with the DFLT script.


anthrotype:
the feature file spec still needs to be updated though:

> The only permitted language tag for the 'DFLT script is 'dflt'.

https://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#2.c

I guess we can merge this into feaLib now, and then also change that spec later.

anthrotype:
@mhosken thanks!

PR #1282, opened 2018-06-27
Optimize the charstring's bytecode

miguelsousa:
Optimizes the charstring's bytecode by encoding as integers all float values that have no decimal portion. This optimization is expected to reduce the file size of CFF2 fonts that have intermediate masters.

Related to adobe-type-tools/afdko#444

Many thanks to @cjchapman for the help on the fixed point math.

### DIFF
--- a/Lib/fontTools/misc/psCharStrings.py
+++ b/Lib/fontTools/misc/psCharStrings.py
@@ -216,8 +216,12 @@ encodeIntT1 = getIntEncoder("t1")
 encodeIntT2 = getIntEncoder("t2")
 
 def encodeFixed(f, pack=struct.pack):
-	# For T2 only
-	return b"\xff" + pack(">l", otRound(f * 65536))
+	"""For T2 only"""
+	value = otRound(f * 65536)  # convert the float to fixed point
+	if value & 0xFFFF == 0:  # check if the fractional part is zero
+		return encodeIntT2(value >> 16)  # encode only the integer part
+	else:
+		return b"\xff" + pack(">l", value)  # encode the entire fixed point value
 
 def encodeFloat(f):
 	# For CFF only, used in cffLib
@@ -944,7 +948,7 @@ class T2CharString(object):
 	decompilerClass = SimpleT2Decompiler
 	outlineExtractor = T2OutlineExtractor
 	isCFF2 = False
-	
+
 	def __init__(self, bytecode=None, program=None, private=None, globalSubrs=None):
 		if program is None:
 			program = []
@@ -1006,8 +1010,7 @@ class T2CharString(object):
 		while i < end:
 			token = program[i]
 			i = i + 1
-			tp = type(token)
-			if issubclass(tp, basestring):
+			if isinstance(token, basestring):
 				try:
 					bytecode.extend(bytechr(b) for b in opcodes[token])
 				except KeyError:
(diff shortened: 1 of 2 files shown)


Lib/fontTools/misc/psCharStrings.py
```diff
@@ -1002,25 +1002,23 @@ def compile(self, isCFF2=False):
 		encodeInt = self.getIntEncoder()
 		encodeFixed = self.getFixedEncoder()
 		i = 0
-		end = len(program)
-		while i < end:
+		while i < len(program):
```
behdad: Any reason for this? Sure it's correct, but also calls `len(program)` more than is needed if `len(program)` doesn't change.

Lib/fontTools/misc/psCharStrings.py
```diff
@@ -1031,9 +1032,12 @@ def compile(self, isCFF2=False):
 		self.setBytecode(bytecode)
 
 		if isCFF2:
+			self.isCFF2 = True
 			# If present, remove return and endchar operators.
 			if self.bytecode and (byteord(self.bytecode[-1]) in (11, 14)):
 				self.bytecode = self.bytecode[:-1]
+		else:
+			self.isCFF2 = False
```
behdad: Better to just have `self.isCFF2 = isCFF2`.

Lib/fontTools/misc/psCharStrings.py
```diff
@@ -1031,9 +1032,12 @@ def compile(self, isCFF2=False):
 		self.setBytecode(bytecode)
 
 		if isCFF2:
+			self.isCFF2 = True
```
behdad: We specifically decided that we don't want `self.isCFF2` state on objects if it can be avoided.  In this case, it can.  Nothing about a charstring makes it a CFF or CFF2 charstring, as they are fully compatible.  It's only during compile that we decide whether it conforms to the resulting format.  Also, compile should not leave a state on the object.  Please rework without it.

Lib/fontTools/misc/psCharStrings.py
behdad: (But of course see my previous comment.)

Tests/misc/psCharStrings_test.py
```diff
@@ -9,22 +9,26 @@ class T2CharStringTest(unittest.TestCase):
...
 
     def test_calcBounds_empty(self):
         cs = self.stringToT2CharString("endchar")
         bounds = cs.calcBounds(None)
         self.assertEqual(bounds, None)
 
     def test_calcBounds_line(self):
-        cs = self.stringToT2CharString("100 100 rmoveto 40 10 rlineto -20 50 rlineto endchar")
+        cs = self.stringToT2CharString(
+            "100 100 rmoveto 40 10 rlineto -20 50 rlineto endchar")
```
behdad: Not sure these add any value.  We don't adhere to 80 columns.

Lib/fontTools/misc/psCharStrings.py
miguelsousa: brain fart; shows my inexperience using `while`

Lib/fontTools/misc/psCharStrings.py
miguelsousa: reworked. Required setting more things in the tests themselves.

miguelsousa:
@behdad is this good to merge now?

behdad:
Yes. Thanks.

PR #1289, opened 2018-07-07
[merge] CFF tables

BlackFoundry:
desubroutinizing with Subsetter
No compreffor. No final compression (may happen after, outside merge)

### DIFF
--- a/Lib/fontTools/merge.py
+++ b/Lib/fontTools/merge.py
@@ -227,7 +227,7 @@ ttLib.getTableClass('head').mergeMap = {
 	'macStyle': first,
 	'lowestRecPPEM': max,
 	'fontDirectionHint': lambda lst: 2,
-	'indexToLocFormat': equal,
+	'indexToLocFormat': first,
 	'glyphDataFormat': equal,
 }
 
--- a/Lib/fontTools/merge.py
+++ b/Lib/fontTools/merge.py
@@ -13,11 +13,15 @@ from fontTools.ttLib.tables import otTables, _h_e_a_d
 from fontTools.ttLib.tables.DefaultTable import DefaultTable
 from fontTools.misc.loggingTools import Timer
 from fontTools.pens.recordingPen import DecomposingRecordingPen
+from fontTools.subset import _DesubroutinizingT2Decompiler
 from functools import reduce
 import sys
 import time
 import operator
 import logging
+import compreffor
+import os
+import tempfile
 
 
 log = logging.getLogger("fontTools.merge")
@@ -205,7 +209,7 @@ ttLib.getTableClass('head').mergeMap = {
 	'macStyle': first,
 	'lowestRecPPEM': max,
 	'fontDirectionHint': lambda lst: 2,
-	'indexToLocFormat': recalculate,
+	'indexToLocFormat': equal,
 	'glyphDataFormat': equal,
 }
 
@@ -362,6 +366,76 @@ def _glyphsAreSame(glyphSet1, glyphSet2, glyph1, glyph2):
 	return (pen1.value == pen2.value and
 		g1.width == g2.width and
 		(not hasattr(g1, 'height') or g1.height == g2.height))
+
+@_add_method(ttLib.getTableClass('CFF '))
+def merge(self, m, tables):
+	newcff = tables[0]
+	newfont = newcff.cff[0]
+	private = newfont.Private
+	storedNamesStrings = []
+	glyphOrderStrings = []
+	for name in newfont.strings.strings:
+		if name not in newfont.getGlyphOrder():
+			storedNamesStrings.append(name)
+		else:
+			glyphOrderStrings.append(name)    
+	chrset = list(newfont.charset)
+	newcs = newfont.CharStrings
+	newcsi = newcs.charStringsIndex
+	log.info("Font 0 global subrs: %s." % len(newcff.cff.GlobalSubrs))
+	ls = None
+	try:
+		ls = newfont.Private.Subrs
+	except:
+		pass
+	if ls is not None:
+		log.info("Font 0 local subrs: %s." % str(len(ls)))
+	else:
+		log.info("Font 0 has no local subrs.")
+	log.info("FONT 0 CharStrings: %s, %s." % (str(len(newcs)), str(len(newcsi))))
+	baseIndex = len(newcsi)
+	for i, table in enumerate(tables[1:]):
+		font = table.cff[0] 
+		font.Private = private
+		for name in font.strings.strings:
+			if name in font.getGlyphOrder():
+				glyphOrderStrings.append(name)
+		cs = font.CharStrings
+		gs = table.cff.GlobalSubrs
+		log.info("Font %s global subrs: %s." % (str(i+1), str(len(gs))))
+		ls = None
+		try:
+			ls = font.Private.Subrs
+		except:
+			pass
+		if ls is not None:
+			log.info("Font %s global subrs: %s." % (str(i+1), str(len(ls))))
+		else:
+			log.info("Font %s has no local subrs." % str(i+1))
+		log.info("Font %s CharStrings: %s, %s." % (str(i+1), str(len(cs)), str(len(cs.charStringsIndex))))
+		if hasattr(font, "FDSelect"):
+			sel = font.FDSelect
+			log.info("HAS FDSelect %s." % str(sel))
+		else:
+			log.info("HAS NO FDSelect.")
+		numCharsExcludingNotDef = len(cs.charStringsIndex)-1
+		newcsi.items.extend([None] * numCharsExcludingNotDef)
+		j = baseIndex
+		for name,i in cs.charStrings.items():
+			if name in ['.notdef']:
+				continue
+			chrset.append(name)
+			newcs.charStrings[name] = j
+			ch = cs.charStringsIndex[i]
+			newcsi[j] = ch
+			j += 1
+		baseIndex = j
+
+	newfont.charset = chrset
+	newfont.numGlyphs = len(chrset)
+	newfont.strings.strings = glyphOrderStrings + storedNamesStrings
+
+	return newcff
 
 @_add_method(ttLib.getTableClass('cmap'))
 def merge(self, m, tables):
@@ -919,21 +993,58 @@ class Merger(object):
 		self.options = options
 
 	def merge(self, fontfiles):
-
-		mega = ttLib.TTFont()
+		# Take first input file sfntVersion
+		sfntVersion = ttLib.TTFont(fontfiles[0]).sfntVersion
+		mega = ttLib.TTFont(sfntVersion=sfntVersion)
 
 		#
 		# Settle on a mega glyph order.
 		#
 		fonts = [ttLib.TTFont(fontfile) for fontfile in fontfiles]
+
+		
+		# Check if all fonts have CFF
+		allOTFs = all(cff is not None for cff in [font.get('CFF ') for font in fonts])
+
+		cffTables = []
+		if allOTFs:
+			for font in fonts:
+				compreffor.decompress(font)
+				cffTables.append(font.get('CFF '))
+		# 	cffTables = [font.get('CFF ') for font in fonts]
+		# 	desubr_cffTables = []
+		# 	for cff in cffTables:
+		# 		# Desubroutinize
+		# 		cs = cff.cff[0].CharStrings
+		# 		for g in cff.cff[0].charset:
+		# 			c, _ = cs.getItemAndSelector(g)
+		# 			c.decompile()
+		# 			subrs = getattr(c.private, "Subrs", [])
+		# 			decompiler = _DesubroutinizingT2Decompiler(subrs, c.globalSubrs)
+		# 			decompiler.execute(c)
+		# 			c.program = c._desubroutinized
+		# 		desubr_cffTables.append(cff)
+		# else:
+		# 	desubr_cffTables = []
+
 		glyphOrders = [font.getGlyphOrder() for font in fonts]
-		megaGlyphOrder = self._mergeGlyphOrders(glyphOrders)
+
+		# Handle glyphOrder merging and cff glyphs renaming together
+		megaGlyphOrder, newcffTables = self._mergeGlyphOrders(glyphOrders, cffTables)
+
 		# Reload fonts and set new glyph names on them.
 		# TODO Is it necessary to reload font?  I think it is.  At least
 		# it's safer, in case tables were loaded to provide glyph names.
 		fonts = [ttLib.TTFont(fontfile) for fontfile in fontfiles]
-		for font,glyphOrder in zip(fonts, glyphOrders):
-			font.setGlyphOrder(glyphOrder)
+
+		if len(newcffTables) > 0:
+			for font, glyphOrder, cffTable in zip(fonts, glyphOrders, newcffTables):
+				font.setGlyphOrder(glyphOrder)
+				font['CFF '] = cffTable
+		else:
+			for font, glyphOrder in zip(fonts, glyphOrders):
+				font.setGlyphOrder(glyphOrder)
+
 		mega.setGlyphOrder(megaGlyphOrder)
 
 		for font in fonts:
@@ -945,13 +1056,16 @@ class Merger(object):
 		allTags = reduce(set.union, (list(font.keys()) for font in fonts), set())
 		allTags.remove('GlyphOrder')
 
-		# Make sure we process cmap before GSUB as we have a dependency there.
+		# Make sure we process CFF before cmap before GSUB as we have a dependency there.
 		if 'GSUB' in allTags:
 			allTags.remove('GSUB')
 			allTags = ['GSUB'] + list(allTags)
 		if 'cmap' in allTags:
 			allTags.remove('cmap')
 			allTags = ['cmap'] + list(allTags)
+		if 'CFF ' in allTags:
+			allTags.remove('CFF ')
+			allTags = ['CFF '] + list(allTags)
 
 		for tag in allTags:
 			with timer("merge '%s'" % tag):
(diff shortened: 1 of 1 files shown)


Lib/fontTools/merge.py
```diff
@@ -363,6 +364,74 @@ def _glyphsAreSame(glyphSet1, glyphSet2, glyph1, glyph2):
...
+	newcs = newfont.CharStrings
+	newcsi = newcs.charStringsIndex
+	log.info("Font 0 global subrs: %s." % len(newcff.cff.GlobalSubrs))
+	ls = None
+	try:
+		ls = newfont.Private.Subrs
+	except:
+		pass
+	if ls is not None:
+		log.info("Font 0 local subrs: %s." % str(len(ls)))
```
anthrotype: you don't need to call `str` to convert an integer to string if you are using the `%s` format

anthrotype: and when you use logging, you don't need to evaluate the `%` formatting operators immediately, but you should do instead:

```python
log.info("Font 0 local subrs: %s", len(ls))
```

the positional argument after the first one will be interpolated only if the logging message is handled (i.e. if the client app has configured logging, and if the logger is configured for that level of verbosity, etc.)

Lib/fontTools/merge.py
```diff
@@ -919,21 +988,63 @@ def __init__(self, options=None):
...
+		mega = ttLib.TTFont(sfntVersion=sfntVersion)
 
 		#
 		# Settle on a mega glyph order.
 		#
 		fonts = [ttLib.TTFont(fontfile) for fontfile in fontfiles]
+
+		
+		# Check if all fonts have CFF
+		allOTFs = all(cff is not None for cff in [font.get('CFF ') for font in fonts])
```
anthrotype: `allOTFs = all("CFF " in font for font in fonts)`

Lib/fontTools/merge.py
```diff
@@ -919,21 +988,63 @@ def __init__(self, options=None):
...
+
+		cffTables = []
+		if allOTFs:
+			for font in fonts:
+				# Desubroutinize with Subsetter (like compreffor does)
```
anthrotype: we could factor this out into a separate function `_desubroutinize(otf)` or similar

Lib/fontTools/merge.py
```diff
@@ -919,21 +988,63 @@ def __init__(self, options=None):
...
+		# Handle glyphOrder merging and cff glyphs renaming together
+		megaGlyphOrder = self._mergeGlyphOrders(glyphOrders, cffTables)
+
 		# Reload fonts and set new glyph names on them.
 		# TODO Is it necessary to reload font?  I think it is.  At least
 		# it's safer, in case tables were loaded to provide glyph names.
-		fonts = [ttLib.TTFont(fontfile) for fontfile in fontfiles]
-		for font,glyphOrder in zip(fonts, glyphOrders):
-			font.setGlyphOrder(glyphOrder)
+		# fonts = [ttLib.TTFont(fontfile) for fontfile in fontfiles]
```
anthrotype: why did you comment this out? Did you test nothing that it was not necessary to reload? In general, I don't like leaving commented-out code in, which becomes stale very quickly.

Lib/fontTools/merge.py
```diff
@@ -919,21 +988,63 @@ def __init__(self, options=None):
...
+				options.desubroutinize = True
+				options.notdef_outline = True
+				subsetter = subset.Subsetter(options=options)
+				subsetter.populate(glyphs=tmpfont.getGlyphOrder())
+				subsetter.subset(tmpfont)
+				data = tmpfont['CFF '].compile(tmpfont)
+				table = ttLib.newTable('CFF ')
+				table.decompile(data, font)
+				font['CFF '] = table
+				cffTables.append(font.get('CFF '))
```
anthrotype: you can unconditionally do `font["CFF "]` instead of `font.get("CFF ")` if you are certain, like in this particular case, that font has "CFF " table inside.
well, actually here you can just do `cffTables.append(table)`

Lib/fontTools/merge.py
```diff
@@ -919,21 +988,63 @@ def __init__(self, options=None):
...
 		# it's safer, in case tables were loaded to provide glyph names.
-		fonts = [ttLib.TTFont(fontfile) for fontfile in fontfiles]
-		for font,glyphOrder in zip(fonts, glyphOrders):
-			font.setGlyphOrder(glyphOrder)
+		# fonts = [ttLib.TTFont(fontfile) for fontfile in fontfiles]
+
+		if len(cffTables):
+			for font, glyphOrder, cffTable in zip(fonts, glyphOrders, cffTables):
+				font.setGlyphOrder(glyphOrder)
+				font['CFF '] = cffTable
```
anthrotype: why are you setting the CFF tables again in this line? You already set them to the same same objects a few lines above, no?
https://github.com/fonttools/fonttools/pull/1289/files#diff-4ac66bcb14ce18ebedd3df81c04259deR1025

Lib/fontTools/merge.py
```diff
@@ -945,13 +1056,16 @@ def merge(self, fontfiles):
 		allTags = reduce(set.union, (list(font.keys()) for font in fonts), set())
 		allTags.remove('GlyphOrder')
 
-		# Make sure we process cmap before GSUB as we have a dependency there.
+		# Make sure we process CFF before cmap before GSUB as we have a dependency there.
```
anthrotype:
```diff
- Make sure we process CFF before cmap before GSUB as we have a dependency there.
+ Make sure we process CFF before cmap, and cmap before GSUB as we have a dependency there.
```

Lib/fontTools/merge.py
```diff
@@ -975,20 +1089,52 @@ def merge(self, fontfiles):
...
 		# TODO Even this simplistic numbering can result in conflicts.
 		# But then again, we have to improve this soon anyway.
+
+		# If we have CFF tables,
+		# rename topDictIndex charStrings
+		for n,cffTable in enumerate(cffTables):
+			td = cffTable.cff.topDictIndex[0]
+			d = {}
+			for i, (glyphName, v) in enumerate(td.CharStrings.charStrings.items()):
+				if glyphName not in ['.notdef']:
```
anthrotype: wouldn't `if glyphName != ".notdef":` accomplish the same thing here?

Lib/fontTools/merge.py
```diff
@@ -975,20 +1089,52 @@ def merge(self, fontfiles):
...
-		for n,glyphOrder in enumerate(glyphOrders):
-			for i,glyphName in enumerate(glyphOrder):
-				glyphName += "#" + repr(n)
-				glyphOrder[i] = glyphName
-				mega.append(glyphName)
+
+		# If we have CFF tables,
+		# make sure to keep .notdef as first glyph,
+		# don't rename it, and skip any additional .notdef
+		if len(cffTables):
```
anthrotype: a non empty list evaluates True in boolean expressions, False if empty. You can change to `if cffTables:`

Lib/fontTools/merge.py
```diff
@@ -975,20 +1089,52 @@ def merge(self, fontfiles):
...
 		mega = []
-		for n,glyphOrder in enumerate(glyphOrders):
-			for i,glyphName in enumerate(glyphOrder):
-				glyphName += "#" + repr(n)
-				glyphOrder[i] = glyphName
-				mega.append(glyphName)
+
+		# If we have CFF tables,
+		# make sure to keep .notdef as first glyph,
+		# don't rename it, and skip any additional .notdef
```
anthrotype: the special handling of .notdef could be extended to non-CFF as well.

Lib/fontTools/merge.py
```diff
@@ -205,7 +206,7 @@ def merge(self, m, tables):
 	'macStyle': first,
 	'lowestRecPPEM': max,
 	'fontDirectionHint': lambda lst: 2,
-	'indexToLocFormat': recalculate,
+	'indexToLocFormat': equal,
```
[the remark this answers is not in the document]
BlackFoundry: it has to do with this:
https://github.com/fonttools/fonttools/blob/87178f413a97bd725c7a62e0c89108933e35c746/Lib/fontTools/ttLib/tables/_h_e_a_d.py#L61

with 'recalculate' I get this error:
  File "/Library/Python/2.7/site-packages/fonttools-3.28.1.dev0-py2.7.egg/fontTools/ttLib/ttFont.py", line 173, in save
    writer_reordersTables = self._save(tmp)
  File "/Library/Python/2.7/site-packages/fonttools-3.28.1.dev0-py2.7.egg/fontTools/ttLib/ttFont.py", line 212, in _save
    self._writeTable(tag, writer, done, tableCache)
  File "/Library/Python/2.7/site-packages/fonttools-3.28.1.dev0-py2.7.egg/fontTools/ttLib/ttFont.py", line 633, in _writeTable
    tabledata = self.getTableData(tag)
  File "/Library/Python/2.7/site-packages/fonttools-3.28.1.dev0-py2.7.egg/fontTools/ttLib/ttFont.py", line 651, in getTableData
    return self.tables[tag].compile(self)
  File "/Library/Python/2.7/site-packages/fonttools-3.28.1.dev0-py2.7.egg/fontTools/ttLib/tables/_h_e_a_d.py", line 69, in compile
    data = sstruct.pack(headFormat, self)
  File "/Library/Python/2.7/site-packages/fonttools-3.28.1.dev0-py2.7.egg/fontTools/misc/sstruct.py", line 68, in pack
    value = obj[name]
KeyError: 'indexToLocFormat'

Lib/fontTools/merge.py
BlackFoundry: I should remove it I think or reload before because, if reload happens here we reload the old subroutinized CFF tables with old glyph names too.

BlackFoundry: yes, that's a mistake, better keep it I think

Lib/fontTools/merge.py
BlackFoundry: that's because the fonts 'should' be reloaded… (see commented line 1038)

Lib/fontTools/merge.py
BlackFoundry: sure ;)

Lib/fontTools/merge.py
BlackFoundry: ok

Lib/fontTools/merge.py
BlackFoundry: right

Lib/fontTools/merge.py
BlackFoundry: makes sense

Lib/fontTools/merge.py
BlackFoundry: doing it

Lib/fontTools/merge.py
BlackFoundry: good :)

Lib/fontTools/merge.py
BlackFoundry: fixing these

Lib/fontTools/merge.py
```diff
@@ -363,6 +364,74 @@ def _glyphsAreSame(glyphSet1, glyphSet2, glyph1, glyph2):
...
+			glyphOrderStrings.append(name)    
+	chrset = list(newfont.charset)
+	newcs = newfont.CharStrings
+	newcsi = newcs.charStringsIndex
+	log.info("Font 0 global subrs: %s." % len(newcff.cff.GlobalSubrs))
+	ls = None
+	try:
+		ls = newfont.Private.Subrs
+	except:
+		pass
```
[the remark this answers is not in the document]
BlackFoundry: making a change

Lib/fontTools/merge.py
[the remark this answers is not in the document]
behdad: That's because with CFF fonts, no one will set this value.  Please use `first` instead.

Lib/fontTools/merge.py
```diff
@@ -371,48 +392,44 @@ def merge(self, m, tables):
...
-		ls = newfont.Private.Subrs
-	except:
-		pass
-	if ls is not None:
-		log.info("Font 0 local subrs: %s." % str(len(ls)))
-	else:
-		log.info("Font 0 has no local subrs.")
-	log.info("FONT 0 CharStrings: %s, %s." % (str(len(newcs)), str(len(newcsi))))
+	if hasattr(newfont, 'Private'):
+		if hasattr(newfont.Private, 'Subr'):
```
behdad: Just do `if hasattr(newfont, 'Private') and hasattr(newfont.Private, 'Subr'):`

Lib/fontTools/merge.py
```diff
@@ -371,48 +392,44 @@ def merge(self, m, tables):
...
-			ls = font.Private.Subrs
-		except:
-			pass
-		if ls is not None:
-			log.info("Font %s global subrs: %s." % (str(i+1), str(len(ls))))
-		else:
-			log.info("Font %s has no local subrs." % str(i+1))
-		log.info("Font %s CharStrings: %s, %s." % (str(i+1), str(len(cs)), str(len(cs.charStringsIndex))))
+		if hasattr(font, 'Private'):
+			if hasattr(font.Private, 'Subr'):
```
behdad: Same.

BlackFoundry:
ping

Lib/fontTools/merge.py
```diff
@@ -363,6 +385,68 @@ def _glyphsAreSame(glyphSet1, glyphSet2, glyph1, glyph2):
...
+	newcsi = newcs.charStringsIndex
+	log.info("Font 0 global subrs: %s.", len(newcff.cff.GlobalSubrs))
+	ls = None
+	if hasattr(newfont, 'Private') and hasattr(newfont.Private, 'Subr'):
+		ls = newfont.Private.Subrs
+	lenls = len(ls) if ls is not None else 0
+	log.info("Font 0 local subrs: %s.", lenls)
+	log.info("FONT 0 CharStrings: %s, %s.",(len(newcs), len(newcsi)))
+	baseIndex = len(newcsi)
+	for i, table in enumerate(tables[1:]):
```
anthrotype: `enumerate` can take a `start` argument, if you wish to start counting from 1, you can do

```python
for i, table in enumerate(tables[1:], start=1):
```

Lib/fontTools/merge.py
```diff
@@ -363,6 +385,68 @@ def _glyphsAreSame(glyphSet1, glyphSet2, glyph1, glyph2):
...
+		if hasattr(font, "FDSelect"):
+			sel = font.FDSelect
+			log.info("HAS FDSelect %s.", sel)
+		else:
+			log.info("HAS NO FDSelect.")
+		numCharsExcludingNotDef = len(cs.charStringsIndex)
+		newcsi.items.extend([None] * numCharsExcludingNotDef)
+		chrset.extend(list(font.charset))
+		j = baseIndex
+		for name,i in cs.charStrings.items():
```
anthrotype: the local variable `i` is already used for the outer for loop, better call this one with another name

Lib/fontTools/merge.py
```diff
@@ -363,6 +385,68 @@ def _glyphsAreSame(glyphSet1, glyphSet2, glyph1, glyph2):
...
+		if hasattr(font, 'Private') and hasattr(font.Private, 'Subr'):
+			ls = font.Private.Subrs
+		lenls = len(ls) if ls is not None else 0
+		log.info("Font %s global subrs: %s.", (i+1, lenls))
+		log.info("Font %s CharStrings: %s, %s.", (i+1, len(cs), len(cs.charStringsIndex)))
+		if hasattr(font, "FDSelect"):
+			sel = font.FDSelect
+			log.info("HAS FDSelect %s.", sel)
+		else:
+			log.info("HAS NO FDSelect.")
```
anthrotype: maybe these should be `log.debug` rather than `log.info`

Lib/fontTools/merge.py
```diff
@@ -363,6 +385,68 @@ def _glyphsAreSame(glyphSet1, glyphSet2, glyph1, glyph2):
...
+		numCharsExcludingNotDef = len(cs.charStringsIndex)
+		newcsi.items.extend([None] * numCharsExcludingNotDef)
+		chrset.extend(list(font.charset))
```
anthrotype: `extend` method takes any iterable, don't need to create a temporary list first

anthrotype:
Thanks Jeremie. Would be really nice to also have a small test case that exercise this code, or
if you can't/don't have time right now, maybe you could provide a test font.

BlackFoundry:
Not sure where to share this, but here is a merge OTF testing script:
[MergeOTFtesting.zip](https://github.com/fonttools/fonttools/files/2384210/MergeOTFtesting.zip)

BlackFoundry:
Cosimo thanks for the script simplification and running the test.
The test works fine on my fork here, I guess you're not using the right merge.py because line 962 looks like this on my fork: 
https://github.com/BlackFoundry/fonttools/blob/35ec0cf0f3b78bd942dff5e3ff90d684551acf87/Lib/fontTools/merge.py#L962

anthrotype:
Ouch, you’re right 🤪

anthrotype:
Thanks! i'll add a test tomorrow.

Lib/fontTools/merge.py
```diff
@@ -919,21 +1003,43 @@ def __init__(self, options=None):
...
+
+		
+		# Check if all fonts have CFF
+		allOTFs = all("CFF " in font for font in fonts)
+
+		cffTables = []
+		if allOTFs:
+			for font in fonts:
+				_desubtoutinize(font)
+				cffTables.append(font['CFF '])
```
anthrotype: what if only some of the font contain CFF table. I think we should error here

Lib/fontTools/merge.py
```diff
@@ -975,20 +1084,43 @@ def merge(self, fontfiles):
...
+				if glyphName != '.notdef':
+					glyphName += "#" + repr(n)
+				elif n > 0:
+					glyphName += "#" + repr(n)
+				d[glyphName] = v
+			cffTable.cff.topDictIndex[0].CharStrings.charStrings = d
+
 		mega = []
+
+		# If we have CFF tables,
```
anthrotype: @BlackFoundry this comment "If we have CFF tables, make sure to keep .notdef..." seems to also apply for non-CFF fonts, so the first part of the comment is misleading

Lib/fontTools/merge.py
```diff
@@ -363,6 +385,68 @@ def _glyphsAreSame(glyphSet1, glyphSet2, glyph1, glyph2):
...
+		if hasattr(font, 'Private') and hasattr(font.Private, 'Subr'):
+			ls = font.Private.Subrs
+		lenls = len(ls) if ls is not None else 0
+		log.info("Font %s global subrs: %s.", (i, lenls))
+		log.info("Font %s CharStrings: %s, %s.", (i, len(cs), len(cs.charStringsIndex)))
+		if hasattr(font, "FDSelect"):
+			sel = font.FDSelect
+			log.debug("HAS FDSelect %s.", sel)
+		else:
+			log.debug("HAS NO FDSelect.")
```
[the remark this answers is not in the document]
BlackFoundry: It does nothing more indeed.

Lib/fontTools/merge.py
```diff
@@ -363,6 +385,68 @@ def _glyphsAreSame(glyphSet1, glyphSet2, glyph1, glyph2):
...
+		numCharsExcludingNotDef = len(cs.charStringsIndex)
```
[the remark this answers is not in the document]
BlackFoundry: That is an old naming, should be something like 'numChars' only

Lib/fontTools/merge.py
BlackFoundry: True

Lib/fontTools/merge.py
BlackFoundry: True as well (old comment, the code changed after writing it)

Lib/fontTools/merge.py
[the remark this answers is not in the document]
anthrotype: so is merging CID-keyed CFF tables supported by this patch?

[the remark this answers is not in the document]
BlackFoundry: not in this patch, no.

Lib/fontTools/merge.py
```diff
@@ -363,6 +385,68 @@ def _glyphsAreSame(glyphSet1, glyphSet2, glyph1, glyph2):
...
+		for name in font.strings.strings:
+			if name in fontGlyphOrder:
+				glyphOrderStrings.append(name)
+		cs = font.CharStrings
+		gs = table.cff.GlobalSubrs
+		log.info("Font %s global subrs: %s." % (str(i), str(len(gs))))
+		ls = None
+		if hasattr(font, 'Private') and hasattr(font.Private, 'Subr'):
+			ls = font.Private.Subrs
+		lenls = len(ls) if ls is not None else 0
```
anthrotype: @BlackFoundry you're getting the local subroutines from the Private dict just to print the length, but you don't actually do anything with it.. Same for the global subroutines. :thinking:

BlackFoundry: yes you're right, that was for testing and informative purposes actually, they are not used for anythng else.

anthrotype:
hm, I merged this PR too hastily. It looks incomplete.
Since I need to tag a new release today, I think I'm going to revert this for now.
Sorry Jeremie.

Lib/fontTools/merge.py
anthrotype: should we actually raise an error if they are not empty? The code expects the cff tables to be desubroutinized

anthrotype:
I reverted the commit from master, and pushed a new `merge-cff` work-in-progress branch here on the upstream repository:
https://github.com/fonttools/fonttools/tree/merge-cff

BlackFoundry:
yeah you're right Cosimo, it's is more a WIP version for now, a bit early for a merge

PR #1292, opened 2018-07-09
More depths to allowing languages in DFLT

mhosken:
Need to allow DFLT based languagesystem commands.

Fix the problem that if a sublanguage is used in the DFLT script that all those lookups end up in DFLT dflt. Just treat it all like any other script.

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -704,10 +704,7 @@ class Builder(object):
                 if key[:2] in self.get_default_language_systems_():
                     lookups = [l for l in lookups if l not in dflt_lookups]
                 self.features_.setdefault(key, []).extend(lookups)
-        if self.script_ == 'DFLT':
-            langsys = set(self.get_default_language_systems_())
-        else:
-            langsys = set()
+        langsys = set()
         langsys.add((self.script_, language))
         self.language_systems = frozenset(langsys)
 
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -1207,10 +1207,6 @@ class Parser(object):
         script = self.expect_script_tag_()
         language = self.expect_language_tag_()
         self.expect_symbol_(";")
-        if script == "DFLT" and language != "dflt":
-            raise FeatureLibError(
-                'For script "DFLT", the language must be "dflt"',
-                self.cur_token_location_)
         return self.ast.LanguageSystemStatement(script, language,
                                                 location=location)
 


mhosken:
NB The PairPosTable test ttx showed the faultiness of the current code before the fix with one feature listing the lookup twice.

anthrotype:
thanks, overall this lgtm.
would be nice to also add an explicit test for this situation of DFLT script, non-dflt language.

Lib/fontTools/feaLib/builder.py
```diff
@@ -704,10 +704,7 @@ def set_language(self, location, language, include_default, required):
                 if key[:2] in self.get_default_language_systems_():
                     lookups = [l for l in lookups if l not in dflt_lookups]
                 self.features_.setdefault(key, []).extend(lookups)
-        if self.script_ == 'DFLT':
-            langsys = set(self.get_default_language_systems_())
-        else:
-            langsys = set()
+        langsys = set()
```
anthrotype: This and the lines following can be simplified now: instead of creating a mutable empty set, calling add and then converting it to a frozenset, you could directly initialise the frozenset in one go

Lib/fontTools/feaLib/builder.py
```diff
@@ -704,10 +704,7 @@ def set_language(self, location, language, include_default, required):
                 if key[:2] in self.get_default_language_systems_():
                     lookups = [l for l in lookups if l not in dflt_lookups]
                 self.features_.setdefault(key, []).extend(lookups)
-        if self.script_ == 'DFLT':
-            langsys = set(self.get_default_language_systems_())
```
anthrotype: Wait, I’m not 100% sure it’s correct that you got rid of that call to  get_default_language_systems_ here.
Supposedly that was a “feature” (ie. falling back to DFLT dflt when no languagesystems statement are exlicitly defined in feature file). I need to check again again the fea spec. I’ll look at this tomorrow, I’m out of the office now.

anthrotype:
also i'm not very happy that we are unilaterally implementing this in feaLib while the Feature FIle spec is not also updated
https://github.com/adobe-type-tools/afdko/issues/438

mhosken:
I agree that the spec should be updated. I do hope that this PR doesn't await that change since it is blocking production. TIA.

readroberts:
@anthrotype I will be updating the afdko feature file spec to match the open type spec sometime in the next few days.

mhosken:
ping

readroberts:
@anthrotype @mhosken I have updated wording for the feature file spec, and a fix for makeotf, but I have another change in feature file spec wording pending ( see issue 1273) and wanted to wait so as to submit both as one update to the feature file spec.

mhosken:
ping2

anthrotype:
sorry Martin, I still have some concerns and would like to play with this a bit more. I promise this won't take another week.

readroberts:
@mhosken @anthrotype Note that I made an update to the feature file spec and makeotf code in afdko PR's 503 and 498.

Lib/fontTools/feaLib/builder.py
anthrotype: ok, I finally checked this. Martin is correct, the whole `if` block here was wrong.

The parser asserted that the only valid language tag that could be associated with script "DFLT" was "dflt"; so, what this `if self.script_ == 'DFLT':` block was doing basically was that, if a feature contained explicit `script DFLT; language dflt` statements then any rules or lookup defined after them would be registered not only for DFLT/dflt, but also (and that's incorrect) for all the globally defined languagesystems. And that is why in the PairPosSubtable.ttx expected test result, the same kernlookup was incorrectly registered twice under latn/dflt language system: i.e. once because of the explicit DFLT/dflt that precedes, and again because added specifically to latn/dflt in the following statements.

https://github.com/fonttools/fonttools/blob/632192804fefc4a0f7d37f54e2cc25b82f6b460e/Tests/feaLib/data/PairPosSubtable.fea

What the Feature File spec says, instead, is:

> Rules that are specified after the start of a feature and before the first "script" and/or "language" statement will be included in all the language systems specified by the "languagesystem" statements. If you do not want any of the rules in the feature to be registered under the language systems specified by the "languagesystem" statements, then a "script" and/or "language" statement for a script must be present before the first rule in the feature.

which is exactly what that feature file snippet is doing: i.e. it defines an explicit script/language combination, in this case DFLT/dflt, so the here meaning is: "I do not want to register the following lookup for all the global languagesystems, but only for this specific one".  There is nothing special about DFLT/dflt combination as far as the determination of the _default lookups_. The spec only says that the rules/lookups that occur between the start of the feature and the _first script_ (doesn't say the first non-DFLT script) are are added to all languagesystems.

Thanks @mhosken for spotting this!

anthrotype:
@readroberts I see you added the requirement that 

> Any other language statements using the 'DFLT' script tag must precede all other language statements.

why exactly is that? feaLib currently doesn't enforce this.

anthrotype:
anyway, now it does (d3229d5)

readroberts:
@anthrotype I think you are correct that there is no need to require that "Any other language statements using the 'DFLT' script tag must precede all other language statements." I'd be fine with removing this restriction.

Issue #1296, opened 2018-07-18
[varlib] Cannot build VF font for SourceSerifPro with fontmake


readroberts:
varlib used to be able to merge the master designs of SourceSerifPro. This broke after the application of varIDXmap optimization to ValueRecords in in varStore.py:: _visit() of (added 2/27 with commit [54fd71f7]), and the use of caching in varStore.py:: OnlineVarStoreBuilder.storeMasters() (added 2/21 with commit [e299650f]). The immediate problem is that an offset overflow error is not being handled, but this didn't happen before the changes. You can reproduce this by downloading the sources for Source Serif Pro from:
https://github.com/adobe-fonts/source-serif-pro

cd to the Roman/Master directory, and say:
fontmake -m SourceSerifVariable-Roman.designspace  -o variable

This will fail the first time, as the interpolatable ttf file for each master design will be built in the directory Masters/master_ttf_interpolatable/, but the varlib merge codes expects them to be in the same directories as the source ufo fonts, Masters/master_[0,1,2]. You can work around this by copying them to the ufo directories, and running fontmake again.

I hope to look at this sometime fairly soon, but would welcome any suggestions, if anyone else has seen similar problems.

miguelsousa:
@readroberts those instructions didn't work for me. I tried them with a clean installation of fontmake version 1.6.0. I did run into two issues though.

One was in fontmake and I logged it at https://github.com/googlei18n/fontmake/issues/451 . My solution was to comment line 585.

The other one was in the feature files of Source Serif, and relates to this change on how the `include()` statements are handled https://github.com/googlei18n/fontmake/issues/157 . To fix this one I created a branch named `fontmake-1.6.0` that contains the necessary changes https://github.com/adobe-fonts/source-serif-pro/tree/fontmake-1.6.0

miguelsousa:
@behdad this is the traceback. Any idea of what's wrong?

```python
Traceback (most recent call last):
  File "bin/fontmake", line 11, in <module>
    sys.exit(main())
  File "lib/python2.7/site-packages/fontmake/__main__.py", line 256, in main
    project.run_from_designspace(designspace_path, **args)
  File "lib/python2.7/site-packages/fontmake/font_project.py", line 623, in run_from_designspace
    **kwargs)
  File "lib/python2.7/site-packages/fontmake/font_project.py", line 689, in run_from_ufos
    master_bin_dir=master_bin_dir)
[...]

PR #1297, opened 2018-07-19
atttempt to fix offset overflows in MarkBasePos subtables

anthrotype:
if an offset overflows within a MarkBasePos (GPOS lookup type 4), we attempt to split it by moving half of the mark classes to the new subtable.

Fixes https://github.com/googlei18n/fontmake/issues/450

### DIFF
--- a/Lib/fontTools/ttLib/tables/otTables.py
+++ b/Lib/fontTools/ttLib/tables/otTables.py
@@ -1281,6 +1281,67 @@ def splitPairPos(oldSubTable, newSubTable, overflowRecord):
 	return ok
 
 
+def splitMarkBasePos(oldSubTable, newSubTable, overflowRecord):
+	# split half of the mark classes to the new subtable
+	classCount = oldSubTable.ClassCount
+	if classCount < 2:
+		# oh well, not much left to split...
+		return False
+
+	oldClassCount = classCount // 2
+	newClassCount = classCount - oldClassCount
+
+	oldMarkCoverage, oldMarkRecords = [], []
+	newMarkCoverage, newMarkRecords = [], []
+	for glyphName, markRecord in zip(
+		oldSubTable.MarkCoverage.glyphs,
+		oldSubTable.MarkArray.MarkRecord
+	):
+		if markRecord.Class < oldClassCount:
+			oldMarkCoverage.append(glyphName)
+			oldMarkRecords.append(markRecord)
+		else:
+			newMarkCoverage.append(glyphName)
+			newMarkRecords.append(markRecord)
+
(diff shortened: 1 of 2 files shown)


anthrotype:
this should also fix the first of the two errors reported by @miguelsousa in https://github.com/fonttools/fonttools/issues/1296#issuecomment-406156935

miguelsousa:
@anthrotype this patch does NOT fix the Source Serif overflow error reported in #1296.
These are the messages before the traceback happens:
```
...
INFO:fontTools.varLib:Merging TT hinting
INFO:fontTools.ttLib.tables.otBase:Attempting to fix OTLOffsetOverflowError ('GPOS', 'LookupIndex:', 8, 'SubTableIndex:', None, 'ItemName:', None, 'ItemIndex:', None)
INFO:fontTools.ttLib.tables.otBase:Attempting to fix OTLOffsetOverflowError ('GPOS', 'LookupIndex:', 0, 'SubTableIndex:', 0, 'ItemName:', 'BaseArray.BaseAnchor.YDeviceTable', 'ItemIndex:', None)
INFO:fontTools.ttLib.tables.otBase:Attempting to fix OTLOffsetOverflowError ('GPOS', 'LookupIndex:', 0, 'SubTableIndex:', 0, 'ItemName:', 'BaseArray.BaseAnchor.YDeviceTable', 'ItemIndex:', None)
...
fontTools.ttLib.tables.otBase.OTLOffsetOverflowError: ('GPOS', 'LookupIndex:', 0, 'SubTableIndex:', 0, 'ItemName:', 'BaseArray.BaseAnchor.YDeviceTable', 'ItemIndex:', None)
```

anthrotype:
Yeah, I checked and indeed, this PR doesn't fix that Source Serif Pro build issue.
The MarkToBase subtable whose offsets are overflowing already only contains ClassCount of 1 so there is nothing else to split according to the method implemented here (which only splits the mark classes and respective records between the two new split subtables). When there is only one mark class defined like in this case, it just bails out.

I'm going to merge this PR, as it fixes some other Noto build failures. We'll try to fix #1296 in a different patch.

behdad:
> Yeah, I checked and indeed, this PR doesn't fix that Source Serif Pro build issue.
> In that font, the MarkToBase subtable whose offsets are overflowing already only contains ClassCount of 1

Can you check how many marks and how many bases in there?

anthrotype:
not many: 28 marks and 71 bases

behdad:
> not many: 28 marks and 71 bases

That's unusual.  How can I test this easily?

behdad:
Ok I follow up on the other issue.

PR #1302, opened 2018-07-26
Fix subtable break

moyogo:
The PairPos subtable break was being forced after each 'pos' statement following a 'subtable' statement, it should only be forced the time.


anthrotype:
Thanks Denis! You're right, the state of the builder needed to be reset after building each subtables!

PR #1307, opened 2018-08-17
Fix default lookups for a language

mhosken:
We still don't have the whole script language thing quite right.

In the existing code, the comment that says that one extends the list of lookups for a language that has been set by a languagesystem defaults, with the lookups from the dflt language in that script, is wrong. It should say, one simply replaces the lookups that may have been added by the languagesystems with the lookups from the dflt language in the script.

The existing code has a bug in that for any language in the DFLT script (which hasn't been available until now) the lookups will not be copied since any lookups in DFLT/dflt are not included in the lookups extension list. So the set of (script/dflt - DFLT/dflt) for script=DFLT, is empty. What is needed is much simpler. No adding of lookups, one simply copies the script/dflt lookups into the new language's lookup list as per the spec. The result is that only if there is no language statement for a language that has a languagesystem, will the list of feature default (before the first script statement) lookups be used for that script/language.

Hmm. This description is more complicated than the code!

Eliminating the bug simplified the code, by removing any use of the dflt_lookups, and replacing .extend by copy. This reduced the code to the point where it could be refactored into something quite small and natural, but structured differently.

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -706,7 +706,7 @@ class Builder(object):
 
         key = (self.script_, language, self.cur_feature_name_)
         lookups = self.features_.get((key[0], 'dflt', key[2]))
-        if language != 'dflt' and include_default and lookups:
+        if (language == 'dflt' or include_default) and lookups:
             self.features_[key] = lookups[:]
         else:
             self.features_[key] = []
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -705,19 +705,11 @@ class Builder(object):
         self.cur_lookup_ = None
 
         key = (self.script_, language, self.cur_feature_name_)
-        if not include_default:
-            # don't include any lookups added by script DFLT in this feature
+        lookups = self.features_.get((key[0], 'dflt', key[2]))
+        if language != 'dflt' and include_default and lookups:
+            self.features_[key] = lookups[:]
+        else:
             self.features_[key] = []
-        elif language != 'dflt':
-            # add rules defined between script statement and its first following
-            # language statement to each of its explicitly specified languages:
-            # http://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#4.b.ii
-            lookups = self.features_.get((key[0], 'dflt', key[2]))
-            dflt_lookups = self.features_.get(('DFLT', 'dflt', key[2]), [])
-            if lookups:
-                if key[:2] in self.get_default_language_systems_():
-                    lookups = [l for l in lookups if l not in dflt_lookups]
-                self.features_.setdefault(key, []).extend(lookups)
         self.language_systems = frozenset([(self.script_, language)])
 
         if required:
--- a/Tests/feaLib/data/spec4h1.fea
+++ b/Tests/feaLib/data/spec4h1.fea
@@ -33,8 +33,8 @@ feature liga {
     sub c s by c_s;
 
     # The rules above will be placed in a lookup that is registered
-    # for all the specified languages for the script latn, but not any
-    # other scripts.
+    # for only the dflt language for this script, and not any other
+    # languages.
 
     language DEU;
     # script latn;       (stays the same)
@@ -43,16 +43,15 @@ feature liga {
     sub c k by c_k;
 
     # The rules above will be placed in a lookup that is registered
-    # only under the script latn, language DEU.
+    # only under the script latn, language DEU, along with the lookup
+    # registered for the dflt language for this script.
 
     language TRK;
 
-    # This will inherit both the top level default rules - the rules
-    # defined before the first 'script' statement, and the
-    # script-level default rules for 'latn': all the lookups of this
-    # feature defined after the 'script latn' statement, and before
-    # the language DEU statement.  If TRK were not named here, it
-    # would not inherit the default rules for the script latn.
+    # This will inherit only the lookup defined for the dflt language
+    # for this script.
+    # If this statement were not here, the language would only inherit
+    # the lookup at the start of this feature before script latn.
 } liga;
 
 # TODO(sascha): Uncomment once we support 'pos' statements.
--- a/Tests/feaLib/data/spec4h1.fea
+++ b/Tests/feaLib/data/spec4h1.fea
@@ -33,8 +33,8 @@ feature liga {
     sub c s by c_s;
 
     # The rules above will be placed in a lookup that is registered
-    # for only the dflt language for this script, and not any other
-    # languages.
+    # for all the specified languages for the script latn, but not any
+    # other scripts.
 
     language DEU;
     # script latn;       (stays the same)
@@ -43,15 +43,16 @@ feature liga {
     sub c k by c_k;
 
     # The rules above will be placed in a lookup that is registered
-    # only under the script latn, language DEU, along with the lookup
-    # registered for the dflt language for this script.
+    # only under the script latn, language DEU.
 
     language TRK;
 
-    # This will inherit only the lookup defined for the dflt language
-    # for this script.
-    # If this statement were not here, the language would only inherit
-    # the lookup at the start of this feature before script latn.
+    # This will inherit both the top level default rules - the rules
+    # defined before the first 'script' statement, and the
+    # script-level default rules for 'latn': all the lookups of this
+    # feature defined after the 'script latn' statement, and before
+    # the language DEU statement.  If TRK were not named here, it
+    # would not inherit the default rules for the script latn.
 } liga;
 
 # TODO(sascha): Uncomment once we support 'pos' statements.
--- a/Tests/feaLib/builder_test.py
+++ b/Tests/feaLib/builder_test.py
@@ -65,7 +65,7 @@ class BuilderTest(unittest.TestCase):
         spec9a spec9b spec9c1 spec9c2 spec9c3 spec9d spec9e spec9f spec9g
         spec10
         bug453 bug457 bug463 bug501 bug502 bug504 bug505 bug506 bug509
-        bug512 bug514 bug568 bug633
+        bug512 bug514 bug568 bug633 bug1307
         name size size2 multiple_feature_blocks omitted_GlyphClassDef
         ZeroValue_SinglePos_horizontal ZeroValue_SinglePos_vertical
         ZeroValue_PairPos_horizontal ZeroValue_PairPos_vertical
--- a/Tests/feaLib/data/bug1307.fea
+++ b/Tests/feaLib/data/bug1307.fea
@@ -1,0 +1,65 @@
+# Test of features and languagesystems
+
+lookup a {
+    sub a by A;
+} a;
+
+lookup b {
+    sub b by B;
+} b;
+
+lookup c {
+    sub c by C;
+} c;
+
+lookup d {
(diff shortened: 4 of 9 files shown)


mhosken:
I've looked at the tests and what the semantics change to. I consider them to be nearly right now. There is another bug to be discussed about the merging of features.

BUT these changes should be reviewed because they do change what people may assume to be the default behaviour for fea. For example:

```
languagesystem DFLT dflt;
languagesystem DFLT kht;

feat liga {
    sub f i by f_i;

    script DFLT;
    language kht;
        sub f i by f_i.other;
} liga;
```

Someone might assume that the DFLT/dflt will contain the first lookup. But due to the script statement containing an implicit language dflt statement, the list of lookups for DFLT/dflt will be cleared as if that statement were in the fea file after the script statement. This applies to all scripts and the dflt language.

This may break existing fonts in hard to discover ways.

We can, if we want, change the behaviour to say that a script statement does not clear the dflt. But then if you add lookups after a script statement and before a language statement, they will be appended in the dflt language to the feature default lookups (those before any script statement).

mhosken:
More thinking on this question (please no, I hear you cry!):

As the fea spec stands now, the usefulness of the languagesystem statement is limited. As soon as you have a script statement in a feature, you may as well not have had any languagesystem statements involving that script. You have to list all the languages every time, under the script, otherwise you don't get what you might want. Consider:

```
languagesystem DFLT dflt;
languagesystem latn dflt;
languagesystem latn DEU;

feat liga {
    script latn;
    language dflt;   # not required, this is implicit
        sub f i by fi_i;
} liga;
```

You might think that having a languageystem statement involving latn/DEU will save you, but it will not. The latn/DEU liga feature will have no lookups in it. You have to include a `language DEU;` statement at the end of the feature if you want it to pick up the dflt lookup.

Now the fea spec is a little ambiguous in its wording:

    When a "script" statement is seen, the language attribute is implicitly set to 'dflt', 

Does that mean that the dflt inherits the feature level lookups and adds to them? Does it mean that the dflt language has no lookups in by default? If the dflt inherits the feature level lookups, does it mean that any additional lookups are added to that list or is the list reset on the first lookup?

Any which way it doesn't solve the languagesystem problem. The ideal would be to say that for all languagesystem statements involving a script, an implicit language statement will be inserted after the dflt language for the script (before the first language statement or next script or end of feature). This would cause all language to inherit the dflt lookups (which may in turn inherit the feature lookups) and life would be more intuitive. Any subsequent language statements may add to the list or reset it if exclude_dflt is set. But doing this would deviate considerably from the fea spec as it stands. And since this would be a change to existing behaviour, is unlikely to be accepted.

So, that leaves us with having to specify language every time and, if we are wise not mixing feature level lookups with script statements. Clear? !!!

mhosken:
ping

readroberts:
@miguelsousa  @cjchapman Thank you for this analysis.  It is indeed a complicated issue, and I am all for getting this fixed. However, makes my head hurt at this hour. I need to think about this more. I will be on vacation the next week and half, but will take some time to review, so that when I come back we can finish this discussion and agree on what is right.

Tests/feaLib/data/spec4h1.fea
```diff
@@ -43,16 +43,15 @@ feature liga {
...
-    # This will inherit both the top level default rules - the rules
-    # defined before the first 'script' statement, and the
-    # script-level default rules for 'latn': all the lookups of this
-    # feature defined after the 'script latn' statement, and before
-    # the language DEU statement.  If TRK were not named here, it
-    # would not inherit the default rules for the script latn.
+    # This will inherit only the lookup defined for the dflt language
+    # for this script.
+    # If this statement were not here, the language would only inherit
+    # the lookup at the start of this feature before script latn.
```
anthrotype: The wording of these comments is copied verbatim from the Feature File spec (example 1 from paragraph 4h).
We can't simply unilaterally change it.

Lib/fontTools/feaLib/builder.py
```diff
@@ -705,19 +705,11 @@ def set_language(self, location, language, include_default, required):
...
-        elif language != 'dflt':
-            # add rules defined between script statement and its first following
-            # language statement to each of its explicitly specified languages:
-            # http://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#4.b.ii
-            lookups = self.features_.get((key[0], 'dflt', key[2]))
-            dflt_lookups = self.features_.get(('DFLT', 'dflt', key[2]), [])
-            if lookups:
-                if key[:2] in self.get_default_language_systems_():
-                    lookups = [l for l in lookups if l not in dflt_lookups]
-                self.features_.setdefault(key, []).extend(lookups)
```
anthrotype: In the previous code, the occurrence of a non-dflt language statement (with include_dflt implicitly true) would _extend_ the lookup list of the feature record for that script/language combo with the current script's default lookups (i.e. those between the script statement and the first non-dflt language statement). And the occurence of a `language dflt` (either explicitly or implicitly via a script statement) would do nothing, besides setting the current `self.language_systems` to this specific script/dflt (replacing the global set from languagesystem statements). 

Now, with your code, the occurrence of a non-dflt (including) language statement sets the feature's lookup list for this script/language to the script's default lookups, discarding any feature-level default lookups (i.e. occurring before any script statement). Moreover with your patch, when a `language dflt` statement is seen , the current script/language lookup list is always emptied.

The result of these modifications can be seen in the adjusted test results in `spec4h1.ttx`, in which lookup 1 (the one with the f-ligatures rules defined at the top of liga feature before any script statement) is no longer included in the lookup lists of the feature records of the latn/dflt, latn/DEU and latn/TRK.

I believe this is a deviation from the current spec.

readroberts:
@mhosken I think you make two objections to the current feature file spec. 
1) "But due to the script statement containing an implicit language dflt statement, the list of lookups for DFLT/dflt will be cleared as if that statement were in the fea file after the script statement. "

2)Also, you state that when the 'dflt' language tag is used,  "You have to list all the languages every time, under the script, otherwise you don't get what you might want."

I think this is a misunderstanding. The  spec says that there are special rules for the 'dflt' language, which I think address your issues. I read that paragraph as saying that neither implicit or explicit declaration of the language 'dflt' clears the list of lookups for the current script/dflt, for any script, including 'DFLT'. I think that this is what you are saying should happen. For your example:
```
languagesystem DFLT dflt;
languagesystem latn dflt;
languagesystem latn DEU;

feat liga {
    script latn;
    language dflt;   # not required, this is implicit
        sub f i by fi_i;

} liga;
```
I would expect that  the rule 'sub f i by fi_i;' would  end up in both latn/dflt and latn/DEU, and not in DFLT/dflt. Likewise, in your first code sample above, 'sub f i by f_i;' would in end up in both DFLT/dflt and DFLT/kht, and 'sub f i by f_i.other;' should end up in only DFLT/kht. If this isn't clear, then we should add more examples to the spec.

anthrotype:
I deleted my previous comment, as I evidently misunderstood that Read was actually summarising what according to Martin it _should_ happen, not what the current spec specifies. I also typed it in a rush from my phone and in between tube stations... Sorry for the noise.

readroberts:
@anthrotype Actually, your first assumption was correct, and re-reading all this, I agree that I was wrong and according to the spec,  the fi_i ligature shouldn't be registered under latn/DEU in the second example,  which is what I think you meant - it should be registered only under latn/dflt, and latn/DEU will be empty. However, latn/DEU would, in a more real-world example, contain any feature-level defaults.

I think that Martin's original complaint about the first example:
"""Someone might assume that the DFLT/dflt will contain the first lookup. But due to the script statement containing an implicit language dflt statement, the list of lookups for DFLT/dflt will be cleared as if that statement were in the fea file after the script statement. This applies to all scripts and the dflt language."""
is something that should not happen, but is also behavior that differs from the spec. In the first example 'sub f i by f_i;' is a feature-level rule, and should end up in both the DFLT/dflt and latn/dflt.

I think there spec may not be clear enough about feature-level defaults and script-level defaults. This is what I think the spec says:
- All script/language pairs specified by a languagesystem keyword inherit all feature-level defaults
- All script/languages specified by the script and language keywords inherit all feature-level defaults, unless the 'exclude_dflt'  keyword is used with  the language keyword.
 - script-level default rules are those between the use of a 'script' keyword and the first language tag other than 'dflt'
- specifying the language 'dflt' explicitly has no effect at all.
- script-level default rules are added only to languages specified explicitly within a feature following a script keyword.

I think that Martin's suggestion is to go beyond changing code to fix the original complaint, and to say that script-level default rules ( i.e. those between a script statement and the first non-default language) should be added to all the languages defined for the same script, whether  in the languagesystem statements or those languages explicitly defined for a script in the same feature.

The advantage to this is avoiding the requirement (mentioned in the spec) to add an empty language keyword in the feature for every languagesystem statement for the current script that should inherit the script-level defaults. This is certainly requires some work, and is is prone to the error of overlooking some combinations. However, the problem is that the new rule forces all languagesystem statements for a script to inherit the script-level defaults for the script from all features. The reason for the current spec is because you may not want all the script-level defaults for all features to show up in all languages for a given script. The new rule would not allow you to prevent this.


@anthrotype comments that the new code adds additional new rules:
-  the occurrence of a non-dflt (including) language statement sets the feature's lookup list for this script/language to the script's default lookups, discarding any feature-level default lookups (i.e. occurring before any script statement). 
- when a language dflt statement is seen , the current script/language lookup list is always emptied.

I would argue against adding these rules, as they would mean that feature-level rules could not be inherited by any rule sets within an explicit script statement block.

mhosken:
Given I need the bug described in my previous comment, fixed soon (since it's blocking a project), I continue onward. Let's agree what we want the behaviour to be. Consider this fea fragment:
```
languagesystem DFLT dflt;
languagesystem DFLT FRE;
languagesystem DFLT ABC;
languagesystem latn dflt;
languagesystem latn ABC;

feat smcp {
    lookup a;
} smcp;

feat liga {
    lookup b;
    script DFLT;
        lookup c;
        language dflt;
            lookup d;
        language FRE;
            lookup e;
    script latn;
        lookup f;
        lang dflt;
            lookup g;
        lang FRE;
            lookup h;
        lang DEF exclude_dflt;
            lookup i;
} liga;
```
This would result in:
```
DFLT/dflt --> smcp(a), liga(b,c,d)
DFLT/ABC  --> smcp(a), liga(b)
DFLT/FRE  --> smcp(a), liga(b,c,d,e)
latn/dflt --> smcp(a), liga(b,f,g)
latn/FRE  --> liga(b,f,g,h)
latn/ABC  --> smcp(a), liga(b)
latn/DEF  --> liga(i)
```
With the current code we get:
```
DFLT/dflt --> smcp(a), liga(b,c,d)
DFLT/ABC  --> smcp(a), liga(b)
DFLT/FRE  --> smcp(a), liga(b,e)
latn/dflt --> smcp(a), liga(b,f,g)
latn/FRE --> liga(b,f,g,h)
latn/ABC  --> smcp(a), liga(b)
latn/DEF  --> liga(i)
```
Everything is very close but DFLT/FRE is wrong. I would also point out that if you want things like latn/ABC to pick up latn/dflt and be liga(b,f,g), then one has to include language statements for them. I think this is a bug in the spec, since in complex situations with multiple scripts, few lookups end up before the first script statement, but many end up in script language dflt. But I am giving up seeing that fixed.

OK I think i can implement this. Look for an update...

mhosken:
I think this meets the requirements. Please review.

PR #1310, opened 2018-08-22
TSIC table support a Visual TrueType source table

paullinnerud:
TSIC is a source table for Visual TrueType to maintain sources for the cvar variation table. Visual TrueType will generate this table to maintain the user intended CVT values as various point in variation space. This table is removed prior to shipping a font. Similar in concept to other TSI* tables already supported in fonttools.

### DIFF
--- a/Lib/fontTools/ttLib/tables/T_S_I_C_.py
+++ b/Lib/fontTools/ttLib/tables/T_S_I_C_.py
@@ -1,0 +1,7 @@
+from __future__ import print_function, division, absolute_import
+from fontTools.misc.py23 import *
+from .otBase import BaseTTXConverter
+
+
+class table_T_S_I_C_(BaseTTXConverter):
+    pass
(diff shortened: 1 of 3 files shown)


moyogo:
The axis tags are backwards. Either VTT needs to write then correctly or fonttools needs to handle them properly.
```xml
    <AxisArray index="0" value="thgw"/>
    <AxisArray index="1" value="htdw"/>
```

moyogo:
Per https://docs.microsoft.com/en-us/typography/opentype/spec/otff
> Tag | Array of four uint8s (length = 32 bits) used to identify a table,  design-variation axis, script, language system, feature, or baseline

VTT is not writing the Tag properly.

paullinnerud:
Yes, the current release version of VTT is not writing the tags in the TSIC table correctly. I have fixed this and it will be in the next release. Fortunately all that VTT uses these tags for is to verify the axis structure of the TSIC is consistent with the fvar. The new fixed VTT will either accept correct or backwards tags and then write the correct tag so the impact of the bug is pretty minimal.

paullinnerud:
Are there any additional process steps needed for this pull request?

anthrotype:
@paullinnerud sorry, I was on holiday a few weeks ago, then it slipped my mind. Thank you for the pull request by the way! I promise I'll have a look tomorrow.

paullinnerud:
Thanks

anthrotype:
looks good to me, thanks!

madig:
@paullinnerud Do I understand correctly that the `TSIC` table is supposed to contain complete information to generate a `cvar` table? In my testing here, VTT writes to the `cvar` directly and the `TSIC` table appears to be mostly unused?

Issue #1224, opened 2018-03-27
[varLib] Use designspaceLib instead of varLib.designspace


behdad:
Port the latter to use the former and keep for compatibility.

Also, coordinate with https://github.com/fonttools/fonttools/issues/1223

PR #1313, opened 2018-09-11
[varLib] use designspaceLib, and depreacate varLib.designspace module

anthrotype:
fixes https://github.com/fonttools/fonttools/issues/1224


Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -346,6 +363,20 @@ def serialize(self):
             map=self.map,
         )
 
+    def map_forward(self, v):
+        from fontTools.varLib.models import piecewiseLinearMap
```
behdad: Move import to top level?

Issue #713, opened 2016-10-20
[feaLib etc.] Support for FeatureVariations?


twardoch:
I've opened a discussion about support for FeatureVariations and a more general support for OTVar in the Adobe FEA syntax over at https://github.com/adobe-type-tools/afdko/issues/153

I have some questions:
1. Are FeatureVariations (such as "rvrn") supported in fontTools at this time? (In the object model and .ttx)? 
2. I wonder what the opinion of the feaLib contributors is about the topics I've raised over at the afdko issue. 

makeOTF and feaLib are now two significant active implementations of FEA, while the FontForge one can probably be considered dead. 

In the past, Adobe did define the changes to the FEA syntax and implemented them. But with OTVar, it seems that @readroberts is open for more community-driven effort to figure out whether FEA needs to be extended, and if so, how, to support variations -- also because the best way would be if both makeOTF and feaLib implementers agree on the common way. 

The FEA spec is published under the Apache 2 license, so in theory it could diverge, but we surely don't want that:
http://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html

readroberts:
I strongly do not want to see different branches  of the feature file syntax appear. I also strongly support community contributions and comments, and the general idea that the fea syntax is common property, not an Adobe only playground, and that makeotf does not have to lead the way: as long as we all agree on the direction; others may well have the time to implement extensions first.

twardoch:
Thanks @readroberts , that's exactly what I imagined your response would be. :) 

I've just posted an alternative FEA syntax extension proposal for FeatureVariations (rvrn etc.) which I actually like. I believe it's "in the spirit" of the previous FEA constructs: 
https://github.com/adobe-type-tools/afdko/issues/153#issuecomment-255094780

As I've explained in the other thread, I think FeatureVariations are the one thing that would require a syntax extension. The other aspects can be done via "blending" of separate per-master FEA files, at least for the most common cases, as `fontmake` already does.

behdad:
We have some code for now, thanks to Just https://github.com/fonttools/fonttools/pull/1240

I need to hook it up to varLib.

anthrotype:
Fixed by #1314

PR #1314, opened 2018-09-12
[varLib] implement building GSUB FeatureVariations from DesignSpace rules

anthrotype:
I hooked up designspaceLib into varLib, and now we can translate the rule elements into equivalent list of conditional substitutions to be used with featureVars.addFeatureVariations.
This seems to be working, but I will add a test tomorrow to confirm.

Fixes https://github.com/fonttools/fonttools/pull/1240 and https://github.com/fonttools/fonttools/issues/713


behdad:
lgtm.

justvanrossum:
Thanks, Cosimo!

Issue #730, opened 2016-10-31
Expose bit 6 of the `glyf` first outline flags byte


twardoch:
My [160413-EvenOddTT](https://github.com/twardoch/test-fonts/tree/master/varia/160413-EvenOddTT) issue exposes a problem that the macOS Sierra rasterizer sometimes displays overlapping TT (`glyf`) contours as "white". This behavior can be suppressed if one sets "bit 6 of the first Outline Flag byte to 1". While the official Apple TrueType spec and the OpenType spec treat this bit as "reserved", Apple's introductory chapter to AAT explains that this bit is actually used to control the rendering of overlapping contours in the Apple rasterizer. Currently, no easy-to-use tools expose this bit, so setting it requires binary work. 

It would be useful if this bit was exposed to fontTools and/or to the `.ttx` dump. 

I'm not sure how to do it elegantly, especially since the `contour.pt` structure in fontTools is not a straight binary dump but is a bit prettified. Since the official spec treats the flag as reserved, and the AAT chapter only mentions the first flags byte of a glyph, an option in fontTools could trigger `.ttx` to dump/merge the following syntax, and of course a `.flag6` boolean attribute could be added to the `pt` object: 

```xml
<TTGlyph name="O" xMin="59" yMin="0" xMax="768" yMax="702">
      <contour>
        <pt x="674" y="440" on="1" flag6="1"/>
        <pt x="674" y="269" on="1"/>
        <pt x="140" y="269" on="1"/>
        <pt x="140" y="440" on="1"/>
      </contour>
      <instructions><assembly>
        </assembly></instructions>
    </TTGlyph>
```

A prettier implementation would be to introduce a boolean attribute on a `fontTools.ttLib.tables._g_l_y_f.Glyph` level, akin to `isComposite` but named `hasOvelaps` or `hasOverlapsAAT`, that would conditionally appear in the `.ttx` dump as an optional attribute of the `<TTGlyph>` element (if it is set in the font), and would be coordinated with the bit 6 of the first flag byte of the `glyf` strucutre for the glyph. The name `hasOverlapsAAT` would hint at its origin. 

```xml
<TTGlyph name="O" xMin="59" yMin="0" xMax="768" yMax="702" hasOverlapsAAT="1">
      <contour>
        <pt x="674" y="440" on="1"/>
        <pt x="674" y="269" on="1"/>
        <pt x="140" y="269" on="1"/>
        <pt x="140" y="440" on="1"/>
      </contour>
[...]

PR #1316, opened 2018-09-14
Add glyf flags bit6 to ttx output

jenskutilek:
Bit 6 of the flags of the first outline point in a glyph, indicating that there are overlaps in the glyph’s contours, got lost when roundtripping with ttx.

This patch preserves bit 6 in the most simple way, by adding it as an attribute of its contour point:

```xml
    <TTGlyph name="A" xMin="12" yMin="0" xMax="1172" yMax="1430">
      <contour>
        <pt x="501" y="1430" on="1" overlap="1"/>
        <pt x="683" y="1430" on="1"/>
        <pt x="1172" y="0" on="1"/>
        <pt x="983" y="0" on="1"/>
        <pt x="591" y="1193" on="1"/>
        <pt x="199" y="0" on="1"/>
        <pt x="12" y="0" on="1"/>
      </contour>
      <contour>
        <pt x="249" y="514" on="1"/>
        <pt x="935" y="514" on="1"/>
        <pt x="935" y="352" on="1"/>
        <pt x="249" y="352" on="1"/>
      </contour>
      <instructions/>
    </TTGlyph>
```

See issue #730.


Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -263,6 +263,9 @@ def __len__(self):
 flagReserved1 = 0x40
 flagReserved2 = 0x80
 
+# These flags are kept for XML output after decompiling the coordinates
+keepFlags = flagOnCurve + flagReserved1
```
khaledhosny: Shouldn’t `flagReserved1` be renamed now? It is no longer a reserved flag in the current version of the spec (`OVERLAP_SIMPLE` in https://docs.microsoft.com/en-us/typography/opentype/spec/glyf).

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -408,10 +411,15 @@ def toXML(self, writer, ttFont):
...
 				for j in range(last, self.endPtsOfContours[i] + 1):
-					writer.simpletag("pt", [
+					attrs = [
 							("x", self.coordinates[j][0]),
 							("y", self.coordinates[j][1]),
-							("on", self.flags[j] & flagOnCurve)])
+							("on", self.flags[j] & flagOnCurve),
+						]
+					if self.flags[j] & flagReserved1:
+						# Apple's AAT uses flagReserved1 in the first contour/first pt to flag glyphs that contain overlapping contours
```
khaledhosny: That is an Apple rasterizer feature rather than an AAT-specific one (i.e. not related to layout tables being used).

Lib/fontTools/ttLib/tables/_g_l_y_f.py
anthrotype: good point

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -441,7 +449,11 @@ def fromXML(self, name, attrs, content, ttFont):
 				if name != "pt":
 					continue  # ignore anything but "pt"
 				coordinates.append((safeEval(attrs["x"]), safeEval(attrs["y"])))
-				flags.append(not not safeEval(attrs["on"]))
+				flag = not not safeEval(attrs["on"])
+				if "overlap" in attrs:
+					f = safeEval(attrs["overlap"])
```
anthrotype: I would use the same trick as in the line above for the `on` flag, by negating twice to obtain a boolean, or just call `bool`.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -441,7 +449,11 @@ def fromXML(self, name, attrs, content, ttFont):
...
+					flag += f * flagReserved1
```
anthrotype: better use "or" bitwise operator instead of "+" here: `flag |= f * flagReserved1`
let's rename `flagReserved1` to OVERLAP_SIMPLE

Tests/ttLib/tables/_g_l_y_f_test.py
```diff
@@ -158,3 +162,63 @@ def test__checkFloat_overflow(self):
...
+HEAD_BIN = os.path.join(DATA_DIR, "_g_l_y_f_outline_flag_bit6.head.bin")
+LOCA_BIN = os.path.join(DATA_DIR, "_g_l_y_f_outline_flag_bit6.loca.bin")
+MAXP_BIN = os.path.join(DATA_DIR, "_g_l_y_f_outline_flag_bit6.maxp.bin")
+
+
+def strip_ttLibVersion(string):
+    return re.sub(' ttLibVersion=".*"', '', string)
+
+
+class glyfTableTest(unittest.TestCase):
```
anthrotype: the rest of the tests in this file already use pytest-style tests (with plain `assert` statements), I would prefer that over unittest-style tests.. but anyway some tests is better than none, thanks anyway.

Tests/ttLib/tables/_g_l_y_f_test.py
```diff
@@ -158,3 +162,63 @@ def test__checkFloat_overflow(self):
...
+        with open(GLYF_TTX, 'r') as f:
+            cls.glyfXML = strip_ttLibVersion(f.read()).splitlines()
+
+    def test_toXML(self):
+        font = TTFont(sfntVersion="\x00\x01\x00\x00")
+        glyfTable = font['glyf'] = newTable('glyf')
+        font['head'] = newTable('head')
+        font['loca'] = newTable('loca')
+        font['maxp'] = newTable('maxp')
+        font['maxp'].data = self.maxpData
```
anthrotype: why do you set `data` attribute here? I don't think it's needed

jenskutilek: I thought it was necessary, but actually isn’t. The only important thing is the tables need to be decompiled in the correct order because they depend on each other.

anthrotype: right

anthrotype:
oh .. and what about the similar OVERLAP_COMPOUND flag? Don't we want to expose that too?

jenskutilek:
> oh .. and what about the similar OVERLAP_COMPOUND flag? Don't we want to expose that too?

That flag is at least already preserved because components show the flags as hex numbers in ttx:

```xml
    <TTGlyph name="Adieresis" xMin="-6" yMin="0" xMax="1495" yMax="1845">
      <component glyphName="A" x="0" y="0" flags="0x604"/>
      <component glyphName="dieresis.cap" x="233" y="0" flags="0x4"/>
    </TTGlyph>
```

Perhaps we should display all the flags in more human readable form here?

anthrotype:
Right ok. I’d say let’s keep it as is for now

jenskutilek:
Thanks!

anthrotype:
Thank you!

PR #1317, opened 2018-09-15
Minor README tweaks

nedbat:
I hope this doesn't seem pedantic. I'm starting to investigate using fonttools, and noticed some typos.


anthrotype:
Thank you!

PR #1318, opened 2018-09-16
[feaLib] Skip noop class PairPos subtables

khaledhosny:
If Coverage is None then the subtable is a noop that does nothing and there is no point in building it. As a bonus, it makes OTS happy since it will reject subtables where coverageOffset is NULL (0), and the spec does not say whether this is allowed or not.


anthrotype:
Thanks!

PR #1319, opened 2018-09-18
WIP: merge CFF

anthrotype:
continuation of #1289

### DIFF
--- a/Lib/fontTools/merge.py
+++ b/Lib/fontTools/merge.py
@@ -387,6 +387,11 @@ def _glyphsAreSame(glyphSet1, glyphSet2, glyph1, glyph2):
 
 @_add_method(ttLib.getTableClass('CFF '))
 def merge(self, m, tables):
+	if any(hasattr(table, "FDSelect") for table in tables):
+		raise NotImplementedError(
+			"Merging CID-keyed CFF tables is not supported yet"
+		)
+
 	newcff = tables[0]
 	newfont = newcff.cff[0]
 	private = newfont.Private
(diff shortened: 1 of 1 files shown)


behdad:
Is this still valid?

anthrotype:
@BlackFoundry are you going to finish work on this? If you can't right now, it's ok. Just remove from the "Assignees" list, thanks.

BlackFoundry:
More likely later, beginning 2019, right now is quite busy. Could @behdad  help too?

BlackFoundry:
Next step would be to implement CIDFonts merging. Would be nice to have some Adobe CFF expert guidance.

readroberts:
@BlackFoundry Sure. The only real difference is that you have to track the FDArray list, and update the fdSelect index value in each charstring.  I would first merge all the charstrings specified, keeping track of the charstrings added. For each charstring that you added from the second font to the first, you need to check the FontDict referenced by the fdSelect index value.  IF you can find a a pre-existing FontDict in the first font's FDArray that you can use instead, you just need to update all the relevant charstrings to reference that FDArray, else you just add the second fonts's FontDict to the first font's FDArray. ( you can use any pre-existing FontDict with has then same PrivateDict values). When you are all done, the two fonts will be cross-linked - for example, any new FontDicts added to the first font will be pointing to PrivateDicts that are in the second font - but this will all be flattened correctly when you compile. Feel free to e-mail me directly, rroberts@adobe.com

anthrotype:
@BlackFoundry any updates on this?

indirectlylit:
refs #1319 I believe?

khaledhosny:
I’d like to get this merged without CID support since it is both less and less relevant these days and apparently no one is going to work on it anytime soon. @anthrotype anything you want to fix here other than CID support?

khaledhosny:
I have this locally rebased against main, but I had to drop the `.notdef` stuff it conflicts with #1729. Should I force-push to this branch and PR, or open a new one?

behdad:
> open a new one?

Do that.

khaledhosny:
https://github.com/fonttools/fonttools/pull/2447

PR #1323, opened 2018-09-24
[morx] Support AAT action type to insert glyphs

brawer:


behdad:
Thanks Sascha!

anthrotype:
Thanks Sascha! Welcome back :+1:

behdad:
FWIW, I already implemented Insertion in HarfBuzz.  So, looking forward to the tests!

PR #1326, opened 2018-09-25
Fix several ResourceWarning: unclosed file and some related improvement

BoboTiG:
Hello,

This is a little patch to fix several file descriptor leaks first, and also to prefer the use of the `with` context manager when dealing with files to ensure file descriptors are always cleanly closed.

I tried to separate commits as much as possible. Let me know if it needs more work/rebase or if some changes are not acceptable as is :)

About the change in `unicode.py` I think this patch could be better:
```diff
diff --git a/Lib/fontTools/unicode.py b/Lib/fontTools/unicode.py
index 00e13bb6..ef3ed695 100644
--- a/Lib/fontTools/unicode.py
+++ b/Lib/fontTools/unicode.py
@@ -4,7 +4,6 @@ from fontTools.misc.py23 import *
 def _makeunicodes(f):
 	import re
 	lines = iter(f.readlines())
-	f.close()
 	unicodes = {}
 	for line in lines:
 		if not line: continue
@@ -19,8 +18,11 @@ class _UnicodeCustom(object):
 
 	def __init__(self, f):
 		if isinstance(f, basestring):
-			f = open(f)
-		self.codes = _makeunicodes(f)
+			with open(f) as fd:
+				codes = _makeunicodes(fd)
+		else:
+			codes = _makeunicodes(f)
+		self.codes = codes
 
 	def __getitem__(self, charCode):
 		try:
```
WDYT?

### DIFF
--- a/Lib/fontTools/t1Lib/__init__.py
+++ b/Lib/fontTools/t1Lib/__init__.py
@@ -164,9 +164,8 @@ def readLWFN(path, onlyHeader=False):
 			elif code in [3, 5]:
 				break
 			elif code == 4:
-				f = open(path, "rb")
-				data.append(f.read())
-				f.close()
+				with open(path, "rb") as f:
+					data.append(f.read())
 			elif code == 0:
 				pass # comment, ignore
 			else:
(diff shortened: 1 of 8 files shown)


Lib/fontTools/t1Lib/__init__.py
```diff
@@ -179,43 +178,41 @@ def readLWFN(path, onlyHeader=False):
...
 
 def readOther(path):
 	"""reads any (font) file, returns raw data"""
-	f = open(path, "rb")
-	data = f.read()
-	f.close()
-	assertType1(data)
+	with open(path, "rb") as f:
+		data = f.read()
+		assertType1(data)
```
anthrotype: this line could be de-dented, doesn't need to be inside the with block

Lib/fontTools/ttx.py
```diff
@@ -362,7 +362,8 @@ def parseOptions(args):
 			output = makeOutputFileName(input, options.outputDir, extension, options.overWrite)
 			# 'touch' output file to avoid race condition in choosing file names
 			if action != ttList:
-				open(output, 'a').close()
+				with open(output, 'a') as f:
+					pass
```
anthrotype: the change in this line is mostly cosmetic, and I actually prefer the previous one as it's more compact

Lib/fontTools/t1Lib/__init__.py
```diff
@@ -179,43 +178,41 @@ def readLWFN(path, onlyHeader=False):
...
+				data.append(chunk)
+			elif code == 3:
+				break
+			else:
+				raise T1Error('bad chunk code: ' + repr(code))
+			if onlyHeader:
+				break
+		data = bytesjoin(data)
+		assertType1(data)
+		return data
```
anthrotype: these last three lines can also be un-indented as you're done with the file

Lib/fontTools/t1Lib/__init__.py
```diff
@@ -179,43 +178,41 @@ def readLWFN(path, onlyHeader=False):
...
-		else:
-			raise T1Error('bad chunk code: ' + repr(code))
-		if onlyHeader:
-			break
-	f.close()
-	data = bytesjoin(data)
-	assertType1(data)
-	return data
+	with open(path, "rb") as f:
+		data = []
```
anthrotype: maybe move this `data = []` just before the with context manager.

Lib/fontTools/t1Lib/__init__.py
anthrotype: actually the whole block after this line need not be indented under `with`.
in general I like to release the file resource as soon as possible and avoid too long indented blocks

Lib/fontTools/t1Lib/__init__.py
```diff
@@ -255,13 +251,10 @@ def writePFB(path, data):
 			f.write(longToString(len(chunk)))
 			f.write(chunk)
 		f.write(bytechr(128) + bytechr(3))
-	finally:
-		f.close()
```
anthrotype: the `try: finally:` was doing exactly the same thing, but if you prefer using `with`, I'm fine with it.

BoboTiG:
@anthrotype, thank you for the review. You are totally right and I applied changes as requested :)

BoboTiG:
@anthrotype what is the status? Do you think it is good enough to be merged?

anthrotype:
looks good, thanks!

Issue #1329, opened 2018-10-08
otf2ttf  OverflowError: unsigned short is greater than maximum


plutext:
I want to convert the Noto CJK fonts from OTF to TTF (so they work better with Microsoft Word).

Here's what happens with NotoSansCJKtc-Regular.otf

```
$ ./otf2ttf.py NotoSansCJKtc-Regular.otf -o NotoSansCJKtc-Regular.ttf
Traceback (most recent call last):
  File "./otf2ttf.py", line 91, in <module>
    sys.exit(main())
  File "./otf2ttf.py", line 87, in main
    font.save(output)
  File "/home/me/.local/lib/python3.7/site-packages/fontTools/ttLib/ttFont.py", line 173, in save
    writer_reordersTables = self._save(tmp)
  File "/home/me/.local/lib/python3.7/site-packages/fontTools/ttLib/ttFont.py", line 212, in _save
    self._writeTable(tag, writer, done, tableCache)
  File "/home/me/.local/lib/python3.7/site-packages/fontTools/ttLib/ttFont.py", line 633, in _writeTable
    tabledata = self.getTableData(tag)
  File "/home/me/.local/lib/python3.7/site-packages/fontTools/ttLib/ttFont.py", line 651, in getTableData
    return self.tables[tag].compile(self)
  File "/home/me/.local/lib/python3.7/site-packages/fontTools/ttLib/tables/_p_o_s_t.py", line 51, in compile
    data = data + self.encode_format_2_0(ttFont)
  File "/home/me/.local/lib/python3.7/site-packages/fontTools/ttLib/tables/_p_o_s_t.py", line 173, in encode_format_2_0
    indices.append(index)
OverflowError: unsigned short is greater than maximum
```

miguelsousa:
Have you tried using `--post-format 3` option?

plutext:
I hadn't, but using that does work, thanks!

```
$ ./otf2ttf.py NotoSansCJKtc-Regular.otf -o NotoSansCJKtc-Regular.ttf --post-format 3
```

miguelsousa:
https://github.com/googlei18n/cu2qu/issues/47#issuecomment-272054508

khaledhosny:
I think the otf2ttf.py should check if the number of glyphs does not fit in `post` table format 2 and automatically switch to format 3, probably with a warning.

PR #1330, opened 2018-10-08
[Snippets] Handle post table overflow in otf2ttf.py

khaledhosny:
Fixes https://github.com/fonttools/fonttools/issues/1329

### DIFF
--- a/Snippets/otf2ttf.py
+++ b/Snippets/otf2ttf.py
@@ -43,6 +43,7 @@ def otf_to_ttf(ttFont, post_format=POST_FORMAT, **kwargs):
     glyf.glyphOrder = glyphOrder
     glyf.glyphs = glyphs_to_quadratic(ttFont.getGlyphSet(), **kwargs)
     del ttFont["CFF "]
+    glyf.compile(ttFont)
 
     ttFont["maxp"] = maxp = newTable("maxp")
     maxp.tableVersion = 0x00010000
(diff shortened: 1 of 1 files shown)


Snippets/otf2ttf.py
```diff
@@ -56,12 +57,19 @@ def otf_to_ttf(ttFont, post_format=POST_FORMAT, **kwargs):
...
     post = ttFont["post"]
     post.formatType = post_format
     post.extraNames = []
     post.mapping = {}
     post.glyphOrder = glyphOrder
+    try:
+        post.compile(ttFont)
+    except OverflowError:
+        post.formatType = 3
+        print("Glyph names do not fit in 'post' table format 2, using format 3 instead.")
```
behdad: Maybe say "dropping glyph names"?

miguelsousa:
Shouldn't `logging` be used instead of `print`?

khaledhosny:
I tried logging first, but for some reason it didn’t show any difference than print; the output of `logger.warning()` and `print()` looked exactly the same, so I thought why bother.

anthrotype:
If you want the same logging formatting as the other fonttools scripts, you can import the configLogger function from fontTools.misc.loggingTools, which uses a custom formatter

khaledhosny:
Still does not make any difference for me, but I made the change anyway.

anthrotype:
Because configLogger is meant to be used to configure fontTools loggers, so by default it only configures loggers whose namespace starts with “fontTools”. You can either rename the logger to something like “fontTools.otf2ttf”, or pass your log instance to the “logger” keyword argument. It’s all explained in the docstring

khaledhosny:
Thanks, looks fine now.

Snippets/otf2ttf.py
```diff
@@ -1,13 +1,17 @@
 #!/usr/bin/env python
-from __future__ import print_function, division, absolute_import
```
anthrotype: maybe just leave absolute_import

khaledhosny: What would it do here?

anthrotype: on py2, would prevent from accidentally do a relative import without an explicit "." notation

anthrotype: in the case of this snippet it probably doesn't matter, but I prefer to always add absolute_import in my lib modules

anthrotype:
thanks khaled

Snippets/otf2ttf.py
behdad: Easier to just have all three there across the codebase, even if unused.

PR #1334, opened 2018-10-17
Add PointPen support to TT Glyph objects

jenskutilek:
See https://github.com/unified-font-object/ufo-spec/issues/74

Using Glyph.drawPoints() will make sure the point indices are not changed from the original. This is vital when the font has TrueType instructions which are based on point indices.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -1063,7 +1063,7 @@ class Glyph(object):
 			segmentType = "line" if cFlags[-1] == 1 else "qcurve"
 			for i, pt in enumerate(contour):
 				if cFlags[i] == 1:
-					pen.addPoint(pt, segmentType = segmentType)
+					pen.addPoint(pt, segmentType=segmentType)
 					segmentType = "line"
 				else:
 					pen.addPoint(pt)
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -1037,6 +1037,39 @@ class Glyph(object):
 					cFlags = cFlags[nextOnCurve:]
 			pen.closePath()
 
+	def drawPoints(self, pen, glyfTable, offset=0):
+		"""Draw the glyph using the supplied pointPen. Opposed to Glyph.draw(),
+		this will not change the point indices.
+		"""
+
+		if self.isComposite():
+			for component in self.components:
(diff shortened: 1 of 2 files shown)


Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -1037,6 +1037,39 @@ def draw(self, pen, glyfTable, offset=0):
...
+			coordinates.translate((offset, 0))
+		start = 0
+		for end in endPts:
+			end = end + 1
+			contour = coordinates[start:end]
+			cFlags = flags[start:end]
+			start = end
+			pen.beginPath()
+			# Start with the appropriate segment type based on the final segment
+			s = "line" if cFlags[-1] == 1 else "qcurve"
```
anthrotype: maybe find a better name for this variable "s", e.g. "segmentType"

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -755,12 +755,6 @@ def draw(self, pen):
 		"""
 		self._glyph.draw(pen)
 
-	def drawPoints(self, pen):
```
anthrotype: maybe you could leave this method in for _TTGlyphCFF class and raise NotImplementedError, so instead that an AttributeError the user would get a more meaningful error

jenskutilek:
I was thinking about tests ... but that would require a PointPen, which as far as I can see is not part of fontTools. Would it be OK to copy it from ufoLib just into the test module, or is it even possible to add dependencies just for tests?

anthrotype:
i was actually on working today on moving ufoLib inside fontTools.ufoLib. I would also move the pointPen module alongside it. 
Anyway, for the test you don't need to subclass the AbstractPointPen, just make a test pen object that has the same beginPath, addPoint, endPath interface

anthrotype:
it is possible to add dependencies just for the test, yes, but this is not the case I think

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -755,6 +755,10 @@ def draw(self, pen):
 		"""
 		self._glyph.draw(pen)
 
+	def drawPoints(self, pen):
+		# drawPoints is only implemented for _TTGlyphGlyf at this time.
+		raise NotImplementedError
```
anthrotype:
```suggestion
		raise NotImplementedError()
```

see https://www.python.org/dev/peps/pep-0317/

anthrotype: or you can put that inline comment as the message for the NotImplementedError exception, even better

PR #1333, opened 2018-10-11
Sphinx documentation expansion for continuous documentation on Read the Docs

chrissimpkins:
This PR expands on the efforts by @gferreira @anthrotype @belluzj @LettError @mashabow to create Sphinx formatted `*.rst` documentation for the fontTools project.  With this PR, all fontTools modules are now included in the Sphinx documentation available in the `Doc` directory and we *should*™ be ready to push to RTD for continuous documentation of the fontTools project.

The PR adds the following:

- new `*.rst` files for modules that were not previously documented with support for automation of source docstring parsing --> formatted documentation with paths organized to match the module paths in the source
- translation of the fontTools README to `*.rst` format for use as the home page (index.html) of the Sphinx documentation (with minor formatting updates)
- Sphinx extension support for local builds of documentation that match the RTD theme (see `Doc/README.md` for install/use details
- Sphinx [Napoleon extension](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/) support for automated parsing of Google Style Guide formatted Python docstrings to documentation files.  This is a much more human legible format than the default Sphinx docstring approach.
- new documentation README in the root of the `Doc` directory with docstring style guidelines, local build instructions, and documentation contributor guidelines
- new CC by-SA 4.0 International License for documentation (defined in `Doc/source/conf.py` file)

I took some liberties with formatting on the `*.rst` doc side but did not modify existing documentation in Python source files or make significant changes to the documentation text used for the translations of existing README text files to this format.  There were no changes to existing documentation in the project README or any other existing documentation files.  These changes support the start of the source documentation work and a great deal of new and modified docstring text will be necessary if we want to use this approach for comprehensive, continuous source documentation.  

For those who would like to build a local set of docs to view these changes, use the following instructions:

(1) setup a development environment for this branch of the fontTools repository and install doc build dependencies with:

```
pip install -e ".[all]"
pip install sphinx sphinx_rtd_theme
```

(2) navigate to the `Doc` directory of the fontTools repository and compile the documentation with one of the two following approaches (based upon availability of `make` on your system):

```
make html
```

or

```
sphinx-build -b html source build
```

The documentation is built on the path `Doc/build/html`.  Open the `index.html` file on that path in your browser to view the home page.

Here are a couple of previews of a local build so that you can see where we are:

### Docs Home

z40r5-image

### ttLib.ttFont module
1awdd-image-crunch

### TODO:

[...]

PR #1335, opened 2018-10-18
merge ufoLib into fontTools.ufoLib

anthrotype:
Part of #1095


anthrotype:
basically, I fetched from the upstream ufoLib's remote, moved files to Lib/fontTools, then removed all duplicate files (setup.py, etc.), and finally used `git merge --allow-unrelated-histories` to merge all git history from one repository to the other.
It's a lot of history, but the most relevant commits are the ones following https://github.com/fonttools/fonttools/pull/1335/commits/49c7bd1e1d08500c6ddb783df3ae61184982d89d, which is the last commit from ufoLib/master.

anthrotype:
I've laso copied the `*.rst` files from the ufoLib/Documentation to the Doc/sources/ufoLib subfolder, but didn't also copy the Sphinx setup, since there is #1333 pending which overlaps with this.

I reorganized a few modules and moved them in other places where they make more sense, but kept shims that re-export all the symbols from the old module location, to make the transition easier.
These are:

- `ufoLib.etree` => `fontTools.misc.etree`
- `ufoLib.plistlib` => `fontTools.misc.plistlib`
- `ufoLib.pointPen` => `fontTools.pens.pointPen`

Finally, I added `extras_require` to the setup.py script, to allow to specify the optional dependencies:
https://github.com/fonttools/fonttools/pull/1335/commits/783707644bfa0d92d924faea745e34f3c0555219

Currently, the optional dependencies are grouped under the following extras keys:
`ufo`, `woff`, `lxml`, `unicode`, `interpolatable`, `symfont`, `type1`
There is also a catch-all extra key, called simply `all`, that installs all the optional dependencies in one go.

Just a reminder: each of these strings can be added to the `pip install fonttools[...]` invocation (or when specifying fonttools as a dependency in one's setup.py). They go between the square brackets and are  separated by commas. E.g.:

`pip install fonttools[ufo,lxml,unicode]` will install fonttools plus all the optional dependencies needed to, in this case, read/write UFOs, using lxml (instead of ElementTree), and with unicodedata from current Unicode 11.

anthrotype:
So, to make this clear, the core of fontTools (e.g. ttx, ttLib, etc.) stays dependency-free. But in order to use the new `fontTools.ufoLib`, one needs to install the `fs` module (aka pyfilesystem2), and the `enum34` backport only if running under python2.7.
This is all specified under the `[ufo]` extra requirement, and the user only needs to remember to type `pip install fonttools[ufo]` iff they want to read/write UFOs with fontTools. Otherwise nothing changes for them.

chrissimpkins:
Will update #1333 with the new docs and rebase source (incl. setup.py) on your new version here.  We can add the ufo extras requirement to the RTD settings file so that all of those new dependencies are pulled for remote doc builds.

anthrotype:
@chrissimpkins thanks. let's first merge this then you can rebase on master

typesupply:
I will have time for a quick review mid-day tomorrow. Is that soon enough?

anthrotype:
that's great, thank you :+1:

anthrotype:
another legacy thing that I would love to get rid of, for the sake of consistency between lxml and ElementTree, is the custom xml declaration that uses double instead of single quotes.  Both lxml and ElementTree use single quotes:

```
<?xml version='1.0' encoding='UTF-8'?>
```

but in ufoLib we write it with double quotes because that's what it had always done.
This is the time to change these little annoying things.

benkiel:
@anthrotype I'm ok changing it, it's a one-time break to everyone's git histories. @miguelsousa you ok with such a thing? I know it's super annoying, but....

miguelsousa:
@anthrotype go ahead with all the cleanup/simplification/breakage. Now is the time.
After the merge, your UFO-breakage allowance will be freezed for 10 years 😄

benkiel:
Congrats @anthrotype!

anthrotype:
Thanks to all the ufoLib contributors!

PR #1344, opened 2018-10-23
make UFOWriter a subclass of UFOReader, use mixins for shared methods

anthrotype:
UFOWriter is already sharing some methods with UFOReader.
I think it would make sense to make it a subclass so that it extends the reader class with its additional writer-specific methods, effectively becoming a UFOReaderWriter.

This would allow, for example, to use the same UFOWriter instance for initializing a Font object, do stuff with it, and finally writing to it inplace.

Comments?

### DIFF
--- a/Lib/fontTools/ufoLib/__init__.py
+++ b/Lib/fontTools/ufoLib/__init__.py
@@ -466,7 +466,7 @@ class UFOReader(_UFOBaseIO):
 			raise UFOLibError("fontinfo.plist is not properly formatted.")
 		return data
 
-	def readInfo(self, info, validate=None):
+	def readInfo(self, info=None, validate=None):
 		"""
 		Read fontinfo.plist. It requires an object that allows
 		setting attributes with names that follow the fontinfo.plist
@@ -509,12 +509,17 @@ class UFOReader(_UFOBaseIO):
 		# validate data
 		if validate:
 			infoDataToSet = validateInfoVersion3Data(infoDataToSet)
-		# populate the object
-		for attr, value in list(infoDataToSet.items()):
-			try:
-				setattr(info, attr, value)
-			except AttributeError:
-				raise UFOLibError("The supplied info object does not support setting a necessary attribute (%s)." % attr)
+		if info is not None:
+			# populate the object
+			for attr, value in infoDataToSet.items():
+				try:
+					setattr(info, attr, value)
+				except AttributeError:
+					raise UFOLibError(
+						"The supplied info object does not support setting a "
+						"necessary attribute (%s)." % attr
+					)
+		return infoDataToSet
 
 	# kerning.plist
 
--- a/Lib/fontTools/ufoLib/__init__.py
+++ b/Lib/fontTools/ufoLib/__init__.py
@@ -5,6 +5,7 @@ from copy import deepcopy
 import logging
 import zipfile
 import enum
+from collections import OrderedDict
 import fs
 import fs.base
 import fs.subfs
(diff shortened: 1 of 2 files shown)


justvanrossum:
I think a combined reader/writer class makes sense. Perhaps an alias `UFOReaderWriter = UFOWriter` could be introduced, to make it possible for client code to more clearly show what's going on?

I am wondering whether it would be simpler to merge the Plist reader and writer mixins into one class.

anthrotype:
> I am wondering whether it would be simpler to merge the Plist reader and writer mixins into one class.

the reason they are split is that UFOReader only needs the _getPlist method, UFOWriter and GlyphSet need both _getPlist and _writePlist

justvanrossum:
> the reason they are split is that UFOReader only needs the _getPlist method, UFOWriter and GlyphSet need both _getPlist and _writePlist

I understand :) I was just wondering how bad it would be if the UFOReader class has a _writePlist method it will never use, especially if it makes the rest of the code a little simpler (one class less is one class less).

anthrotype:
ah ok! ;)
it's private (underscore prefixed) so I guess it's ok

anthrotype:
I'll put all the three shared methods under the same `_UFOBaseIO` class and use that as parent for all UFOReader, UFOWriter and GlyphSet

justvanrossum:
If merging UFOReader and UFOWriter has to terrible consequences for the read functionality, would it be an idea to just do:
```python
class UFOReaderWriter(...):
    ...

UFOReader = UFOReaderWriter  #b/w compat
UFOWriter = UFOReaderWriter  #b/w compat
```

anthrotype:
hm.... :thinking: 
i'll have to think about that.

anthrotype:
the `__init__` methods of UFOReader and UFOWriter are still very different from each other, and for good reasons. Maybe that's not a good idea. Anyway, I can't think straight now, it's late here. Let's keep thinking

justvanrossum:
Also: @typesupply's opinion on such a change would be valuable.

Originally, it made sense to have separate reader and writer classes. With all the common functionality fs and all) perhaps that idea makes less sense now. Not sure, just wondering.

justvanrossum:
Good point. Let's not move too fast here.

typesupply:
I've been trying to remember why we separated the reader and writer instead of merging the two, but I can't recall. `glifLib.GlyphSet` was already in existence and in use, so we had _some_reason for not following that same model. I remember discussing the split model on iChat with @justvanrossum way back in the day but I don't recall the contents of the conversation.

I'm not opposed to a unified reader/writer now. My only hesitation at the moment is that I haven't had time to think through [this issue](https://github.com/typesupply/defcon/pull/204#issuecomment-429320866) and, given the complexities of .zip, a unified reader/writer could make that even more difficult. It could also make it easier. I dunno, it makes my head hurt right now. I'm deep in drawing work right now so the code/logic part of my brain is non-functional.

anthrotype:
No problem Tal!
Anyway this PR as it stands is quite conservative, does’t change the public API (besides adding the UFOReaderWriter alias) not change the current behaviour of the existing reader and writer.
So I guess we can merge and continue to experiment with this until we familiarise with this new fs world?

typesupply:
> So I guess we can merge and continue to experiment with this until we familiarise with this new fs world?

Fine with me!

justvanrossum:
Hm, my comment was meant to go to the last commit, the one changing the return value of readInfo(). Just to be clear.

anthrotype:
Ok, no big deal, I’ll revert for now.

anthrotype:
@justvanrossum btw thanks for keeping an eye on this!
I guess I should merge after the CI passes, otherwise I'm tempted to keep tinkering.

justvanrossum:
Yeah, looks good so far, let's not add unrelated changes :)

behdad:
Why not just have one `UFOReaderWriter` object?

anthrotype:
> Why not just have one UFOReaderWriter object?

this PR is a cautious step towards that direction.

PR #1347, opened 2018-10-24
Fix normalization error when condition’s minimum/maximum are missing in designspace, update test too

### LINKED ISSUE #1346: Omitting maximum for a rule condition in a designspace document causes an error
When using rules in a designspace document.

```
<rules>
        <rule name="dollar-stroke">
            <conditionset>
                <condition name="weight" minimum="500" maximum="1000" />
            </conditionset>
            <sub name="uni0024" with="uni0024.nostroke" />
        </rule>
</rules>
```

When I omit the `maximum`, according to the spec the maximum of the corresponding axis should be used, however, I get an error during compilation with fontmake.

```
  File "/Users/davidbrezina/anaconda3/bin/fontmake", line 11, in <module>
    sys.exit(main())
  File "/Users/davidbrezina/anaconda3/lib/python3.6/site-packages/fontmake/__main__.py", line 266, in main
    project.run_from_designspace(designspace_path, **args)
  File "/Users/davidbrezina/anaconda3/lib/python3.6/site-packages/fontmake/font_project.py", line 646, in run_from_designspace
    **kwargs)
  File "/Users/davidbrezina/anaconda3/lib/python3.6/site-packages/fontmake/font_project.py", line 715, in run_from_ufos
    master_bin_dir=master_bin_dir)
  File "/Users/davidbrezina/anaconda3/lib/python3.6/site-packages/fontmake/font_project.py", line 268, in build_variable_font
    font, _, _ = varLib.build(designspace_path, finder)
  File "/Users/davidbrezina/anaconda3/lib/python3.6/site-packages/fontTools/varLib/__init__.py", line 750, in build
    _add_GSUB_feature_variations(vf, ds.axes, ds.internal_axis_supports, ds.rules)
  File "/Users/davidbrezina/anaconda3/lib/python3.6/site-packages/fontTools/varLib/__init__.py", line 591, in _add_GSUB_feature_variations
    maximum = normalize(axis_name, condition["maximum"])
  File "/Users/davidbrezina/anaconda3/lib/python3.6/site-packages/fontTools/varLib/__init__.py", line 575, in normalize
    {name: value}, internal_axis_supports
  File "/Users/davidbrezina/anaconda3/lib/python3.6/site-packages/fontTools/varLib/models.py", line 69, in normalizeLocation
    out[tag] = normalizeValue(v, triple)
  File "/Users/davidbrezina/anaconda3/lib/python3.6/site-packages/fontTools/varLib/models.py", line 19, in normalizeValue
    v = max(min(v, upper), lower)
TypeError: '<' not supported between instances of 'float' and 'NoneType'
```

MrBrezina:
Fixes #1346

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -389,7 +389,7 @@ def _merge_TTHinting(font, model, master_ttfs, tolerance=0.5):
 	# cvt table
 
 	all_cvs = [Vector(m["cvt "].values) for m in master_ttfs if "cvt " in m]
-	
+
 	if len(all_cvs) == 0:
 		# There is no cvt table to make a cvar table from, we're done here.
 		return
@@ -587,8 +587,14 @@ def _add_GSUB_feature_variations(font, axes, internal_axis_supports, rules):
 			space = {}
 			for condition in conditions:
 				axis_name = condition["name"]
-				minimum = normalize(axis_name, condition["minimum"])
-				maximum = normalize(axis_name, condition["maximum"])
+				if condition["minimum"]:
+					minimum = normalize(axis_name, condition["minimum"])
+				else:
+					minimum = -1.0
+				if condition["maximum"]:
+					maximum = normalize(axis_name, condition["maximum"])
+				else:
+					maximum = 1.0
 				tag = axis_tags[axis_name]
 				space[tag] = (minimum, maximum)
 			region.append(space)
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -587,11 +587,11 @@ def _add_GSUB_feature_variations(font, axes, internal_axis_supports, rules):
 			space = {}
 			for condition in conditions:
 				axis_name = condition["name"]
-				if condition["minimum"]:
+				if condition["minimum"] is not None:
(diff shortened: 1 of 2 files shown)


anthrotype:
yes, I think this is correct. Maybe @justvanrossum can also take a look

Lib/fontTools/varLib/__init__.py
```diff
@@ -587,8 +587,14 @@ def normalize(name, value):
 			space = {}
 			for condition in conditions:
 				axis_name = condition["name"]
-				minimum = normalize(axis_name, condition["minimum"])
-				maximum = normalize(axis_name, condition["maximum"])
+				if condition["minimum"]:
```
anthrotype: wait... I think you meant `if "minimum" in condition` here, no?

anthrotype: actually I don't remember, is the key always there but maybe None?

anthrotype: if you're checking for None, it's better to do that explicitly because as you know, in Python `0` or (`0.0`) mean False

anthrotype: ok, I checked and designspaceLib `_readConditionElements` method sets those to `None` when not specified, so you should check `if condition["minimum"] is not None`

PR #1348, opened 2018-10-24
Some languagesystems might not have any features assigned (FeatureIndex is missing)

MrBrezina:
Feel free to reject this. It seemed easier to make a PR than issue for this one.


anthrotype:
when does that happen?

MrBrezina:
I think (!) I have some languagesystems which are not used in any features yet or they end up empty after compilation/optimization of the code (they appear in the code, but they do not really make a difference). I guess this is primarily a problem with my unfinished features’ code, but I thought it should not crash the compilation.

anthrotype:
weird, I would expect that `FeatureIndex` attribute to be present (as an empt list)..

anthrotype:
Ah! Look at the fontTools.subset module. It is calling an `ensureDecompiled` method before attempting to access attributes from an otBase.BaseTable (mother class of all the otData tables).
Maybe that code in featureVars should do the same?

anthrotype:
@MrBrezina How can I reproduce the AttribuetError?

PR #1351, opened 2018-10-26
transform: add getters for xScale, xyScale, yxScale, yScale, xOffset …

anthrotype:
…and yOffset

UFO images have a transformation matrix with those names.
http://unifiedfontobject.org/versions/ufo3/glyphs/glif/
This makes the Transform look more like a namedtuple, in that it can be accessed both by index and by property name.

### DIFF
--- a/Lib/fontTools/misc/transform.py
+++ b/Lib/fontTools/misc/transform.py
@@ -345,6 +345,60 @@ class Transform(object):
 		return "<%s [%g %g %g %g %g %g]>" % ((self.__class__.__name__,) \
 				+ self.__affine)
 
+	@property
+	def xScale(self):
+		"""
+		>>> t = Transform(1, 2, 3, 4, 5, 6)
+		>>> t.xScale
+		1
+		"""
+		return self.__affine[0]
+
+	@property
+	def xyScale(self):
+		"""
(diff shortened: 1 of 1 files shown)


anthrotype:
the change is trivial, merging.

justvanrossum:
a) Why do you request my review if you're going to merge anyway? :) Why are you in such a rush?
b) Where do you actually need this? Usually one passes a transformation as a whole. I _never_ use indexing, except implicitly when unpacking:
`xScale, xyScale, yxScale, yScale, xOffset, yOffset = t`
c) These names do not match the keyword args of `__init__`

This change seems trivial, but was not thought trough enough and should be reconsidered.

justvanrossum:
d) Now there are two ways of accessing the elements: tuple unpacking and attrs. More ways of doing one thing means more possibility for confusion. Not a big fan.

anthrotype:
the change seemed innocuous to me. But it turns out I don't actually need it. I was confused because on the one hand the UFOWriter expects a Glyph.image to be a dict-like object containing those keys (plus "fileName" and "color") whereas the defcon's Image class has a "transformation" property getter that returns a 6-tuple from those.
I'll take it back, sorry.

anthrotype:
reverted in master

PR #1352, opened 2018-10-29
Some Pointpen fixes, and a point pen test file

justvanrossum:


anthrotype:
reminder to self: disable Appveyor builds from PR branches.
This is one of the reasons I always work from my anthrotype fork and send PRs from there.

Thanks Just, I'll take a look now

anthrotype:
i think the tests are failing because of the `u""` unicode literals on py27, as you're comparing strings. try to normalize them to native str type (using py23.tostr) or remove unicode_literals import if you prefer.

justvanrossum:
Sorry for the AppVeyor hassle, next time I'll try to set up a fork (I always get lost with the upstream stuff).

Tests/pens/pointPen_test.py
```diff
@@ -0,0 +1,408 @@
...
+        if identifier is not None:
+            items.append("identifier=%r" % identifier)
+        items.extend(_reprKwargs(kwargs))
+        self._commands.append("beginPath(%s)" % ", ".join(items))
+
+    def addPoint(self, pt, segmentType=None, smooth=False, name=None,
+                 identifier=None, **kwargs):
+        items = ["%s" % (pt,)]
+        if segmentType is not None:
+            items.append("segmentType=%r" % segmentType)
```
anthrotype: here, don't use `%r`, use just `%s`. the problem is the repr() of unicode objects in py27.

justvanrossum: By not using the unicode_literals directive in pointPen.py, the py27 tests seem to pass here locally.

justvanrossum: Not sure whether there may be other consequences of not using `unicode_literals` in pointPen.py. If you want I can use `'%s'` instead of `%r`. (Except the one in `_reprKwargs()`, but that's currently not even properly used/tested.)

anthrotype: hm, I would actually prefer this latter option. I like to have all text strings as unicode (more future proof)

anthrotype:
FYI, I set up Travis/Appveyor so that they will only run builds from either `master` branch (which includes PR to master branch) or from branches that start with the string `wip/`.
https://github.com/fonttools/fonttools/commit/139030e1fbd5abc11b4e368415f9670cb774c461

This is the same setup that @miguelsousa uses in afdko, and it prevents duplicate CI builds when working from PR branches on a shared upstream github repository.

anthrotype:
please merge once appveyor/pr builds passes (I cancelled the appveyor/branch build, no prob)

justvanrossum:
Thanks!

miguelsousa:
> This is the same setup that @miguelsousa uses in afdko, and it prevents duplicate CI builds when working from PR branches on a shared upstream github repository.

Correction: that setup is used in the [psautohint repo](https://github.com/adobe-type-tools/psautohint) and was put in place by @khaledhosny.
In the [afdko repo](https://github.com/adobe-type-tools/afdko) we use `skip_branch_with_pr: true` instead.

anthrotype:
That option is specific to Appveyor, I think. How do you do the same with Travis?

miguelsousa:
I don't. Travis runs the jobs in parallel, so there's no slowdown even if both checks (PR and branch) run. AppVeyor, on the other hand, runs everything sequentially (even across separate repos) so there was a big backlog when a PR got updates.

PR #1354, opened 2018-10-31
[WIP] FontBuilder

justvanrossum:
(I quickly demoed this at the FontTools meeting at Robothon in The Hague.)

This PR contains a new fontTools submodule called "fontBuilder", which contains a convenient wrapper class around TTFont called FontBuilder, making it a lot easier to create a working font from scratch with code. It supports TTF and OTF, and a little bit of variations.

Some features may be a little arbitrary, and possibly reflect what I needed the most at the time I wrote this.

I realize work may be needed before this can be accepted. It definitely needs some documentation/examples (at least in the module doc string).

### DIFF
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -1,6 +1,7 @@
 from __future__ import print_function, division, absolute_import
 from __future__ import unicode_literals
 
+from .misc.py23 import *
 from .ttLib import TTFont, newTable
 from .ttLib.tables._c_m_a_p import cmap_classes
 from .ttLib.tables._n_a_m_e import NameRecord, makeName
@@ -249,7 +250,11 @@ class FontBuilder(object):
                 nameID = nameName
             else:
                 nameID = _nameIDs[nameName]
-            nameTable.addMultilingualName(dict(en=nameValue), ttFont=self.font, nameID=nameID)
+            if not isinstance(nameValue, basestring):
+                nameValue, language = nameValue
+            else:
+                language = "en"
+            nameTable.addMultilingualName({language: nameValue}, ttFont=self.font, nameID=nameID)
 
     def setupOS2(self, **values):
         if "xAvgCharWidth" not in values:
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -6,12 +6,23 @@ __all__ = ["FontBuilder"]
 """
 This module is *experimental*, meaning it still may evolve and change.
 
-The `FontBuilder` class is a convenient helper to construct working TTF or OTF
-fonts from scratch.
+The `FontBuilder` class is a convenient helper to construct working TTF or
+OTF fonts from scratch.
 
-Note that the various setup method calls cannot be done in arbitrary order,
-due to various interdependencies between OpenType tables. See the examples
-below for an order that works.
+Note that the various setup methods cannot be called in arbitrary order,
+due to various interdependencies between OpenType tables. Here is an order
+that works:
+
+    fb = FontBuilder(...)
+    fb.setupGlyphOrder(...)
+    fb.setupCharacterMap(...)
+    fb.setupGlyf(...) --or-- fb.setupCFF(...)
+    fb.setupMetrics("hmtx", ...)
+    fb.setupHorizontalHeader()
+    fb.setupNameTable(...)
+    fb.setupOS2()
+    fb.setupPost()
+    fb.save(...)
 
 Here is how to build a minimal TTF:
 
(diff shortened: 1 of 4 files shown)


Lib/fontTools/fontBuilder.py
```diff
@@ -0,0 +1,548 @@
...
+nameStrings = dict(familyName=dict(en="HelloTestFont", nl="HalloTestFont"),
+                   styleName=dict(en="TotallyNormal", nl="TotaalNormaal"))
+nameStrings['psName'] = familyName + "-" + styleName
+
+pen = TTGlyphPen(None)
+drawTestGlyph(pen)
+glyph = pen.glyph()
+glyphs = {".notdef": glyph, "A": glyph, "a": glyph, ".null": glyph}
+fb.setupGlyf(glyphs)
+fb.calcGlyphBounds()
```
anthrotype: this call to `fb.calcGlyphBounds()` is not needed here, no? The `setupGlyf` method does it already by default.

Lib/fontTools/fontBuilder.py
```diff
@@ -0,0 +1,548 @@
...
+
+    def setupOS2(self, **values):
+        if "xAvgCharWidth" not in values:
+            gs = self.font.getGlyphSet()
+            widths = [gs[glyphName].width for glyphName in gs.keys() if gs[glyphName].width > 0]
+            values["xAvgCharWidth"] = int(round(sum(widths) / float(len(widths))))
+        self._initTableWithValues("OS/2", _OS2Defaults, values)
+        if not ("ulUnicodeRange1" in values or "ulUnicodeRange2" in values or
+                "ulUnicodeRange3" in values or "ulUnicodeRange3" in values):
+            assert "cmap" in self.font
```
anthrotype: it would be good to document any of these inter-table dependencies, so that it's clear what's the expected order that one may call the various setupTable methods.

Lib/fontTools/fontBuilder.py
```diff
@@ -0,0 +1,548 @@
...
+            postTable.formatType = 3.0
+
+    def setupMaxp(self):
+        if self.isTTF:
+            defaults = _maxpDefaultsTTF
+        else:
+            defaults = _maxpDefaultsOTF
+        self._initTableWithValues("maxp", defaults, {})
+
+    def setupDSIG(self):
```
anthrotype: hm, is this really needed? I've been reluctant in adding support for similar "dummy" DSIG in fontmake/ufo2ft, as I think this is a dirty hack, which is easy to do anyway as a post-process step if really necessary. We need to deprecate the table for good, instead of keeping it "alive" even though invalid.

justvanrossum: Well, it's not called by default, but it is definitely needed in some cases. I believe MS Word doesn't do OT features in a TTF unless there's a DSIG. Or something like that.

It's one of those ugly legacy things, and since FontBuilder is mostly about convenience, I think it's a good thing to have that functionality here.

justvanrossum:
Yeah, the expected order of calls needs to be documented. It's indeed a weak point of this design.

I'd be fine marking it as experimental and merging it as is.

Lib/fontTools/fontBuilder.py
anthrotype: > I believe MS Word doesn't do OT features in a TTF unless there's a DSIG. Or something like that.

I'd be nice if somebody could verify this is still the case in current versions of MS Office and let us know. I tried several times in the OT mailing list to raise this issue about DSIG table spec being insufficient to implement, if not even incorrect (MS signtool utility seems to calculate checksums differently than what's written in the spec), but without success. But I gave up on DSIG long time ago.

anthrotype:
ok let's merge it then!

Lib/fontTools/fontBuilder.py
justvanrossum: Yeah, DSIG is a mess, a pain, and should die. However, there remains a need to produce fonts that work in legacy Office versions.

How about I rename `setupDSIG()` to `setupDummyDSIG()` and add a doc string saying something about the legacy status of this?

anthrotype: that would be better, I think, yeah

justvanrossum:
Do you prefer a "Squash and merge" or a normal merge commit?

anthrotype:
if you like, you could locally squash into a smaller set of logical commits, push --force and then merge.
I don't like much the automatic "squash and merge" thing. Up to you, I don't mind if you leave the intermediate commits.

anthrotype:
I'll update the changelog now and then cut a 3.32.0 release. Is that ok?

justvanrossum:
No problem. I need to write some var font tests, but that won't likely be today, so let's not wait for that.

PR #1357, opened 2018-11-01
FontBuilder improvements

justvanrossum:
- some doc strings
- new setupHorizontalMetrics and setupVerticalMetrics methods
- test for variations (but: the variation support is very minimal and low level, I need to figure out how varLib could be involved here)

### DIFF
--- a/Tests/fontBuilder/fontBuilder_test.py
+++ b/Tests/fontBuilder/fontBuilder_test.py
@@ -64,7 +64,7 @@ def test_build_ttf(tmpdir):
     glyphTable = fb.font["glyf"]
     for gn, advanceWidth in advanceWidths.items():
         metrics[gn] = (advanceWidth, glyphTable[gn].xMin)
-    fb.setupMetrics("hmtx", metrics)
+    fb.setupHorizontalMetrics(metrics)
 
     fb.setupHorizontalHeader()
     fb.setupNameTable(nameStrings)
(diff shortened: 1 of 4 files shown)


Tests/fontBuilder/fontBuilder_test.py
```diff
@@ -111,3 +113,81 @@ def test_build_otf(tmpdir):
...
+
+    fb.setupOS2()
+    fb.setupPost()
+    fb.setupDummyDSIG()
+
+    fb.save(outPath)
+
+    f = TTFont(outPath)
+    f.saveXML(outPath + ".ttx")
+    f.saveXML('test_var.ttf.ttx')  # XXXX
```
anthrotype: is this saved in the current directory? tmpdir would be better

Tests/fontBuilder/fontBuilder_test.py
```diff
@@ -111,3 +113,81 @@ def test_build_otf(tmpdir):
...
+    rightDeltas = [(0, 0), (0, 0), (200, 0), (200, 0), None, None, None, None]
+    upDeltas = [(0, 0), (0, 200), (0, 200), (0, 0), None, None, None, None]
+    downDeltas = [(0, -200), (0, 0), (0, 0), (0, -200), None, None, None, None]
+    variations['a'] = [
+        TupleVariation(dict(RGHT=(0, 1, 1)), rightDeltas),
+        TupleVariation(dict(LEFT=(0, 1, 1)), leftDeltas),
+        TupleVariation(dict(UPPP=(0, 1, 1)), upDeltas),
+        TupleVariation(dict(DOWN=(0, 1, 1)), downDeltas),
+    ]
+    fb.setupGvar(variations)
```
anthrotype: nice test case 👍

Tests/fontBuilder/fontBuilder_test.py
justvanrossum: Oops, that line shouldn't have been there. Fixed in master.

PR #1358, opened 2018-11-02
Fix python3 file writing issues and table compression

mhosken:
This fixes a number of python3 issues affecting Graphite tables particularly regarding writing binary fonts.

The only issue outside of Graphite tables is that under python3 writing an sfnt resulted in a table length including the padding 0 bytes (to a 4 byte boundary) when it should not.

### DIFF
--- a/Lib/fontTools/ttLib/sfnt.py
+++ b/Lib/fontTools/ttLib/sfnt.py
@@ -556,7 +556,6 @@ def calcChecksum(data):
 		3655064932
 	"""
 	remainder = len(data) % 4
-	data = data[:]          # py3 data is a bytearray so copy it before modifying
 	if remainder:
 		data += b"\0" * (4 - remainder)
 	value = 0
--- a/Lib/fontTools/ttLib/sfnt.py
+++ b/Lib/fontTools/ttLib/sfnt.py
@@ -556,6 +556,7 @@ def calcChecksum(data):
 		3655064932
 	"""
 	remainder = len(data) % 4
+	data = data[:]          # py3 data is a bytearray so copy it before modifying
 	if remainder:
 		data += b"\0" * (4 - remainder)
 	value = 0
--- a/Lib/fontTools/ttLib/tables/G__l_a_t.py
+++ b/Lib/fontTools/ttLib/tables/G__l_a_t.py
@@ -141,8 +141,8 @@ class table_G__l_a_t(DefaultTable.DefaultTable):
     def compileAttributes12(self, attrs, fmt):
         data = b""
         for e in grUtils.entries(attrs):
-            data += sstruct.pack(fmt, {'attNum' : e[0], 'num' : e[1]})
-            data += struct.pack(('>%dh' % len(e[2])), *e[2])
+            data += sstruct.pack(fmt, {'attNum' : e[0], 'num' : e[1]}) + \
+                    struct.pack(('>%dh' % len(e[2])), *e[2])
         return data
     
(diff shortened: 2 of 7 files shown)


Lib/fontTools/ttLib/sfnt.py
```diff
@@ -556,6 +556,7 @@ def calcChecksum(data):
 		3655064932
 	"""
 	remainder = len(data) % 4
+	data = data[:]          # py3 data is a bytearray so copy it before modifying
```
anthrotype: no, data is an immutable `bytes` object, it's not a mutable `bytearray` as the comment incorrectly is saying

Lib/fontTools/ttLib/tables/G__l_a_t.py
```diff
@@ -139,11 +139,11 @@ def compile(self, ttFont):
...
 
     def compileAttributes12(self, attrs, fmt):
-        data = []
+        data = b""
         for e in grUtils.entries(attrs):
-            data.extend(sstruct.pack(fmt, {'attNum' : e[0], 'num' : e[1]}))
-            data.extend(struct.pack(('>%dh' % len(e[2])), *e[2]))
-        return "".join(data)
+            data += sstruct.pack(fmt, {'attNum' : e[0], 'num' : e[1]})
+            data += struct.pack(('>%dh' % len(e[2])), *e[2])
```
anthrotype: appending and then joining once at the end of the loop is generally more efficient.
there's a `bytesjoin` function in fontTools.misc.py23 which you could use here

Lib/fontTools/ttLib/tables/grUtils.py
```diff
@@ -13,7 +13,7 @@ def decompress(data):
     if scheme == 0:
         pass
     elif scheme == 1 and lz4:
-        res = lz4.decompress(struct.pack("<L", size) + data[8:])
+        res = lz4.block.decompress(struct.pack("<L", size) + data[8:])
```
anthrotype: I must confess, this is the first time I notice that this Graphite tables in fontTools require `lz4` compression library
Is this the one you are using here?
https://pypi.org/project/lz4/
https://python-lz4.readthedocs.io/en/stable/

If so we need to document it in the README, like we do with other optional dependencies, and also add a new `[graphite2]` (or just graphite) "extra" in setup.py, in the `extras_require` setup option.

anthrotype:
> under python3 writing an sfnt resulted in a table length including the padding 0 bytes

really? how I can reproduce this?

Lib/fontTools/ttLib/sfnt.py
mhosken: I agree. I'll take this out. I was passing a bytearray from another program. So the bug is in the call site not here. Sorry about that.

Lib/fontTools/ttLib/tables/G__l_a_t.py
mhosken: Yeh. I did that as a lazy man's pep8. But I notice that line length isn't a strong value. Anyway I have tidied this up. I don't think a malloc + memcpy is that expensive over managing an array.

Lib/fontTools/ttLib/tables/grUtils.py
mhosken: yes. Will do.

mhosken:
You can't from within fonttools. Hence removing the code.

anthrotype:
@mhosken thanks

Issue #930, opened 2017-04-19
Improve BCP-47 matching in name table


behdad:
Currently name table thinks "fa-IR" cannot be encoded in Windows names, because the table only has a "fa" entry.  How should this be resolved?  Keep dropping last item in the langcode while there's no match?

brawer:
Unicode defines an algorithm to minimize BCP47 language identifiers, with supporting data in CLDR. For example, this will transform all of `fa-IR`, `fa-Arab` and `fa-Arab-IR` to `fa`. The algorithm preserves some subtags; for example, `fa-Arab-AF` gets mapped to `fa-AF` because Iran (and not Afghanistan) is assumed to be the default region for the Farsi language.

ICU has implemented this algorithm, so we could call ICU. Alternatively, we could write a pure Python implementation somewhere in fonttools.misc. I think we should do the latter, because the algorithm isn’t all that complicated, and ICU would add a rather hefty dependency to fonttools.

I wouldn’t recommend blindly stripping language tags until we find a match in the old-style enums. The very point of BCP47 is to go beyond limited enums, and the reason for adding `ltag` to TrueType (respectively `name` format 1 to OpenType) was to support arbitrary BCP47 tags.

behdad:
I agree a fontTools.bcp47 would be nice.

brawer:
Here’s two independent Python implementations. They both seem a bit heavy to me, but I’m putting the links here for reference.

* https://github.com/OnroerendErfgoed/language-tags
* https://github.com/LuminosoInsight/langcodes/

anthrotype:
I see that varLib's fvar builder is not using that `addMultilingualNames` method yet, and only adds English names (via `addNames` method). For the time being, couldn't we just use the hard-coded mappings we already have, or do must we wait for full support for BCP-47 tags normalization?

brawer:
Agree, I think the current function is OK for now. It can always be improved later.

anthrotype:
The `addMultilingualNames` method adds Macintosh (platform 1) nameIDs only when it can't find a match for a given BCP-47 tag among Windows (platform 3) language ids.

The `addNames` method currently used by default adds both Mac and Windows (English) nameIDs.

If we un-comment that line and use the former instead of the latter, that means we would no longer add both sets of nameIDs for fvar axes names, but only one set (most often, the Windows ones).

[...]

PR #1359, opened 2018-11-02
[name] make addMultilingualNames() add mac names by default

justvanrossum:
As discussed in #930.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_n_a_m_e.py
+++ b/Lib/fontTools/ttLib/tables/_n_a_m_e.py
@@ -161,7 +161,8 @@ class table__n_a_m_e(DefaultTable.DefaultTable):
 			raise ValueError("nameID must be less than 32768")
 		return nameID
 
-	def addMultilingualName(self, names, ttFont=None, nameID=None):
+	def addMultilingualName(self, names, ttFont=None, nameID=None,
+	                        windows=True, mac=True):
 		"""Add a multilingual name, returning its name ID
 
 		'names' is a dictionary with the name in multiple languages,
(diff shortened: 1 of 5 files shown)


Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -184,15 +188,16 @@ def addMultilingualName(self, names, ttFont=None, nameID=None):
...
-			else:
+			if windows:
+				windowsName = _makeWindowsName(name, nameID, lang)
+				if windowsName is not None:
+					self.names.append(windowsName)
+				else:
+					# We cannot not make a Windows name: make sure we add a
+					# Mac name as a fallback. This can happen for exotic
+					# BCP47 language tags that have no Windows language code.
+					mac = True
```
anthrotype: what if there is neither windows nor mac compatible name? shouldn't we error, or return None or something?

justvanrossum:
Hm, not sure why the tests are failing, maybe I messed something up locally when merging #1357

Lib/fontTools/ttLib/tables/_n_a_m_e.py
justvanrossum: Good question. In the previous revision this situation passed silently also. Leaning towards an error.

anthrotype:
@justvanrossum what's the status of this? Can it be merged?

this TODO in varLib may be related:
https://github.com/fonttools/fonttools/blob/1fd9bf6c298c7aaceb7b87bb0d4b71041f2af740/Lib/fontTools/varLib/__init__.py#L78-L83

justvanrossum:
Can be merged as far as I'm concerned, thanks!

PR #1361, opened 2018-11-05
Ignore the length of kern table sutable format 0

### LINKED ISSUE #314: ignore length of the last 'kern' subtable"

moyogo:
The "length" value of kern table subtable format 0 is ignored since some fonts (like OpenSans and Calibri) have a subtable larger than its value.

Fixes #314.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_k_e_r_n.py
+++ b/Lib/fontTools/ttLib/tables/_k_e_r_n.py
@@ -48,6 +48,19 @@ class table__k_e_r_n(DefaultTable.DefaultTable):
 				# This "version" is always 0 so we ignore it here
 				_, length, subtableFormat, coverage = struct.unpack(
 					">HHBB", data[:6])
+				if nTables == 1 and subtableFormat == 0:
+					# The "length" value is ignored since some fonts
+					# (like OpenSans and Calibri) have a subtable larger than
+					# its value.
+					nPairs, = struct.unpack(">H", data[6:8])
+					calculated_length = (nPairs * 6) + 14
+					if length != calculated_length:
+						log.warning(
+							"'kern' subtable longer than defined: "
+							"%d bytes instead of %d bytes" %
+							(calculated_length, length)
+						)
+					length = calculated_length
 			if subtableFormat not in kern_classes:
 				subtable = KernTable_format_unkown(subtableFormat)
 			else:
@@ -128,7 +141,6 @@ class KernTable_format_0(object):
 			">HHHH", data[:8])
 		data = data[8:]
 
-		nPairs = min(nPairs, len(data) // 6)
 		datas = array.array("H", data[:6 * nPairs])
 		if sys.byteorder != "big": datas.byteswap()
 		it = iter(datas)
@@ -153,6 +165,7 @@ class KernTable_format_0(object):
 	def compile(self, ttFont):
 		nPairs = len(self.kernTable)
 		searchRange, entrySelector, rangeShift = getSearchRange(nPairs, 6)
+		searchRange &= 0xFFFF
 		data = struct.pack(
 			">HHHH", nPairs, searchRange, entrySelector, rangeShift)
 
@@ -175,6 +188,10 @@ class KernTable_format_0(object):
 		if not self.apple:
 			version = 0
 			length = len(data) + 6
+			if length >= 0x10000:
+				log.warning('"kern" subtable overflow, '
+							'truncating length value while preserving pairs.')
+				length &= 0xFFFF
 			header = struct.pack(
 				">HHBB", version, length, self.format, self.coverage)
 		else:
--- a/Tests/ttLib/tables/_k_e_r_n_test.py
+++ b/Tests/ttLib/tables/_k_e_r_n_test.py
@@ -5,6 +5,7 @@ from fontTools.ttLib.tables._k_e_r_n import (
     KernTable_format_0, KernTable_format_unkown)
 from fontTools.misc.textTools import deHexStr
 from fontTools.misc.testTools import FakeFont, getXML, parseXML
+import itertools
 import pytest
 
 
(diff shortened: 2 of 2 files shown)


Lib/fontTools/ttLib/tables/_k_e_r_n.py
```diff
@@ -48,6 +48,12 @@ def decompile(self, data, ttFont):
 				# This "version" is always 0 so we ignore it here
 				_, length, subtableFormat, coverage = struct.unpack(
 					">HHBB", data[:6])
+				if subtableFormat == 0:
+					# The "length" value is ignored since some fonts
+					# (like OpenSans and Calibri) have a subtable larger than
+					# its value.
+					nPairs,  = struct.unpack(">H", data[6:8])
+					length = (nPairs * 6) + 14
```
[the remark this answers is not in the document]
moyogo: I guess, since these fonts only have one subtable.
Like this?
```python
 				if len(nTables) == 1 and subtableFormat == 0: 
```

Lib/fontTools/ttLib/tables/_k_e_r_n.py
```diff
@@ -153,6 +158,7 @@ def decompile(self, data, ttFont):
 	def compile(self, ttFont):
 		nPairs = len(self.kernTable)
 		searchRange, entrySelector, rangeShift = getSearchRange(nPairs, 6)
+		searchRange = min(searchRange, 32768)
```
anthrotype: @moyogo why did you clamp searchRange to maximum 32768?

searchRange is defined as:

> The largest power of two less than or equal to the value of nPairs, multiplied by the size in bytes of an entry in the table.

The size of an entry is 6. The searchRange is a uint16, so it cannot exceed 0xFFFF (65535).
I think the maximum searchRange that we should allow is `(2 ** 13) * 6`, which is 0xC000 (49152).
The next power of two would be bigger than the maximum unsigned short: `(2 ** 14) * 6 = 0x18000` (98304).

anthrotype: actually, it doesn't make sense to only clamp searchRange, as the latter always goes hand in hand with the other two (entrySelector and rangeShift).

The maximum number of pairs that still produce a valid searchRnge, entrySelector and rangeShift triplet is 0x3FFF (16383). Anything greater than that will overflow USHORT

```python
>>> from fontTools.ttLib import getSearchRange
>>> getSearchRange(16383, 6)
(49152, 13, 49146)

>>> getSearchRange(16384, 6)
(98304, 14, 0)


```

Lib/fontTools/ttLib/tables/_k_e_r_n.py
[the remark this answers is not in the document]
anthrotype: ok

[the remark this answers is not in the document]
anthrotype: I think we should warn here while decompiling such kern subtables, if the length computed from nPairs exceeds the one reported in the header

Lib/fontTools/ttLib/tables/_k_e_r_n.py
moyogo: Indeed, I should have looked at this better we don’t want to clamp at 32768. OpenSans and calibri fit 0x18000 into 0xFFFF, which is 0x8000. So I’ll do searchRange = searchRange % 0x10000.

Lib/fontTools/ttLib/tables/_k_e_r_n.py
```diff
@@ -48,6 +48,16 @@ def decompile(self, data, ttFont):
...
 					">HHBB", data[:6])
+				if nTables == 1 and subtableFormat == 0:
+					# The "length" value is ignored since some fonts
+					# (like OpenSans and Calibri) have a subtable larger than
+					# its value.
+					nPairs,  = struct.unpack(">H", data[6:8])
+					log.warning(
+						"'kern' subtable longer than defined: "
+						"%d bytes instead of %d bytes",
+						(nPairs * 6) + 14, length)
```
anthrotype: @moyogo you only want to print this warning when the `length` stored in the kern subtable is actually different from the one computed from `(nPairs * 6) + 14`. Here it is being logged all the time, whenever nTables == 1 and subtableFormat == 0 (i.e. 99 % of kern tables).

anthrotype:
@moyogo are you sure that MS Calibri has such a kern subtable with more pairs than can fit the length field?
I just tried to dump one copy I found on my VM, Calibri 2.10, and it looks ok.

Lib/fontTools/ttLib/tables/_k_e_r_n.py
anthrotype: if you use modulo like that, then you may be setting searchRange to an incorrect value. 65535 cannot be divided by 6 (the size of the kern pair records).
I am not sure whether those three related fields, searchRange, entrySelector and rangeShift are there only for optimizing the binary search, or what would happen if they don't match exactly the wording of the spec. That's why I suggested to not allow more than 16383, which is the maximum number that still produces a valid searchRnge, entrySelector and rangeShift triplet (one that does not overflow USHORT)

anthrotype:
correction: I was testing Calibri Light, which doesn't have this kern overflowing issue. The Regular style does show this problem.

Lib/fontTools/ttLib/tables/_k_e_r_n.py
anthrotype: >> I’ll do searchRange = searchRange % 0x10000.

> if you use modulo like that, then you may be setting searchRange to an incorrect value. 

@moyogo sorry, I need to correct myself. Doing `n % 0x10000` as you originally suggested does exactly the same thing as doing `n & 0xFFFF`(i.e. truncates n to a 16-bit unsigned integer). Apologies for the confusion.

PR #1363, opened 2018-11-06
italicAngle should be set to -(slnt value)

nyshadhr9:
slnt angle is measured clockwise while italicAngle is measured counter-clockwise (as per spec). 

https://docs.microsoft.com/en-us/typography/opentype/spec/post#header
https://drafts.csswg.org/css-fonts-4/#valdef-font-style-oblique-angle

@anthrotype does that seem right?


behdad:
Nope.  Both are supposed to be same direction.  The bug you are probably hitting is in Chrome / CSS.  @drott can share details.

anthrotype:
https://github.com/w3c/csswg-drafts/issues/3091#issuecomment-431780259

Issue #1370, opened 2018-11-09
VF-ttf goes from ~330kb to ~3mb after simple addFeatureVariation


weiweihuanghuang:
This seems excessive.

[Archive.zip](https://github.com/fonttools/fonttools/files/2564941/Archive.zip)

justvanrossum:
You need to use only one substitution dict per condition:

```python
### Based on a script by Stephen Nixon

import os
import sys
import fontTools
from fontTools.ttLib import TTFont
from fontTools.varLib.featureVars import addFeatureVariations

inputTTF = sys.argv[1]
# style = sys.argv[2]

f = TTFont(inputTTF)

condSubst = [	  
    ([{"wght": (0.158, 0.564)}], {"uniF8FF": "uniF8FF.001"}),  
	([{"wght": (0.564, 1)}], {"Adieresis.titl": "Adieresis.titl.136_230",
	                          "Udieresis.titl": "Udieresis.titl.136_230"}),
    ([{"wght": (0.424, 1)}], {"Odieresis.titl": "Odieresis.titl.106_230"}),
    ([{"wght": (0.868, 1)}], {"uni20B5": "uni20B5.201_230",
                              "cent": "cent.201_230",
                              "colonmonetary": "colonmonetary.201_230",
                              "dollar": "dollar.201_230",
                              "uni20B2": "uni20B2.201_230",
                              "dollar.tf": "dollar.tf.201_230",
                              "cent.tf": "cent.tf.201_230"}),
    ([{"wght": (0.71, 1)}], {"uni20A6": "uni20A6.169_230",
                             "peseta": "peseta.169_230",
                             "uni20A9": "uni20A9.169_230"}),
    ([{"wght": (0.564, 1)}], {"uni20B1": "uni20B1.136_230"}),
    ([{"wght": (0.424, 1)}], {"uni2761": "uni2761.79_230",
                              "paragraph": "paragraph.79_230"}),
]

addFeatureVariations(f, condSubst)

outputTTF = inputTTF.replace('.ttf', '-swap.ttf')
f.save(outputTTF)
```

This adds a mere 2336 bytes to the font.

This could possibly be automated by `addFeatureVariations()`. Let's keep this issue open.

weiweihuanghuang:
I see, thank you for the explanation!

behdad:
Yeah the space partition code should be smarter.

Even without space partitioning, maybe we should just hack the current code to merge same ranges.

behdad:
[...]

PR #1372, opened 2018-11-09
[varLib] Far faster featureVars

behdad:
Motivated by https://github.com/fonttools/fonttools/issues/1370

### DIFF
--- a/Lib/fontTools/varLib/featureVars.py
+++ b/Lib/fontTools/varLib/featureVars.py
@@ -284,8 +284,10 @@ def overlayBox(top, bot):
 
     remainder = dict(bot)
     exactlyOne = False
+    fullyInside = False
     for axisTag in bot:
         if axisTag not in intersection:
+            fullyInside = False
             continue # Lies fully within
         min1, max1 = intersection[axisTag]
         min2, max2 = bot[axisTag]
@@ -301,6 +303,7 @@ def overlayBox(top, bot):
         if exactlyOne:
             return intersection, bot
         exactlyOne = True
+        fullyInside = False
 
         # Otherwise, cut remainder on this axis and continue.
         if max1 < max2:
@@ -316,6 +319,10 @@ def overlayBox(top, bot):
             return intersection, bot
 
         remainder[axisTag] = minimum,maximum
+
+    if fullyInside:
+        # bot is fully within intersection.  Remainder is empty.
+        return intersection, None
 
     return intersection, remainder
 
--- a/Lib/fontTools/varLib/featureVars.py
+++ b/Lib/fontTools/varLib/featureVars.py
@@ -317,6 +317,10 @@ def overlayBox(top, bot):
 
         remainder[axisTag] = minimum,maximum
 
+    if not exactlyOne:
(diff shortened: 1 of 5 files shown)


behdad:
While the merging of duplicate regions fixed https://github.com/fonttools/fonttools/issues/1370, I like to rewrite the algorithm to be roughly O(num-out-regions x num-input-regions) region operations instead of current more than O(2^num-input-regions) region operations.

Lib/fontTools/varLib/featureVars.py
```diff
@@ -29,22 +29,62 @@ def addFeatureVariations(font, conditionalSubstitutions):
     The minimum and maximum values are expressed in normalized coordinates.
 
     A Substitution is a dict mapping source glyph names to substitute glyph names.
+
+    Example:
+
+    # >>> f = TTFont(srcPath)
+    #>>> condSubst = [
```
justvanrossum: Inconsistent formatting with the previous line.

behdad:
> This is a sensible refactoring. Curious how the algorithm will improve.

See the test.  Current algorithm generates 2**n-1 regions whereas only 2*n-1 is needed.

behdad:
Okay!! @justvanrossum this is ready for review.  Someone needs to inspect failing test for correctness of new output and update expectation.

behdad:
From commit logs of the two main commits:
```
[varLib.featureVars] Rewrite algorithm

Whereas previous algorithm had exponential running time and return
value size, new one has quadratic.

For featureVars_test.py test case, for example, which is a pathological
test case of n sliding intervals, the number of output intervals of
various algorithms are:

- Previous algorithm: 2**n - 1
- New algorithm: n*(n-1)/2
- Optimal algorithm: 2*n - 1

Ie, we go from exponential to quadratic, whereas in this case the optimal
solution is linear.

Running time of said test, for n=20, goes from over 20s, to 0.06s.

The algorithm can be improved.  The overlayBox() function currently does
not try to shrink the remainder box.  Doing that will probably bring us
to optimal solution for this test case.

Fixes #1372

One test is failing.  Needs to be investigated that new output is correct,
and test expectations updated.
```
and later:
```
[varLib.featureVars] Improve new algorithm to its full potential

See 0283b1f

By finishing overlayBox() to cull remainder box when possible, we generate far fewer
spurious boxes.

For the featureVars_test.py test case, this now generates optimal output, that is, 2*n - 1
boxes.

Running time for n=100 is up from 0.06s to 0.16s, but output size goes down from
5050 boxes to the optimal 199.

Fixes #1372 again
```

anthrotype:
Bravo! 👏

behdad:
Meh.  Found bugs.  Reworking.  The optimal solution might not happen.

behdad:
Bugs resolved.  Same results.

justvanrossum:
That one test failure seems harmless, in that two FeatureVariationRecords appear to have swapped places. As far as I can see these are non overlapping regions, so that change in order should not make a difference in behavior.

Would be nice to have explicit tests for complex configurations of regions.

Issue #1371, opened 2018-11-09
addFeatureVariation doesn't work on a glyph that's also being swapped by another opentype feature


weiweihuanghuang:
Please see the attached file, the ".tf" glyphs have an extra shape inserted in front so I can test and see if they're being swapped properly. I built these with fontmake then ran the attached script.

This is the progression:
![nov-09-2018 18-54-26](https://user-images.githubusercontent.com/1271897/48250029-06a8a680-e451-11e8-88fc-8de7527076f2.gif)

This is how it should look at the heaviest end where the swap is supposed to have kicked in
<img width="602" alt="screen shot 2018-11-09 at 6 55 15 pm" src="https://user-images.githubusercontent.com/1271897/48250033-090b0080-e451-11e8-8226-96a306215470.png">

[Archive.zip](https://github.com/fonttools/fonttools/files/2565029/Archive.zip)

justvanrossum:
Hm, I can't find anything wrong with your font, but I can reproduce the (non)behavior with harfbuzz/hb-shape and axis-praxis on Chrome.

@behdad, can you see why the font doesn't behave as intended?

Here's an improved build script (see #1370):

```python
### Based on a script by Stephen Nixon

import os
import sys
import fontTools
from fontTools.ttLib import TTFont
from fontTools.varLib.featureVars import addFeatureVariations

inputTTF = sys.argv[1]
##style = sys.argv[2]

f = TTFont(inputTTF)
condSubst = [
    # A list of (Region, Substitution) tuples.
    ([{"wght": (0.158, 0.564)}], {"uniF8FF": "uniF8FF.001"}),
    ([{"wght": (0.868, 1)}], {"uni20B5": "uni20B5.201_230",
                              "cent": "cent.201_230",
                              "colonmonetary": "colonmonetary.201_230",
                              "dollar": "dollar.201_230",
                              "uni20B2": "uni20B2.201_230",
                              "dollar.tf": "dollar.tf.201_230",
                              "cent.tf": "cent.tf.201_230"}),
    ([{"wght": (0.71, 1)}], {"uni20A6": "uni20A6.169_230",
                             "peseta": "peseta.169_230",
                             "uni20A9": "uni20A9.169_230"}),
    ([{"wght": (0.564, 1)}], {"uni20B1": "uni20B1.136_230",
                              "Adieresis.tf": "Adieresis.tf.136_230",
                              "Udieresis.tf": "Udieresis.tf.136_230"}),
[...]

PR #1374, opened 2018-11-10
[varLib] FeatureTableSubstitution.Version can be 0x00010001

justvanrossum:
I ran into this assert while running `fontTools.varLib.mutator` on the font posted in #1371.

Is this fix correct, @behdad?


justvanrossum:
Wait, I think this is surely not correct, looking at featureVars.py now.

justvanrossum:
Fixed by #1375.

PR #1377, opened 2018-11-13
Varlib interpolate cff2

readroberts:
Add support in varLib.mutator.py for interpolating CFF2 VF fonts.

I note that a number of tables are omitted from the interpolation, like HVAR/hmtx. Is this just not yet implemented, or is this by design?


behdad:
> HVAR/hmtx

HVAR is expected to be dropped, since it's not needed in a non-variable font.  hmtx is needed, and we interpolate it.  What are you seeing exactly?

Tests/varLib/varLib_test.py
```diff
@@ -1,8 +1,9 @@
 from __future__ import print_function, division, absolute_import
-from fontTools.misc.py23 import *
```
behdad: Please don't remove this line . All files should start with the above two lines.

Lib/fontTools/varLib/mutator.py
```diff
@@ -30,6 +30,69 @@
...
+		charstring = charstrings[gname]
+		charstring.decompile()
+		vsindex = charstring.private.vsindex if (
+													hasattr(charstring.private,
+													'vsindex')) else 0
+		num_regions = charstring.private.getNumRegions(vsindex)
+		numMasters = num_regions + 1
+		new_program = []
+		last_i = 0
+		for i, token in enumerate(charstring.program):
```
behdad: So, this assumes no subroutines?

Tests/varLib/varLib_test.py
readroberts: Right

Lib/fontTools/varLib/mutator.py
readroberts: Whoops. I spent last week with T2CharstringPen.draw(), where the flattening is all taken care of, and forgot I was workng with non-flattened charstrings. I certainly need to allow for subroutines.

readroberts:
I do expect HVAR to be dropped, I should have said that I didn't were where it was applied. For CFF2, the glyph widths come out of the hmtx table, and I don't see the widths I am expecting in the interpolated font.

Lib/fontTools/varLib/varStore.py
```diff
@@ -120,6 +120,15 @@ def __getitem__(self, varidx):
 			delta += d * s
 		return delta
 
+	def applyScalars(self, varDataIndex, deltas):
```
behdad: Please rename to `interpolateFromDeltas`, to be consistent with `VariationModel`.

Lib/fontTools/varLib/varStore.py
```diff
@@ -120,6 +120,15 @@ def __getitem__(self, varidx):
...
+		varData = self._varData
+		scalars = [self._getScalar(ri) for ri in
+					varData[varDataIndex].VarRegionIndex]
+		delta = 0.
+		for d,s in zip(deltas, scalars):
+			delta += d * s
+		return delta
```
behdad: Please share this last block with the function before it in a static method `interpolateFromDeltasAndScalars`. Again, like `VariationModel` does.

behdad:
> I do expect HVAR to be dropped, I should have said that I didn't were where it was applied. For CFF2, the glyph widths come out of the hmtx table, and I don't see the widths I am expecting in the interpolated font.

Right.  `__init__._SetCoordinates()` does that.  So, variation width comes from glyf table in this case.  We should implement HVAR/VVAR application.

behdad:
> Right. `__init__._SetCoordinates()` does that. So, variation width comes from glyf table in this case. We should implement HVAR/VVAR application.

Please file an issue for that.  I'll do it later this week.

behdad:
LGTM. Thanks.

Lib/fontTools/varLib/mutator.py
readroberts: Done. Let me know if you would like the CFF2 support functions moved from mutator.py to elsewhere, or anther approach to flattening the charstrings - there are lots of ways to do this.

Lib/fontTools/varLib/varStore.py
readroberts: Done.

Lib/fontTools/varLib/varStore.py
readroberts: Done

readroberts:
I also just noticed that the test should be in mutator_test.py, not varLib_test.py. Will move this when I need to make the next change.

Lib/fontTools/varLib/mutator.py
behdad: > Done.

I'm confused.  Don't we need to flatten recursively?  Isn't this exactly the `_DesubroutinizingT2Decompiler` in fontTools.subset.__init__.py?  Let's move that somewhere else and use it.  I also prefer that naming to "flattening" which can be ambiguous.

Also, won't we need drop to drop subroutines from the font?  Ie. exactly the desubroutinization operation.

> Let me know if you would like the CFF2 support functions moved from mutator.py to elsewhere, or anther approach to flattening the charstrings - there are lots of ways to do this.

Those can stay for now.

behdad: This code needs to be moved to a module under cffLib and reused:
https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/subset/__init__.py#L2204-L2600

@anthrotype ideas for naming?

anthrotype: fontToola.cffLib.decompilers

behdad: Humm.  While true that these are decompilers, I guess the API to them will be just functions, like `font['CFF '].cff.desubroutinize()` or `.drop_hints()`, since there's some work to do after the decompiler.  The internals can be exposed (with no API guarantee) for those who need more access.  Oh, one other useful function, that is used after both operations, is to remove unused subroutines.  This involves renumbering all other subroutines and call sites.  Is a very useful optimization.  The subsetter calls those functionalities `prune`.

Same about code that removes / renumbers unused lookups in GSUB/GPOS.  We should move that code somewhere reusable.  For example, varLib.mutator needs to call that after applying `FeatureVariations` such that unused lookups are dropped.

behdad: Maybe it's fine to expose these from submodules under the subset module?

anthrotype: > Maybe it's fine to expose these from submodules under the subset module?

sure

behdad: > Maybe it's fine to expose these from submodules under the subset module?

Subset?  I'm not ready to move interpolation-space-subsetting (of which mutator is a special-case of) to subset yet.  I suggest exposing them from varLib.mutator.* instead.

anthrotype: well, if they are needed both by subset and varLib, then cffLib would be the best candidate.
anyway, i'm ok either way.

behdad: Oh.  I confused myself again.  The github email showed your comment in reply to the code, which is very varLib-specific.  Now I see you were replying to my proposal.  Yeah, I think submodules in subset are best right now.  I thought about cffLib, but then some more will go in GSUB/GPOS, etc.  That's one way to order thing.  The other we have been using is to group by functionality.  So let's stick to the latter for now.

readroberts: I was not aware of _DesubroutinizingT2Decompiler in fontTools.subset. Certainly no need to re-invent the wheel. I would like to tweak _DesubroutinizingT2Decompiler some: CFF2 charstrings have neither endchar not return ops, so lines 2431-2436 are not needed for that charstring flavor. I would also like to at least have a mode where the charString is not desubroutinized every time it is called. Some CJK fonts max out at the limit of 64K subrs , most of which get called many times, so this would add up to serious overhead.

readroberts: I think the naming discussion about concluded that _DesubroutinizingT2Decompiler and everything else on L2204-2600 should stay in in fontTools.subset, and that interpolate_cff2_PrivateDict and interpolate_cff2_charstrings should stay in mutator.py, is that correct? 

Also, interpolate_cff2_charstrings does drop Subrs from the font, rather crudely by simply deleting them all, which is fine because we flatten all the charstrings.

behdad: > I would also like to at least have a mode where the charString is not desubroutinized every time it is called.

Right.  When I wrote the code originally I did it this way, to ensure correctness.  Code seems to have performed solidly in the years since, so I'm fine using the cache now.

behdad: > Also, interpolate_cff2_charstrings does drop Subrs from the font, rather crudely by simply deleting them all, which is fine because we flatten all the charstrings.

Correct.  But then again, the desubroutinize() function should do that (the code in subsetter does).

So, let's introduce `subset.cff` and `varLib.cff` modules?

readroberts: @behdad I switched to using a new desubroutinize() function that I added to subset.__init__.py - take a look.

After looking at subset.__init.py, I'd prefer to extract the CFF-related functions from subset.__init__.py into a new subset.cffLib module after we finish and merge the changes needed for this PR. I see a lot of lines which are CFF specific ( vs CFF2) that I'd like to fix at the same time, and I think it would be easier to review as a separate branch and PR.

behdad: Sure.  `subset.cff` please, no Lib suffix.

readroberts:
@behdad  Please let me know if you want any more changes.

Warning - you will need to pull/fetch/reset, as I ended up doing a force push to this branch.  I messed up merging the master/head changes into this branch, and ended up in a circular merge conflict with version on origin, that  I couldn't figure out how to fix. I finally ended up fixing this by renaming the local branch elsewhere, starting a new local branch with same name from the current master/head, cherry-picking the changes from the old branch, and then force-pushing the new local branch with the same name to origin.

behdad:
Thanks.  I'll check tomorrow.

PR #1345, opened 2018-10-23
build cff2 variable fonts

readroberts:
The changes in this PR should allow fonttools to build a CFF2 variable font from any OTF fonts that are built from UFO's from the Superpolator/designLib workflows. Unlike the current afdko 'buildmasterotfs' and 'buildcff2vf', there are no dependencies outside of fonttools.

PR #1378, opened 2018-11-14
[varLib]Add support for building CFF2 variable font

readroberts:
This pull request replaces https://github.com/fonttools/fonttools/pull/1345; I moved the changes to branch off a more recent state of master.


Lib/fontTools/varLib/__init__.py
```diff
@@ -620,6 +620,21 @@ def normalize(name, value):
...
+	from fontTools.cffLib.cff2_merge_funcs import (convertCFFtoCFF2, 
+								addCFFVarStore, 
+								merge_region_fonts)
+	glyphOrder = varFont.getGlyphOrder()
+	convertCFFtoCFF2(varFont)
+	# I sort the master font in the model order so as to
+	# not have to index through model.mapping when building
+	# a blend vector from the master font data.
+	ordered_fonts_list = [master_fonts[idx] for idx in model.reverseMapping]
+	model.mapping = model.reverseMapping = range(len(master_fonts))
```
behdad: Maybe we should add a method `VariationModel.reorderMasters()`, so we don't tweak its internal data?

Lib/fontTools/cffLib/cff2_merge_funcs.py
```diff
@@ -0,0 +1,246 @@
...
+	# At this point, the Subrs and Charstrings are all still T2Charstring class
+	# easiest to fix this by compiling, then decompiling again
+	cff.major = 2
+	file = BytesIO()
+	cff.compile(file, otFont, isCFF2=True)
+	file.seek(0)
+	cff.decompile(file, otFont, isCFF2=True)
+
+
+def pointsDiffer(pointList):
```
behdad: `varLib.models.allEqual()`?

Lib/fontTools/cffLib/cff2_merge_funcs.py
```diff
@@ -0,0 +1,246 @@
...
+		error_msg = os.linesep.join(error_msg)
+
+
+def conv_to_int(num):
+	if num % 1 == 0:
+		return int(num)
+	return num
+
+
+pd_blend_fields = ("BlueValues", "OtherBlues", "FamilyBlues",
```
behdad: This list seems to be duplicated in your mutator patch as well.  Maybe put it somewhere in cffLib?

Lib/fontTools/cffLib/cff2mergePen.py
```diff
@@ -0,0 +1,292 @@
...
+	def _p(self, pt):
+		""" Unlike T2CharstringPen, this class stores absolute values.
+		This is to allow the logic in check_and_fix_closepath() to work,
+		where the current or previous absolute point has to be compared to
+		the path start-point.
+		"""
+		self._p0 = pt
+		return list(self._p0)
+
+	def make_flat_curve(self, prev_coords, cur_coords):
```
behdad: Why do we need this again?

Lib/fontTools/cffLib/cff2mergePen.py
```diff
@@ -0,0 +1,292 @@
...
+			success = True
+		else:
+			success = False
+		return success, pt_coords
+
+	def check_and_fix_closepath(self, cmd, point_type, pt_coords):
+		""" Some workflows drop a lineto which closes a path.
+		Also, if the last segment is a curve in one master,
+		and a flat curve in another, the flat curve can get
+		converted to a closing lineto, and then dropped.
```
behdad: I don't think you should elevate a line to a flat curve, as there are infinite ways to do that and there's no way to know which one was the designer's intention.  This just masks a workflow bug, and introduces bugs in the font.  We discussed this in length before.

Lib/fontTools/cffLib/cff2_merge_funcs.py
readroberts: Sure.

Lib/fontTools/varLib/__init__.py
readroberts: Sounds good to me. Will do.

Lib/fontTools/cffLib/cff2_merge_funcs.py
readroberts: I don't see public method varLib.models.allEqual(). I do see varLib.merger._all_equal, which does do the same thing as cffLib.cff_merge_funcs.pointsDiffer, but I like the latter implementation a little better. Not enough to really care, though. Would you prefer I make varLib.merger._all_equal a public method, and use that instead?

Lib/fontTools/cffLib/cff2_merge_funcs.py
readroberts: Yes, I will move them to cffLib. There actually is a reason to have two separate lists, bit not three.

Lib/fontTools/cffLib/cff2mergePen.py
readroberts: Can you point me to the discussion about the problems with guessing the designers intentions? I agree that converting a line to curve arbitrarily, which I did do this PR, is a bad idea. However, I would have thought that we do have a guide to the designer's intent in the other master designs. We could  use the first master which has a curve as a model for the proportional length of the of the BCP arms. This is mentioned as away to avoid kinks in the 1997 Adobe Tech Note 5091 "Designing Multiple Master Tyepfaces", page 12, on 
https://www.adobe.com/devnet/font.html.

 I'll also need to take a look at where this happens. I think that makeotf and Robofont and Prepolator all convert flat curves to lines. I can fix the latter to have a non-optimizing mode, but if it is possible to adequately fix this problem, I'd argue that it is useful for robustness to handle this case.

Lib/fontTools/varLib/__init__.py
behdad: It should, in the VariationModel, also reorder self.origLocations, and possibly clear self._subModels or something.

Lib/fontTools/cffLib/cff2_merge_funcs.py
behdad: It should be there now.  Was a branch.  Just merged it.

Your pointsDiffer implementation is multiple times slower than allEqual().

Lib/fontTools/cffLib/cff2mergePen.py
readroberts: The starts of the last sentence should have been  'I can fix the first', not 'the latter', i.e. makeotf.

behdad: Previous discussion is not as heated as I thought it was: https://github.com/googlei18n/cu2qu/issues/55

The way to avoid kinks is to either have same BCP ratios OR same angles.  So, you can't just use BCP ratios...  Also, the BCP ratios are ratios of two BCPs of one point, NOT two BCPs of one curve.  So we end up having to look at neighboring curves.  This is messy.  I don't like to see such hacks.  Definitely not in the name of "robustness".



> I'll also need to take a look at where this happens. I think that makeotf and Robofont and Prepolator all convert flat curves to lines. I can fix the latter to have a non-optimizing mode, but if it is possible to adequately fix this problem, I'd argue that it is useful for robustness to handle this case.

Then we also fix Robofont and Prepolator.  Both have maintainers that frequent this space and can be talked to and fix things properly.  No hacks needed.

readroberts: Fair enough. I will back out flat curve code.

Lib/fontTools/cffLib/specializer.py
```diff
@@ -62,12 +63,58 @@ def programToCommands(program):
...
-def commandsToProgram(commands):
+
+def commandsToProgram(commands, var_model=None, round_func=None):
 	"""Takes a commands list as returned by programToCommands() and converts
-	it back to a T2CharString program list."""
+	it back to a T2CharString or CFF2Charstring program list."""
 	program = []
-	for op,args in commands:
-		program.extend(args)
+	for op, args in commands:
```
behdad: Can you please explain what this piece of code does?  Probably should be added as comments.

behdad: Right.  Agree with both points.

madig:
I tested this with Cantarell again:
```
$ fontmake -g Cantarell.glyphs -o otf --keep-overlaps --optimize-cff 0
$ cp master_otf/Cantarell-* master_ufo/
$ cd master_ufo
$ python3
>>> from fontTools.varLib import build
>>> varfont, model, _ = build("Cantarell.designspace", lambda s:s.replace(".ufo", ".otf"))
>>> varfont.save("test.otf")
$ ots-sanitize test.otf 
ERROR at src/layout.cc:1387 (ParseDeviceTable)
ERROR: Layout: bad delta format: 32768
ERROR at src/gpos.cc:165 (ParseAnchorTable)
ERROR: GPOS: Failed to parse device table in anchor table
ERROR at src/gpos.cc:211 (ParseMarkArrayTable)
ERROR: GPOS: Faled to parse anchor table for mark table 0
ERROR at src/gpos.cc:648 (ParseMarkToAttachmentSubtables)
ERROR: GPOS: Failed to parse mark array
ERROR at src/layout.cc:1176 (Parse)
ERROR: Layout: Failed to parse lookup subtable 3
ERROR at src/layout.cc:238 (ParseLookupTable)
ERROR: Layout: Failed to parse subtable 0
ERROR at src/layout.cc:1326 (ParseLookupListTable)
ERROR: Layout: Failed to parse lookup 0
ERROR: GPOS: Failed to parse lookup list table
ERROR at src/ots.cc:675 (ProcessGeneric)
ERROR: GPOS: Failed to parse table
Failed to sanitize file!
```
At least the outlines aren't corrupted now.

madig:
`curl -O https://gitlab.gnome.org/GNOME/cantarell-fonts/raw/master/src/Cantarell.glyphs`

behdad:
BTW, don't we need a `otf-interpolatable` mode for fontmake?

behdad:
>  ERROR at src/layout.cc:1387 (ParseDeviceTable) ERROR: Layout: bad delta format: 32768

Oh wait.  This is perfectly expected.  ots doesn't support variable fonts.  That's all.  That's what Delta format 32768 is about.

cc @khaledhosny

behdad:
This generally looks good to me.

I wish CFF2CharStringMergePen had more tests, and I hope to be able to merge some of that code back with the specializer and otherwise simplify it.

khaledhosny:
> > ERROR at src/layout.cc:1387 (ParseDeviceTable) ERROR: Layout: bad delta format: 32768
> 
> Oh wait. This is perfectly expected. ots doesn't support variable fonts. That's all. That's what Delta format 32768 is about.
> 
> cc @khaledhosny

That looks like an old version of OTS that without variable fonts support. Please test with latest release.

readroberts:
> This generally looks good to me.
> 
> I wish CFF2CharStringMergePen had more tests, and I hope to be able to merge some of that code back with the specializer and otherwise simplify it.

What additional tests would you like to see? FYI, I hope to be adding a test font with sparse masters in a week or so.

madig:
> FYI, I hope to be adding a test font with sparse masters in a week or so.

Maybe https://github.com/googlei18n/fontmake/issues/477#issuecomment-441201373 can help bootstrap such a test font :)

Lib/fontTools/varLib/models.py
```diff
@@ -264,7 +264,17 @@ def upperBound(value, lst):
 			return min(v for v in lst if v > value)
 		else:
 			return value
-
+	def reorderMasters(self, master_list):
+		# Re-order the master item list from the original master font
+		# list order to the sorted location order. This puts the 
+		# default master first, and makes building up the blend data
+		# simpler in some workflows, such as for CFF2 charstrrings.
```
behdad: This comment doesn't belong here.  Also, I was meaning for `reorderMasters` to take a mapping and reorder to that.  You can then pass `model.reverseMapping` to it.

readroberts: @behdad. Will fix.

Issue #1381, opened 2018-11-15
[varLib] Shifting composite glyphs in macOS


jenskutilek:
https://github.com/googlei18n/fontmake/issues/481

I had noticed this macOS rendering behaviour a long time ago, but the fix occurred to me only today. varLib could work around it by adding dummy entries to glyphs without necessary deltas to make sure they get an entry in the gvar table.

You can see it very clearly in [Normschrift](https://www.axis-praxis.org/specimens/normschrift) e.g. in ö (composites at default positions, no gvar entry), but not in Ö (dieresis is y-shifted so there is a gvar entry).

behdad:
Right.  I saw the same when I read your comment in other issue.  It's slightly trickier than I like it to be, as the empty entries are optimized out during the compile process, so varLib needs a way to communicate it to the compiler that we want to retain empty entries for those glyphs..

behdad:
Humm, or maybe not.

Lorp:
This bug in macOS 12.4 seems to apply to fonts that have the workaround:

[Variable fonts rendering bug: composite glyphs disappear when all deltas are zero](https://feedbackassistant.apple.com/feedback/11009662)

anthrotype:
I get "Feedback not found". @Lorp can you copy/paste the content of the bug report here?

 > apply to fonts that have the workaround:

so IIUC, removing the workaround fixes the bug on macOS 12.4?

PR #1383, opened 2018-11-15
[varLib] Always write tuple entries for composite glyphs

behdad:
Fixes https://github.com/fonttools/fonttools/issues/1381
Fixes https://github.com/googlei18n/fontmake/issues/481

@jenskutilek can you test this?


Lib/fontTools/varLib/__init__.py
```diff
@@ -313,7 +317,7 @@ def _add_gvar(font, model, master_ttfs, tolerance=0.5, optimize=True):
 		endPts = control[1] if control[0] >= 1 else list(range(len(control[1])))
 
 		for i,(delta,support) in enumerate(zip(deltas[1:], supports[1:])):
-			if all(abs(v) <= tolerance for v in delta.array):
+			if all(abs(v) <= tolerance for v in delta.array) and not isComposite():
```
anthrotype:
```suggestion
			if all(abs(v) <= tolerance for v in delta.array) and not isComposite:
```

behdad: Thanks.  Fixed.

behdad: Oh, these github suggestions are interesting.  No more manual fixing.

jenskutilek:
Will test next week! Thanks.

anthrotype:
This doesn't seem to have any effect on the issue with WorkSans as reported by @weiweihuanghuang in googlei18n/fontmake#481

I checked out this PR branch locally and installed fonttools from there; I downloaded [WorkSans-Italic.glyphs](https://github.com/weiweihuanghuang/Work-Sans/blob/master/sources/WorkSans-Italic.glyphs) and built it with `fontmake -g WorkSans-Italic.glyphs -o variable`.

I still looks like this in Axis Praxis on Chrome on MacOS High Sierra (10.13.6):

screen shot 2018-11-19 at 11 00 50 am

On FontView (using freetype+harfbuzz) it renders correctly:
screen shot 2018-11-19 at 11 00 08 am

Note that if I build the same font using fonttools from master branch (in which the current PR's patch is not applied), I get exactly the same gvar table. For Adieresis (composite glyph), it looks like this:

```xml
    <glyphVariations glyph="Adieresis">
      <tuple>
        <coord axis="wght" value="-1.0"/>
        <delta pt="0" x="0" y="0"/>
        <delta pt="1" x="-37" y="0"/>
        <delta pt="2" x="0" y="0"/>
        <delta pt="3" x="-74" y="0"/>
        <delta pt="4" x="0" y="-54"/>
        <delta pt="5" x="0" y="0"/>
      </tuple>
      <tuple>
        <coord axis="wght" value="1.0"/>
        <delta pt="0" x="0" y="0"/>
        <delta pt="1" x="14" y="0"/>
        <delta pt="2" x="0" y="0"/>
        <delta pt="3" x="29" y="0"/>
        <delta pt="4" x="0" y="24"/>
        <delta pt="5" x="0" y="0"/>
      </tuple>
    </glyphVariations>
```

jenskutilek:
Maybe Apple fixed a bug from 10.13 to 10.14? Building with fontmake and this fonttools branch, Work Sans Italic looks good in Chrome on 10.14.1:

bildschirmfoto 2018-11-20 um 11 45 35

anthrotype:
@jenskutilek and if you build it _without_ this patch? I didn't see any differences, neither visually, nor in the TTX output

jenskutilek:
Work Sans would not have been affected by the ‘missing gvar entries problem’ in the first place, because it has deltas for Adieresis anyway. Sorry for the confusion.

Here is a glyphs file with fonts to demonstrate the bug in odieresis (ö). The `varLib-composites` branch indeed does not change the binary output of fontmake, the odieresis is omitted from gvar in both cases.

[NormschriftVarTest.zip](https://github.com/fonttools/fonttools/files/2599387/NormschriftVarTest.zip)

jenskutilek:
This is what it looks like on Axis-Praxis on 10.14.1:

bildschirmfoto 2018-11-20 um 12 09 15

Lib/fontTools/varLib/__init__.py
```diff
@@ -313,7 +317,7 @@ def _add_gvar(font, model, master_ttfs, tolerance=0.5, optimize=True):
 		endPts = control[1] if control[0] >= 1 else list(range(len(control[1])))
 
 		for i,(delta,support) in enumerate(zip(deltas[1:], supports[1:])):
-			if all(abs(v) <= tolerance for v in delta.array):
+			if all(abs(v) <= tolerance for v in delta.array) and not isComposite:
 				continue
 			var = TupleVariation(support, delta)
 			if optimize:
```
jenskutilek:
```suggestion
			if optimize and not isComposite:
```

We would have to skip the optimization as well to make sure the gvar record stays intact, but then it would be longer than necessary (0 for all points when just one point would be enough).

jenskutilek:
I think this solves it. Note that for a glyph with 2 components, the unoptimized tupleVariation is smaller.

jenskutilek:
Any further comments? Should I merge this even though it does not solve the original issue?

Lib/fontTools/varLib/__init__.py
```diff
@@ -313,13 +317,17 @@ def _add_gvar(font, model, master_ttfs, tolerance=0.5, optimize=True):
...
 				delta_opt = iup_delta_optimize(delta, origCoords, endPts, tolerance=tolerance)
 
 				if None in delta_opt:
+					# In composite glyphs, there should be one 0 entry
+					# to make sure the gvar entry is written to the font.
```
anthrotype: it's not clear from this comment why "in composite glyphs, there should be one 0 entry to make sure the gvar entry is written to the font". Have we established whether this is a requirement in the OT spec or implementation detail (or even a bug) in CoreText?

jenskutilek: The spec is not very explicit about if glyphs (composites) that are not referenced in `gvar` should have variations applied (through component glyph variations).

The closest statement I could find is ‘Note: inferring of deltas for un-referenced points applies only to simple glyphs, not to composite glyphs.’

This could be interpreted in a way that composite glyphs must have explicit `gvar` entries if variations should be applied to them.

anthrotype: btw, what does Windows do?

jenskutilek: There is no problem on Window (10, tested in Edge) if the `gvar` entries are not present.

behdad: > The closest statement I could find is [‘Note: inferring of deltas for un-referenced points applies only to simple glyphs, not to composite glyphs.’](https://docs.microsoft.com/de-de/typography/opentype/spec/gvar#inferred-deltas-for-un-referenced-point-numbers)
> 
> This could be interpreted in a way that composite glyphs must have explicit `gvar` entries if variations should be applied to them.

That cannot be deduced from that, no.   What that means simply is that if a point has no deltas in gvar for composite glyphs, then the value should be assumed to be zero.

Don't commit this unless you can produce a test and this fixes.

jenskutilek: OK ... so do we agree the different rendering is a macOS bug and we don’t want a (partial) workaround?

jenskutilek:
Thanks @Lorp, what you describe is consistent with my observations.

(I can’t read your Apple bug report, I think nobody can view other people’s bugs there.)

So, in case it was too hard to follow along the various issues mentioned in this thread, let me conclude:

The effect @weiweihuanghuang sees is caused by a rendering bug in macOS up to 10.13, which is fixed in 10.14 and can not be worked around from the font side as far as we know.

This pull request will fix a different rendering bug that is still present in macOS 10.14 which is triggered by composite glyphs without any gvar entry.

anthrotype:
Thanks Jens for investigating this.

> This pull request will fix a different rendering bug that is still present in macOS 10.14...

So do we conclude this is indeed a rendering bug in CoreText? If so then I think we should make that clear in the above inline comment

anthrotype:
@jenskutilek can you confirm that this test font you published in https://github.com/unicode-org/text-rendering-tests/pull/52 is a reproducer for the

> different rendering bug that is still present in macOS 10.14 which is triggered by composite glyphs without any gvar entry.

anthrotype:
For reference, this is how the Normschrift variable font renders on Axis Praxis with Chrome on MacOS 10.14.1. The gvar table contains no entry for glyph odieresis:

out

This is the fixed-up version with a gvar entry for odieresis containing an explicit 0 delta:

out2

jenskutilek:
> @jenskutilek can you confirm that this test font you published in [unicode-org/text-rendering-tests#52](https://github.com/unicode-org/text-rendering-tests/pull/52) is a reproducer for the
> 
> > different rendering bug that is still present in macOS 10.14 which is triggered by composite glyphs without any gvar entry.

Yes, that is correct. Thanks @anthrotype for putting together the illustrated sample.

jenskutilek:
I tried to add a test for this, but get an error:

```bash
======================================================================
ERROR: test_varlib_gvar_explicit_delta (__main__.BuildTest)
The variable font contains a composite glyph n.001 which does not
----------------------------------------------------------------------
  File "varLib_test.py", line 221, in test_varlib_gvar_explicit_delta
    expected_ttx_name=test_name
  File "varLib_test.py", line 110, in _run_varlib_build_test
    varfont, model, _ = build(ds_path, finder)
  File "/Users/jens/Library/Python/3.7/lib/python/site-packages/fontTools/varLib/__init__.py", line 730, in build
    master_fonts = [TTFont(ttf_path) for ttf_path in master_ttfs]
  File "/Users/jens/Library/Python/3.7/lib/python/site-packages/fontTools/varLib/__init__.py", line 730, in <listcomp>
    master_fonts = [TTFont(ttf_path) for ttf_path in master_ttfs]
  File "/Users/jens/Library/Python/3.7/lib/python/site-packages/fontTools/ttLib/ttFont.py", line 123, in __init__
    file = open(file, "rb")
  File "/Users/jens/Library/Python/3.7/lib/python/site-packages/fontTools/misc/py23.py", line 250, in open
    file, mode, buffering, encoding, errors, newline, closefd)
FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/dp/c9fjc9m107v4pqjyprys509h0000gn/T/tmpnc_t6k8x/TestFamily4-Regular.ttf'
```

Tried digging around but I can't figure out where the file name comes from or what the expected name should be. Could you have a look @anthrotype?

Tests/varLib/varLib_test.py
```diff
@@ -204,6 +204,23 @@ def test_varlib_build_feature_variations(self):
             save_before_dump=True,
         )
 
+    def test_varlib_gvar_explicit_delta(self):
+        """The variable font contains a composite glyph n.001 which does not
```
anthrotype: isn't the glyph in question named "odieresis" in the test ufo?

Tests/varLib/varLib_test.py
```diff
@@ -204,6 +204,23 @@ def test_varlib_build_feature_variations(self):
...
+        need a gvar entry, because all its deltas are 0, but it must be added
+        anyway to work around an issue with macOS 10.14.
+
+        This is again to to work around an issue with some rasterizers:
```
anthrotype: typo: `to to`

Tests/varLib/varLib_test.py
jenskutilek: Ah, that was from an older test ufo. Will fix.

jenskutilek:
I can build the var font fine from the data directory:

```bash
fontmake -m BuildGvarComposite.designspace -o variable
```

but the test doesn’t build it ... ?

anthrotype:
The varLib_test doesn't compile the UFO into master ttfs, it takes TTX files for the latter and build temporary TTF files from them. You only added the TTX for the resulting variable font. To make it work like the other test cases, you also need to add TTX files for the master TTFs (you can create them with fontmake, then dump to TTX)

anthrotype:
fonttools.varLib (and its tests) doesn't know how to build variable fonts from UFO+designspace, only from TTFs+designspace. You need to create the master TTFs with some other tool (e.g. fontmake) and check them in as TTX files.

anthrotype:
the fact that we also commit the UFO sources to the fonttools repository is just in case we need to be able to rebuild those test fonts.

jenskutilek:
OK, thanks, I think I got everything working now. Also added the comment about macOS 10.14.

anthrotype:
thanks Jens, just merge when the test pass and you're done here

anthrotype:
@jenskutilek mind adding an entry at the top of NEWS.rst as well? so it'll be read for next release.
No need to manually write the header with the version string for the next version (that's added automatically by the release script).

Lib/fontTools/varLib/__init__.py
```diff
@@ -315,13 +319,23 @@ def _add_gvar(font, masterModel, master_ttfs, tolerance=0.5, optimize=True):
...
+					"""In composite glyphs, there should be one 0 entry
+					to make sure the gvar entry is written to the font.
+
+					This is to work around an issue with macOS 10.14 and can be
+					removed once the behaviour of macOS is changed.
+
+					https://github.com/fonttools/fonttools/issues/1381
+					"""
+					if all(d is None for d in delta_opt):
+						delta_opt = [(0, 0)] + [None] * (len(delta_opt) - 1)
```
anthrotype: @behdad @jenskutilek 
I was rereading this today, and I think that it doesn't make sense that here we try to run `iup_delta_optimize` on a composite glyph's deltas. The gvar spec is clear in saying that "inferring of deltas for un-referenced points applies only to simple glyphs, not to composite glyphs"
varLib should just write the composite glyph deltas as they are, without any inferred delta. Well ok, if they are all (0, 0) we can write the first (0, 0) and the rest as None, but that's about it.
iup_delta_optimize should only be invoked on simple glyph deltas.

anthrotype: @behdad in an earlier comment wrote, referring to the gvar spec sentence I quoted above

> What that means simply is that if a point has no deltas in gvar for composite glyphs, then the value should be assumed to be zero.

I actually couldn't find any reference in the gvar spec saying that when a point has no delta for composite glyphs it's assumed to be (0, 0). I wish this could be clarified in the spec.

If that's confirmed, then we need two different optimization methods for gvar deltas: one `iup_delta_optimize` that is to be used for simple contour glyphs, and another for composite glyphs that replaces (0,0) deltas by `None`; and if all None, leave at least the first delta as (0, 0) to keep the gvar entry for the composite glyph and make macOS happy.

anthrotype:
Ok it looks like this was intentional. I was just a little confused

behdad:
IUP only works within a contour.  Since composite components don't share contour, doesn't make sense to use IUP on them.  As such, we pass them to the optimizer and declare each point being its own contour.  This way we get the trivial 0,0->None out of the same optimizer code.

anthrotype:
Thanks for the clarification.

anthrotype:
@behdad Interestingly, I noticed that compiling a TupleVariation containing a series of explicit (0, 0) deltas ends up being slightly more compact (2 bytes less) than writing the same tuple with deltas like `[(0, 0), None, None, ...]`.
Only after writing more than 128 such (0, 0) explicit deltas, the size of the un-optimized tuple exceeds the size of an optimized one containing inferred None deltas for all except the first delta.
But the odds that a composite glyph has so many components are pretty low.
So basically the un-optimized (but smaller) tuple gets chosen 99.99 % of the time.

Lib/fontTools/varLib/__init__.py
behdad: > I was rereading this today, and I think that it doesn't make sense that here we try to run `iup_delta_optimize` on a composite glyph's deltas. The gvar spec is clear in saying that "inferring of deltas for un-referenced points applies only to simple glyphs, not to composite glyphs"

I didn't notice these comments before...

We already handle what you specify in glyf.getCoordinatesAndControl. Basically we assign position of each component its own contour, by way of this line:
```
                if glyph.isComposite():
                        coords = GlyphCoordinates(
                                [(getattr(c, 'x', 0), getattr(c, 'y', 0)) for c in glyph.components]
                        )
                        controls = _GlyphControls(
                                numberOfContours=glyph.numberOfContours,
                                endPts=list(range(len(glyph.components))),
                                flags=None,
                                components=[c.glyphName for c in glyph.components],
                        )
```

A single-point contour never gets any interpolation done for it. So that part is taken care of. We still can call iup optimization on it because a contour that is not moved at all, can avoid encoding the 0,0 movement, at the expense of having to encode point numbers explicitly. So it still can possibly benefit from IUP.

The bug seems to have simply been a wrong optimization in MacOS. That's all.

PR #1384, opened 2018-11-16
Add IDEF for GETVARIATION opcode

nyshadhr9:
(1) Checks if bytecode of any glyph uses GETVARIATION opcode
(2) If true, adds an IDEF at the end of the fpgm table and modifies 2 attributes in maxp table.

### DIFF
--- a/Lib/fontTools/varLib/mutator.py
+++ b/Lib/fontTools/varLib/mutator.py
@@ -5,8 +5,9 @@ $ fonttools varLib.mutator ./NotoSansArabic-VF.ttf wght=140 wdth=85
 """
 from __future__ import print_function, division, absolute_import
 from fontTools.misc.py23 import *
-from fontTools.misc.fixedTools import floatToFixedToFloat, otRound
-from fontTools.ttLib import TTFont
+from fontTools.misc.fixedTools import floatToFixedToFloat, otRound, floatToFixed
+from fontTools.ttLib import TTFont, newTable
+from fontTools.ttLib.tables import ttProgram
 from fontTools.ttLib.tables._g_l_y_f import GlyphCoordinates
 from fontTools.varLib import _GetCoordinates, _SetCoordinates
 from fontTools.varLib.models import (
@@ -63,11 +64,11 @@ def instantiateVariableFont(varfont, location, inplace=False):
 	glyf = varfont['glyf']
 	# get list of glyph names in gvar sorted by component depth
 	glyphnames = sorted(
-		gvar.variations.keys(),
-		key=lambda name: (
-			glyf[name].getCompositeMaxpValues(glyf).maxComponentDepth
-			if glyf[name].isComposite() else 0,
-			name))
+			gvar.variations.keys(),
+			key=lambda name: (
+				glyf[name].getCompositeMaxpValues(glyf).maxComponentDepth
+				if glyf[name].isComposite() else 0,
+				name))
 	for glyphname in glyphnames:
 		variations = gvar.variations[glyphname]
 		coordinates,_ = _GetCoordinates(varfont, glyphname)
@@ -112,7 +113,7 @@ def instantiateVariableFont(varfont, location, inplace=False):
 			if not delta:
 				continue
 			setattr(varfont[tableTag], itemName,
-				getattr(varfont[tableTag], itemName) + delta)
+							getattr(varfont[tableTag], itemName) + delta)
 
 	if 'GDEF' in varfont:
 		log.info("Mutating GDEF/GPOS/GSUB tables")
@@ -120,6 +121,35 @@ def instantiateVariableFont(varfont, location, inplace=False):
 
 		log.info("Building interpolated tables")
 		merger.instantiate()
+
+	for glyph in glyf.glyphs.values():
+		if hasattr(glyph, "program"):
+			instructions = glyph.program.getAssembly()
+			# If GETVARIATION opcode is used in bytecode of any glyph add IDEF
+			addidef = False
+			for instruction in instructions:
+				if instruction.startswith("GETVARIATION"):
+					addidef = True
+					break
+			if addidef:
+				asm = []
+				if varfont.has_key('fpgm'):
+					fpgm = varfont['fpgm']
+					asm = fpgm.program.getAssembly()
+				else:
+					fpgm = newTable('fpgm')
+					fpgm.program = ttProgram.Program()
+					varfont['fpgm'] = fpgm
+				log.info("Adding IDEF to fpgm table for GETVARIATION opcode")
+				asm.append("PUSHB[000] 145")
+				asm.append("IDEF[ ]")
+				args = [str(len(loc))]
+				for val in loc.values():
+					args.append(str(floatToFixed(val, 14)))
+				asm.append("NPUSHW[ ] " + ' '.join(args))
+				asm.append("ENDF[ ]")
+				fpgm.program.fromAssembly(asm)
+				break
 
 	if 'name' in varfont:
 		log.info("Pruning name table")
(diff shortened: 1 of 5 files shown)


Lib/fontTools/varLib/mutator.py
```diff
@@ -63,11 +64,11 @@ def instantiateVariableFont(varfont, location, inplace=False):
 	glyf = varfont['glyf']
 	# get list of glyph names in gvar sorted by component depth
 	glyphnames = sorted(
-		gvar.variations.keys(),
-		key=lambda name: (
-			glyf[name].getCompositeMaxpValues(glyf).maxComponentDepth
-			if glyf[name].isComposite() else 0,
-			name))
+			gvar.variations.keys(),
```
behdad: Are these whitespace changes intentional, or your editor settings do that?

Lib/fontTools/varLib/mutator.py
```diff
@@ -121,6 +122,35 @@ def instantiateVariableFont(varfont, location, inplace=False):
 		log.info("Building interpolated tables")
 		merger.instantiate()
 
+	for glyph in glyf.glyphs.values():
+		if hasattr(glyph, "program"):
+			instructions = glyph.program.getAssembly()
+			# If GETVARIATION opcode is used in bytecode of any glyph add IDEF
+			addidef = False
```
behdad: This for loop can be written more succinctly as:
`addidef = any(op.startswith("GETVARIATION") for op in instructions)`.

BTW, any reason to use startswith?

Lib/fontTools/varLib/mutator.py
```diff
@@ -121,6 +122,35 @@ def instantiateVariableFont(varfont, location, inplace=False):
...
+					fpgm = newTable('fpgm')
+					fpgm.program = ttProgram.Program()
+					varfont['fpgm'] = fpgm
+				log.info("Adding IDEF to fpgm table for GETVARIATION opcode")
+				asm.append("PUSHB[000] 145")
+				asm.append("IDEF[ ]")
+				args = [str(len(loc))]
+				for val in loc.values():
+					args.append(str(floatToFixed(val, 14)))
+				asm.append("NPUSHW[ ] " + ' '.join(args))
```
behdad: Isn't it easier to produce the program skipping the string parsing stage?

Lib/fontTools/varLib/mutator.py
```diff
@@ -149,6 +149,12 @@ def instantiateVariableFont(varfont, location, inplace=False):
 				asm.append("NPUSHW[ ] " + ' '.join(args))
 				asm.append("ENDF[ ]")
 				fpgm.program.fromAssembly(asm)
+
+				if varfont.has_key('maxp'):
+					if hasattr(varfont['maxp'], "maxInstructionDefs"):
+						varfont['maxp'].maxInstructionDefs += 1
+					if hasattr(varfont['maxp'], "maxStackElements"):
+						varfont['maxp'].maxStackElements += len(loc)
```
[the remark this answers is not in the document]
nyshadhr9: Thanks for catching that. Fixed it.

Lib/fontTools/varLib/mutator.py
nyshadhr9: Done.
Do you think I should check if op contains GETVARIATION? I thought a GETVARIATION opcode will be on its own line and easy to check for with startsWith.

Lib/fontTools/varLib/mutator.py
nyshadhr9: I think its because of my editor settings. Let me check how to revert them.

Lib/fontTools/varLib/mutator.py
behdad: I don't remember what the program data structure looks like. I thought the entire string is equal to the op?

Lib/fontTools/varLib/mutator.py
```diff
@@ -121,6 +122,46 @@ def instantiateVariableFont(varfont, location, inplace=False):
...
 		merger.instantiate()
 
+	addidef = False
+	for glyph in glyf.glyphs.values():
+		if addidef:
+			break
+		if hasattr(glyph, "program"):
+			instructions = glyph.program.getAssembly()
+			# If GETVARIATION opcode is used in bytecode of any glyph add IDEF
+			addidef = any(op.startswith("GETVARIATION") for op in instructions)
```
anthrotype: You could break the loop here instead of at the next iteration.

Lib/fontTools/varLib/mutator.py
```diff
@@ -121,6 +122,46 @@ def instantiateVariableFont(varfont, location, inplace=False):
...
+
+	if addidef:
+		log.info("Adding IDEF to fpgm table for GETVARIATION opcode")
+		asm = []
+		if varfont.has_key('fpgm'):
```
anthrotype: has_ key is a bit old-fashioned, you should use if ‘fpgm’ in varfont

Same for maxp below

Lib/fontTools/varLib/mutator.py
behdad: Maybe a bit ugly, but you can remove the addifdef completely and use a "for else" block.

Lib/fontTools/varLib/mutator.py
nyshadhr9: Yes the entire string including any comments, which is why I'm checking for startsWith.

Lib/fontTools/varLib/mutator.py
nyshadhr9: Done.

Lib/fontTools/varLib/mutator.py
nyshadhr9: Hmm I don't think "for else" is applicable here, since we want to implement logic if breaking out of loop?  Also, I intended to have addidef block separately to make it obvious in the code that the idef is not always added.

Lib/fontTools/varLib/mutator.py
nyshadhr9: Sorry I'm not sure if I completely understood your comment. Can you explain :)

Lib/fontTools/varLib/mutator.py
behdad: Right.

Lib/fontTools/varLib/mutator.py
behdad: Ignore me.  I thought the program data types is more than just strings.

Lib/fontTools/varLib/mutator.py
nyshadhr9: Reverted them.

Lib/fontTools/varLib/mutator.py
```diff
@@ -121,6 +122,46 @@ def instantiateVariableFont(varfont, location, inplace=False):
...
+		asm.append("ENDF[ ]")
+		fpgm.program.fromAssembly(asm)
+
+		# Change maxp attributes as IDEF is added
+		if 'maxp' in varfont:
+			maxp = varfont['maxp']
+			if hasattr(maxp, "maxInstructionDefs"):
+				maxp.maxInstructionDefs += 1
+			else:
+				setattr(maxp, "maxInstructionDefs", 1)
```
behdad: Don't know if we need the else blocks.  If maxp table is version 0.5 then we just don't touch it?

LGTM otherwise.

behdad: Right. But quotation is what to do if it isn't :).  Current code is fine.

anthrotype: I think we should unconditionally assume that maxInstructionDefs attribute is present in a TrueType maxp table, and let it raise AttributeError in the unlikely event that the attribute is missing, either because it’s a maxp version 0.5, or because the table hasn’t been fully constructed/decompiled.

behdad: Ignoring what spec says, it's perfectly fine using a version .5 for TrueType fonts.  The rest of fields in 1.0 are unused by most moderns systems.  I wouldn't want something like mutator police what people use.

anthrotype: alright.. in that case, setting the attribute seems unnecessary because it won't be compiled in the end.

anthrotype:
thanks @nyshadhr9!

PR #1385, opened 2018-11-19
Set up Windows CI on Travis

anthrotype:
let's see how it works
https://blog.travis-ci.com/2018-10-11-windows-early-release

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,16 @@ matrix:
         - PYENV_VERSION_STRING='Python 2.7.6'
         - PYENV_ROOT=$HOME/.travis-pyenv
         - TRAVIS_PYENV_VERSION='0.4.0'
+    - language: sh
+      os: windows
+      env:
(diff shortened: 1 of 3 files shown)


anthrotype:
also, the Travis windows builds seem to be taking much longer than Appveyor's.
maybe it's a bit too early for this.

behdad:
Oooh. Can someone do for HarfBuzz please? @ebraminio @khaledhosny

anthrotype:
Azure Pipelines also supports all the three desktop platforms in the same CI and integrates well with GitHub. It's worth a try
also see https://medium.com/@anthonypjshaw/azure-pipelines-with-python-by-example-aa65f4070634

anthrotype:
Github Actions also supports CI/CD for all three major OSes (and I believe it is also powered by Azure) and integrates naturally with Github's UI. I'll see if I can switch to that instead.
https://help.github.com/en/github/automating-your-workflow-with-github-actions/about-github-actions

PR #1387, opened 2018-11-23
[varLib] Implement passing the layer name to MasterFinder

madig:
This is preparatory work for https://github.com/googlei18n/ufo2ft/pull/295.

The idea is that `ufo2ft` should be able to produce sparse intermediate master OT binaries from ("brace") layers that varLib can then merge into a variable font, thanks to Behdad's sparse master work. varLib needs a way to associate a sparse "source" with an OT binary, so this PR adds passing the layer name, if existant, to `MasterFinder`.

To test-drive this, get [smastertest.zip](https://github.com/fonttools/fonttools/files/2610835/smastertest.zip), extract it somewhere and do in that folder:
```
$ python3 -m venv venv && . venv/bin/activate
$ pip install fonttools[ufo]  # to pull in ufoLib dependencies
$ pip install git+https://github.com/madig/fonttools.git@varLib-MasterFinder-add-layerName git+https://github.com/madig/ufo2ft.git@sparse-master-by-layerName
$ python build.py
```

Side-thought: in the spirit of avoiding assigning `None` to anything, maybe the `layer_name` argument should default to `"public.default"` instead?

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -23,6 +23,7 @@ from __future__ import unicode_literals
 from fontTools.misc.py23 import *
 from fontTools.misc.fixedTools import otRound
 from fontTools.misc.arrayTools import Vector
+from fontTools.misc.filenames import userNameToFileName
 from fontTools.ttLib import TTFont, newTable
 from fontTools.ttLib.tables._n_a_m_e import NameRecord
 from fontTools.ttLib.tables._f_v_a_r import Axis, NamedInstance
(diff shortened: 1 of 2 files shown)


madig:
> > Side-thought: in the spirit of avoiding assigning `None` to anything, maybe the `layer_name` argument should default to `"public.default"` instead?
> 
> Makes sense to me.

On a second though, this is putting UFO semantics into varLib. Not sure this is a good idea 🤔

behdad:
> On a second thought, this is putting UFO semantics into varLib. Not sure this is a good idea thinking

That also makes sense to me :D.

> Edit: @anthrotype just contacted me and said that he'd like to ponder the mapping of source to master OT binary some more (maybe map things explicitly in the designspace document?). So I suppose this PR is on a hold until we come to a conclusion.

Yeah, sounds a bit hacky here.

That said, here's another suggestion'ish.  Use CSS-like `filename#subtag` to refer to layers / TTC indices.

madig:
I.e. use a `#` instead of `@`? I think Cosimo is unhappy that this approach is adding finder magic when maybe the .designspace should contain an explicit mapping. Although `f"{pathlib.Path(source.filename).stem}#{userNameToFileName(source.layerName)}"` is.. ok I guess? Not sure what a better option would look like...

madig:
I noticed that with this approach, `varLib.load_designspace()` chooses the wrong base master when the axis defaults point to a sparse master/layer. In `varLib.build()`, the base font `vf` is then made the sparse master with its limited glyph set, making OT feature compilation fail with cryptic error messages.

anthrotype:
so I guess we shall do this differently.
In this PR on ufo2ft repo https://github.com/googlei18n/ufo2ft/pull/300 we added a new `compileInterpolatableTTFsFromDS` function that takes a `designspaceLib.DesignSpaceDocument` object, where each SourceDescriptor has a `font` attribute with an altready loaded defcon.Font (or compatible class). This function compiles the TTFonts and store them in the respective source.font attributes and returns a copy of the DesignSpaceDocument (or the same if inplace=True).

Given this, we could modify varLib.build to take a DesignSpaceDocument object (in addition of a path to a designspace file) and first look at the source.font attributes, if these are already loaded TTFont objects, we can use them for as master fonts, otherwise do run finder on the filenames.

PR #1395, opened 2018-12-04
[designspaceLib] Documentation: Clarify what min/max/default on an AxisDescriptor mean

madig:
The mapping stuff continues to confuse me, so I wrote down what fontmake makes of it.


madig:
This is related to https://github.com/LettError/designSpaceDocument/issues/16 in a sense. This PR provides a clarification until a decision is made.

@behdad?

behdad:
I'm not a fan of the wording.  The "Otherwise" somehow implies that if *this* value is not present or something...

Just say in user space?  Then in the help for mapping talk about user/design space and how if mapping is not present the two are the same.

behdad:
lgtm

madig:
Squash?

anthrotype:
you can sqash then merge

PR #1396, opened 2018-12-04
[varLib] addFeatureVariationsRaw: handle non-default LangSysRecord

madig:
This change handles feature variation appending to non-DefaultLangSys `Langsys`es.


madig:
Ugh! I was on 3.32.0, it's fixed in 3.33.0. Sorry for the noise.

PR #1399, opened 2018-12-05
[fontBuilder] some support for CFF2

justvanrossum:


Lib/fontTools/fontBuilder.py
```diff
@@ -478,19 +479,99 @@ def setupCFF(self, psName, fontInfo, charStringsDict, privateDict):
...
+        from .cffLib import VarStoreData
+
+        assert "fvar" in self.font, "fvar must to be set up first"
+        assert "CFF2" in self.font, "CFF2 must to be set up first"
+        axisTags = [a.axisTag for a in self.font["fvar"].axes]
+        varRegionList = ot.VarRegionList()
+        varRegionList.RegionAxisCount = len(regions)
+        varRegionList.Region = []
+        for regionDict in regions:
+            region = ot.VarRegion()
```
behdad: Use buildVarRegion()?

justvanrossum: I went one further with buildVarRegionList(). Thanks! Somehow I didn't see that as a possibility earlier.

PR #1402, opened 2018-12-05
cff2 vf fixes. Make calc_bounds work, fix subsetting bug.

readroberts:
Fixed psCharstrings so that calc_bounds will run. I would guess no-one has tried to use a BoundsPen on a CFF2 VF before - thanks to Chris Chapman. It now returns a result only for the default instance.

Fixed bug in subsetting: removed assert that a Subr is not empty after subsetting or de-hinting. CFF2 Charstrings do not have terminal "return' op.

### DIFF
--- a/Lib/fontTools/misc/psCharStrings.py
+++ b/Lib/fontTools/misc/psCharStrings.py
@@ -407,7 +407,7 @@ class SimpleT2Decompiler(object):
 		numBlends = self.pop()
 		numOps = numBlends * (self.numRegions + 1)
 		blendArgs = self.operandStack[-numOps:]
-		del self.operandStack[:-(numOps-numBlends)] # Leave the default operands on the stack.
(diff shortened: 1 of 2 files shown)


readroberts:
A note on the calc_bounds fix. There were several problems. First, the op_blend handler for T2OutlineExtractor needed to leave the default font args on the stack, rather than popping off all blend args.  So, the T2Decompiler op_blend handler does that - just don't override it. Then the  T2Decompiler op_blend handler had a bug - it was deleting from the top of the stack, rather than the bottom. Finally, in order for  T2Decompiler op_blend handler to work in the context of a T2OutlineExtractor, we needed to pass the 'private' argument down the chain of super'd __init__() functions.

readroberts:
@miguelsousa @anthrotype  Can you review this soon?

miguelsousa:
@readroberts if this is related to calc_bounds @cjchapman is a better reviewer. You need to push the request button tho.

Issue #1403, opened 2018-12-06
[CFF] remove_hints() wrong for CFF2


behdad:
Read recently changed CFF2 Private object with this:
```
↦       ↦       ↦       # Provide dummy values. This avoids needing to provide 
↦       ↦       ↦       # an isCFF2 state in a lot of places. 
↦       ↦       ↦       self.nominalWidthX = self.defaultWidthX = 0 
```
I suggest using 0 instead of None there.  That can also be used to detect CFF later.  Currently, subset.cff's drop_hints() is wrong in that it puts a width back into the charstring:
```
↦       ↦       if hasattr(self, 'width'): 
↦       ↦       ↦       # Insert width back if needed 
↦       ↦       ↦       if self.width != self.private.defaultWidthX: 
↦       ↦       ↦       ↦       self.program.insert(0, self.width - self.private.nominalWidthX) 
```
which should not happen for CFF2.  Using None above can be used to detect that.

readroberts:
Fixed by PR 1405

PR #1405, opened 2018-12-10
Clean cff2 3

readroberts:
varLib subset CFF2] Set PrivateDict nominal and default WidthX to None
Address issues raised in #1403

I do think setting the dummy CFF2 PrivateDict nominalWidthX and defaultWidthX to None, which leads to the charstring.width also being None, is a good idea. I originally set them to 0, which produces a charstring width of 0, in order to avoid problems with logic that assumes that the field is good for math. However, I now think that it is better to find errors around charstring type assumptions earlier than later.

"drop_hints()" is actually not wrong - I did look at this when making the changes. For CFF2 charstrings, self.width is always equal to self.private.defaultWidthX, so the width is never inserted. This is because in psCharstrings.py::T2WidthExtractor.popallWidth(), the test "evenOdd ^ (len(args) % 2)" is alway False. Left to myself, I would not change this code. If the CFF2 charstring is correct, there is not a problem. if the CFF2 charstring is not correct, then both in drop_hints() and in T2WidthExtractor.popallWidth(), the logic will now hit an assert that I added.

### DIFF
--- a/Lib/fontTools/misc/psCharStrings.py
+++ b/Lib/fontTools/misc/psCharStrings.py
@@ -461,7 +461,7 @@ class T2WidthExtractor(SimpleT2Decompiler):
 		if not self.gotWidth:
 			if evenOdd ^ (len(args) % 2):
 				# For CFF2 charstrings, this should never happen
-				assert self.defaultWidthX != None, "CFF2 CharStrings must not have an initial width value"
+				assert self.defaultWidthX is not None, "CFF2 CharStrings must not have an initial width value"
 				self.width = self.nominalWidthX + args[0]
 				args = args[1:]
 			else:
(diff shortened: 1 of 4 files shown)


Lib/fontTools/misc/psCharStrings.py
```diff
@@ -460,6 +460,8 @@ def popallWidth(self, evenOdd=0):
 		args = self.popall()
 		if not self.gotWidth:
 			if evenOdd ^ (len(args) % 2):
+				# For CFF2 charstrings, this should never happen
+				assert self.defaultWidthX != None, "CFF2 CharStrings must not have an initial width value"
```
behdad: Use `is not None` instead of `!= None`.

Lib/fontTools/subset/cff.py
```diff
@@ -168,6 +168,8 @@ def drop_hints(self):
 		if hasattr(self, 'width'):
 			# Insert width back if needed
 			if self.width != self.private.defaultWidthX:
+				# For CFF2 charstrings, this should never happen
+				assert self.private.defaultWidthX != None, "CFF2 CharStrings must not have an initial width value"
```
behdad: Same.

PR #1406, opened 2018-12-13
[plistlib] Treat Mapping objects as dicts

madig:
This is needed to make ufoLib2's fontinfo guidelines and gasp records serialize correctly.


anthrotype:
looks good. Could you add a simple test as well?

PR #1413, opened 2018-12-17
Pin STAT to version 1.1

moyogo:
The Windows 10 Font preview can handle fonts with STAT version 1.2 but other applications do not and only display the default named instance.
For example the Windows\Fonts folder, Word or WordPad only show the default named instance when the STAT is version 1.2 but they display all the named instances when the STAT is version 1.1.

This was tested in Windows 10 Pro, Version 1803, build 17134.407.


anthrotype:
We probably aren’t even using the fields from 1.2, are we?

moyogo:
STAT 1.2 only differs from STAT 1.1 with the addition of format 4 axis value table.

anthrotype:
Right, and we currently don’t set any axis value tables. Ok then

Issue #1411, opened 2018-12-17
Minor Kerning Interpolation Issue.


VivaRado:
https://github.com/fonttools/fonttools/blob/6e1e0b73f1d82a0a3d1e6307164dddcadcf20df6/Lib/fontTools/varLib/merger.py#L677

Reversing cols when first values are None, solves minor kerning interpolation issue on compile.
Tested only for Group to Group based kerning pairs.

```
...
	...
		for cols in zip(*list(r.Class2Record for r in rows)):
			#
			if cols[0] == None:
				cols = cols[::-1]
			#
	...
...
```

anthrotype:
you propose a "solution" but I still don't get whant the issue is

VivaRado:
Simple problem im trying to solve here. I have set extreme values between kerning the regular and italic. Values like -200 and 200. And I am trying to generate that.

By providing a small pair set (250 Group to Group), modifying the italics and regular to have opposite exaggerated kerning values (reg -200 and ital 200), and testing  it until varlib generates the interpolation between those font instances, we experience inconsistently generated VadData items. 

If you compare the OTFs, from two separate generation occasions of the same sources, that where **successful** at interpolating the extreme values, you get very small differences in parts of the header where i think its just the date.

Whereas if you compare the generated OTFs of an occasion where one interpolates and the other doesn't, the files have significant differences in DIFF.

If we open the files as TTX we see some data missing in the VarData tag and later on the ClassRecord. If we replace that data in the TTX with the data from the successfully generated kerning interpolating font and convert back to a TTF, we got kerning interpolation working on the file that it didn't work.

I hope that is clear, I could send you some files so you can check if you want. I've been all over this piece of code varLib, fontTools, fontproject following the trace. and it leads me to this issue, which my opinion is that is a surface issue. During I have experienced other issues with class assignments and mixed kerning pair sets getting miss assigned, or even classes created without existing in the provided classes which might be a feaLib issue, which I'm also looking into.

To be clear this is not an easy to identify bug and one might think that all the generated values look ok but in reality they are not. that is if you put some exaggerated values for good measure and see if they actually get outputted as they should. On small kerning pair sets, the kerning interpolation succeeds more than if the kerning.fea is larger with more pairs, involving Group to Group, Group to Letter, letter to Group, and Letter to Letter. 

Thank you!

anthrotype:
i'm sorry, I'm even more confused than before. Maybe somebody else can help you.

[...]

Issue #1410, opened 2018-12-17
much space


VivaRado:
https://github.com/fonttools/fonttools/blob/6e1e0b73f1d82a0a3d1e6307164dddcadcf20df6/Lib/fontTools/varLib/merger.py#L861

anthrotype:
sorry but this is just noise, if you care about that you can send a PR. It takes you the same amont of time as opening the issue.

VivaRado:
Done: https://github.com/VivaRado/fonttools

anthrotype:
done what?

VivaRado:
Forked and PR

PR #1414, opened 2018-12-17
Kerning Interpolation Inconsistency When First Values of Cols are None

VivaRado:
Not a real solution, more hotfix. As this problem probably originates elsewhere.
Will keep testing further. Thank you.

Follow up PR on issues:
- https://github.com/fonttools/fonttools/issues/1411#issuecomment-447809403
- https://github.com/fonttools/fonttools/issues/1410

Might be relevant to:
- https://github.com/googlei18n/fontmake/issues/470

### DIFF
--- a/Lib/fontTools/varLib/merger.py
+++ b/Lib/fontTools/varLib/merger.py
@@ -698,6 +698,10 @@ def _Lookup_PairPosFormat2_subtables_flatten(lst, font):
 		row.Class2Record = []
 		row = row.Class2Record
 		for cols in zip(*list(r.Class2Record for r in rows)):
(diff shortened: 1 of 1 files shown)


behdad:
> Not a real solution, more hotfix.

We don't do that around here.

I'll try to read your bug report and make sense of it...

VivaRado:
> I'll try to read your bug report and make sense of it...

Awesome! thank you @behdad

VivaRado:
Also @behdad if you ever get to testing this.
My files are equal - meaning the PLIST and FEA have the same structure across weights and just the kerning integers change.

PR #1415, opened 2018-12-18
[varLib.cff] Fix issues in building cff2 variable font

readroberts:
in varLib.cff2: shouldn't iterate through a dict while deleting keys
in cffLib.specializer.py: when doing peephole optimization, both the first and second args may be lists. Test for both cases.

It does bother me to have this test for list type args in specializeCommands(), since it is needed only for the one case where CFF2CharStringMergePen is being used to build up a list of source master coordinates in place of a single  coordinate.  I would rather add a function argument "doPeephole=True" for specializeCommands(), to allow suppressing this test. It will be  rare that it would be possible to satisfy the conditions for doing this optimization in a variable font, and suppressing the test makes more sense to me than complicating the common case for the sake of the special case.

### DIFF
--- a/Lib/fontTools/cffLib/specializer.py
+++ b/Lib/fontTools/cffLib/specializer.py
@@ -421,10 +421,11 @@ def specializeCommands(commands,
 					(op == commands[i-1][0]) and
 					(not isinstance(args[0], list))):
 				_, other_args = commands[i-1]
-				assert len(args) == 1 and len(other_args) == 1
-				commands[i-1] = (op, [other_args[0]+args[0]])
-				del commands[i]
-				continue
+				if not isinstance(other_args[0], list):
+					assert len(args) == 1 and len(other_args) == 1
+					commands[i-1] = (op, [other_args[0]+args[0]])
+					del commands[i]
+					continue
 
 	# 4. Peephole optimization to revert back some of the h/v variants back into their
 	#    original "relative" operator (rline/rrcurveto) if that saves a byte.
--- a/Lib/fontTools/varLib/cff.py
+++ b/Lib/fontTools/varLib/cff.py
@@ -77,7 +77,7 @@ def lib_convertCFFToCFF2(cff, otFont):
 		privateOpOrder = buildOrder(privateDictOperators2)
 		for fontDict in fdArray:
 			fontDict.setCFF2(True)
-			for key in fontDict.rawDict.keys():
+			for key in list(fontDict.rawDict.keys()):
 				if key not in fontDict.order:
 					del fontDict.rawDict[key]
 					if hasattr(fontDict, key):


anthrotype:
>  I would rather add a function argument "doPeephole=True" for specializeCommands(), to allow suppressing this test.

sounds good to me, though maybe the proposed argument name is not immediately clear

behdad:
Can we step back?

First, the trivial fix to not modifying dict, either separate it, or just push it to master.

For specializer change, can you tell me again, how the CFF2 varfont code is using the specializer?  You are calling specializer on the merged charstring, before producing blend operators?  So, instead of each operand, there's a list of operand values from multiple masters?  And you want the specializer to work for that?   Does the rest of it work?

readroberts:
Yes, the rest of specializer.py::specializeCommands() works fine on the merged charstring. This is because most of the optimizations are about removing a 0 value. This should be done only if the relative coord is is 0 in all the masters. If it is, then it is represented by a single 0 value rather than a list, and the logic works. This is a harder to do after the coordinate lists are reduced to a blend operator and its args, because you have to figure out what index to test for the 0 value lineto or curveto operator. 

Discussing makes me realize that this is actually a larger issue, as right now, programToCommands() is broken for CFF2Charstrings. For example, the program byte sequence "x0 <y blendargs> blend rlineto" turns into the command list items   ["blend", [ x0] + <y blendargs>], ["rlineto", []]. Note that the blend operator ends up with all the args, and the 'rlineto" operator ends up with an empty stack. If you call commandsToProgram() on this command list, the correct CFF2Charstring program will come out, but any operation that dependson referencing an expected number of args for an operator in command list will fail.  I suggest that the simplest way to deal with this issue to have programToCommands convert a blend operator by removing it, and pushing the blend args into the marking operator as a list rather than a value. This require new logic in a number of places, but so will dealing with the current case.

behdad:
BTW, I think tuples are more appropriate than lists to represent a set of values / deltas.  If we can at least detect / allow both in code, that's better.

Lib/fontTools/cffLib/specializer.py
```diff
@@ -421,10 +421,11 @@ def specializeCommands(commands,
 					(op == commands[i-1][0]) and
 					(not isinstance(args[0], list))):
 				_, other_args = commands[i-1]
-				assert len(args) == 1 and len(other_args) == 1
-				commands[i-1] = (op, [other_args[0]+args[0]])
-				del commands[i]
-				continue
+				if not isinstance(other_args[0], list):
```
behdad: Combine the if with the conditions before it...

Actually, if both are list, should we add them up?  I think so.

behdad: And if one is a list and other is a scalar, should the scalar be added to the first item of the list, or all of them?  Depends on whether the list is all master values or master+deltas.  I'm fine skipping that part.

behdad:
> I would rather add a function argument "doPeephole=True" for specializeCommands(), to allow suppressing this test.

It's not the peephole optimization BTW.  It's merging consecutive operations.  I'm fine working with it here, if we integrate blend operator handling as well.

Lib/fontTools/cffLib/specializer.py
readroberts: Just to make sure that I read your comments correctly:
I think you agree that blend operators should not be represented directly in the charstring.command list. Instead, each blend operator and its args will be replaced by a series of tuples, one for each of the blended args. leaving the expected number of args on the stack for the following marking operator. This makes sense to me, as the current intent of a command list item is that it be self-contained: an operator name, and a list of all its arguments, which is not the case if the blend operator and its args are represented directly. The design cost of the proposed change is that when the program is converted to a command list, the number of regions must be available. For CFF2 Charstrings, this true in the workflows I have seen with the current code base. However, this is true of any logic that needs to make sense of the  blend operator args.

The commandsToProgram does not need a maxstack argument, although it does need the value in order to avoid overrunning the stack, because maxstack for a CFF2 Charstring is now a constant that can be imported from the cffLib. 

And as you suggest, since blend args are all source font values, merging consecutive operations by adding their args is actually pretty simple for any mix of blend args and regular coordinate value args.

behdad: > And as you suggest, since blend args are all source font values, merging consecutive operations by adding their args is actually pretty simple for any mix of blend args and regular coordinate value args.

Aren't blend args normally deltas?  Those need different treatment adding with a scalar; just the first one should be adjusted.  Right?

behdad: The rest sounds good.

anthrotype:
@readroberts I just came across this debugger call in `varLib.cff`
https://github.com/fonttools/fonttools/blob/f88a1f6465d6b045481eceba88f5f7aaa85fd1fe/Lib/fontTools/varLib/cff.py#L474-L476

it looks like it was forgotten there..

readroberts:
@anthrotype You are correct.  I have fixed this in a the  branch 'cff2-saprse-sources', which I expect to have done in the next workday or so.

readroberts:
I am closing this PR, as after Behdad's comments above, I am taking a different approach in specializer.py than in the branch 'clean-cff2vf-4'.

anthrotype:
Thanks Read!

PR #1416, opened 2018-12-19
[varLib] Accept DesignspaceDocument objects in build() and load_designspace()

madig:
We need a way to integrate sparse masters that come from ("brace") layers instead of stand-alone sparse UFOs. Instead of modifying `varLib.build()` and `MasterFinder` to go look for specially named files like in https://github.com/fonttools/fonttools/pull/1387, modify `varLib.build()` to take either 

1. a path and continue as before (masters from layers not supported then, unless you write the compiled `TTFont`s to disk and refer to them with the `filename` attribute and probably adjust `master_finder`)
2. or a `DesignspaceDocument` and demand that the caller already read and attached the `TTFont` objects to all `SourceDescriptor.font` attributes. ufo2ft will do this when calling https://github.com/googlei18n/ufo2ft/pull/300/files#diff-cd32d8abceb8eff27b2b4994f7322bc1R256.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -755,26 +755,23 @@ def build(designspace, master_finder=lambda s:s, exclude=[], optimize=True):
 	else:  # Assume a path
 		basedir = os.path.dirname(designspace)
 
-	for master in ds.masters:
-		# 1. If a SourceDescriptor has a layer name, demand that the compiled TTFont 
-		# be supplied by the caller. This spares us from modifying MasterFinder.
-		if master.layerName:
-			if not master.font:
+	for master_index, master in enumerate(ds.masters):
+		# 1. If the caller already supplies a TTFont for a source, just take it.
+		if master.font:
+			font = master.font
+			master_fonts.append(font)
+			master_ttfs.append(None)  # No file path for in-memory object.
+		else:
+			# If a SourceDescriptor has a layer name, demand that the compiled TTFont
+			# be supplied by the caller. This spares us from modifying MasterFinder.
+			if master.layerName:
 				raise AttributeError(
 					"Designspace source '%s' specified a layer name but lacks the "
 					"then required TTFont object in the 'font' attribute."
 					% getattr(master, "name", "<Unknown>")
 				)
-			master_fonts.append(master.font)
-			master_ttfs.append(None)  # No file path for in-memory object.
-		else:
-			# 2. If the caller already supplies a TTFont for a source, just take it.
-			if master.font:
-				font = master.font
-				master_fonts.append(font)
-				master_ttfs.append(None)  # No file path for in-memory object.
 			else:
-				# 3. A SourceDescriptor's filename might point to a UFO or an OpenType
+				# 2. A SourceDescriptor's filename might point to a UFO or an OpenType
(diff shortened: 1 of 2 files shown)


behdad:
Sgtm.

madig:
After a discussion with Cosimo, we want to support three use cases:

1. The CLI should continue to work as before, as in `python3 -m varLib my.designspace --master-finder "somewhere/{stem}.otf"`. Usually, the pointed to Designspace file contains UFO sources, so we need to go find the master binaries.
2. In-memory merging without intermediate writing to disk must be possible, so `build()` must be able to take a `DesignspaceDocument` where the `SourceDescriptor.font` attributes have been preloaded by e.g. fontmake/ufo2ft.
3. It should be possible to `fontmake -m my.designspace -o ttf-interpolatable --output-path somewhere && python3 -m varLib somewhere/my.designspace`, i.e. to write a Designspace file where the `SourceDescriptor.filename`s point to OT binaries. In that case, there must be no `layerName` as that doesn't make sense for OT binaries and the master finder should not be invoked.

Lib/fontTools/varLib/__init__.py
```diff
@@ -740,15 +744,32 @@ def build(designspace_filename, master_finder=lambda s:s, exclude=[], optimize=T
...
-	basedir = os.path.dirname(designspace_filename)
-	master_ttfs = [master_finder(os.path.join(basedir, m.filename)) for m in ds.masters]
-	master_fonts = [TTFont(ttf_path) for ttf_path in master_ttfs]
-	# Reload base font as target font
-	vf = TTFont(master_ttfs[ds.base_idx])
+
+	if hasattr(designspace, "sources"):  # Assume a DesignspaceDocument
+		for master in ds.masters:
+			if master.font is None:
+				raise AttributeError(
```
anthrotype: instead of asserting master.font is not None for the case when designspace is a DesignSpaceDocument object, and loading TTFont from master.filename using master_finder when designspace is a file path...

how about you:
- check if master.font is not None (in which case varLib.build _must_ have been passed a DesignSpaceDocument instance), and if so then use it as is; otherwise
- try to load it as a TTFont from master.filename; if it doesn't raise TTLibError, it must be a valid binary master TTF
- else if it can't be loaded as TTFont, we use the master_finder to locate the binary file from the master.filename

readroberts:
This is a useful idea, if I understand it correctly, even without brace layers.  I am currently dealing with a workflow where all the source fonts will have a full glyph complement, and I will have to subset some of them to temporary files to make sparse source fonts. With this PR, I think i can read in the original source fonts and attach the TTFont objects to the DesignSpace SourceDescriptor.font attributes, and subset them in memory.

anthrotype:
@readroberts correct

behdad:
> It's a bit unfortunate that `build()` takes a file path instead of expecting a fully loaded `DesignspaceDocument` and letting the caller care about the nitty-gritty...

Change it as you wish.

madig:
Changing it might be something for 4.0. I'd like to think more on the API. Maybe we can move MasterFinder to the entry points instead of having them in `__init__`...

anthrotype:
@madig merged manually to https://github.com/fonttools/fonttools/tree/varlib-build-accept-ds-object branch, let's continue there

PR #1421, opened 2018-12-30
[designspaceLib] [t1Lib] fix cases where pathlib.Path objects were not accepted

justvanrossum:


justvanrossum:
Also: interesting failures on Python 3.5 :(

madig:
`os.fspath` is 3.6+ only 😢

anthrotype:
@justvanrossum have you tried this?
https://github.com/fonttools/fonttools/blob/8e080f2a82f407eae8cd5040c9c3d4fa24f17de0/Lib/fontTools/ufoLib/__init__.py#L206

Tests/designspaceLib/designspace_test.py
```diff
@@ -785,3 +791,18 @@ def test_documentLib(tmpdir):
...
+    import pathlib
+    source = (pathlib.Path(__file__) / "../data/test.designspace").resolve()
+    assert source.exists()
+    doc = DesignSpaceDocument()
+    doc.read(source)
+
+@pytest.mark.skipif(pathlib is None, reason="pathlib not installed")
+def test_with_with_path_object(tmpdir):
+    import pathlib
+    dest = pathlib.Path(tmpdir) / "test.designspace"
```
anthrotype: Do `pathlib.Path(str(tmpdir))` here
The tmpdir fixture returns a py.path.local object (yet another path library)

madig: Or use `tmp_path` instead of `tmpdir` in pytest 3.9+.

anthrotype: Does pytest now requires pathlib2 backport on py27?

Lib/fontTools/t1Lib/__init__.py
```diff
@@ -108,11 +108,12 @@ def parse(self):
 
 def read(path, onlyHeader=False):
 	"""reads any Type 1 font file, returns raw data"""
-	normpath = path.lower()
+	_, ext = os.path.splitext(path)
```
anthrotype: It looks like splitext doesn’t work with PathLike objects on python 3.5
You may have to convert to str/bytes first, then splitext

I’d love to drop 3.5 support and require 3.6+ (but let’s discuss this in the appropriate issue)

Tests/designspaceLib/designspace_test.py
madig: Yes if you use tmp_path.

anthrotype: Whatever works. I’m ok with requiring the latest pytest to run tests (which is what the CI runs anyway).

Lib/fontTools/t1Lib/__init__.py
madig: I guess the fspath thing has to be done everywhere where a str is expected. Or: refactor everything to work on paths, as a long-term solution.

justvanrossum: I'm tempted to skip the tests below Python 3.6, somehow I can't get it to work in 3.5.

anthrotype: that makes sense

PR #1425, opened 2019-01-02
[varLib] make `build` also accept DesignSpaceDocument objects

anthrotype:
with source descriptors containing pre-loaded TTFont instances

Continuation of https://github.com/fonttools/fonttools/pull/1416

### DIFF
--- a/Lib/fontTools/designspaceLib/__init__.py
+++ b/Lib/fontTools/designspaceLib/__init__.py
@@ -1282,13 +1282,15 @@ class DesignSpaceDocument(LogMixin, AsDictMixin):
         'path' attribute, or if the latter is not set, the concatenation of
         the source's (relative) 'filename' attribute and the designspace's
         directory.
+        The returned path uses the platform-specific path separator: i.e.
+        forward slash on POSIX, backslash on Windows.
 
         Raises DesignSpaceDocumentError if sourceDescriptor.path is None and
         self.path is None, or sourceDescriptor.filename is None.
         """
         if sourceDescriptor.path:
             # prefer absolute path if present
-            return sourceDescriptor.path
+            return os.path.normpath(sourceDescriptor.path)
         if self.path is None:
             raise DesignSpaceDocumentError(
                 "DesignSpaceDocument 'path' attribute is not defined; cannot "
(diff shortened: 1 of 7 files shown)


madig:
Can we use Python 3.x features yet? The whole path handling stuff makes my head hurt. I'd like to have all paths be `pathlib.Path` objects, no POSIX gymnastics, that are written to disk with `as_posix()`. Since filenames (not `.path`) must be relative according to the spec, we don't have to deal with drive roots on Windows.

I'm not super familiar with all the use-cases for designspace objects, but what about making `filename` a write-only property and always have `path` point to the absolute location of sources and instances? On writing the DS, the filename property is calulated as `filename = SourceOrInstanceDescriptor.path.relative_to(Path(designspace_file))`. Then there is also no need for a `getSourcePath()` method.

Let me think about what that would mean for client code. 

1. A piece of code with an in-memory designspace object with in-memory UFOs or TTFs attached and no path for them set yet will write all font objects to disk somewhere, write the paths to the DS object and write it last.
2. Loading a designspace file would resolve the filename attributes on loading and instantiate absolute `Path` objects.
3. A piece of code moving font objects somewhere else on disk would overwrite the path attributes and write the DS object again.

Maybe I'm forgetting something, but path modification would mean simply overwriting the `path` attribute with the known new absolute path and writing the DS object to disk would take care of the rest.

madig:
Aside from this, the changes look mostly fine to me. One thing that has me scratching my head is the `reload_font` thing in `varLib_test.py`. Are we testing the `__deepcopy__` changes properly? The TTX dumps we load from the test data are already "properly" serialized data structures with no temporary data, no? Shouldn't we test freshly compiled fonts that still have some working data in them that gets removed on `save()`?

anthrotype:
what you suggest make sense, and i've also always been confused by this duality of path/filename attributes in designspace document. However, I don't want to break existing code for the sake of it. For now, let's keep both path and filename attributes as posix path strings, mutable, and nullable (with all the problems that this implies), and deal with them later in separate PRs.

About the `reload_font` thing in the test, that's only done because we are comparing TTX dumps for string equality, and there are a few redundant things which are only computed at compile time: e.g. count values (written out inside xml comments in the ttx dump), or in this particular case, the name records in the name table are sorted upon compile. No big deal, the data doesn't change.

I changed the test to load compiled binary test fonts instead of importing them as ttx to address your concern, and it actually caught a bug in that custom deepcoy implementation for SFNTReader.

Issue #1426, opened 2019-01-02
Error in error reporting


schriftgestalt:
When this file (that is messed up on purpose), ttx responds with this error message:

https://github.com/moonAgirl/Bugs/blob/master/otfcc/2018-12-30-01-global-buffer-overflow.otf/
the file was posted here: https://github.com/caryll/otfcc/issues/59

```
  File "/fonttools/Lib/fontTools/ttx.py", line 398, in main
    process(jobs, options)
  File "/fonttools/Lib/fontTools/ttx.py", line 372, in process
    action(input, output, options)
  File "/fonttools/Lib/fontTools/misc/loggingTools.py", line 375, in wrapper
    return func(*args, **kwds)
  File "/fonttools/Lib/fontTools/ttx.py", line 269, in ttDump
    newlinestr=options.newlinestr)
  File "/fonttools/Lib/fontTools/ttLib/ttFont.py", line 228, in saveXML
    self._saveXML(writer, **kwargs)
  File "/fonttools/Lib/fontTools/ttLib/ttFont.py", line 283, in _saveXML
    self._tableToXML(tableWriter, tag, splitGlyphs=splitGlyphs)
  File "/fonttools/Lib/fontTools/ttLib/ttFont.py", line 314, in _tableToXML
    table.toXML(writer, self)
  File "/fonttools/Lib/fontTools/ttLib/tables/C_F_F_.py", line 42, in toXML
    self.cff.toXML(writer)
  File "/fonttools/Lib/fontTools/cffLib/__init__.py", line 156, in toXML
    font.toXML(xmlWriter)
  File "/fonttools/Lib/fontTools/cffLib/__init__.py", line 2322, in toXML
    self.decompileAllCharStrings()
  File "/fonttools/Lib/fontTools/cffLib/__init__.py", line 2338, in decompileAllCharStrings
    log.error("Error in charstring %s", i)
NameError: global name 'i' is not defined
```

anthrotype:
ouch. thank @schriftgestalt for reporting this and @justvanrossum for the quick fix

anthrotype:
Fixed by #1427

PR #1427, opened 2019-01-02
[cffLib] fixing #1426

justvanrossum:
However, an AssertionError is raised further on, when dumping the test font.


anthrotype:
the assertion error is fine because the mentioned test font is invalid, as it was intentionally produced to solicit undefined behavior.

PR #1428, opened 2019-01-02
Update etree.py

frankrolf:
To conform with the output from UFOnormalizer (and the rest of the XML file).
Along similar lines, I wonder what needs to happen to write a tabbed XML file (instead of two-spaced, as lxml does)?

### DIFF
--- a/Lib/fontTools/misc/etree.py
+++ b/Lib/fontTools/misc/etree.py
@@ -15,7 +15,7 @@ from __future__ import absolute_import, unicode_literals
 from fontTools.misc.py23 import basestring, unicode, tounicode, open
 
 
-XML_DECLARATION = """<?xml version='1.0' encoding='%s'?>"""
+XML_DECLARATION = """<?xml version="1.0" encoding="%s"?>"""
 
 __all__ = [
     # public symbols


anthrotype:
thanks @frankrolf, however I am inclined to keep it as it is.
We decided to intentionally modify the default XML declaration to use single quotes, even though it would constitute a breaking change from the output of the standalone ufoLib. The reason was to make the output be consistent with that of lxml and the built-in etree library, which both use single quotes in their XML declaration (for historical reason, there is no better or worse, they are both valid). 

see https://github.com/fonttools/fonttools/pull/1335#issuecomment-431118568

Same thing about using tabs vs two-space indentation. Lxml doesn't allow to customize the output from its pretty_print. For simplicity I'd like to keep it that way.

frankrolf:
Why is lxml consistency more urgent than ufoLib consistency?

anthrotype:
speed

anthrotype:
the purpose of that `fontTools.misc.etree` is to provide a compatibility shim for when lxml is not installed, as we didn't feel ready to add a hard requirement on lxml (being a native extension module). So the idea is to keep consistent the output of fontTools.ufoLib when installed with or without lxml. And since the former cannot be customized (and it's good enough, and it's super fast), the latter is made to conform to it.
I hope that makes sense

justvanrossum:
It's rather unfortunate that everyone who has UFOs under version control will now have to commit this otherwise nonsensical change in hundreds or thousands of files.

It's probably too late to go back now, but really, this can't happen again in the future. Same with the plist indentation change. For users it's a gratuitous change that causes many headaches.

anthrotype:
I know..

Anyway, lxml does have support for XSLT transformations, maybe that could be a way for the user to adopt a particular indentation style and the like. I remember @moyogo experimented with something like that

anthrotype:
More info at https://lxml.de/xpathxslt.html#xslt

PR #1429, opened 2019-01-04
Adds test for misc.psCharStrings.encodeFloat that exposes a bug

rsms:
Adds test for misc.psCharStrings.encodeFloat that exposes a bug with "E+"

For a value like `9.876537e+24` the encodeFloat raises a KeyError exception.

This patch adds a test for `misc.psCharStrings.encodeFloat`

Note: I'm submitting a separate PR with an updated version of `encodeFloat` that passes the tests.

```
$ pytest Tests/misc/psCharStrings_test.py
...
Lib/fontTools/misc/psCharStrings.py:244: KeyError
==== short test summary info ====
FAIL Tests/misc/psCharStrings_test.py::T2CharStringTest::test_encodeFloat
==== 1 failed, 4 passed in 0.15 seconds ====
```

PR #1430, opened 2019-01-04
Fixes bug in misc.psCharStrings.encodeFloat and adds test

rsms:
This patch:

- Adds a unit test for `misc.psCharStrings.encodeFloat`
- Fixes a bug in `encodeFloat` where a number like `1.23e+4` would raise a KeyError exception
- Fixes an issue where the produced data would be of too high precision for 32-bit parsers, like macOS. See https://github.com/googlei18n/ufo2ft/issues/306

This is a follow-up PR to this small proof-of-issue PR: https://github.com/fonttools/fonttools/pull/1429

A faster implementation of `encodeFloat` is available here separately: https://gist.github.com/rsms/efd5663749a9b66bc53be8f85becc9d2 — introducing this more efficient implementation caused test failures in ttLib/tables/C_F_F_test which seemed like too much changes for one PR.

### DIFF
--- a/Lib/fontTools/misc/psCharStrings.py
+++ b/Lib/fontTools/misc/psCharStrings.py
@@ -242,11 +242,9 @@ def encodeFloat(f):
 		if c == "E":
 			c2 = s[:1]
 			if c2 == "-":
-				s = s[2:]
+				s = s[1:]
 				c = "E-"
 			elif c2 == "+":
-				s = s[2:]
-			else:
 				s = s[1:]
 		nibbles.append(realNibblesDict[c])
 	nibbles.append(0xf)
--- a/Lib/fontTools/misc/psCharStrings.py
+++ b/Lib/fontTools/misc/psCharStrings.py
@@ -223,9 +223,14 @@ def encodeFixed(f, pack=struct.pack):
 	else:
 		return b"\xff" + pack(">l", value)  # encode the entire fixed point value
 
+
+realZeroBytes = bytechr(30) + bytechr(0xf)
+
 def encodeFloat(f):
 	# For CFF only, used in cffLib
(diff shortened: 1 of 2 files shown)


justvanrossum:
It seems Adobe never emits more than 8 digits:

https://github.com/adobe-type-tools/afdko/blob/d61a88009c5781088ce926846aeb659e091e2617/c/public/lib/source/cffwrite/cffwrite_dict.c#L230

rsms:
I see. I’ll change this to match.

rsms:
Looking into AFDKO, [`ctuDtostr `](https://github.com/adobe-type-tools/afdko/blob/107849ab9851cb78a1d90ee82add394e54119a9c/c/public/lib/source/ctutil/ctutil.c#L244) used to format a double into a string never produces output with exponent (i.e. "e") as it uses `%.{PRECISION}lf` with `printf`. This means that not only is some code paths of [`cfwDictSaveReal`](https://github.com/adobe-type-tools/afdko/blob/d61a88009c5781088ce926846aeb659e091e2617/c/public/lib/source/cffwrite/cffwrite_dict.c#L230) never taken, but that also some numbers might be _very_ long (since exponent is not used in the output.) For instance a double constant like `1.e+30` would produce a 40 byte long ASCII string, which with exponent notation would be able to be expressed in as little as 5 bytes. For this reason I think it's best to keep using `G` for formatting the string to keep output as compact as possible.

anthrotype:
oh.. I now realised I shouldn't have merged #1429 before this one, I misunderstood the two were interdependent while actually this one incorporates and supersedes the previous one. I'll revert #1429.

rsms:
Thanks for a fantastic piece of software ❤️

PR #1434, opened 2019-01-09
[fontBuilder[ added support for cmap format 14 Unicode Variation Sequences

justvanrossum:


Tests/fontBuilder/fontBuilder_test.py
```diff
@@ -244,3 +244,34 @@ def test_setupNameTable_no_windows():
...
+def test_unicodeVariationSequences(tmpdir):
+    familyName = "UVSTestFont"
+    styleName = "Regular"
+    nameStrings = dict(familyName=familyName, styleName=styleName)
+    nameStrings['psName'] = familyName + "-" + styleName
+    glyphOrder = [".notdef", "space", "zero", "zero.slash"]
+    cmap = {ord(" "): "space", ord("0"): "zero"}
+    uvs = [
+        (0x0030, 0xFE00, "zero.slash"),
+        (0x0030, 0xFE01, "zero"),  # not an official sequence, just testing
```
anthrotype: maybe you could also add a test case where the third item is None

anthrotype: would be nice if the test data reflected some realistic unicode variation sequence, e.g. the zero.slashed that @moyogo used elsewhere. I think it was `U+0030 U+FE00`

anthrotype: oops.. never mind (i must be drunk)

Tests/fontBuilder/data/test_uvs.ttf.ttx
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ttFont sfntVersion="\x00\x01\x00\x00" ttLibVersion="3.35">
+
+  <cmap>
+    <tableVersion version="0"/>
+    <cmap_format_4 platformID="0" platEncID="3" language="0">
+      <map code="0x20" name="space"/><!-- SPACE -->
+      <map code="0x30" name="zero"/><!-- DIGIT ZERO -->
+    </cmap_format_4>
+    <cmap_format_14 platformID="0" platEncID="5" format="14" length="49" numVarSelectorRecords="2">
```
anthrotype: unrelated but.. isn't it weird that format14 cmap subtables are dumped with that `lenght` attribute there?

justvanrossum: Yeah, I have no idea what that's doing there.

Tests/fontBuilder/data/test_uvs.ttf.ttx
```diff
@@ -0,0 +1,20 @@
...
+      <map uvs="0xfe00" uv="0x30" name="zero.slash"/>
+      <map uvs="0xfe01" uv="0x30" name="None"/>
```
anthrotype: i also don't like that special "None" glyph name.. but let's fix this and the "length" thing in a separate PR maybe

justvanrossum: Yeah, both those thing should not be part of this PR.

Tests/fontBuilder/fontBuilder_test.py
justvanrossum: It's a mix of realistic and unrealistic :)

PR #1437, opened 2019-01-10
[ttLib] cmap format 14 cleanup

### LINKED ISSUE #1435: [ttLib] clean up XML for cmap format 14
As a follow up to #1434:
* default values should no use “None” for glyph names
* the `cmap_format_14` element shouldn’t have a `lenght` attribute.

justvanrossum:
Fixes #1435

### DIFF
--- a/Tests/ttLib/tables/data/_c_m_a_p_format_14.ttx
+++ b/Tests/ttLib/tables/data/_c_m_a_p_format_14.ttx
@@ -4,8 +4,8 @@
   <cmap>
     <tableVersion version="0"/>
     <cmap_format_14 platformID="0" platEncID="5">
-      <map uvs="0xfe00" uv="0x30" name="zero.slash"/>
-      <map uvs="0xfe01" uv="0x30"/>
+      <map uv="0x30" uvs="0xfe00" name="zero.slash"/>
+      <map uv="0x30" uvs="0xfe01"/>
     </cmap_format_14>
   </cmap>
 
--- a/Tests/ttLib/tables/data/_c_m_a_p_format_14.ttx
+++ b/Tests/ttLib/tables/data/_c_m_a_p_format_14.ttx
@@ -5,7 +5,7 @@
     <tableVersion version="0"/>
     <cmap_format_14 platformID="0" platEncID="5">
       <map uvs="0xfe00" uv="0x30" name="zero.slash"/>
-      <map uvs="0xfe01" uv="0x30" name=""/>
+      <map uvs="0xfe01" uv="0x30"/>
     </cmap_format_14>
   </cmap>
 
--- a/Tests/ttLib/tables/data/_c_m_a_p_format_14.ttx
+++ b/Tests/ttLib/tables/data/_c_m_a_p_format_14.ttx
@@ -3,7 +3,7 @@
 
   <cmap>
     <tableVersion version="0"/>
-    <cmap_format_14 platformID="0" platEncID="5" format="14" length="49" numVarSelectorRecords="2">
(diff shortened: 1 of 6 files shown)


Tests/ttLib/tables/data/_c_m_a_p_format_14.ttx
```diff
@@ -3,7 +3,7 @@
 
   <cmap>
     <tableVersion version="0"/>
-    <cmap_format_14 platformID="0" platEncID="5" format="14" length="49" numVarSelectorRecords="2">
+    <cmap_format_14 platformID="0" platEncID="5">
       <map uvs="0xfe00" uv="0x30" name="zero.slash"/>
       <map uvs="0xfe01" uv="0x30" name="None"/>
```
khaledhosny: I think `name="None"` should go as well, since `None` is a legitimate glyph name, the absence of a `name` should be treated as `None` or something like that.

justvanrossum: Right, I meant to do that and now forgot, thanks!

anthrotype: maybe only upon reading a ttx file, you could check if a glyph named "None" actually exist in the font, and if not, treat that special string as it was, to not break backward compatibility with files produced with older fonttools?

justvanrossum: Hm, it would be possible if we postpone that test to compile-time. I don't think we know which glyphs are in the font when parsing the xml. Hmhm.

anthrotype: hmmmmmm

justvanrossum: Hm actually, we have a TTFont instance, and I think ttFont.getGlyphOrder() can't fail at this stage. At least not for a full TTX load, as the `<GlyphOrder>` element always comes first.

When merging a partial TTX file with an existing font there is also a valid glyph order. Right now I can't think of a case where this would fail.

Tests/ttLib/tables/data/_c_m_a_p_format_14.ttx
```diff
@@ -3,9 +3,9 @@
 
   <cmap>
     <tableVersion version="0"/>
-    <cmap_format_14 platformID="0" platEncID="5" format="14" length="49" numVarSelectorRecords="2">
+    <cmap_format_14 platformID="0" platEncID="5">
       <map uvs="0xfe00" uv="0x30" name="zero.slash"/>
-      <map uvs="0xfe01" uv="0x30" name="None"/>
+      <map uvs="0xfe01" uv="0x30" name=""/>
```
anthrotype: how about, instead of an empty string, we don't output the `name` attribute in the TTX dump when it is `None`, and similarly assume that name is `None` when the attribute is not present?

justvanrossum: Ah, that's not a bad suggestion. Let me see.

anthrotype: also maybe, shouldn't the unicode variation selector "uvs" attribute be written _after_ the base unicode character value "uv" in the sequence of xml attributes for the `map` element?
That's how one writes them, e.g. 0x30 0xFE00, not viceversa

justvanrossum: Yeah, also a fair point.

PR #1439, opened 2019-01-14
{v,h}mtx: allow to compile/decompile even without {v,h}hhea table

anthrotype:
when no horizontal/vertical header table is present, we can assume numberOfHMetrics == maxp.numGlyphs.
This allows to compile/decompile incomplete (i.e. technically invalid) OpenType binary fonts that only containing head, maxp, hmtx, loca and glyf (or CFF), that is the bare minimum to temporarily store glyph outline and metrics data for use as sparse/intermediate master TTFs to build variable fonts with varLib.
See https://github.com/googlei18n/ufo2ft/issues/308


behdad:
Bad hack ;).

PR #1441, opened 2019-01-14
[varLib] allow loading masters from TTX files as well

anthrotype:
in a designspace document, the sources' filename attribute can now contain a relative path to a .TTX file for that master, in addition to .TTF or .UFO (the latter, resolved through a master_finder callable as before).

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -811,14 +811,36 @@ def build(designspace, master_finder=lambda s:s, exclude=[], optimize=True):
 	return vf, model, master_ttfs
 
 
+def _open_font(path, master_finder):
+	# load TTFont masters from given 'path': this can be either a .TTX or an
+	# OpenType binary font; or if neither of these, try use the 'master_finder'
+	# callable to resolve the path to a valid .TTX or OpenType font binary.
+	from fontTools.ttx import guessFileType
+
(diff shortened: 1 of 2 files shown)


anthrotype:
@justvanrossum would you be ok if we have TTFont constructor also accept .TTX files?
e.g. doing `TTFont("MyFont.ttx")` would be the same as doing `font = TTFont(); font.importXML("MyFont.ttx")`

justvanrossum:
> would you be ok if we have TTFont constructor also accept .TTX files?
e.g. doing TTFont("MyFont.ttx") would be the same as doing font = TTFont(); font.importXML("MyFont.ttx")

What will that buy us? I don't really have a solid argument against it, but it does _feel_ a little odd and different, as parsing a TTX file has to be done in one go, and cannot be done lazily. To me, `f = TTFont(...)` is always a fairly quick and efficient operation, and that will definitely not be the case with a TTX file.

justvanrossum:
> sometimes is useful to load a TTFont from either binary or TTX input files,

That is clear, but I'm not 100% convinced it's the TTFont contructor's responsibility to do that.

> but if you prefer that can be done in a separate function or classmethod.

I tend to think it's app-specific behavior, and am not even sure it belongs in ttLib. But yeah, something separate from the constructor is definitely better IMO.

PR #1443, opened 2019-01-14
[subset] Add --retain-gids to retain glyph indices

behdad:


anthrotype:
we can work on that HVAR todo together next week

anthrotype:
ah, and CFF too

Issue #1424, opened 2019-01-02
rename deprecated methods in latest pyfilesystem2 API (fs >= 2.2.0)


anthrotype:
A few methods like getbytes, setbytes and similar have been renamed in fs module 2.2.0, they now issue a deprecation warning:

https://github.com/PyFilesystem/pyfilesystem2/releases/tag/2.1.2

we should rename them and bump required fs module in install_requires

willmcgugan:
Apologies, I messed up the release version. Here's the link:

https://github.com/PyFilesystem/pyfilesystem2/releases/tag/v2.2.0

plant99:
Hey, @willmcgugan .
I would like to work on this issue, if you are busy to change the function names, and up the version in 'install_requires'.

willmcgugan:
@plant99 not my project, but I imagine they would be happy to accept a PR...

anthrotype:
> I would like to work on this issue
@plant99 of course, you're welcome to send a PR! thanks

PR #1445, opened 2019-01-15
ufoLib, fs calls: Replace deprecated getbytes with readbytes

madig:


anthrotype:
You also need to bump the setup.py requirements for fs module

anthrotype:
setbytes -> writebytes deprecation warnings are still triggered.
#1424

Issue #1446, opened 2019-01-15
[subset] handle CFF and HVAR in --retain-gids feature


anthrotype:
https://github.com/fonttools/fonttools/pull/1443 added support for subsetting TrueType glyph outlines/metrics without modifying glyph indexes (making dropped glyphs unprintable, no contours, no advance).
We need to do the same for CFF (and "endchar" in a glyph's charstring program should take care of both advance width and contours, right?), and HVAR (with no mapping, glyphIDs are used as implicit delta-set indices).
What else?

behdad:
> HVAR (with no mapping, glyphIDs are used as implicit delta-set indices).

A bit more involved.  Let's work out together next week.

> What else?

Those are what I can think of.  Not sure what other tables require all glyphs to be listed in.

Probably also post table?  Or maybe not.  If user is keeping glyph names, best to keep all glyph names?   Hard to know without knowing *why* this feature is being used...

behdad:
Changed my mind.  I think we should drop glyph names for them as well if we can...

PR #1447, opened 2019-01-15
[subset] set emptied CFF charstrings to 'endchar' with --retain-gids

anthrotype:
Part of #1446

is it just setting the program to `["endchar"]` or am I missing something?

### DIFF
--- a/Lib/fontTools/subset/cff.py
+++ b/Lib/fontTools/subset/cff.py
@@ -66,8 +66,13 @@ def closure_glyphs(self, s):
 		s.glyphs.update(components)
 		decompose = components
 
-def _empty_charstring(font, glyphName, width=None):
+def _empty_charstring(font, glyphName, isCFF2, width=None):
 	c, fdSelectIndex = font.CharStrings.getItemAndSelector(glyphName)
+	if isCFF2:
+		# CFF2 charstrings have no widths nor 'endchar' operators
+		c.decompile()
+		c.program = []
+		return
 	if hasattr(font, 'FDArray') and font.FDArray is not None:
 		private = font.FDArray[fdSelectIndex].Private
 	else:
@@ -94,7 +99,7 @@ def prune_pre_subset(self, font, options):
 	if options.notdef_glyph and not options.notdef_outline:
 		for fontname in cff.keys():
 			font = cff[fontname]
-			_empty_charstring(font, ".notdef")
+			_empty_charstring(font, ".notdef", isCFF2=cff.major > 1)
 
 	# Clear useless Encoding
 	for fontname in cff.keys():
@@ -113,7 +118,7 @@ def subset_glyphs(self, s):
 
 		if s.options.retain_gids:
 			for g in s.glyphs_emptied:
-				_empty_charstring(font, g, width=0)
+				_empty_charstring(font, g, isCFF2=cff.major > 1, width=0)
 		else:
 			# Load all glyphs
 			for g in font.charset:
--- a/Lib/fontTools/subset/cff.py
+++ b/Lib/fontTools/subset/cff.py
@@ -93,9 +93,10 @@ def prune_pre_subset(self, font, options):
 	cff.fontNames = cff.fontNames[:1]
 
 	if options.notdef_glyph and not options.notdef_outline:
+		isCFF2 = cff.major > 1
 		for fontname in cff.keys():
 			font = cff[fontname]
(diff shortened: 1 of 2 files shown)


Lib/fontTools/subset/cff.py
```diff
@@ -104,37 +104,44 @@ def subset_glyphs(self, s):
...
+					c = cs[g]
+					c.decompile()
+					c.program = ["endchar"]
 		else:
-			cs.charStrings = {g:v
-					  for g,v in cs.charStrings.items()
-					  if g in s.glyphs}
-		font.charset = [g for g in font.charset if g in s.glyphs]
-		font.numGlyphs = len(font.charset)
+			# Load all glyphs
```
anthrotype: i didn't touch this whole block, simply indented

miguelsousa:
I think the modified glyphs then inherit the defaultWidth as their advance width, so there may be more work needed.

anthrotype:
oh right, so we need to do something similar to what we already do when we empty the .notdef glyph.
i'll do that after lunch, thanks

anthrotype:
On the other hand, given that these glyphs will be unreachable from cmap, hence virtually inexistent, I wonder if that matters...

miguelsousa:
Yeah Cosimo, you can't think about encoded glyphs only 😉

khaledhosny:
I *think* the glyph ids are simply kept so that the subset font have the same glyph ids as the original ones, all other properties are of not much interest.

miguelsousa:
But the glyphs are still in the font, so to make them appear non-existent (i.e "subset") most everything about them should be zeroed, I think.

readroberts:
@anthrotype Don't forget that CFF2 charstrings have no width and no 'endchar'.

anthrotype:
Ok but #1448

anthrotype:
@readroberts ok, done with https://github.com/anthrotype/fonttools/commit/9b296463bde96e6032dedd4c886580cae8ee171e -- which also fixed the issue in #1448. 
I think this PR is ready now

Lib/fontTools/subset/cff.py
```diff
@@ -104,37 +116,41 @@ def subset_glyphs(self, s):
...
-				# https://github.com/khaledhosny/ots/pull/31
-				#sel.format = None
-				sel.format = 3
-				sel.gidArray = [sel.gidArray[i] for i in indices]
-			cs.charStrings = {g:indices.index(v)
-					  for g,v in cs.charStrings.items()
-					  if g in s.glyphs}
+		if s.options.retain_gids:
+			for g in s.glyphs_emptied:
+				_empty_charstring(font, g, isCFF2=cff.major > 1, width=0)
```
behdad: I think we want to pass private.defaultWidthX here.

anthrotype: I don't understand.. i'm passing the cff font in as parameter, then for CFF 1.0 I'm getting it's private.defaultWidthX and compute/store the charstring width. I thought for CFF2 we never want to store the width.

behdad:
> I think the modified glyphs then inherit the defaultWidth as their advance width, so there may be more work needed.

That's fine to me, and in fact, we should do exactly that.

anthrotype:
> That's fine to me, and in fact, we should do exactly that.

I had a feeling you preferred that way... Want me to revert/simplify?

Lib/fontTools/subset/cff.py
behdad: You're passing width=0, which requires storage in the charstring in CFF 1.0.  We don't care about width of the emptied glyphs, so we want to not store anything, which is equivalent to default width.  For CFF2 of course, there is no width whatsoever.

behdad:
> > That's fine to me, and in fact, we should do exactly that.
> 
> I had a feeling you preferred that way... Want me to revert/simplify?

Yes please.

anthrotype:
yeah, I did that intentionally because others pointed that out, but I realise it makes more sense to not waste storage for data which is not going to be used

behdad:
> yeah, I did that intentionally because others pointed that out, but I realise it makes more sense to not waste storage for data which is not going to be used

That, and OpenType 1.6 on say width comes from hmtx, not CFF table.

anthrotype:
it would be tempting to also drop the custom width of the ".notdef" glyph, but I kept it as it was (it's just one glyph after all).

PR #1449, opened 2019-01-15
[varLib] ignore post.underline* from sparse masters when building MVAR

anthrotype:
When varLib.build receives "sparse" masters only containing a subset of the glyph outlines/metrics and minimal set of tables compared to the full base master font, we need a way to exclude the post.underlineThickness and post.underlinePosition values from being used to build the MVAR delta sets.
We need a post table (for truetype) if we want to use the glyph names only defined in there when we build gvar, HVAR, etc. from such sparse masters -- the glyph name is used as the key to group glyphs from different masters.
At the same time, a post table must have these two underline values defined to something.
Currently, varLib `_add_MVAR` function assumes that, if a given master contains a table (OS/2, hhea, post, etc), then all the MVAR-specific values in that table will be used when generating MVAR table for the variable font.
So, we cannot do without post (otherwise we lose glyph names in TrueType), but at the same time we want underline values to be ignored for these incomplete masters.

In this PR, I came up with an unlikely, arbitrary value of 0x8000 (the minimum int16 integer), and take that as a sentinel value to decide whether to exclude a master when computing MVAR's `unds` and `undo` tags.
It's a hack, but I can't see any alternative solution -- besides maybe requiring the use of .TTX files instead of binary fonts for such sparse masters, in that a TTX file may contain a pseudo "GlyphOrder" table without an actual `post` table. 

I also described the issue here: https://github.com/googlei18n/ufo2ft/issues/308#issuecomment-454079704
Related ufo2ft PR: https://github.com/googlei18n/ufo2ft/pull/310

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -512,13 +512,31 @@ def _add_MVAR(font, masterModel, master_ttfs, axisTags):
 	lastTableTag = None
 	fontTable = None
 	tables = None
+	# HACK: we need to special-case post.underlineThickness and .underlinePosition
+	# and unilaterally/arbitrarily define a sentinel value to distinguish the case
+	# when a post table is present in a given master simply because that's where
+	# the glyph names in TrueType must be stored, but the underline values are not
+	# meant to be used for building MVAR's deltas. The value of 0x8000 (-36768)
+	# the minimum FWord (int16) value, was chosen for its unlikelyhood to appear
+	# in real-world underline position/thickness values.
+	specialTags = {"unds": 0x8000, "undo": 0x8000}
 	for tag, (tableTag, itemName) in sorted(MVAR_ENTRIES.items(), key=lambda kv: kv[1]):
 		if tableTag != lastTableTag:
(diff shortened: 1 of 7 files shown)


anthrotype:
I actually meant -0x8000 (-32768) with a minus sign; the +0x8000 of course doesn't fit a signed int16.

Lib/fontTools/varLib/__init__.py
```diff
@@ -512,13 +512,31 @@ def _add_MVAR(font, masterModel, master_ttfs, axisTags):
...
 		if tableTag != lastTableTag:
 			tables = fontTable = None
 			if tableTag in font:
 				fontTable = font[tableTag]
-				tables = [master[tableTag] if tableTag in master else None
-					  for master in master_ttfs]
+				tables = []
+				for master in master_ttfs:
+					if tableTag in master:
+						if (
```
behdad: Please rewrite this block like this:
```
if (tableTag not in master or
  (tag in ... and getattr...)):
  tables.append(None)
  continue
tables.append(master[tableTag])

PR #1450, opened 2019-01-15
[voltLib] Fix check for duplicate anchors

khaledhosny:
When checking for duplicate anchors, the component number should be taken into account since the same anchors can be used for different components i.e. over ligatures.


moyogo:
👍 LGTM!

PR #1453, opened 2019-01-16
[voltLib] Don’t try to read past END

khaledhosny:
The parser was still trying to read the next token even when the current token was END, but I think it should just stop reading here. When reading from `TSIV` table there can be null bytes at the end when would cause an exception in the lexer.

### DIFF
--- a/Lib/fontTools/voltLib/parser.py
+++ b/Lib/fontTools/voltLib/parser.py
@@ -44,10 +44,7 @@ class Parser(object):
                 func = getattr(self, PARSE_FUNCS[self.cur_token_])
                 statements.append(func())
             elif self.is_cur_keyword_("END"):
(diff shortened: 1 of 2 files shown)


moyogo:
Good point.
I was worried vtp files manually or programmatically generated may have statements after END.
Should we trim null bytes instead?

khaledhosny:
I considered trimming null bytes (that is what fontTools do when writing to XML), but it felt like the more sensible approach is to ignore anything after `END` as otherwise its presence is meaningless (but we don’t have a spec, so we can just speculate about its exact meaning).

moyogo:
OK.

moyogo:
Let’s merge this, MS VOLT ignores what is after END.

Issue #1451, opened 2019-01-16
[ttx] AttributeError "isCFF2" from T2CharString when dumping CFF2 toXML()


anthrotype:
Dumping "SourceSerifVariable-Roman.otf" with ttx yields an AttributeError from T2CharString.toXML method:

```
$ ttx ~/Downloads/SourceSerifVariable-Roman.otf 
Dumping "/home/clupo/Downloads/SourceSerifVariable-Roman.otf" to "/home/clupo/Downloads/SourceSerifVariable-Roman.ttx"...
Dumping 'GlyphOrder' table...
Dumping 'head' table...
Dumping 'hhea' table...
Dumping 'maxp' table...
Dumping 'OS/2' table...
Dumping 'hmtx' table...
Dumping 'cmap' table...
Dumping 'name' table...
Dumping 'post' table...
Dumping 'BASE' table...
Dumping 'CFF2' table...
ERROR: Unhandled exception has occurred
Traceback (most recent call last):
  File "/home/clupo/Github/fonttools/Lib/fontTools/ttx.py", line 398, in main
    process(jobs, options)
  File "/home/clupo/Github/fonttools/Lib/fontTools/ttx.py", line 372, in process
    action(input, output, options)
  File "/home/clupo/Github/fonttools/Lib/fontTools/misc/loggingTools.py", line 375, in wrapper
    return func(*args, **kwds)
  File "/home/clupo/Github/fonttools/Lib/fontTools/ttx.py", line 269, in ttDump
    newlinestr=options.newlinestr)
  File "/home/clupo/Github/fonttools/Lib/fontTools/ttLib/ttFont.py", line 228, in saveXML
    self._saveXML(writer, **kwargs)
  File "/home/clupo/Github/fonttools/Lib/fontTools/ttLib/ttFont.py", line 283, in _saveXML
    self._tableToXML(tableWriter, tag, splitGlyphs=splitGlyphs)
  File "/home/clupo/Github/fonttools/Lib/fontTools/ttLib/ttFont.py", line 314, in _tableToXML
    table.toXML(writer, self)
  File "/home/clupo/Github/fonttools/Lib/fontTools/ttLib/tables/C_F_F_.py", line 42, in toXML
    self.cff.toXML(writer)
  File "/home/clupo/Github/fonttools/Lib/fontTools/cffLib/__init__.py", line 156, in toXML
    font.toXML(xmlWriter)
[...]

PR #1456, opened 2019-01-17
[cffLib] Fix for #1451

justvanrossum:
A charstring is not guaranteed to end in an operator, so the final bytecodes 11 and 14 can be part of an encoded numeric value; so remove 'return' or 'endchar' at the program level instead of bytecode.

Fixes next issue from #1451

### DIFF
--- a/Lib/fontTools/misc/psCharStrings.py
+++ b/Lib/fontTools/misc/psCharStrings.py
@@ -983,6 +983,12 @@ class T2CharString(object):
 			return
 		opcodes = self.opcodes
 		program = self.program
+
+		if isCFF2:
+			# If present, remove return and endchar operators.
+			if program and program[-1] in ("return", "endchar"):
+				program = program[:-1]
+
 		bytecode = []
 		encodeInt = self.getIntEncoder()
 		encodeFixed = self.getFixedEncoder()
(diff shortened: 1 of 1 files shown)


justvanrossum:
The "if not isCFF2" test later could actually be the else clause of this new if. That error could be raised even before compilation starts.

justvanrossum:
> The "if not isCFF2" test later could actually be the else clause of this new if. That error could be raised even before compilation starts.

Yes, let me incorporate that change in this PR.

Lib/fontTools/misc/psCharStrings.py
```diff
@@ -983,6 +983,16 @@ def compile(self, isCFF2=False):
 			return
 		opcodes = self.opcodes
 		program = self.program
+
+		if isCFF2:
+			# If present, remove return and endchar operators.
+			if program and program[-1] in ("return", "endchar"):
+				program = program[:-1]
```
anthrotype: `del program[-1]` would avoid the copy

justvanrossum: Yeah, but I thought it was nicer to not alter self.program. But I can change it if you want.

anthrotype: ah, you mean, in case the reference to `program` list was shared by some other object.. Ok, let's leave it like that.

PR #1462, opened 2019-01-21
[voltLib] Fix parsing glyph groups

### LINKED ISSUE #1442: [voltLib] GroupDefinition.glyphSet() is broken
Take the following simple test vtp file:
```
DEF_GROUP "test"
 ENUM
   GLYPH "a"
 END_ENUM
END_GROUP
```

Processing it with this simple program:
```python
import sys
from fontTools.voltLib.ast import GroupDefinition
from fontTools.voltLib.parser import Parser

ast = Parser(sys.argv[1]).parse()
for statement in ast.statements:
    if isinstance(statement, GroupDefinition):
        print(statement.name)
        print(statement.glyphSet())
```

This raises:
```
test
  File "test.py", line 10, in <module>
    print(statement.glyphSet())
  File "/lib/python3.7/site-packages/fontTools/voltLib/ast.py", line 75, in glyphSet
    self.glyphs_ = self.enum.glyphSet(groups)
AttributeError: 'tuple' object has no attribute 'glyphSet'
```

It looks like there is some confusion here; the code seems to expect the enum to be `Lexer.Enum` which is not used anywhere actually as `Parser.parse_coverage_()` returns a tuple. If this is indeed the intention, then `Parser.parse_coverage_()` should be fixed to return `Lexer.Enum` (I tried that quickly, but it seems to a chain reaction of other things that need to be fixed, so I’d like to have some confirmation before attempting that.

@moyogo

khaledhosny:
Fixes #1442.

### DIFF
--- a/Lib/fontTools/voltLib/ast.py
+++ b/Lib/fontTools/voltLib/ast.py
@@ -96,6 +96,15 @@ class Enum(Expression):
         for e in self.glyphSet():
             yield e
 
+    def __len__(self):
+        return len(self.enum)
+
+    def __eq__(self, other):
+        return self.glyphSet() == other.glyphSet()
+
+    def __hash__(self):
+        return hash(self.glyphSet())
+
     def glyphSet(self, groups=None):
         glyphs = set()
         for element in self.enum:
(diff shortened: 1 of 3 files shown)


moyogo:
LGTM, thank you!

Issue #1141, opened 2018-01-05
svgLib: support approximating arcs to cubic bezier curves


anthrotype:
Follow up from @santhoshtr attempt here:
https://github.com/fonttools/fonttools/pull/1136

Let's continue the discussion here.

I found this piece of python code that @behdad wrote, with an `Arc` class with an `approximate_bezier` metohd:
https://github.com/behdad/gl-bezier/blob/master/behdad.py#L114

He said cairo has a similar but better implementation. Need to check that out

anthrotype:
I think what we want is this `_cairo_arc_segment` function in `cairo-arc.c`
https://cgit.freedesktop.org/cairo/tree/src/cairo-arc.c#n124

anthrotype:
also @justvanrossum's snippet here:
https://github.com/fonttools/fonttools/pull/1136#issuecomment-355552910

anthrotype:
I just remembered that @pomax’s Bezier Primer also contains a chapter on approximating circles with cubic or quadratic beziers
https://pomax.github.io/bezierinfo/#circles_cubic

behdad:
Use this: https://github.com/behdad/cairo/blob/color-emoji/src/cairo-arc.c

anthrotype:
ok, I'll use that when I get to this

rsheeter:
Oh very nice. I was reading Pomax but cairo-arc.c is even better.

behdad:
> Oh very nice. I was reading Pomax but cairo-arc.c is even better.

I see you're getting pulled into the curve math blackhole :D.  There's a lot to read, and even more unwritten...

rsheeter:
Good news, people are very determined to ensure I do nothing but attend meetings so I guess I can dodge that bullet :)

PR #1464, opened 2019-01-22
svgLib: add support for converting elliptical arcs to cubic bezier curves

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/1141

I basically translated the code from Chromium Blink's SVG path parser, in SVGPathNormalizer::DecomposeArcToCubic
https://github.com/chromium/chromium/blob/master/third_party/blink/renderer/core/svg/svg_path_parser.cc#L169-L278

Each elliptical arc is approximated by series of cubic bezier curves, one cubic every 90-degree portion of an arc.

I decided to use this instead of the other suggested code from Cairo, because this one also handles the conversion from endpoint to center parametrization of elliptical arcs, whereas the cairo-arc.c code assumed a center parametrization. The Cairo code also allows greater precision than the current one, but I believe 4 cubic bezier (each 90 degrees) approximating for a full circle or ellipsis are OK for font design. As far as I remember, for drawing the geometric primitive shape of circles or ellipses, font editors always approximate them with four cubic bezier curves.

Maybe I can further work on this in a second iteration of this PR. 

Tomorrow I'll add some tests too.


davelab6:
Side note: Sometimes 8 cubics are used instead of 4, but I don't think that
changes anything here :)

anthrotype:
Sometimes.. when?

davelab6:
For ellipses that can't be reached with 4

PR #1465, opened 2019-01-22
[voltLib] Distinguish different PROCESS_MARKS

khaledhosny:
`PROCESS_MARKS` followed by a group name is used for `markAttachmentType` lookup flag, while followed by `MARK_GLYPH_SET` is used for `useMarkFilteringSet`. The code parsed both correctly but did not
distinguish between the two in the generated AST as it should since they compile to different lookup flags.

### DIFF
--- a/Tests/voltLib/parser_test.py
+++ b/Tests/voltLib/parser_test.py
@@ -622,9 +622,8 @@ class ParserTest(unittest.TestCase):
             'END_SUB\n'
             'END_SUBSTITUTION'
         ).statements
-        process_base = lookup.process_base
         self.assertEqual(
-            (lookup.name, process_base),
+            (lookup.name, lookup.process_base),
             ("SomeSub", False))
 
     def test_substitution_process_base(self):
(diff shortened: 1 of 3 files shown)


Tests/voltLib/parser_test.py
```diff
@@ -683,6 +683,24 @@ def test_substitution_process_marks(self):
...
+            'AS_SUBSTITUTION\n'
+            'SUB GLYPH "A"\n'
+            'WITH GLYPH "A.c2sc"\n'
+            'END_SUB\n'
+            'END_SUBSTITUTION'
+        ).statements
+        mark_glyph_set = lookup.mark_glyph_set
+        self.assertEqual(
+            (lookup.name, mark_glyph_set),
+            ("SomeSub", "SomeMarks"))
```
moyogo: `mark_glyph_set = lookup.mark_glyph_set` is not needed, use `lookup.mark_glyph_set` directly in the assertEqual.

moyogo:
Thanks!

Tests/voltLib/parser_test.py
khaledhosny: I just copied the test before it, but it makes sense.

PR #1469, opened 2019-01-26
[voltLib] Return a tuple from parse_coverage_()

khaledhosny:
This is a followup to commit 94633e9f46975c356ec3a2d21ed30768c2fa0cd5, where I mistakenly made it return an `Enum` but `parse_coverage_()` it not used only for `ENUM`s and in many (most?) places returning an `Enum` is wrong as you have a list of separate items that has to remain separate.


khaledhosny:
@moyogo ping.

moyogo:
Thanks Khaled. Sorry for the delay.

PR #1475, opened 2019-01-30
Support sparse sources and multiple submodels for CFF2 fonts

readroberts:
Changes to sparse sources and multiple sub-models for CFF2 fonts, and test cases.


miguelsousa:
@readroberts all checks have failed

Lib/fontTools/cffLib/specializer.py
```diff
@@ -26,26 +27,63 @@ def programToString(program):
...
+	values.
+	'deltasToSourceValues' can be from varLib.varStore.VarDataItem."""
 
 	width = None
 	commands = []
 	stack = []
+	seen_blend = False
+	seen_op = False
+	if num_regions != 0:
+		num_sources = num_regions + 1
```
cjchapman: I think you should set `num_sources` to some value, even if `num_regions` is zero, to reduce the risk of `num_sources` being undefined at line 65.

readroberts: Thanks! I agree that this will make code safer against future editing. 'num_sources' can't currently be used unless num_regions is > 0, but this could  be changed in the future.

behdad:
I'll take a look this week..

readroberts:
@anthrotype @behdad  I just updated this branch, with all of Cosimo's suggestions folded in, at least as I understood them. Since I edited the history to clean up the commits and force-pushed it, you will need to update your version.  The build did fail on Windows Python 27, but it looks like an installer issue rather than a code issue. The style suggestions should be much more widely applied, but I limited my changes to those specifically suggested, in order to not confuse the logic change history.

anthrotype:
I re-run the appveyour build and it passes now (not sure why it failed earlier

@behdad pretty ping (i know you're also busy)

@readroberts I don't feel confident enough with the CFF/CFF2 specs to review this PR. FYI I'll in Mountain View next week and the following one. Maybe we could meet in person or over VC and you could walk me through it? It's not necessary of course, you can proceed without me.

Lib/fontTools/cffLib/specializer.py
```diff
@@ -26,26 +27,62 @@ def programToString(program):
...
 		if not isinstance(token, basestring):
 			stack.append(token)
 			continue
-
-		if width is None and token in {'hstem', 'hstemhm', 'vstem', 'vstemhm',
-					       'cntrmask', 'hintmask',
-					       'hmoveto', 'vmoveto', 'rmoveto',
-					       'endchar'}:
+		if token == 'blend':
+			# if we see a blend, it is a CFF charstring, and there
```
anthrotype: did you mean perhaps this
```suggestion
			# if we see a blend, it is a CFF2 charstring, and there
```

Tests/varLib/varLib_test.py
```diff
@@ -361,87 +378,6 @@ def test_varlib_build_sparse_masters(self):
         tables = [table_tag for table_tag in varfont.keys() if table_tag != "head"]
         self.expect_ttx(varfont, expected_ttx_path, tables)
 
-    def test_varlib_build_sparse_masters_MVAR(self):
```
anthrotype: any reasons why you had to remove this test case?

Lib/fontTools/varLib/models.py
```diff
@@ -298,8 +293,13 @@ def _computeMasterSupports(self, axisPoints, axisOrder):
...
 			# Walk over previous masters now
-			for j,m in enumerate(locations[:i]):
-				# Master with extra axes do not participte
+			for j,m in enumerate(locations[:i+2]):
+				if i == j:
+					continue
+				# Need to iterate through i+2 in order to set both the
+				# lower and upper values of the (lower, peak, upeer)
+				# triplet for intermediate masters.
+				# Master with extra axes do not participate
```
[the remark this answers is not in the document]
anthrotype: As discussed with Read and Behdad, I have unbundled the change to varLib.models from the current cff2-sparse-masters PR. I will push that commit to a separate branch. Behdad had some reservation about it, while Read argued that it was necessary. I admit I haven't fully understood the issue yet.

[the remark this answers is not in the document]
anthrotype: I pushed the above mentioned commit (https://github.com/fonttools/fonttools/commit/898c7f76602b22ecda6a81fd9a6cd223f7343947) to this branch:
https://github.com/fonttools/fonttools/tree/varlib-models-intermediate-masters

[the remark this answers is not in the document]
behdad: That change is just wrong.  Please don't push in.

Lib/fontTools/varLib/varStore.py
```diff
@@ -184,6 +184,71 @@ def interpolateFromDeltas(self, varDataIndex, deltas):
 		return self.interpolateFromDeltasAndScalars(deltas, scalars)
 
 
+class VarDataItem(object):
```
behdad: I don't think this belongs here.

anthrotype: yeah, maybe should be moved to `fontTools.varLib.cff`

behdad:
I've tried reviewing this over ten times.  It just hurts the way it changes specializer, and varStore modules.  I think it's confusing and makes assumptions that are not necessarily correct.  Like, going from deltas to source values and back.  I don't support this change as is, but also don't have time to fully understand it and suggest alternative approach.

Lib/fontTools/varLib/models.py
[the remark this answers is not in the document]
anthrotype: Plotting the `Tests/varLib/data/TestSparseCFF2VF.designspace` file used in this PR's tests _without_ commit 898c7f7 currently gives this:

testsparsecff2_without_read_patch

Whereas this is the result _after_ applying that commit:

testsparsecff2_with_read_patch

Lib/fontTools/cffLib/specializer.py
```diff
@@ -26,26 +27,62 @@ def programToString(program):
...
+				default_value = stack[argi]
+				blend_list = deltasToSourceValues(default_value, deltas)
+				stack[argi] = blend_list
+				tuplei = next_ti
+				argi += 1
+			del stack[-(num_args- num_blends):]
+			# convert blend op to blend value list args
+			continue
+			
+		if not seen_op and token in {'hstem', 'hstemhm', 'vstem', 'vstemhm',
```
blueshade7: This code attempts to process the width in a CFF2 CharString, if no `blend` op has been seen so far and an extra argument is on the stack. It shouldn't be triggered by a well-formed CFF2 CharString, but seems odd and maybe better to explicitly check for CFF2 instead?

Lib/fontTools/cffLib/specializer.py
```diff
@@ -26,26 +27,62 @@ def programToString(program):
 	return ' '.join(str(x) for x in program)
 
 
-def programToCommands(program):
+def programToCommands(program, deltasToSourceValues=None, numRegions=0):
```
blueshade7: Maybe we might want to separate VF versions of `programToCommands()` and `commandsToProgram()` from the existing non-VF functions. It seems cleaner & clearer at expense of some code duplicates.

Lib/fontTools/varLib/models.py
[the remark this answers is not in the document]
readroberts: If you haven't looked at the font data, then the 'after' graphs do look odd. What you can't see in the graphs is that each dot is not one, but two region locations, separated by one design space unit. If the 'after'  graphs are incorrect, then I am missing something basic. My model is that for a series of regions along one axis, that the area of influence for each region is specified by the triplet of location axis coordinates (low, peak,high), where the scalar for the region is 0 below 'low' and above 'high', and is 1.0 at 'peak'. 'Low' and 'high' are are the 'peak' location of respectively the preceding and following regions. Can you tell me what is wrong with this model, or point me to a discussion that will explain it?

Lib/fontTools/cffLib/specializer.py
readroberts: The 'is _CFF2' state would have to be passed in as a function argument: it is not currently available in the function. This would certainly be a simpler solution.

Lib/fontTools/cffLib/specializer.py
readroberts: I chose to represent the blend values as the actual master values in the 'commands' list rather than deltas, as I thought it useful to have access to this representation - there is no need to do so for technical reasons.  Doing this does require a lot more complexity in the API. I just built a version that does not convert between the delta and master values, and the changes were easy. I'm fine with going this way.

This also eliminates the need for the varStore.py::VarDataItem class in merging CFF2 charstrings. However, I'd like to keep this class anyway, as it is is useful for converting deltas to region values and back. The support for doing so with the current models.py::VariationModel class is all that is needed in this context, but there is no support for doing so when working with a built VF font and only a VarStore.

Some additional complexity is still required to convert a blend op in the 'program' list to a list with some blend values in the 'commands' list, even without the conversion of delta values. I do see that having a VF-specific version of programsToCommand() and commandsToProgram() will allow keeping the original much simpler version of these functions, but the more complex versions will still need to exist, and the developer then needs to be aware of both versions and call the right one.

readroberts:
I discussed this PR with Ariza, Cosimo and Behdad. I will close this PR, and open a  new branch and PR with the changes. 
-  In specializer.py, I will move the changes to convert between a blend operator and blend list arguments from programToCommand() and commandToProgram() to generalizeCommand() and specializeCommand(). 
- I will also preserve the blend values as a default font value and its delta tuple, rather than converting the deltas to source font values and back. 
- I will back out my change to models.py: I now understand that the original logic is correct, and is necessary to deal with a more complex problem than I had envisaged. For the case of a series of intermediate masters on one axis, I saw that the existing fonttools code extended the region influence of each intermediate master all the way to the last masters on the axis, and thought this was a bug. However, using this model, and sorting the master correctly, it becomes much easier  to solve for the region of influence of off-axis intermediate masters. The downside is that for simple cases, you end up with a lot more scaling factors to apply than is necessary with the simpler solutions. The upside is that the size of the deltas are smaller, and can take less space, the design space can have axes without a master at the end point, and you can implement virtual axes that act as a control comprising several real axes.

PR #1477, opened 2019-01-30
[varLib] Fix MVAR generation for sparse masters

madig:
MVAR table generation currently blows up if you throw ufo2ft-2.7.0-style sparse masters (SFNT objects without an `OS/2` and other non-outline-related tables) into the mix.

The core issue seems to be that `VariationModel::getDeltas` assumes that are as many master values (values from an existing `OS/2` table) as there are `deltaWeights`. This assumption breaks because there are no master values from sparse masters, yet they have `deltaWeights`.

This blocks https://github.com/googlei18n/fontmake/pull/501.

Investigation ongoing.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -886,7 +886,7 @@ def load_masters(designspace, master_finder=lambda s: s):
 				# 2. A SourceDescriptor's path might point an OpenType binary, a
 				# TTX file, or another source file (e.g. UFO), in which case we
 				# resolve the path using 'master_finder' function
-				font = _open_font(master.path, master_finder)
+				master.font = font = _open_font(master.path, master_finder)
 				master_fonts.append(font)
 
 	return master_fonts
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -521,7 +521,11 @@ def _add_MVAR(font, masterModel, master_ttfs, axisTags):
 	# the minimum FWord (int16) value, was chosen for its unlikelyhood to appear
 	# in real-world underline position/thickness values.
 	specialTags = {"unds": -0x8000, "undo": -0x8000}
+
 	for tag, (tableTag, itemName) in sorted(MVAR_ENTRIES.items(), key=lambda kv: kv[1]):
+		# For each tag, fetch the associated table from all fonts (or not when we are
+		# still looking at a tag from the same tables) and set up the variation model
+		# for them.
 		if tableTag != lastTableTag:
 			tables = fontTable = None
 			if tableTag in font:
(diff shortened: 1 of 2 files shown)


readroberts:
I agree there is a bug, but argue with how it is stated. It is by design that VariationModel::getDeltas assumes that there are as many deltas as master values - this leads to useful behavior. I suggest that problem is that the logic needs to make a different subModel for each different set of master values. See varLib.__init__._add_HVAR() lines 447-450 in fonttools 37.1. I don't immediately understand what is broken in  _add_MVAR(), as I see code that looks like it should take care of this case, but I think the answer is to use the right subModel and build a new VariationData table, not to change the behavior of getDeltas().

behdad:
These lines are supposed to take care of this:
```
                model, tables = masterModel.getSubModel(tables)
                store_builder.setModel(model)
```

I need test interpolatable-ttf masters and a designspace to debug / fix this.

behdad:
Or just paste the error message!

madig:
The test case in this PR provides it.

```
_______________ BuildTest.test_varlib_build_sparse_masters_MVAR _______________

self = <varLib.varLib_test.BuildTest testMethod=test_varlib_build_sparse_masters_MVAR>

    def test_varlib_build_sparse_masters_MVAR(self):
        ds_path = self.get_test_input("SparseMasters.designspace")
        ds = DesignSpaceDocument.fromfile(ds_path)
        load_masters(ds)

        # Trigger MVAR generation so varLib is forced to create deltas with a
        # sparse master inbetween.
        ds.sources[0].font["OS/2"].sxHeight = 100
        ds.sources[2].font["OS/2"].sxHeight = 800

>       varfont, _, _ = build(ds)

Tests\varLib\varLib_test.py:374:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
lib\fontTools\varLib\__init__.py:808: in build
    _add_MVAR(vf, model, master_fonts, axisTags)
lib\fontTools\varLib\__init__.py:551: in _add_MVAR
    base, varIdx = store_builder.storeMasters(master_values)
lib\fontTools\varLib\varStore.py:88: in storeMasters
    deltas = self._model.getDeltas(master_values)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <fontTools.varLib.models.VariationModel object at 0x000001EB563306D8>
masterValues = [100, 800]

    def getDeltas(self, masterValues):
>       assert len(masterValues) == len(self.deltaWeights)
E    AssertionError
```

Haven't gotten around to actually trying to fix this yet :(

madig:
If I change the values for sCapHeight and generate a MVAR entry for that, it works. sCapHeight also happens to be the first item the code iterates over. What seems to happen is that the model is set with each iteration over the MVAR tags. The first time tables are fetch, a `None` is inserted. `getSubModel` removes it, so subsequent calls to it will not see the `None` of the first time and the code path is different. Setting the model just once after tables have been fetched seems to make stuff work?

madig:
@behdad can I merge?

behdad:
Yes please.

PR #1484, opened 2019-02-04
[feaLib] Consistently escape glyph names

khaledhosny:
Add table names to the reserved keywords and make sure glyph names are always escaped written written back to feature files. This probably only affects AST tree that is created programmatically, not by parsing existing feature files.

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -102,7 +102,8 @@ fea_keywords = set([
     "required", "righttoleft", "reversesub", "rsub",
     "script", "sub", "substitute", "subtable",
     "table",
-    "usemarkfilteringset", "useextension", "valuerecorddef"]
+    "usemarkfilteringset", "useextension", "valuerecorddef",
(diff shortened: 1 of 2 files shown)


khaledhosny:
I attempted that, but we don’t have pure ast tests and a parser test won’t work as the glyph names will have to be escaped already.

khaledhosny:
The AppVeyor failure seems unrelated.

anthrotype:
it's strange.. it's attempting to build scipy from source instead of downloading/installing the pre-compiled wheel
https://ci.appveyor.com/project/fonttools/fonttools/builds/22132850/job/f6plp3eum41l649r#L106
PyPI has wheels for scipy 1.2.0 for windows python 2.7
https://pypi.org/project/scipy/1.2.0/#files

I hope this gets fixed by itself, I don't have a windows box at hand to test this out locally.

khaledhosny:
Seems to have fixed itself, master is all green now.

anthrotype:
even weirder

PR #1488, opened 2019-02-05
WIP: Fix deepcopy-ing of CFFs

madig:
Deepcopying a TTFont with a CFF table can result in an infinite recursion on Python 3 under certain circumstances.

@anthrotype found that CFFFontSet keeps a reference to the TTFont object that holds it to get at the glyph order at some point: https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/cffLib/__init__.py#L34. Maybe this messes with `deepcopy`.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -890,7 +890,7 @@ def load_masters(designspace, master_finder=lambda s: s):
 				# 2. A SourceDescriptor's path might point an OpenType binary, a
 				# TTX file, or another source file (e.g. UFO), in which case we
 				# resolve the path using 'master_finder' function
-				font = _open_font(master.path, master_finder)
+				master.font = font = _open_font(master.path, master_finder)
 				master_fonts.append(font)
 
 	return master_fonts
--- a/Tests/cffLib/cffLib_test.py
+++ b/Tests/cffLib/cffLib_test.py
@@ -2,6 +2,8 @@ from __future__ import print_function, division, absolute_import
 from fontTools.cffLib import TopDict, PrivateDict, CharStrings
 from fontTools.misc.testTools import parseXML, DataFilesHandler
 from fontTools.ttLib import TTFont
+import copy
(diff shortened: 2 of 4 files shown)


anthrotype:
I was checking as well during my flight, but I don’t think it’s that. It has to do with the custom `BaseDict.__getattr__` method, it seems to not play nicely with deepcopy. I need to study the code a bit more

madig:
Yeah, implementing

```Python
	def __deepcopy__(self, memo):
		from copy import deepcopy

		cls = self.__class__
		obj = cls.__new__(cls)
		for k, v in self.__dict__.items():
			if k == "otFont":
				obj.otFont = None
			else:
				obj.__dict__[k] = deepcopy(v, memo)
		return obj
```

on CFFFontSet doesn't help.

madig:
Stepping through this, it seems that `BaseDict.__getattr__` is getting repeatedly called with `name="rawDict"`, which executes `self.rawDict.get(name, None)` (rawDict being a PrivateDict), which in it's `__getattr__` calls `BaseDict.__getattr__`, i.e. loops back.

madig:
Actually, there's infinite recursion even when I remove the `__getattr__` on PrivateDict. Stopping on the first line of BaseDict's `__getattr__` when name is `rawDict`, even accessing `self.rawDict` in the debugger causes infinite recursion.

madig:
Maybe because `self.rawDict` doesn't exist and as such triggers calling `__getattr__` on itself, which we overrode to call ourselves.

anthrotype:
I have no internet on my laptop so I just sent you a photo of a git patch via WhatsApp LOL 😂

anthrotype:
We need to implement custom `__getstate__` and `__setstate__` for BaseDict so that both deepcopy and pickle work. I can pick up from here

anthrotype:
fixed on master with https://github.com/fonttools/fonttools/commit/649dc49dbaf056545a117278190a9586d3eb9af5

PR #1495, opened 2019-02-09
Prevent UnicodeDecodeError on py27; this fixes #1494

### LINKED ISSUE #1494: [ufoLib] UnicodeDecodeError in py27
```python
# coding: utf-8
from fontTools.ufoLib import UFOReader
UFOReader('zá意.txt', validate=False)
```
```
  File "ft.py", line 5, in <module>
    UFOReader('zá意.txt', validate=False)
  File "fontTools/ufoLib/__init__.py", line 209, in __init__
    structure = _sniffFileStructure(path)
  File "fontTools/ufoLib/__init__.py", line 1602, in _sniffFileStructure
    "The specified UFO does not have a known structure: '%s'" % ufo_path
UnicodeDecodeError: 'ascii' codec can't decode byte 0xcc in position 2: ordinal not in range(128)
```

justvanrossum:
I forgot, there wasn't something agains using `%r` for paths? Using it here does prevent #1494 from happening. I could add a test, but it seems it would require a new test file, as all ufoLib test files seem to have a special subject, and none seems to fit the bill for this little thing.

### DIFF
--- a/Lib/fontTools/ufoLib/__init__.py
+++ b/Lib/fontTools/ufoLib/__init__.py
@@ -1599,10 +1599,10 @@ def _sniffFileStructure(ufo_path):
 		return UFOFileStructure.PACKAGE
 	elif os.path.isfile(ufo_path):
 		raise UFOLibError(
-			"The specified UFO does not have a known structure: '%s'" % ufo_path
+			"The specified UFO does not have a known structure: %r" % ufo_path
 		)
 	else:
-		raise UFOLibError("No such file or directory: '%s'" % ufo_path)
+		raise UFOLibError("No such file or directory: %r" % ufo_path)
 
 
 def makeUFOPath(path):


anthrotype:
Wouldn’t this use the bytes repr `b’some encoded path name’`?
How about instead decoding using `sys.gefilesystemencoding()`?

justvanrossum:
> Wouldn’t this use the bytes repr `b’some encoded path name’`?

Yes. Would that be bad? It's an error message after all. Also: I _think_ in Py3 in general file paths are `str`, not `bytes`, and this is purely a Py2 issue.

> How about instead decoding using `sys.gefilesystemencoding()`?

Sure, but that would need to be done conditionally, no?

I would not be opposed if we just close this as "won't fix", as it's not a critical bug, and it's Py2-only.

anthrotype:
Wontfix sounds good

justvanrossum:
My main reason to reject this PR is that `%r` is not nice to `os.PathLike` objects. They should look like a path with `%s`, but may look like `PosixPath('test.ufo')` when using `%r`.

PR #1499, opened 2019-02-12
[subset CFF] Fix de-subroutinizing bug when subroutines contain hints

readroberts:
Fixes issue 1493

The  code was skipping executing a subroutine if it had already been desubroutinized. The initial set of vstemhm and hstemhm operators and values may be in a subroutine. If a charstring is being executed which calls such subroutines, they still need to be executed in order to count the number of hint values seen, so that the byte length of the hintmask can be calculated.

I fixed this bug by executing subroutines even if they have already been desubroutinized, as long  as  (we don't know yet if we are doing hintmasks) or ( we do need a hintmask, but have not yet seen it).

### DIFF
--- a/Lib/fontTools/subset/cff.py
+++ b/Lib/fontTools/subset/cff.py
@@ -352,28 +352,45 @@ class _DehintingT2Decompiler(psCharStrings.T2WidthExtractor):
 				if subr_hints.status == 0:
 					hints.last_hint = index
 				else:
-					hints.last_hint = index - 2 # Leave the subr call in
+					hints.last_hint = index - 2  # Leave the subr call in
 
 		elif subr_hints.status == 0:
 			hints.deletions.append(index)
 
 		hints.status = max(hints.status, subr_hints.status)
 
+class StopHintCountEvent(Exception):
+	pass
+
+
+stop_hintcount_ops = ("op_hstem", "op_vstem", "op_rmoveto", "op_hmoveto",
+						"op_vmoveto")
+
+
 class _DesubroutinizingT2Decompiler(psCharStrings.SimpleT2Decompiler):
 
 	def __init__(self, localSubrs, globalSubrs, private=None):
-		psCharStrings.SimpleT2Decompiler.__init__(self,
-							  localSubrs,
-							  globalSubrs, private)
+		psCharStrings.SimpleT2Decompiler.__init__(self, localSubrs, globalSubrs,
+												private)
 
 	def execute(self, charString):
+		self.need_hintcount = True  # until proven otherwise
+		for op_name in stop_hintcount_ops:
+			setattr(self, op_name, self.stop_hint_count)
+
 		if hasattr(charString, '_desubroutinized'):
+			if self.need_hintcount and self.callingStack:
+				try:
+					psCharStrings.SimpleT2Decompiler.execute(self, charString)
+				except StopHintCountEvent:
+					del self.callingStack[-1]
+				return
 			return
 
 		charString._patches = []
 		psCharStrings.SimpleT2Decompiler.execute(self, charString)
 		desubroutinized = charString.program[:]
-		for idx,expansion in reversed (charString._patches):
+		for idx, expansion in reversed(charString._patches):
 			assert idx >= 2
 			assert desubroutinized[idx - 1] in ['callsubr', 'callgsubr'], desubroutinized[idx - 1]
 			assert type(desubroutinized[idx - 2]) == int
@@ -401,9 +418,24 @@ class _DesubroutinizingT2Decompiler(psCharStrings.SimpleT2Decompiler):
 		psCharStrings.SimpleT2Decompiler.op_callgsubr(self, index)
 		self.processSubr(index, subr)
 
+	def stop_hint_count(self, index=None):
+		self.need_hintcount = False
+		for op_name in stop_hintcount_ops:
+			setattr(self, op_name, None)
+		cs = self.callingStack[-1]
+		if hasattr(cs, '_desubroutinized'):
+			raise StopHintCountEvent()
+
+	def op_hintmask(self, index):
+		psCharStrings.SimpleT2Decompiler.op_hintmask(self, index)
+		if self.need_hintcount:
+			self.stop_hint_count()
+
 	def processSubr(self, index, subr):
 		cs = self.callingStack[-1]
-		cs._patches.append((index, subr._desubroutinized))
+		if not hasattr(cs, '_desubroutinized'):
+			cs = self.callingStack[-1]
+			cs._patches.append((index, subr._desubroutinized))
 
 
 @_add_method(ttLib.getTableClass('CFF '))
@@ -592,4 +624,3 @@ def remove_unused_subroutines(self):
 		# Cleanup
 		for subrs in all_subrs:
 			del subrs._used, subrs._old_bias, subrs._new_bias
-
--- a/Lib/fontTools/subset/cff.py
+++ b/Lib/fontTools/subset/cff.py
@@ -363,11 +363,11 @@ class StopHintCountEvent(Exception):
 	pass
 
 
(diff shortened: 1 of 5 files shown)


Lib/fontTools/subset/cff.py
```diff
@@ -352,28 +352,45 @@ def processSubr(self, index, subr):
...
 
 	def __init__(self, localSubrs, globalSubrs, private=None):
-		psCharStrings.SimpleT2Decompiler.__init__(self,
-							  localSubrs,
-							  globalSubrs, private)
+		psCharStrings.SimpleT2Decompiler.__init__(self, localSubrs, globalSubrs,
+												private)
 
 	def execute(self, charString):
+		self.need_hintcount = True  # until proven otherwise
```
cjchapman: I suggest moving the initialization of `need_hintcount` into the `__init__()` method.

Lib/fontTools/subset/cff.py
```diff
@@ -401,9 +418,24 @@ def op_callgsubr(self, index):
 		psCharStrings.SimpleT2Decompiler.op_callgsubr(self, index)
 		self.processSubr(index, subr)
 
+	def stop_hint_count(self, index=None):
```
cjchapman: I think you can remove `index` from the parameter list as it's not used in this method, nor in the calls to this method.

Lib/fontTools/subset/cff.py
readroberts: The problem with that is that the class instance can be re-used for more than one charstring. Running it on one charstring will leave self.need_hintcount set to False, and it needs to be True at the start of execute(). I could set the value in init, and then extend the parent class's reset() method to set this to True. However, since this class is pretty specialized, I chose to set close to were it is used.

Lib/fontTools/subset/cff.py
readroberts: Actually, it is used. All of the functions in the stop_hintcount_ops list are called with two arguments: self , and index. stop_hint_count() needs accept the second argument, even though it isn't used in this functions. These functions are all called by  misc.psCharstrings.SimpleT2Decompiler.execute(), with the same handler call.

Lib/fontTools/subset/cff.py
cjchapman: OK.

Lib/fontTools/subset/cff.py
cjchapman: OK.

anthrotype:
when calling the subsetter with `--desubroutinize` as well as with `--no-hinting`, we still get AttributeError for `GlobalSubrs` inside `remove_unused_subroutines` method, which is called at the end of `remove_hinting` method. This is because the `desubroutinize` method deletes the attribute on the FontDict.
This patch fixes this. Do you mind applying this and also add a little test case for using the two options in combination?

```diff
diff --git a/Lib/fontTools/subset/cff.py b/Lib/fontTools/subset/cff.py
index ee6048a0..9dc099b8 100644
--- a/Lib/fontTools/subset/cff.py
+++ b/Lib/fontTools/subset/cff.py
@@ -486,9 +486,7 @@ def desubroutinize(self):
                        decompiler.execute(c)
                        c.program = c._desubroutinized
                        del c._desubroutinized
-               # Delete All the Subrs!!!
-               if font.GlobalSubrs:
-                       del font.GlobalSubrs
+               # Delete all the local subrs
                if hasattr(font, 'FDArray'):
                        for fd in font.FDArray:
                                pd = fd.Private
@@ -502,6 +500,7 @@ def desubroutinize(self):
                                del pd.Subrs
                        if 'Subrs' in pd.rawDict:
                                del pd.rawDict['Subrs']
+       # as well as the global subrs
        cff.GlobalSubrs.clear()
```

Note that it is not necessary to call `del font.GlobalSubrs` to delete the global subroutines, since the attribute is shared by all the font dicts in the CFFFontSet and we call `clear()` at the end of `desubroutinize` to set it empty.

Lib/fontTools/subset/cff.py
anthrotype: if it's unused, it's clearer if you just use `*args, **kwargs` then

Lib/fontTools/subset/cff.py
```diff
@@ -401,9 +418,24 @@ def op_callgsubr(self, index):
...
+	def op_hintmask(self, index):
+		psCharStrings.SimpleT2Decompiler.op_hintmask(self, index)
+		if self.need_hintcount:
+			self.stop_hint_count()
+
 	def processSubr(self, index, subr):
 		cs = self.callingStack[-1]
-		cs._patches.append((index, subr._desubroutinized))
+		if not hasattr(cs, '_desubroutinized'):
+			cs = self.callingStack[-1]
```
anthrotype: this line 437 is repeated two lines above

Lib/fontTools/subset/cff.py
```diff
@@ -352,28 +352,45 @@ def processSubr(self, index, subr):
...
+		for op_name in stop_hintcount_ops:
+			setattr(self, op_name, self.stop_hint_count)
+
 		if hasattr(charString, '_desubroutinized'):
+			if self.need_hintcount and self.callingStack:
+				try:
+					psCharStrings.SimpleT2Decompiler.execute(self, charString)
+				except StopHintCountEvent:
+					del self.callingStack[-1]
+				return
```
anthrotype: this `return` statement is also unnecessary, the following line does the same

Lib/fontTools/subset/cff.py
```diff
@@ -352,28 +352,45 @@ def processSubr(self, index, subr):
...
 		elif subr_hints.status == 0:
 			hints.deletions.append(index)
 
 		hints.status = max(hints.status, subr_hints.status)
 
+class StopHintCountEvent(Exception):
+	pass
+
+
+stop_hintcount_ops = ("op_hstem", "op_vstem", "op_rmoveto", "op_hmoveto",
```
anthrotype: i'd make this `stop_hintcount_ops` a class variable, it's only used within `_DesubroutinizingT2Decompiler`

Lib/fontTools/subset/cff.py
readroberts: Agreed.

Lib/fontTools/subset/cff.py
readroberts: Right.

Lib/fontTools/subset/cff.py
readroberts: Right.

Lib/fontTools/subset/cff.py
readroberts: Right. Thanks for the review! Will fix all these, including the patch for the GlobalSubrs bug.

readroberts: @anthrotype I see that there is already a test, subset_test.py::test_no_hinting_desubroutinize_CFF, which should encounter the AttributeError for `GlobalSubrs`. I can't reproduce this bug, either with this test, or using the command-line "pyftsubset Lobster.subset.otf  --gids=0-11 --desubroutinize --no-hinting". How did you reproduce this problem?

readroberts: @anthrotype Never mind - I realize that the test needs a font with a non-empty GlobalSubrs. Have test case, will add.

Lib/fontTools/subset/cff.py
```diff
@@ -470,6 +498,7 @@ def desubroutinize(self):
 				del pd.Subrs
 			if 'Subrs' in pd.rawDict:
 				del pd.rawDict['Subrs']
+	# as well as the global subrs
```
behdad: How is this change relevant to the issue being fixed?

anthrotype: I asked Read to apply that patch since he was working on related issues with the —desubroutinize option:
https://github.com/fonttools/fonttools/pull/1499#issuecomment-463309891

PR #1500, opened 2019-02-12
[svgLib] add rudimentary support for shapes

rsheeter:
Make it possible to import SVGs with simple uses of `<rect>`, `<circle>`, `<polygon>` by converting them to `<path>`


Lib/fontTools/svgLib/path/shapes.py
```diff
@@ -0,0 +1,120 @@
+def _prefer_non_zero(*args):
+    for arg  in args:
+      if arg != 0:
+        return arg
```
anthrotype: indentation should use four spaces like in the rest of the module

Lib/fontTools/svgLib/path/shapes.py
```diff
@@ -0,0 +1,120 @@
...
+    return 0.
+
+def _ntos(n):
+  # %f likes to add unnecessary 0's, %g isn't consistent about # decimals
```
anthrotype: same

anthrotype:
closing this, we'll continue on #1508

Issue #1468, opened 2019-01-26
[feaLib] Special case of contextual positioning not supported


punchcutter:
Just ran into this trying to run fontmake on a Glyphs source file with added contextual kerning. 

From https://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#6.h.v
> In order to make pair positioning rules easier to read and write as contextual kern pairs, the feature file syntax will identify a special case of contextual rule which contains only one marked glyph or glyph class, followed by one or more un-marked glyph or glyph class, plus a value record. This will be treated as a contextual pair positioning statement, and will be the only one case where a value record may follow an un-marked glyph.

The example shown is 

`position L' quoteright -150;` 

but makeotf will also allow a more complicated sequence like 

`pos @someclass' @anotherclass [someglyph anotherglyph moreglyphs] <30 0 30 0>;`

When this format is used feaLib loses the ValueRecord at https://github.com/fonttools/fonttools/blob/18ac453781029783e1d340ada7699df966c2b04b/Lib/fontTools/feaLib/parser.py#L407

Handling this would need an update to `asFea()` at https://github.com/fonttools/fonttools/blob/18ac453781029783e1d340ada7699df966c2b04b/Lib/fontTools/feaLib/ast.py#L1066 if 100% round tripping were desired. As is it will only do the standard format and put the ValueRecord after the `'`.

anthrotype:
@punchcutter thanks for the report. I can have a look at this after I'm back to work in the second week of February -- unless sombody else wishes to tackle this before.

marekjez86:
@punchcutter : thanks for debugging the problem

marekjez86:
@anthrotype : any update on this?
cc @davelab6 
I cannot assign bugs in this repository hence I'm using comments

anthrotype:
Thanks for checking. I am busy with internal training for at least another two weeks, I am afraid I won't be able to look into this before then.

punchcutter:
It's not a huge deal and isn't blocking anything, but I'll send a pull request today.

anthrotype:
that'd be awesome! thanks

PR #1501, opened 2019-02-12
Read special case of contextual positioning

punchcutter:
As explained here: https://github.com/fonttools/fonttools/issues/1468
If the ValueRecord comes after a suffix instead of directly after the affected glyphs then we still want to read it. Currently the ValueRecord is silently skipped, resulting in missing kerning. 

There's really no way to roundtrip with asFea() because from a compiled source we can't know that the original FDK code was written this way. That means if the original code is 

`position L' quoteright -150;`

then the asFea() output will become

`position L' -150 quoteright;`

I don't see any problem with this as this "special case" format is really just a convenience for FDK source and I assume most code is written the normal way. If not we probably would have run into this long ago.

### DIFF
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -404,7 +404,9 @@ class Parser(object):
             return ([], prefix, [None] * len(prefix), values, [], hasMarks)
         else:
             assert not any(values[:len(prefix)]), values
-            values = values[len(prefix):][:len(glyphs)]
+            format1 = values[len(prefix):][:len(glyphs)]
+            format2 = values[len(prefix) + len(glyphs):][:len(suffix)]
+            values = format2 if format2 and isinstance(format2[0], self.ast.ValueRecord) else format1
             return (prefix, glyphs, lookups, values, suffix, hasMarks)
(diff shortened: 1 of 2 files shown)


Lib/fontTools/feaLib/parser.py
```diff
@@ -404,7 +404,9 @@ def parse_glyph_pattern_(self, vertical):
             return ([], prefix, [None] * len(prefix), values, [], hasMarks)
         else:
             assert not any(values[:len(prefix)]), values
-            values = values[len(prefix):][:len(glyphs)]
+            format1 = values[len(prefix):][:len(glyphs)]
+            format2 = values[len(prefix) + len(glyphs):][:len(suffix)]
```
anthrotype: i got confused here by the relative priority of `+` and `:` operators. Some parentheses may help readability maybe

```suggestion
            format2 = values[(len(prefix) + len(glyphs)):][:len(suffix)]
```

anthrotype:
> There's really no way to roundtrip with asFea()

i'm ok if asFea() doesn't roundtrip this special case, as long as the meaning of the feature code doesn't change.

PR #1509, opened 2019-02-14
A couple ufoLib.pointPen → pens.pointPen

khaledhosny:


Lib/fontTools/ttLib/ttFont.py
```diff
@@ -774,7 +774,7 @@ def draw(self, pen):
 		glyph.draw(pen, glyfTable, offset)
 
 	def drawPoints(self, pen):
-		"""Draw the glyph onto PointPen. See ufoLib.pointPen for details
+		"""Draw the glyph onto PointPen. See pens.pointPen for details
```
justvanrossum: `fontTools.pens.pointPen` (I suppose this is a remnant from when ufoLib was not part of fontTools yet.)

khaledhosny: Sure.

justvanrossum:
Thanks!

PR #1511, opened 2019-02-19
[feaLib] Add forceChain to MultipleSubstStatement

khaledhosny:
For consistency with other substitution statements.

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -905,20 +905,22 @@ class MarkMarkPosStatement(Statement):
 
 
 class MultipleSubstStatement(Statement):
-    def __init__(self, prefix, glyph, suffix, replacement, location=None):
+    def __init__(self, prefix, glyph, suffix, replacement, forceChain, location=None):
         Statement.__init__(self, location)
         self.prefix, self.glyph, self.suffix = prefix, glyph, suffix
         self.replacement = replacement
+        self.forceChain = forceChain
 
     def build(self, builder):
         prefix = [p.glyphSet() for p in self.prefix]
         suffix = [s.glyphSet() for s in self.suffix]
         builder.add_multiple_subst(
-            self.location, prefix, self.glyph, suffix, self.replacement)
+            self.location, prefix, self.glyph, suffix, self.replacement,
+            self.forceChain)
 
     def asFea(self, indent=""):
         res = "sub "
-        if len(self.prefix) or len(self.suffix):
+        if len(self.prefix) or len(self.suffix) or self.forceChain:
             if len(self.prefix):
                 res += " ".join(map(asFea, self.prefix)) + " "
             res += asFea(self.glyph) + "'"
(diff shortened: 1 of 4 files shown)


anthrotype:
this looks good to me, however I wonder why this wasn't implemented already.. Does the FEA spec allow this? What does makeotf do in this case?

khaledhosny:
For the record, `makeotf` didn’t support short contextual multiple substitution syntax (a named lookup was needed), but this was fixed in https://github.com/adobe-type-tools/afdko/pull/729. However, feaLib did support this syntax but it was incomplete and this PR fixes this.

dscorbett:
@khaledhosny, isn’t this change backwards-incompatible? Before this change, the fifth argument passed to `MultipleSubstStatement` was `location`, but now it is `forceChain`.

khaledhosny:
Yes, I guess, but I don’t think we have API stability guarantees for these new libs.

dscorbett:
I had assumed all the APIs were stable except for those marked experimental. How can I tell which APIs are stable?

khaledhosny:
That is a question for @anthrotype.

anthrotype:
> How can I tell which APIs are stable?

that's a hard question, especially since python doesn't enforce any strict rules. Ideally, functions/classes that are not prefixed with an underscore, and that have some documentation (e.g. docstrings) are intended to be public, but our documentation efforts have not been systematic...

Some modules have indeed been marked as experimental (e.g. varLib), but they have been so for a while so much that I suspect people are treating them as stable.

I haven't looked into it, but if you can make that specific change mentioned here backward compatible, i'm ok with that.

khaledhosny:
We can make `forceChain` last (which would require making it a keyed argument), but that would be inconsistent with other functions. Unless we have evidence this change is breaking existing code, I suggest we keep things consistent.

dscorbett:
To be backwards-compatible, the signature would have to be `__init__(self, prefix, glyph, suffix, replacement, location=None, forceChain=False)`, but putting a parameter after `location` is inconsistent with the rest of the module. Under the assumption that any caller that sets `location` does so with a named argument, a practically backwards-compatible signature is `__init__(self, prefix, glyph, suffix, replacement, forceChain=False, location=None)`. That is still inconsistent with the other classes that take `forceChain`, where it is a required parameter, though. On the other hand, keeping it as it is will break any code that uses `MultipleSubstStatement` (unless, like me, they’ve added `<=3.38.0` to their requirements.txt). It all depends how you rate API stability vs. API consistency. What do you think?

dscorbett:
>Unless we have evidence this change is breaking existing code

Oh, it definitely breaks my code. That’s why I brought this up. But since I’ve changed my requirements.txt, maybe that doesn’t matter.

anthrotype:
the location argument is intended to be keyword only, and I believe internal library code is consistent with that. So the suggestion to make forceChain an optional keyword argument that defaults to False seems reasonable to me. The inconsistency can be documented as for not breaking backward compatibility.

We haven't released this patch yet in a stable version on PyPI, so unless you are installing from git repo (which is by definition in-development and subject to break) you should be fine.

anthrotype:
@dscorbett I just did that in https://github.com/fonttools/fonttools/commit/abae80d7305caf5739d701b92a2ca4edeaf87455

let me know if that works for you, thanks

dscorbett:
Thanks, that looks good.

Issue #1516, opened 2019-02-22
[subset/var] subsetting VVAR table fails if VOrgMap non-empty


blueshade7:
Line 1856 - 1857 in `fonttools/Lib/fontTools/subset/__init__.py` reads:
```
	if table.VOrgMap:

		table.RsbMap.mapping = {k:varidx_map[v] for k,v in table.RsbMap.mapping.items()}
```
It should be:
```
	if table.VOrgMap:
		table.VOrgMap.mapping = {k:varidx_map[v] for k,v in table.VOrgMap.mapping.items()}
```
Otherwise subsetting VVAR table with non-empty VOrgMap fails as:
```
Traceback (most recent call last):
  File "/usr/local/bin/pyftsubset", line 11, in <module>
    load_entry_point('fonttools==3.37.4.dev0', 'console_scripts', 'pyftsubset')()
  File "/usr/local/lib/python2.7/site-packages/fonttools-3.37.4.dev0-py2.7.egg/fontTools/misc/loggingTools.py", line 375, in wrapper
    return func(*args, **kwds)
  File "/usr/local/lib/python2.7/site-packages/fonttools-3.37.4.dev0-py2.7.egg/fontTools/subset/__init__.py", line 2827, in main
    subsetter.subset(font)
  File "/usr/local/lib/python2.7/site-packages/fonttools-3.37.4.dev0-py2.7.egg/fontTools/subset/__init__.py", line 2630, in subset
    self._subset_glyphs(font)
  File "/usr/local/lib/python2.7/site-packages/fonttools-3.37.4.dev0-py2.7.egg/fontTools/subset/__init__.py", line 2576, in _subset_glyphs
    retain = table.subset_glyphs(self)
  File "/usr/local/lib/python2.7/site-packages/fonttools-3.37.4.dev0-py2.7.egg/fontTools/subset/__init__.py", line 1857, in subset_glyphs
    table.RsbMap.mapping = {k:varidx_map[v] for k,v in table.RsbMap.mapping.items()}
  File "/usr/local/lib/python2.7/site-packages/fonttools-3.37.4.dev0-py2.7.egg/fontTools/ttLib/tables/otBase.py", line 546, in __getattr__
    raise AttributeError(attr)
```

blueshade7:
[AdobeVFPrototype.abc_vvar.otf.zip](https://github.com/fonttools/fonttools/files/2895949/AdobeVFPrototype.abc_vvar.otf.zip)
[...]

PR #1517, opened 2019-02-22
Fix issue #1516: subset VVAR

blueshade7:
This PR fixes issue #1516.
It looks like a simple copy & paste & forgot-to-edit bug.
Fix is simple & straightforward too.


behdad:
Thanks.  Can you add tests please?

blueshade7:
The two fonts attached to #1516 fails before & succeeds after with the following command line:
```
pyftsubset --text=ac AdobeVFPrototype.abc_vvar.otf --output-file=output.otf
```

PR #1520, opened 2019-02-26
[feaLib] Support subtable statement in more places

khaledhosny:
Although AFDKO’s makeotf supports subtable breaks only in pair pos lookups, I believe that is an implementation limit that we don’t have to follow; all OpenType lookups has subtables and it should be possible to code that in the feature files.

This PR adds `subtable` statement support to many lookups, but not all of them; I added it to the lookups I had tests for or was easy enough to construct tests for.

This also make unsupported `subtable` a warning instead of an error, to follow makeotf.

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -1509,11 +1509,22 @@ class SingleSubstBuilder(LookupBuilder):
                 self.mapping == other.mapping)
 
     def build(self):
-        subtable = otl.buildSingleSubstSubtable(self.mapping)
-        return self.buildLookup_([subtable])
+        subtables = []
+        mapping = {}
+        for key in self.mapping:
+            if key[0] == self.SUBTABLE_BREAK_:
+                subtables.append(otl.buildSingleSubstSubtable(mapping))
+                mapping = {}
+            else:
+                mapping[key] = self.mapping[key]
+        subtables.append(otl.buildSingleSubstSubtable(mapping))
+        return self.buildLookup_(subtables)
 
     def getAlternateGlyphs(self):
         return {glyph: set([repl]) for glyph, repl in self.mapping.items()}
+
+    def add_subtable_break(self, location):
+        self.mapping[(self.SUBTABLE_BREAK_, location)] = self.SUBTABLE_BREAK_
 
 
 class ClassPairPosSubtableBuilder(object):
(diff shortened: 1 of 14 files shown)


khaledhosny:
Python < 3.6 fails as I was suspecting, I guess I need to through in a few `OrderedDicts()`, is Python < 3.6 dead yet?

khaledhosny:
The warning vs. error issue is blocking a project I’m working on, so if this PR is going to take long to review I can split that part into its own PR.

anthrotype:
Can I ask why do you need that?

I'm slightly worried that feaLib implementation may end up diverging too much from makeotf. The FEA spec explicitly says that `subtable` statement

> is intended for use with only Pair Adjustment Positioning Format 2 (i.e. pair class kerning). 

If anything we should also agree on modifying the spec itself to lift that restriction.

/cc @readroberts @miguelsousa

anthrotype:
@khaledhosny I'm ok if you want to push the commit that demotes unsupported `subtable` statement from error to warning as a separate PR.

anthrotype:
@khaledhosny maybe you need this to manually fix offset overflow errors?

khaledhosny:
> Can I ask why do you need that?

I’m creating feature files from input sources that allow explicitly creating subtables anywhere, and I need to capture that in the feature files. Subtables may have semantic or mere structural implications, but I don’t want to second guess the original sources. It would be helpful also when “decompiling” layout tables into feature files.

> I'm slightly worried that feaLib implementation may end up diverging too much from makeotf. The FEA spec explicitly says that `subtable` statement
> 
> > is intended for use with only Pair Adjustment Positioning Format 2 (i.e. pair class kerning).
> 
> If anything we should also agree on modifying the spec itself to lift that restriction.

That would be nice thing, but if anything I don’t want to restrict us to the limitations of makeotf implementation. After what is the point if having a Python-based implementation, if we can’t develop and extend it.

FontForge already allow `subtable` anywhere in the feature files, by the way.

readroberts:
The feature file syntax should be able to evolve beyond the current limitations of any implementation, including makeotfexe, as long as there is good reason and past behavior isn't broken by the change. In this case, I see no harm in have supporting subtable breaks everywhere, as long as implementations are allowed to warn when this isn't supported. The current language in the spec isn't because Sairus Patel and I were against allowing subtable breaks elsewhere, it was just because we didn't think it was needed anywhere else.

PR #1521, opened 2019-02-27
[designspaceLib] use axis maps in DesignSpaceDocument.normalizeLocation

### LINKED ISSUE #1226: [designspaceLib] normalize() does not consider axisMap

anthrotype:
Fixes #1226

### DIFF
--- a/Lib/fontTools/designspaceLib/__init__.py
+++ b/Lib/fontTools/designspaceLib/__init__.py
@@ -1188,31 +1188,21 @@ class DesignSpaceDocument(LogMixin, AsDictMixin):
         return None
 
     def normalizeLocation(self, location):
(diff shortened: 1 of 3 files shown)


LettError:
Something is failing though.

anthrotype:
importing `fontTools.varLib.models` at the top of `designspaceLib/__init__.py` leads to a circular import, I'll move the import statement inside the method that uses it.

anthrotype:
After this PR, the `fonttools varLib.plot` script should now work as intented, with either a .designspace file as input, or with a list of locations, e.g.:

```
$ fonttools varLib.plot NotoSans-MM.designspace
$ fonttools varLib.plot -1,0 0,0 1,0 0,-1 0,1 0.5,0.5
```

PR #1523, opened 2019-03-03
[cmap,VORG,COLR,otTables] misc idiomatic updates

justvanrossum:
- changed some uses of list(map(...)) to list comprehensions and explicit loops, avoiding possibly large unneeded tmp lists
- replaced use of eval() with int()
- adding tests to verify all changes are covered

(There's still a lot of duplicate logic in various decompile methods, but that's for another time.)

### DIFF
--- a/Lib/fontTools/ttLib/tables/_c_m_a_p.py
+++ b/Lib/fontTools/ttLib/tables/_c_m_a_p.py
@@ -444,11 +444,11 @@ class cmap_format_2(CmapSubtable):
 		nameMap = ttFont.getReverseGlyphMap()
 		lenCharCodes = len(charCodes)
 		try:
-			gids = list(map(operator.getitem, [nameMap]*lenCharCodes, names))
+			gids = [nameMap[name] for name in names]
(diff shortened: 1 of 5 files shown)


Lib/fontTools/ttLib/tables/_c_m_a_p.py
```diff
@@ -444,11 +444,11 @@ def compile(self, ttFont):
 		nameMap = ttFont.getReverseGlyphMap()
 		lenCharCodes = len(charCodes)
```
khaledhosny: Looks like `lenCharCodes` is now unused.

justvanrossum: Thanks, fixed.

justvanrossum:
Ugh, there's a lot more of this list(map(operator...)) stuff in various sources that should be either list comprehensions or explicit loops: V_O_R_G_.py, C_O_L_R_.py and otTables.py

Maybe I'll do them in this PR as well.

PR #1528, opened 2019-03-05
varLib: use vmtx to compute phantom pts; fix the sign of bottomSideY

anthrotype:
when 'vmtx' is present in a font, use that to compute the third and fourth 'phantom points'.
When not present, use the glyph bbox yMax and yMin coords.

NOTE that previously the bottomSideY was incorrectly set to `-glyph.yMin` (with a minus sign).
However, the minus is not needed when we use the bbox.

Positive vertical advance grows towards negative Y axis.
In the absence of vmtx, we default to using `yMax` as the y coordinate of the top origin and `yMax - yMin` as the vertical advance width; so the y coordinate of the advance height _point_ (bottomSideY) is `yMax - (yMax - yMin) => yMin`

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -209,8 +209,40 @@ def _add_stat(font, axes):
 	# TODO make this user-configurable via designspace document
 	stat.ElidedFallbackNameID = 2
 
+
+def _get_phantom_points(font, glyphName, defaultVerticalOrigin=None):
+	glyf = font["glyf"]
+	glyph = glyf[glyphName]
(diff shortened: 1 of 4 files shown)


behdad:
When `vmtx` is not present, glyph vertical origin and advance are computed based on the font's horizontal metrics.  Ie. ascent/descent.  I *think* we use ascent for the top, and use upem as advance in HarfBuzz.  Anyone, using yMin/yMax is just wrong.

anthrotype:
Gotcha, I’ll fix that, thanks

behdad:
lgtm.  Thanks.

PR #1531, opened 2019-03-06
updated inline github issue URLs

justvanrossum:
The forwards still work, but I thought it'd be better to update them to the current repo.


khaledhosny:
There are a few old URLs remaining, under Tests mostly.

PR #1532, opened 2019-03-06
more updated issue URLs

justvanrossum:
Followup to #1531


justvanrossum:
Thanks for having a look, @khaledhosny! I think I covered everything now. Let me know if you see I missed something. (I left the reference in README.rst, as it's a historical reference.)

khaledhosny:
That seems to have caught all of them (I was a bout to make a pull request when you made this one :smile:).

PR #1535, opened 2019-03-10
designspaceLib, findDefault: consider axis mapping

madig:
Consider axis mappings when determining the default source in a Designspace. The minimum/default/maximum values of an axis are always in user space while source and instance locations are always in design space. If a mapping exists for an axis, user and design space diverge.

### DIFF
--- a/Lib/fontTools/designspaceLib/__init__.py
+++ b/Lib/fontTools/designspaceLib/__init__.py
@@ -1176,15 +1176,26 @@ class DesignSpaceDocument(LogMixin, AsDictMixin):
         return None
 
     def findDefault(self):
-        # new default finder
-        # take the sourcedescriptor with the location at all the defaults
-        # if we can't find it, return None, let someone else figure it out
+        """Set and return SourceDescriptor at the default location or None.
+
+        The default location is the set of all `default` values in user space
+        of all axes.
+        """
         self.default = None
+
+        # Convert the default location from user space to design space before comparing
+        # it against the SourceDescriptor locations (always in design space).
+        # Note: given no map, piecewiseLinearMap will simply return the value.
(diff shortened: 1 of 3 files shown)


Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -1176,15 +1176,29 @@ def getAxis(self, name):
...
+        from fontTools.varLib.models import piecewiseLinearMap
+
         self.default = None
+
+        # Convert the default location from user space to design space before comparing
+        # it against the SourceDescriptor locations (always in design space).
+        # Note: given no map, piecewiseLinearMap will simply return the value.
+        axis_mapping = {axis.name: {k: v for k, v in axis.map} for axis in self.axes}
+        default_design_location = {
+            name: piecewiseLinearMap(value, axis_mapping[name])
```
anthrotype: There’s map_forward and map_backward methods of Axis for that

behdad:
Are instance locations in design space? I know we've been talking about needing both.

anthrotype:
Yes so far, instance locations have been only in internal design coordinates, not user-space.
We could specify a new attribute (e.g. `user=“1”`) to allow defining instance coordinates in user space (not in this PR though).

madig:
While fixing that one test case, I noticed that when locations are written out, locations not explicitly set are generated by the writer... in user-space, despite the presence of a map. D'oh.

madig:
Uh... what did I just do :confounded:

madig:
Not sure this is as intended, but `newDefaultLocation` is only used when making a new location in the writer.

madig:
I suspect all of designspaceLib needs to be reviewed for user and design space assumptions.

PR #1539, opened 2019-03-15
Compress type 1 GPOS tables better

mhosken:
The current code has a tendency to create lots of little subtables for a big set of key values. This is not only wasteful (it takes 5 uint16 overhead for a new type 1 subtable, and if your value record is a single value, that means you can get 5 keys in the main coverage table for the same space), it is slower since an engine can only search subtables linearly.

This PR does a more careful calculation of value record size and then decides whether it really needs a subtable or not based on the number of keys and the size of the value.

This PR fails the same tests that the master it is based on fails.


behdad:
Thanks.  I see how this is an improvement.  One day I'll implement an optimal packer.  This is quite similar to the VarStore optimizer.

Lib/fontTools/otlLib/builder.py
```diff
@@ -491,6 +493,14 @@ def _makeDeviceTuple(device):
...
             tuple(device.DeltaValue))
 
+def _getSinglePosValueSize(valueKey):
+    count = 0
+    for k in valueKey[1:]:
+        if hasattr(k[1], '__len__') and len(k[1]):
+            count += len(k[1][3]) + 3
+        else:
+            count += 1
+    return count
```
anthrotype: I'd suggest to give more descriptive names to the variables or add comments, because it's not immediately clear what valueKey is here, and what it contains

Lib/fontTools/otlLib/builder.py
```diff
@@ -428,8 +430,8 @@ def buildSinglePos(mapping, glyphMap):
     # is unique as well. We encode these in format 1 again.
     for key, glyphs in coverages.items():
         if key not in handled:
-            assert len(glyphs) == 1, glyphs
-            st = buildSinglePosSubtable({glyphs[0]: values[key]}, glyphMap)
+            for g in glyphs:
+                st = buildSinglePosSubtable({g: values[key]}, glyphMap)
```
anthrotype: you removed the assertion that `len(glyphs) == 1` but the comment still says that remaining ValueRecords are supposed to be used by a single glyph..

Lib/fontTools/otlLib/builder.py
```diff
@@ -408,7 +408,7 @@ def buildSinglePos(mapping, glyphMap):
     # If a ValueRecord is shared between multiple glyphs, we generate
     # a SinglePos format 1 subtable; that is the most compact form.
     for key, glyphs in coverages.items():
-        if len(glyphs) > 1:
+        if len(glyphs) * _getSinglePosValueSize(key) > 5:
```
anthrotype: this magic number 5 could also be documented better

anthrotype:
@mhosken please let me know if you intend to address the review comments above, so I can go on merging this, thank you.

mhosken:
Sorry. Yes I do plan to tidy this up, although I have a conference next
week so it may be easy or hard ;) But I do intend to get this merged.

Lib/fontTools/otlLib/builder.py
mhosken: It's hard to come up with another name given `valueKey` comes from `_getSinglePosValueKey` which kind of introduces a ValueKey as a concept in the code. Hence using that name here. I've added a comment to the function to help clarify things a little.

Lib/fontTools/otlLib/builder.py
mhosken: comment adjusted.

anthrotype:
Thanks

Lib/fontTools/otlLib/builder.py
```diff
@@ -491,6 +493,14 @@ def _makeDeviceTuple(device):
     return (device.DeltaFormat, device.StartSize, device.EndSize,
             tuple(device.DeltaValue))
 
+def _getSinglePosValueSize(valueKey):
+    count = 0
+    for k in valueKey[1:]:
+        if hasattr(k[1], '__len__') and len(k[1]):
+            count += len(k[1][3]) + 3
```
behdad: The calculation here does not take delta-format into account BTW.

anthrotype: oh.. when the device table has DeltaFormat == 0x8000 (aka VariationIndex table), it is always 3-ushort long, right?

anthrotype: btw, this `otlLib.buildSinglePos` function is currently only used within feaLib and mtiLib, which only ever output DeltaFormat 1, 2 or 3.

behdad: The DeltaFormat == 0x8000 is not handled by code in that module that converts Device table to a tuple.  But that's not what I'm talking about.  The code above assumes that each delta value takes 2 bytes.  That's simply not true.  Each delta value takes one of 1, 2, or 4 *bits* depending on DeltaFormat.

anthrotype: ok now i'm proper confused, so i'll leave that to you

anthrotype: cc @mhosken

behdad: I don't feel touching it either.  Martin can fix at some point.

anthrotype: OK I got it

https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#device-and-variationindex-tables

> Format values 0x0001 to 0x0003 are used for Device tables, and indicate the format of delta adjustment values contained directly within the device table: signed 2-, 4,- or 8-bit values

Martin is just taking the `len()` of the device.DeltaValue (thus assuming one uint16 or 2 bytes per adjustment delta value) but we should instead count the length the packed representation.

As far as I understand, this is not a major issue, we simply don't optimize as much as we could, right?

behdad: Correct.

mhosken: I agree. I saw this when I wrote it, but decided not to complexify the code to handle a pretty rare use case. If someone who is actually using device records wants to refine the code, they can.

readroberts:
Device tables are indeed very rare, but now with variable fonts, we have a lot of VariationIndex tables. I am having to touch these lines in order to support variable fonts with a GPSO vpal feature, so I can fix it. I see the spec is confusing: in the text, it implies that there is a uint16 in the delta array for each integer size in the inclusive size range. Fortunately, the example makes it clear that there is a delta value for each ppem size in the size range, the bit size of a delta value is specified by the delta flag, and these are packed into the uint16 array sequentially as if the array as a continuous byte string.

PR #1544, opened 2019-03-16
Fix CursivePos subtable overflows

dscorbett:


behdad:
Did you got this error, or just implemented it?

This doesn't work. At least not with HarfBuzz. Two neighboring glyphs with their EntryExit records in different subtables will NOT attach.

dscorbett:
I got this error. I guess I’ll have to think of a different solution.

behdad:
Interesting.  For one script?

dscorbett:
Yes, for Duployan. It has a lot of contextual variants.

dscorbett:
When a shaper is checking coverage to choose a subtable in a CursivePos lookup, does it consider both the exit and entry glyphs, just the exit glyph, or just the entry glyph?

behdad:
> When a shaper is checking coverage to choose a subtable in a CursivePos lookup, does it consider both the exit and entry glyphs, just the exit glyph, or just the entry glyph?

Just one.  It used to be checking the entry.  But @mhosken reported that Windows seem to be doing it the other way, so we changed it.  It checks the exit glyph and looks backward now.

dscorbett:
Right: by “table” I meant “lookup”.

behdad:
I'll close this then, until a better solution comes up.

PR #1546, opened 2019-03-17
[ttLib/glyf] Raise more specific error when encountering recursive components

### LINKED ISSUE #1545: TTX crashes when compiling recursive glyph component defintion
If the glyf table in a ttx file (nonsensically) declares a glyph to be a component of itself, ttx in Version 3.0 of fonttools recursively calls getCoordinates until the "maximum recursion depth exceeded while calling a Python object".  The tool ttx ought to report which glyph has the problem.  (I encountered this bug in a glyph which is not used to render text.)

justvanrossum:
Raise TTLibError with the offending glyph name in the error message when a component (indirectly) references itself. Fixes #1545

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -843,7 +843,7 @@ class Glyph(object):
 				g = glyfTable[compo.glyphName]
 				try:
 					coordinates, endPts, flags = g.getCoordinates(glyfTable)
-				except RuntimeError:
+				except RecursionError:
 					raise ttLib.TTLibError("glyph '%s' contains a recursive component reference" % compo.glyphName)
 				if hasattr(compo, "firstPt"):
 					# move according to two reference points
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -844,7 +844,7 @@ class Glyph(object):
 				try:
 					coordinates, endPts, flags = g.getCoordinates(glyfTable)
 				except RuntimeError:
-					raise ttLib.TTLibError("glyph '%s' contains recursive component reference" % compo.glyphName)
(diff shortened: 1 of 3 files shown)


justvanrossum:
Does anyone see a better phrasing for the error message? I'm not 100% convinced of:

`"glyph 'xyz' contains a recursive component reference"`

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -841,7 +841,10 @@ def getCoordinates(self, glyfTable):
 			allEndPts = []
 			for compo in self.components:
 				g = glyfTable[compo.glyphName]
-				coordinates, endPts, flags = g.getCoordinates(glyfTable)
+				try:
+					coordinates, endPts, flags = g.getCoordinates(glyfTable)
+				except RuntimeError:
```
anthrotype: python3.5 has `RecursionError` built-in exception, maybe add a TODO for later as we drop py27 support.
should you do instead:

```python
except RuntimeError as e:
    if e.args[0] == "maximum recursion depth exceeded":
        raise ttLib.TTLibError(...)
    raise
```

anthrotype:
the phrasing of the error message looks good to me

Lib/fontTools/ttLib/tables/_g_l_y_f.py
justvanrossum: Or maybe: add RecursionError as an alias for RuntimeError in py23?

anthrotype: No because RuntimeError is broader than RecursionError

justvanrossum: I understand, but: the alias would only be for 2.7, and in Python 3 we'd get the "correct" exception. That means, yes, we catch a little broader on 2.7, but there'll be no todo for Python 3, and no deed to parse the exception message. For me that's 2 pluses for Py3 and one slight minus for Py2.

anthrotype: Fair enough. It’ll be a temporary hack anyway.

PR #1547, opened 2019-03-18
Build CFF2 Variable Font from sparse sources, and with more than one VarData table.

readroberts:
in varLib/cffLib.py, add support for sparse sources, and sources with more than one model, and hence more than one VarData element in the VarStore.

Support blend ops in cffLib/specializer.py generalizeCommands() and specializeCommands().


readroberts:
Note: this is a PR for a revision of an earlier branch. See https://github.com/fonttools/fonttools/pull/1475

anthrotype:
I cherry-picked the HVAR commit into master since I am about about to cut a new release, and don't feel ready to merge this one just yet (would like to take another look and also give @behdad another change to review).
https://github.com/fonttools/fonttools/commit/635537604de1d21f2d18c9f64847af0bfe313e27

readroberts:
Of course, there is certainly enough rework to need a new review. Let's keep the momentum going!

Lib/fontTools/cffLib/specializer.py
```diff
@@ -267,11 +309,62 @@ def _negateCategory(a):
...
+			for arg in blendlist:
+				blend_args.append(arg[0])
+			for arg in blendlist:
+				blend_args.extend(arg[1:])
+			blend_args.append(num_blends)
+			blend_cmds.append(('blend', blend_args))
+			blend_args = []
+			stack_use = prev_stack_use + num_blends
+
+	return blend_cmds, blend_args
```
behdad: I know spec doesn't allow it, but we should support blends where the args are also blended themselves.  Or at least be able to pass those through without error.

Lib/fontTools/cffLib/specializer.py
```diff
@@ -203,11 +213,43 @@ def rlinecurve(args):
...
+	# which is also the first default font value. There are 'num_blends'
+	# default font values, followed by 'num_blends' delta tuples.
+	blend_args = op_args[:argi]
+	first_delta_idx = tuplei = argi + num_blends
+	while argi < first_delta_idx:
+		next_ti = tuplei + numRegions
+		blendList = [op_args[argi]] + op_args[tuplei:next_ti]
+		blend_args.append(blendList)
+		tuplei = next_ti
+		argi += 1
```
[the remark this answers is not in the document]
behdad: > Is this block correct? I thought there's all default values first, then all deltas for first region, then all deltas for second region, etc. No?

Apparently no.

Lib/fontTools/cffLib/specializer.py
```diff
@@ -4,6 +4,7 @@
 
 from __future__ import print_function, division, absolute_import
 from fontTools.misc.py23 import *
+from fontTools.cffLib import maxStackLimit
```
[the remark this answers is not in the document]
readroberts: My concern here is that the default value for the specializeCommands() maxstack argument is reasonable for CFF fonts, but not for CFF2 fonts with a blend operator.  How about a check to set maxstack to cffLib.maxStackLimit if a) it is the default value of 48, and numRegions is not None?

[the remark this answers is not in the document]
behdad: Yes I like that.  If numRegions is not None, then assume CFF2.

Lib/fontTools/cffLib/specializer.py
```diff
@@ -417,12 +559,24 @@ def specializeCommands(commands,
...
+				arg0 = args[0]
+				arg1 = other_args[0]
+				if isinstance(arg0, list) and isinstance(arg1, list):
+					new_args = [[a1 + a2 for a1, a2 in zip(arg0, arg1)]]
+				elif isinstance(arg0, list):
+					new_args = _combineLineArgs(arg0, arg1)
+				elif isinstance(arg1, list):
+					new_args = _combineLineArgs(arg1, arg0)
+				else:
+					new_args = [arg0 + arg1]
```
behdad: This block should simply become `new_args = [_addArgs(args[0], other_args[0])]

Lib/fontTools/cffLib/specializer.py
```diff
@@ -267,11 +350,68 @@ def _negateCategory(a):
...
+			blend_args.append(num_blends)
+			new_args.append(blend_args)
+			stack_use = prev_stack_use + num_blends
+
+	return new_args
+
+def _combineLineArgs(listArg, valArg):
+	listArg[0] += valArg
+	newArgList = listArg
+	return newArgList
```
behdad: This function seems incomplete.  The whole `if` hierarchy below should come here, and it should recurse somethings.  This should become:
```
def _addArgs(a, b):
  if isinstance(a, list) and isinstance(b, list):
    return [_addArgs(va, vb) for va,vb in zip(a, b)]
  if isinstance(b, list):
    a, b = b, a
  if isinstance(a, list):
    return [_addArgs(a[0], b)] + a[1:]]
  return a + b
```

Lib/fontTools/cffLib/specializer.py
```diff
@@ -26,13 +27,15 @@ def programToString(program):
 	return ' '.join(str(x) for x in program)
 
 
-def programToCommands(program):
+def programToCommands(program, numRegions=None, **kwargs):
```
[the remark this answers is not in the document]
readroberts: The extra args are there because of specializeProgram(program, **kwargs) and generalizeProgram(program, **kwargs). **kwargs needs to hold both numRegions  and generalizeFirst. If these args  are both in **kwargs, then programToCommnd and specializeCommand complain about the unused arg unless you provide the additional **kwargs to swallow the unused arg. I don't really like what I did, but it is better than the alternatives I thought of.

Lib/fontTools/cffLib/specializer.py
readroberts: Yep, that is better.

Lib/fontTools/cffLib/specializer.py
readroberts: see below

Lib/fontTools/cffLib/specializer.py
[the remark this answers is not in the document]
behdad: Why not just spell out the arguments that are needed?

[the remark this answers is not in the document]
readroberts: Because I saw that the prior code used **kwargs rather than the arguments actually needed. I assumed this was a deliberate choice. I do see that it allows for adding future new arguments to the called functions, without having to change specializeProgram() and generalizeProgram(). That said, I am perfectly happy to follow your suggestion. I would then change **kwargs in specializeProgram() and generalizeProgram  to the two args, and pass only the needed args to the callees. Just confirm that this is what you would prefer.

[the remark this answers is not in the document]
behdad: I see.  That definitely explains your choice.  Thank you.

Generally I like using kwargs for methods that just pass along arguments, but not in methods that consume them.  What I like to see here, I think, is to keep things as they are, just add `numRegion` anywhere it's needed.  You can add `numRegions=None` just before `**kwargs` in the methods that take that.  This will separate that one argument from the rest, then you decide which methods you pass the numRegions to and which one the `**kwargs`.  I *think* that works but up to you whichever works.  Thanks.

Lib/fontTools/cffLib/specializer.py
readroberts: @behdad Just pushed the changes above.

readroberts:
@behdad About your question: "Shouldn't we pass a map/function that when passed the vsindex, returns numRegion?". I think this will add complexity to the logic. 'vsindex' may not ( and indeed is usually not) present in the program, so you would have to pass in both the mapping function and the default vsindex, which may be set only in the private dict, or not at all. You can see how numRegions is derived at line 945 in specializer_test.py, and line varLib/cff.py: whether or not there is a 'vsindex' in the charstring is already handled by the T2Charstring.vsindex property.

behdad:
> @behdad About your question: "Shouldn't we pass a map/function that when passed the vsindex, returns numRegion?". I think this will add complexity to the logic. 'vsindex' may not ( and indeed is usually not) present in the program, so you would have to pass in both the mapping function and the default vsindex, which may be set only in the private dict, or not at all. You can see how numRegions is derived at line 945 in specializer_test.py, and line varLib/cff.py: whether or not there is a 'vsindex' in the charstring is already handled by the T2Charstring.vsindex property.

I still think the specializer should do the right thing / be generic.

I suggest this: take numRegions as is.  When we need to use it:
* if it's None, err,
* if it's an integer, use it,
* otherwise call it with one argument.  The argument is the last vsindex seen so far (I know spec says vsindex can only occur once and at the beginning, but there's no reason NOT to write the code more generally here).  If we have not seen vsindex so far, pass None as parameter.  Use the return value.

I suggest you do this in specializer and document it in the docstring.  Whether you use it in varLib/cff is another issue and up to you.

Thanks.  Or feel free to ignore, and I do it after you land.

readroberts:
@behdad That creates an extra problem case: what if the numRegions passed in as an integer, and the program contains a vsindex operator? How about a variation on your original suggestion: the numRegions arg is always None or a function. If the function is not passed an arg, it returns the default numRegions for the charstring, else it takes vsindex as an argument, and returns the numRegion implied by the vsindex.

behdad:
> @behdad That creates an extra problem case: what if the numRegions passed in as an integer, and the program contains a vsindex operator? How about a variation on your original suggestion: the numRegions arg is always None or a function. If the function is not passed an arg, it returns the default numRegions for the charstring, else it takes vsindex as an argument, and returns the numRegion implied by the vsindex.

Sure, that's what I originally had in mind.  I thought about not passing an argument, vs passing None.  The latter is easier since you can initialize a variable to None, and update it if you see vsindex op, and pass that variable to function, as opposed to conditionalize the call to the function.

readroberts:
Sounds good. I propose adding a numRegions property in psCharstrings.py::T2CharString. This not only provides the needed function, but can reset the T2CharString current vsindex value whenever T2CharString.numRegions() is called with a not None value.  I think I will need to update T2CharString anyway - it now assumes that any vsindex op occurs only at the start of the charstring.

behdad:
> Sounds good. I propose adding a numRegions property in psCharstrings.py::T2CharString. This not only provides the needed function, but can reset the T2CharString current vsindex value whenever T2CharString.numRegions() is called with a not None value. I think I will need to update T2CharString anyway - it now assumes that any vsindex op occurs only at the start of the charstring.

I'm not sure how that works, but sure, if you think so.

readroberts:
@behdad Updated specializer with change in handling of numRegions and vsindex; please review. Other files updated to match -see commit message.

Lib/fontTools/misc/psCharStrings.py
```diff
@@ -944,6 +944,16 @@ def __init__(self, bytecode=None, program=None, private=None, globalSubrs=None):
 		self.program = program
 		self.private = private
 		self.globalSubrs = globalSubrs if globalSubrs is not None else []
+		self._cur_vsindex = None
+
+	def numRegions(self, vsindex=None):
```
behdad: Should this be renamed `getNumRegions`?

Lib/fontTools/cffLib/specializer.py
```diff
@@ -26,14 +27,20 @@ def programToString(program):
...
 
-def programToCommands(program):
+def programToCommands(program, numRegions=None):
 	"""Takes a T2CharString program list and returns list of commands.
 	Each command is a two-tuple of commandname,arg-list.  The commandname might
 	be empty string if no commandname shall be emitted (used for glyph width,
 	hintmask/cntrmask argument, as well as stray arguments at the end of the
-	program (¯\_(ツ)_/¯)."""
+	program (¯\_(ツ)_/¯).
+	'numRegions' may be None, or a function that returns the number
```
behdad: Instead of `function` say `callable`?

Lib/fontTools/cffLib/specializer.py
```diff
@@ -267,6 +360,68 @@ def _negateCategory(a):
...
+			blend_args.append(num_blends)
+			new_args.append(blend_args)
+			stack_use = prev_stack_use + num_blends
+
+	return new_args
+
+def _addArgs(a, b):
+	if isinstance(b, list):
+		if isinstance(a, list):
+			return [_addArgs(va, vb) for va,vb in zip(a, b)]
```
behdad: Should we check that the lengths match?

readroberts: @behdad Sure. This can't happen if the command list is build by programToCommands, but a developer could build one independently and  do it incorrectly. I would raise an error if the lengths don't match, catch it at line 585 and continue without changing the command. Sound OK?

behdad: Yes. Thanks.

readroberts:
@behdad Thanks, the changes you suggested yesterday were all useful. Please take a look at the latest commit, which I think implements all of them

readroberts:
Change submitted in https://github.com/fonttools/fonttools/pull/1591: built new branch and PR to allow clean commit

PR #1551, opened 2019-03-21
[varLib] Add support for building VVAR table from vmtx and VORG tables.

readroberts:
Added new function __init__.py::_add_VVAR() to build VVAR table, and shared code from _add_HVAR().
Added test case.


Lib/fontTools/varLib/__init__.py
```diff
@@ -463,46 +463,134 @@ def _merge_TTHinting(font, masterModel, master_ttfs, tolerance=0.5):
...
+	'sb1': 'LsbMap',
+	'sb2': 'RsbMap',
+	'mapping_name': 'AdvWidthMap',
+	}
+vvar_fields = { 'table_tag': 'VVAR',
+	'metrics_tag': 'vmtx',
+	'bearing1': 'TsbMap',
+	'bearing2': 'BsbMap',
+	'mapping': 'AdvHeightMap',
+	}
```
anthrotype: These two dicts hvar_fields and vvar_fields are unused and replaced by the namedtuples with the same name below, should be removed

Lib/fontTools/varLib/__init__.py
```diff
@@ -463,46 +463,134 @@ def _merge_TTHinting(font, masterModel, master_ttfs, tolerance=0.5):
...
+
+	if v_orig_metricses:
+		singleModel = False
+		for glyph in glyphOrder:
+			# We need to supply a vOrigs tuple with non-None default values
+			# for each glyph. v_orig_metricses contains values only for those
+			# glyphs which have a non-default vOrig.
+			vOrigs = [metrics[glyph] if glyph in metrics else defaultVOrig
+				for metrics, defaultVOrig in v_orig_metricses]
+			print(glyph, vOrigs)
```
anthrotype: Please remove stray print statement

Lib/fontTools/varLib/__init__.py
```diff
@@ -463,46 +463,134 @@ def _merge_TTHinting(font, masterModel, master_ttfs, tolerance=0.5):
...
+	adv_mapping = {}
 	for glyphName in glyphOrder:
-		deltas,supports = hAdvanceDeltasAndSupports[glyphName]
+		deltas, supports = vhAdvanceDeltasAndSupports[glyphName]
 		storeBuilder.setSupports(supports)
-		mapping[glyphName] = storeBuilder.storeDeltas(deltas)
+		adv_mapping[glyphName] = storeBuilder.storeDeltas(deltas)
+
+	if v_orig_metricses:
+		v_orig_mapping_i = {}
```
anthrotype: I don’t immediately understand what “_i” in the v_orig_mapping_i variable name means, please use more descriptive name

Lib/fontTools/varLib/__init__.py
```diff
@@ -463,46 +463,134 @@ def _merge_TTHinting(font, masterModel, master_ttfs, tolerance=0.5):
...
+			deltas, supports = vOrigDeltasAndSupports[glyphName]
+			storeBuilder.setSupports(supports)
+			v_orig_mapping_i[glyphName] = storeBuilder.storeDeltas(deltas)
+
 	indirectStore = storeBuilder.finish()
 	mapping2 = indirectStore.optimize()
-	mapping = [mapping2[mapping[g]] for g in glyphOrder]
-	advanceMapping = builder.buildVarIdxMap(mapping, glyphOrder)
+	adv_mapping = [mapping2[adv_mapping[g]] for g in glyphOrder]
+	advanceMapping = builder.buildVarIdxMap(adv_mapping, glyphOrder)
```
anthrotype: At some point we should really agree on a consistent variable name style convention. All this mixing camelCase and snake_case not just within the same module, even on the same line, it’s a bit distracting.

Tests/varLib/varLib_test.py
```diff
@@ -442,6 +442,20 @@ def test_varlib_build_sparse_masters_MVAR(self):
...
         assert all(tag in mvar_tags for tag in fontTools.varLib.mvar.MVAR_ENTRIES)
 
+    def test_varlib_build_VVAR_CFF2(self):
+        ds_path = self.get_test_input('TestVVAR.designspace')
+        suffix = '.otf'
+        expected_ttx_name = 'TestVVAR'
+        tables = ["VVAR"]
+
+        finder = lambda s: s.replace('.ufo', suffix)
+        varfont, model, _ = build(ds_path, finder)
```
anthrotype: You know that the varLib.build function can also build from designspace file that has sources’s filenames pointing to TTX files? I think it’d be better to check in the repo only text files instead of binary OTF files as the test data files (unless the binary file is the only means of storing a piece of data).

readroberts: All your points make sense. Will fix.

readroberts: My apologies for the mixed case style in one line - that is ugly. The standard I see in fontTools is:
- new modules should follow PEP8
- new functions added to an existing module should follow the module's predominating style. (for example, varLib/__init__.py has some functions with snake-case, and but has mostly camel-case code, and some examples of both in the same function).
- new logic added to a function should follow the function style.

readroberts:
@anthrotype Please take another look. I think I have fixed the code issues. I did not fix the test font issue, as I see that all the CFF2 tests I have added over the last several months use binary OTF files. I would prefer to open a new issue/PR to fix all of these.

Lib/fontTools/varLib/__init__.py
```diff
@@ -463,46 +463,120 @@ def _merge_TTHinting(font, masterModel, master_ttfs, tolerance=0.5):
...
+	_add_VHVAR(font, masterModel, master_ttfs, axisTags, vvarFields)
+
+def _add_VHVAR(font, masterModel, master_ttfs, axisTags, tableFields):
+
+	tableTag = tableFields.tableTag
+	assert tableTag not in font
+	log.info("Generating " + tableTag)
+	VHVAR = newTable(tableTag)
+	table_class = getattr(ot, tableTag)
+	vhvar = VHVAR.table = table_class()
```
anthrotype: `table_class` -> `tableClass` for consistency with the rest of the local vars

readroberts:
Add support for building VVAR table from vmtx and VORG tables.

PR #1553, opened 2019-03-24
DesignSpaceDocument.defaultLoc needs to be updated

LettError:
defaultLoc is used in findDefault(), but it is not updated after first read. So any axes that are added after reading the doc or creating the object are ignored.

### DIFF
--- a/Lib/fontTools/designspaceLib/__init__.py
+++ b/Lib/fontTools/designspaceLib/__init__.py
@@ -998,7 +998,7 @@ class DesignSpaceDocument(LogMixin, AsDictMixin):
         self.axes = []
         self.rules = []
         self.default = None         # name of the default master
-        self.defaultLoc = None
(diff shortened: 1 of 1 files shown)


anthrotype:
Lgtm. We should also add a test for this.

madig:
This will fail if axes are removed or switched out completely (`designspace.axes = [...]`).  Maybe get rid of `defaultLoc` and always iterate over axes in `findDefault` or make it a property that always iterates over all axes.

anthrotype:
Yeah, a property is a good idea for this.

anthrotype:
actually, no need for a new property, as `newDefaultLocation` method already does that. I simply got rid of defaultLoc attribute

anthrotype:
https://github.com/fonttools/fonttools/commit/b327287f977f45de52ce956787d4e33c441c657d

PR #1557, opened 2019-03-27
[fontBuilder] Enable making CFF2 fonts with 'post' table format 2

miguelsousa:
Before this change all OTFs wound up with 'post' table format 3

### DIFF
--- a/Tests/fontBuilder/fontBuilder_test.py
+++ b/Tests/fontBuilder/fontBuilder_test.py
@@ -135,9 +135,11 @@ def test_build_otf(tmpdir):
     charString = pen.getCharString()
     charStrings = {".notdef": charString, "A": charString, "a": charString, ".null": charString}
     fb.setupCFF(nameStrings['psName'], {"FullName": nameStrings['psName']}, charStrings, {})
+
+    lsb = {gn: cs.calcBounds(None)[0] for gn, cs in charStrings.items()}
     metrics = {}
     for gn, advanceWidth in advanceWidths.items():
-        metrics[gn] = (advanceWidth, 100)  # XXX lsb from glyph
+        metrics[gn] = (advanceWidth, lsb[gn])
     fb.setupHorizontalMetrics(metrics)
 
     fb.setupHorizontalHeader(ascent=824, descent=200)
--- a/Tests/fontBuilder/fontBuilder_test.py
+++ b/Tests/fontBuilder/fontBuilder_test.py
@@ -2,14 +2,13 @@ from __future__ import print_function, division, absolute_import
 from __future__ import unicode_literals
 
 import os
-import shutil
+import pytest
 import re
 from fontTools.ttLib import TTFont
 from fontTools.pens.ttGlyphPen import TTGlyphPen
 from fontTools.pens.t2CharStringPen import T2CharStringPen
 from fontTools.fontBuilder import FontBuilder
 from fontTools.ttLib.tables.TupleVariation import TupleVariation
-from fontTools.ttLib.tables._g_l_y_f import GlyphCoordinates
 from fontTools.misc.psCharStrings import T2CharString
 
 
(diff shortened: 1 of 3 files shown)


anthrotype:
For CFF 1.0 the post table must always be 3.0, because the postscript names are already embedded in the table, you don’t want to duplicate them in the post table as well.
But yes, for CFF2 one should be able to use either post v2 or v3.

justvanrossum:
> For CFF 1.0 the post table must always be 3.0, because the postscript names are already embedded in the table, you don’t want to duplicate them in the post table as well.

The tests in this PR contain a test for CFF with post 2.0. Does that even make sense then?

`fb.setupPost()` currently does the wrong thing for CFF in this PR, as `keepGlyphNames` default value is `True`, so this is a non-backwards-compatible change.

I want `setupPost()` to do the right thing automatically by default, so we should somehow detect cff2-ness.

Perhaps something like:

```python
if (self.isTTF or "CFF2" in self.font) and keepGlyphNames: ...`
```
Which admittedly only works if `setupCFF2()` has been called before `setupPost()`. Hmm.

anthrotype:
Hm. Maybe we could deprecate the `isTTF` boolean parameter and substitute it with a `flavor` parameter which takes a set of (enum?) constant values like "ttf", "cff", "cff2" (even "woff", "woff2"?).

anthrotype:
(actually forget about woff, those are just container formats, not OpenType flavours)

justvanrossum:
> Hm. Maybe we could deprecate the isTTF boolean parameter and substitute it with a flavor parameter which takes a set of (enum?) constant values like "ttf", "cff", "cff2".

Yeah, I'm open for something like that.

miguelsousa:
You both are absolutely right about a CFF 1.0 font with post v2 not making any sense. Got carried away there.
I've reworked the changes. Please take another look.

> Which admittedly only works if `setupCFF2()` has been called before `setupPost()`.

The documentation already prescribes that [the calls need to happen in a certain order](https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/fontBuilder.py#L12), so I think we're covered there.

> Maybe we could deprecate the isTTF boolean parameter and substitute it with a flavor parameter which takes a set of (enum?) constant values like "ttf", "cff", "cff2" (even "woff", "woff2"?).
> (actually forget about woff, those are just container formats, not OpenType flavours)

[`flavor` is already a parameter of **ttFont** which can be `None`, `woff` and `woff2`](https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/ttLib/ttFont.py#L46).

miguelsousa:
@justvanrossum @anthrotype OK to merge?

Tests/fontBuilder/fontBuilder_test.py
```diff
@@ -81,16 +84,21 @@ def test_build_ttf(tmpdir):
...
 
     fb.save(outPath)
 
     _verifyOutput(outPath)
 
 
-def test_build_otf(tmpdir):
-    outPath = os.path.join(str(tmpdir), "test.otf")
+@pytest.mark.parametrize('out_filename, keep_glyph_names', [
+    # CFF 1.0 fonts should always use post table format 3.0
```
justvanrossum: So, what we're effectively testing here is whether the `keepGlyphNames` flag is ignored in the case of CFF 1.0, correct?

miguelsousa: That's right

anthrotype:
I wish we didn't have to parametrize the fontBuilder integration tests and add so many test data files just to check if we keep or drop the glyph names. A more atomic unit test jsut for the setupPost method would be preferable IMO.

justvanrossum:
> I wish we didn't have to parametrize the fontBuilder integration tests and add so many test data files just to check if we keep or drop the glyph names. A more atomic unit test just for the setupPost method would be preferable IMO.

Good point, and I agree. Pretty much we want to ensure that `keepGlyphNames` does the right thing for TFF and CFF2, and is being ignored for CFF. That could be done without saving the font at all, and just inspect `fb.font["post"].formatType`. So maybe leave the existing tests for what they are, and add a lighter-weight parameterized test function. How does that sound?

miguelsousa:
Sounds good. I'll rework it.

justvanrossum:
> Sounds good. I'll rework it.

Thanks. Ideally, also check whether `setupPost()` called with no arguments does the right thing.

miguelsousa:
@justvanrossum @anthrotype please have another look.
With this change the default `post` format for a CFF2 made with FontBuilder changes from 3.0 to 2.0, akin to TTF.

PR #1561, opened 2019-03-28
refactor instantiateItemVariationStore for better test-ability

anthrotype:
The function now takes a VarStore instance, the fvar axes and a partial location, and returns an array of delta-sets to be applied to the default instance.

The algorithm is now more similar to the one used for instantiating the tuple variation store.

Tests are coming soon.

### DIFF
--- a/Lib/fontTools/varLib/instancer.py
+++ b/Lib/fontTools/varLib/instancer.py
@@ -17,9 +17,7 @@ from fontTools.varLib.models import supportScalar, normalizeValue, piecewiseLine
 from fontTools.varLib.iup import iup_delta
 from fontTools.ttLib import TTFont
 from fontTools.ttLib.tables._g_l_y_f import GlyphCoordinates
-from fontTools.varLib.varStore import VarStoreInstancer
 from fontTools.varLib.mvar import MVAR_ENTRIES
-import collections
 from copy import deepcopy
 import logging
 import os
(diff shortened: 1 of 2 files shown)


Lib/fontTools/varLib/instancer.py
```diff
@@ -132,132 +130,152 @@ def instantiateCvar(varfont, location):
...
 
-def setMvarDeltas(varfont, location):
+def setMvarDeltas(varfont, deltaArray):
     log.info("Setting MVAR deltas")
 
     mvar = varfont["MVAR"].table
-    fvar = varfont["fvar"]
-    varStoreInstancer = VarStoreInstancer(mvar.VarStore, fvar.axes, location)
     records = mvar.ValueRecord
-    # accumulate applicable deltas as floats and only round at the end
```
anthrotype: turns out it's actually not necessary to accumulate and round at the end in this case, because each MVAR record can only refer to one item in the VarStore at a time.

anthrotype:
oh.. I mean to open this against the googlefonts/fonttools fork.
Please ignore this, we shall send a PR to upstream fonttools/fonttools once this work is complete.
I prefer to do that since I'm working on it with Nyshadh from different offices/timezones, and I'd like to use PRs but at same time don't want to create too much noise here for a WIP branch.

behdad:
I wouldn't mind monitoring your progress.

In general I think most of us prefer that, to seeing a big PR appear from thin air.  People can unsubscribe from individual issues if they do not want to follow...

anthrotype:
Fair enough 😅

PR #1564, opened 2019-04-02
[svgLib] Support line, polyline and matrix transforms

miguelsousa:
* Adds support for [**matrix** transform function](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform#Matrix)
* Adds support for [`<polyline>` element](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/polyline)
* Adds support for [`<line>` element](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/line)


Lib/fontTools/svgLib/path/shapes.py
```diff
@@ -16,12 +19,25 @@ def _strip_xml_ns(tag):
     return tag.split('}', 1)[1] if '}' in tag else tag
 
 
+def _transform(raw_value):
+    # start simple: if you aren't exactly matrix(...) then no love
+    match = re.match(r'matrix\((.*)\)', raw_value)
+    if not match:
+        raise ValueError('unknown transform %s' % raw_value)
+    matrix = [float(p) for p in re.split(r'\s+|,', match.group(1))]
```
anthrotype: you can directly make a tuple here with a generator expression like `tuple(float(p) for p in ...` instead of list comprehension and then `tuple(matrix)` at the end

Lib/fontTools/svgLib/path/__init__.py
```diff
@@ -55,6 +55,13 @@ def draw(self, pen):
         # xpath | doesn't seem to reliable work so just walk it
         for el in self.root.iter():
             pb.add_path_from_element(el)
-        for path in pb.paths:
+        original_pen = pen
+        for path, transform in zip(pb.paths, pb.transforms):
+            # XXX assumes a 'matrix' transform.
```
anthrotype: instead "XXX" use "TODO"

Lib/fontTools/svgLib/path/__init__.py
```diff
@@ -55,6 +55,13 @@ def draw(self, pen):
...
+            # No other transform functions are supported at the moment.
+            # https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform
```
anthrotype: maybe move this inline comment to the `_transform` function in shapes module

Lib/fontTools/svgLib/path/shapes.py
```diff
@@ -16,12 +19,25 @@ def _strip_xml_ns(tag):
     return tag.split('}', 1)[1] if '}' in tag else tag
 
 
+def _transform(raw_value):
+    # start simple: if you aren't exactly matrix(...) then no love
+    match = re.match(r'matrix\((.*)\)', raw_value)
+    if not match:
+        raise ValueError('unknown transform %s' % raw_value)
```
anthrotype: you could also use NotImplementedError

miguelsousa:
@anthrotype all done, thanks.

PR #1568, opened 2019-04-05
[cffLib] Allow charset to be a collections.deque

miguelsousa:
While making an OTF with **fontBuilder** I used [`collections.deque`](https://docs.python.org/3/library/collections.html#collections.deque) (instead of a simple `list`) to assemble the font's glyph order.

This patch fixes the following error I ran into,

    TypeError: sequence index must be integer, not 'slice'

### DIFF
--- a/Lib/fontTools/cffLib/__init__.py
+++ b/Lib/fontTools/cffLib/__init__.py
@@ -1382,7 +1382,7 @@ def packCharset0(charset, isCID, strings):
 	else:
 		getNameID = getSIDfromName
 
-	for name in charset[1:]:
+	for name in list(charset)[1:]:
 		data.append(packCard16(getNameID(name, strings)))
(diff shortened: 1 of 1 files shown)


anthrotype:
Hm. Can’t you cast it to a list before you pass it to the fontBuilder?

justvanrossum:
> Hm. Can’t you cast it to a list before you pass it to the fontBuilder?

I was thinking the same. Right now it does an unnecessary copy if the value is already a list.

miguelsousa:
Sure can. This seemed like the forward looking way to fix it though.

justvanrossum:
I think it's ok to require a list here, but _if_ more general iterable support were desired, `itertools.islice()` would be the solution.

miguelsousa:
Makes sense. Thanks.

PR #1577, opened 2019-04-16
[partial-instancer] merge regions with same axes and optimize ItemVarStore

anthrotype:
instantiateItemVarStore now calls VarStore.optimize() at the end and returns the mapping from old to new VarIdx (or None if varStore has been emptied).

We also merge regions that end up with identical axes configuration after partial-instancing, so that they only occupy a single column in the delta-set rows.

### DIFF
--- a/Lib/fontTools/varLib/instancer.py
+++ b/Lib/fontTools/varLib/instancer.py
@@ -163,12 +163,17 @@ def setMvarDeltas(varfont, deltaArray):
 def instantiateMvar(varfont, location):
     log.info("Instantiating MVAR table")
 
-    varStore = varfont["MVAR"].table.VarStore
+    mvar = varfont["MVAR"].table
     fvarAxes = varfont["fvar"].axes
-    defaultDeltas = instantiateItemVariationStore(varStore, fvarAxes, location)
+    defaultDeltas, varIndexMapping = instantiateItemVariationStore(
+        mvar.VarStore, fvarAxes, location
+    )
     setMvarDeltas(varfont, defaultDeltas)
 
-    if not varStore.VarRegionList.Region:
+    if varIndexMapping:
+        for rec in mvar.ValueRecord:
+            rec.VarIdx = varIndexMapping[rec.VarIdx]
+    else:
         # Delete table if no more regions left.
         del varfont["MVAR"]
 
@@ -222,10 +227,12 @@ def _getVarDataDeltasForRegions(varData, regionIndices, rounded=False):
 
 def _subsetVarStoreRegions(varStore, regionIndices):
     # drop regions not in regionIndices
-    newVarDatas = []
     for varData in varStore.VarData:
         if regionIndices.isdisjoint(varData.VarRegionIndex):
-            # drop VarData subtable if we remove all the regions referenced by it
+            # empty VarData subtable if we remove all the regions referenced by it
+            varData.Item = [[] for _ in range(varData.ItemCount)]
+            varData.VarRegionIndex = []
+            varData.VarRegionCount = varData.NumShorts = 0
             continue
 
         # only retain delta-set columns that correspond to the given regions
@@ -237,10 +244,7 @@ def _subsetVarStoreRegions(varStore, regionIndices):
 
         # recalculate NumShorts, reordering columns as necessary
         varData.optimize()
-        newVarDatas.append(varData)
 
-    varStore.VarData = newVarDatas
-    varStore.VarDataCount = len(varStore.VarData)
     # remove unused regions from VarRegionList
     varStore.prune_regions()
 
@@ -281,7 +285,13 @@ def instantiateItemVariationStore(varStore, fvarAxes, location):
     }
     _subsetVarStoreRegions(varStore, newRegionIndices)
 
-    return defaultDeltaArray
+    if varStore.VarRegionList.Region:
+        # optimize VarStore, and get a map from old to new VarIdx after optimization
+        varIndexMapping = varStore.optimize()
+    else:
+        varIndexMapping = None  # VarStore is empty
+
+    return defaultDeltaArray, varIndexMapping
 
 
 def instantiateFeatureVariations(varfont, location):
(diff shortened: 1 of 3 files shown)


Lib/fontTools/varLib/instancer.py
```diff
@@ -150,25 +150,29 @@ def setMvarDeltas(varfont, deltaArray):
...
         itemIndex = rec.VarIdx & 0xFFFF
-        deltaRow = deltaArray[varDataIndex][itemIndex]
-        delta = sum(deltaRow)
+        delta = deltaArray[varDataIndex][itemIndex]
         if delta != 0:
             setattr(
                 varfont[tableTag],
                 itemName,
-                getattr(varfont[tableTag], itemName) + otRound(delta),
+                getattr(varfont[tableTag], itemName) + delta,
```
rsheeter: seems worth commenting why we do/do not round

Lib/fontTools/varLib/instancer.py
```diff
@@ -196,59 +200,108 @@ def _getVarRegionScalar(location, regionAxes):
...
-    varRegionCount = len(varData.VarRegionIndex)
-    scalars = [regionScalars[regionIndex] for regionIndex in varData.VarRegionIndex]
+    # multiply all varData deltas in-place by the corresponding region scalar, and
+    # remove regions whose scalar is 0
+    scalars = [regionScalars[ri] for ri in varData.VarRegionIndex]
     for item in varData.Item:
-        assert len(item) == varRegionCount
-        item[:] = [delta * scalar for delta, scalar in zip(item, scalars)]
+        item[:] = [
+            delta * scalar for delta, scalar in zip(item, scalars) if scalar != 0
```
rsheeter: do you need != 0 or is just `if scalar` sufficient?

Lib/fontTools/varLib/instancer.py
```diff
@@ -196,59 +200,108 @@ def _getVarRegionScalar(location, regionAxes):
...
 
 
-def _getVarDataDeltasForRegions(varData, regionIndices, rounded=False):
-    # Get only the deltas that correspond to the given regions (optionally, rounded).
-    # Returns: list of lists of float
+def _popVarDataDeltas(varData, regionIndex):
+    # For each item's delta-set row, pop the column that correspond to regionIndex.
+    # If the region is not referenced by the varData.VarRegionIndex, return all zeros
+    # and keep varData as is.
+    # Returns: list of deltas, one per VarData item.
```
rsheeter: general comment: some of these comments are explaining *what* we are doing (which one can generally tell from the code itself) and not *why* (which is harder to know from the code)

Tests/varLib/instancer_test.py
```diff
@@ -279,57 +299,81 @@ def test_scaleVarDataDeltas(self):
...
         )
 
-        assert instancer._getVarDataDeltasForRegions(varData, {1}) == [[33.5], [-100]]
-        assert instancer._getVarDataDeltasForRegions(varData, {0}) == [[67.9], [-200]]
-        assert instancer._getVarDataDeltasForRegions(varData, set()) == [[], []]
-        assert instancer._getVarDataDeltasForRegions(varData, {1}, rounded=True) == [
-            [34],
-            [-100],
-        ]
+        instancer._mergeVarDataRegions(varData, [0, 0, 0, 3, 4])
```
nyshadhr9: Can you add another test case where the number of regions referenced in varData is less than the actual number of regions:

varData = builder.buildVarData([3, 1, 2, 0, 4], [[12, 34, 56, 78, 0], [91, 23, 45, 67, -1]])
instancer._mergeVarDataRegions(varData, [0, 0, 0, 3, 4, **5**])

Lib/fontTools/varLib/instancer.py
```diff
@@ -196,59 +200,108 @@ def _getVarRegionScalar(location, regionAxes):
 
 
 def _scaleVarDataDeltas(varData, regionScalars):
-    # multiply all varData deltas in-place by the corresponding region scalar
-    varRegionCount = len(varData.VarRegionIndex)
-    scalars = [regionScalars[regionIndex] for regionIndex in varData.VarRegionIndex]
+    # multiply all varData deltas in-place by the corresponding region scalar, and
+    # remove regions whose scalar is 0
```
nyshadhr9: Won't there be a problem with indices if we drop the columns whose regionScalar == 0? I'm asking because the delta column gets dropped but the region still exists.  
If not can you add the explanation as a comment.

Tests/varLib/instancer_test.py
```diff
@@ -279,57 +299,81 @@ def test_scaleVarDataDeltas(self):
...
+        assert instancer._popVarDataDeltas(varData, 1) == [34, -100]
+        assert varData.VarRegionCount == 1
+
+        varData = builder.buildVarData([1, 0], [[34, 68], [-100, -200]], optimize=False)
+        assert instancer._popVarDataDeltas(varData, 0) == [68, -200]
+        assert varData.VarRegionCount == 1
+
+        varData = builder.buildVarData([1, 0], [[34, 68], [-100, -200]], optimize=False)
+        assert instancer._popVarDataDeltas(varData, 2) == [0, 0]  # missing
+        assert varData.VarRegionCount == 2
```
nyshadhr9: Can you add another test case: instancer._popVarDataDeltas(varData, None). I don't think we would want [0, 0] as the output here?

Lib/fontTools/varLib/instancer.py
anthrotype: `if scalar` works too but I like to be explicit when comparing numbers to 0

Lib/fontTools/varLib/instancer.py
anthrotype: because we have already summed and rounded them in _mergeVarDataItemColumns

Lib/fontTools/varLib/instancer.py
anthrotype: The delta column gets dropped and the associated region is no longer referenced by this varData.VarRegionIndex, however the region stays in the VarRegionList until we call `VarStore.prune_regions()` at the end.
If we modified VarRegionList, all the VarRegionIndex would be invalidated.

anthrotype:
closing in favour of #1579

PR #1580, opened 2019-04-18
[partial-instancer] support instantiating GDEF and GPOS

anthrotype:
this uses the same MutatorMerger class as is used in mutator module, slightly modified to optionally allow to keep VariationIndex tables when VarStore was only partially instanced.

### DIFF
--- a/Lib/fontTools/varLib/merger.py
+++ b/Lib/fontTools/varLib/merger.py
@@ -833,17 +833,10 @@ class MutatorMerger(AligningMerger):
 	the operation can benefit from many operations that the
 	aligning merger does."""
 
-	def __init__(self, font, location):
+	def __init__(self, font, instancer, deleteVariations=True):
 		Merger.__init__(self, font)
-		self.location = location
-
-		store = None
-		if 'GDEF' in font:
-			gdef = font['GDEF'].table
-			if gdef.Version >= 0x00010003:
-				store = gdef.VarStore
-
-		self.instancer = VarStoreInstancer(store, font['fvar'].axes, location)
+		self.instancer = instancer
+		self.deleteVariations = deleteVariations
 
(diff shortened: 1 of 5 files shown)


Lib/fontTools/varLib/mutator.py
```diff
@@ -284,7 +284,7 @@ def instantiateVariableFont(varfont, location, inplace=False, overlap=True):
 		gdef = varfont['GDEF'].table
 		instancer = VarStoreInstancer(gdef.VarStore, fvar.axes, loc)
 
-		merger = MutatorMerger(varfont, loc)
+		merger = MutatorMerger(varfont, instancer)
```
anthrotype: note how this `instancer` local variable in the line above this was being thrown away (a new VarStoreInstancer was being created inside MutatorMerger's constructor from the same gdef.VarStore and location).
I changed MutatorMerger to take an already loaded VarStoreInstancer object.
But unlike in mutator module, we have already pre-computed all the deltas at that location, so we pass them (as a dict) to the "instancer" parameter of MutatorMerger. The latter expects such an "instancer" to behave like a mapping or sequence (i.e. with a `__getitem__` method) where the keys are VarIdx integers, and the values the deltas at location.

Lib/fontTools/varLib/merger.py
```diff
@@ -900,7 +897,6 @@ def merge(merger, self, lst):
 	self.__dict__ = lst[0].__dict__.copy()
 
 	instancer = merger.instancer
-	# TODO Handle differing valueformats
```
anthrotype: this comment was a copy-paste leftover. There are no "differing valueformats" across multiple fonts since the MutatorMerger is called with a single font (see merger.mergeTables() second parameter).

PR #1582, opened 2019-04-20
(Re)calculate OS/2.usMaxContext value

### LINKED ISSUE #466: Can OS/2's usMaxContext be automatically determined during compilation?
If so, that might be a nice feature.

### LINKED ISSUE #1191: Subsetting layout features does not update OS/2 → usMaxContext
1. `pyftsubset SourceSansPro-Black.ttf  --glyphs="*" --layout-features=""`
2. `ttx -o - -t OS/2 SourceSansPro-Black.subset.ttf | grep usMaxContext`

Here, usMaxContext is 3 when it should  be zero (no layout features).

Is this even relevant?

miguelsousa:
The value is only modified in the context of **fontBuilder** or **subset**. It will NOT be modified when using **ttx**.

`maxContextCalc.py` was copied from https://github.com/googlefonts/ufo2ft/blob/master/Lib/ufo2ft/maxContextCalc.py

Fixes #1191
Fixes #466

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -7,6 +7,7 @@ from fontTools.feaLib.error import FeatureLibError
 from fontTools.feaLib.parser import Parser
 from fontTools.feaLib.ast import FeatureFile
 from fontTools.otlLib import builder as otl
+from fontTools.otlLib.maxContextCalc import maxCtxFont
 from fontTools.ttLib import newTable, getTableModule
 from fontTools.ttLib.tables import otBase, otTables
 from collections import defaultdict, OrderedDict
@@ -137,6 +138,9 @@ class Builder(object):
                 fontTable.table = table
             elif tag in self.font:
                 del self.font[tag]
+        if (any(tag in self.font for tag in ("GPOS", "GSUB")) and
+                "OS/2" in self.font):
+            self.font["OS/2"].usMaxContext = maxCtxFont(self.font)
         if "GDEF" in tables:
             gdef = self.buildGDEF()
             if gdef:
--- a/Lib/fontTools/otlLib/maxContextCalc.py
+++ b/Lib/fontTools/otlLib/maxContextCalc.py
@@ -1,0 +1,101 @@
+from __future__ import print_function, division, absolute_import, unicode_literals
+
+__all__ = ['maxCtxFont']
+
+
+def maxCtxFont(font):
+    """Calculate the usMaxContext value for an entire font."""
+
+    maxCtx = 0
+    for tag in ('GSUB', 'GPOS'):
+        if tag not in font:
+            continue
+        table = font[tag].table
+        if not table.LookupList:
+            continue
+        for lookup in table.LookupList.Lookup:
+            for st in lookup.SubTable:
+                maxCtx = maxCtxSubtable(maxCtx, tag, lookup.LookupType, st)
+    return maxCtx
+
+
+def maxCtxSubtable(maxCtx, tag, lookupType, st):
+    """Calculate usMaxContext based on a single lookup table (and an existing
+    max value).
+    """
+
+    # single positioning, single / multiple substitution
+    if (tag == 'GPOS' and lookupType == 1) or (
+        tag == 'GSUB' and lookupType in (1, 2, 3)):
+        maxCtx = max(maxCtx, 1)
+
+    # pair positioning
+    elif tag == 'GPOS' and lookupType == 2:
+        maxCtx = max(maxCtx, 2)
+
+    # ligatures
+    elif tag == 'GSUB' and lookupType == 4:
+        for ligatures in st.ligatures.values():
+            for ligature in ligatures:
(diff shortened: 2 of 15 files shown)


anthrotype:
Hm, I don’t know if it’s a good idea to always recalculate that field upon compiling OS/2. Maybe it would make more sense to do that (optionally) in fontBuilder or in the subsetter?
For a font that has already the correct value, recalculating that field may waste a few cycles every time font is recompiled. 
Whereas for a font that has it set incorrectly, recompiling always fixes it even if the user didn’t ask for a fix.
In any case I’m ok with including this code here.

anthrotype:
this could also be the chance to add some unit tests for this module, if you feel like working on that.
E.g. one could use feaLib addOpenTypeFeaturesFromString on an empty TTFont (it just needs a glyphOrder set) to build each type of lookups, the call maxCtxFont on it and verifying the value is set accordingly
In OS2.

anthrotype:
I think the place where this sort of calculation can be done by default is when building the open type features inside feaLib.builder

anthrotype:
In which case, maybe the module should be moved to fontTools.otlLib package

anthrotype:
I have always wondered whether that OS/2.usMaxContext is ever actually used anywhere.

The spec says: 

> This field could be useful to sophisticated line-breaking engines in determining how far they should look ahead to test whether something could change that effects the line breaking.

Anybody knows if or where this is implemented?

behdad:
> Anybody knows if or where this is implemented?

This field was designed under wrong assumptions, and as far as I know is unused.  It cannot be correctly used in any way that I can imagine.  See:

  https://github.com/harfbuzz/harfbuzz/issues/1463#issuecomment-445431431

That said, I think recalculating this is no worse that recalculating all-glyphs bounding box, so I have no strong opinion here.

miguelsousa:
Doing the recalculation only for subset and fontBuilder sounds fine to me. I can also add tests for the new file. @anthrotype what should be the next steps?

anthrotype:
1) moving module to fontTools.otlLib
2) call it from feaLib.builder after GSUB/GPOS have been built (this will covert fontBuilder too); I don't think we need an option to disable this, do it all the time.
3) call it from subset module (with option to --no-recalc-max-context)

miguelsousa:
@anthrotype please have another look and let me know if I should add tests for the new file.

Lib/fontTools/feaLib/builder.py
```diff
@@ -137,6 +138,9 @@ def build(self, tables=None):
                 fontTable.table = table
             elif tag in self.font:
                 del self.font[tag]
+        if any([tag in self.font for tag in ("GPOS", "GSUB")]) \
```
anthrotype:
```suggestion
        if any(tag in self.font for tag in ("GPOS", "GSUB")) \
```

anthrotype: pls use parentheses instead of backslash to split multi-line statements

Lib/fontTools/otlLib/maxContextCalc.py
```diff
@@ -0,0 +1,101 @@
...
+
+def maxCtxFont(font):
+    """Calculate the usMaxContext value for an entire font."""
+
+    maxCtx = 0
+    for tag in ('GSUB', 'GPOS'):
+        if tag not in font:
+            continue
+        table = font[tag].table
+        if getattr(table, 'LookupList', None) is None:
```
anthrotype:
```suggestion
        if not table.LookupList:
```

the attribute is guaranteed to be there (as of current master)

Lib/fontTools/otlLib/maxContextCalc.py
```diff
@@ -0,0 +1,101 @@
...
+
+    # pair positioning
+    elif tag == 'GPOS' and lookupType == 2:
+        maxCtx = max(maxCtx, 2)
+
+    # ligatures
+    elif tag == 'GSUB' and lookupType == 4:
+        for ligatures in st.ligatures.values():
+            for ligature in ligatures:
+                maxCtx = max(maxCtx, getattr(ligature, 'CompCount', 0))
```
anthrotype:
```suggestion
                maxCtx = max(maxCtx, len(ligature.Component))
```

anthrotype:
> let me know if I should add tests for the new file.

ideally yes. Current coverage of maxContextCalc.py module is only 53%

Lib/fontTools/otlLib/maxContextCalc.py
miguelsousa: sorry, copied the wrong file. Thanks for catching it

miguelsousa:
@anthrotype PTAL. The coverage went up but not by much. I've ran out of ideas on how to increase it with feaLib. Perhaps some of the conditional branches require hand-made TTX files. One thing I'm suspicious about is not getting the extension lookups covered despite using `useExtension`.

miguelsousa:
using TTX files I was able to squeeze out a bit more coverage

PR #1583, opened 2019-04-20
[instancer] support instantiating HVAR and VVAR  (TTF only for now)

anthrotype:
We don't actually apply deltas to hmtx/vmtx from HVAR/VVAR since these have already been applied from the gvar deltas when we call `glyf.setCoordinates` method using the glyph's four phantom points.

We simply call `instantiateItemVariationStore` on `{H,V}VAR.VarStore` in order to remove regions and scale remaining deltas, but we ignore the return value.

Also, we run `VarStore.optimize()` (which modifies the VarDatas layout) only when the HVAR/VVAR originally have an AdvWidthMap/AdvHeightMap.
If they do not (they use a direct implicit GID->VariationIndex mapping for advance deltas), then we keep the VariationIndex unchanged by not optimizing the VarStore.

If all the axes in fvar are being instanced, then we simply delete HVAR (just like varLib.mutator currently does).

Finally, supporting CFF2 would need more work, in that HVAR is required and we need to apply the deltas to hmtx in here. Let's do that after we add support for partial instancing CFF2 charstrings.

I'll add some tests later.

### DIFF
--- a/Lib/fontTools/varLib/instancer.py
+++ b/Lib/fontTools/varLib/instancer.py
@@ -17,6 +17,7 @@ from fontTools.varLib.models import supportScalar, normalizeValue, piecewiseLine
 from fontTools.ttLib import TTFont
 from fontTools.ttLib.tables.TupleVariation import TupleVariation
 from fontTools.ttLib.tables._g_l_y_f import GlyphCoordinates
+from fontTools import varLib
 from fontTools.varLib import builder
 from fontTools.varLib.mvar import MVAR_ENTRIES
 from fontTools.varLib.merger import MutatorMerger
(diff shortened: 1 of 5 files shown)


Lib/fontTools/varLib/instancer.py
```diff
@@ -158,24 +159,71 @@ def setMvarDeltas(varfont, deltas):
...
+        log.info("Dropping %s table", tableTag)
+        del varfont[tableTag]
+        return
+
+    log.info("Instantiating %s table", tableTag)
+    vhvar = varfont[tableTag].table
+    varStore = vhvar.VarStore
+    # since deltas were already applied, the return value here is ignored
+    instantiateItemVariationStore(varStore, fvarAxes, location)
+
```
nyshadhr9: Can you add a comment here explaining what the below code is trying to do.

Lib/fontTools/varLib/instancer.py
```diff
@@ -158,24 +159,71 @@ def setMvarDeltas(varfont, deltas):
...
+    if varStore.VarRegionList.Region:
+        if getattr(vhvar, tableFields.advMapping):
+            varIndexMapping = varStore.optimize()
+            glyphOrder = varfont.getGlyphOrder()
+            _remapVarIdxMap(vhvar, tableFields.advMapping, varIndexMapping, glyphOrder)
+            if getattr(vhvar, tableFields.sb1):
```
nyshadhr9: Comment explaining what sb1 stands for in case of HVAR/VVAR will be helpful.

Lib/fontTools/varLib/instancer.py
```diff
@@ -158,24 +159,71 @@ def setMvarDeltas(varfont, deltas):
...
+                _remapVarIdxMap(vhvar, tableFields.sb1, varIndexMapping, glyphOrder)
+            if getattr(vhvar, tableFields.sb2):
+                _remapVarIdxMap(vhvar, tableFields.sb2, varIndexMapping, glyphOrder)
+            if tableFields.vOrigMapping and getattr(vhvar, tableFields.vOrigMapping):
```
nyshadhr9: If the first check is to verify the table is VVAR, can we directly use tableTag == 'VVAR' instead, just to make the code easier to read.

Issue #1587, opened 2019-04-23
[subset] assert if AdvWidthMap None, LsbMap & RsbMap non-None


blueshade7:
HVAR subsetter has the following line of assert in `subset/__init__.py` expecting if AdvWidthMap is missing, then LsbMap and RsbMap must be also both missing.
```
assert table.LsbMap is None and table.RsbMap is None, "File a bug."
```
OpenType spec does not appear to state such a requirement.
The attached test font triggering the assert passes both ots-sanitize and ttx.

VVAR subsetter also has a similar assert.
[TestHVVAR.ttf.zip](https://github.com/fonttools/fonttools/files/3109489/TestHVVAR.ttf.zip)

behdad:
Right. Just not implemented.

blueshade7:
Let me take care on this. My HVAR/VVAR fix for #1446 appears to fix this as well.

Issue #1448, opened 2019-01-15
[subset] support subsetting glyphs from CFF2 fonts


anthrotype:
Right now, if one passes an OpenType font containing CFF2 table, the subsetter chokes like this

```
$ fonttools subset /home/clupo/Downloads/SourceSerifVariable-Roman.otf A B C
Traceback (most recent call last):
  File "/home/clupo/.pyenv/versions/fontmake-py37/bin/fonttools", line 11, in <module>
    load_entry_point('fonttools', 'console_scripts', 'fonttools')()
  File "/home/clupo/Github/fonttools/Lib/fontTools/__main__.py", line 29, in main
    runpy.run_module(mod, run_name='__main__')
  File "/home/clupo/.pyenv/versions/3.7.0/lib/python3.7/runpy.py", line 208, in run_module
    return _run_code(code, {}, init_globals, run_name, mod_spec)
  File "/home/clupo/.pyenv/versions/3.7.0/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/clupo/Github/fonttools/Lib/fontTools/subset/__main__.py", line 7, in <module>
    sys.exit(main())
  File "/home/clupo/Github/fonttools/Lib/fontTools/misc/loggingTools.py", line 375, in wrapper
    return func(*args, **kwds)
  File "/home/clupo/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2824, in main
    subsetter.subset(font)
  File "/home/clupo/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2625, in subset
    self._prune_pre_subset(font)
  File "/home/clupo/Github/fonttools/Lib/fontTools/subset/__init__.py", line 2433, in _prune_pre_subset
    retain = table.prune_pre_subset(font, self.options)
  File "/home/clupo/Github/fonttools/Lib/fontTools/subset/cff.py", line 80, in prune_pre_subset
    private = font.FDArray[fdSelectIndex].Private
  File "/home/clupo/Github/fonttools/Lib/fontTools/cffLib/__init__.py", line 585, in __getitem__
    item = self.items[index]
TypeError: list indices must be integers or slices, not NoneType
```

anthrotype:
[...]

PR #1588, opened 2019-04-24
[subset] HVAR/VVAR with --retain-gids

blueshade7:
Fixes subsetting HVAR/VVAR issues #1448 & #1587

When AdvWidth(Height)Map is missing from HVAR/VVAR, unused deltas are zeroed (with or without --retain-gids).
With --retain-gids, major/minor indices for an unused glyph is set to 0/0 if gid <= last used gid
With --retain-gids, if unused gid > last used gid, then its major/minor is set to that of last used gid (they are optimized out when the table is written out).

Added test cases to test_subset.py for HVAR & VVAR with and without --retain-gids

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -1803,15 +1803,17 @@ def _remap_index_map(s, varidx_map, table_map):
 def subset_glyphs(self, s):
 	table = self.table
 
-	# TODO Update for retain_gids
-
 	used = set()
+	advIdxes_ = set()
+	retainAdvMap = False
 
 	if table.AdvWidthMap:
 		table.AdvWidthMap.mapping = _dict_subset(table.AdvWidthMap.mapping, s.glyphs)
 		used.update(table.AdvWidthMap.mapping.values())
 	else:
 		used.update(s.reverseOrigGlyphMap.values())
+		advIdxes_ = used.copy()
+		retainAdvMap = s.options.retain_gids
 
 	if table.LsbMap:
 		table.LsbMap.mapping = _dict_subset(table.LsbMap.mapping, s.glyphs)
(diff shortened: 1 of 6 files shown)


Lib/fontTools/subset/__init__.py
```diff
@@ -1789,6 +1789,13 @@ def subset_glyphs(self, s):
 	self.glyphCount = len(self.variations)
 	return bool(self.variations)
 
+def _remap_index_map(s, varidx_map, table_map):
+	map = {k:varidx_map[v] for k,v in table_map.mapping.items()}
+	last_idx = varidx_map[table_map.mapping[s.last_retained_glyph]]
+	for g,i in s.reverseEmptiedGlyphMap.items():
+		map[g] = last_idx if i > s.last_retained_order else 0
```
anthrotype: you could add a comment here explaining why the varIdx for emptied glyphs are set to either 0/0 or to the varIdx of the last retained glyph (e.g. mention the fact they are optimized out as table is compiled).

anthrotype: `map` is a built-in fuction name; most syntax highlighter can't distinguish between builtins and local variables that shadow the latter (as you see it appears in blue). It's distracting when reading at code. Conventionally one would add an underscore to the name e.g. `map_` or use a different name altogether

Lib/fontTools/varLib/varStore.py
```diff
@@ -217,10 +217,16 @@ def VarStore_subset_varidxes(self, varIdxes, optimize=True):
...
 		items = data.Item
 		newItems = []
-		for minor in sorted(usedMinors):
-			newMinor = len(newItems)
-			newItems.append(items[minor])
-			varDataMap[(major<<16)+minor] = (newMajor<<16)+newMinor
+		if major == 0 and retainFirstMap:
+			zeros = [0] * len(items[0])
+			for minor in range(len(items)):
+				newItems.append(items[minor] if minor in usedMinors else zeros)
```
anthrotype: it would be better to append a copy of that `zeros` list instead of a reference to the original (mutable) object. The expectation is that each VarData item is a distinct list object

Lib/fontTools/subset/__init__.py
blueshade7: added comment & renamed map to map_

Lib/fontTools/varLib/varStore.py
blueshade7: inlined zero array creation

blueshade7:
I realized this PR produces wrong VarData[0] without retain-gids for no AdvWidthMap.
In test data expect_HVVAR.ttx, HVAR.VarData[0] must be compacted but it isn’t (expected result is wrong as well as code).
Let me refix.

blueshade7:
With no retain-gids, no AdvWidthMap, VarData[0] must be compacted.
Things get complicated if LsbMap / RsbMap reference deltas in VarData[0]:

If deltas used as width of a retained glyph, then use its (new) index
If deltas not used as width, then the delta entry must be moved to the end of VarData[0] (after all deltas for widths) then use its index

How’s that?

blueshade7:
updated code & test result as outlined above

PR #1591, opened 2019-04-26
Sparse cff2vf support

readroberts:
See history on PR #1547.
Built a new branch to allow a clean merge, without overwriting any of the main branch fixes  in the many weeks since PR #1547 was opened.

In varLib/cffLib.py, add support for sparse sources, and sources with more than one model, and hence more than one VarData element in the VarStore.

CFF2 source fonts with multiple FontDicts in the FDArray needed some extra work. With sparse fonts, some of the source fonts may have a fewer FontDicts than the default font. The getfd_map function() builds a map from the FontDict indices in the default font to those in each region font. This is needed when building up the blend value lists in the master font FontDict PrivateDicts, in order to fetch PrivateDict values from the correct FontDict in each region font.

In specializer.py, add support for CFF2 CharStrings with blend operators.

Add test case for sparse CFF2 sources.
The test font has 55 glyphs. 2 glyphs use only 2 sources (weight = 0 and 100). The rest use 4 source fonts: the two end points of the weight axis, and two intermediate masters. The intermediate masters are only 1 design space unit apart, and are used to change glyph design at the point in design space. For the rest, at most 2 glyphs use the same set of source fonts. There are 12 source fonts.

Add test case for specializer programToCommands() and commandsToProgram by converting each CharString.program in the font to a command list, and back again, and comparing original and final versions.


readroberts:
See history in PR https://github.com/fonttools/fonttools/pull/1547

Issue #1590, opened 2019-04-26
[subset] AssertionError when using --desubroutinize


khaledhosny:
Running the following command on the attached font, raises an `AssertionError` exception:
```
$ pyftsubset --glyphs='*' --desubroutinize LibertinusSerif-Regular.subset.otf 
Traceback (most recent call last):
  File "/bin/pyftsubset", line 10, in <module>
    sys.exit(main())
  File "/lib/python3.7/site-packages/fontTools/misc/loggingTools.py", line 375, in wrapper
    return func(*args, **kwds)
  File "/lib/python3.7/site-packages/fontTools/subset/__init__.py", line 2834, in main
    subsetter.subset(font)
  File "/lib/python3.7/site-packages/fontTools/subset/__init__.py", line 2638, in subset
    self._prune_post_subset(font)
  File "/lib/python3.7/site-packages/fontTools/subset/__init__.py", line 2621, in _prune_post_subset
    retain = table.prune_post_subset(font, self.options)
  File "/lib/python3.7/site-packages/fontTools/subset/cff.py", line 457, in prune_post_subset
    self.desubroutinize()
  File "/lib/python3.7/site-packages/fontTools/subset/cff.py", line 484, in desubroutinize
    decompiler.execute(c)
  File "/lib/python3.7/site-packages/fontTools/subset/cff.py", line 390, in execute
    psCharStrings.SimpleT2Decompiler.execute(self, charString)
  File "/lib/python3.7/site-packages/fontTools/misc/psCharStrings.py", line 299, in execute
    rv = handler(index)
  File "/lib/python3.7/site-packages/fontTools/misc/psCharStrings.py", line 357, in op_hintmask
    hintMaskBytes, index = self.callingStack[-1].getBytes(index, self.hintMaskBytes)
  File "/lib/python3.7/site-packages/fontTools/misc/psCharStrings.py", line 1061, in getBytes
    assert len(bytes) == nBytes
AssertionError
```

Not sure if this is a font or subsetter bug, but it happens only when `--desubroutinize` is used.

[LibertinusSerif-Regular.subset.zip](https://github.com/fonttools/fonttools/files/3121374/LibertinusSerif-Regular.subset.zip)

behdad:
Maybe @blueshade7 can take a look?

blueshade7:
This line of code in `_DesubroutinizingT2Decompiler` looks suspicious:
```
[...]

PR #1593, opened 2019-04-30
Fix bug in subsetting T2 Charstrings

readroberts:
Fixes issue '[subset] AssertionError when using --desubroutinize #1590'


readroberts:
My apologies - I see I branched off another branch rather than master, hence al the commits. Will re-do

PR #1594, opened 2019-04-30
Fix bug in subsetting T2 CharStrings from CFF font, issue 1590

readroberts:
Fixes issue '[subset] AssertionError when using --desubroutinize #1590'


anthrotype:
Thanks. Would be great if you could also add a test, maybe using the font that @khaledhosny uploaded to reproduce the issue in https://github.com/fonttools/fonttools/issues/1590

anthrotype:
Thanks

readroberts:
I'll add a test when I am back work on Tues.

anthrotype:
Khaled already added one https://github.com/fonttools/fonttools/pull/1594/commits/b1999a1a0e7d7fb232e496eff5c6215d8b6aaebd

PR #1595, opened 2019-04-30
Convert OTF font files in test data to ttx files.

readroberts:
Convert OTF font files in test data to ttx files.

I did not touch the aots files, nor Tests/ttx/data/TestOTF.otf. The former can't all be round tripped with ttx, and the latter exists specifically to test going from binary to ttx XML file.

This required a fix to fontTools/cffLib. When reading a CFF2 variable font (VF) from XML, the VF state in FontDict and PrivateDict gets set correctly only if the VarStore element is read in before the FDArray. Fixed by adding code to force the VarStore element to be read in before the FDArray.

When debugging, the inconsistency in how the VF state in FontDict and PrivateDict was set bothered me, so I converted these in getters and setters. This required minor touches to other files.


readroberts:
@anthrotype I agree with your comments. I have also spent some time looking at how the isCFF2 attribute is used, and am disappointed to see that simply calling cff.compile() or decompile() with isCFF2=True is not enough to make the right thing happen. I think this is pretty easy to fix, along with the issue of propagating vstore that you mention above. Also, I can think I can get rid of varLib.cff.py::lib_convertCFFToCFF2() altogether, as much of the logic in it is no longer necessary. However, this is a larger change than this PR is aimed at. I suggest that I cancel this PR, resubmit a new one with a just the font data and the current change to TopDict.fromXML to fix up vstore after the fact. I will then re-do the compile/decompile/fromXML tree so that isCFF2 is not needed as a persistent or public field, but as new PR, and a single task.

anthrotype:
Ok

readroberts:
Closing per discussion with Cosimo.

PR #1596, opened 2019-05-01
[instancer] partially instantiate avar, STAT and fvar

anthrotype:


behdad:
Technically STAT can exist in fully-instantiated font as well.

anthrotype:
> Technically STAT can exist in fully-instantiated font as well.

I know (see commit msg).. I was lazy, but I added a TODO

anthrotype:
ok, I also implemented partial STAT instancing. The whole table is only dropped when all its DesignAxis records have been instanced.

It's actually hard to find a real-world font with complete STAT table (i.e. containing non-empty AxisValue records) given varLib doesn't build them currently.

I'll have to make one up for the sake of testing this code. Tomorrow.

anthrotype:
@nyshadhr9 I had forgotten to also re-map the AxisIndex of the AxisValue records that weren't dropped to point to the new DesignAxis indices. I also added a STAT table with non-empty AxisValueArray to the test font (PartialInstancer-VF.ttx) to exercise this code.

anthrotype:
about STAT partial instancing, I'm thinking.. maybe instead of removing all the DesignAxes (and corresponding AxisValues) for the (fvar) axes that are being instanced, the partial instancer should instead keep all the STAT DesignAxes, and only drop the AxisValues corresponding to the NamedInstances that were removed from fvar?

PR #1597, opened 2019-05-01
Fix unmasked bug from previous typo fix

mhosken:
OOps. sorry. This one is even tested. Sigh.


anthrotype:
no worries, I should have checked as well before merging https://github.com/fonttools/fonttools/pull/1586
but if you say it's tested, why didn't the tests fail?

mhosken:
It's tested against a production font. Adding a full test would be a pain to do to trim the thing down to something manageable for pytest.

PR #1598, opened 2019-05-01
Convert OTF font files in test data to ttx files.

readroberts:
This required a fix to fontTools/cffLib. When reading a CFF2 variable font (VF) from XML, the VF state in FontDict and PrivateDict does not get set. I made a temporary fix by adding a loop to set PrivateDict.vstore for all the PrivateDict objects after the XML file has been read. This should not be necessary, and in the near future I will revisit both this issue, and the related use of isCFF2 when compiling/decompiling.


Tests/cffLib/specializer_test.py
```diff
@@ -918,28 +919,21 @@ def test_maxstack(self):
...
-            self.assertRaisesRegex = self.assertRaisesRegexp
+class CFF2VFTestSpecialize(DataFilesHandler):
 
     @staticmethod
     def get_test_input(test_file_or_folder):
         path, _ = os.path.split(__file__)
         return os.path.join(path, "data", test_file_or_folder)
 
     def test_blend_round_trip(self):
-        otfvf_path = self.get_test_input('TestSparseCFF2VF.otf')
```
anthrotype: this `get_test_input` is method is no longer used in this class so you may well remove it

Tests/varLib/varLib_test.py
```diff
@@ -229,37 +229,51 @@ def test_varlib_gvar_explicit_delta(self):
...
+        expected_ttx_path = self.get_test_output("BuildTestCFF2.ttx")
 
-        finder = lambda s: s.replace('.ufo', suffix)
-        varfont, model, _ = build(ds_path, finder)
-        # some data (e.g. counts printed in TTX inline comments) is only
-        # calculated at compile time, so before we can compare the TTX
-        # dumps we need to save to a temporary stream, and realod the font
+        self.temp_dir()
+        for path in self.get_file_list(ttx_dir, '.ttx', 'TestCFF2_'):
+            print(path)
```
anthrotype: not sure if this print(path) was intentionally left here. pytest would capture stdout by default, only shows it when a test fails.

PR #1603, opened 2019-05-08
instancer: prune unused name records and set glyf 'overlap' flags

anthrotype:


anthrotype:
> Would --no-overlap-flag be better? This seems confusable with a potential instancer option to really remove overlaps.

makes sense, I change the wording. 👍

Lorp:
Maybe a very minor point, but do you think the option needs a “don’t change”? (We have “set to 0” and “set to 1”.) I can imagine a reasonable variable font build workflow that sets the flag only if there is any designspace location where overlap happens. How about this?

* --overlap-flag=0
* --overlap-flag=1 (default)
* --overlap-flag=-1 (don’t change)

anthrotype:
but the flag only makes a difference for static (non variable) instances, no? For variable font, the rendering engine should support overlaps hence must always assume even-odd fill rule.

anthrotype:
--no-overlap-flag option already does what you want: i.e. don't change.

PR #1605, opened 2019-05-09
Fix: ttLib _h_e_a_d FontBBox rounding

### LINKED ISSUE #1604: Float yMin value: required argument is not an integer
If a font file has a float value in `yMin`—and I assume equally in `xMin`, `xMax` or `yMax`—it will fail to save with the error `required argument is not an integer` ([`fontTools/misc/sstruct.py in pack at line 75`](https://github.com/fonttools/fonttools/blob/3.40.0/Lib/fontTools/misc/sstruct.py#L75), fonttools v3.40.0).

Trace:
```
fontTools/misc/sstruct.py in pack at line 75
fontTools/ttLib/tables/_h_e_a_d.py in compile at line 69
fontTools/ttLib/ttFont.py in getTableData at line 651
fontTools/ttLib/ttFont.py in _writeTable at line 633
fontTools/ttLib/ttFont.py in _save at line 212
fontTools/ttLib/ttFont.py in save at line 173
```

Variables at point of error:
```python
formatstring = ">llIIHHQQhhhhHHhhh"
elements = [
    65536, 
    65601, 
    1208942685, 
    1594834165, 
    3, 
    1000, 
    3551183604, 
    3640213847, 
    -132, 
    -170.009, 
    788, 
    835, 
    0, 
    3, 
    2, 
    0, 
    0
]
```

As you can see the value `-170.009` would trigger the error. If integers are expected then rounding should probably be applied.

editkid:
Ensures that CFF FontBBox float values are rounded before being read into `_h_e_a_d` table.

Note: I had a look in regards to adding a unit test for this fix, but couldn't find any existing tests for class `table__h_e_a_d`. Do let me know if I missed an obvious place to add a test.

Resolves #1604

### DIFF
--- a/Lib/fontTools/ttLib/tables/_h_e_a_d.py
+++ b/Lib/fontTools/ttLib/tables/_h_e_a_d.py
@@ -4,6 +4,7 @@ from fontTools.misc import sstruct
 from fontTools.misc.textTools import safeEval, num2binary, binary2num
 from fontTools.misc.timeTools import timestampFromString, timestampToString, timestampNow
 from fontTools.misc.timeTools import epoch_diff as mac_epoch_diff # For backward compat
+from fontTools.misc.arrayTools import intRect
 from . import DefaultTable
 import logging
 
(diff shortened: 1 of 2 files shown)


anthrotype:
Thanks!

editkid:
Thank you!

PR #1606, opened 2019-05-10
designspaceLib: add loadSourceFonts method with custom opener

anthrotype:
Added `loadSourceFonts` method to `DesignSpaceDocument` object, which allows to use a custom opener function to load a font object from a SourceDescriptor's `path`, and store the result in memory inside the SourceDescriptor's `font` attribute.

The opener function is any callable that takes a path and returns a font -- _any_ font really.

In ufo2ft we assume source descriptor `font` holds a UFO (defcon.Font or similar), whereas in varLib.build we assume it is set to a ttLib.TTFont.

### DIFF
--- a/Lib/fontTools/designspaceLib/__init__.py
+++ b/Lib/fontTools/designspaceLib/__init__.py
@@ -1272,6 +1272,15 @@ class DesignSpaceDocument(LogMixin, AsDictMixin):
         If the font attribute is already not None, it is not loaded again.
         Fonts with the same path are only loaded once and shared among SourceDescriptors.
 
+        For example, to load UFO sources using defcon:
+
+            designspace = DesignSpaceDocument.fromfile("path/to/my.designspace")
+            designspace.loadSourceFonts(defcon.Font)
+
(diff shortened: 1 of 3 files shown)


Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -1262,3 +1262,41 @@ def normalize(self):
...
             rule.conditionSets = newConditionSets
+
+    def loadSourceFonts(self, opener, **kwargs):
+        """Ensure SourceDescriptor.font attributes are loaded, and return list of fonts.
+
+        Takes a callable which initializes a new font object (e.g. TTFont, or
+        defcon.Font, etc.) from the SourceDescriptor.path, and sets the
+        SourceDescriptor.font attribute.
+        If the font attribute is already not None, it is not loaded again.
+        Fonts with the same path are only loaded once and shared among SourceDescriptors.
```
madig: Maybe add an example:
```Python
designspace.loadSourceFonts(defcon.Font)
designspace.loadSourceFonts(ufoLib2.Font.open, lazy=False)
```

PR #1609, opened 2019-05-13
varLib: keep CFF2 glyph names in post format2 table

anthrotype:
OpenType CFF fonts normally have post table format 3 with no glyph names, because the names themselves are stored already in the CFF 1.0 table.
When varLib builds a CFF2 variable font, it merges the master's CFF tables into a new CFF2 table but it throws away all the glyph names, because the post table format of the base master font (from which the varfont is built) also has post format 3. Hence we lost all the glyph names when building a CFF2 variable font with varlib.

I think the default behavior should be that varLib.build downgrades the varfont's post table to format 2, so that we keep the original glyph names.
If a user is not interested in them, they can always drop them later, but it is not varLib's job to do that.


anthrotype:
(tests will fail since the test fonts in the integration tests assume post table isn't changed hence on decompile they get auto-assigned glyph names generated from the cmap)

anthrotype:
I'll merge #1598 first, then rebase this one and fixup the failing tests

PR #1610, opened 2019-05-15
instancer: add IDEF to fpgm for GETVARIATION opcode

anthrotype:
as required by https://docs.microsoft.com/en-us/typography/opentype/spec/tt_instructions#get-variation
to support legacy environment that do not know about the GETVARIATION[] (0x91) opcode.

mostly copied from the mutator module.

### DIFF
--- a/Lib/fontTools/varLib/instancer.py
+++ b/Lib/fontTools/varLib/instancer.py
@@ -669,6 +669,10 @@ def addGetVariationIdef(font, coordinates):
     asm.append("ENDF[ ]")
     fpgm.program.fromAssembly(asm)
 
+    maxp = font["maxp"]
+    maxp.maxInstructionDefs += 1
+    maxp.maxStackElements = max(len(coordinates), maxp.maxStackElements)
+
 
 def normalize(value, triple, avar_mapping):
     value = normalizeValue(value, triple)
--- a/Lib/fontTools/varLib/instancer.py
+++ b/Lib/fontTools/varLib/instancer.py
@@ -12,11 +12,12 @@ NOTE: The module is experimental and both the API and the CLI *will* change.
 """
 from __future__ import print_function, division, absolute_import
 from fontTools.misc.py23 import *
-from fontTools.misc.fixedTools import floatToFixedToFloat, otRound
+from fontTools.misc.fixedTools import floatToFixedToFloat, otRound, floatToFixed
 from fontTools.varLib.models import supportScalar, normalizeValue, piecewiseLinearMap
-from fontTools.ttLib import TTFont
+from fontTools.ttLib import TTFont, newTable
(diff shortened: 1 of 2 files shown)


behdad:
Actually, even if it's not a full instance, you will need to interject an opcode and update all hinting code to use it, such that the custom opcode returns the values that the hinting code expects, which are the values that the original font did.

anthrotype:
oh right...
So, if I understood correctly, for a partial instance (still a varfont), we would have to define a custom IDEF instruction (an arbitrary opcode, not 0x91) that would, in turn, call GETVARIATION[] to obtain the coordinates of the current instance (that are only available at runtime), and push to the stack these _as well as_ the coordinates of the axes that were instanced, in the same order as they were previously.
And we would have to update the glyph programs that use GETVARIATION to instead use the custom opcode.

I haven't seen any font in the wild that actually uses this GETVARIATION, nor I am aware of a hinting tool that actually makes use of it. Last time I checked VTT did not produce this.
I'm inclined to post-pone this for when we actually find a need for this.

/cc @laerm0 @moyogo

behdad:
> So, if I understood correctly, for a partial instance (still a varfont), we would have to define a custom IDEF instruction (an arbitrary opcode, not 0x91) that would, in turn, call GETVARIATION[] to obtain the coordinates of the current instance (that are only available at runtime), and push to the stack these _as well as_ the coordinates of the axes that were instanced, in the same order as they were previously.
> And we would have to update the glyph programs that use GETVARIATION to instead use the custom opcode.

That, and possibly remapping values returned by GETVARIATION if for axes that are retained, you changed any of min/default/max.

> I haven't seen any font in the wild that actually uses this GETVARIATION, nor I am aware of a hinting tool that actually makes use of it. Last time I checked VTT did not produce this.
> I'm inclined to post-pone this for when we actually find a need for this.

Sure, just document it.  But then it would not have been necessary to implement it for instancing either.  Humm. I thought @nyshadhr9 had seen it used.

nyshadhr9:
No, we actually never saw this used in variable fonts. I think we just assumed this will be a common thing in VFs with hints, so ended up implementing it to be on the safe side.

moyogo:
[Selawik Variations](https://github.com/unicode-org/text-rendering-tests/blob/master/fonts/Selawik-variable.ttf) uses `GETVARIATION`.

anthrotype:
@moyogo thanks.

I noticed that Selawik calls GETVARIATION instruction inside a FDEF function (191) in the fpgm table, whereas in varLib.mutator code (as well as in this PR) we are only checking the individual glyph programs, so it doesn't trigger with Selawik-variable.ttf.

Still, I think this is getting a bit too complicated. Maybe I'll finish working on this another time.

PR #1611, opened 2019-05-15
Make --retain-gids truncate empty glyphs after the last non-empty glyph.

garretrieger:


Lib/fontTools/subset/__init__.py
```diff
@@ -2556,8 +2556,8 @@ def _closure_glyphs(self, font):
 
 		self.glyphs_emptied = frozenset()
 		if self.options.retain_gids:
-			self.glyphs_emptied = realGlyphs - self.glyphs_retained
-			# TODO Drop empty glyphs at the end of GlyphOrder vector.
+			max_retained_glyph_id = max(font.getGlyphID(g) for g in self.glyphs_retained)
+			self.glyphs_emptied = {g for g in realGlyphs - self.glyphs_retained if font.getGlyphID(g) <= max_retained_glyph_id}
```
behdad: There's a better way to write this one to avoid calling getGlyphID on all... Just take the first `max_raintained_glyph_id+1` members of the GlyphOrder and turn into set and subtract glyphs_retained... Feel free to ignore.

Lib/fontTools/subset/__init__.py
```diff
@@ -2556,8 +2556,8 @@ def _closure_glyphs(self, font):
 
 		self.glyphs_emptied = frozenset()
 		if self.options.retain_gids:
-			self.glyphs_emptied = realGlyphs - self.glyphs_retained
-			# TODO Drop empty glyphs at the end of GlyphOrder vector.
+			max_retained_glyph_id = max(font.getGlyphID(g) for g in self.glyphs_retained)
```
anthrotype: in https://github.com/fonttools/fonttools/pull/1588, @blueshade7 added a `self.last_retained_order` attribute to the Subsetter which should be the same as his `max_retained_glyph_id` local variable. Have a look

anthrotype: @garretrieger if you rebase your branch on origin/master, you should see the lines I'm referring to from #1588 appear just below this block

garretrieger: Thanks that simplifies things, rebased and updated to use it.

Lib/fontTools/subset/__init__.py
garretrieger: Rewrote using last_retained_order.

Lib/fontTools/subset/__init__.py
```diff
@@ -2610,12 +2612,17 @@ def _subset_glyphs(self, font):
...
-			with timer("subset GlyphOrder"):
-				glyphOrder = font.getGlyphOrder()
+		with timer("subset GlyphOrder"):
+			glyphOrder = font.getGlyphOrder()
+			if not self.options.retain_gids:
 				glyphOrder = [g for g in glyphOrder if g in self.glyphs_retained]
-				font.setGlyphOrder(glyphOrder)
-				font._buildReverseGlyphOrderDict()
+			else:
+				max_retained_glyph_id = max(font.getGlyphID(g) for g in self.glyphs_retained)
```
anthrotype: you could reuse the `self.last_retained_order` here instead of recomputing `max_retained_glyph_id`

garretrieger: Good catch, fixed.

anthrotype:
thanks

PR #1612, opened 2019-05-17
[subset] Update font extents in head table

khaledhosny:
When --recalc-bounds option is used the font extents in the head table need to be updated, but since tables are lazy-loaded by default the table will not be recompiled and will keep the old value. Force
recompiling the table by adding it to prune_post_subset tables, though I’m not 100% sure this is the best approach.


behdad:
LGTM.

PR #1617, opened 2019-05-20
instancer: make 'None' mean to drop an axis, or pinning at default

anthrotype:
This allows to drop an axis (aka "L1" partial instancing) without knowing the axis' actual default value from fvar table. One can simply call `instantiateVariableFont` function with a `None` value for a particual axis (i.e. `axis_limits={'wght': None}`); the `None` value is replaced by the axis default value as per fvar table.

The same can be done from the console script as well. The special string literal 'None' is parsed as the Python `None` object. E.g.:

```
$ fonttools varLib.instancer MyFont-VF.ttf wght=None
```

### DIFF
--- a/Lib/fontTools/varLib/instancer.py
+++ b/Lib/fontTools/varLib/instancer.py
@@ -677,6 +677,17 @@ def sanityCheckVariableTables(varfont):
         raise NotImplementedError("Instancing CFF2 variable fonts is not supported yet")
 
 
+def populateAxisDefaults(varfont, axis_limits):
+    if any(value is None for value in axis_limits.values()):
+        fvar = varfont["fvar"]
+        defaultValues = {a.axisTag: a.defaultValue for a in fvar.axes}
+        return {
+            axisTag: defaultValues[axisTag] if value is None else value
+            for axisTag, value in axis_limits.items()
+        }
+    return axis_limits
+
+
 def instantiateVariableFont(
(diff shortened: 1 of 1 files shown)


rsheeter:
Why None? Something like "drop" would seem more intuitive?

anthrotype:
> Something like "drop" would seem more intuitive?

only for the CLI arguments or also for the `axis_limits` argument of `instantiateVariableFont`?

To me `None` conveys the meaning of a not-value. Also I wanted to have the CLI and the function arguments be similar to each other: the "None" string literal directly parsed as the `None` object. 
But I can change if you prefer "drop" or believe None is confusing.

rsheeter:
I think that the parsing is an implementation detail that shouldn't inform
the choice of cli args, we should aim for "nice" (as in as
readable/intuitive as possible) CLI.

For the internal code None seems fine, or you could convert to wght: <an
actual value> before you call instantiate. I confess to liking the pass me
actual values option, CLI wrapper can reach out to fvar for default before
calling instantiate, other consumers can do whatever they please.

anthrotype:
ok, i'll change that

anthrotype:
I changed the CLI arg to accept 'drop', but I kept the possibility to pass `None` values in the `axis_limits` dict input. `populateAxisDefaults` needs fvar so it should be called after `sanityCheckVariableTables`, which is what checks that input font is indeed a variable font.

PR #1619, opened 2019-05-23
fix sign of fsType

schriftgestalt:


schriftgestalt:
Can someone have a look. The tests seem to be broke. The tests fail only in some python version.

anthrotype:
That’s correct, the spec says it should be uint16. CI failures are unrelated to the change so I’ll merge. Will fix those later

PR #1621, opened 2019-05-24
fontBuilder: Don't add a stub signature to DSIG

miguelsousa:
An empty DSIG is enough to make MS applications happy and also fixes an error reported by MS Font Validator.

### DIFF
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -712,25 +712,14 @@ class FontBuilder(object):
         self._initTableWithValues("maxp", defaults, {})
 
     def setupDummyDSIG(self):
-        """This adds a dummy DSIG table to the font to make some MS applications
+        """This adds an empty DSIG table to the font to make some MS applications
         happy. This does not properly sign the font.
         """
-        from .ttLib.tables.D_S_I_G_ import SignatureRecord
-
-        sig = SignatureRecord()
-        sig.ulLength = 20
-        sig.cbSignature = 12
-        sig.usReserved2 = 0
-        sig.usReserved1 = 0
-        sig.pkcs7 = b'\xd3M4\xd3M5\xd3M4\xd3M4'
-        sig.ulFormat = 1
-        sig.ulOffset = 20
-
         values = dict(
             ulVersion = 1,
-            usFlag = 1,
-            usNumSigs = 1,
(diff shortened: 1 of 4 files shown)


justvanrossum:
Why do we need two different fake/dummy DSIG tables? If the empty one is the more canonical way to do it, I'd be happy to just have that.

anthrotype:
I wonder if it’s actually still the case that “some MS applications require this”, do you have concrete examples with recent versions?
I have tried unsuccessfully a few years ago on the opentype mailing list to raise the issue that the DSIG spec does not match the MS implementation; and that since nobody would or could implement a tool for signing/verifying such signatures it would have been best to deprecate the table and remove it from the OT spec.
I would prefer that fonttools did _not_ add such “dummy” DSIG table, as that would contribute to maintain this situation.
I believe the anybody can write a simple script based on fonttools to add such table if they wish (and there are numerous snippets floating around such as this one).

moyogo:
Plenty of people are still using versions of Microsoft Office applications that work fine for them and that require the DSIG for OT features to work.
I’d recomend fontBuilder add a dummy DSIG by default, or at least provide the option to do it easily.

chrissimpkins:
Is there more information about what OT features this issue influences for MS Word users at any affected version?

anthrotype:
If we have to add one I prefer it is an empty one rather than an invalid but non-empty DSIG. And possibly disabled by default so if one cares and knows what they are doing the can opt in

chrissimpkins:
> I prefer it is an empty one rather than an invalid but non-empty DSIG

Is the public/private key authentication "feature" of applications that use this table removed and, for some reason, they simply need the table to be present to support some OT features as Denis suggested above? If DSIG certificate authentication is still occurring out there, then the approach in your comment seems mandatory Cosimo.

miguelsousa:
@justvanrossum an empty DSIG is enough. That's what makeotf has been doing for a while and we know of no side effects. Also, MS Font Validator complains about the stub signature. I've updated the PR.

@anthrotype FontBuilder only adds a DSIG when `setupDummyDSIG` is called.

justvanrossum:
> an empty DSIG is enough

Then let's make it do that instead of the current behavior. 

Yes, FontBuilder never adds a dummy DSIG implicitly.

PR #1626, opened 2019-05-29
instancer: bugfixes and tests for instantiateFeatureVariations

anthrotype:
we had forgotten to remap the ConditionTable.AxisIndex of remaining FeatureVariationRecords after dropping axes from fvar.

Also, we can drop lookups that are no longer referenced after we have subsetted or completely removed the FeatureVariations table.

This also adds unit tests for the instantiateFeatureVariations function, using the same test file "FeatureVars.ttx" used in varLib_test.py (the one built from Tests/varLib/data/FeatureVars.designspace with the dollar trick).


anthrotype:
gonna merge this now, and send the big PR to master soon afterwards

Issue #1537, opened 2019-03-13
[varLib.mutator] Support partial VF instances


anthrotype:
The current fontTools.varlib.mutator module can fully instantiate a variable font into a static font at a given location along all the axes.

The Google Fonts team is interested in adding support for generating "partial" instances: i.e one wants to keep `wght` axis but pin the `wdth` axis to a specific location, either the default or a custom one,  dropping all the variation data relating to the pinned axes.
Or alternatively one may want to restrict the range (min and/or max) of an axis, producing a new VF with possibly smaller file size.

We have created a new branch `partial-instancer` in here where we pushed a new module called `fontTools.varLib.partialInstancer` (the name may change, I'm thinking just "instancer" sounds better). It's is based on the mutator module and has a similar interface. When it's complete, potentially it could subsume the functionality of mutator, since the partial instancing can be seen as a more general problem.

The module is still in early stages and can currently only pin-and-drop gvar variation data. Limiting axis ranges is not supported either. We shall probably focus on implementing the axis dropping and pinning first for the rest of the variation tables, as that appear to be a relatively easier problem than limiting ranges, and one that guarantees some size reduction.

Feel free to chime in with suggestions or comments. I'll keep you updated with the progress.

behdad:
I suggest we break this operation down to separate operations that can be chained together.  That way the code would be much easier to reason about.  I suggest implementing the following:

- Dropping an entire axis.  This is simplest: any deltaset that has a non-zero peak for the said axis is just dropped completely.

- Instantiating an axis to a particular value.  The previous operation is a special-case of this one.  In this operation, any deltaset that interacts with the axis will get its deltas multiplied by the scalar support of the axis at the desired instantiation point for the deltaset's tent.  If the scalar is 0, drop the deltaset.  If it's 1, leave as is, otherwise multiply the values by it.

- Limit min/max of an axis.  For each deltaset that participates in the axis, if the deltaset's tent is fully outside new min/max of axis, drop it.  If the tent is fully inside the new range, keep it (just scale the tent min/peak/max to the new normalized axis).  If the tent overlaps but also extends beyond new min/max, it becomes interesting.  I'll address that separately in the following comment.

- Move the base location.  This is harder in the general case and to be done later, separately, if ever.

behdad:
[...]

PR #1628, opened 2019-05-31
add new varLib.instancer module

anthrotype:
Part of https://github.com/fonttools/fonttools/issues/1537

This adds a new `varLib.instancer` module and CLI tool, similar to `varLib.mutator`, from which it borrows a much of the logic (but is self-contained and independent from the latter), and extend it to allow creating not just full static instances from a variable font but also "partial" variable fonts without some of the original axes.

Currently it can either drop one or more axes while leaving the default tables unmodified, or it can pin or freeze them at non-default locations. A third level of partial instancing, whereby axes are restricted within a new given min..max range, is planned to be implemented at a later stage.

FWIW, the test coverage for the new module alone is at a 97%.


anthrotype:
(ignore the appveyor test failure, something is up with the logging configuration, I'll investigate later)

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -446,6 +444,125 @@ def getTupleSize_(flags, axisCount):
 			size += axisCount * 4
 		return size
 
+	def getDeltaType(self):
+		""" Check if deltas are (x, y) as in gvar, or single values as in cvar.
+		Returns a string ("gvar" or "cvar"), or None if empty.
+		"""
+		firstDelta = next((c for c in self.coordinates if c is not None), None)
+		if firstDelta is None:
+			return  # empty or has no impact
```
rsheeter: us hasImpact?

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -446,6 +444,125 @@ def getTupleSize_(flags, axisCount):
...
+		if type(firstDelta) is tuple and len(firstDelta) == 2:
+			return "gvar"
+		elif type(firstDelta) in (int, float):
+			return "cvar"
```
rsheeter: would it make sense to return 1 or 2 [or 0 if no impact?], the number of active values, rather than tags?

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -54,10 +55,7 @@ def hasImpact(self):
 		If the result is False, the TupleVariation can be omitted from the font
 		without making any visible difference.
 		"""
-		for c in self.coordinates:
-			if c is not None:
-				return True
-		return False
+		return any(c is not None for c in self.coordinates)
```
rsheeter: if you had a function for first active delta you could use it here and in getDeltaType. Currently more or less duplicates the loop over coordinates.

Lib/fontTools/ttLib/tables/TupleVariation.py
anthrotype: yes, I could call `hasImpact` here, but doing this saves me an extra for loop and method call

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -446,6 +444,125 @@ def getTupleSize_(flags, axisCount):
...
+			return "gvar"
+		elif type(firstDelta) in (int, float):
+			return "cvar"
+		else:
+			raise TypeError("invalid type of delta: %s" % type(firstDelta))
+
+	def scaleDeltas(self, scalar):
+		if scalar == 1.0:
+			return  # no change
+		deltaType = self.getDeltaType()
```
rsheeter: Looks repetitive and self-similar.

if you got the tuple width here could you generalize the code? E.g. if scaler 0 self.coordinates = [tuple of <width> 0's] * len(self.coordinates). Then we just need to collapse the else branches :)

anthrotype: sorry, I didn't get this. WDYM by "tuple width"? The value of self.coordinates is either a tuple (as in the python built-in type) or a float.

Lib/fontTools/varLib/instancer.py
```diff
@@ -0,0 +1,1031 @@
...
+    entire regions of variations that fall completely outside this new range.
+L4) moving the default location of an axis.
+
+Currently only TrueType-flavored variable fonts (i.e. containing 'glyf' table)
+are supported, but support for CFF2 variable fonts will be added soon.
+
+The discussion and implementation of these features are tracked at
+https://github.com/fonttools/fonttools/issues/1537
+
+NOTE: The module is experimental and both the API and the CLI *may* change.
```
rsheeter: clarify how it will become non-experimental and, even roughly, when

Lib/fontTools/ttLib/tables/TupleVariation.py
rsheeter: (it bugs me we have all the branches with almost identical code on both branches)

anthrotype: TupleVariation.coordinates is either a list of tuples of floats ("gvar"), or a list of floats ("cvar"). I cannot collapse the else branches in this case. I prefer not to change the TupleVariation class too much.

Lib/fontTools/varLib/instancer.py
anthrotype: I'll remove that line, it was copy-pasted from mutator. I don't foresee any meaningful changes in the API (e.g. the only publicly exported function is `instantiateVariableFont`).

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -243,6 +243,175 @@ def __len__(self):
...
+		# Add phantom points for (left, right, top, bottom) positions.
+		phantomPoints = self.getPhantomPoints(
+			glyphName, ttFont, defaultVerticalOrigin=defaultVerticalOrigin
+		)
+		coords.extend(phantomPoints)
+		return coords, controls
+
+	def getCoordinates(self, glyphName, ttFont, defaultVerticalOrigin=None):
+		"""Same as `getCoordinatesAndControls` but only returns coordinates array,
+		or None if the glyph is missing.
```
rsheeter: This is a weird pattern (why doesn't one fn use the other?), definitely want to explain why we do this.

On that note, why not have just the one function? you could take a param for whether you want controls or not if it's expensive.

Lib/fontTools/varLib/instancer.py
```diff
@@ -0,0 +1,1031 @@
+""" Partially instantiate a variable font.
+
+This is similar to fontTools.varLib.mutator, but instead of creating full
```
rsheeter: Not sure it's helpful to define this in terms of mutator, just define what it is directly.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
anthrotype: yeah, i'll remove it.

Lib/fontTools/varLib/instancer.py
```diff
@@ -0,0 +1,1031 @@
...
+import logging
+from itertools import islice
+import os
+import re
+
+
+log = logging.getLogger("fontTools.varLib.instancer")
+
+
+def instantiateTupleVariationStore(variations, location, origCoords=None, endPts=None):
```
rsheeter: This would benefit from comments about what the args are, absent both types and comments it's difficult to tell what a variations or location *is*

Lib/fontTools/varLib/instancer.py
```diff
@@ -0,0 +1,1031 @@
...
+            var.calcInferredDeltas(origCoords, endPts)
+
+        var.scaleDeltas(scalar)
+
+        # merge TupleVariations with overlapping "tents"
+        axes = tuple(var.axes.items())
+        if axes in newVariations:
+            newVariations[axes] += var
+        else:
+            newVariations[axes] = var
```
rsheeter: get w/default, no if?

anthrotype: I'm setting, not getting. Maybe you mean `setdefault` method?
Here I'm grouping the TupleVariations that have the same axes (same axis tags and and same (min/default/max)) and summing their deltas as I go.

Lib/fontTools/varLib/instancer.py
```diff
@@ -0,0 +1,1031 @@
...
+                    applies = False
+                    condition.AxisIndex = axisIndexMap[axisTag]
+                    newConditions.append(condition)
+            else:
+                log.warning(
+                    "Condition table {0} of FeatureVariationRecord {1} has "
+                    "unsupported format ({2}); ignored".format(j, i, condition.Format)
+                )
+                applies = False
+                newConditions.append(condition)
```
rsheeter: suggest rewrite to not have the deep nesting, either reordering and/or using narrow purpose functions.

For example [very pseudocode]:

```
# _instantiateFeatureVariations
for i, record in ...
  instantiateFeatureVariation

# instantiateFeatureVariation
# loop and call fn for every condition

# fn for every condition
if format != 1
  log
  return False
if axisTag not in pinnedAxes
  ...
  return False
...now do stuff...
```

Lib/fontTools/ttLib/tables/TupleVariation.py
rsheeter: Consider the branches to multiply by 0. You are just trying to multiply by 0 or by (0,0). So, pseudo-code:

```
width = getDeltaWidth()
if scalar == 0:
  self.coordinates = _coordMultiplier(width, 0) # returns 0 for width 1, (0, 0) for width 2
```

But ... why branch on scalar 0 at all? - multiplying by 0 often gets you to 0. So, replace L465..480 with something like:

```
_scaleCoordinates(scalar)

def scaleCoordinates(scalar):
  coord_width = getCoordWidth()
  multiplier = scalar
  if coord_width > 1:
    multiplier = tuple([scalar] * coord_width)
  # ...now multiply all the coordinates by multiplier...
  # if you really care that for scalar 0 that None => 0 you can code that here too
```

anthrotype: > You are just trying to multiply by 0 or by (0,0)

i'm confused. I'm not "multiplying" by 0 or (0, 0). I'm using the `*` operator on a list object together with an int operand in order to obtain a new list of given length, with the content of the original list repeated.
One cannot just multiply tuples together. I want to multiply their content (floats) by another float (scalar).
The code is complicated by the fact that TupleVariation.coordinates is a `List[Optional[Union[Tuple[float, float], float]]]`, e.g.: these are all valid values: `[None, None]`, `[(0, 0), (1, 1), None, (3, 3)]`, or `[1, 2, None, 4, 5]`.

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -446,6 +444,125 @@ def getTupleSize_(flags, axisCount):
...
+						"cannot sum gvar deltas with inferred points"
+					)
+		else:
+			for i, d2 in zip(range(length), deltas2):
+				d1 = deltas1[i]
+				if d1 is not None and d2 is not None:
+					deltas1[i] = d1 + d2
+				elif d1 is None and d2 is not None:
+					deltas1[i] = d2
+				# elif d2 is None do nothing
```
rsheeter: Here again we're writing code twice to handle one vs two values; lets figure out a way to not do that. For example, we could have a width-aware add fn. I vaguely imagine you end up with something like:

```
if their deltas length != ours err
if thier coord width != ours err [I think we're missing this currently?]
deltas = [_addCoord(d1, d2) for d1, d2 in zip(self.coordinates, other.coordinates)]
# _addCoord can get and use the width and do the right thing
```

Lib/fontTools/ttLib/tables/TupleVariation.py
rsheeter: My suggestion is essentially implement a general purpose multiply coords instead of all this branching.

Looking at 0s specifically it appears to me the code that says if scalar = 0 coordinates = [0] * # of coords is just short-circuit for the notion that if we multiplied each coord by 0 we'd get all 0's. And then we code it all over again to support the case where we have two coordinates. Then repeat for non-zero multiplication once for each width.

Lib/fontTools/ttLib/tables/TupleVariation.py
anthrotype: I don't think it's worth to turn a one-liner into its own method. The generator expression `next((c for c in self.coordinates if c is not None), None)` is efficient and readable to me.

Lib/fontTools/ttLib/tables/TupleVariation.py
anthrotype: I removed the special case for scalar = 0, but we still need to branch out when scaling, adding, rounding, etc. depending on whether the coordinate is a bare number or a tuple of numbers.

Lib/fontTools/ttLib/tables/TupleVariation.py
anthrotype: I renamed getDeltaType to getCoordWidth (returns 0, 1 or 2)

Lib/fontTools/ttLib/tables/TupleVariation.py
anthrotype: > if thier coord width != ours err [I think we're missing this currently?]

no, it is possible that `self.getCoordWidth() == 0` but `other.getCoordWidth() != 0`, e.g. all self.coordinates are None but other's aren't, in which case adding self with the other gives the other.

anthrotype: Another reason for all the branching is that the `__iadd__` method needs to deal with the different meanings that the `None` coordinates values have in gvar vs cvar TupleVariations: within the gvar, when deltas are not provided explicitly for some points, they need to be inferred; however, this does not apply for the 'cvar' table, where if deltas are not provided for some CVT values, then no adjustments are made (i.e. `None` is same as 0).
So when we have a gvar-style TupleVariation (coordWidth=2) we cannot meaningful "add" its coordinates to another such TupleVariation unless we have first computed all the inferred None values (hence the defensive ValueError when that's not the case).
When the `None` values occur in a cvar-style TupleVariation (coordWidth=1) we can "add" them up by effectively treating `None` as a 0.

anthrotype: and the reason I don't bother checking if we are adding TupleVariation of different width is that it is unlikely that one ever tries to sum a TupleVariation from gvar with another from cvar or viceversa.
These two flavors of TupleVariation really should be different types, rather than same class parametrised by the type of its content (tuple of numbers vs plain numbers). But I didn't write this code originally and I wanted to make the new instancer's code more or less self-contained without introducing too sweeping changes to the rest of the library.

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -446,6 +444,118 @@ def getTupleSize_(flags, axisCount):
...
+		if coordWidth == 2:
+			self.coordinates = [
+				(d[0] * scalar, d[1] * scalar) if d is not None else None
+				for d in self.coordinates
+			]
+		elif coordWidth == 1:
+			self.coordinates = [
+				d * scalar if d is not None else None
+				for d in self.coordinates
+			]
```
evanadams: I understand Rod's comments about the coding being duplicated between the coordWidth == 1 and coordWidth == 2 cases. It's annoying that both the loop over self.coordindates and the check for None are duplicated.

You mentioned below that there are some semantical differences regarding None in some situations, but i'm not seeing any in this function or in roundDeltas below.  

The biggest problem is that coordWidth == 2 is operating on tuples while coordWidth == 1 is operating on scalars.

I did come up with this:
   self.coordinates = [None if d is None else d * scalar if coordWidth == 1 else (d[0] * scalar, d[1] * scalar) for d in deltas]

This has the loop and the check for None once rather than twice.

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -446,6 +444,118 @@ def getTupleSize_(flags, axisCount):
...
+
+	def roundDeltas(self):
+		coordWidth = self.getCoordWidth()
```
evanadams: Should coordWidth be computed just once and then assigned to field?

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -446,6 +444,118 @@ def getTupleSize_(flags, axisCount):
...
+						"cannot sum gvar deltas with inferred points"
+					)
+		else:
+			for i, d2 in zip(range(length), deltas2):
+				d1 = deltas1[i]
+				if d1 is not None and d2 is not None:
+					deltas1[i] = d1 + d2
+				elif d1 is None and d2 is not None:
+					deltas1[i] = d2
+				# elif d2 is None do nothing
```
evanadams: Is it possible for d1 to be None?

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -243,6 +243,154 @@ def __len__(self):
...
+		If the ttFont doesn't contain a "vmtx" table, the hhea.ascent is used as the
+		vertical origin, and the head.unitsPerEm as the vertical advance.
+
+		The "defaultVerticalOrigin" (Optional[int]) is used when the ttFont contains
+		neither a "vmtx" nor an "hhea" table.
+
+		https://docs.microsoft.com/en-us/typography/opentype/spec/tt_instructing_glyphs#phantoms
+		"""
+		glyph = self[glyphName]
+		assert glyphName in ttFont["hmtx"].metrics, ttFont["hmtx"].metrics
```
evanadams: Looks like ttFont["hmtx"].metrics is being checked twice.

Lib/fontTools/ttLib/tables/TupleVariation.py
anthrotype: yes, either could be None. `d1 is None` is already handled when `d2 is not None`; when both d1 and d2 are None, nothing happens.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
anthrotype: I don't think so. The optional expression after the comma in an `assert` statement is printed (for debugging purposes) when the assertion fails.

Lib/fontTools/ttLib/tables/TupleVariation.py
anthrotype: the thing is, the way it's defined, this coordWidth can actually change as result of `optimize` (it can become 0 if all deltas are set to None), or `__iadd__` (if self coordWidth is 0 and other has coordWidth 1) or `calcInferredDeltas` (coordWidth can change from 0 to 2). So we need to compute it each time.

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -446,6 +444,118 @@ def getTupleSize_(flags, axisCount):
...
+		as in cvar, or 0 if empty.
+		"""
+		firstDelta = next((c for c in self.coordinates if c is not None), None)
+		if firstDelta is None:
+			return 0  # empty or has no impact
+		if type(firstDelta) is tuple and len(firstDelta) == 2:
+			return 2
+		elif type(firstDelta) in (int, float):
+			return 1
+		else:
```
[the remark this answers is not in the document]
anthrotype: changed to use all if-stmt and reordered as suggested.

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -446,6 +444,118 @@ def getTupleSize_(flags, axisCount):
...
+			raise TypeError("invalid type of delta: %s" % type(firstDelta))
```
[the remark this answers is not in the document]
anthrotype: added more info to error message expalining what the expected types are

Lib/fontTools/ttLib/tables/TupleVariation.py
anthrotype: ok, I chained the two if statements in one line like you proposed.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -243,6 +243,154 @@ def __len__(self):
...
+		leftSideX = coord[-4][0]
+		rightSideX = coord[-3][0]
+		topSideY = coord[-2][1]
+		bottomSideY = coord[-1][1]
+
+		coord = coord[:-4]
+
+		if glyph.isComposite():
+			assert len(coord) == len(glyph.components)
+			for p,comp in zip(coord, glyph.components):
```
evanwadams: A nit: the rest of file appears to have a space after a comma.  That's missing here and two lines below.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -243,6 +243,154 @@ def __len__(self):
...
+			glyphName, ttFont, defaultVerticalOrigin=defaultVerticalOrigin
+		)
+		coords.extend(phantomPoints)
+		return coords, controls
+
+	def setCoordinates(self, glyphName, coord, ttFont):
+		"""Set coordinates and metrics for the given glyph.
+
+		"coord" is an array of GlyphCoordinates which must include the four
+		"phantom points".
```
evanwadams: A nit: probably worth mentioning that the "phantom points" must be the last 4 coords.

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -446,6 +444,117 @@ def getTupleSize_(flags, axisCount):
...
+
+	def scaleDeltas(self, scalar):
+		if scalar == 1.0:
+			return  # no change
+		coordWidth = self.getCoordWidth()
+		self.coordinates = [
+			None
+			if d is None
+			else d * scalar
+			if coordWidth == 1
```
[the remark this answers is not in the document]
anthrotype: it is handled implicitly by the first `None if d is None else ...`, because coordWidth == 0 iff the self.coordinates are _all_ None (or empty).

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -446,6 +444,117 @@ def getTupleSize_(flags, axisCount):
...
+			else otRound(d)
+			if coordWidth == 1
+			else (otRound(d[0]), otRound(d[1]))
+			for d in self.coordinates
+		]
+
+	def calcInferredDeltas(self, origCoords, endPts):
+		from fontTools.varLib.iup import iup_delta
+
+		if self.getCoordWidth() == 1:
```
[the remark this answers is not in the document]
anthrotype: yes -- if self.coordinates are all None; that's handled already. Only when coordWidth == 1 (e.g. for 'cvar' tuples), it's invalid to call `calcInferredDeltas` method, hence the ensuing error.

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -446,6 +444,117 @@ def getTupleSize_(flags, axisCount):
...
+			raise TypeError(
+				"Only 'gvar' TupleVariation can have inferred deltas"
+			)
+		if None in self.coordinates:
+			if len(self.coordinates) != len(origCoords):
```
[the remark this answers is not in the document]
anthrotype: `None in self.coordinates` tests if there are any deltas to be inferred, to decide whether to proceed calling `iup_delta`. The check wherby `len(self.coordinates)` be equal to `len(origCoords)` is only relevant/useful when we actually call `iup_delta`.

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -446,6 +444,117 @@ def getTupleSize_(flags, axisCount):
...
+		if len(deltas2) != length:
+			raise ValueError(
+				"cannot sum TupleVariation deltas with different lengths"
+			)
+		if self.getCoordWidth() == 2:
+			for i, d2 in zip(range(length), deltas2):
+				d1 = deltas1[i]
+				try:
+					deltas1[i] = (d1[0] + d2[0], d1[1] + d2[1])
+				except TypeError:
```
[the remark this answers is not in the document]
anthrotype: it's just a sanity check, if one were to call `__iadd__` method on a TupleVariation before having calcInferredDeltas (which is not the case for the instancer).

Lib/fontTools/ttLib/tables/TupleVariation.py
anthrotype: ok, I can add a comment.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -243,6 +243,154 @@ def __len__(self):
...
+		else:
+			# without vmtx, use ascent as vertical origin and UPEM as vertical advance
+			# like HarfBuzz does
+			verticalAdvanceWidth = ttFont["head"].unitsPerEm
+			try:
+				topSideY = ttFont["hhea"].ascent
+			except KeyError:
+				# sparse masters may not contain an hhea table; use the ascent
+				# of the default master as the vertical origin
+				assert defaultVerticalOrigin is not None
```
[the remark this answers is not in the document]
anthrotype: caller is required to pass a `defaultVerticalOrigin` when getting phantom points from a sparse master font that may not contain an hhea table. I will change it so that the KeyError("hhea") is raised when defaultVerticalOrigin is not provided.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
anthrotype: it was copy-pasted from elsewhere, but i'll fix that, thanks.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
anthrotype: ok, I'll clarify that.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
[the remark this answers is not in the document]
anthrotype: not worth raising an error here, a warning shold be ok; if no vmtx and hhea tables are present and caller didn't bother passing a defaultVerticalOrigin, they probably don't care about the vertical phantom points. I can use the UPEM for both verticalAdvanceWidth and topSideY coord.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -243,6 +243,154 @@ def __len__(self):
...
+		if "vmtx" in ttFont:
+			verticalAdvanceWidth, topSideBearing = ttFont["vmtx"].metrics[glyphName]
+			topSideY = topSideBearing + glyph.yMax
+		else:
+			# without vmtx, use ascent as vertical origin and UPEM as vertical advance
+			# like HarfBuzz does
+			verticalAdvanceWidth = ttFont["head"].unitsPerEm
+			try:
+				topSideY = ttFont["hhea"].ascent
+			except KeyError:
```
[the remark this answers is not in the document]
anthrotype: I see what you mean; my thinking was that 99 % of the fonts do have hhea table so the fact that it's missing is an "exception". But since 'sparse' masters (without hhea) are considered part of the normal input for this method, I can instead check for its existence.

anthrotype:
I hereby press the green button. Thanks everybody for the reviews!

PR #1638, opened 2019-06-11
varLib: Fix merging of class kerning tables

madig:
The current code in `varLib.merger` may produce corrupted kerning in variable fonts, which manifests itself in two ways:

1) Random classes against class zero (1st against 2nd) may have a "VariationIndex" table attached, causing kerning to be applied in certain variable regions.
2) Some kerning specified for one class may end up in another or actually come from another master?

I found this "ghost kerning" in several of our variable library fonts and also in IBM Plex Sans Variable. I used the script in https://gist.github.com/madig/248e3090de0f6c77001dee266181e59b to detect case 1. For case 2, I used the kern-dump utilities to dump the kerning of e.g. the regular static master (from `fontmake -m designspace -o ttf-interpolatable`) and a static instance cut from the VF (`fonttools varLib.mutator myVF.ttf -o test.ttf`).

The cause seems to be a forgotten deep-copying of the class kerning tables before aligning matrices in `_PairPosFormat2_merge`, so the code works on unintentionally shared data structures while assuming they are independant (https://github.com/fonttools/fonttools/commit/6f41c2ab53ccf308e2a3f4b789860ce5d60d607b).


Tests/varLib/varLib_test.py
```diff
@@ -496,6 +496,64 @@ def test_varlib_build_VVAR_CFF2(self):
...
+        class_kerning_table = class_kerning_tables[0]
+
+        # Test that no class kerned against class zero (containing all glyphs not
+        # classed) has a `XAdvDevice` table attached, which in the variable font
+        # context is a "VariationIndex" table and points to kerning deltas in the GDEF
+        # table. Variation deltas of any kerning class against class zero should
+        # probably never exist.
+        for class1_record in class_kerning_table.Class1Record:
+            class2_zero = class1_record.Class2Record[0]
+            assert vars(class2_zero.Value1).get("XAdvDevice", None) is None
```
anthrotype:
```suggestion
            assert getattr(class2_zero.Value1, "XAdvDevice", None) is None
```

Tests/varLib/data/master_kerning_merging/0.ttx
```diff
@@ -0,0 +1,296 @@
...
+    <mtx name="C" width="500" lsb="0"/>
+    <mtx name="D" width="500" lsb="0"/>
+  </hmtx>
+
+  <cmap>
+    <tableVersion version="0"/>
+    <cmap_format_4 platformID="0" platEncID="3" language="0">
+    </cmap_format_4>
+    <cmap_format_4 platformID="3" platEncID="1" language="0">
+    </cmap_format_4>
```
anthrotype: maybe assign unicode codepoints to your test master fonts so we get a non-empty cmap table and one could test the kerning pairs in a real shaper.

anthrotype:
.@behdad's [suggestion](https://github.com/fonttools/fonttools/commit/6f41c2ab53ccf308e2a3f4b789860ce5d60d607b#commitcomment-33880880) to deepcopy this line:

https://github.com/fonttools/fonttools/blob/a56b1af2f6ba0115480dd2898e202eb73e688fe8/Lib/fontTools/varLib/merger.py#L494

doesn't work because [deepcopy](https://docs.python.org/3.7/library/copy.html#copy.deepcopy) function keeps a `memo` dictionary of objects already copied, and `matrices[0]` may already contain some shared `Class2Record` objects.

@madig's fix ensure that all Class2Record objects in `matrices` are distinct instances, so when later they are mutated (a VariationIndex table may be assigned to them if they vary across masters) all works as expected.

PR #1639, opened 2019-06-11
[woff2] support hmtx transform + glyf/loca without transformation

### LINKED ISSUE #1636: [woff2] encoder should allow to compress font without transforming glyf/loca
in order to preserve the OVERLAP_SIMPLE glyf flags

Also see https://github.com/google/woff2/issues/123

and https://github.com/fonttools/fonttools/issues/1281

anthrotype:
Fixes #1636 and https://github.com/fonttools/fonttools/issues/1281

The ~~`Snippets/woff2_compress.py`~~ `fonttools ttLib.woff2 compress` script now has a `--no-glyf-transform` that allows to turn off the default `glyf` (and `loca`) preprocessing transformations.
This allows to have the `glyf` table data passed through unchanged, so that the OVERLAP_SIMPLE flags are kept instead of dropped by the transformation.

I tested manually locally, but I will add proper tests later.

/cc @Lorp


anthrotype:
Since I'm touching the woff2 module, I figured that it was time to also add support for hmtx trasformation (still not on by default though). I added an `--hmtx-transform` command line option to the `Snippets/woff2_compress.py` script.

anthrotype:
I have added a console entry point for `fonttools ttLib.woff2` (which is the same as doing  `python -m fontTools.ttLib.woff2`), and removed the two `woff2_compress.py` and `woff2_decompress.py` scripts from Snippets/ folder.

To compress a font to WOFF2 now one can do:

```bash
$ fonttools ttLib.woff2 compress MyFont-Regular.ttf
```

To decompress:

```bash
$ fonttools ttLib.woff2 decompress MyFont-Regular.woff2
```

The `compress` sub-command allows to disable default glyf+loca transformation, as well as enable the optional hmtx transformation (cf. `fonttools ttLib.woff2 compress --help`).

Issue #1624, opened 2019-05-28
[varLib] Merging vpal GPOS feature can't handle differing number of subtables


readroberts:
The vpal GPOS feature can be constructed differently in different source fonts. A glyph may require both and xAdvance and xPlacement adjustment in one font, and only a xAdvance in another fonts. Also in one font, a long sequence of glyphs may use the same value record, but the same sequence of glyphs may use the same value record format, but have different adjustment values. These differences lead to the source fonts having a different number of subtables after optimization for size. The current code can't merge the features if the source fonts differ in number of subtables, or if the subtables  differ in coverage or number of value records. I attached a small test case. The GPOS data is edited in rather arbitrarily. If you have the afdko, you can build this with 'buildcff2vf -d test_vpal.designspace'.
[test_vpal.zip](https://github.com/fonttools/fonttools/files/3230051/test_vpal.zip)

miguelsousa:
The test files provided can be exercised with this command,

    fonttools varLib --master-finder master_vpal_test/{stem}.otf test_vpal.designspace

This is the traceback it will produce:

```
Merging OpenType Layout tables
Traceback (most recent call last):
  File "fonttools", line 10, in <module>
    sys.exit(main())
  File "fontTools/__main__.py", line 29, in main
    runpy.run_module(mod, run_name='__main__')
  File "lib/python3.7/runpy.py", line 208, in run_module
    return _run_code(code, {}, init_globals, run_name, mod_spec)
  File "lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "fontTools/varLib/__main__.py", line 7, in <module>
    sys.exit(main())
  File "fontTools/varLib/__init__.py", line 1064, in main
    optimize=options.optimize
  File "fontTools/varLib/__init__.py", line 916, in build
    _merge_OTL(vf, model, master_fonts, axisTags)
  File "fontTools/varLib/__init__.py", line 688, in _merge_OTL
    merger.mergeTables(font, master_fonts, ['GSUB', 'GDEF', 'GPOS'])
  File "fontTools/varLib/merger.py", line 108, in mergeTables
    for m in master_ttfs])
  File "fontTools/varLib/merger.py", line 950, in mergeThings
    super(VariationMerger, self).mergeThings(out, lst)
  File "fontTools/varLib/merger.py", line 94, in mergeThings
    self.mergeObjects(out, lst)
  File "fontTools/varLib/merger.py", line 73, in mergeObjects
[...]

PR #1641, opened 2019-06-12
Fix singlepos merge

readroberts:
Fix issue [varLib] Merging vpal GPOS feature can't handle differing number of subtables #1624.


Lib/fontTools/varLib/merger.py
```diff
@@ -726,6 +728,31 @@ def _Lookup_PairPos_subtables_canonicalize(lst, font):
 
 	return lst
 
+def _Lookup_SinglePos_subtables_flatten(lst, font, min_inclusive_rec_format):
+	lst = list(lst)
```
anthrotype: what's the purpose of the shallow copy `list(lst)` here? is it because you don't want to modify the original master fonts' SinglePos subtables?

Lib/fontTools/varLib/merger.py
```diff
@@ -743,19 +770,37 @@ def merge(merger, self, lst):
 			sts.extend(new_sts)
 
 	isPairPos = self.SubTable and isinstance(self.SubTable[0], ot.PairPos)
-
+	isSinglePos = False
```
anthrotype: this line `isSinglePos = False` is redundant, given that below you do `if isPairPos: ... else:` and then evaluate again `isSinglePos = ...`

Lib/fontTools/varLib/merger.py
```diff
@@ -743,19 +770,37 @@ def merge(merger, self, lst):
...
+					_Lookup_SinglePos_subtables_flatten(st, merger.font, mirf) for st in subtables]
+				flattened = True
+			else:
+				flattened = False
+	
 	merger.mergeLists(self.SubTable, subtables)
 	self.SubTableCount = len(self.SubTable)
 
 	if isPairPos:
+		isSinglePos = False
```
anthrotype: you could remove this line `isSinglePos = False` and turn the following `if isSinglePos:` branch into an `elif isSinglePos:`.
I feel this function is becoming too long and complex.

Lib/fontTools/varLib/merger.py
```diff
@@ -768,6 +813,17 @@ def merge(merger, self, lst):
 			self.SubTable.pop(-1)
 			self.SubTableCount -= 1
 
+	if isSinglePos:
+		singlePosTable = self.SubTable[0]
+		glyphs = singlePosTable.Coverage.glyphs
+		if flattened:
```
anthrotype: i would collapse `if isSinglePos and flattened:`

Lib/fontTools/varLib/merger.py
```diff
@@ -768,6 +813,17 @@ def merge(merger, self, lst):
...
+			# We know that singlePosTable is Format 2, as this is set
+			# in _Lookup_SinglePos_subtables_flatten.
+			recs = singlePosTable.Value
+			numRecs = len(recs)
+			recList = [ (glyphs[i], recs[i]) for i in range(numRecs)]
+			singlePosMapping = {gname: valRecord for gname, valRecord in recList}
```
anthrotype: you could simplify the four lines above like so
```python
singlePosMapping = {gname: valRecord for gname, valRecord in zip(glyphs, singlePosTable.Value)}
```

anthrotype:
@readroberts I took the liberty of applying the fix myself since I would like to include this in the upcoming release, I hope you don't mind. Thanks

readroberts:
@anthrotype This is good timing - I was about to do this today. I am always delighted when someone else does work instead of me - I always have much more to do that time allows, even more so now that I am now working only three days a week, Tues-Thurs. This also limits my ability to respond quickly, or at all when I am away during my days off.

Issue #1632, opened 2019-06-03
KeyError in varLib.cff.merge_PrivateDicts


miguelsousa:
Getting the error below while trying to build the fonts at https://github.com/adobe-fonts/adobe-blank-vf

Test case attached. The command was,

    fonttools varLib --master-finder fonts/{stem}.otf AdobeBlankVF.designspace

```
Traceback (most recent call last):
  File "fonttools", line 10, in <module>
    sys.exit(main())
  File "fontTools/__main__.py", line 29, in main
    runpy.run_module(mod, run_name='__main__')
  File "runpy.py", line 208, in run_module
    return _run_code(code, {}, init_globals, run_name, mod_spec)
  File "runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "fontTools/varLib/__main__.py", line 7, in <module>
    sys.exit(main())
  File "fontTools/varLib/__init__.py", line 1064, in main
    optimize=options.optimize
  File "fontTools/varLib/__init__.py", line 924, in build
    _add_CFF2(vf, model, master_fonts)
  File "fontTools/varLib/__init__.py", line 767, in _add_CFF2
    merge_region_fonts(varFont, model, ordered_fonts_list, glyphOrder)
  File "fontTools/varLib/cff.py", line 310, in merge_region_fonts
    merge_PrivateDicts(top_dicts, cvData.vsindex_dict, model, fd_map)
  File "fontTools/varLib/cff.py", line 183, in merge_PrivateDicts
    sub_model, model_keys = vsindex_dict[vsindex]
KeyError: 0
```
[AdobeBlankVF.zip](https://github.com/fonttools/fonttools/files/3245532/AdobeBlankVF.zip)

kenlunde:
I am sure that it has something to do with its _blank_ glyphs, as the corresponding **black** VFs build successfully.

readroberts:
I can reproduce this. Am looking at it now.

readroberts:
varLib.cff.py didn't support building a variation font where none of the glyphs have any variation. I have a fix for this, will submit after more testing.

anthrotype:
Fixed by #1642

PR #1642, opened 2019-06-12
Fixes merge bug when VF source have no blends or no marking glyphs in any charstrings.

readroberts:
Fixes issue: "KeyError in varLib.cff.merge_PrivateDicts #1632".

### DIFF
--- a/Lib/fontTools/varLib/cff.py
+++ b/Lib/fontTools/varLib/cff.py
@@ -393,7 +393,7 @@ def merge_charstrings(glyphOrder, num_masters, top_dicts, masterModel):
 	# never gets built. This could still be needed if there is variation 
 	# in the PrivatDict, so we will build the default data for vsindex = 0.
 	if not vsindex_dict:
-		key = (True)*num_masters
+		key = (True,) * num_masters
 		_add_new_vsindex(model, key, masterSupports, vsindex_dict,
 			vsindex_by_key, varDataList)
 	cvData = CVarData(varDataList=varDataList, masterSupports=masterSupports,
--- a/Lib/fontTools/varLib/cff.py
+++ b/Lib/fontTools/varLib/cff.py
@@ -161,7 +161,7 @@ def merge_PrivateDicts(top_dicts, vsindex_dict, var_model, fd_map):
 		For each key, step through each relevant source font Private dict, and
 		build a list of values to blend.
 	The 'relevant' source fonts are selected by first getting the right
-	submodel using model_keys[vsindex]. The indices of the
+	submodel using vsindex_dict[vsindex]. The indices of the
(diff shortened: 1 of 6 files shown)


Lib/fontTools/varLib/cff.py
```diff
@@ -365,24 +378,24 @@ def merge_charstrings(glyphOrder, num_masters, top_dicts, masterModel):
...
-			for support in model.supports[1:]:
-				if support not in masterSupports:
-					masterSupports.append(support)
-				varTupleIndexes.append(masterSupports.index(support))
-			var_data = varLib.builder.buildVarData(varTupleIndexes, None, False)
-			vsindex = len(vsindex_dict)
-			vsindex_by_key[key] = vsindex
-			vsindex_dict[vsindex] = (model, [key])
-			varDataList.append(var_data)
+			vsindex = _add_new_vsindex(model, key,masterSupports, vsindex_dict,
```
anthrotype: whitespace
```suggestion
			vsindex = _add_new_vsindex(model, key, masterSupports, vsindex_dict,
```

Lib/fontTools/varLib/cff.py
```diff
@@ -365,24 +378,24 @@ def merge_charstrings(glyphOrder, num_masters, top_dicts, masterModel):
...
+				vsindex_by_key, varDataList)
 		# We do not need to check for an existing new_cs.private.vsindex,
 		# as we know it doesn't exist yet.
 		if vsindex != 0:
 			new_cs.program[:0] = [vsindex, 'vsindex']
+			
+	# If there is no variation in any of the charstrings, then vsindex_dict
+	# never gets built. This is could still be needed if there is variation
```
anthrotype:
```suggestion
	# never gets built. This could still be needed if there is variation 
```

Lib/fontTools/varLib/cff.py
```diff
@@ -365,24 +378,24 @@ def merge_charstrings(glyphOrder, num_masters, top_dicts, masterModel):
...
 		# We do not need to check for an existing new_cs.private.vsindex,
 		# as we know it doesn't exist yet.
 		if vsindex != 0:
 			new_cs.program[:0] = [vsindex, 'vsindex']
+			
+	# If there is no variation in any of the charstrings, then vsindex_dict
+	# never gets built. This is could still be needed if there is variation 
+	# in the PrivatDict, so we will build the default data for vsindex = 0.
+	if not vsindex_dict:
+		key = (True)*num_masters
```
anthrotype: I think you meant a tuple here:

```suggestion
		key = (True,) * num_masters
```

otherwise the value of `(True) * num_masters` is equivalent to multiplying boolean `True` (which is the same as integer `1` in python) by another integer (`num_masters`).

PR #1653, opened 2019-06-20
varLib CFF fixes

miguelsousa:


Lib/fontTools/varLib/cff.py
```diff
@@ -205,11 +205,10 @@ def merge_PrivateDicts(top_dicts, vsindex_dict, var_model, fd_map):
 				try:
 					values = [pd.rawDict[key] for pd in pds]
 				except KeyError:
-					del private_dict.rawDict[key]
```
anthrotype: are you sure you want to delete this line?

anthrotype:
I'm not sure I fully understand the issue here (especially the last commit 19f7e31), so i'd like to ask @readroberts to take a look instead.

Lib/fontTools/varLib/cff.py
miguelsousa: Yes. The VF will have the item with the values of the default master. No blends. I think that's better than dropping it entirely.

miguelsousa:
@anthrotype any chance that a new FontTools version be released this week? We need to integrate these fixes in the AFDKO.

anthrotype:
I'm busy attending the TDi programme in Reading, but I'll try to find the time later this week.

anthrotype:
https://github.com/fonttools/fonttools/releases/tag/3.43.2

miguelsousa:
@anthrotype thanks

Issue #1536, opened 2019-03-10
Optimizing fontTools.ttLib.TTFont.getGlyphID


dscorbett:
I have a project with many thousands of glyphs, which translates to thousands of calls to `fontTools.ttLib.TTFont.getGlyphID` when calling `fontTools.ttLib.TTFont.save`. I was able to make it faster (17.66&nbsp;s to 4.236&nbsp;s of cumulative time) by adding the following to the beginning of the function.
```python
		try:
			return self._reverseGlyphOrderDict[glyphName]
		except (AttributeError, KeyError):
			pass
```

The function is slow because it checks that `_reverseGlyphOrderDict` and `glyphOrder` are up to date and consistent. My version skips those checks. Is that safe? I think it is, for my use case, because I have finished creating glyphs by the time the function is first called.

Would it be possible to add something like this optimization to fontTools? If `getGlyphID` is called because `save` was called, it seems like it could assume that `_reverseGlyphOrderDict` and `glyphOrder` are frozen for the duration of `save`.

behdad:
> I was able to make it faster (17.66 s to 4.236 s of cumulative time) by adding the following to the beginning of the function.

That's super weird.  It shouldn't take that long no matter what.  Do we know which callsites are taking the most time?

behdad:
How long does the save take in total?

dscorbett:
I reran the build and have attached [the profiles](https://github.com/fonttools/fonttools/files/2959176/profiles.tar.gz). `save` took 50.375&nbsp;s in the slower build vs. 42.159&nbsp;s in the faster build. The most time-consuming callers of `getGlyphID` in the slower build were [`GlyphID.write`](https://github.com/fonttools/fonttools/blob/3.26.0/Lib/fontTools/ttLib/tables/otConverters.py#L281), [`Coverage.preWrite`](https://github.com/fonttools/fonttools/blob/3.26.0/Lib/fontTools/ttLib/tables/otTables.py#L487), and [`SingleSubst.preWrite`](https://github.com/fonttools/fonttools/blob/3.26.0/Lib/fontTools/ttLib/tables/otTables.py#L648).

anthrotype:
@dscorbett How did you create those .profile files? They don't seem to be the output of `python -m cProfile -o ...` -- I can't visualize them or extract the data.

dscorbett:
[...]

PR #1654, opened 2019-06-23
Optimize TTFont.getGlyphID

dscorbett:
If a `TTFont` has the attribute `_reverseGlyphOrderDict`, it must be consistent with `_glyphOrder`. The attribute `glyphOrder` is now a property whose setter upholds this invariant. This was suggested in https://github.com/fonttools/fonttools/issues/1536#issuecomment-472505536.

### DIFF
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -622,8 +622,8 @@ class FontBuilder(object):
         self.font["loca"] = newTable("loca")
         self.font["glyf"] = newTable("glyf")
         self.font["glyf"].glyphs = glyphs
-        if hasattr(self.font, "glyphOrder"):
-            self.font["glyf"].glyphOrder = self.font.glyphOrder
+        if hasattr(self.font, "_glyphOrder"):
+            self.font["glyf"].glyphOrder = self.font._glyphOrder
         if calcGlyphBounds:
             self.calcGlyphBounds()
 
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -1815,7 +1815,7 @@ def subset_glyphs(self, s):
 @_add_method(ttLib.getTableClass('gvar'))
 def prune_pre_subset(self, font, options):
 	if options.notdef_glyph and not options.notdef_outline:
-		self.variations[font.glyphOrder[0]] = []
+		self.variations[font.getGlyphOrder()[0]] = []
 	return True
 
 @_add_method(ttLib.getTableClass('gvar'))
--- a/Lib/fontTools/ttLib/ttFont.py
+++ b/Lib/fontTools/ttLib/ttFont.py
@@ -426,16 +426,20 @@ class TTFont(object):
 			return default
 
 	def setGlyphOrder(self, glyphOrder):
-		self.glyphOrder = glyphOrder
+		try:
+			del self._reverseGlyphOrderDict
+		except AttributeError:
+			pass
+		self._glyphOrder = tuple(glyphOrder)
 
 	def getGlyphOrder(self):
 		try:
-			return self.glyphOrder
+			return list(self._glyphOrder)
 		except AttributeError:
 			pass
 		if 'CFF ' in self:
 			cff = self['CFF ']
-			self.glyphOrder = cff.getGlyphOrder()
+			self._glyphOrder = tuple(cff.getGlyphOrder())
 		elif 'post' in self:
 			# TrueType font
 			glyphOrder = self['post'].getGlyphOrder()
@@ -447,10 +451,10 @@ class TTFont(object):
 				#
 				self._getGlyphNamesFromCmap()
 			else:
-				self.glyphOrder = glyphOrder
+				self._glyphOrder = tuple(glyphOrder)
 		else:
 			self._getGlyphNamesFromCmap()
-		return self.glyphOrder
+		return list(self._glyphOrder)
 
 	def _getGlyphNamesFromCmap(self):
 		#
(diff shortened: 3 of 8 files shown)


behdad:
Problem is when the contents of the glyph order list change though.  The property doesn't catch that.

dscorbett:
How about now? (The AppVeyor builds failed for a reason that doesn’t seem related to my changes.)

anthrotype:
this feels to me a bit over-engineered.
First, lot of code assumes that TTFont.glyphOrder is actually a plan python list, I wouldn't like to change that. If anything, you should subclass from MutableSequence or even easier from UserList (which wraps a plan list in a self.data attribute and only override the mutating methods).
Second, I don't like the cyclical reference between TTFont and this _GlyphOrderProxy object. I think the Windows Appveyor CI failures actually are related to the change, because the cyclical reference prevents or delays the objects from being garbage collected, thus maitaining the open file handle and windows disallows removing a file that's kept open by a process. To prevent this sorts of issue, one should use the weakref module.

Anyway, I am not suggesting you to do these changes. I would rather prefer a more K.I.S.S approach like the one I suggested in the other thread:

https://github.com/fonttools/fonttools/issues/1536#issuecomment-472505536

> Would it be enough to just have setGlyphOrder method be the only supported way of modifying the self.glyphOrder, and have that method invalidate (i.e. remove the attribute) the self._reverseGlyphOrderDict cache? Then the getGlyphID method can simply assume that whenever a _reverseGlyphOrderDict attribute is present, it must be in sync with the glyphOrder, and do try get glyphName from it, except rebuilding the reverse order when missing.

anthrotype:
in addition to the suggestion above, I would also have `TTFont.getGlyphOrder` always return a copy of `self.glyphOrder `(i.e. `return list(self.glyphOrder)`) to prevent any attempts at intentionally or accidentally modifying this inner `self.glyphOrder` without first going through the `TTFont.setGlyphOrder` method (which should invalidate the _reverseGlyphOrderDict).

anthrotype:
> return a copy of self.glyphOrder 

actually, perhaps that's too heavy handed.

Maybe it's ok if we document that the supported way to modify a TTFont's glyphOrder is by calling setGlyphOrder. And if one modifies it without invalidating the cached _reverseGlyphOrderDict it's undefined behavior.

Alternatively, we could make the self.glyphOrder an immutable tuple of strings, so we avoid the copy and get most of the list-like methods without worrying about it being modified outside of setGlyphOrder.
This would still break assumptions about it being a list (e.g. unittests asserting that TTFont.getGlyphOrder == ["a", "b", "c"] would fail if a tuple is returned).

dscorbett:
Currently, because `glyphOrder` doesn’t begin with an underscore, a user might get or set the field directly. Even if a user uses the methods, they might mutate the list after calling the method. I wrote the proxy class to avoid breaking the implicit API. If none of that was ever actually supported, though, I’ll rename the field `_glyphOrder` and implement it as a tuple. That means the setter will have to call `tuple`. The reason to rename it is to crash users’ code instead of letting it break subtly. Even so, it will still break subtly if they use the setter method, then mutate the original list, because that will no longer have the expected effect.

Another approach is to keep the current implicit API and add a method to explicitly freeze the glyph order.

anthrotype:
> Currently, because glyphOrder doesn’t begin with an underscore, a user might get or set the field directly. 

Direct usages of glyphOrder attribute outside of get/setGlyphOrder were never supported. I don't think even fontTools library does that internally, even though technically it could. So I am not worried too much if user code is relying on that attribute.

> Even if a user uses the methods, they might mutate the list after calling the method. 

yeah, since getGlyphOrder returns a reference to the internal mutable list. But I don't think that the intention was to allow clients to mutate the list in place. It's probably an unintentional side-effect.

I would be ok to rename the attribute _glyphOrder and make it a tuple, however since getGlyphOrder is a very commonly used method, that would be a breaking change and I'd prefer to introduce a change like that when we will bump the major version to 4.0 as we drop py2 support, hopefully sometimes next month.

anthrotype:
> a method to explicitly freeze the glyph order

what would that do exactly?

anthrotype:
(to clarify: it's not renaming to self._glyphOrder that would be break API, but the change of getGlyphOrder() return value from list to tuple object)

dscorbett:
By freezing, I meant notifying the `TTFont` that it should internally convert everything to this new API, where the glyph order is a tuple. But if version 4.0 is coming so soon, I’ll just implement it as always a tuple, and you can merge this after the last 3.\* is released.

By the way, for 4.0, could you prefix underscores to all unsupported things, or document the API policy? I’m never sure which parts of FontTools I can rely on.

behdad:
I imagine some will use the glyph order API like this:
```
  glyphOrder = []
  font.setGlyphOrder(glyphOrder)
  for glyph in ...
    glyphOrder.append(glyph)
```

It's hard to keep supporting the above.

I've got to ask, is this even a problem?  How much cost does the current check have?  My idea for optimizing this was to add mapping API that maps an entire array of glyph names to indices.  Such API can reverse the glyph-list and skip the extra checks if length of array is of a certain minimum...

dscorbett:
It’s a problem for me: I have a big font. I attached some profiles to the original issue. More recently, I ran builds without and with this patch: `TTFont.save` went from 28.32&nbsp;s to 20.85&nbsp;s, which is about the savings I get by monkey-patching it in my own project.

anthrotype:
> add mapping API that maps an entire array of glyph names to indices. Such API can reverse the glyph-list and skip the extra checks if length of array is of a certain minimum

I don't understand what @behdad exactly means here. I may be missing other details.

How about we add an option (`update=True`) to TTFont.getGlyphID() and turn that on in the hottest call sites where the method is called hundred of times, as long as we can guarantee that the glyphOrder list doesn't change and is consistent with the cached _reverseGlyphOrderDict?

Or, instead of an option, a new _getGlyphID method (used by the public getGlyphID) that just looks up _reversedGlyphOrderDict, again only used internally where we know glyphOrder won't change.

anthrotype:
> Add a getGlyphIDs that maps an array the fastest way possible.

we already have that, I think. It's called TTFont.getReversedGlyphMap()

behdad:
> > add mapping API that maps an entire array of glyph names to indices. Such API can reverse the glyph-list and skip the extra checks if length of array is of a certain minimum
> 
> I don't understand what @behdad exactly means here. I may be missing other details.

This: https://github.com/fonttools/fonttools/issues/1536#issuecomment-472581569

behdad:
Closing in preference to https://github.com/fonttools/fonttools/pull/2398

PR #1656, opened 2019-07-01
Snippet to alter sidebearings

simoncozens:


jenskutilek:
What happens to composite glyphs that reference the modifed glyphs? Or if the glyph to be modified is a composite glyph itself?

simoncozens:
See, that's why it's a draft PR. :-) Will work on it; also thinking how to work on CFF.

behdad:
Why is this a useful snippet? :)

simoncozens:
Somebody asked for it way back when. I can't remember why now.

OTOH anything which shows people how to use fonttools to do stuff to their fonts is a useful snippet.

Issue #1655, opened 2019-06-24
designspaceLib.newDefaultLocation must also make non-mapped location


LettError:
`designspaceLib.newDefaultLocation` must be able to generate a location with the actual axis defaults, with "designspace", non-mapped values. I will also review the other uses of `axisDescriptor.map_forward`.

https://github.com/fonttools/fonttools/blob/18f969ab7dca9b5ec16dcf480ffe3c46ae70beb7/Lib/fontTools/designspaceLib/__init__.py#L1128

anthrotype:
Hey Erik,

> the actual axis defaults, with "designspace", non-mapped values

well, that's what it _is_ doing.
Mapping "forward" in this context means from user to internal design coordinates.
The first column of the axes maps is "user" (the 'input'); the second column (the 'output') is internal "design" coordinates.
Also -- and that may be the source of the confusion here -- the axes minimum, default and maximum are always specified as "user"-space coordinates. That is why, when you ask "give me a new default location, in design coordinates", you first take the axis.default (always in 'user' coordinates) and map that forward to 'design' coordinate space.

anthrotype:
See also my other comment in https://github.com/googlefonts/fontmake/pull/552#issuecomment-497431705

LettError:
>  the axes minimum, default and maximum are always specified as "user"-space coordinates.
is it really? that's not how I built it originally and I'm not sure that is a good idea.

anthrotype:
yes, that's how varLib interprets them...

Quoting @behdad from https://github.com/LettError/designSpaceDocument/issues/16

> One thing that still bothers me is that the axis default / minimum / maximum values are in the user space, whereas everything else is by default in the design space.

and then, in another later comment, he added

> I think axis default/min/max should default to "input" space. Ie. they define the user-visible range, and the mapping defines how to convert that to internal representation. For masters and instances allowing both would be great.

LettError:
That is not always convenient.

madig:
(FWIW, I stumbled over this, too, and added clarifications to https://github.com/fonttools/fonttools/tree/master/Doc/source/designspaceLib#axisdescriptor-object)

LettError:
Great I think google is happy then.

anthrotype:
> That is not always convenient. And there is a lot of data that does not work like that.

this is not good, and I do share your concern. We need to come up together with a reasonable plan to make so that these location coordinates are set explicitly and unambiguously as either user-space or design-space.

[...]

Issue #1631, opened 2019-06-03
Default master coordinates cannot hit zero while normalized


yarmola:
I have an issue with rounding the default master location when axis map is present and that map does not have a point for the default master. I know that's wrong, but I don't think current designSpace doc requires to map default location.

Perfectly integer design coordinate of the default master maps to some floating-point user coordinate which then is recorded (in slightly rounded form, 4 points after dot) as axis default.

It will never get back to normalized zero, which will make this piece of code fail:

<img width="461" alt="Screenshot 2019-06-03 at 03 21 24" src="https://user-images.githubusercontent.com/407038/58769450-d755b480-85af-11e9-8609-abe684e3ca0a.png">

(which is what is happening here)

I will add a work-around (by inserting "rounded" default location to the map), but I think it will be good to 1) add a point to the designSpace document asking to map default location and 2) maybe replace that  == 0 with something that is more floating-point compatible, so it will tolerate "almost zero values" (0.0001213, etc.).

anthrotype:
I think we should update the doc to say that a mapping for the default master (as well as for minimum and maximum) is required if an axis has any map elements.

can you show what is your designspace axes map configuration?

I don't thin we need to do any rounding. A default master is currently required to be present in a DesignSpace to build a varfont, and that master needs to be given an explicit mapping if for a given axis the user vs design locations are different.

yarmola:
I agree that designSpace spec should include this addition (which is very logical, as "implied" map contains min, def, max values by definition). 

My current fix is doing exactly that: adding the point for the default master to the map. It solves the problem.

PR #1658, opened 2019-07-04
 Write something about the different requirements of DS consumers 

madig:
As noted in

- https://github.com/fonttools/fonttools/issues/1655#issuecomment-505388564
- https://github.com/fonttools/fonttools/issues/1631

Also fix the code sample at the beginning of the specification.

Todo:
- Cross-check information with information from http://letterror.com/dev/mutmath_varmod/ and https://github.com/googlefonts/fontmake/issues/264


madig:
I think this should be part of a larger spec refactoring/updating, closing for now.

Issue #1660, opened 2019-07-10
fontBuilder minimal example does not work on windows


massaad:
I am just starting with font generation, and I am trying to generate a font through code, so fontBuilder seemed like the ideal solution for me. Unfortunately, I am not sure how to start... using the minimal example included in the code is not working for me on windows. Even though the font files are created, If I try to view it, I get a "The requested file sample.otf is not a valid font file."

Are there any steps to get that code generate a minimal font that is viewable?

Thanks.

miguelsousa:
I suspect that your font doesn't have one of the `name` table IDs required by Windows.
My colleague @cjchapman recently did some tests on that and these were his observations:

> Windows — `name`  IDs 1,2,3,4,6 required (platform=Windows)
> Mac — `name` IDs 1,2,3,4,6 required (platform=Mac, if *any* Mac IDs, otherwise uses Windows IDs)
> variable fonts — `name` IDs 16 & 17 required per MS
>
> Note that in my own experiments in Windows 10, it seems that only `name` IDs 1,2,3,4 are required (not 6).

justvanrossum:
Yeah, it seems the minimal example in the code misses 3 (`uniqueFontIdentifier`) and 4 (`fullName`).

Untested:

```python
from fontTools.fontBuilder import FontBuilder
from fontTools.pens.ttGlyphPen import TTGlyphPen

def drawTestGlyph(pen):
    pen.moveTo((100, 100))
    pen.lineTo((100, 1000))
    pen.qCurveTo((200, 900), (400, 900), (500, 1000))
    pen.lineTo((500, 100))
    pen.closePath()

fb = FontBuilder(1024, isTTF=True)
fb.setupGlyphOrder([".notdef", ".null", "A", "a"])
fb.setupCharacterMap({65: "A", 97: "a"})

advanceWidths = {".notdef": 600, "A": 600, "a": 600, ".null": 600}

familyName = "HelloTestFont"
styleName = "TotallyNormal"
nameStrings = dict(familyName=dict(en="HelloTestFont", nl="HalloTestFont"),
                   styleName=dict(en="TotallyNormal", nl="TotaalNormaal"),
                   uniqueFontIdentifier="ABC; HelloTestFont TotallyNormal",
                   fullName="HelloTestFont TotallyNormal")
nameStrings['psName'] = familyName + "-" + styleName

pen = TTGlyphPen(None)
drawTestGlyph(pen)
glyph = pen.glyph()
[...]

PR #1661, opened 2019-07-11
make the minimal examples work on windows 10

massaad:
Partially addresses #1660. This makes the otf minimal example work. The ttf example still doesn't work.

### DIFF
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -31,6 +31,7 @@ Here is how to build a minimal TTF:
 from fontTools.fontBuilder import FontBuilder
 from fontTools.pens.ttGlyphPen import TTGlyphPen
 
+
 def drawTestGlyph(pen):
     pen.moveTo((100, 100))
     pen.lineTo((100, 1000))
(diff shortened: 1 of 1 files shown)


justvanrossum:
Looks good to me. Do we want to keep this open until we find the problem with the ttf example or merge it now?

massaad:
I will spend a day or two trying to figure out what's wrong with the ttf example, but I am really no expert at all with fonts... just beginning. So, any help would be appreciated :-)

miguelsousa:
@massaad try running [FontValidator](https://github.com/HinTak/Font-Validator) or [Font Bakery](https://github.com/googlefonts/fontbakery) on the TTF.

massaad:
Now both samples (otf and ttf) work in windows 10. This is ready for merging. Thanks.

justvanrossum:
Thanks! Looks good to me, although I would prefer the two examples to be as similar as possible, eg. the name strings and OS/2 values should probably be the same for OTF and TTF. Is that something you would be willing to improve?

massaad:
They should be matching now, as much as I know how.

justvanrossum:
Perfect, thanks!

PR #1668, opened 2019-07-23
Create scale-font.py

BlackFoundry:
Script to change UnitsPerEm of a Font
USAGE:
scale-font.py path/to/inputFont.ttf targetUPM path/to/outputFont.ttf

Only works on .ttf (no CFF) for now

### DIFF
--- a/Snippets/scale-font.py
+++ b/Snippets/scale-font.py
@@ -4,42 +4,42 @@ USAGE:
 scale-font.py path/to/inputFont.ttf targetUPM path/to/outputFont.ttf
 
 Only works on .ttf (no CFF) for now
+
+TO DO: add support for scaling GPOS lookups type 7 and 8
 """
 
 from fontTools import ttLib
 from sys import argv
+from fontTools.misc.fixedTools import otRound
 
-def scaleValue_toUPM(value, UPM, target_UPM):
-    if value < 0.0 :
-        return int ( ( value * target_UPM -0.5 ) / UPM )
-    else:
-        return int ( ( value * target_UPM +0.5 ) / UPM )
+def scale_value_fromUPM_toUPM(value, UPM, target_UPM):
+    return otRound((value * target_UPM)/UPM)
 
-def scaleLookupType1(subTable, UPM, target_UPM):
+def _scaleLookupType1(subTable, UPM, target_UPM):
     if subTable.Format == 1:
         for k, v in subTable.Value.__dict__.items():
             if k == 'XAdvance':
-                subTable.Value.XAdvance = scaleValue_toUPM(v, UPM, target_UPM)
+                subTable.Value.XAdvance = scale_value_fromUPM_toUPM(v, UPM, target_UPM)
             elif k == 'YAdvance':
-                subTable.Value.YAdvance = scaleValue_toUPM(v, UPM, target_UPM)
+                subTable.Value.YAdvance = scale_value_fromUPM_toUPM(v, UPM, target_UPM)
             elif k == 'XPlacement':
-                subTable.Value.XPlacement = scaleValue_toUPM(v, UPM, target_UPM)
+                subTable.Value.XPlacement = scale_value_fromUPM_toUPM(v, UPM, target_UPM)
             elif k == 'YPlacement':
-                subTable.Value.YPlacement = scaleValue_toUPM(v, UPM, target_UPM)                  
+                subTable.Value.YPlacement = scale_value_fromUPM_toUPM(v, UPM, target_UPM)                  
     if subTable.Format == 2:
         for value in subTable.Value:
             for k, v in value.__dict__.items():
                 if k == 'XAdvance':
-                    value.XAdvance = scaleValue_toUPM(v, UPM, target_UPM)
+                    value.XAdvance = scale_value_fromUPM_toUPM(v, UPM, target_UPM)
                 elif k == 'YAdvance':
-                    value.YAdvance = scaleValue_toUPM(v, UPM, target_UPM)
+                    value.YAdvance = scale_value_fromUPM_toUPM(v, UPM, target_UPM)
                 elif k == 'XPlacement':
-                    value.XPlacement = scaleValue_toUPM(v, UPM, target_UPM)
+                    value.XPlacement = scale_value_fromUPM_toUPM(v, UPM, target_UPM)
                 elif k == 'YPlacement':
-                    value.YPlacement = scaleValue_toUPM(v, UPM, target_UPM)
+                    value.YPlacement = scale_value_fromUPM_toUPM(v, UPM, target_UPM)
 
 
-def scaleLookupType2(subTable, UPM, target_UPM):
+def _scaleLookupType2(subTable, UPM, target_UPM):
     # format 1: Single Pairs
     if subTable.Format == 1:
         for pairSet in subTable.PairSet:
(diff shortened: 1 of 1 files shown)


Snippets/scale-font.py
```diff
@@ -0,0 +1,273 @@
...
+"""
+
+from fontTools import ttLib
+from sys import argv
+
+def scaleValue_toUPM(value, UPM, target_UPM):
+    if value < 0.0 :
+        return int ( ( value * target_UPM -0.5 ) / UPM )
+    else:
+        return int ( ( value * target_UPM +0.5 ) / UPM )
```
anthrotype: please use `otRound` from `fontTools.misc.fixedTools` to round the scaled value. Import it at the top

Snippets/scale-font.py
```diff
@@ -0,0 +1,273 @@
...
+USAGE:
+scale-font.py path/to/inputFont.ttf targetUPM path/to/outputFont.ttf
+
+Only works on .ttf (no CFF) for now
+"""
+
+from fontTools import ttLib
+from sys import argv
+
+def scaleValue_toUPM(value, UPM, target_UPM):
```
anthrotype: i don't like the mixing of camelCase and snake_case, try to stick to one or the other (I prefer the latter, but in fontTools there's probably more of the former).

Snippets/scale-font.py
```diff
@@ -0,0 +1,273 @@
...
+    tt['hhea'].ascent = scaleValue_toUPM(tt['hhea'].ascent, UPM, target_UPM)
+    tt['hhea'].descent = scaleValue_toUPM(tt['hhea'].descent, UPM, target_UPM)
+    tt['hhea'].lineGap = scaleValue_toUPM(tt['hhea'].lineGap, UPM, target_UPM)
+    tt['hhea'].advanceWidthMax = scaleValue_toUPM(tt['hhea'].advanceWidthMax, UPM, target_UPM)
+    tt['hhea'].minLeftSideBearing = scaleValue_toUPM(tt['hhea'].minLeftSideBearing, UPM, target_UPM)
+    tt['hhea'].minRightSideBearing = scaleValue_toUPM(tt['hhea'].minRightSideBearing, UPM, target_UPM)
+    tt['hhea'].xMaxExtent = scaleValue_toUPM(tt['hhea'].xMaxExtent, UPM, target_UPM)
+    #################################
+    # print tt['hhea'].numberOfHMetrics #Number of entries of the hmtx table
+    # print len(glyphOrder)
```
anthrotype: please get rid of stray commented-out lines like these

Snippets/scale-font.py
```diff
@@ -0,0 +1,273 @@
...
+            mark2Anchor.XCoordinate = scaleValue_toUPM(mark2Anchor.XCoordinate, UPM, target_UPM)
+            mark2Anchor.YCoordinate = scaleValue_toUPM(mark2Anchor.YCoordinate, UPM, target_UPM)
+
+def scaleLookupType7(subTable, UPM, target_UPM):
+    pass 
+    
+def scaleLookupType8(subTable, UPM, target_UPM):
+    pass        
+
+def scaleFont(tt, target_UPM):
```
anthrotype: `tt` is not the most readable name for a variable that holds an instance of TTFont. Use `font` or `ttFont`

Snippets/scale-font.py
```diff
@@ -0,0 +1,273 @@
...
+    inputPath = argv[1]
+    targetUPM = int(argv[2])
+    outputPath = argv[3]
+
+    inputFont = ttLib.TTFont(inputPath)
+    scaleFont(inputFont, targetUPM)
+    inputFont.save(outputPath)
+
+def argsCheck():
+    return(argv)
```
anthrotype: this code is no-op and looks incomplete. clean up or remove

Snippets/scale-font.py
```diff
@@ -0,0 +1,273 @@
...
+    for mark1Record in subTable.Mark1Array.MarkRecord:
+        mark1Record.MarkAnchor.XCoordinate = scaleValue_toUPM(mark1Record.MarkAnchor.XCoordinate, UPM, target_UPM)
+        mark1Record.MarkAnchor.YCoordinate = scaleValue_toUPM(mark1Record.MarkAnchor.YCoordinate, UPM, target_UPM)
+    for mark2Record in subTable.Mark2Array.Mark2Record:
+        for mark2Anchor in mark2Record.Mark2Anchor:
+            mark2Anchor.XCoordinate = scaleValue_toUPM(mark2Anchor.XCoordinate, UPM, target_UPM)
+            mark2Anchor.YCoordinate = scaleValue_toUPM(mark2Anchor.YCoordinate, UPM, target_UPM)
+
+def scaleLookupType7(subTable, UPM, target_UPM):
+    pass
```
anthrotype: what's the use of these?

justvanrossum:
@anthrotype:
> Another way would be doing like the subset or merge module do, i.e. dynamically assign methods to existing classes defined elsewhere using decorators (but it's ok at this stage if keep it simple).

I'm not a fan of that method (pun intended), and I wonder: wouldn't `functools.singledispatch()` be a better approach these days?

anthrotype:
I also like singledispatch

behdad:
> How about, instead of passing UPM and target_UPM around as fuction arguments, you make a UpemScaler class that holds them as attributes, then make those `scale*` functions into instance methods of that class, each taking a table or subtable object as a parameter?
> Another way would be doing like the subset or merge module do, i.e. dynamically assign methods to existing classes defined elsewhere using decorators (but it's ok at this stage if keep it simple).

Right.  I also think something using visitor pattern and list of fields to rescale...  Def not acceptable all the repetition.

jenskutilek:
Maybe out of scope for this, but what happens if the font is hinted? At least the cvt and cvar table values (all of them?) would need to be scaled as well. Other hinting instructions may also refer to font unit measurements.

behdad:
Yeah hinting cannot uséfully be scaled.  Same about cvt.

anthrotype:
I made a script that uses [fontMath](https://github.com/robotools/fontMath) to scale a whole UFO and change its unitsPerEm, I put it here in case other may find it useful:
https://gist.github.com/anthrotype/62d0bfe1d38b8f11a199bd3b66574bcc

Of course that only works if one has a UFO source to work with, so it doesn't completely replace a tool for scaling a binary font's UPEM. But working on the sources does make things a bit easier.

JeremieHornus:
Nice! Compact and clean, will be very useful when sources UFO are available!
I am not forgetting the scale-font project for TTF fonts but need to find and take the time for finishing it.

Snippets/scale-font.py
JeremieHornus: yes I quite like snake_case as well to list the arguments implicitly

Snippets/scale-font.py
JeremieHornus: thanks!

Snippets/scale-font.py
JeremieHornus: contextual positioning: 
https://docs.microsoft.com/en-us/typography/opentype/spec/gpos#lookup-type-7-contextual-positioning-subtables

quite rare in fonts I came across (never seen it used yet)

Snippets/scale-font.py
JeremieHornus: okay!

Snippets/scale-font.py
JeremieHornus: right

Snippets/scale-font.py
JeremieHornus: probably need more details for this specific comment, can you specify?

JeremieHornus: ah yes ! I se what you mean… ;-)

behdad:
This should be done similar to merge.py.  I'm putting together design for a unified "processor" that should also be used for this.

behdad:
> I'm not a fan of that method (pun intended), and I wonder: wouldn't `functools.singledispatch()` be a better approach these days?

Interesting.  I'll check that out.

behdad:
> > I'm not a fan of that method (pun intended), and I wonder: wouldn't `functools.singledispatch()` be a better approach these days?
> 
> Interesting. I'll check that out.

I'll use it.  It's essentially the same as what I'm doing, just implemented in the standard library now.  So I'll embrace it.

behdad:
> I'll use it. It's essentially the same as what I'm doing, just implemented in the standard library now. So I'll embrace it.

FWIW in varLib I was doing very similar.  Ie. not adding methods to the classes, but registering them for dispatch somewhere else.  It's *very* similar to singledispatch.

behdad:
> FWIW in varLib I was doing very similar. Ie. not adding methods to the classes, but registering them for dispatch somewhere else. It's _very_ similar to singledispatch.

Humm.  I don't think I can reuse it though, since we need inheritance in a way that singledispatch doesn't allow.  If you check `fontTools.varLib.merger` you see what I mean, eg. `AligningMerger` is inherited by others.

The module I'm proposing is a similar to singledispatch implemented for our fonts; ie. it uses more information than just that.  I'll review singledispatch to inform my design.  I'll create a new request for discussing the design and will link from here.

BlackFoundry:
this all sounds good @behdad, in the meantime I have made some improvements to my base

anthrotype:
@behdad I like the idea of centralizing all the code in subset/merge/varLib that deals with traversing a font object hierarchy to do interesting stuff with it.

> I don't think I can reuse it though, since we need inheritance in a way that singledispatch doesn't allow.

Python 3.8 has `functools.singledispatchmethod` to make single-dispatch generic instance methods which should work with inheritance
https://docs.python.org/3/library/functools.html#functools.singledispatchmethod
There's also a backport for pythons < 3.8 [on PyPI](https://pypi.org/project/singledispatchmethod)

A separate project called [multipledispatch](https://github.com/mrocklin/multipledispatch/) also allows to do the dispatching on instance methods with inheritance, and has a nicer API than built-in singledispatch; plus, as the name tells, it can displatch on multiple arguments' type, not just one (though maybe we don't need that). Though I don't know if we want to add a third party dep for this. Still worth having a look.

Ok we are digressing. We should discuss this in a separate issue, sorry Jeremie :)

anthrotype:
just a quick correction to my previous post: singledispatchmethod does _not_ play well with inheritance https://bugs.python.org/issue36457
So we either use mutilpledispatch or whatever behdad comes up with.

JeremieHornus:
It's ok Cosimo :) it is interesting stuff! I hope my base will be of some help for this upcoming task. It was a time-consuming reverse-engineering task to get the relevant scalable data parsed in the tables and in GPOS lookup in particular.

anthrotype:
definitely, thanks for that @JeremieHornus 👍

behdad:
I've finally sketched a `fontTools.misc.visitor` and how this scale-font would look like using it, so far just the GPOS part:
https://github.com/fonttools/fonttools/pull/2718/files
```

@ScalerVisitor.register(otTables.ValueRecord)
def visit(visitor, obj):
    attrs = ['XAdvance', 'YAdvance', 'XPlacement', 'YPlacement']
    for attr in attrs:
        v = getattr(obj, attr, None)
        if v is not None:
            v = visitor.scale(v)
            setattr(obj, attr, v)
    return False

@ScalerVisitor.register(otTables.Anchor)
def visit(visitor, obj):
    attrs = ['XCoordinate', 'YCoordinate']
    for attr in attrs:
        v = getattr(obj, attr)
        v = visitor.scale(v)
        setattr(obj, attr, v)
    return False
```

That's a lot less code than writing custom code for each lookup type.

behdad:
Closing in favor of the other PR, which handles variable-fonts as well:
https://github.com/fonttools/fonttools/pull/2718#issuecomment-1215739512

PR #1669, opened 2019-07-23
[cffLib specializer] Fix bug in programToCommands with CFF2 charstrings

readroberts:
When checking if there was an initial width argument, old logic did not correctly sum the number of blend and non-blend arguments before the first hint or moveto operator.

Added test case. Old logic dropped the arguments for the hmoveto in the last glyph.

### DIFF
--- a/Lib/fontTools/cffLib/specializer.py
+++ b/Lib/fontTools/cffLib/specializer.py
@@ -37,14 +37,28 @@ def programToCommands(program, getNumRegions=None):
 	number of regions. 'getNumRegions' takes a single argument, vsindex. If
 	the vsindex argument is None, getNumRegions returns the default number
 	of regions for the charstring, else it returns the numRegions for
-	the vsindex."""
+	the vsindex.
+	The Charstring may or may not start with a width value. If the first
+	non-blend operator has an odd number of arguments, then the first argument is
+	a width, and is popped off. This is complicated with blend operators, as
+	there may be more than one before the first hint or moveto operator, and each
+	one reduces several arguments to just one list argument. We have to sum the
+	number of arguments that are not part of the blend arguments, and all the
+	'numBlends' values. We could instead have said that by definition, if there
+	is a blend operator, there is no width value, since CFF2 Charstrings don't
+	have width values. I discussed this with Behdad, and we are allowing for an
+	initial width value in this case because developers may assemble a CFF2
+	charstring from CFF Charstrings, which could have width values.
+	"""
 
-	width = None
 	seenWidthOp = False
 	vsIndex = None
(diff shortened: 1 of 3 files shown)


anthrotype:
possibly stupid question, I haven't checked the spec: can Type2 charstrings from CFF 1.0 table contain blend operators? I thought that was CFF2 only.

Lib/fontTools/cffLib/specializer.py
```diff
@@ -37,14 +37,29 @@ def programToCommands(program, getNumRegions=None):
...
+	non-blend operator has an odd number of arguments, then the first argument is
+	a width, and is popped off. This is complicated with blend operators, as
+	there may be more than one before the first hint or moveto operator, and each
+	one reduces several arguments to just one list argument. We have to sum the
+	number of arguments that are not part of the blend arguments, and all the
+	'numBlends' values. We could instead have said that by definition, if there
+	is a blend operator, there is no width value, since CFF2 Charstrings don't
+	have width values. I discussed this with Behdad, and we are allowing for an
+	initial width value in this case because developers may assemble a CFF2
+	charstring from CFF Charstrings, which coudl have width values.
```
anthrotype:
```suggestion
	charstring from CFF Charstrings, which could have width values.
```

readroberts: You are correct.No blend operators in Type1, no width operators in CFF2 char strings. However,  Behdad thought that the logic should survive data from one type being in the others, specifically in the case where someone is converting CFF to CFF2.

Lib/fontTools/cffLib/specializer.py
```diff
@@ -37,14 +37,29 @@ def programToCommands(program, getNumRegions=None):
...
+	"""
 
 	width = None
```
cjchapman: This initialization of `width` is now superfluous and can be removed.

Lib/fontTools/cffLib/specializer.py
readroberts: @behdad @anthrotype Actually, the conversation I referred to is quite some time ago. Now I think I would disallow a CFF2 Charstring having a width arg. I would keep the same logic, but use it to post an assert that would be more informative that the much more obscure error that you get otherwise. Comments?

readroberts: Fixed the typo. I also had a brief conversation with Behdad today about the width value. He doesn't have a strong opinion either way, but has a mild preference for retaining the original intent, and not move towards specializer.py be a specification enforcer.

readroberts:
@anthrotype @behdad I think this is ready to go, just waiting for one of you to review and merge.

Issue #1670, opened 2019-07-26
Invalid RegionAxisCount in varLib for single master VF's


underware:
Compiling a single master designspace (where the master is located at the default of the axis) generates a VF which is not sanitizing. Duplicating this master solves the problem, but is not practical. 
See also  https://github.com/googlefonts/fontmake/issues/565

behdad:
Not sanitizing with what?  It's *their* bug :|

behdad:
Using a null offset to varStore should also be supported IMO...

PR #1671, opened 2019-07-26
varLib: don't build empty HVAR or gvar tables

anthrotype:
When asked to build a "variable" font with only one source master, there's no variation data, we can skip building HVAR and gvar tables. The fvar and STAT table are still built.

Fixes https://github.com/fonttools/fonttools/issues/1670
Also see https://github.com/googlefonts/fontmake/issues/565

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -311,15 +311,13 @@ def _add_gvar(font, masterModel, master_ttfs, tolerance=0.5, optimize=True):
 
 	log.info("Generating gvar")
 	assert "gvar" not in font
-	gvar = font["gvar"] = newTable('gvar')
-	gvar.version = 1
-	gvar.reserved = 0
-	gvar.variations = {}
 
 	glyf = font['glyf']
 
 	# use hhea.ascent of base master as default vertical origin when vmtx is missing
 	defaultVerticalOrigin = font['hhea'].ascent
+
+	variations = {}
 	for glyph in font.getGlyphOrder():
 
 		isComposite = glyf[glyph].isComposite()
@@ -339,7 +337,6 @@ def _add_gvar(font, masterModel, master_ttfs, tolerance=0.5, optimize=True):
 		del allControls
 
 		# Update gvar
-		gvar.variations[glyph] = []
 		deltas = model.getDeltas(allCoords)
 		supports = model.supports
 		assert len(deltas) == len(supports)
@@ -378,7 +375,14 @@ def _add_gvar(font, masterModel, master_ttfs, tolerance=0.5, optimize=True):
 					if optimized_len < unoptimized_len:
 						var = var_opt
 
-			gvar.variations[glyph].append(var)
+			variations.setdefault(glyph, []).append(var)
+
+	if variations:
+		gvar = font["gvar"] = newTable('gvar')
+		gvar.version = 1
+		gvar.reserved = 0
+		gvar.variations = variations
+
 
 def _remove_TTHinting(font):
 	for tag in ("cvar", "cvt ", "fpgm", "prep"):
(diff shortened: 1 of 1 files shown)


behdad:
I disagree.  If font is deemed variable (ie. has fvar, axes..) then if HVAR is missing, one should load glyph to get any *possible* advance variation.  That's extra cost.  Empty HVAR should be encoded IMO.

anthrotype:
ok.. I tried that first, but I couldn't figure out why the VarRegionList.RegionAxisCount was been set to 0 upon compilation, even though it was correctly set to 1 (the fvar.axisCount) by buildVarRegionList.
The mismatch only seems to happen when VarRegionList.Region is empty, i.e. there are no supports because only one master is provided.

@behdad can you revert and fix up as necessary?

behdad:
> ok.. I tried that first, but I couldn't figure out why the VarRegionList.RegionAxisCount was been set to 0 upon compilation, even though it was correctly set to 1 (the fvar.axisCount) by buildVarRegionList.
> The mismatch only seems to happen when VarRegionList.Region is empty, i.e. there are no supports because only one master is provided.

I don't understand.  What's wrong with RegionAxisCount being 0?  That sounds legit to me, but yeah need to double-check.

> @behdad can you revert and fix up as necessary?

Not soon.

anthrotype:
apparently, as @underware reported in https://github.com/googlefonts/fontmake/issues/565, the opentype-sanitizer is flagging that as an error, as the spec says it should match the fvar.axisCount. At least that's my understanding.

behdad:
> I don't understand. What's wrong with RegionAxisCount being 0? That sounds legit to me, but yeah need to double-check.

Okay.  I see what you mean / what's going on.  Let me take a look.  No promise I can fix immediately.

behdad:
> the opentype-sanitizer

Please say OTS if that's what you mean.  There's no "the opentype-sanitizer".

anthrotype:
> a null offset to varStore should also be supported IMO...

No luck. I get this error if I set HVAR.VarStore to None:

```
ERROR: Variations: Invalid variation region list offset
ERROR: HVAR: Failed to parse item variation store
ERROR: Dropping all Variation tables
ERROR: HVAR: Discarding Variations table
ERROR: Table discarded
ERROR: fvar: Discarding Variations table
ERROR: Table discarded
File sanitized successfully!
```

anthrotype:
the problem is that VarRegionList.RegionAxisCount is a "propagated count" field which is set to the length of VarRegion.VarRegionAxis array; however, when VarRegionList contains no regions, the default value for RegionAxisCount is 0 (whereas OTS expects it to always be equal to fvar.axisCount, whether or not the VarRegionList is empty).

I am not sure how to ovverride this default behavior (in otBase.BaseTable.compile) for VarRegionList, so that even when compiling a virtually empty HVAR with no regions (e.g. for single-master VF) we get a RegionAxisCount that always matches fvar.axisCount, to appease OTS.

For the time being, I'm inclined to revert my PR and roll back to the original behavior, and file an issue with OTS.

behdad:
> the problem is that VarRegionList.RegionAxisCount is a "propagated count" field which is set to the length of VarRegion.VarRegionAxis array; however, when VarRegionList contains no regions, the default value for RegionAxisCount is 0 (whereas OTS expects it to always be equal to fvar.axisCount, whether or not the VarRegionList is empty).

Yes that's what I thought...  Sucks that OTS enforces such stupid constraints.  I mean, sucks that spec requires those....

Adding a preWrite might do the trick...

anthrotype:
@behdad I fixed OTS with https://github.com/khaledhosny/ots/pull/192 -- do we still want to do that preWrite trick?

behdad:
I think yes, to comply with spec and older software.

behdad:
I will take a look.

behdad:
> I will take a look.

I don't think I can bring myself to hack on this anything soon :(.

Issue #684, opened 2016-10-01
Problem with OpenType bitmap font (Ambrosia.otb)


lemzwerg:
[This zipped SFNT bitmap-only font](https://github.com/fonttools/fonttools/files/504565/Ambrosia.zip), created a long time ago by George Williams with FontForge, can't be correctly disassembled with current git (1eba7b3).
I get (with paths shortened for readability)
```
Dumping "Ambrosia.otb" to "Ambrosia.ttx"...
Dumping 'GlyphOrder' table...
Dumping 'head' table...
Dumping 'hhea' table...
Dumping 'maxp' table...
Dumping 'OS/2' table...
Dumping 'hmtx' table...
Dumping 'cmap' table...
WARNING: corrupt 'loca' table, or wrong numGlyphs in 'maxp': 0 250
Dumping 'loca' table...
ERROR: An exception occurred during the decompilation of the 'glyf' table
Traceback (most recent call last):
  File ".../fontTools/ttLib/__init__.py", line 418, in __getitem__
    table.decompile(data, self)
  File ".../fontTools/ttLib/tables/_g_l_y_f.py", line 62, in decompile
    if len(data) - next >= 4:
UnboundLocalError: local variable 'next' referenced before assignment
Dumping 'glyf' table...
Dumping 'name' table...
Dumping 'post' table...
Dumping 'EBDT' table...
Dumping 'EBLC' table...
Dumping 'EBSC' table...
Dumping 'GDEF' table...
Dumping 'GPOS' table...
Dumping 'GSUB' table...
```

The `glyf` table is empty, and ttx doesn't like this while assembling.

```
Parsing 'GlyphOrder' table...
Parsing 'head' table...
Parsing 'hhea' table...
Parsing 'maxp' table...
Parsing 'OS/2' table...
Parsing 'hmtx' table...
Parsing 'cmap' table...
Parsing 'loca' table...
Parsing 'glyf' table...
Parsing 'name' table...
Parsing 'post' table...
Parsing 'EBDT' table...
Parsing 'EBLC' table...
Parsing 'EBSC' table...
Parsing 'GDEF' table...
Parsing 'GPOS' table...
Parsing 'GSUB' table...
ERROR: Unhandled exception has occurred
Traceback (most recent call last):
  File ".../fontTools/ttx.py", line 381, in main
    process(jobs, options)
  File ".../fontTools/ttx.py", line 355, in process
    action(input, output, options)
[...]

PR #1680, opened 2019-08-03
[ttLib] fix UnboundLocalError for empty loca/glyph tables

justvanrossum:
This partially adresses #684.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -55,28 +55,28 @@ class table__g_l_y_f(DefaultTable.DefaultTable):
 
 	def decompile(self, data, ttFont):
 		loca = ttFont['loca']
-		last = int(loca[0])
+		pos = int(loca[0])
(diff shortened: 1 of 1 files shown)


Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -59,6 +59,7 @@ def decompile(self, data, ttFont):
 		noname = 0
 		self.glyphs = {}
 		self.glyphOrder = glyphOrder = ttFont.getGlyphOrder()
+		next = 0
 		for i in range(0, len(loca)-1):
```
anthrotype: Maybe you can rename the local variable nextOffset to avoid shadowing the built-in next() function. I know it is not used in this scope but syntax highlighters aren't that smart and the blue coloured label is sometimes confusing/distracting.

justvanrossum: Agreed. I renamed `last` to `pos`, so `next` became `nextPos`.

PR #1681, opened 2019-08-03
[ttLib] allow the glyf table to be incomplete when dumping to XML

justvanrossum:
Allow the glyf table to be incomplete — to not contain all glyphs listed in the glyph order — when writing to XML. Partially addresses #684.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -140,7 +140,6 @@ class table__g_l_y_f(DefaultTable.DefaultTable):
 		for glyphName in glyphNames:
 			if glyphName not in self:
 				log.warning("glyph '%s' does not exist in glyf table", glyphName)
-				log.warning
 				continue
 			glyph = self[glyphName]
 			if glyph.numberOfContours:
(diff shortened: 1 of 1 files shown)


anthrotype:
hm.. instead of `continue` shouldn't you dump an empty TTGlyph element instead?

anthrotype:
what happens when loading fromXML where not all glyphs from GlyphOrder are actually present in the glyf table dump? You end up with an incomplete glyf table? 
I think there is an implicit guarantee that the glyf table keys and ttFont.getGlyphOrder() be the same, at least after loading a font from file (whether binary or xml). 
There's lots of code around that loops over `for glyphName in ttFont.getGlyphOrder(): glyph = ttFont["glyf"][glyphName]`

anthrotype:
oh I see now, you want to support #684...

justvanrossum:
Well, to some extent. It allows the font from #684 to be dumped, but not recompiled. I'm not sure we need to support either, actually. I was just thinking it doesn't harm us if we allow an incomplete glyf table to be dumped, even if it's broken according to the spec.

anthrotype:
ok. how about at least logging a warning for the missing glyphs?

justvanrossum:
> how about at least logging a warning for the missing glyphs?

Good idea, done. Let me know if you want the warning to be worded differently.

PR #1683, opened 2019-08-05
Clarify InstanceDescriptor lib semantics

madig:


madig:
Actually, wait with this.

PR #1684, opened 2019-08-08
Make glyphSet param consistently optional for pens

baileyparker:
The following pens have `glyphSet` as a required param to their constructor:

  - `ControlBoundsPen`, `BoundsPen`
  - `CocoaPen`
  - `PointInsidePen`
  - `QtPen`
  - `ReportLabPen`
  - `SVGPathPen`
  - `T2CharStringPen`
  - `TTGlyphPen`
  - `WxPen`

All other pens make it optional (defaulting to None):

  - `AreaPen`
  - `FilterPen`, `ContourFilterPen`, `ReverseContourPen`
  - `MomentsPen`
  - `PerimeterPen`
  - `PointToSegmentPen`, `SegmentToPointPen`, `GuessSmoothPointPen`,
    `ReverseContourPointPen`
  - `RecordingPen`, `DecomposingRecordingPen`
  - `StatisticsPen`
  - `TeePen`
  - `TransformPen`

Of the above pens that require `glyphSet` (to be passed into the parent constructor for `DecomposingPen`--a superclass of `BasePen`), only `TTGlyphPen` actually uses `self.glyphSet` and `addComponent` (a method added by `DecomposingPen`). Several pens override `addComponent` or call it on pens they own (such as `TeePen`, which calls it on all of its subpens).

The fact that so many pens (which on the surface seem like they may need to support `addComponent`--like `AreaPen`, `StatisticsPen`, etc.) make `glyphSet` optional indicates that it should be safe to make it more uniformly optional. This way, it isn't so cumbersome to construct an `SVGPathPen`, for example.

Note that `PointInsidePen` cannot have `glyphSet` made optional, because it is followed by a required param:

    def __init__(self, glyphSet, testPoint, evenOdd=False):

Reordering the params to accommodate would be a breaking change.

### DIFF
--- a/Lib/fontTools/pens/boundsPen.py
+++ b/Lib/fontTools/pens/boundsPen.py
@@ -22,7 +22,7 @@ class ControlBoundsPen(BasePen):
 	If 'ignoreSinglePoints' is True, single points are ignored.
 	"""
 
-	def __init__(self, glyphSet, ignoreSinglePoints=False):
+	def __init__(self, glyphSet=None, ignoreSinglePoints=False):
 		BasePen.__init__(self, glyphSet)
 		self.ignoreSinglePoints = ignoreSinglePoints
 		self.init()
--- a/Lib/fontTools/pens/cocoaPen.py
+++ b/Lib/fontTools/pens/cocoaPen.py
@@ -8,7 +8,7 @@ __all__ = ["CocoaPen"]
 
 class CocoaPen(BasePen):
 
-	def __init__(self, glyphSet, path=None):
+	def __init__(self, glyphSet=None, path=None):
 		BasePen.__init__(self, glyphSet)
 		if path is None:
 			from AppKit import NSBezierPath
--- a/Lib/fontTools/pens/qtPen.py
+++ b/Lib/fontTools/pens/qtPen.py
@@ -8,7 +8,7 @@ __all__ = ["QtPen"]
 
 class QtPen(BasePen):
 
-	def __init__(self, glyphSet, path=None):
+	def __init__(self, glyphSet=None, path=None):
 		BasePen.__init__(self, glyphSet)
 		if path is None:
 			from PyQt5.QtGui import QPainterPath
--- a/Lib/fontTools/pens/reportLabPen.py
+++ b/Lib/fontTools/pens/reportLabPen.py
@@ -11,7 +11,7 @@ class ReportLabPen(BasePen):
 
 	"""A pen for drawing onto a reportlab.graphics.shapes.Path object."""
 
-	def __init__(self, glyphSet, path=None):
+	def __init__(self, glyphSet=None, path=None):
 		BasePen.__init__(self, glyphSet)
 		if path is None:
 			path = Path()
--- a/Lib/fontTools/pens/svgPathPen.py
+++ b/Lib/fontTools/pens/svgPathPen.py
@@ -9,7 +9,7 @@ def pointToString(pt):
 
 class SVGPathPen(BasePen):
 
-    def __init__(self, glyphSet):
+    def __init__(self, glyphSet=None):
         BasePen.__init__(self, glyphSet)
         self._commands = []
         self._lastCommand = None
--- a/Lib/fontTools/pens/t2CharStringPen.py
+++ b/Lib/fontTools/pens/t2CharStringPen.py
@@ -44,7 +44,7 @@ class T2CharStringPen(BasePen):
     which are close to their integral part within the tolerated range.
     """
 
-    def __init__(self, width, glyphSet, roundTolerance=0.5, CFF2=False):
+    def __init__(self, width, glyphSet=None, roundTolerance=0.5, CFF2=False):
         super(T2CharStringPen, self).__init__(glyphSet)
         self.roundPoint = makeRoundFunc(roundTolerance)
         self._CFF2 = CFF2
--- a/Lib/fontTools/pens/wxPen.py
+++ b/Lib/fontTools/pens/wxPen.py
@@ -8,7 +8,7 @@ __all__ = ["WxPen"]
 
 class WxPen(BasePen):
 
-	def __init__(self, glyphSet, path=None):
+	def __init__(self, glyphSet=None, path=None):
 		BasePen.__init__(self, glyphSet)
 		if path is None:
 			import wx


justvanrossum:
I will review this later, but here's a historical note:
Originally all pens (derived from BasePen) required the glyphSet argument (which is often just passed to BasePen, which deals with components in a default way), and it's mostly later pens that aren't consistent with the original idea, which was: make it explicit you don't need support for composite glyphs by passing None as the glyphSet argument.

That said, I'm not against making it more consistently optional now, and I've often considered doing it as it is indeed more convenient in some cases.

anthrotype:
I don't know if it's a good idea.. Letting glyphSet be None by default means that, e.g. in ControlBoundsPen if there's a composite glyph, the addComponent method of DecomposingPen will be called with a None glyphSet, which means TypeError exception.
If caller is sure there are no composite glyphs, then one can pass None in the constructor, thus the responsibility is on the user.

anthrotype:
> Of the above pens that require glyphSet... only TTGlyphPen actually uses self.glyphSet and addComponent

that's not true. All the subclasses of BasePen that you listed, and that require glyphSet argument in their constructor, may need to call addComponent if they are used to draw a composite glyph.
And to be able to draw a composite glyph, you need a non-None glyphSet. That's why it is required for those.

anthrotype:
If we want to relax this requirement and allow BasePen subclasses to skip over composite glyphs (instead of letting a TypeError exception propagate as this PR does), then we may change the BasePen constructor like this: 

```diff
$ git diff
diff --git a/Lib/fontTools/pens/basePen.py b/Lib/fontTools/pens/basePen.py
index e0d3ae06..10415e2c 100644
--- a/Lib/fontTools/pens/basePen.py
+++ b/Lib/fontTools/pens/basePen.py
@@ -194,7 +194,7 @@ class BasePen(DecomposingPen):
        """
 
        def __init__(self, glyphSet=None):
-               super(BasePen, self).__init__(glyphSet)
+               super(BasePen, self).__init__(glyphSet if glyphSet is not None else {})
                self.__currentPoint = None
 
        # must override
```

Thene DecomposingPen by default will skip over missing components and only issue a logging warning (unless the pen's `skipMissingComponents` attribute is set to False, in which case KeyError exception is raised on missing components).

justvanrossum:
If we make the glyphSet argument optional (I don't care much either way), I do want to always get an exception when a component is being drawn, not just a logging warning. A KeyError may be confusing, as it suggests there is a glyph set, that just happens to not contain the requested glyph.

anthrotype:
that's why I prefer _not_ making it optional, if it the default value can produce an error. If the client knows what they are doing, they can set it to None (meaning "I don't have or need to resolve component references in this case") or to an empty dict (meaning "I don't have a glyphSet at hand, even if there are components but I don't really care if they draw empty so just skip them"); or the client can set skipMissingComponents attribute to False (like the RecordingPen does) if they do have a glyphSet available but some component may actually be missing from it. 
If this is not clear, the docstrings could be changed to make this clearer.

baileyparker:
Oh! I fundamentally misunderstood what composite glyph meant (the doc strings for `DecomposingPen.addComponent` led me to believe that it was doing something different--`BasePen.addComponent` is more clear). Apologies.

If it is the case that you do need the glyphset to render a glyph (because it is possible a glyph contains components--is this the proper terminology?), then I'd agree with @anthrotype that making it optional seems like the wrong move.

If I may, I think giving the Pen a glyphset might be going about this wrong. Pens shouldn't need to be concerned with internals of a glyphs they've been given.

Say we made `glyphSet` a required constructor param for all pens. Then you could do something like this:

```
comic_sans = TTFont('Comic Sans.ttf')
wingdings = TTFont('Wingdings.ttf')

p = SomePen(wingdings.getGlyphSet())
p.draw(comic_sans.getGlyphSet()['a'])
```

Here, the developer building the pen needs to remember where the glyph they're about to draw came from. If they forget or mess up, you end up with something nonsensical like this. And if the `a` glyph from Comic Sans had components, this would potentially `KeyError` because Wingdings might not have those component glyphs.

It seems like the better solution is to have a glyph know which glyphset it came from. Then a situation like this can never arise.

This is a bit more substantial of a rework, but I think the design is more sound. It eliminates the case of `None`/`{}` being ambiguous about whether a glyphset exists. `glyph.glyphSet is None` implies a glyph didn't come directly from a font. If it's possible to add glyphs to a font, you just need to set this when adding.

In effect, this solution is what the API currently requires of consumers (keeping the glyphset and glyph together if you ever want to do anything with pens). Since it's possible to get wrong, why not do it correctly once in the API so usage can look like:

```
p = SomePen()
p.draw(glyph)
```

anthrotype:
> Pens shouldn't need to be concerned with internals of a glyphs they've been given.

which is why Glyph objects have a draw method that takes a Pen, and not viceversa. Only the glyph knows its internals, and only calls method from the public interface of a Pen. The glyph draws itself using the given pen.
But the pen may need to know where to search components, hence the need to pass in a glyphSet.

I like to keep the Glyph and the glyphSet (i.e. Dict[str, Glyph]) separate to prevent the circularity of the references between the container and the contained.

justvanrossum:
Everything @anthrotype said. It was to keep the requirements for glyph objects to a minimum, and at the same time give pens the options to either retain component information or expand into outlines.

This design is close to 20 years old and has proven to be very useful for people. Sure, alternate designs are possible, but I really don't think it makes sense to change it now.

Let's leave it as it is, and live with the fact that there are some pens in our code base that _do_ treat the glyphSet constructor argument as optional. As I said earlier, those are the ones that aren't consistent...

baileyparker:
Oops, realized I flipped the order in my example. It should've been `glyph.draw(pen)`.

Sure, but doesn't passing the glyphset to the pen make the pen need to know specifics about the glyph (in particular, from which glyphset it came)?

I partially understand the aversion to cyclic references, but shouldn't it be weighed with ease of use of the API? The current way feels rather cumbersome to me:

```
pen = SomePen(font.getGlyphSet())
font.getGlyphSet()['a'].draw(pen)
```

Versus what adding the cyclic reference would allow:

```
pen = SomePen()
glyph.draw(pen)
```

Additionally, is passing the wrong glyphset not a concern?

As an API consumer, within my code I can't just pass around glyph objects. I need to pass around glyphs with their glyphsets if I ever want to be able to use pens. FWIW, I've wrapped the different font types in a wrapper class for a current project I'm working on. I also have a glyph wrapper. And it currently does this tying the knot thing (technically not really cyclic--but it's the spirit) so that you can indeed just pass my glyphs around:

```
class Font:
    def get_glyph(name: str) -> Glyph:
        return Glyph(self._font.getGlyphSet()[name], self)

class Glyph:
    def __init__(self, glyph: FontToolsGlyph, font: Font) -> None:
        self._glyph = glyph
        self._font = font

    @property
    def svg_path(self) -> str:
        pen = SVGPathPen(self._font._font.getGlyphSet())
        self._glyph.draw(pen)

        return pen.getCommands()
```

In effect, I see not doing this at the fontTools level just kicking the can down the road. At some point an API consumer will need to do this. So the net effect of this can kicking is the API is harder to use (and learn--I didn't understand components until now, and I don't think I should've needed to understand that to use a pen) and more error prone.

justvanrossum:
> Versus what adding the cyclic reference would allow:
> ```
> pen = SomePen()
> glyph.draw(pen)
> ```

What exactly does this allow?

It would make the `glyph.draw()` method responsible for either calling `pen.addComponent()` or expanding its components. Right now, this is a decision made by the pen object.

> Additionally, is passing the wrong glyphset not a concern?

I have not heard of a single case where this caused problems.

justvanrossum:
glyph.draw() is indeed supposed to call pen.addComponent(). For the pen to be able to decide what to do next (expand the component to an outline or not) it's pen.addComponent() that needs to use the glyph set.

Cyclic references have been accidentally added to the code base by people who didn't realize that having no cyclic references was a design goal. There are more examples like the one you mention. It's not a problem these days, but when fonttools was designed, it was still a concern.

Still, to _require_ a glyph to have a reference to its owning glyphset is a burden on the API, too. Right now, a glyphset needs to be dict-like, which needs to support `__getitem__`, and the returned glyph object should have a draw() method that takes a pen as its single argument. So, while it may be a little odd that a pen needs a glyph set reference, it does allow for a lot of simplicity on the glyph side.

fonttools by no means offers a perfect API:

> The issue is that API allows this mistake.

Oh dear, you are in for a world of pain using fonttools... There are way worse things that can happen, and there are many more examples of (pieces of) API that could be better or cleaner in this respect. It's a low level toolbox.

For higher level stuff, you don't want to use (eg.) TTGlyph objects, but your own objects, that are most suitable for the tasks you need to do. For example, the defcon and fontParts libraries offer glyph objects that _do_ have back references to their glyph sets. That is totally understandable at that level. But fonttools is designed to make little assumptions about its host app, and there's a lot of duck typing going on. The pen protocol is one example of this. It's relatively light-weight, which makes it easy to integrate with other code. Sure, you need to remember to pass a (correct) glyphset object to certain pens, but I stand by this tradeoff, that asks very little of glyphset and glyph objects.

baileyparker:
> fonttools by no means offers a perfect API

Fair enough. My aim was to improve the API, but that's obviously in conflict with the current design standards.

Thanks for indulging me! I appreciate the references too. I'll check out defcon and fontParts. I may not need to be working at this low of a level indeed.

justvanrossum:
> My aim was to improve the API

There's a lot of code out there that depends on how these things work. Changes that don't offer 100% API compatibility are usually problematic. We necessarily have to be very conservative in these areas.

behdad:
I wanted to change this for a while... Easiest would be to pass glyphSet to the addComponent() call directly.  Not sure how to roll that out with backwards compatibility though.

PR #1686, opened 2019-08-09
Make code Python 3 only

madig:
This PR breaks off a part of https://github.com/fonttools/fonttools/pull/1685 to make review more managable.

1. Mark as Python 3.6+ only in setup.py
2. Reduce misc.py23 to just the Python 3 symbols
3. Remove `__future__` imports


Lib/fontTools/misc/py23.py
```diff
@@ -1,525 +1,148 @@
...
+class Py23Error(NotImplementedError):
+    pass
 
 
-class Py23Error(NotImplementedError):
-	pass
-
-
-PY3 = sys.version_info[0] == 3
-PY2 = sys.version_info[0] == 2
```
anthrotype: you may as well keep these PY3 and PY2 constants, there may be code around still using them (can't remember)

anthrotype: actually scrap that, it's ok to remove.

benkiel:
Could this just throw a deprecation warning for now for `py23`, then change/go away? There are a bunch of things outside of fontTools that use the `py23` module. I know that the switch to Python3 is here for fontTools, but other packages that don't have as active maintainers are going to get caught off-guard here.

anthrotype:
yes, that's exactly what we did

benkiel:
@anthrotype ok, sorry for my confusion, I saw a bunch of code in `py23` gone and assumed the worst. thanks for helping ease the transition, really really really appreciated.

anthrotype:
no problem. what Nikolaus removed are the python2-specific code paths only.

Issue #1633, opened 2019-06-05
[subset] KeyError in EBDT table with msgothic.ttc


john45676:
The font is Microsoft Windows embeded font.

C:\Windows\Fonts\MS UI Gothic

Font number 0 & 1 are fine, but number 2 will subset fail. 

My command code is:
pyftsubset msgothic.ttc --text-file=XXX.txt --output-file=msgothic_subset.ttf --drop-tables= --font-number=2

Here is the message:
`    load_entry_point('fonttools==3.42.0', 'console_scripts', 'pyftsubset')()
  File "fonttools-3.42.0-py3.6.egg\fontTools\misc\loggingTools.py", line 375, in wrapper
    return func(*args, **kwds)
  File "fonttools-3.42.0-py3.6.egg\fontTools\subset\__init__.py", line 2873, in main
    save_font(font, outfile, options)
  File "fonttools-3.42.0-py3.6.egg\fontTools\misc\loggingTools.py", line 375, in wrapper
    return func(*args, **kwds)
  File "fonttools-3.42.0-py3.6.egg\fontTools\subset\__init__.py", line 2720, in save_font
    font.save(outfile, reorderTables=options.canonical_order)
  File "fonttools-3.42.0-py3.6.egg\fontTools\ttLib\ttFont.py", line 173, in save
    writer_reordersTables = self._save(tmp)
  File "fonttools-3.42.0-py3.6.egg\fontTools\ttLib\ttFont.py", line 212, in _save
    self._writeTable(tag, writer, done, tableCache)
  File "fonttools-3.42.0-py3.6.egg\fontTools\ttLib\ttFont.py", line 633, in _writeTable
    tabledata = self.getTableData(tag)
  File "fonttools-3.42.0-py3.6.egg\fontTools\ttLib\ttFont.py", line 651, in getTableData
    return self.tables[tag].compile(self)
  File "fonttools-3.42.0-py3.6.egg\fontTools\ttLib\tables\E_B_D_T_.py", line 101, in compile
    glyph = curGlyphDict[curName]
KeyError: 'uni4E0A'`

Thanks.

anthrotype:
`EBDT` table is currently not supported by the subsetter. In fact it is included in the default list of `drop-tables`. Since you used `--drop-tables=`, you requested to not drop any the tables. The `EBDT` table is thus kept, but it fails to compile because the rest of the font was subsetted.

[...]

PR #1698, opened 2019-08-16
[subset] Prune empty strikes in EBDT data

khaledhosny:
After subsetting some strikes might be empty (strikes don’t have to cover all glyphs equally) which would cause a key error later when saving the table.

Fixes https://github.com/fonttools/fonttools/issues/1633

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -1388,6 +1388,9 @@ def subset_glyphs(self, s):
 def subset_glyphs(self, s):
   self.strikeData = [{g: strike[g] for g in s.glyphs if g in strike}
 					 for strike in self.strikeData]
+  # Prune empty strikes
+  # https://github.com/fonttools/fonttools/issues/1633
+  self.strikeData = list(filter(bool, self.strikeData))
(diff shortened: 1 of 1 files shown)


Lib/fontTools/subset/__init__.py
```diff
@@ -1388,6 +1388,9 @@ def subset_glyphs(self, s):
 def subset_glyphs(self, s):
   self.strikeData = [{g: strike[g] for g in s.glyphs if g in strike}
 					 for strike in self.strikeData]
+  # Prune empty strikes
+  # https://github.com/fonttools/fonttools/issues/1633
+  self.strikeData = list(filter(bool, self.strikeData))
```
anthrotype: this works, but the canonical way for pruning empty/null items from a sequence is passing `None` as filter

```suggestion
  self.strikeData = list(filter(None, self.strikeData))
```

https://docs.python.org/3/library/functions.html#filter

> If function is None, the identity function is assumed, that is, all elements of iterable that are false are removed.

A list comprehension would be equally fine:

```python
self.strikeData = [s for s in self.strikeData if s]
```

Issue #1541, opened 2019-03-15
[feaLib] Octal numbers are parsed as decimal


dscorbett:
[The OpenType feature file spec](https://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#9.e) says, “Decimal numbers must begin with a non-0 digit, octal numbers with a 0 digit, and hexadecimal numbers with a 0x prefix to numbers and hexadecimal letters a-f or A-F.” Octal numbers are accepted after `nameid`, `name`, and `sizemenuname`; makeotf also accepts them after `Character`. feaLib’s parser does not recognize octal numbers and just treats them as decimal.

khaledhosny:
Do you have a sample feature file to test with?

dscorbett:
```pycon
$ python3 -q
>>> from io import StringIO
>>> from fontTools.feaLib.parser import Parser
>>> Parser(StringIO("feature dist { position a 010; } dist;")).parse().statements[0].statements[0].pos[0][1].xAdvance
10
```

khaledhosny:
Can octal or hexadecimal numbers be prefixed by `-`?

dscorbett:
[It doesn’t look like it.](https://github.com/adobe-type-tools/afdko/blob/be1e5d33f09f14e67d1c3b57e0cdb43b826532be/c/makeotf/makeotf_lib/source/hotconv/featpars.dlg#L615-L618) It also looks like non-decimal numbers are only allowed in certain contexts, after the keywords in my original comment. So I think my 'dist' example should be a syntax error.

khaledhosny:
Looks like feaLib already accepts hexadecimals everywhere, but I think with octals it might be worse. Need to find what makeotf does here and see if we can follow it.

PR #1700, opened 2019-08-17
[feaLib] Correctly handle octal numbers

khaledhosny:
From https://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#9.e:
> Decimal numbers must begin with a non-0 digit, octal numbers with a 0
> digit, and hexadecimal numbers with a 0x prefix to numbers and
> hexadecimal letters a-f or A-F.

Fixes https://github.com/fonttools/fonttools/issues/1541

### DIFF
--- a/Lib/fontTools/feaLib/lexer.py
+++ b/Lib/fontTools/feaLib/lexer.py
@@ -6,6 +6,9 @@ import os
 
 class Lexer(object):
     NUMBER = "NUMBER"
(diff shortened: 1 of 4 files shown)


khaledhosny:
I think his accepts octal numbers anywhere hexadecimal or decimal numbers are accepted, which may be wrong. Need to double check.

PR #1706, opened 2019-08-24
Fix MultipleSubst subtable overflows

dscorbett:


anthrotype:
this looks correct, thanks!
Would be great to also add unit tests in `Tests/ttLib/tables/otTables_test.py`, something along the lines of the `test_splitMarkBasePos`. Please let me know if you want to work on this.

dscorbett:
Sure, I’ll write a test.

PR #1715, opened 2019-09-05
[ttLib] Add aliases for renamed entries in hhea table

### LINKED ISSUE #1714: ascender and ascent
The [opentype spec ](https://docs.microsoft.com/en-gb/typography/opentype/spec/hhea) calls the first two substantive entries in the `hhea` table "`ascender`" and "`descender`". fonttools calls them "`ascent`" and "`descent`".

This was surprising! Maybe it's too late to change then but can we at least have an alias?

simoncozens:
Closes #1714

### DIFF
--- a/Tests/ttLib/tables/_h_h_e_a_test.py
+++ b/Tests/ttLib/tables/_h_h_e_a_test.py
@@ -126,6 +126,13 @@ class HheaCompileOrToXMLTest(unittest.TestCase):
             len([r for r in captor.records
                  if "Table version value is a float" in r.msg]) == 1)
 
+    def test_aliases(self):
+        hhea = self.font['hhea']
+        self.assertEqual(hhea.ascent, hhea.ascender)
+        self.assertEqual(hhea.descent, hhea.descender)
+        hhea.ascender = 800
+        self.assertEqual(hhea.ascent, 800)
+        hhea.ascent = 750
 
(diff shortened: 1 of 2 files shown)


justvanrossum:
I wonder if it should be the other way around: rename the actual attributes and provide getters/setters for the old names?

simoncozens:
Huh. Maybe not, as it would mean the XML files become incompatible.

behdad:
XML as well as properties dict...

justvanrossum:
> Maybe not, as it would mean the XML files become incompatible

Right, your original proposal is safest.

Tests/ttLib/tables/_h_h_e_a_test.py
```diff
@@ -126,6 +126,13 @@ def test_toXML_version_as_float(self):
             len([r for r in captor.records
                  if "Table version value is a float" in r.msg]) == 1)
 
+    def test_aliases(self):
+        hhea = self.font['hhea']
+        self.assertEqual(hhea.ascent, hhea.ascender)
+        self.assertEqual(hhea.descent, hhea.descender)
+        hhea.ascender = 800
+        self.assertEqual(hhea.ascent, 800)
+        hhea.ascent = 750
```
justvanrossum: Did you mean to continue here? You're not testing the descender setter.

anthrotype:
thanks. I'll complete the test cases for the descender setter

PR #1716, opened 2019-09-05
Remove the import of redirect_stderr

hhb:
They are not used. And redirect_stderr is introduced in python 3.5. It prevents execution on old python.

### DIFF
--- a/Lib/fontTools/misc/py23.py
+++ b/Lib/fontTools/misc/py23.py
@@ -3,7 +3,6 @@
 import decimal as _decimal
 import math as _math
 import warnings
-from contextlib import redirect_stderr, redirect_stdout
 from io import BytesIO
 from io import StringIO as UnicodeIO
 from types import SimpleNamespace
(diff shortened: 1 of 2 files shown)


hhb:
Hmm I don't quite understand AppVeyor error. Is there anything actually depending on redirect_err? My search didn't show anything...

justvanrossum:
We don't support Python < 3.6 anymore, but still, that import indeed seems unused and unneeded.

anthrotype:
The symbol was kept there because the module contained a backport with that name. I don't remember why it was originally added, but it I think it's safer to keep it there just in case it may be used elsewhere.
And yes, we don't support < 3.6 any more.

justvanrossum:
These symbols aren't listed in `__all__`, though. (But I don't mind if we keep them.)

anthrotype:
true, but the intention in keeping the symbols was to avoid unintended breakages on python versions that we do support.

anthrotype:
in a few months, we can get rid of the py23 module altogether.

PR #1718, opened 2019-09-09
ufoLib: Clean up legacy code

madig:
Use Python 3.6+ syntax, remove redundant code.

### DIFF
--- a/Lib/fontTools/ufoLib/__init__.py
+++ b/Lib/fontTools/ufoLib/__init__.py
@@ -1,6 +1,7 @@
 import sys
 import os
 from copy import deepcopy
+from os import fsdecode
 import logging
 import zipfile
 import enum
@@ -20,7 +21,7 @@ from fontTools.ufoLib.validators import *
 from fontTools.ufoLib.filenames import userNameToFileName
 from fontTools.ufoLib.converters import convertUFO1OrUFO2KerningToUFO3Kerning
 from fontTools.ufoLib.errors import UFOLibError
-from fontTools.ufoLib.utils import datetimeAsTimestamp, fsdecode, numberTypes
+from fontTools.ufoLib.utils import numberTypes
 
 """
 A library for importing .ufo files and their descendants.
@@ -120,7 +121,7 @@ class _UFOBaseIO:
 		except (fs.errors.MissingInfoNamespace, fs.errors.ResourceNotFound):
 			return None
 		else:
-			return datetimeAsTimestamp(dt)
+			return dt.timestamp()
 
 	def _getPlist(self, fileName, default=None):
 		"""
--- a/Lib/fontTools/ufoLib/__init__.py
+++ b/Lib/fontTools/ufoLib/__init__.py
@@ -1323,8 +1323,6 @@ class UFOWriter(UFOReader):
 			for layerName in layerOrder:
 				if layerName is None:
 					layerName = DEFAULT_LAYER_NAME
-				else:
-					layerName = tostr(layerName)
 				newOrder.append(layerName)
 			layerOrder = newOrder
 		else:
--- a/Lib/fontTools/ufoLib/__init__.py
+++ b/Lib/fontTools/ufoLib/__init__.py
@@ -671,7 +671,7 @@ class UFOReader(_UFOBaseIO):
 
 	def getCharacterMapping(self, layerName=None, validate=None):
 		"""
-		Return a dictionary that maps str values (ints) to
+		Return a dictionary that maps unicode values (ints) to
 		lists of glyph names.
 		"""
 		if validate is None:
--- a/Lib/fontTools/ufoLib/__init__.py
+++ b/Lib/fontTools/ufoLib/__init__.py
@@ -1,4 +1,3 @@
-import sys
 import os
 from copy import deepcopy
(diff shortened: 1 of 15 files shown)


Lib/fontTools/ufoLib/__init__.py
```diff
@@ -2042,17 +2042,17 @@ def convertFontInfoValueForAttributeFromVersion1ToVersion2(attr, value):
 		if attr == "fontStyle":
 			v = _fontStyle1To2.get(value)
 			if v is None:
-				raise UFOLibError("Cannot convert value (%s) for attribute %s." % (repr(value), attr))
+				raise UFOLibError(f"Cannot convert value ({repr(value)}) for attribute {attr}.")
```
anthrotype: you could use `{value!r}` with `!r` conversion flag, which is the same as calling `repr(value)`

Lib/fontTools/ufoLib/utils.py
```diff
@@ -1,50 +1,12 @@
...
-
-# TODO: should import from fontTools.misc.py23
-try:
-	long = long
-except NameError:
-	long = int
-
-integerTypes = (int, long)
-numberTypes = (int, float, long)
+integerTypes = int
```
anthrotype: you may as well get rid of this `integerTypes` now

Lib/fontTools/ufoLib/__init__.py
```diff
@@ -671,7 +671,7 @@ def getGlyphSet(self, layerName=None, validateRead=None, validateWrite=None):
 
 	def getCharacterMapping(self, layerName=None, validate=None):
 		"""
-		Return a dictionary that maps unicode values (ints) to
+		Return a dictionary that maps str values (ints) to
```
justvanrossum: This should still read "unicode"

Lib/fontTools/ufoLib/glifLib.py
```diff
@@ -695,12 +695,12 @@ def _writeUnicodes(glyphObject, element, validate):
...
 	note = getattr(glyphObject, "note", None)
-	if validate and not isinstance(note, basestring):
-		raise GlifLibError("note attribute must be str or unicode")
+	if validate and not isinstance(note, str):
+		raise GlifLibError("note attribute must be str")
 	note = note.strip()
 	note = "\n" + note + "\n"
-	# ensure text is unicode, if it's bytes decode as ASCII
-	etree.SubElement(element, "note").text = tounicode(note)
+	# ensure text is str, if it's bytes decode as ASCII
```
anthrotype: the comment is stale, you're no longer ensuring text is `str`, you just setting text attribute directly (assuming it's already a `str`, which it is). remove the comment

Lib/fontTools/ufoLib/__init__.py
```diff
@@ -1325,7 +1325,7 @@ def writeLayerContents(self, layerOrder=None, validate=None):
 				if layerName is None:
 					layerName = DEFAULT_LAYER_NAME
 				else:
-					layerName = tounicode(layerName)
+					layerName = tostr(layerName)
```
anthrotype: you can just remove the `tostr` (or `tounicode` which are the same on py3) altogether here. Doesn't make sense to continue supporting passing in bytes strings for layer names. The `layerOrder` argument is a `Optional[List[str]]`, so the decode is unnecessary.

Lib/fontTools/ufoLib/__init__.py
```diff
@@ -1439,9 +1439,9 @@ def _getGlyphSetFormatVersion3(self, validateRead, validateWrite, layerName=None
 				# not caching this could be slightly expensive,
 				# but caching it will be cumbersome
 				existing = {d.lower() for d in self.layerContents.values()}
-				if not isinstance(layerName, unicode):
+				if not isinstance(layerName, str):
```
anthrotype: you can remove the entire `if not isinstance(layerName, str):` block, since a `layerName` as anything else than `str` should now be unsupported/impossible.

Lib/fontTools/ufoLib/__init__.py
anthrotype: good catch!

anthrotype:
about the appveyor failures, I don't know what exactly is going on (the `caplog` fixture seems unable to capture the logging messages, which are sent to stderr instead). If you look at the appveyor build history the same failures started happening since pytest was updated to 5.1.2. Maybe try pinning pytest to 5.1.1 and see if the errors go away.

Lib/fontTools/ufoLib/__init__.py
madig: Ha, I was wondering why the test case has a byte name.

madig:
I just discovered that UFO3_test.py defines "def testRemoveDefaultLayer" twice, i.e. the first one doesn't get run. Making it run makes it fail. Oops :o

dscorbett:
Was it intentional to not modify format expressions without parentheses, like `"Images are not allowed in UFO %d." % self._formatVersion`?

madig:
An oversight because pyupgrade is very timid about this and I was too lazy to check everything 😊 there's more stuff left, like `dict(type=something)` that could be replaced with literals.

PR #1719, opened 2019-09-10
[name] add removeName method to table__n_a_m_e

m4rc1e:
We can currently get and set names but we don't have a method to remove them. I've written similar code in several throwaway scripts before so I think it warrants its own method.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_n_a_m_e.py
+++ b/Lib/fontTools/ttLib/tables/_n_a_m_e.py
@@ -147,6 +147,31 @@ class table__n_a_m_e(DefaultTable.DefaultTable):
 		else:
 			self.names.append(makeName(string, nameID, platformID, platEncID, langID))
 
+	def removeNames(self, nameID=None, platformID=None, platEncID=None, langID=None):
+		"""Remove any name records identified by 'nameID', 'platformID', 'platEncID'
+		or 'langID'.
+		"""
+		args = {
+			argName: argValue
+			for argName, argValue in (
+				("nameID", nameID),
+				("platformID", platformID),
+				("platEncID", platEncID),
+				("langID", langID),
+			)
+			if argValue is not None
+		}
+		if not args:
+			# no arguments, nothing to do
+			return
+		self.names = [
+			rec for rec in self.names
+			if any(
+				argValue != getattr(rec, argName)
+				for argName, argValue in args.items()
+			)
+		]
+
 	def _findUnusedNameID(self, minNameID=256):
 		"""Finds an unused name id.
 
--- a/Tests/ttLib/tables/_n_a_m_e_test.py
+++ b/Tests/ttLib/tables/_n_a_m_e_test.py
@@ -75,6 +75,39 @@ class NameTableTest(unittest.TestCase):
 		with self.assertRaises(TypeError):
 			table.addName(b"abc")  # must be unicode string
 
+	def test_removeNames(self):
+		table = table__n_a_m_e()
+		table.setName("Regular", 2, 1, 0, 0)
+		table.setName("Regular", 2, 3, 1, 0x409)
+		table.removeNames(nameID=2)
+		self.assertEqual(table.names, [])
+
+		table = table__n_a_m_e()
+		table.setName("FamilyName", 1, 1, 0, 0)
+		table.setName("Regular", 2, 1, 0, 0)
+		table.setName("FamilyName", 1, 3, 1, 0x409)
+		table.setName("Regular", 2, 3, 1, 0x409)
+		table.removeNames(platformID=1)
+		self.assertEqual(len(table.names), 2)
+
+		table = table__n_a_m_e()
+		table.setName("FamilyName", 1, 1, 0, 0)
+		table.setName("Regular", 2, 1, 0, 0)
+		table.removeNames(nameID=1)
+		self.assertEqual(len(table.names), 1)
+
+		table = table__n_a_m_e()
+		table.setName("FamilyName", 1, 1, 0, 0)
+		table.setName("Regular", 2, 1, 0, 0)
+		table.removeNames(2, 1, 0, 0)
+		self.assertEqual(len(table.names), 1)
+
+		table = table__n_a_m_e()
+		table.setName("FamilyName", 1, 1, 0, 0)
+		table.setName("Regular", 2, 1, 0, 0)
+		table.removeNames()
+		self.assertEqual(len(table.names), 2)
+
 	def test_addMultilingualName(self):
 		# Microsoft Windows has language codes for “English” (en)
 		# and for “Standard German as used in Switzerland” (de-CH).


anthrotype:
would also be nice to have a `removeNames` (plural) method to remove _any_ name record that matches given combination of nameID, platformID, platEncID or langID.
E.g. say one wants to drop all records with platformID=1.
The signature of `removeNames` could have all of arguments default to None 

```python
def removeNames(self, nameID=None, platformID=None, platEncID=None, langID=None):
```

the methods iterates over all name records and only keeps those which don't match the input
(no-op if all args are None).

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -147,6 +147,18 @@ def setName(self, string, nameID, platformID, platEncID, langID):
...
+		"""Remove a name record identified by 'nameID', 'platformID', 'platEncID'
+                and 'langID'.
+		"""
+		for i, namerecord in enumerate(self.names):
+			if (namerecord.nameID == nameID and
+				namerecord.platformID == platformID and
+				namerecord.platEncID == platEncID and
+				namerecord.langID == langID):
+				    del self.names[i]
+				    break
```
anthrotype: how about
```python
self.names = [
    n for n in self.names
    if n.nameID != nameID
    and n.platformID != platformID
    and n.platEncID != platEncID
    and n.langID != langID
]
```

m4rc1e: This approach will definitely work better for removeNames. I'll update and include both methods.

m4rc1e:
I've added the removeNames by using [functools.partialmethod](https://docs.python.org/3/library/functools.html#functools.partialmethod). Python 3.4+ is needed in order to use it. Happy to refactor if we need to support earlier versions of Python.

m4rc1e:
Another approach without using partialmethod could be:

```
	def removeNames(self, nameID=None, platformID=None, platEncID=None, langID=None):
		self.names = [n for n in self.names
				if n.nameID != nameID
				if n.platformID != platformID
				if n.platEncID != platEncID
				if n.langID != langID
		]

        def removeName(self, nameID, platformID, platEncID, langID):
            self.removeNames(nameID, platformID, platEncID, langID)
```

justvanrossum:
I think `removeNames` is enough. I'm not sure what the added benefit of `removeName` is to be honest.

anthrotype:
I agree with Just, just removeNames is sufficient.

m4rc1e:
Thanks for the feedback both of you. I've dropped `removeName` and kept `removesNames`.

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -147,6 +147,17 @@ def setName(self, string, nameID, platformID, platEncID, langID):
 		else:
 			self.names.append(makeName(string, nameID, platformID, platEncID, langID))
 
+	def removeNames(self, nameID=None, platformID=None, platEncID=None, langID=None):
+		"""Remove name records identified by 'nameID', 'platformID', 'platEncID'
+		and 'langID'.
+		"""
+		self.names = [n for n in self.names
```
anthrotype: I prefer if add a newline after the opening `[`, with a single indentation level

anthrotype:
please do `git fetch origin && git rebase origin/master` to get rid of unrelated extra commits

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -147,6 +147,18 @@ def setName(self, string, nameID, platformID, platEncID, langID):
 		else:
 			self.names.append(makeName(string, nameID, platformID, platEncID, langID))
 
+	def removeNames(self, nameID=None, platformID=None, platEncID=None, langID=None):
+		"""Remove name records identified by 'nameID', 'platformID', 'platEncID'
+		and 'langID'.
```
anthrotype:
```suggestion
		or 'langID'.
```

Tests/ttLib/tables/_n_a_m_e_test.py
```diff
@@ -75,6 +75,28 @@ def test_addName(self):
 		with self.assertRaises(TypeError):
 			table.addName(b"abc")  # must be unicode string
 
+	def test_removeNames(self):
```
anthrotype: you could also add explicit tests for `removeNames` when called with more than one argument, especially one with all four of them.

Tests/ttLib/tables/_n_a_m_e_test.py
```diff
@@ -75,6 +75,28 @@ def test_addName(self):
...
+		table = table__n_a_m_e()
+		table.setName("Regular", 2, 1, 0, 0)
+		table.setName("Regular", 2, 3, 1, 0.409)
```
anthrotype: typo?

```suggestion
		table.setName("Regular", 2, 3, 1, 0.409)
		table.setName("Regular", 2, 3, 1, 0x409)
```

anthrotype:
this should work:

```python
def removeNames(self, nameID=None, platformID=None, platEncID=None, langID=None):
    """Remove any name records identified by 'nameID', 'platformID', 'platEncID'
    or 'langID'.
    """
    args = {
        argName: argValue
        for argName, argValue in (
            ("nameID", nameID),
            ("platformID", platformID),
            ("platEncID", platEncID),
            ("langID", langID),
        )
        if argValue is not None
    }
    if not args:
        # no arguments, nothing to do
        return
    self.names = [
        rec for rec in self.names
        if any(
            argValue != getattr(rec, argName)
            for argName, argValue in args.items()
        )
    ]
```

m4rc1e:
I came up with:

```
def removeNames(self, nameID=None, platformID=None, platEncID=None, langID=None):
        """Remove name records identified by 'nameID', 'platformID', 'platEncID'
        and 'langID'.
        """
        if not any ([nameID, platformID, platEncID, langID]):
                return
        nameIDs, platformIDs, platEncIDs, langIDs = set(), set(), set(), set()
        for name in self.names:
                if name.nameID == nameID or nameID is None:
                        nameIDs.add(name)
                if name.platformID == platformID or platformID is None:
                        platformIDs.add(name)
                if name.platEncID == platEncID or platEncID is None:
                        platEncIDs.add(name)
                if name.langID == langID or langID is None:
                        langIDs.add(name)
        to_remove = nameIDs & platformIDs & platEncIDs & langIDs
        self.names = sorted(set(self.names) - to_remove)
```

m4rc1e:
Yours is cleaner so I'm going to use it :-). Thank you for your assistance on this.

justvanrossum:
Why is Cosimo's original listcomp not good enough?

anthrotype:
> Why is Cosimo's original listcomp not good enough?

because we want the method to drop name records when _all_ the given (non-None) parameters match. My simpler list comprehension was too greedy, as it dropped any records when _any_ of the input parameters match the corresponding attribute.

justvanrossum:
> My simpler list comprehension was too greedy

Duh of course, now I see it. Thanks!

PR #1720, opened 2019-09-10
PointToSegmentPen: preserve duplicate last point

anthrotype:
The PointToSegmentPen translates between PointPen and (Segment)Pen protocol.

In the SegmentPen protocol, closed contours always imply a final 'lineTo' segment from the last oncurve point to the starting point.
So the PointToSegmentPen omits the final 'lineTo' segment for closed contours -- unless the option 'outputImpliedClosingLine' is True (it is False by default, and defcon.Glyph.draw method initializes the
converter pen without this option).

However, if the last oncurve point is on a "line" segment and has same coordinates as the starting point of a closed contour, the converter pen must always output the closing 'lineTo' explicitly (regardless of the value of the 'outputImpliedClosingLine' option) in order to disambiguate this case from the implied closing 'lineTo'.

If it doesn't do that, a duplicate 'line' point at the end of a closed contour gets lost in the conversion.

See https://github.com/googlefonts/fontmake/issues/572.

### DIFF
--- a/Lib/fontTools/pens/pointPen.py
+++ b/Lib/fontTools/pens/pointPen.py
@@ -180,18 +180,36 @@ class PointToSegmentPen(BasePointToSegmentPen):
 			pen.moveTo(movePt)
 		outputImpliedClosingLine = self.outputImpliedClosingLine
 		nSegments = len(segments)
+		lastPt = movePt
 		for i in range(nSegments):
 			segmentType, points = segments[i]
 			points = [pt for pt, smooth, name, kwargs in points]
 			if segmentType == "line":
 				assert len(points) == 1, "illegal line segment point count: %d" % len(points)
 				pt = points[0]
-				if i + 1 != nSegments or outputImpliedClosingLine or not closed:
+				# For closed contours, a 'lineTo' is always implied from the last oncurve
+				# point to the starting point, thus we can omit it when the last and
+				# starting point don't overlap.
+				# However, when the last oncurve point is a "line" segment and has same
+				# coordinates as the starting point of a closed contour, we need to output
+				# the closing 'lineTo' explicitly (regardless of the value of the
+				# 'outputImpliedClosingLine' option) in order to disambiguate this case from
+				# the implied closing 'lineTo', otherwise the duplicate point would be lost.
+				# See https://github.com/googlefonts/fontmake/issues/572.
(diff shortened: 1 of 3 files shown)


anthrotype:
ignore for now the unrelated appveyor failure (see https://github.com/fonttools/fonttools/pull/1718#issuecomment-529802170)

Lib/fontTools/pens/pointPen.py
```diff
@@ -180,18 +180,36 @@ def _flushContour(self, segments):
...
+				# coordinates as the starting point of a closed contour, we need to output
+				# the closing 'lineTo' explicitly (regardless of the value of the
+				# 'outputImpliedClosingLine' option) in order to disambiguate this case from
+				# the implied closing 'lineTo', otherwise the duplicate point would be lost.
+				# See https://github.com/googlefonts/fontmake/issues/572.
+				if (
+					i + 1 != nSegments
+					or outputImpliedClosingLine
+					or not closed
+					or pt == lastPt
```
anthrotype: the counterpart of this is in `SegmentToPointPen.closePath`, where there is a similar comparison of the coordinates of the last and first points, which determines whether there's an implied line or not:

https://github.com/fonttools/fonttools/blob/afdb967a2841547161f72b559fca7845e45a409e/Lib/fontTools/pens/pointPen.py#L247-L250

anthrotype:
I am confident this is the correct solution. I am surprised the bug didn't surface earlier, given how widely used the ufoLib pointPens are. I guess the loss of a duplicate point was not big issue in normal circumstances, unless strict point-wise compatibility of contours is required as in variable font workflows.

justvanrossum:
Yeah, the varfont use case exposes this as an issue indeed. Before, I think it was pretty much never desirable to have a duplicate point anywhere.

jansindl3r:
thanks for having look on that.. testing it with fontmake <i>(origin of this issue)</i> gives now different error 
`WARNING:fontTools.varLib:glyph caron has incompatible masters; skipping`
which is now 
`ERROR:cu2qu.ufo:Glyphs named 'caron' have different number of segments`

madig:
If `caron` is actually compatible and the conversion to quadratic outlines is still botched, please post more detail on the original issue over at the fontmake issue tracker.

PR #1721, opened 2019-09-10
tox.ini: pin pytest==5.1.1 to workaround issue with caplog fixture on windows

anthrotype:
let's see if this fixes the recent failues on Appveyor CI: e.g. https://ci.appveyor.com/project/fonttools/fonttools/builds/27126458/job/iomtsu4prflws4xu

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,7 +5,7 @@ envlist = py3{6,7}-cov, htmlcov
 [testenv]
 deps =
     cov: coverage>=4.3
-    pytest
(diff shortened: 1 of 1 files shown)


anthrotype:
The tests seem to pass on Windows if I freeze pytest version to 5.1.1. I wish I had a local windows machine available to debug why they fail on pytest 5.1.2. I'll merge this for the time being, and open an issue for it so we can retry un-pin pytest at some point in the future.

PR #1726, opened 2019-09-17
group fix side check

typoman:
This fixes the group side checking. The current state considers any non-kerning group as the right side kerning.


justvanrossum:
I don't think this fix is needed as any non `public.kern1` or `public.kern2` groups are skipped:
https://github.com/fonttools/fonttools/blob/7cb8e01edaac66801f250e1e84434d7df3898cdb/Lib/fontTools/ufoLib/validators.py#L885

typoman:
Ow I see, I was getting errors for glyph happening on non-kerning groups. I'm gonna check again if the error was caused by another problem. Thanks Just.

justvanrossum:
Writing a doctest for this case would have shown you that the code here is in fact correct :)

PR #1740, opened 2019-10-08
don't round in fixedToFloat, unless we're dumping floats to XML

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/737

I changed `fixedToFloat` to NOT do any magic, simply return `value / (1 << precisionBits)`
The code of the previous fixedToFloat is now called `fixedToStr`, but returns a string and is used in `toXML` methods to write fixed-point numbers as compact decimal floats.

The function actually used in `toXML` methods is `floatToFixedToStr` (that in turn uses `fixedToStr`), because internally table objects store these fixed-point numbers as floats, not as integers.
But I still kept `fixedToStr` (and `strToFixed`) in `fixedTools` module, even though they are not used directly anywhere.

There is also a corresponding `strToFixedToFloat` that does the reverse and it's equivalent of doing `floatToFixedToFloat(float(string))`. This is used in `fromXML` methods to ensure that the internal un-rounded float values are the same whether a table was decompiled from binary (hence used `fixedToFloat`) or loaded from XML.

I updated all the calling sites accordingly, including `Fixed` and `F2Dot14` types in `otConverters`, fvar, avar, glyf, head, trak, TupleVariations and psCharStrings.

/cc @Lorp

### DIFF
--- a/Lib/fontTools/misc/fixedTools.py
+++ b/Lib/fontTools/misc/fixedTools.py
@@ -11,7 +11,11 @@ __all__ = [
 	"otRound",
 	"fixedToFloat",
 	"floatToFixed",
-    "floatToFixedToFloat",
+	"floatToFixedToFloat",
+	"floatToFixedToStr",
+	"fixedToStr",
+	"strToFixed",
+	"strToFixedToFloat",
 	"ensureVersionIsLong",
 	"versionToFixed",
 ]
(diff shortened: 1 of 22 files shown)


behdad:
Thanks for working on this!

Thiss sounds like what I remember needed to be done.  lgtm.

Lorp:
Looks like a very good separation of responsibility :)

anthrotype:
Sorry it took me so long to get to it. I was actually hit by it while working on the partial instancer...

PR #1742, opened 2019-10-11
[name] Handle duplicate name records with different `string` types

madig:
Found this one when I tried to add a name ID 25 string (type `str`) to a font that already had one (type `bytes`) and overlooked the `setName` method. Turns out there is no sanitization when you append name records manually to the `names` list.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_n_a_m_e.py
+++ b/Lib/fontTools/ttLib/tables/_n_a_m_e.py
@@ -459,22 +459,36 @@ class NameRecord(object):
 		if type(self) != type(other):
 			return NotImplemented
 
-		# implemented so that list.sort() sorts according to the spec.
-		selfTuple = (
-			getattr(self, "platformID", None),
-			getattr(self, "platEncID", None),
-			getattr(self, "langID", None),
-			getattr(self, "nameID", None),
-			getattr(self, "string", None),
-		)
-		otherTuple = (
-			getattr(other, "platformID", None),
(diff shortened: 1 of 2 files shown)


Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -459,22 +459,36 @@ def __lt__(self, other):
...
+			)
+			otherTuple = (
+				getattr(other, "platformID", None),
+				getattr(other, "platEncID", None),
+				getattr(other, "langID", None),
+				getattr(other, "nameID", None),
+				other.toBytes(),
+			)
+			return selfTuple < otherTuple
+		except UnicodeEncodeError:
```
anthrotype: Maybe also catch `AttributeError`, since `toBytes` unconditionally gets the `string` attribute, whereas the others are (maybe too cautiously) using `gettattr(self, ..., None)`
actually, let's turn all these `getattr` to unconditional self.platformID, self.platEncID, etc. and catch AttributeError and return NotImplemented if any of these attributes is missing. This can only happen for 1) an object that is not a NameRecord, or an unlikely incomplete NameRecord object which has not been fully populated. In both cases the best is to return NotImplemented.
I also think the warning is too detailed and verbose for an occurrance which is not very likely to ever occur in real world. It will raise an UnicodeEncodeError _anyway_ as soon as you attempt to compile that thing, and it's clear where the error is coming from. Just remove the logging.warning, thanks

PR #1743, opened 2019-10-11
Remove some Python 2 era import gaming

madig:
Happens to help with mypy checking.

### DIFF
--- a/Lib/fontTools/misc/plistlib.py
+++ b/Lib/fontTools/misc/plistlib.py
@@ -1,5 +1,6 @@
 import sys
 import re
+import warnings
 from io import BytesIO
 from datetime import datetime
 from base64 import b64encode, b64decode
@@ -149,6 +150,12 @@ class PlistTarget:
         if use_builtin_types is None:
             self._use_builtin_types = USE_BUILTIN_TYPES
         else:
+            if use_builtin_types is False:
+                warnings.warn(
+                    "Setting use_builtin_types to False is deprecated and will be "
+                    "removed soon.",
+                    DeprecationWarning,
+                )
             self._use_builtin_types = use_builtin_types
         self._dict_type = dict_type
 
--- a/Lib/fontTools/misc/plistlib.py
+++ b/Lib/fontTools/misc/plistlib.py
@@ -5,17 +5,15 @@ from datetime import datetime
 from base64 import b64encode, b64decode
 from numbers import Integral
 
+from types import SimpleNamespace
 from collections.abc import Mapping
 from functools import singledispatch
 
 from fontTools.misc import etree
 
 from fontTools.misc.py23 import (
-    basestring,
     tounicode,
     tobytes,
-    SimpleNamespace,
-    range,
 )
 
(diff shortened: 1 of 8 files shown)


Lib/fontTools/misc/plistlib.py
```diff
@@ -5,18 +5,8 @@
...
-
-try:
-    from functools import singledispatch
-except ImportError:
-    try:
-        from singledispatch import singledispatch
-    except ImportError:
-        singledispatch = None
+from collections.abc import Mapping
+from functools import singledispatch
```
anthrotype: In plislitb module if you look further down there is an if singldispatch is not None branch, you can simplify that one since we no longer support the case of singldispatch not being present.

Also there is a different PY2 vs PY3 treatment for use_builtin_types argument default value, that can be simplified too

Tests/misc/plistlib_test.py
```diff
@@ -12,11 +12,8 @@
...
 import pytest
+from collections.abc import Mapping
 
-try:
-    from collections.abc import Mapping # python >= 3.3
-except ImportError:
-    from collections import Mapping
 
 PY2 = sys.version_info < (3,)
 if PY2:
```
anthrotype: if PY2, code is never executed, so remove this

Lib/fontTools/misc/plistlib.py
madig: Should I make `use_builtin_types` a no-op and print a deprecation warning when it's used?

madig:
I suppose the `Data` class and the `use_builtin_types` handling can also go, but I'm having difficulties adjusting `_string_or_data_element` so that `test_bytes_string` succeeds. Maybe needs a closer look at a later stage.

anthrotype:
I don't think it makes sense to keep supporting `use_builtin_types` option. I know it is still supported by the stdlib `plistlib` module, but the only use for it is when supporting a legacy py2.py3 codebase. Since fonttools is now py3 only, we never want to _not_ use builting types (i.e. we always want to treat bytes as binary data and str as unicode text, so we don't need that Data class to distinguish bytes-as-binary-data from bytes-as-ascii-text). Just get rid of it

anthrotype:
though I'm not sure if any external code is using our plistlib module with `use_builting_types=False`. We may want to add a deprecation warning for another release cycle before we actually remove the code.

madig:
How should `_string_or_data_element` behave then in the face of `b"hello"`?

anthrotype:
get rid of `_string_or_data_element` and just register `_data_element` for the bytes data type.
But as I said, we may wish to hold on and just do a deprecation warning, then later we'll remove the `use_builtin_types` and the `Data` class handling in the next release.

anthrotype:
you'd also get rid of the `test_bytes_string` test because bytes is never considered to be a (ascii text) string now, only ever binary data.

madig:
Hm, should I do this in a separate PR? It does seem to be an API breaking change, even though no one should ever use that part of the API.

anthrotype:
yeah, that's what I meant. for now, only a deprecation warning if `use_builtin_types` is set to False

Issue #1749, opened 2019-10-18
<designspace format="4.0"> vs. <designspace format="3">


twardoch:
Is there any info about what is valid or invalid in designspace documents with format 4.0 vs. 3?

PR #1750, opened 2019-10-18
[WIP] [designspaceLib] designspace docs edits

justvanrossum:
- misc clarifications and edits from Erik
- proposed addition to allow rules be applied _after_ other features. Relates to #1747 and #1371

### DIFF
--- a/Doc/source/designspaceLib/readme.rst
+++ b/Doc/source/designspaceLib/readme.rst
@@ -14,8 +14,6 @@ Python API
 
 
 
-
-
 .. _designspacedocument-object:
 
 DesignSpaceDocument object
@@ -79,7 +77,7 @@ Methods
    location. Returns None if there isn't one.
 -  ``normalizeLocation(aLocation)``: return a dict with normalized axis values.
 -  ``normalize()``: normalize the geometry of this designspace: scale all the
-  locations of all masters and ins tnces to the ``-1 - 0 - 1`` value.
+  locations of all masters and instances to the ``-1 - 0 - 1`` value.
 -  ``loadSourceFonts()``: Ensure SourceDescriptor.font attributes are loaded,
    and return list of fonts.
 -  ``tostring(encoding=None)``: Returns the designspace as a string. Default 
@@ -123,8 +121,7 @@ Attributes
 -  ``copyLib``: bool. Indicates if the contents of the font.lib need to
    be copied to the instances. MutatorMath.
 -  ``copyInfo`` bool. Indicates if the non-interpolating font.info needs
-   to be copied to the instances. Also indicates this source is expected
-   to be the default font. MutatorMath + Varlib
+   to be copied to the instances. MutatorMath
 -  ``copyGroups`` bool. Indicates if the groups need to be copied to the
    instances. MutatorMath.
 -  ``copyFeatures`` bool. Indicates if the feature text needs to be
@@ -166,6 +163,7 @@ InstanceDescriptor object
 =========================
 
 .. attributes-1:
+
 
 Attributes
 ----------
@@ -269,10 +267,9 @@ AxisDescriptor object
 -  ``default``: number. The default value for this axis, i.e. when a new
    location is created, this is the value this axis will get in user
    space. MutatorMath + Varlib.
--  ``map``: list of input / output values that can describe a warp of user space
-   to design space coordinates. If no map values are present, it is assumed user
-   space is the same as design space, as in [(minimum, minimum), (maximum, maximum)].
-   Varlib.
+-  ``map``: list of input / output values that can describe a warp
+   of user space to design space coordinates. If no map values are present, it is assumed user space is the same as design space, as
+   in [(minimum, minimum), (maximum, maximum)]. Varlib.
 
 .. code:: python
 
@@ -666,7 +663,7 @@ Example
 -  MutatorMath uses the ``glyphs`` element to describe how certain
    glyphs need different masters, mainly to describe the effects of
    conditional rules in Superpolator.
--  Location in designspace coordinates.   
+-  Location in designspace coordinates.
 
 .. attributes-8:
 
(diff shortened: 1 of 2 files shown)


justvanrossum:
Would the new `processing` attribute in the `rules` element need a format version update? Say 4.1?

(See also #1749)

Doc/source/designspaceLib/readme.rst
```diff
@@ -28,87 +29,71 @@ adding them to the document. This makes it easy to integrate this object
...
-
--  The source with the ``copyInfo`` flag indicates this is the default
-   font.
--  In mutatorMath the default font is selected automatically. A warning
-   is printed if the mutatorMath default selection differs from the one
-   set by ``copyInfo``. But the ``copyInfo`` source will be used.
--  If no source has a ``copyInfo`` flag, mutatorMath will be used to
-   select one. This source gets its ``copyInfo`` flag set. If you save
-   the document this flag will be set.
--  Use ``doc.checkDefault()`` to set the default font.
```
belluzj: Is this bit of information going to be moved to another part of the document? Or is it removed completely?

Doc/source/designspaceLib/readme.rst
```diff
@@ -846,15 +844,28 @@ Example
...
+glyphname pairs: the glyphs that need to be substituted. For a rule to be triggered
+**only one** of the conditionsets needs to be true, ``OR``. Within a conditionset 
+**all** conditions need to be true, ``AND``.
+
+
+Attributes
+----------
+
+-  ``processing``: flag, optional. Valid values are [``first``, ``last``]. This flag indicates whether the substitution rules should be applied before or after other glyph substitution features.
+-  If no ``processing`` attribute is given, interpret as ``first``.
```
anthrotype: in the other thread @twardoch raised a valid point about the undefined meaning of processing=post when generating instance UFOs from a designspace. Should the spec say something here?

justvanrossum: The "UFO instances" section about rules currently doesn't say anything about other glyph substitutions, so I think a general clarification of how to deal with that would be good.

justvanrossum: An important detail is (under 5.0 rules element) "The rules are evaluated in this order". This is currently not true with varLib.featureVars, and I'm not sure it should be. The order is only relevant if rules react on each other (eg. rule N+1 acts on the output of rule N).

While this may be fixable (generate substitution lookups in the order of the original rules), I'm not sure what the use case is.

justvanrossum: (Perhaps this discussion deserves its own issue.)

Doc/source/designspaceLib/readme.rst
LettError: Maybe these need to moved to a paragraph marked **deprecated**, rather than be removed?

LettError: I'm sorry for removing it without proper discussion, I guess I was trying to clean the document up. The motives for removing this text were:

* Finding or assigning the default font in a designspace used to be more informal and automatic, so there was code to support that. But the role of the default has become much more prominent and it should be assigned explicitly.
* This was formalised a while ago when the default is understood to be the font located on the (mapped!) default value on each axis. The author of the designspace has to make sure this is the case, there are no reliable methods of guessing what the default should be.
* Assigning the default by setting the **copyInfo** flag then only creates ambiguity. 

I could add a paragraph explaining that there may be older designspaces in which the **copyInfo** flag could mean that the source was intended to be the default.

justvanrossum: > I could add a paragraph explaining that there may be older designspaces in which the copyInfo flag could mean that the source was intended to be the default.

That sounds perfect, thanks.

Doc/source/designspaceLib/readme.rst
```diff
@@ -28,87 +29,71 @@ adding them to the document. This makes it easy to integrate this object
...
-different objects, as long as they have the same attributes.
+different objects, as long as they have the same attributes. Reader and
+Writer objects can be subclassed as well.
+
+.. example-1:
 
 .. code:: python
 
-    from designSpaceDocument import DesignSpaceDocument
+    from designspaceLib import DesignSpaceDocument
```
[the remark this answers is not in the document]
LettError: lgtm

Doc/source/designspaceLib/readme.rst
belluzj: Thanks for the explanation! So just to make sure I understand, the clarified spec is:
* the default font (from a point of view of the origin of deltas) is the source whose coordinates match the default of each axis
* the copyInfo flag:
  * if absent, copy info from the default font
  * if present, copy info from the source that has the flag, but don't change the default source for deltas

Is that correct?

LettError: > the default font (from a point of view of the origin of deltas) is the source whose coordinates match the default of each axis
the copyInfo flag

Correct

> the `copyInfo` flag:
> * if absent, copy info from the default font
> * if present, copy info from the source that has the flag, but don't change the default source for deltas

Would there still be use-cases for taking the info from another source?

twardoch: I think there will be such use-cases. Let’s say you have 2 axes with 4 real, designed masters at the corners (wght 100 wdth 75, wght 900 wdth 75, wght 100 wdth 125, wght 900 wdth 125). You provide the non-interpolating font info in one UFO (e.g. the wght 100 wdth 75), similarly to where you have your lib, groups and features. That’s convenient. 

Now, you may use build DesignSpace A with just those 4 corner masters and use fontmake to produce a bunch of UFO instances, one of them being at wght 400 wdth 100. Then you derive DesignSpace B from DS A, where you set the axes defaults to wght 400 wdth 100, and use the wght 400 wdth 100 instance as your new neutral masters, plus 4 more instances as on-axis masters. So you now have 9 masters, but it’s still convenient to use the wght 100 wdth 75 master as the source of font info, even because you don’t want to bother if the software wrote all the right font info into the generated UFO instance which will now serve as your neutral master. 

Finally, you build your variable font using DesignSpace B. 

Currently, https://github.com/fonttools/fonttools/tree/master/Doc/source/designspaceLib#sourcedescriptor-object says: 

> copyInfo bool. Indicates if the non-interpolating font.info needs to be copied to the instances. Also indicates this source is expected to be the default font. MutatorMath + Varlib

I think it’d be best to reduce copyInfo to the “original” function:

> copyInfo bool. Indicates if the non-interpolating font.info needs to be copied to the instances. MutatorMath + Varlib

anthrotype: Howerver, keep in mind that when producing a VF, varLib (neither its client, fontmake) has never honoured the copyInfo flag in its original meaning of "copying the non-interpolating font.info to the instances" because that doesn't apply for a VF output (varLib takes TTF masters, the meaning of "info" is undefined in this case). varLib only ever understood it as a marker for identifying the neutral master. So if we reduce copyInfo to its original function, we should not mention varLib.

LettError: @twardoch I have no problem coming up with theoretical entangled designspaces. But given the limited amount of time on this earth, I'd argue we should just consider deprecating the `copyFlag`.

schriftgestalt: From my point of view the copyInfo (and copyFeatrues...) tag makes total sense I would strongly encourage to keep it. 

Most of the reason was explained by Adam. 

Defining where to get the info and what outlines are used as a default in a variable font has nothing to do with each other. 

The default master in the axis is defined in variable font coordinates but that masters can have a different range. So I need to first go through all axis, collect all default coordinates, transform them to the master coordinate space and hope that I find a master with that coordinates. This is to complicated and introduces to many possibilities for failure. 

This is as wrong as using the copyInfo flag to define the origin of the variable font.

twardoch:
On the other hand, using the axes default makes .designspace potentially more useful. We can potentially have software that uses .designspace files plus a compiled VF to produce “slices” (subset) etc. I know it’s a bit annoying at first but I think the axes default method is conceptually cleaner.

Doc/source/designspaceLib/readme.rst
twardoch: On the other hand, using the axes default makes .designspace potentially more useful. We can potentially have software that uses .designspace files plus a compiled VF to produce “slices” (subset) etc. I know it’s a bit annoying at first but I think the axes default method is conceptually cleaner.

LettError: The default master on the default values of the axes is very clear and clear. It can be a problem that these values are in userspace coordinates. As long as there is no axis map this is fine because user space == design space. But adding a map can easily move the expected location of the default. That just means editors need to keep track of that. I updated mine.

madig: To come back to the second part of Jany's question, after the decision to keep `copy*` in:

The `copy*` flags:
- if absent, copy `*` from the default font.
- if present, copy `*` from the source that has the flag if you feel like it, or don't, but don't change the source for the default outlines (which is _always_ determined by the axis defaults).

Is that correct?

LettError: Correct.

PR #1752, opened 2019-10-24
enforce VarStore RegionAxisCount == fvar.AxisCount

anthrotype:
... even when the ItemVariationStore contains no Regions and thus we can't take the length of their VarRegionAxis arrays.

This is to appease older versions of OTS (< 8.0.0) which blindly enforce this rule and reject a VF that has, e.g., an empty HVAR table with no regions when the HVAR.VarStore.VarRegionList.RegionAxisCount != fvar.AxisCount.

Fixes https://github.com/fonttools/fonttools/issues/1670

Related:
https://github.com/fonttools/fonttools/pull/1671
https://github.com/googlefonts/fontmake/issues/565
https://github.com/khaledhosny/ots/pull/192

### DIFF
--- a/Lib/fontTools/ttLib/tables/otBase.py
+++ b/Lib/fontTools/ttLib/tables/otBase.py
@@ -498,6 +498,11 @@ class OTTableWriter(object):
 		return OverflowErrorRecord( (self.tableTag, LookupListIndex, SubTableIndex, itemName, itemIndex) )
 
 
+class LiteralCount(int):
+	"""A count value that should be taken literally, rather than recomputed on compile."""
+	pass
+
+
 class CountReference(object):
 	"""A reference to a Count value, not a count of references."""
 	def __init__(self, table, name, size=None, value=None):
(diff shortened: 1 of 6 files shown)


behdad:
There might be an easier fix to just see the Count if unset. Maybe in a preWrite method or something, without adding the StaticCount hack.

behdad:
Or just always take current value of count var and use if never set.

behdad:
Not sure what bad implications that might have though.

anthrotype:
I'm not sure what you mean exactly..
At least the way I did it makes sure that the static count and the computed one are synchronized (see the assert in CountReference.setValue).

behdad:
> At least the way I did it makes sure that the static count and the computed one are synchronized (see the assert in CountReference.setValue).

I know that.  So, you just need to make a setValue call with the value from fvar.  Right?  That's what I'm trying to say.  I thought that should be doable in a shorter patch than what you've got.

behdad:
Yeah okay I see what you are doing.  Fine with me.  The name StaticCount is confusing.  What about you just reuse CountReference?  So in the preWrite you set a CountReference which is already initialized to the value from fvar.  No need for new type.

PR #1754, opened 2019-10-25
checkAxis and checkDefault were removed, but not from doc

benkiel:
Update docs for the removal in 3.27.0


madig:
Rebased. The readme.rst file was probably removed in the big DSv5 update.

PR #1757, opened 2019-10-30
[pens] Improve error message if pen.moveTo() is omitted.

justvanrossum:
See robotools/fontParts#470


justvanrossum:
Or should it be a "proper" error? If so, which one? I'm happy enough with an assert myself.

anthrotype:
LGTM. Maybe instead of "no current contour" would it be good to say "contour missing required initial moveTo" or something like that

anthrotype:
i think assert is ok. it is a code error

PR #1762, opened 2019-11-21
ufoLib groupnames conversion 2 → 3

### LINKED ISSUE #1761: [ufoLib] Forced up-conversion to UFO3
When opening a UFO2 generated by `vfb2ufo`, it is forced to UFO3. There I run into an error with upgrading the kerning group names. The code responsible for this is basically the UFO sample implementation, so I opened an issue there: https://github.com/unified-font-object/ufo-spec/issues/98

Any way from the ufoLib side to avoid the ambiguity around how group names are extracted from the kerning items?

Or, because my UFO process is one-way only, I would be happy to open the UFO keeping version 2, then processing the groups and throwing out the ambiguous names before converting up to 3. But it seems a forced upgrade to version 3 now.

jenskutilek:
Pass glyphSet to `convertUFO1OrUFO2KerningToUFO3Kerning` so it can exclude glyph names from kerning group renaming. Fixes #1761.

### DIFF
--- a/Lib/fontTools/ufoLib/converters.py
+++ b/Lib/fontTools/ufoLib/converters.py
@@ -6,9 +6,7 @@ Conversion functions.
 
 # adapted from the UFO spec
 
-def convertUFO1OrUFO2KerningToUFO3Kerning(kerning, groups, glyphSet=None):
-    if glyphSet is None:
-        glyphSet = set()
+def convertUFO1OrUFO2KerningToUFO3Kerning(kerning, groups, glyphSet=()):
     # gather known kerning groups based on the prefixes
     firstReferencedGroups, secondReferencedGroups = findKnownKerningGroups(groups)
     # Make lists of groups referenced in kerning pairs.
--- a/Lib/fontTools/ufoLib/converters.py
+++ b/Lib/fontTools/ufoLib/converters.py
@@ -6,7 +6,9 @@ Conversion functions.
 
 # adapted from the UFO spec
 
-def convertUFO1OrUFO2KerningToUFO3Kerning(kerning, groups, glyphSet=set()):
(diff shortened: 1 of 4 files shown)


anthrotype:
thanks. got a test too?

anthrotype:
the tests are failing because the KerningUpConversionTestCase setup code is not also creating the required "glyphs" subfolder in the test UFO.

Lib/fontTools/ufoLib/converters.py
```diff
@@ -6,16 +6,16 @@
 
 # adapted from the UFO spec
 
-def convertUFO1OrUFO2KerningToUFO3Kerning(kerning, groups):
+def convertUFO1OrUFO2KerningToUFO3Kerning(kerning, groups, glyphSet):
```
justvanrossum: For maximum bw compatibility, maybe give `glyphSet` a default value of `{}` or `set()`?

jenskutilek: Sure!

jenskutilek:
I fixed the test and also addes some pairs that would trigger the original problem.

Lib/fontTools/ufoLib/converters.py
```diff
@@ -6,16 +6,16 @@
 
 # adapted from the UFO spec
 
-def convertUFO1OrUFO2KerningToUFO3Kerning(kerning, groups):
+def convertUFO1OrUFO2KerningToUFO3Kerning(kerning, groups, glyphSet=set()):
```
khaledhosny: Using a mutable as default argument is adviced against in Python.

anthrotype: good point. An empty tuple `()` or `frozenset()` will do

anthrotype: even though it is not modified, but only read by the function. still a good idea

PR #1763, opened 2019-11-26
ttGlyphPen must round float coordinates and offsets

anthrotype:
otherwise the computed glyph xMin, yMin, xMax and yMax bounds can be computed incorrectly.

https://github.com/googlefonts/fontmake/issues/593

See the failing test.

### DIFF
--- a/Lib/fontTools/pens/ttGlyphPen.py
+++ b/Lib/fontTools/pens/ttGlyphPen.py
@@ -1,6 +1,6 @@
 from fontTools.misc.py23 import *
 from array import array
-from fontTools.misc.fixedTools import MAX_F2DOT14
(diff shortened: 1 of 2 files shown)


anthrotype:
here's the test failure showing that the composite glyph's bounds are compute incorrectly if the components have float coordinates and/or their x/y offsets are floats 

https://travis-ci.org/fonttools/fonttools/jobs/617287097#L606-L609

In order for recalcBounds method to work correctly, the simple glyphs' coordinates and the components' offsets need to be integers. It's not sufficient to only round at the end (in calcIntBounds) because components may be translated, and subtracting two floats and then rounding is not always the same as subtracting their rounded values.

Thus, when TTGlyphPen receives float coordinates and offsets, we need to round them to integer.

PR #1766, opened 2019-11-28
[pointPens] Add {Filter,Recording,Transform,Rounding}PointPen

anthrotype:
This PR adds a few point pens that are equivalent to the corresponding (segment) pens:

- FilterPointPen: passes through points/components unchanged to another 'slave' pen; mostly used as base class for more interesting filter pens, such as..
- TransformPointPen: a filter point pen that takes an Affine transforms and another point pen;
- RoundingPointPen: filter point pen that rounds point coordinates and components' offsets to integer;
- RecordingPointPen: records and replays point-pen commands (beginPath, addPoint, endPath)

### DIFF
--- a/Lib/fontTools/pens/filterPen.py
+++ b/Lib/fontTools/pens/filterPen.py
@@ -1,12 +1,13 @@
 from fontTools.misc.py23 import *
 from fontTools.pens.basePen import AbstractPen
+from fontTools.pens.pointPen import AbstractPointPen
 from fontTools.pens.recordingPen import RecordingPen
 
 
 class _PassThruComponentsMixin(object):
 
-    def addComponent(self, glyphName, transformation):
-        self._outPen.addComponent(glyphName, transformation)
+    def addComponent(self, glyphName, transformation, **kwargs):
+        self._outPen.addComponent(glyphName, transformation, **kwargs)
 
 
 class FilterPen(_PassThruComponentsMixin, AbstractPen):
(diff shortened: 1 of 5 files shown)


justvanrossum:
Have you looked at the RecordingPointPen and TransformPointPen implementations from https://github.com/robotools/fontPens/tree/master/Lib/fontPens?

anthrotype:
oops.. I didn't 😅 
lemme have a look and see if I can merge them with mine

anthrotype:
Ok, the TransformPointPen and RecordingPointPen now work the same as the ones from fontPens package. The only difference with my original implementation was that I included `identifier` in the positional argument and didn't have a slot for the extra `kwargs`. But I agree it's safer not to assume identifier is a positional arg (as it was introduced later), and it's good to keep the kwargs.

I haven't added the `replayRecording` method. Do we need one? Probably not.

justvanrossum:
> I haven't added the replayRecording method. Do we need one? Probably not.

I added that for symmetry with recordingPen.py, but I don't think I've ever used it.

What would be a great is a PR for fontPens that imports the ones from fonttools, and issues a deprecation warning upon import. It's a bit silly to keep both.

anthrotype:
> What would be a great is a PR for fontPens that imports the ones from fonttools, and issues a deprecation warning upon import. It's a bit silly to keep both.

Sure, that can be done once these are merged.
Let me know if you have other comments, or I'll go on with the merge.

PR #1767, opened 2019-12-02
[feaLib] Warning instead of error for duplicate identical substitutions

jenskutilek:
We have lots of source files which contain substitutions that do nothing or are duplicated because classes are reused for different features. Those fail compilation with feaLib currently.

 I think if an identical substitution is added to a feature, it could be ignored instead of raising an error. The problem is only when a duplicate source glyph would introduce an ambiguity, i.e. the code would demand to replace one glyph with two different target glyphs.

Example:

```
@fig = [one two three Euro];
@figOldstyle = [one.osf two.osf Euro];
@figSC = [one.sc two.sc Euro.sc];

feature onum {
  sub @fig by @figOldstyle;
} onum;

feature c2sc {
  sub @fig by @figSC;
  sub @figOldstyle by @figSC; # <-- adds sub Euro by Euro.sc; again
} cs2c;
```

Which is equal to:

```
feature onum {
  sub [one two three Euro] by [one.osf two.osf Euro];
} onum;

feature c2sc {
  sub [one two three Euro] by [one.sc two.sc Euro.sc];
  sub [one.osf two.osf Euro] by [one.sc two.sc Euro.sc]; # <-- adds sub Euro by Euro.sc; again
} cs2c;
```

Adobe FDK logs an info in this situation and continues compilation.

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -917,8 +917,8 @@ class Builder(object):
                 if to_glyph == lookup.mapping[from_glyph]:
                     log.info(
                         'Removing duplicate single substitution from glyph'
-                        ' "%s" to "%s" at %s:%i:%i' %
-                        (from_glyph, to_glyph, *location),
+                        ' "%s" to "%s" at %s:%i:%i',
+                        from_glyph, to_glyph, *location,
                     )
                 else:
                     raise FeatureLibError(
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -914,10 +914,19 @@ class Builder(object):
         lookup = self.get_lookup_(location, SingleSubstBuilder)
         for (from_glyph, to_glyph) in mapping.items():
             if from_glyph in lookup.mapping:
-                raise FeatureLibError(
-                    'Already defined rule for replacing glyph "%s" by "%s"' %
-                    (from_glyph, lookup.mapping[from_glyph]),
-                    location)
+                if to_glyph == lookup.mapping[from_glyph]:
+                    # log warning?
+                    # FDK logs "[NOTE] Removing duplicate single substitution in 'xxxx' feature: from_glyph, to_glyph" in this case
+                    log.info(FeatureLibError(
+                        'Removing duplicate single substitution for glyph "%s" by "%s"' %
+                        (from_glyph, to_glyph),
+                        location
+                    ))
+                else:
+                    raise FeatureLibError(
+                        'Already defined rule for replacing glyph "%s" by "%s"' %
+                        (from_glyph, lookup.mapping[from_glyph]),
+                        location)
             lookup.mapping[from_glyph] = to_glyph
 
     def add_single_subst_chained_(self, location, prefix, suffix, mapping):
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -915,13 +915,11 @@ class Builder(object):
         for (from_glyph, to_glyph) in mapping.items():
             if from_glyph in lookup.mapping:
                 if to_glyph == lookup.mapping[from_glyph]:
-                    # log warning?
-                    # FDK logs "[NOTE] Removing duplicate single substitution in 'xxxx' feature: from_glyph, to_glyph" in this case
-                    log.info(FeatureLibError(
-                        'Removing duplicate single substitution for glyph "%s" by "%s"' %
-                        (from_glyph, to_glyph),
-                        location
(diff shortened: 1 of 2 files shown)


jenskutilek:
Hm, now I see there's a flaw in my example that I just made up. Will have to dig for a real example from our files.

jenskutilek:
We want to be able to switch default figures by just assigning the Unicode values to a different set of figures, so we include substitutions for all possible figure sets in the figure features.

```
@Fig_LinTab = [one.LinTab uni20B4];
@Fig_OnumTab = [one.OnumTab uni20B4];
@Fig_lowTab = [one.lowTab uni20B4];
@Fig_sc.tab = [one.sc.tab uni20B4.sc];
@Fig_LinProp = [one uni20B4];
@Fig_oldstyle = [one.oldstyle uni20B4];
@Fig_lowProp = [one.lowProp uni20B4];
@Fig_sc.prop = [one.sc.prop uni20B4.sc];

feature onum { # Old Style Numerals
# tabular
	sub @Fig_LinTab by @Fig_OnumTab;
	sub @Fig_lowTab by @Fig_OnumTab;
	sub @Fig_sc.tab by @Fig_OnumTab;
# proportional
	sub @Fig_LinProp by @Fig_oldstyle;
	sub @Fig_lowProp by @Fig_oldstyle;
	sub @Fig_sc.prop by @Fig_oldstyle;
} onum;
```

This is equal to:

```
feature onum { # Old Style Numerals
# tabular
	sub [one.LinTab uni20B4] by [one.OnumTab uni20B4];
	sub [one.lowTab uni20B4] by [one.OnumTab uni20B4];
	sub [one.sc.tab uni20B4.sc] by [one.OnumTab uni20B4];
# proportional
	sub [one uni20B4] by [one.oldstyle uni20B4];
	sub [one.lowProp uni20B4] by [one.oldstyle uni20B4];
	sub [one.sc.prop uni20B4.sc] by [one.oldstyle uni20B4];
} onum;
```

You see, the duplications are unnecessary, but not ambiguous.

FDK just says:

```
[NOTE] <XYZ-Light> Removing duplicate single substitution in 'onum' feature: uni20B4, uni20B4
```

Lib/fontTools/feaLib/builder.py
```diff
@@ -914,10 +914,19 @@ def add_single_subst(self, location, prefix, suffix, mapping, forceChain):
         lookup = self.get_lookup_(location, SingleSubstBuilder)
         for (from_glyph, to_glyph) in mapping.items():
             if from_glyph in lookup.mapping:
-                raise FeatureLibError(
-                    'Already defined rule for replacing glyph "%s" by "%s"' %
-                    (from_glyph, lookup.mapping[from_glyph]),
-                    location)
+                if to_glyph == lookup.mapping[from_glyph]:
+                    # log warning?
+                    # FDK logs "[NOTE] Removing duplicate single substitution in 'xxxx' feature: from_glyph, to_glyph" in this case
```
anthrotype: logging.info is fine. Also, we don't need these comments

Lib/fontTools/feaLib/builder.py
```diff
@@ -914,10 +914,19 @@ def add_single_subst(self, location, prefix, suffix, mapping, forceChain):
...
+                    log.info(FeatureLibError(
```
anthrotype: i believe you're using the FeatureLibError exception so that it logs the line number, right?
if so make it more explicit, e.g. turn the exception into a string first and pass a string to log.info, instead of the exception object.

justvanrossum:
Would it be possible to add a test for this?

jenskutilek:
I added a simple test that checks if the warning is logged.

Lib/fontTools/feaLib/builder.py
```diff
@@ -914,10 +914,17 @@ def add_single_subst(self, location, prefix, suffix, mapping, forceChain):
...
             if from_glyph in lookup.mapping:
-                raise FeatureLibError(
-                    'Already defined rule for replacing glyph "%s" by "%s"' %
-                    (from_glyph, lookup.mapping[from_glyph]),
-                    location)
+                if to_glyph == lookup.mapping[from_glyph]:
+                    log.info(
+                        'Removing duplicate single substitution from glyph'
+                        ' "%s" to "%s" at %s:%i:%i' %
+                        (from_glyph, to_glyph, *location),
```
anthrotype: please do this

```diff
-                        ' "%s" to "%s" at %s:%i:%i' %
-                        (from_glyph, to_glyph, *location),
+                        ' "%s" to "%s" at %s:%i:%i',
+                        from_glyph, to_glyph, *location
```

anthrotype:
Thank you Jens!

PR #1772, opened 2019-12-06
[ttLib.glyf] Make sure to use the flagOnCurve mask in glyph.draw()

### LINKED ISSUE #1771: Bad font data after instantiating variable font
I'm coming across bad font data when using fontTools pens to interact with instantiated variable fonts. The problem can be reproduced in every contour in every Quadratic variable font I've tried.

When reading the point data with a pen, the point order is offset by one (which in this case I don't mind) and the last two `lineTo` points are joined together into a `qCurve` segment (which distorts the shape). The problem happens in both the `instancer` and `mutator` variations of the `instantiateVariableFont` function.

However, the font data in the TTF looks fine to me when I save it to a file and open it in an editor or read through a TTX dump.

This demo just happens to use Monserrat (to make sure that the problem wasn't in my own font) and I set the location to the default weight to make the output comparison more clear, but the problem happens at any location (and with any Quadratic TTF)

Can anyone think of what I might be doing wrong, or where the problem might be coming from?

```python
from fontTools.varLib import mutator, instancer
from fontTools.pens.recordingPen import RecordingPen


def printReport(f):
    g = f.getGlyphSet()["I"]
    pen = RecordingPen()
    g.draw(pen)
    for i in pen.value:
        print(i)  

fontPath = "Montserrat-VF.ttf"
location = {"wght":100}

print("Default:")
f = TTFont(fontPath)
printReport(f)

print("\nmutator.instantiateVariableFont")
instM = mutator.instantiateVariableFont(f, location)
#instM.save("mutator.ttf")
printReport(instM)

print("\ninstancer.instantiateVariableFont")
instI = instancer.instantiateVariableFont(f, location)
#instI.save("instancer.ttf")
printReport(instI)
```

Output:

```
Default:
('moveTo', ((133, 700),))
('lineTo', ((153, 700),))
('lineTo', ((153, 0),))
('lineTo', ((133, 0),))
('lineTo', ((133, 700),))
('closePath', ())

mutator.instantiateVariableFont
('moveTo', ((153, 700),))
('lineTo', ((153, 0),))
('lineTo', ((133, 0),))
('qCurveTo', ((133, 700), (153, 700)))
('closePath', ())

instancer.instantiateVariableFont
('moveTo', ((153, 700),))
('lineTo', ((153, 0),))
('lineTo', ((133, 0),))
('qCurveTo', ((133, 700), (153, 700)))
('closePath', ())
```

justvanrossum:
...so we don't trip on the overlap flag, that is set when instantiating variable fonts. Fixes #1771.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -1179,7 +1179,7 @@ class Glyph(object):
 		for end in endPts:
 			end = end + 1
 			contour = coordinates[start:end]
-			cFlags = flags[start:end]
+			cFlags = [flagOnCurve & f for f in flags[start:end]]
 			start = end
 			if 1 not in cFlags:
 				# There is not a single on-curve point on the curve,
--- a/Tests/ttLib/tables/_g_l_y_f_test.py
+++ b/Tests/ttLib/tables/_g_l_y_f_test.py
@@ -288,6 +288,8 @@ class glyfTableTest(unittest.TestCase):
     def test_bit6_draw_to_pen_issue1771(self):
         # https://github.com/fonttools/fonttools/issues/1771
         font = TTFont(sfntVersion="\x00\x01\x00\x00")
+        # glyph00003 contains a bit 6 flag on the first point,
+        # which triggered the issue
         font.importXML(GLYF_TTX)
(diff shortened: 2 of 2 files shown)


anthrotype:
Thanks for the quick fix!

justvanrossum:
Please don't merge yet, making a test!

PR #1775, opened 2019-12-08
[ttLib.glyf] make glyph.draw() skip final lineTo()

justvanrossum:
In glyphs from the `glyf` table, glyph.draw() has always emitted a final `pen.lineTo()` back to the first point. This is redundant.

However, this may be a breaking change for client code, so I'm not sure it's wise to do this.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -1198,7 +1198,10 @@ class Glyph(object):
 				while contour:
 					nextOnCurve = cFlags.index(1) + 1
 					if nextOnCurve == 1:
-						pen.lineTo(contour[0])
+						# Skip a final lineTo(), as it is implied by
+						# pen.closePath()
+						if len(contour) > 1:
+							pen.lineTo(contour[0])
 					else:
 						pen.qCurveTo(*contour[:nextOnCurve])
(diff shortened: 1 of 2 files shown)


anthrotype:
oh, i had missed this one.
I think this change is safe, because it doesn't really change the drawn results. The segment pen API is clear that closePath will add an implied lineTo the closed contour's moveTo point if the last and first points in the contour have different coordinates.
So, LGTM

justvanrossum:
Thanks. Btw. this PR also ensures that

`glyph.draw(pen)`

now has exactly the same effect as

`glyph.drawPoints(PointToSegmentPen(pen))`

which I think is also a good thing. Let me see if I can enhance the test to reflect that.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -1198,7 +1198,10 @@ def draw(self, pen, glyfTable, offset=0):
 				while contour:
 					nextOnCurve = cFlags.index(1) + 1
 					if nextOnCurve == 1:
-						pen.lineTo(contour[0])
+						# Skip a final lineTo(), as it is implied by
+						# pen.closePath()
+						if len(contour) > 1:
+							pen.lineTo(contour[0])
```
behdad: Can we skip lineTo if the end point is same as start point already?

PR #1782, opened 2019-12-12
[subset] remap FeatureVariations SubstitutionRecord.FeatureIndex

### LINKED ISSUE #1777: [subset] FeatureVariations subsetted incorrectly
Take this font by @arrowtype which contains loads of GSUB rvrn FeatureVariations substitutions:
https://github.com/arrowtype/recursive/releases/download/1.027/recursive-MONO_CASL_wght_slnt_ital--full_gsub--2019_11_30-20_07.ttf

If you run `fonttools subset --unicodes='*'` on it, the "rvrn" feature and all its lookups are dropped, even though "rvrn" is in the subsetter's default list of layout features to keep. Also, if one explicitly adds "rvrn" to the `--layout-features+=rvrn` option, still it is being subsetted away.

I think the problem is that the `prune_pre_subset` stage is modifying the GSUB.table.FeatureList.FeatureRecord array (dropping feature records for tags not included in the options.layout_features). But then later, since the FeatureVariations references the feature records by their indexes and these may have changed inthe pre-subset pruning, the FeatureVariations record may be incorrectly dropped.

I have this patch that fixes the problem, but before I send a PR I want to make a good test case that reproduces the issue. I'll work on that later.

```diff
diff --git a/Lib/fontTools/subset/__init__.py b/Lib/fontTools/subset/__init__.py
index 581c3948..fea85f0e 100644
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -1307,6 +1307,9 @@ def subset_features(self, feature_indices):
        self.ensureDecompiled()
        self.SubstitutionRecord = [r for r in self.SubstitutionRecord
                                     if r.FeatureIndex in feature_indices]
+       feature_index_map = dict(zip(feature_indices, range(len(feature_indices))))
+       for r in self.SubstitutionRecord:
+               r.FeatureIndex = feature_index_map[r.FeatureIndex]
        self.SubstitutionCount = len(self.SubstitutionRecord)
        return bool(self.SubstitutionCount)

```

Thanks @arrowtype for reporting this bug! :+1

anthrotype:
Fixes #1777

The 'rvrn' feature is being incorrectly dropped because the feature indexes change if some other feature occuring before 'rvrn' is dropped. The feature indexes need to be remapped.

I add a failing test case to show the issue, then I'll follow up with the actual patch.

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -1307,6 +1307,9 @@ def subset_features(self, feature_indices):
 	self.ensureDecompiled()
 	self.SubstitutionRecord = [r for r in self.SubstitutionRecord
 				     if r.FeatureIndex in feature_indices]
+	# remap feature indices
+	for r in self.SubstitutionRecord:
+		r.FeatureIndex = feature_indices.index(r.FeatureIndex)
 	self.SubstitutionCount = len(self.SubstitutionRecord)
 	return bool(self.SubstitutionCount)
 
--- a/Tests/subset/subset_test.py
+++ b/Tests/subset/subset_test.py
@@ -1,5 +1,7 @@
+import io
 from fontTools.misc.py23 import *
 from fontTools import subset
+from fontTools.fontBuilder import FontBuilder
 from fontTools.ttLib import TTFont, newTable
 from fontTools.misc.loggingTools import CapturingLogHandler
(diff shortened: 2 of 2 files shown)


behdad:
lgtm as well.  Thanks!

PR #1786, opened 2019-12-16
ufoLib: support UFO formatVersionMinor and GLIF formatMinor

anthrotype:
As discussed in https://github.com/unified-font-object/ufo-spec/issues/78, this PR adds support for reading/writing a new `formatVersionMinor` element in `metainfo.plist` as well as a `formatMinor` attribute to the `glyph` element of the GLIF data structure.
The idea is to be able to add new data to the UFO spec without breaking existing implementations, (which should be able to ignore this new data).


anthrotype:
This PR is now ready for review, I cleaned it up, rebased and added tests.
If we like this, I will work on a PR to update the UFO spec to have an additional `formatVersionMinor` element in metainfo.plist (0 if omitted), and an additional `formatMinor` attribute for the `<glyph>` element in GLIF format (also defaulting to 0).

Summary of the changes:
* when reading metainfo.plist, check for a `formatVersionMinor` (int), assume 0 if missing.
* if the parsed formatVersion (major, minor) tuple is unsupported, raise UnsupportedUFOFormat exception (subclass of UFOLibError) when validate=True (default); if validate=False, assume the latest supported version and warn that some data may be skipped or parsed incorrectly.
* in UFOWriter constructor, allow to pass formatVersion as a tuple (major, minor) in addition to a single int (i.e. major only, minor=0); raise UnsupportedUFOFormat if specified formatVersion is unknown. If no explicit formatVersion, default to latest;
* UFOReader and UFOWriter `formatVersion` property is deprecated but kept for backward compatibility; you should now use `formatVersionTuple` which return a tuple-enum object (with minor and major attributes).
* glifLib.GlyphSet similarly accepts a tuple as ufoFormatVersion; added a `ufoFormatVersionTuple` attribute, but kept old `ufoFormatVersion` (only returning the major int) to avoid breaking clients that may use that.
* when parsing GLIF, check for a `formatMinor` attribute of glyph element (defaults to 0 if missing); if the (major, minor) format tuple thus parsed is unsupported raise UnsupportedGLIFFormat exception (subclass of GlifLibError) when validate=True (default); if validate=False, assume the latest supported version and log a warning that data may be skipped/parsed incorrectly.
* when serializing glyph element, a `formatMinor` attribute is added (in addition to the existing `format` attribute) but only if version tuple.minor is != 0
* GlyphSet.writeGlyph and writeGlyphToString can take optional formatVersion as a tuple.
* In readGlyphFromString, the set of allowed formatVersions can contain either ints or tuples.

Note that I didn't define any UFO 3.1 or GLIF 2.1 in this PR yet. This is only to allow us in the future to be able to make backward compatible changes to the UFO spec by bumping the minor version of the format.

Lib/fontTools/ufoLib/utils.py
```diff
@@ -36,6 +36,39 @@ def wrapper(*args, **kwargs):
     return deprecated_decorator
 
 
+# To be mixed with enum.Enum in UFOFormatVersion and GLIFFormatVersion
+class _VersionTupleEnumMixin(object):
```
justvanrossum, line 40: No need for `(object)` here.

anthrotype: right. bad py23 habit :)

anthrotype:
@typemytype can you please check out this PR branch https://github.com/anthrotype/fonttools/tree/ufo-minor-format-version and see if we didn't break anything in RoboFont, as @benkiel suggested?

anthrotype:
@typemytype you should expect no changes at all, since I didn't define any new UFO 3.x or GLIF 2.x in this PR, the new element/attribute is optional on reading and only written when != 0. Once we will have a minor UFO/GLIF update we can specify that in the spec.

typemytype:
To many fontTools issues, so this one got lost... (is there way to set it up I got notifications only when Im mentioned, instead of all fontTools notifications from github)

maybe you should send a PR to defcon as a warning is now popping up. Defcon already requires the latest fontTools with this change...

> DeprecationWarning: The 'formatVersion' attribute is deprecated; use the 'formatVersionTuple'

thanks!

typemytype:
ehm, Im looking at patching this change to defcon and fontParts and I have some issues with comparing. It should be able to compare with ints, tuples and UFOFormatVersion objects

```
from fontTools.ufoLib import UFOFormatVersion

m = UFOFormatVersion(1)
print(m == 1) # this is False, but that is not true

m = UFOFormatVersion(1)
print(m == (1, 0)) # this is True, thats true

m = UFOFormatVersion(1)
print(m == UFOFormatVersion(1)) # this is True, thats true

m = UFOFormatVersion(1)
print(m == UFOFormatVersion((1, 0)) # this is True, thats true
```

> The idea is to be able to add new data to the UFO spec without breaking existing implementations, (which should be able to ignore this new data).

it breaks when comparing ufo format version... lot of scripts have a `defconFont.save(... formatVersion=2)` this will fail while overwriting existing ufo with a different format version.

It could be solved by improving __eq__, __ne__ by allowing comparing the tuple format with an int. Only the major version should be compared then.

anthrotype:
It shouldn't fail because UFO writer still accepts the integer form (and converts automatically to a version tuple) and similarly the reader has the formatVersion property returning the major integer just like before. The deprecation warning is to nudge applications to use the tuple form from now onwards, but anything existing should continue to work by passing/getting integer versions to ufoLib API.
The additional formatVersionTuple attribute will now return a 2-tuple, you can treat it and compare it as such. When you only need to access the major (or minor) version you can by using the major/minor getters. So this works like a namedtuple if you like.
I would prefer to keep the old and new formatVersion formats distinct (instead of adding comparison methods like you suggest), and eventually drop support for the old integer only format.
In defcon you can always convert incoming integer-only formatVersion values from the user into a UFOFormatVersion tuple, or even just a tuple tuple and let ufoLib do the casting.

typemytype:
this is endless...

a simple `__eq__` could solve this...

fe in: 

https://github.com/adobe-type-tools/afdko/blob/172fea0f689940a2a23125dcfe0d204a1f282a8d/python/afdko/checkoutlinesufo.py#L68-L70

```python
from fontTools.ufoLib import UFOFormatVersion as UFOFormatVersionUnableToCompareWithAnythingExecptItselfAndIsUnalbeToBeSubclassed
from fontTools.ufoLib.utils import _VersionTupleEnumMixin
import enum


class UFOFormatVersion(tuple, _VersionTupleEnumMixin, enum.Enum):
    FORMAT_1_0 = (1, 0)
    FORMAT_2_0 = (2, 0)
    FORMAT_3_0 = (3, 0)
	    
    def __eq__(self, other):
        if isinstance(other, (int, float)):
            other = self.__class__(other)
        return super().__eq__(other)
        
    
    
version = UFOFormatVersion(3)
print(version)
print(version == 3)
print(3 == version)

cannotCompareVersion = UFOFormatVersionUnableToCompareWithAnythingExecptItselfAndIsUnalbeToBeSubclassed(3)

print(cannotCompareVersion)
print(cannotCompareVersion == 3)
print(3 == cannotCompareVersion)
```

typemytype:
obviously the same for `__lt__`, `__le__`, `__gt__`, `__ge__`

typemytype:
more places where comparison will go wrong, but could be fixed easily...

https://github.com/LettError/ufoProcessor/blob/master/Lib/ufoProcessor/__init__.py#L310

@LettError

justvanrossum:
I think this PR was not correct: https://github.com/robotools/defcon/pull/289.

You can't just replace the existing `decon.Font().ufoFormatVersion` attribute with the new UFOFormatVersion object and expect everything to work in a 100% bw-compatible way. For exactly those reasons ufoLib does _not_ take that approach.

Either you do as ufoLib, and provide a _new_ attribute, eg. called `font.ufoFormatVersionTuple`, or you provide a new attribute called `font.ufoFormatVersionMinor`.

typemytype:
It should, otherwise its a breaking change for fe. defcon.... the promise was not a breaking change

In the near future defcon also would like to know if data is according the (3, 1) spec... and being able to compare

justvanrossum:
> It should, otherwise its a breaking change for fe. defcon.... the promise was not a breaking change

It should what exactly? It's the defcon PR that is the breaking change.

typemytype:
no, the PR in defcon tries to work around the deprecated warning in fontTools, related to the use of `ufoFormatVersion`, solving that moves the breaking change to other packages... 

and this could be easily be solved to improve comparing UFOFormatVersions, I dont see the issue with by adding support for comparing with ints... the object already takes simple ints as argument and adds missing minor version... 

a non breaking change means **nothing** has to be changed in other packages, this is not the case

justvanrossum:
So a deprecation warning makes an otherwise non-breaking change a breaking change? I think that's an unrealistic position.

It's possible to adapt defcon to take advantage of the new format version scheme that is backwards compatible (as per my suggestion). You chose not to do that, and now blame fonttools. That's just not right.

> I dont see the issue with by adding support for comparing with ints...

I think it this would be a terrible design decision, but apart from that: it wouldn't even solve everything. There are other ways client code can expect an int (say, write it to json or plist) which would then still break. Again, the reasons why ufoLib doesn't do it that way (but opted for a new attribute) are applicable to defcon just the same.

The approach of ufoLib is: the old attribute stays _exactly_ the same (int), and if you want to take advantage of the minor version you have to use the tuple variant of the format version attribute. defcon should do the same, then the problem is solved. Old code just works, but yeah, will face a deprecation warning, which I think is a fair price for a change like this.

typemytype:
dont get it, why  a deprecated warning? this is a signal for other packages to make change... cause it will break...

justvanrossum:
It's a new API, and we want people to move to the new API. A deprecation warning doesn't mean the old API is going to go away soon, it just means there's something better now, and please use it if you can.

It's a new API because it's not possible to guarantee backwards compatibility by changing the existing API.

Would you have preferred to keep the old API but have it break in subtle ways that are hard to predict? That's pretty much what happened in the defcon PR, and why I think that was the wrong thing to do.

typemytype:
Im just saying it can be solved differently :) and the solution now is not so friendly for others... 

fe a breaking change now: (without that PR in defcon) imagine opening a UFO in the near future with format `(3, 1)`, will save now back to `3` which is not good. 

and moving to a new api has a big impact, bigger then anticipated, bigger then the change in GLIFFormatVersion as a lot of tools are comparing, the goal of these attributes is to compar :) IMO this can solved differently and easily. 

and I dont like a deprecated warning each time I open a UFO, and no I dont want to import warning each time too

why does the `UFOFormatVersion` take a single int as argument? 
looking at similar version objects: `distutils.version` objects can always be compared with `str`, init argument is a `str` --> compare is possible with a str. 

For `UFOFormatVersion` init argument is an `int` --> compare should be possible with an int. But here we are making circles :)

justvanrossum:
> IMO this can solved differently and easily.

This is where we disagree, as it would give way more subtle breakage down the line.

The kind of magic you're proposing always comes back to bite you when you least expect it.

typemytype:
its idd a kind of sorcery where you dont have to hold the wings while flying, but trust the plane to be attached to wings...

I dont see a difference in comparing:
```python
UFOFormatVersion(3) == (3, 0)
UFOFormatVersion(3) == 3
```
why allowing comparing with a tuple?

(now my open source time for this month and next month is over)

justvanrossum:
> why allowing comparing with a tuple?

Because it _is_ a tuple.

Eventually all client code will have to explicitly deal with the ufo format version having two components. What you're proposing encourages client code to not care about this ever.

This is just no way to make progress. If the depreciation warning bothers you so much, I invite you to pin fonttools to an older version. But I guess you'd like to take advantage of the other progress in fonttools as well...

typemytype:
ahah, euh that is what I tried in the defcon PR, until I noticed it has a bigger impact, moving the problem is not solving it...

justvanrossum:
So, defcon exposes ufoLib's format version.

The UFO spec changes so that there can be a minor component of the format version. ufoLib is (obviously) the first to implement this, and comes to the (correct) conclusion that it's not possible to do this in a 100% bw-compatible way without introducing a _new_ API, that will eventually replace the old API. To make client code aware of this, a deprecation warning is issued, as is the usual way of dealing with these kinds of things.

Defcon can choose to abstract this away by applying magic (eg. defcon could provide its own tuple subclass that will do the comparisons how you like them), but that won't invalidate ufoLib's conclusion that such an abstraction is ultimately the wrong thing to do.

Defcon is a library, too, and I see no problem with it adding a new API, too, and deprecating the old one. It's directly exposing a ufoLib API to defcon clients after all, so defcon clients will have to deal with this eventually, too. To be honest I don't understand how you think defcon can get away with something that ufoLib consciously decided is too risky to do to _its_ clients.

Everyone is free to ignore deprecation warnings. Demanding a non-breaking change is one thing, but demanding it with without deprecation warnings is moving the goal posts.

typemytype:
yes defcon exposes ufoLib `UFOFormatVersion`, like it does for the `fileStructure`. 

Defcon cannot make it own subclass as these enum objects are not subclassable... (see the end of the example)

Still dont see the issue why this is so bad.

this example warns AND has a correct comparison result!! the best of both worlds!! 

it does the same thing like on init of a UFOFormatVersion: converting an int to a tuple. 

This can be removed in a couple of years, when all packages depending on this have updated to a tuple version world. 

```python
from fontTools.ufoLib import UFOFormatVersion as UFOFormatVersionUnableToCompareWithAnythingExecptItselfAndIsUnalbeToBeSubclassed
from fontTools.ufoLib.utils import _VersionTupleEnumMixin
import enum

class _CompareMixin:
    
    def __comparableUFOFormatVersionObject(self, other):
        if isinstance(other, (int, float)):
            print("warning here, but allow it, but still the result of the comparison will be correct!!!")
            other = self.__class__(other)
        return other
        
    def __eq__(self, other):
        return super().__eq__(self.__comparableUFOFormatVersionObject(other))
    
    def __ne__(self, other):
        return super().__ne__(self.__comparableUFOFormatVersionObject(other))
    
    def __lt__(self, other):
        return super().__lt__(self.__comparableUFOFormatVersionObject(other))

    def __le__(self, other):
        return super().__le__(self.__comparableUFOFormatVersionObject(other))
    
    def __gt__(self, other):
        return super().__gt__(self.__comparableUFOFormatVersionObject(other))
            
    def __ge__(self, other):
        return super().__ge__(self.__comparableUFOFormatVersionObject(other))


class UFOFormatVersion(_CompareMixin, tuple, _VersionTupleEnumMixin, enum.Enum):
    FORMAT_1_0 = (1, 0)
    FORMAT_2_0 = (2, 0)
    FORMAT_3_0 = (3, 0)        
    


version = UFOFormatVersion(3)
print(version)
print(version)
print(version == 3)
print(3 != version)
print((3, 6) != version)

cannotCompareVersion = UFOFormatVersionUnableToCompareWithAnythingExecptItselfAndIsUnalbeToBeSubclassed(3)

print(cannotCompareVersion)
print("this should be True")
print(cannotCompareVersion == 3)
print("this should be False")
print(3 != cannotCompareVersion)

# this fails: TypeError: Cannot extend enumerations
# class MySubclass(UFOFormatVersionUnableToCompareWithAnythingExecptItselfAndIsUnalbeToBeSubclassed):
#     pass
```

justvanrossum:
First you say the deprecation warning is the deal breaker, and now you introduce a warning in your own proposal. That makes no sense, and you may as well just deal with the existing deprecation warning.

I give up arguing, but I stand by the current design in ufoLib and recommend against adopting comparisons this way. This is ultimately @anthrotype's call.

I also still recommend against defcon adopting this kind of implicit conversion, too, but that's your call.

Btw.:

> Defcon cannot make it own subclass as these enum objects are not subclassable

I didn't say subclass UFOFormatVersion, I said "provide its own tuple subclass". You can cast UFOFormatVersion to a tuple.

PR #1788, opened 2019-12-19
Remove workaround (#1381)

jenskutilek:
Apparently the macOS bug has been fixed. I'm not sure when, but in 10.14.6 the workaround introduced in #1381 is not needed anymore. Can anybody confirm?

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -325,7 +325,6 @@ def _add_gvar(font, masterModel, master_ttfs, tolerance=0.5, optimize=True):
 
     for glyph in font.getGlyphOrder():
         log.debug("building gvar for glyph '%s'", glyph)
-        isComposite = glyf[glyph].isComposite()
 
         allData = [
             m.glyf._getCoordinatesAndControls(glyph, m.hMetrics, m.vMetrics)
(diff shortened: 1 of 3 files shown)


jenskutilek:
Anecdotally, removing the workaround saves me 11kB (61kB to 50kB) on one of the TTFs from https://github.com/jenskutilek/homecomputer-fonts.

anthrotype:
> in 10.14.6 the workaround introduced in #1381 is not needed anymore

That's great to hear. 

> removing the workaround saves me 11kB (61kB to 50kB)

Hm that's a lot, but I am inclined to keep this in by default, given that 10.14 is relatively recent.
Maybe we could add a command-line option to fonttools varLib to disable this.

behdad:
2.5 years later. Let's decide when is a good time to remove the workaround.

justvanrossum:
Since it was a problem introduced with macOS 10.14, and was apparently fixed latest in 10.14.6, the time to remove the workaround should be now...

Lorp:
See https://github.com/fonttools/fonttools/issues/1381#issuecomment-1217119263 about a current macOS bug that appears to hit glyphs containing components with zero deltas.

anthrotype:
one year later, what shal we do here?

> a current macOS bug that appears to hit glyphs containing components with zero deltas.

@Lorp what happened to that ever since?

anthrotype:
5 years later, now really is time to remove the workaround

PR #1796, opened 2020-01-06
Get rid of SyntaxWarning with py3.8.

anntzer:
On a clean clone (make sure there are no pycs present yet),
```
python -c 'import fontTools.ttLib.tables._c_m_a_p'
```
results in
```
.../fontTools/ttLib/tables/_c_m_a_p.py:21: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if gid is 0:
```

Fix that.


justvanrossum:
Thanks!

PR #1809, opened 2020-01-24
[varLib] demonstrate and fix #1808

### LINKED ISSUE #1808: [varLib] merging OTL fails if master TTFonts are loaded with lazy=True
If I populate my designspace source descriptors with TTFont() objects, similar to this:

```python
for source in ds.sources:
    source.font = TTFont(<ttfsource>, lazy=True)
```
I get an AssertionError deep down in `mergeThings()`:

```
    ttFont = varLib.build(self.doc, exclude=['MVAR', 'HVAR', 'VVAR', 'STAT'])
../fonttools_justvanrossum/Lib/fontTools/varLib/__init__.py:880: in build
    _merge_OTL(vf, model, master_fonts, axisTags)
../fonttools_justvanrossum/Lib/fontTools/varLib/__init__.py:601: in _merge_OTL
    merger.mergeTables(font, master_fonts, ['GSUB', 'GDEF', 'GPOS'])
../fonttools_justvanrossum/Lib/fontTools/varLib/merger.py:123: in mergeTables
    for m in master_ttfs])
../fonttools_justvanrossum/Lib/fontTools/varLib/merger.py:1013: in mergeThings
    super(VariationMerger, self).mergeThings(out, lst)
../fonttools_justvanrossum/Lib/fontTools/varLib/merger.py:108: in mergeThings
    self.mergeObjects(out, lst)
../fonttools_justvanrossum/Lib/fontTools/varLib/merger.py:80: in mergeObjects
    mergerFunc(self, value, values)
../fonttools_justvanrossum/Lib/fontTools/varLib/merger.py:1013: in mergeThings
    super(VariationMerger, self).mergeThings(out, lst)
../fonttools_justvanrossum/Lib/fontTools/varLib/merger.py:108: in mergeThings
    self.mergeObjects(out, lst)
../fonttools_justvanrossum/Lib/fontTools/varLib/merger.py:80: in mergeObjects
    mergerFunc(self, value, values)
../fonttools_justvanrossum/Lib/fontTools/varLib/merger.py:1013: in mergeThings
    super(VariationMerger, self).mergeThings(out, lst)
../fonttools_justvanrossum/Lib/fontTools/varLib/merger.py:108: in mergeThings
    self.mergeObjects(out, lst)
../fonttools_justvanrossum/Lib/fontTools/varLib/merger.py:80: in mergeObjects
    mergerFunc(self, value, values)
../fonttools_justvanrossum/Lib/fontTools/varLib/merger.py:1013: in mergeThings
    super(VariationMerger, self).mergeThings(out, lst)
../fonttools_justvanrossum/Lib/fontTools/varLib/merger.py:108: in mergeThings
    self.mergeObjects(out, lst)
../fonttools_justvanrossum/Lib/fontTools/varLib/merger.py:80: in mergeObjects
    mergerFunc(self, value, values)
../fonttools_justvanrossum/Lib/fontTools/varLib/merger.py:1013: in mergeThings
    super(VariationMerger, self).mergeThings(out, lst)
```
```
E      AssertionError: (({'name': <'name' table at 12b0066d0>, 'fvar': <'fvar' table at 12b006610>, 'GPOS': <'GPOS' table at 11cc64850>}, [{'name': <'name' table at 12af841d0>, 'GPOS': <'GPOS' table at 12b00ffd0>}, {'GPOS': <'GPOS' table at 12b00fcd0>}, {'GPOS': <'GPOS' table at 12b019590>}, {'GPOS': <'GPOS' table at 12b019850>}]), 'dict', '.tables', 'TTFont', '.font', 'FeatureList', '.FeatureList', 'GPOS', '.table', 'table_G_P_O_S_')
```

This is because the `FeatureList` object has `font` and `reader` attributes that the merger ties to merge, but can't. When the font is not loaded in "lazy mode", those attrs don't exist and all is well.

I wonder if it can have anything to do with how `deepcopy()` works on the TTFont instance:
https://github.com/fonttools/fonttools/blob/a37dab38243d8c1e88ab83376a80942012bbab48/Lib/fontTools/varLib/__init__.py#L852

I don't have  a ready reproducer yet, will see if I can make one.

justvanrossum:
This change triggers the problem described in #1808.

### DIFF
--- a/Lib/fontTools/designspaceLib/__init__.py
+++ b/Lib/fontTools/designspaceLib/__init__.py
@@ -1322,14 +1322,3 @@ class DesignSpaceDocument(LogMixin, AsDictMixin):
                 loaded[source.path] = source.font
             fonts.append(source.font)
         return fonts
-
-    def closeSourceFonts(self):
-        for source in self.sources:
-            if source.font is not None and hasattr(source.font, "close"):
-                source.font.close()
-
-    def __enter__(self):
-        return self
-
-    def __exit__(self, type, value, traceback):
-        self.closeSourceFonts()
--- a/Lib/fontTools/designspaceLib/__init__.py
+++ b/Lib/fontTools/designspaceLib/__init__.py
@@ -1325,7 +1325,7 @@ class DesignSpaceDocument(LogMixin, AsDictMixin):
 
     def closeSourceFonts(self):
         for source in self.sources:
-            if source.font is not None:
+            if source.font is not None and hasattr(source.font, "close"):
                 source.font.close()
 
     def __enter__(self):
--- a/Lib/fontTools/designspaceLib/__init__.py
+++ b/Lib/fontTools/designspaceLib/__init__.py
@@ -1322,3 +1322,14 @@ class DesignSpaceDocument(LogMixin, AsDictMixin):
                 loaded[source.path] = source.font
             fonts.append(source.font)
         return fonts
+
+    def closeSourceFonts(self):
+        for source in self.sources:
+            if source.font is not None:
+                source.font.close()
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, type, value, traceback):
+        self.closeSourceFonts()
--- a/Lib/fontTools/varLib/merger.py
+++ b/Lib/fontTools/varLib/merger.py
@@ -60,6 +60,11 @@ class Merger(object):
 		return _default
 
 	def mergeObjects(self, out, lst, exclude=()):
+		if hasattr(out, "ensureDecompiled"):
+			out.ensureDecompiled()
+		for item in lst:
+			if hasattr(item, "ensureDecompiled"):
+				item.ensureDecompiled()
 		keys = sorted(vars(out).keys())
 		assert all(keys == sorted(vars(v).keys()) for v in lst), \
 			(keys, [sorted(vars(v).keys()) for v in lst])
@@ -87,11 +92,6 @@ class Merger(object):
 
 	def mergeThings(self, out, lst):
 		try:
-			if hasattr(out, "ensureDecompiled"):
-				out.ensureDecompiled()
-			for item in lst:
-				if hasattr(item, "ensureDecompiled"):
-					item.ensureDecompiled()
 			assert allEqualTo(out, lst, type), (out, lst)
 			mergerFunc = self.mergersFor(out).get(None, None)
 			if mergerFunc is not None:
--- a/Lib/fontTools/varLib/merger.py
+++ b/Lib/fontTools/varLib/merger.py
@@ -87,6 +87,11 @@ class Merger(object):
 
 	def mergeThings(self, out, lst):
 		try:
+			if hasattr(out, "ensureDecompiled"):
+				out.ensureDecompiled()
+			for item in lst:
+				if hasattr(item, "ensureDecompiled"):
+					item.ensureDecompiled()
 			assert allEqualTo(out, lst, type), (out, lst)
 			mergerFunc = self.mergersFor(out).get(None, None)
 			if mergerFunc is not None:
--- a/Tests/varLib/varLib_test.py
+++ b/Tests/varLib/varLib_test.py
@@ -483,7 +483,7 @@ class BuildTest(unittest.TestCase):
 
         ds_path = self.get_test_input("SparseMasters.designspace")
         ds = DesignSpaceDocument.fromfile(ds_path)
-        master_fonts = load_masters(ds)
+        load_masters(ds)
 
         # Trigger MVAR generation so varLib is forced to create deltas with a
         # sparse master inbetween.
--- a/Tests/varLib/varLib_test.py
+++ b/Tests/varLib/varLib_test.py
@@ -485,83 +485,79 @@ class BuildTest(unittest.TestCase):
         ds = DesignSpaceDocument.fromfile(ds_path)
         master_fonts = load_masters(ds)
 
-        try:
-            # Trigger MVAR generation so varLib is forced to create deltas with a
-            # sparse master inbetween.
-            font_0_os2 = ds.sources[0].font["OS/2"]
-            font_0_os2.sTypoAscender = 1
-            font_0_os2.sTypoDescender = 1
-            font_0_os2.sTypoLineGap = 1
-            font_0_os2.usWinAscent = 1
-            font_0_os2.usWinDescent = 1
-            font_0_os2.sxHeight = 1
-            font_0_os2.sCapHeight = 1
-            font_0_os2.ySubscriptXSize = 1
-            font_0_os2.ySubscriptYSize = 1
-            font_0_os2.ySubscriptXOffset = 1
-            font_0_os2.ySubscriptYOffset = 1
-            font_0_os2.ySuperscriptXSize = 1
-            font_0_os2.ySuperscriptYSize = 1
-            font_0_os2.ySuperscriptXOffset = 1
-            font_0_os2.ySuperscriptYOffset = 1
-            font_0_os2.yStrikeoutSize = 1
-            font_0_os2.yStrikeoutPosition = 1
-            font_0_vhea = newTable("vhea")
-            font_0_vhea.ascent = 1
-            font_0_vhea.descent = 1
-            font_0_vhea.lineGap = 1
-            font_0_vhea.caretSlopeRise = 1
-            font_0_vhea.caretSlopeRun = 1
-            font_0_vhea.caretOffset = 1
-            ds.sources[0].font["vhea"] = font_0_vhea
-            font_0_hhea = ds.sources[0].font["hhea"]
-            font_0_hhea.caretSlopeRise = 1
-            font_0_hhea.caretSlopeRun = 1
-            font_0_hhea.caretOffset = 1
-            font_0_post = ds.sources[0].font["post"]
-            font_0_post.underlineThickness = 1
-            font_0_post.underlinePosition = 1
+        # Trigger MVAR generation so varLib is forced to create deltas with a
+        # sparse master inbetween.
+        font_0_os2 = ds.sources[0].font["OS/2"]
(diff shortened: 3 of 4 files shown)


Lib/fontTools/varLib/merger.py
```diff
@@ -87,6 +87,11 @@ def mergeLists(self, out, lst):
 
 	def mergeThings(self, out, lst):
 		try:
+			if hasattr(out, "ensureDecompiled"):
```
anthrotype: they should always have `ensureDecompiled` as they are all subclasses of otTables.BaseTable, no?

anthrotype:
appveyor build failures seem unrelated, we can ignore here and fix separately

anthrotype:
actually, maybe they are _not_ unrelated, since you set `lazy=True` by default in varLib, hence the tests need to releases the filesystem resources before shutil.rmtree can clean up on windows

Lib/fontTools/varLib/merger.py
justvanrossum: No, `mergeThings()` also gets called with the toplevel GDEF/GPOS/GSUB table objects, which do not have it. It's probably true once we're at the `font["GDEF"].table` level, though.

justvanrossum:
> actually, maybe they are not unrelated, since you set lazy=True by default in varLib, hence the tests need to releases the filesystem resources before shutil.rmtree can clean up on windows

Yeah, that entirely possible :( Please advice on how to fix.

Lib/fontTools/varLib/merger.py
anthrotype: oh, I see. We might as well add `ensureDecompiled` to BaseTTXConverter.

anthrotype:
you need to make sure we call `TTFont.close` method (or use `with ...` self-closing context)

Lib/fontTools/varLib/merger.py
justvanrossum: I just tried your suggestion, but it doesn't work, as `mergeThings()` also gets called with ints, lists, None, and maybe more.

anthrotype: ok.. that's why the method name is generic "things". hasattr is fine then

justvanrossum:
Yeah but where? I think this is about the `.font` attribute on SourceDescriptor instances, and SourceDescriptor doesn't know about that attr.

The `lazy=True` flag can easily be removed from that spot, but then we need an alternative test for #1808.

anthrotype:
what I don't get is why one needs to manually call `ensureDecompiled` at all. Supposedly all otTables are decompiled lazily as soon as you get any attribute (see `BaseTable.__getattr__`).

justvanrossum:
> what I don't get is why one needs to manually call ensureDecompiled at all. Supposedly all otTables are decompiled lazily as soon as you get any attribute (see BaseTable.__getattr__).

Same here...

anthrotype:
> I think this is about the .font attribute on SourceDescriptor instances, and SourceDescriptor doesn't know about that attr.

then we may need a `closeSourceFonts` method on DesignSpaceDocument class, to match the existing `loadSourceFonts`

justvanrossum:
> then we may need a closeSourceFonts method on DesignSpaceDocument class, to match the existing loadSourceFonts

Ok. How about making DesignSpaceDocument work as context manage, too?

anthrotype:
the only filesystem resource that DesignSpaceDocument holds is the TTFont.reader.file (and only when loaded lazy=True). Just reading a DS doesn't set the source.font attributes yet. Only if you call `loadSourceFonts` method you get those. The latter method can't easily be turned into a contextmanager since it returns a list of fonts already. Unless you return an object that behaves both like a list of fonts and has `__enter__` and `__exit__` maybe..

justvanrossum:
Right. I went ahead and tried the DSD context manager, mainly to see if it fixes the Windows failures.

anthrotype:
but we aren't consistent with using `loadSourceFonts` either. I see some tests set the source.font attribute directly. Maybe your solution to make the whole DS object a context (no-op in case no source.fonts) is ok.
With hindight, I don't like too much that passing a DS object to varLib.build changes its internal state.

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -1322,3 +1322,14 @@ def loadSourceFonts(self, opener, **kwargs):
                 loaded[source.path] = source.font
             fonts.append(source.font)
         return fonts
+
+    def closeSourceFonts(self):
+        for source in self.sources:
+            if source.font is not None:
+                source.font.close()
```
anthrotype: defcon Font objects don't have a close method... SourceDescriptor.font can be UFO objects too. You may want to check if hasattr(source.font, "close")...

justvanrossum:
Agreed.

I wonder – if you want to go ahead with a DS context manager – whether `varLib.build` should then also use it.

anthrotype:
sure, makes sense that whomever loads the fonts (in this case varLib) ensures they get closed.
maybe the close method should also set source.font back to None?

anthrotype:
> Unless you return an object that behaves both like a list of fonts and has __enter__ and __exit__ maybe..

that's not a bad idea. E.g.

```python
ds = DesignSpaceDocument.fromfile("...")
with ds.loadSourceFonts(opener=TTFont, lazy=True) as master_fonts:
    # do stuff with master_fonts

# auto close them on exit
```

justvanrossum:
> sure, makes sense that whomever loads the fonts (in this case varLib) ensures they get closed.

That's not necessarily always the case: in my project I'm passing a ds document to varLib.build that has source.font already populated. But in my case I wouldn't mind if `varLib.build` would close them.

anthrotype:
... and using `UserList` to wrap the auto-closing list of fonts returned by loadSourceFonts
https://docs.python.org/3/library/collections.html#userlist-objects

anthrotype:
> in my project I'm passing a ds document to varLib.build that has source.font already populated.

right. so varLib should only close what it has opened itself. if the source.font was already populated by user, it's their responsibility to close it. It's a bit like what we do with when you init a TTFont from an already opened file object or a path.

justvanrossum:
However, `varLib.build` can't really know who opened the fonts...

anthrotype:
the custom list returned from loadSourceFonts could remember which source.font were not None and which ones where actually opened by it, and close only the latter.
If you think it gets too complicated, ignore this.

anthrotype:
> varLib.build can't really know who opened the fonts

FWIW varLib.load_masters uses DesignSpaceDocument.loadSourceFonts

justvanrossum:
Alternative idea: we can add a `close_source_fonts arg to `varib.build` that is True by default, which tells `varLib.build` to call close all items in `master_fonts` once it's done. In a try/finally handler. Then if the user really does _not_ want the source fonts to be closed, they have to ask for that explicitly.

Less automatic, but it will make this PR much much smaller...

anthrotype:
closing all is most likely what one wants, so i'm ok with that. I'll explore the possibility of returning a context manager list from loadSourceFonts antother time.

Tests/varLib/varLib_test.py
```diff
@@ -558,6 +559,9 @@ def test_varlib_build_sparse_masters_MVAR(self):
             varfont, _, _ = build(ds)
             mvar_tags = [vr.ValueTag for vr in varfont["MVAR"].table.ValueRecord]
             assert all(tag in mvar_tags for tag in fontTools.varLib.mvar.MVAR_ENTRIES)
+        finally:
+            for master in master_fonts:
+                master.close()
```
anthrotype: aren't these being closed by default inside varLib.build? why you need to close them again

justvanrossum: Ahh, you're right. I saw `load_masters` and didn't look far enough to see that `build` was actually called.

anthrotype:
> if the user really does not want the source fonts to be closed, they have to ask for that explicitly.

then, if one forgets to set it to False, and attempts to run varLib.build more than once on the same DS object (e.g. with different parameters), the second time it will probably fail because the fonts won't be reloaded and were irreversibly closed in the previous run, right?

justvanrossum:
> then, if one forgets to set it to False, and attempts to run varLib.build more than once on the same DS object (e.g. with different parameters), the second time it will probably fail because the fonts won't be reloaded and were irreversibly closed in the previous run, right?

I'm afraid that is indeed the logical consequence of the current state of the PR. Hm, what to do?

anthrotype:
for your use case (where you set source.font attrs directly in your client app), wouldn't it be enough to make that `ensureDecompiled` change inside mergeObjects? Why do we need to change the default behavior of varLib.build (loading lazy=True and always closing fonts, even when user provided and possibly not loaded lazily to begin with)?
I would just make the first change, and leave varLib.build alone. Then you could add a distinct test to verify that it correctly decompiles the lazy fonts.
Sorry if we're going in circles...

justvanrossum:
I'd be totally fine with that (and indeed I proposed it before), except right now I'm out of energy to come up with a separate test for that.

anthrotype:
we could add that later, it's fine. thanks

anthrotype:
We still need to investigate why the ensureDecompiled is needed at all. Maybe @behdad will have some insight

Lib/fontTools/varLib/merger.py
```diff
@@ -60,6 +60,11 @@ def mergersFor(celf, thing, _default={}):
 		return _default
 
 	def mergeObjects(self, out, lst, exclude=()):
+		if hasattr(out, "ensureDecompiled"):
+			out.ensureDecompiled()
+		for item in lst:
+			if hasattr(item, "ensureDecompiled"):
+				item.ensureDecompiled()
 		keys = sorted(vars(out).keys())
```
anthrotype: @justvanrossum I see why we need to do `ensureDecompiled`. It's because we call `vars(out)`, but the latter probably only gets the object's `__dict__` object, so `__getattr__` method is not triggered and the lazy decompilation doesn't occur.

PR #1811, opened 2020-01-26
[feaLib] Do not fail on duplicate multiple substitutions

dscorbett:
This is like #1767 but for multiple substitutions.


anthrotype:
thanks!

PR #1814, opened 2020-01-29
[feaLib] fix mixed single/multiple substitutions

khaledhosny:
If the single substitution involved a glyph class, we were incorrectly using only the first glyph in the class.

Broken since ec6ff821f0e72022d7aec8794b6bb589d8f81808, apparently no one else uses this feature!


khaledhosny:
I had some rule mystetiously being working partially after porting from makeotf to feaLib, otherwise I would have never noticed this.

PR #1815, opened 2020-01-29
[feaLib] Don’t reset lookupflag in nested lookups

khaledhosny:
In makeotf, lookups defined inside feature blocks inherit the current `lookupflag` of the feature, so don’t reset the `lookupflag` for such lookups.


anthrotype:
Thanks. While I agree we should match makeotf as much as possible, I wonder if this behavior is defined in the FEA spec somewhere?

khaledhosny:
AFAICT, this is not documented but I’ll try to open an AFDKO issue for it.

PR #1816, opened 2020-01-31
[varLib] Allow using CFF2 table as source

khaledhosny:
Nothing clever, if the source font has a `CFF2` table use it as if it were a `CFF` table (i.e. non-variable, not sure what would happen in the source `CFF2` was variable already).

Also some other small fixes to bounds recalculation that I found while testing this.

Draft as it needs some tests.


Lib/fontTools/ttLib/tables/_h_e_a_d.py
```diff
@@ -65,6 +65,19 @@ def compile(self, ttFont):
 			if 'CFF ' in ttFont:
 				topDict = ttFont['CFF '].cff.topDictIndex[0]
 				self.xMin, self.yMin, self.xMax, self.yMax = intRect(topDict.FontBBox)
+			if 'CFF2' in ttFont:
```
adrientetar: elif?

khaledhosny: Good point. Probably doesn’t make difference in practice, but good idea nevertheless.

khaledhosny:
The first commit has a test now, the other commits are covered by existing tests.

khaledhosny:
> > non-variable, not sure what would happen in the source CFF2 was variable already
> 
> maybe we could add a check and log a warning if that's the case? I presume the variable data in that case would be ignored/discarded.

Same thing I guess when a font with `gvar` table is given as input. Probably too obscure case to be worried about.

PR #1818, opened 2020-01-31
EBLC: fix padding length calculation for Format 3 IndexSubTable

### LINKED ISSUE #1817: [BUG] Alignment issue with CBLC/EBLC index format 3
According to the spec CBLC/EBLC IndexSubtables using index format 3 need to be 32-bit aligned [1]. However, due to a bug in calculating if padding is required near line 482 of `E_B_L_C_.py` the resulting compiled table will always be 16-bit aligned. See the code for an explanation.

```python
dataList = [EblcIndexSubTable.compile(self, ttFont)]
# At this point len(dataList) == 1 this will be a problem when doing padding calculation.
# The first entry in dataList is already 32-bit aligned as it is:
# uint16 firstGlyphId
# uint16 lastGlyphId
# uint32 imageDataOffset

dataList += [struct.pack(dataFormat, offsetValue) for offsetValue in offsetArray]
# Now len(dataList) == 1 + # of offsets

# Only # of offsets should be considered for padding not the first entry in dataList
# len(dataList) should be len(offsetArray) or len(dataList) - 1
if offsetDataSize * len(dataList) % 4 != 0:
    dataList.append(struct.pack(dataFormat, 0))
```

[1] [Spec for IndexSubtables](https://docs.microsoft.com/en-us/typography/opentype/spec/eblc#indexsubtables)

anthrotype:
Fixes #1817

Ideally we should add a test for this. Currently, the only test that has to do with CBLC/EBLC is for the subset module. The Tests/subset/data/google_color.ttx file however only has format 1 index subtable.
I need to make a test font or modify the existing one.

/cc @ckitagawa-work

### DIFF
--- a/Lib/fontTools/ttLib/tables/E_B_L_C_.py
+++ b/Lib/fontTools/ttLib/tables/E_B_L_C_.py
@@ -482,7 +482,7 @@ def _createOffsetArrayIndexSubTableMixin(formatStringForDataType):
 			dataList = [EblcIndexSubTable.compile(self, ttFont)]
 			dataList += [struct.pack(dataFormat, offsetValue) for offsetValue in offsetArray]
 			# Take care of any padding issues. Only occurs in format 3.
-			if offsetDataSize * len(dataList) % 4 != 0:
+			if offsetDataSize * len(offsetArray) % 4 != 0:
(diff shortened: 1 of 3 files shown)


ckitagawa-work:
Thanks for the fix :)

PR #1820, opened 2020-02-04
[feaLib lexer tests] Compare realpath against realpath

madig:
On Windows, tests may be run in the user's temp directory, so the previous code may compare `C:\Users\nikolaus.waxweiler\` against `C:\Users\NIKOLA~1.WAX\`.


anthrotype:
do we not run tests on windows?

madig:
We do, but this one wasn't caught I guess? Maybe the temp dir is different and unDOSed on Appveyor.

PR #1821, opened 2020-02-04
[varLib] Use proper exceptions for signalling input source errors

madig:
This is a support PR for https://github.com/googlefonts/fontmake/pull/634 and aims to make the varLib variable font building API raise `VarLibError` and derivatives instead of generic exceptions and assertion errors.

I try to differentiate between input errors and programming errors. Input errors should be catchable by fontmake and ideally tell the user what's wrong. Programming errors are unexpected (haha) and should continue to produce stack traces because there is nothing else useful you can do with them.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -79,7 +79,12 @@ def _add_fvar(font, axes, instances):
 		coordinates = instance.location
 
 		if "en" not in instance.localisedStyleName:
-			assert instance.styleName
+			if not instance.styleName:
+				raise VarLibValidationError(
+					f"Instance at location '{coordinates}' must have a default English "
+					"style name ('stylename' attribute on the instance element or a "
+					"stylename element with an 'xml:lang=\"en\"' attribute)."
+				)
 			localisedStyleName = dict(instance.localisedStyleName)
 			localisedStyleName["en"] = tounicode(instance.styleName)
 		else:
@@ -135,20 +140,32 @@ def _add_avar(font, axes):
 		# Current avar requirements.  We don't have to enforce
 		# these on the designer and can deduce some ourselves,
 		# but for now just enforce them.
-		assert axis.minimum == min(keys)
-		assert axis.maximum == max(keys)
-		assert axis.default in keys
-		# No duplicates
-		assert len(set(keys)) == len(keys), (
-			f"{axis.tag} axis: All axis mapping input='...' "
-			"values must be unique, but we found duplicates."
-		)
-		assert len(set(vals)) == len(vals), (
-			f"{axis.tag} axis: All axis mapping output='...' "
-			"values must be unique, but we found duplicates."
-		)
+		if axis.minimum != min(keys):
+			raise VarLibValidationError(
+				f"Axis '{axis.name}': there must be a mapping for the axis minimum "
+				f"value {axis.minimum} and it must be the lowest input mapping value."
+			)
+		if axis.maximum != max(keys):
+			raise VarLibValidationError(
+				f"Axis '{axis.name}': there must be a mapping for the axis maximum "
+				f"value {axis.maximum} and it must be the highest input mapping value."
+			)
+		if axis.default not in keys:
+			raise VarLibValidationError(
+				f"Axis '{axis.name}': there must be a mapping for the axis default "
+				f"value {axis.default}."
+			)
+		# No duplicate input values (output values can be >= their preceeding value).
+		if len(set(keys)) != len(keys):
+			raise VarLibValidationError(
+				f"Axis '{axis.name}': All axis mapping input='...' values must be "
+				"unique, but we found duplicates."
+			)
 		# Ascending values
-		assert sorted(vals) == vals
+		if sorted(vals) != vals:
+			raise VarLibValidationError(
+				f"Axis '{axis.name}': mapping output values must be in ascending order."
+			)
 
 		keys_triple = (axis.minimum, axis.default, axis.maximum)
 		vals_triple = tuple(axis.map_forward(v) for v in keys_triple)
@@ -212,8 +229,8 @@ def _add_stat(font, axes):
 
 
 def _add_gvar(font, masterModel, master_ttfs, tolerance=0.5, optimize=True):
-
-	assert tolerance >= 0
+	if tolerance < 0:
+		raise ValueError("`tolerance` must be a positive number.")
 
 	log.info("Generating gvar")
 	assert "gvar" not in font
@@ -704,6 +721,7 @@ def load_designspace(designspace):
 		raise VarLibValidationError("Designspace must have at least one source.")
 	instances = ds.instances
 
+	# TODO: Use fontTools.designspaceLib.tagForAxisName instead.
 	standard_axis_map = OrderedDict([
 		('weight',  ('wght', {'en': u'Weight'})),
 		('width',   ('wdth', {'en': u'Width'})),
@@ -713,11 +731,15 @@ def load_designspace(designspace):
 		])
 
 	# Setup axes
+	if not ds.axes:
+		raise VarLibValidationError(f"Designspace must have at least one axis.")
+
 	axes = OrderedDict()
-	for axis in ds.axes:
+	for axis_index, axis in enumerate(ds.axes):
 		axis_name = axis.name
 		if not axis_name:
-			assert axis.tag is not None
+			if not axis.tag:
+				raise VarLibValidationError(f"Axis at index {axis_index} needs a tag.")
 			axis_name = axis.name = axis.tag
 
 		if axis_name in standard_axis_map:
@@ -726,7 +748,8 @@ def load_designspace(designspace):
 			if not axis.labelNames:
 				axis.labelNames.update(standard_axis_map[axis_name][1])
 		else:
-			assert axis.tag is not None
+			if not axis.tag:
+				raise VarLibValidationError(f"Axis at index {axis_index} needs a tag.")
 			if not axis.labelNames:
 				axis.labelNames["en"] = tounicode(axis_name)
 
@@ -737,15 +760,28 @@ def load_designspace(designspace):
 	for obj in masters+instances:
 		obj_name = obj.name or obj.styleName or ''
 		loc = obj.location
+		if loc is None:
+			raise VarLibValidationError(
+				f"Source or instance '{obj_name}' has no location."
+			)
 		for axis_name in loc.keys():
-			assert axis_name in axes, "Location axis '%s' unknown for '%s'." % (axis_name, obj_name)
+			if axis_name not in axes:
+				raise VarLibValidationError(
+					f"Location axis '{axis_name}' unknown for '{obj_name}'."
+				)
 		for axis_name,axis in axes.items():
 			if axis_name not in loc:
 				# NOTE: `axis.default` is always user-space, but `obj.location` always design-space.
 				loc[axis_name] = axis.map_forward(axis.default)
 			else:
 				v = axis.map_backward(loc[axis_name])
-				assert axis.minimum <= v <= axis.maximum, "Location for axis '%s' (mapped to %s) out of range for '%s' [%s..%s]" % (axis_name, v, obj_name, axis.minimum, axis.maximum)
+				if not (axis.minimum <= v <= axis.maximum):
+					raise VarLibValidationError(
+						f"Source or instance '{obj_name}' has out-of-range location "
+						f"for axis '{axis_name}': is mapped to {v} but must be in "
+						f"mapped range [{axis.minimum}..{axis.maximum}] (NOTE: all "
+						"values are in user-space)."
+					)
 
 	# Normalize master locations
 
@@ -766,9 +802,15 @@ def load_designspace(designspace):
 	base_idx = None
 	for i,m in enumerate(normalized_master_locs):
 		if all(v == 0 for v in m.values()):
-			assert base_idx is None
+			if base_idx is not None:
+				raise VarLibValidationError(
+					"More than one base master found in Designspace."
+				)
 			base_idx = i
-	assert base_idx is not None, "Base master not found; no master at default location?"
+	if base_idx is None:
+		raise VarLibValidationError(
+			"Base master not found; no master at default location?"
+		)
 	log.info("Index of base master: %s", base_idx)
 
 	return _DesignSpaceData(
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -39,12 +39,10 @@ import os.path
 import logging
 from copy import deepcopy
 from pprint import pformat
+from .errors import VarLibError, VarLibValidationError
 
 log = logging.getLogger("fontTools.varLib")
 
-
-class VarLibError(Exception):
-	pass
 
 #
 # Creation routines
@@ -703,7 +701,7 @@ def load_designspace(designspace):
 
 	masters = ds.sources
 	if not masters:
-		raise VarLibError("no sources found in .designspace")
+		raise VarLibValidationError("Designspace must have at least one source.")
 	instances = ds.instances
 
 	standard_axis_map = OrderedDict([
@@ -927,7 +925,7 @@ def _open_font(path, master_finder=lambda s: s):
 	elif tp in ("TTF", "OTF", "WOFF", "WOFF2"):
 		font = TTFont(master_path)
 	else:
-		raise VarLibError("Invalid master path: %r" % master_path)
+		raise VarLibValidationError("Invalid master path: %r" % master_path)
 	return font
 
 
@@ -947,10 +945,10 @@ def load_masters(designspace, master_finder=lambda s: s):
 		# If a SourceDescriptor has a layer name, demand that the compiled TTFont
 		# be supplied by the caller. This spares us from modifying MasterFinder.
 		if master.layerName and master.font is None:
-			raise AttributeError(
-				"Designspace source '%s' specified a layer name but lacks the "
-				"required TTFont object in the 'font' attribute."
-				% (master.name or "<Unknown>")
+			raise VarLibValidationError(
+				f"Designspace source '{master.name or '<Unknown>'}' specified a "
+				"layer name but lacks the required TTFont object in the 'font' "
+				"attribute."
 			)
 
 	return designspace.loadSourceFonts(_open_font, master_finder=master_finder)
--- a/Lib/fontTools/varLib/errors.py
+++ b/Lib/fontTools/varLib/errors.py
@@ -1,0 +1,39 @@
+class VarLibError(Exception):
+    """Base exception for the varLib module."""
+
+
+class VarLibValidationError(VarLibError):
+    """Raised when input data is invalid from varLib's point of view."""
+
+
+class VarLibMergeError(VarLibError):
+    """Raised when input data cannot be merged into a variable font."""
+
+
+class VarLibCFFDictMergeError(VarLibMergeError):
+    """Raised when a CFF PrivateDict cannot be merged."""
+
+    def __init__(self, key, value, values):
+        error_msg = (
+            f"For the Private Dict key '{key}', the default font value list:"
+            f"\n\t{value}\nhad a different number of values than a region font:"
+        )
+        for region_value in values:
+            error_msg += f"\n\t{region_value}"
+        self.args = (error_msg,)
+
+
+class VarLibCFFPointTypeMergeError(VarLibMergeError):
+    """Raised when a CFF glyph cannot be merged."""
+
+    def __init__(self, point_type, pt_index, m_index, default_type, glyph_name):
+        error_msg = (
+            f"Glyph '{glyph_name}': '{point_type}' at point index {pt_index} in "
+            f"master index {m_index} differs from the default font point type "
+            f"'{default_type}'"
+        )
+        self.args = (error_msg,)
+
+
+class VariationModelError(VarLibError):
+    """Raised when a variation model is faulty."""
--- a/Lib/fontTools/varLib/merger.py
+++ b/Lib/fontTools/varLib/merger.py
@@ -14,6 +14,8 @@ from fontTools.varLib.varStore import VarStoreInstancer
 from functools import reduce
 from fontTools.otlLib.builder import buildSinglePos
 
+from .errors import VarLibMergeError
+
 
 class Merger(object):
 
@@ -66,8 +68,8 @@ class Merger(object):
 			if hasattr(item, "ensureDecompiled"):
 				item.ensureDecompiled()
 		keys = sorted(vars(out).keys())
-		assert all(keys == sorted(vars(v).keys()) for v in lst), \
-			(keys, [sorted(vars(v).keys()) for v in lst])
+		if not all(keys == sorted(vars(v).keys()) for v in lst):
+			raise VarLibMergeError((keys, [sorted(vars(v).keys()) for v in lst]))
 		mergers = self.mergersFor(out)
 		defaultMerger = mergers.get('*', self.__class__.mergeThings)
 		try:
@@ -82,7 +84,8 @@ class Merger(object):
 			raise
 
 	def mergeLists(self, out, lst):
-		assert allEqualTo(out, lst, len), (len(out), [len(v) for v in lst])
+		if not allEqualTo(out, lst, len):
+			raise VarLibMergeError((len(out), [len(v) for v in lst]))
 		for i,(value,values) in enumerate(zip(out, zip(*lst))):
 			try:
 				self.mergeThings(value, values)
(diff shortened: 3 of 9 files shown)


madig:
Ugh, I just noticed that all the blank `assert`s produce empty exception messages. Guess I'll write something for all of them :)

madig:
Change of strategy after a Cosimo talk: Try to refactor varLib's exceptions instead of trying to kludge them at the API boundary.

```
first thing: identify all the possible source of exceptiosn
then: classify them into groups
then: define exception classes that make sense and export those as varLib.errors
replace generic exception with more specific ones
```

madig:
I had a look at every line in varLib that contains `(assert|raise)`, and actual errors seem to fall very broadly into two groups: (Designspace) validation errors and font merger errors. There are miscellaneous errors that I'd broadly call "internal sanity checking", i.e. asserting that a function parameter is a positive number or that a function has certain pre- and/or postconditions (traditional `assert`s). There seems to be a fine line between actual and miscellaneous errors in merger.py.

madig:
I would use `ValueError` when e.g. a function checks that it has received the correct values and `assert` for pre- or postconditions. Both would result in tracebacks on the terminal, but they are really internal programming errors rather than faulty input the user could do something about.

madig:
The `assert`s in iup.py seem to be mostly internal consistency stuff?

madig:
The changes in merger.py need the most scrutiny. Gonna have to do some testing with known bad input.

madig:
Hrm. Testing reveals that the `except Exception as e` statements also caught `ValueError`s when e.g. `_MarkBasePosFormat1_merge` calls into `getMasterLocationsSortKeyFunc` eventually and a base master is not found.

justvanrossum:
Those `except Exception as e` statements only serve to extent the error message, and they all reraise. Why even change those at all?

madig:
I want to coerce everything into a VarLibMergeError. I suppose I could do an `isinstance` check and convert if necessary.

anthrotype:
yes, like Just said, the purpose of those bare Exception is to extend the arguments (to provide context) and they always re-raise the underlying exception. We can keep those

madig:
Ok, and what to do with e.g. https://github.com/fonttools/fonttools/pull/1821/files#diff-f63c79454bc79c02f239be712e5833c7R224-R225? ValueError looks appropriate to me here because it's a generic varLib function, but the error happens during merging, which I want to make catchable. Wrap unless isinstance of VarLibError?

justvanrossum:
Maybe raising the following would be an option?

```python
class VarLibValueError(ValueError, VarLibError): pass
```

justvanrossum:
Hm, never mind, that makes not sense.

justvanrossum:
Btw. isn't this a good use case for exception chaining? At some point you want to present the user with (say) a VarLibMergeError, but you don't really want to hide the fact that the source was a ValueError.

https://www.python.org/dev/peps/pep-3134/

madig:
It should be, indeed. The code now converts any exception into a VLMError with chaining unless it already is one.

madig:
Now, what to do about the remaining asserts in merger.py?

1. The coercion code will use `str(e)` for chaining, but many asserts don't have a message attached, so the error message will be an empty final string. Suboptimal.
2. Asserts are omitted when running python with `-O`, so they should only test for internal consistency and should never be part of the actual algorithm. 
3. Asserts point to a programming error and should ideally crash everything and produce a backtrace, while a proper VarLibMergeError should produce a friendlier error message. The user can only do something about the latter. Coercion muddles this, while also making sure that we catch all left-over asserts that are actually part of the algorithm.

Should I add message to all asserts? There is more than one place though where the intent is unclear to me.

justvanrossum:
The `str(e)` is not good: that loses the exeption type. What I think you should do is this:

```python
try:
    ...some varlib code...
except Exception as e:
    if not isinstance(e, VarLibMergeError):
        raise VarLibMergeError("message...") from e  # exception chaining
    raise
```

justvanrossum:
(sorry for the noise, I was not reading properly as you could tell)

madig:
Cosimo says not to coerce errors but to catch e.g. ValueError at the source and convert as necessary.

Lib/fontTools/varLib/__init__.py
```diff
@@ -700,56 +676,141 @@ def load_designspace(designspace):
...
+			if len(input_locations) != len(input_list):
+				raise VarLibValidationError(f"Designspace {ds_path}: Axis '{axis.name}': mapping input values must be unique but contain duplicates.")
+			if len(output_locations) != len(output_list):
+				raise VarLibValidationError(f"Designspace {ds_path}: Axis '{axis.name}': mapping output values must be unique but contain duplicates.")
+
+			# Input and output values are in ascending order:
+			if sorted(input_locations) != input_list:
+				raise VarLibValidationError(f"Designspace {ds_path}: Axis '{axis.name}': mapping input values must be in ascending order.")
+			if sorted(output_locations) != output_list:
+				raise VarLibValidationError(f"Designspace {ds_path}: Axis '{axis.name}': mapping output values must be in ascending order.")
```
belluzj: Compared to the previous code, the following mapping which was allowed before is now rejected. Why the change?
```xml
<map input="10" output="1"/>
<map input="0" output="0"/>
<map input="20" output="2"/>
```

Lib/fontTools/varLib/__init__.py
```diff
@@ -700,56 +676,141 @@ def load_designspace(designspace):
...
 	log.info("Axes:\n%s", pformat([axis.asdict() for axis in axes.values()]))
 
 	# Check all master and instance locations are valid and fill in defaults
+	instances = ds.instances
 	for obj in masters+instances:
 		obj_name = obj.name or obj.styleName or ''
 		loc = obj.location
-		for axis_name in loc.keys():
-			assert axis_name in axes, "Location axis '%s' unknown for '%s'." % (axis_name, obj_name)
+		# NOTE: designspaceLib filters out axes in locations that are unknown, no need to check here.
```
belluzj: But is designspaceLib the only library that reads designspaces? I would say that the check can go away only if the designspace spec says that it's forbidden to mention locations along axes that are not defined previously?

Lib/fontTools/varLib/__init__.py
```diff
@@ -700,56 +676,141 @@ def load_designspace(designspace):
...
 		for axis_name,axis in axes.items():
 			if axis_name not in loc:
-				loc[axis_name] = axis.default
+				# NOTE: `axis.default` is always user-space, but `obj.location` always design-space.
+				loc[axis_name] = axis.map_forward(axis.default)
```
belluzj: This is a bugfix then, not just refactoring? Mention it in the release notes or changelog?

Lib/fontTools/varLib/__init__.py
```diff
@@ -700,56 +676,141 @@ def load_designspace(designspace):
...
 
-	# Normalize master locations
+				v = axis.map_backward(axis_location)
+				if not (axis.minimum <= v <= axis.maximum):
+					raise VarLibValidationError(f"Designspace {ds_path}: Source or instance '{obj_name}' has out-of-range location for axis '{axis_name}': is {v} but must be in range [{axis.minimum}..{axis.maximum}] (NOTE: all values are in user-space).")
+
+	# Check that we have something to put into fvar.
+	for instance in instances:
+		if "en" not in instance.localisedStyleName and not instance.styleName:
+			raise VarLibValidationError(f"Designspace {ds_path}: Instance at location '{coordinates}' must have a styleName.")
```
belluzj: `coordinates`?

Lib/fontTools/varLib/__init__.py
```diff
@@ -700,56 +676,141 @@ def load_designspace(designspace):
...
+
+	# Sanity-check rules.
+	for rule in ds.rules:
+		for condition_set in rule.conditionSets:
+			for condition in condition_set:
+				# NOTE: All rule values are in design space.
+				if not condition.get("name"):
+					raise VarLibValidationError(f"Designspace {ds_path}: Rule '{rule.name}' contains a condition without an axis name.")
```
belluzj: Better worded as "should have..." (actionable)

Lib/fontTools/varLib/__init__.py
```diff
@@ -763,19 +824,22 @@ def load_designspace(designspace):
 	normalized_master_locs = [models.normalizeLocation(m, internal_axis_supports) for m in internal_master_locs]
 	log.info("Normalized master locations:\n%s", pformat(normalized_master_locs))
 
-	# Find base master
-	base_idx = None
+	# Find default master after normalization.
+	# TODO: Can the base master ever be anything other than what ds.findDefault() returns?
```
belluzj: I remember a discussion about this and also vs. copyKerning = true and friends I think

Lib/fontTools/varLib/__init__.py
```diff
@@ -763,19 +824,22 @@ def load_designspace(designspace):
...
 		if all(v == 0 for v in m.values()):
-			assert base_idx is None
-			base_idx = i
-	assert base_idx is not None, "Base master not found; no master at default location?"
-	log.info("Index of base master: %s", base_idx)
+			if default_master_index is not None:
+				raise VarLibValidationError(f"Designspace {ds_path}: More than one default master found.")
+			default_master_index = i
+	if default_master_index is None:
+		raise VarLibValidationError(f"Designspace {ds_path}: Default master not found after normalizing the locations; no master at default location? Check the mappings.")
```
belluzj: "There should be a master/source at the default location, meaning that the location of that master along each axis is equal to the default value for that axis" or something like that

Lib/fontTools/varLib/__init__.py
```diff
@@ -1019,12 +1092,12 @@ def main(args=None):
 	logging_group = parser.add_mutually_exclusive_group(required=False)
 	logging_group.add_argument(
 		"-v", "--verbose",
-                action="store_true",
-                help="Run more verbosely.")
+				action="store_true",
+				help="Run more verbosely.")
```
belluzj: Why?

Lib/fontTools/varLib/merger.py
```diff
@@ -170,7 +177,8 @@ def _merge_GlyphOrders(font, lst, values_lst=None, default=None):
 	sortKey = font.getReverseGlyphMap().__getitem__
 	order = sorted(combined, key=sortKey)
 	# Make sure all input glyphsets were in proper order
-	assert all(sorted(vs, key=sortKey) == vs for vs in lst), "glyph orders are not consistent across masters"
+	if not all(sorted(vs, key=sortKey) == vs for vs in lst):
+		raise VarLibMergeError("glyph orders are not consistent across masters")
```
belluzj: Could this say e.g. which master is different from the other or something actionable by the user?

Lib/fontTools/varLib/__init__.py
madig: Mixture of tabs and spaces. This tabs everything.

Lib/fontTools/varLib/__init__.py
```diff
@@ -60,17 +59,14 @@ def _add_fvar(font, axes, instances):
...
-
 	log.info("Generating fvar")
 
 	fvar = newTable('fvar')
 	nameTable = font['name']
 
 	for a in axes.values():
 		axis = Axis()
-		axis.axisTag = Tag(a.tag)
+		axis.axisTag = a.tag
```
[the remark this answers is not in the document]
madig: I think it's unnecessary. @anthrotype ?

Lib/fontTools/varLib/__init__.py
```diff
@@ -92,7 +87,7 @@ def _add_fvar(font, axes, instances):
 		inst = NamedInstance()
 		inst.subfamilyNameID = nameTable.addMultilingualName(localisedStyleName)
 		if psname is not None:
-			psname = tounicode(psname)
+			psname = psname
```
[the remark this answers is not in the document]
madig: I think it's unnecessary. @anthrotype ?

Lib/fontTools/varLib/__init__.py
madig: Bugfix :) Must write a test.

Lib/fontTools/varLib/__init__.py
madig: The specification doesn't seem to impose an order here. @anthrotype sort the map by input values like the previous code so Jany's example passes?

Lib/fontTools/varLib/__init__.py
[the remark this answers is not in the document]
anthrotype: it's unnecessary in this case because sstruct.pack encodes them `tobytes` before. The Tag class is to be able to treat bytes or ascii strings as equivalent, like in py2. I'd suggest you keep it there for now, we can clean up all that py23 legacy in separate PR.

Lib/fontTools/varLib/__init__.py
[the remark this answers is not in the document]
anthrotype: ok this one is no longer needed, as any strings parsed with py3 etree or lxml will be unicode strings now (previously ascii strings may be encoded as bytes)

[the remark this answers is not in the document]
anthrotype: but do strip the unneeded tautology pls

Lib/fontTools/varLib/__init__.py
anthrotype: yes, sort

Lib/fontTools/varLib/__init__.py
anthrotype: agree with jany, pls keep this assert as assert

Lib/fontTools/varLib/merger.py
```diff
@@ -1015,15 +1030,17 @@ def buildVarDevTable(store_builder, master_values):
...
 	self.Coordinate, DeviceTable = buildVarDevTable(merger.store_builder, [a.Coordinate for a in lst])
 	if DeviceTable:
 		self.Format = 3
 		self.DeviceTable = DeviceTable
 
 @VariationMerger.merger(ot.Anchor)
 def merge(merger, self, lst):
-	assert self.Format == 1
+	if self.Format != 1:
+		raise VarLibMergeError(f"Anchor format {self.Format} unknown.")
```
anthrotype: it's not like it is unknown, more like the merger doesn't support the second type of anchors (using contour points to position)

Lib/fontTools/varLib/merger.py
```diff
@@ -1015,15 +1030,17 @@ def buildVarDevTable(store_builder, master_values):
 
 @VariationMerger.merger(ot.CaretValue)
 def merge(merger, self, lst):
-	assert self.Format == 1
+	if self.Format != 1:
+		raise VarLibMergeError(f"CaretValue format {self.Format} unknown.")
```
anthrotype: like for anchors, same for the caret (the format 2 uses contour points instead of xy coords). it is unsupported, not unknown

Lib/fontTools/varLib/merger.py
```diff
@@ -766,8 +778,10 @@ def merge(merger, self, lst):
 		if not sts:
 			continue
 		if sts[0].__class__.__name__.startswith('Extension'):
-			assert allEqual([st.__class__ for st in sts])
-			assert allEqual([st.ExtensionLookupType for st in sts])
+			if not allEqual([st.__class__ for st in sts]):
+				raise VarLibMergeError()
```
anthrotype: the problem here is when, while merging a lookup that contains Extension subtables, not all the masters use extensions or have the same extension lookup type (because later it is assuming the first subtable lookup type is the one to be used for all). I know it's hard to come up with a meaningful error message here. But raising a muted exception with no error is no better than before.

Lib/fontTools/varLib/__init__.py
```diff
@@ -127,34 +124,16 @@ def _add_avar(font, axes):
...
-		assert len(set(keys)) == len(keys), (
-			f"{axis.tag} axis: All axis mapping input='...' "
-			"values must be unique, but we found duplicates."
-		)
-		assert len(set(vals)) == len(vals), (
-			f"{axis.tag} axis: All axis mapping output='...' "
-			"values must be unique, but we found duplicates."
-		)
-		# Ascending values
-		assert sorted(vals) == vals
```
anthrotype: i know you are validating and raising exceptions beforehand, but keeping the assertions wouldn't do harm here. It's simply confirming what should always stay true

Lib/fontTools/varLib/__init__.py
```diff
@@ -164,10 +143,12 @@ def _add_avar(font, axes):
 
 		curve.update(zip(keys, vals))
 
-		assert 0.0 in curve and curve[0.0] == 0.0
-		assert -1.0 not in curve or curve[-1.0] == -1.0
-		assert +1.0 not in curve or curve[+1.0] == +1.0
```
anthrotype: these should stay as assertions. given the assumptions and the way the normalizing code works, these will always be true. You can't hit them, hence assertion is the correct error

Lib/fontTools/varLib/__init__.py
madig: Cosimo says this code may predate findDefault and what they do should match. Should add a test for that.

Lib/fontTools/varLib/errors.py
```diff
@@ -0,0 +1,39 @@
...
+        error_msg = (
+            f"For the Private Dict key '{key}', the default font value list:"
+            f"\n\t{value}\nhad a different number of values than a region font:"
+        )
+        for region_value in values:
+            error_msg += f"\n\t{region_value}"
+        self.args = (error_msg,)
+
+
+class VarLibCFFPointTypeMergeError(VarLibMergeError):
```
anthrotype: would be nice to have a similarly detailed exception with glyph name, master index and point index, etc. for TrueType glyphs too (but for later maybe)

Lib/fontTools/varLib/__init__.py
```diff
@@ -870,29 +922,35 @@ def build(designspace, master_finder=lambda s:s, exclude=[], optimize=True):
...
+			_add_CFF2(vf, model, master_fonts)
+			if "post" in vf:
+				# set 'post' to format 2 to keep the glyph names dropped from CFF2
+				post = vf["post"]
+				if post.formatType != 2.0:
+					post.formatType = 2.0
+					post.extraNames = []
+					post.mapping = {}
+	except Exception as e:
+		e.args = (f"In Designspace '{ds_path}': {', '.join(str(a) for a in e.args)}",)
```
anthrotype: i'm not sure this should be done in here. I prefer if the client of varLib.build takes care of including (or not) the designspace.path in the error message

Lib/fontTools/varLib/__init__.py
```diff
@@ -852,11 +900,15 @@ def build(designspace, master_finder=lambda s:s, exclude=[], optimize=True):
...
-	if 'avar' not in exclude:
-		_add_avar(vf, ds.axes)
+	try:
+		fvar = _add_fvar(vf, ds.axes, ds.instances)
+		if 'STAT' not in exclude:
+			_add_stat(vf, ds.axes)
+		if 'avar' not in exclude:
+				_add_avar(vf, ds.axes)
+	except Exception as e:
+		e.args = (f"In Designspace '{ds_path}': {', '.join(str(a) for a in e.args)}",)
```
anthrotype: same here, I prefer if client does this if they care.

Lib/fontTools/varLib/__init__.py
```diff
@@ -713,20 +733,27 @@ def load_designspace(designspace):
...
-	for axis in ds.axes:
+	for axis_index, axis in enumerate(ds.axes):
 		axis_name = axis.name
 		if not axis_name:
-			assert axis.tag is not None
+			if not axis.tag:
+				raise VarLibValidationError(f"Axis at index {axis_index} needs a tag.")
 			axis_name = axis.name = axis.tag
+			# TODO: Set axis_name to the corresponding default name from
+			# standard_axis_map so the code below gives it a default name.
```
anthrotype: i don't understand this TODO

madig: If `axis.name` is unset and we're setting it to the axis tag string, we miss fetching the label name in the `if axis_name in standard_axis_map` conditional.

The follwoing code does it:
```python
	# Provide fallback standard name table entries for known axes.
	standard_names = {
		"weight": ("wght", {"en": "Weight"}),
		"width": ("wdth", {"en": "Width"}),
		"slant": ("slnt", {"en": "Slant"}),
		"optical": ("opsz", {"en": "Optical Size"}),
		"italic": ("ital", {"en": "Italic"}),
	}
	tag_to_standard_names = {v[0]: k for k, v in standard_names.items()}

	# Set up axes.
	# [...]

	axes = OrderedDict()
	for axis_index, axis in enumerate(ds.axes):
		if not axis.name:
			if not axis.tag:
				raise VarLibValidationError(f"Designspace {ds_path}: Axis at index {axis_index} needs a tag.")
			axis.name = tag_to_standard_names[axis.tag]  # Fill in default name below.

		if axis.name in standard_names:
			if axis.tag is None:
				axis.tag = standard_names[axis.name][0]
			if not axis.labelNames:
				axis.labelNames.update(standard_names[axis.name][1])
		# [...]
	# [...]
```

Maybe this can be refactored even more.

Lib/fontTools/varLib/errors.py
madig: I can add a TODO, although most of these cases are going to be caught by cu2qu, no?

Lib/fontTools/varLib/__init__.py
```diff
@@ -81,7 +79,10 @@ def _add_fvar(font, axes, instances):
 		coordinates = instance.location
 
 		if "en" not in instance.localisedStyleName:
-			assert instance.styleName
+			if not instance.styleName:
+				raise VarLibValidationError(
+					f"Instance at location '{coordinates}' must have a styleName."
```
anthrotype: instance must have a styleName ... or an 'en' localizedStyleName (basically the default english style name is required)

Lib/fontTools/varLib/__init__.py
```diff
@@ -137,20 +138,37 @@ def _add_avar(font, axes):
...
-			"values must be unique, but we found duplicates."
-		)
+		if len(set(keys)) != len(keys):
+			raise VarLibValidationError(
+				f"Axis '{axis.name}': All axis mapping input='...' values must be "
+				"unique, but we found duplicates."
+			)
+		if len(set(vals)) != len(vals):
+			raise VarLibValidationError(
+				f"Axis '{axis.name}': All axis mapping output='...' values must be "
```
anthrotype: axis mapping values need not be unique actually, only increasing. Note the "greater than _or equal_" below when discussing the `toCoordinate` value in avar AxisValueMapRecord

https://docs.microsoft.com/en-us/typography/opentype/spec/avar

> for any given record except the first, the toCoordinate value must be greater than or equal to the toCoordinate value of the preceding record

Lib/fontTools/varLib/__init__.py
anthrotype: the whole `standard_axis_map` is kind of legacy, predates the time DS axes were given an explicit `tag` and `labelNames` attributes. Keep as is, and remove the TODO comment.

Lib/fontTools/varLib/__init__.py
```diff
@@ -739,15 +764,28 @@ def load_designspace(designspace):
 	for obj in masters+instances:
 		obj_name = obj.name or obj.styleName or ''
 		loc = obj.location
+		if loc is None:
+			raise VarLibValidationError(
+				f"Source or instance '{obj_name}' has no location."
```
anthrotype: when can happen that an object.location is None? If anything it should be an empty dict..
Do we not fill the implicit locations with axes defaults?

Lib/fontTools/varLib/__init__.py
```diff
@@ -739,15 +764,28 @@ def load_designspace(designspace):
...
 			if axis_name not in loc:
 				# NOTE: `axis.default` is always user-space, but `obj.location` always design-space.
 				loc[axis_name] = axis.map_forward(axis.default)
 			else:
 				v = axis.map_backward(loc[axis_name])
-				assert axis.minimum <= v <= axis.maximum, "Location for axis '%s' (mapped to %s) out of range for '%s' [%s..%s]" % (axis_name, v, obj_name, axis.minimum, axis.maximum)
+				if not (axis.minimum <= v <= axis.maximum):
+					raise VarLibValidationError(
+						f"Source or instance '{obj_name}' has out-of-range location "
+						f"for axis '{axis_name}': is {v} but must be in range "
```
anthrotype: don't say `is {v}`, but `is mapped to {v}`

Lib/fontTools/varLib/__init__.py
madig: Happens when you have no `<location>` element in a DS.

madig: Note to self: maybe use `designspaceLib.tagForAxisName` for what `standard_axis_map` does.

madig:
@josh-hadley I see you wrote code in the AFDKO's `buildcff2vf.py` module that catches one of the errors that I changed here. I inserted aliases for the old names, but you may want to update your code. Also, the same file redefines MergeTypeError?

josh-hadley:
> @josh-hadley I see you wrote code in the AFDKO's `buildcff2vf.py` module that catches one of the errors that I changed here. I inserted aliases for the old names, but you may want to update your code. Also, the same file redefines MergeTypeError?

Duly noted; thanks for the warning :-) I opened https://github.com/adobe-type-tools/afdko/issues/1088 in AFDKO and will make sure we're all in sync.

Issue #1812, opened 2020-01-27
[otData] how to handle Count value appearing after the array?


anthrotype:
I'm trying to convert `COLR` table to `otData`-style class. Here's is what I've got so far:

```python
	('COLR', [
		('uint16', 'Version', None, None, 'Table version number (starts at 0).'),
		('uint16', 'BaseGlyphRecordCount', None, None, 'Number of Base Glyph Records.'),
		('LOffsetTo(BaseGlyphRecordArray)', 'BaseGlyphRecordArray', None, None, 'Offset (from beginning of COLR table) to Base Glyph records.'),
		('LOffsetTo(LayerRecordArray)', 'LayerRecordArray', None, None, 'Offset (from beginning of COLR table) to Layer Records.'),
		('uint16', 'LayerRecordCount', None, None, 'Number of Layer Records.')
	]),

	('BaseGlyphRecordArray', [
		('BaseGlyphRecord', 'BaseGlyph', 'BaseGlyphRecordCount', 0, 'Base Glyph records.'),
	]),

	('BaseGlyphRecord', [
		('GlyphID', 'BaseGlyphID', None, None, 'Glyph ID of reference glyph. This glyph is for reference only and is not rendered for color.'),
		('uint16', 'FirstLayerIndex', None, None, 'Index (from beginning of the Layer Records) to the layer record. There will be numLayers consecutive entries for this base glyph.'),
		('uint16', 'NumLayers', None, None, 'Number of color layers associated with this glyph.'),
	]),

	('LayerRecordArray', [
		('LayerRecord', 'Layer', 'LayerRecordCount', 0, 'Layer records.'),
	]),

	('LayerRecord', [
		('GlyphID', 'LayerGlyphID', None, None, 'Glyph ID of layer glyph (must be in z-order from bottom to top).'),
		('uint16', 'PaletteIndex', None, None, 'Index value to use with a selected color palette.'),
	]),

```

The problem I have is that in COLR table the count value which tells how many records are in the LayerRecordArray only comes _after_ the offset to the LayerRecordArray.
See `numLayerRecords` in https://docs.microsoft.com/en-us/typography/opentype/spec/colr#header

But the assumption in otBase is that count fields always **precede** the actual arrays they are counting, whereas in the case of COLR, the layer count comes after the layers array (i.e. after its offset, but Offset converters are decompiled to the Table they point to).
I believe -- though I'm not 100% sure -- this is the reason why I get the following error when I try to decompile the table:

```
[...]

PR #1822, opened 2020-02-04
Define COLR using otData; add builders for COLRv1

anthrotype:
related to #1812 

this is incomplete. I drop it here in case anybody wants to take a look.

Both BaseTable.decompile and compile methods expect counts to appear before the things they count; if the things counted are in nested subtables, these counts are "propagated": i.e. when decompiling, the reader stores the count value in its local state, so that the nested subtable's converter can get it; when compiling and a propagated count field is found, the writer stores a CountReference object that will be filled later with the actual length by the time the nested array gets written.
The OTTableReader and OTTableWriter have dict-like methods and maintain an inner `localState` dict holding these counts or CountReferences. When an offset to a subtable is found upon decompiling, a sub-reader is forked and positioned at that offset, ready to decompile the subtable (see `otConverters.Table.read` method). If the font is loaded non-lazily (TTFont.lazy is None by default), the subtable decompile method is called immediately with the sub-reader. If font is lazy, the sub-reader is stored in the subtable, so that it is decompiled on request (see `BaseTable.__getattr__`).
The reader and writer `__setitem__` methods copy the inner `localState` (mutable dict) before setting the new value; a sub-reader (or sub-writer) shares the parent's localState, until its own state is modified. Exactly why this is like that is not 100% clear to me. However this makes so that, unless a propagated count appears before the offset to the subtable containing the counted elements, the localState of the reader while decompiling a nested subtable will not contain an entry for the count (nor a writer will contain the CountReference, upon compiling).

My workaround for COLR.LayerRecordCount (appearing after the offset to the LayerRecordArray in the COLRv0 header) is to have a custom `decompile` method that seeks to LayerRecordCount before the rest of the table is decompiled, and store that count in the reader localState.
Similarly, for the compile method, I have a custom preWrite that that sets a CountReference for LayerRecordCount, and make sure that it is propagated to the current writer before any other field is written.

@rsheeter had suggested that we change the logic such that the offsets to subtables aren't traversed until all the other fields in the current table are read. I didn't manage to found a clean way to do that doesn't also break everything else...


anthrotype:
instead of using the `BaseTTXConverter` for `table_C_O_L_R`, like the rest of the otData classes, I opted here for calling `OTTableReader` and `OTTableWriter` in its `decompile` and `compile` methods. Basically I replaced the manual use of `struct` with the otData-driven reader/writer. The fromXML and toXML are also unchanged and don't use otData-style xml serializer, which would give a much lower-level and verbose output than the current one (and would be backward incompatible).

anthrotype:
In commit https://github.com/fonttools/fonttools/pull/1822/commits/d2026bd65165adfc4bcf1b33f68517c3d2f3eb28 I started translating @behdad's [CORLv1 draft proposal](https://docs.google.com/document/d/1EPndWsdMK_M135FOIxQZ--dHpH727rXEBvLZqgIzZvs/edit#) to equivalent otData structures. But the `C_O_L_R_` ttLib class is not hooked to the v1 otTables.COLR yet.

I'm not sure how to do that. On the one hand, we want to keep the same interface as the old `C_O_L_R_` class (with its `ColorLayers` dict of LayerRecords keyed by base glyph name); at the same time I'd like to use otData-style definitions, and all ttLib tables that us those subclass the BaseTTXConverter (the one that have `.table` attribute wrapping the otTables dynamic classes).
So far I haven't sublcassed BaseTTXConverter from ttLib's C_O_L_R_ class, but call OTTableReader and OTTableWriter directly inside decompile/compile methods.

Maybe I should subclass BaseTTXConverter then add a `ColorLayers` property getter/setter that returns a dictionary in the form of the current ttLib class data. But then if one expects to modify those objects in-place (e.g. change the layer glyph name of a LayerRecord, add or remove a layer, etc.), it won't work anymore...

PR #1823, opened 2020-02-08
Make calling the subsetter possible with file-like objects

massaad:
Currently to call the subsetter as a library, the easiest way I found is to call main passing it a list of arguments. 

If it will be used as a step in a pipeline, it would be nice if it does not have to deal with actual files on disk. I found that it is possible to pass a file-like object for the input font as the first item in the argument list, but you cannot do the same with the output font, since all the rest of the arguments are treated as strings. 

This pull request makes it possible to have another file-like object as the output font. This is an extra parameter (named outfile) to main. If outfile is not passed in, everything will work as before, but if it is passed in, the output will be written to that file-like object (for example an `io.BytesIO()` from the standard library).

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -2771,7 +2771,7 @@ def usage():
 	print("Try pyftsubset --help for more information.\n", file=sys.stderr)
 
 @timer("make one with everything (TOTAL TIME)")
-def main(args=None):
+def main(args=None, outfile=None):
 	from os.path import splitext
 	from fontTools import configLogger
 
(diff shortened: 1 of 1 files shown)


anthrotype:
`main()` functions are supposed to be used as command line scripts' entry points, so they only accept a list of strings (the content of sys.argv).

If you want to save the output of the subsetter to a file-like object, you should be importing the `Subsetter` class, initialize it with options, call `populate` with some inputs, then call `subset` with a TTFont instance that you wish to subset (which will be modified in-place), finally save the TTFont to a file-like object.
Basically what the `subset.main` function does, without the the CLI argument parsing.

massaad:
Thanks. That works.

Issue #1725, opened 2019-09-17
add typing support and run mypy static typechecker


anthrotype:
now that fonttools is py3-only we can take advantage of native typing annotation and run mypy to statically check types.

Currently running mypy on the whole fontTools package -- even without any annotations -- gives a lot of errors (376 errors in 49 files, out of 224 files checked). The list is too long to paste here.

I imagine it will not be easy to adapt fonttools to pass strict mypy checks, since fonttools makes extensive use of duck-typing and other dynamic features of the language.
But at least we should give it a try.

madig:
Maybe we can start with one sub-library and go from there gradually.

nbro10:
Are there any plans to actually solve this?

I'm getting this `Skipping analyzing "fontTools.ttLib": module is installed, but missing library stubs or py.typed marker  [import-untyped]`

anthrotype:
Unfortunately no, I'm sorry. I have limited time and I don't think this is urgent. For now, I suggest you just mute/ignore those warnings.

PR #1826, opened 2020-02-10
colorLib: add buildCOLR and buildCPAL

anthrotype:
for now this just adds a `buildCOLR` and `buildCPAL`, e.g. so we can use it in ufo2ft here https://github.com/googlefonts/ufo2ft/pull/359

### DIFF
--- a/Lib/fontTools/colorLib/builder.py
+++ b/Lib/fontTools/colorLib/builder.py
@@ -1,46 +1,45 @@
-from fontTools.ttLib import newTable
+from typing import Dict, List, Tuple
+from fontTools.ttLib.tables.C_O_L_R_ import LayerRecord, table_C_O_L_R_
+from fontTools.ttLib.tables.C_P_A_L_ import Color, table_C_P_A_L_
 from .errors import ColorLibError
 
 
-def buildCOLR(colorLayers):
+def buildCOLR(colorLayers: Dict[str, List[Tuple[str, int]]]) -> table_C_O_L_R_:
     """Build COLR table from color layers mapping.
 
     Args:
-        colorLayers: Dict[str, List[Tuple[str, int]]]: map of base glyph names (str)
-            to lists of layer glyph names (str) and palette indices (int) tuples.
+        colorLayers: : map of base glyph names to lists of (layer glyph names,
+            palette indices) tuples.
 
     Return:
         A new COLRv0 table.
     """
-    from fontTools.ttLib.tables.C_O_L_R_ import LayerRecord
-
     colorLayerLists = {}
     for baseGlyphName, layers in colorLayers.items():
         colorLayerLists[baseGlyphName] = [
             LayerRecord(layerGlyphName, colorID) for layerGlyphName, colorID in layers
         ]
 
-    colr = newTable("COLR")
+    colr = table_C_O_L_R_()
     colr.version = 0
     colr.ColorLayers = colorLayerLists
     return colr
 
 
-def buildCPAL(palettes):
+def buildCPAL(
+    palettes: List[List[Tuple[float, float, float, float]]]
+) -> table_C_P_A_L_:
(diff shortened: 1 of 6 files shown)


Lib/fontTools/colorLib/builder.py
```diff
@@ -0,0 +1,54 @@
...
+    colr.version = 0
+    colr.ColorLayers = colorLayerLists
+    return colr
+
+
+def buildCPAL(palettes):
+    """Build CPAL table from list of color palettes.
+
+    Args:
+        palettes: List[List[Tuple[float, float, float, float]]]: list of lists
```
madig: Maybe use typing hints directly?

anthrotype: yes, that'd be nice for self-documenting purposes, but we don't run any static typechecker. Also, otTables classes are dynamically generated and mypy doesn't seem to like that very much (#1725). Let's do fix separately some other time

madig: Given that docstrings aren't checked either, I'd prefer we actually use type hints so that adding mypy is less work later. My 2 cents :)

anthrotype: fair enough, I added the type hints.

Lib/fontTools/colorLib/builder.py
```diff
@@ -0,0 +1,147 @@
...
+                raise ColorLibError(
+                    f"In palette[{i}][{j}]: expected (R, G, B, A) tuple, got {color!r}"
+                )
+            if any(v > 1 or v < 0 for v in color):
+                raise ColorLibError(
+                    f"palette[{i}][{j}] has invalid out-of-range [0..1] color: {color!r}"
+                )
+            # input colors are RGBA, CPAL encodes them as BGRA
+            red, green, blue, alpha = color
+            colors.append(Color(*(round(v * 255) for v in (blue, green, red, alpha))))
```
khaledhosny: I wonder, was using floats then multiplying by `255` really necessary, as opposed to using the `0..255` range integers directly? For example, Glyphs stores the integers and it feels rather unfortunate to convert to floats when using this function, just for it to convert them back to integers.

anthrotype: hm. In UFO, colors are usually stored as RGBA tuples of floats in 0..1 range. 
http://unifiedfontobject.org/versions/ufo3/conventions/#colors

In ufo2ft similarly we used that notation:
https://github.com/googlefonts/ufo2ft/pull/359/files

Anyway, an int in 0..255 range that is divided by 255, then again multiplied by 255 and rounded, is guaranteed to round-trip..

khaledhosny: If it is guaranteed to round trip, OK then.

anthrotype:
```python
>>> assert list(range(256)) == [round(v/255 * 255) for v in range(256)]
True
```

justvanrossum:
I see 0..255 color values as similar to 0.8 fixed point numbers. It's a low level representation that is tied to the binary format. We generally don't use fixed point numbers directly in text formats, either, but convert to float.

khaledhosny:
I was worried about round tripping, my other problem with floats (if I were to save it to file, but I don’t have to here) is that I get from Python stuff like `[0.8235294117647058, 0.1803921568627451, 0.12156862745098039, 1.0]` (for `[210, 46, 31, 255]`) and my simple brain wouldn’t know what safe rounding can I apply to it to get readable numbers.

anthrotype:
3 decimal positions are enough to roundtrip

this still passes
```
assert list(range(256)) == [round(float(f"{v/255:.03f}") * 255) for v in range(256)]
```

with 2 decimals, it doesn't roundtrip

madig:
We use 3 decimal positions in glyphsLib and it works well.

Issue #1723, opened 2019-09-14
[feaLib] Raise specific exception when a glyph referenced in a feature file does not exist in the font


madig:
So we can print a user-friendly error message for cases like https://github.com/googlefonts/fontmake/issues/584#issuecomment-531458280.

PR #1828, opened 2020-02-11
[feaLib] Check that glyph names referenced in the feature file are part of the glyph set

madig:
This checks that glyph names that appear in a feature file are actually in the glyph set provided in `glyphNames`, unless it is empty. 

Closes https://github.com/fonttools/fonttools/issues/1723.

### DIFF
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -19,6 +19,14 @@ class Parser(object):
 
     def __init__(self, featurefile, glyphNames=(), followIncludes=True,
                  **kwargs):
+        """Initializes a Parser object.
+
+        Note: the `glyphNames` iterable serves a double role to help distinguish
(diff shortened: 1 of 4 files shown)


anthrotype:
makes sense to use `glyphNames` when one is provided.

LGTM once all test pass

madig:
Hm. My change to `expect_glyph_` might be too simplistic. `parse_glyphclass_` contains some logic to distinguish glyph names from ranges using the return value of `expect_glyph_`. So I guess I have to move my checking logic up.

madig:
I have no idea if I _should_ check CID names for membership.

anthrotype:
@madig yes, you should treat cidXXXXX pseudo-names the same as normal glyph names. If parser is provided the `glyphNames` parameter, it should error if a CID glyph is referenced in the features but is missing from the font.

PR #1829, opened 2020-02-12
[glyf] compile empty table as 1 null byte to make OTS and Windows happy

### LINKED ISSUE #899: add 1 zero byte when compiling empty glyf tables...
...to make Windows happy.

The OTS does that too now.

See discussion
https://github.com/khaledhosny/ots/issues/52#issuecomment-289369267

anthrotype:
OTS rejects fonts with empty tables, and Windows considers the font file invalid if a table is empty.
See discussion https://github.com/khaledhosny/ots/issues/52
When a glyf table is empty (e.g. all glyphs are zero contours) we compile it as a single zero byte `b'\x00'`, which will make it pass OTS.

Fixes https://github.com/fonttools/fonttools/issues/899


behdad:
I'm against this.

behdad:
Ah doing it just for `glyf` table.  Yeah am fine with that.

justvanrossum (approved):

anthrotype:
Yes only for glyf. OTS does the same.

PR #1831, opened 2020-02-13
[glyf] fix getCoordinates for composite glyphs that use anchor points

### LINKED ISSUE #1556: [glyf] bbox of transformed & anchored component calculated incorrectly
The attached ttx file of a TrueType font (Variable, but this seems irrelevant) has its “O” glyph composed of a pair of “C” glyphs, one rotated by 180 degrees lined up to the other using anchor points (i.e., flag ARGS_ARE_XY_VALUES unset). The XML description of “O” glyph reads:
```
<TTGlyph name="O" xMin="56" yMin="-12" xMax="612" yMax="672">
  <component glyphName="C" x="0" y="0" flags="0x204"/>
  <component glyphName="C" firstPt="0" secondPt="9" scale="-1.0" flags="0x4"/>
</TTGlyph>
```
When converted to ttf by ttx then back to ttx, the XML description becomes:
```
<TTGlyph name="O" xMin="-334" yMin="-1356" xMax="334" yMax="672">
  <component glyphName="C" x="0" y="0" flags="0x204"/>
  <component glyphName="C" firstPt="0" secondPt="9" scale="-1.0" flags="0x4"/>
</TTGlyph>
```
The glyph bounding box is calculated incorrectly.
It appears as if translation offset for the component is calculated before rotation. These should be done backwards.

[SourceSansVariable-Roman.anchor.ttx.zip](https://github.com/fonttools/fonttools/files/3009833/SourceSansVariable-Roman.anchor.ttx.zip)

anthrotype:
Fixes #1556 

As @blueshade7 noted, there is a bug computing bounding boxes for composite glyphs with components that uses firstPt/secondPt reference anchor points (instead of the more common XY offsets), and the components also have a transform.

This is because we are computing the translation offset between firstPt and secondPt _before_
applying the 2x2 scale/rotation/shear transform. By the time we also apply the translation, the
offset is now incorrect.

We need to compute the translation offset _after_ we have applied the 2x2 transform.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -1012,33 +1012,37 @@ class Glyph(object):
 					coordinates, endPts, flags = g.getCoordinates(glyfTable)
 				except RecursionError:
 					raise ttLib.TTLibError("glyph '%s' contains a recursive component reference" % compo.glyphName)
+				coordinates = GlyphCoordinates(coordinates)
 				if hasattr(compo, "firstPt"):
-					# move according to two reference points
+					# component uses two reference points: we apply the transform _before_
+					# computing the offset between the points
+					if hasattr(compo, "transform"):
+						coordinates.transform(compo.transform)
 					x1,y1 = allCoords[compo.firstPt]
 					x2,y2 = coordinates[compo.secondPt]
 					move = x1-x2, y1-y2
-				else:
-					move = compo.x, compo.y
-
-				coordinates = GlyphCoordinates(coordinates)
-				if not hasattr(compo, "transform"):
 					coordinates.translate(move)
 				else:
-					apple_way = compo.flags & SCALED_COMPONENT_OFFSET
-					ms_way = compo.flags & UNSCALED_COMPONENT_OFFSET
-					assert not (apple_way and ms_way)
-					if not (apple_way or ms_way):
-						scale_component_offset = SCALE_COMPONENT_OFFSET_DEFAULT  # see top of this file
-					else:
-						scale_component_offset = apple_way
-					if scale_component_offset:
-						# the Apple way: first move, then scale (ie. scale the component offset)
+					# component uses XY offsets
+					move = compo.x, compo.y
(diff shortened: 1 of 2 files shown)


Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -1012,33 +1012,37 @@ def getCoordinates(self, glyfTable):
...
-					else:
-						scale_component_offset = apple_way
-					if scale_component_offset:
-						# the Apple way: first move, then scale (ie. scale the component offset)
+					# component uses XY offsets
+					move = compo.x, compo.y
+					if not hasattr(compo, "transform"):
 						coordinates.translate(move)
-						coordinates.transform(compo.transform)
 					else:
```
anthrotype: Note that the whole `else:` block below is simply indented, the diff seems bigger than it really is

anthrotype:
roundtripping through ttx the file that @blueshade7 attached to #1556 [SourceSansVariable-Roman.anchor.ttx.zip](https://github.com/fonttools/fonttools/files/3009833/SourceSansVariable-Roman.anchor.ttx.zip) with this PR patch applied does not modify the bounding boxes any more.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -1012,33 +1012,37 @@ def getCoordinates(self, glyfTable):
...
 					move = x1-x2, y1-y2
-				else:
-					move = compo.x, compo.y
-
-				coordinates = GlyphCoordinates(coordinates)
-				if not hasattr(compo, "transform"):
 					coordinates.translate(move)
 				else:
-					apple_way = compo.flags & SCALED_COMPONENT_OFFSET
-					ms_way = compo.flags & UNSCALED_COMPONENT_OFFSET
```
anthrotype: the whole scaled vs unscaled component _offset_ (Apple vs MS) logic should not apply for components that use firstPt/secondPt anchor points. There is no offset to scale in this case; the only offset possible is the difference between firstPt and secondPt, which must always be aligned to one another (whether the componet is scaled or not).

PR #1844, opened 2020-03-03
Fix composite instructions for empty program

jenskutilek:
I found a problem when a composite has a glyph program which is empty.

When I generate a PDF via InDesign, which contains a subset of such font, composite glyphs with empty programs still have the "WE_HAVE_INSTRUCTIONS" flag set, but there is no instruction data to decompile. This causes the affected glyphs to disappear from the PDF text, and make Adobe Reader show a warning about damaged fonts.

This patch makes sure the flag is only set if there actually is a non-empty glyph program.

It also adds a warning on decompilation of a broken font extracted from such PDF (instead of failing decompilation).

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -789,7 +789,8 @@ class Glyph(object):
 		haveInstructions = 0
 		for i in range(len(self.components)):
 			if i == lastcomponent:
-				haveInstructions = hasattr(self, "program")
+				# Only set haveInstructions if the program is non-zero
+				haveInstructions = hasattr(self, "program") and self.program
 				more = 0
 			compo = self.components[i]
 			data = data + compo.compile(more, haveInstructions, glyfTable)
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -677,6 +677,13 @@ class Glyph(object):
 			haveInstructions = haveInstructions | haveInstr
 			self.components.append(component)
 		if haveInstructions:
+			if not data:
+				log.warning(
+					"Composite glyph flags announce presence of instructions, but there are none. %s",
+					[c.glyphName for c in self.components]
+				)
+				return
+
 			numInstructions, = struct.unpack(">h", data[:2])
 			data = data[2:]
 			self.program = ttProgram.Program()


anthrotype:
How was the subset generated? I thought pyftsubset would take care of adjusting the flags when dropping hinting, does it not do that?

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -677,6 +677,13 @@ def decompileComponents(self, data, glyfTable):
 			haveInstructions = haveInstructions | haveInstr
 			self.components.append(component)
 		if haveInstructions:
+			if not data:
+				log.warning(
+					"Composite glyph flags announce presence of instructions, but there are none. %s",
+					[c.glyphName for c in self.components]
+				)
+				return
```
anthrotype: We usually don't do these kinds of optimizations or fixes upon decompile. Only on compile. Maybe you should warn but keep the font as is

anthrotype:
So the question is who or what generated the supposedly broken font to begin with?

jenskutilek:
The subsetted font was generated by Adobe InDesign PDF export. When we remove the warning, the subsetted font will throw a traceback on decompilation with ttx. But maybe that's what we prefer?

The original font was generated by ufo2ft with hinting from my branch of [htic](https://gitlab.com/jenskutilek/htic) causing the problem by adding empty glyph programs to composite glyphs. The issue is fixed there now.

FontValidator sees nothing wrong with the original font btw…

I stumbled across this before, maybe it could be made more consistent in FontTools?

- simple glyphs always need a program attribute, even if the program is empty (could be made non-requirement)
- composite glyphs may or may not have a program attribute, which has influence on the WE_HAVE_INSTRUCTIONS flag even when the program is empty, which may lead to problems as seen above

Here's the broken subsetted font if you would like to have a look: [broken.zip](https://github.com/fonttools/fonttools/files/4281985/broken.zip)

anthrotype:
> The original font was generated by ufo2ft with hinting from my branch of htic...

so, the font was already broken before it was subsetted by indesign?

What you describe is already what fonttools is doing. Simple glyphs always have program (even if empty). Composite glyphs may or may not have the program, depending on the flag.

The problem here is that, even if the flag is set, the composite glyph data ends abruptly. It should at least have the numInstructions set to 0 (useless but still valid program).

If the flag says there are instructions but in fact there aren't and the data is insufficient, then sure we could fail more nicely than raise an exception. I suppose you could still add the empty program just after issuing the warning, instead of returning early. The message could be similar to the one we log just after that, about "too much data" (in your case it would be "not enough data"). Maybe instead of `if not data` check if `len(data) < 2`, for the numInstructions.

And I would prefer not to "fix" it upon compile, but leave it to user to drop empty programs.

jenskutilek:
> so, the font was already broken before it was subsetted by indesign?

No. It had `<instructions/>` on composite glyphs. FontValidator doesn’t complain about this, ttx decompiles/roundtrips it without issue. The `numInstructions` field is there (0000).

Only when the font is subset and embedded into a PDF by InDesign this becomes a problem, because the Adobe PDF export apparently throws away the `numInstructions` field but keeps the WE_HAVE_INSTRUCTIONS flag.

I don't know yet if the subsetting matters, or the instantiation of the variable font, or if any font embedded in a PDF will have this problem.

So it really may be an Adobe bug, but apparently a longstanding and hidden one.

> And I would prefer not to "fix" it upon compile, but leave it to user to drop empty programs.

Because this error is so hard to detect, I thought we could do something to avoid it by changing the logic on compilation of when to set the WE_HAVE_INSTRUCTIONS flag (= set it only if there are actual instructions). Can you think of any use case where having an empty program would be preferred? It just takes up unnecessary space in the file.

The decompilation is a minor issue, because it only occurs when extracting the subsetted font from the PDF and trying to decompile it. And that font actually is broken.

jenskutilek:
I have tried to make sense of the font binary code to see what happens when the font is embedded:

```
FFFF  numContours = -1
0037  xMin = 55
FFF8  yMin = -8
01A3  xMax = 419
02E4  yMax = 740
0222  flags USE_MY_METRICS & MORE_COMPONENTS & ARGS_ARE_XY_VALUES 
01EF  Glyph index 495
00    xOffset = 0
00     yOffset = 0
0106  flags WE_HAVE_INSTRUCTIONS & ARGS_ARE_XY_VALUES & ROUND_XY_TO_GRID 
06FC  Glyph index 1788
77    xOffset = 119
00    yOffset = 0
0000  instructionLength = 0
(next glyph)
FFFF  numContours = -1
```

After embedding:

```
FFFF  numContours = -1
0037  xMin = 55
FFF8  yMin = -8
01A3  xMax = 419
02E4  yMax = 740
0222  flags USE_MY_METRICS & MORE_COMPONENTS & ARGS_ARE_XY_VALUES
01EF  Glyph index 495
00    xOffset = 0
00     yOffset = 0
0107  flags WE_HAVE_INSTRUCTIONS & ARGS_ARE_XY_VALUES & ROUND_XY_TO_GRID & ARG_1_AND_2_ARE_WORDS
06FC  Glyph index 1788
007A  xOffset = 122
0000  yOffset = 0
(next glyph)
0002  numContours = 2
```

I have no idea why the xOffset of the 2nd component changes… but it looks like the PDF generator just cuts off the `instructionLength` uint16.

anthrotype:
I see, so it seems like a bug in adobe pdf font embedding. But you could work around it on your end by ensuring you don't assign empty programs with 0 numIstructions to composite glyphs.

> Can you think of any use case where having an empty program would be preferred? It just takes up unnecessary space in the file.

I don't see a use case, but it is valid..
I don't know, I'm -0 on this, but I could be swayed.

behdad:
I'm leaning towards closing this.

justvanrossum:
I'd merge: it is a safe recovery from a broken font, which we could otherwise not read.

anthrotype:
FWIW, I reiterate my comment from 2 years ago

> you could still add the empty program just after issuing the warning, instead of returning early. The message could be similar to the one we log just after that, about "too much data" (in your case it would be "not enough data"). Maybe instead of if not data check if len(data) < 2, for the numInstructions.
And I would prefer not to "fix" it upon compile, but leave it to user to drop empty programs.

PR #1850, opened 2020-03-10
Annotate LookupFlag ttx dump

simoncozens:
This is a cosmetic improvement to the TTX dumps which helps the user by explaining what bits are set in a LookupFlag and what they mean.

### DIFF
--- a/Lib/fontTools/ttLib/tables/otConverters.py
+++ b/Lib/fontTools/ttLib/tables/otConverters.py
@@ -1569,6 +1569,19 @@ class VarDataValue(BaseConverter):
 	def xmlRead(self, attrs, content, font):
 		return safeEval(attrs["value"])
 
+class LookupFlag(UShort):
+	def xmlWrite(self, xmlWriter, font, value, name, attrs):
+		xmlWriter.simpletag(name, attrs + [("value", value)])
+		if value > 0:
+			flags = []
+			if value & 0x01: flags.append("rightToLeft")
+			if value & 0x02: flags.append("ignoreBaseGlyphs")
+			if value & 0x04: flags.append("ignoreLigatures")
+			if value & 0x08: flags.append("ignoreMarks")
+			if value & 0x10: flags.append("useMarkFilteringSet")
+			if value & 0xff00: flags.append("markAttachmentType")
+			xmlWriter.comment(" ".join(flags))
+		xmlWriter.newline()
 
 converterMapping = {
 	# type		class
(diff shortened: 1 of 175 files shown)


Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -1569,6 +1569,19 @@ def xmlWrite(self, xmlWriter, font, value, name, attrs):
...
+	def xmlWrite(self, xmlWriter, font, value, name, attrs):
+		xmlWriter.simpletag(name, attrs + [("value", value)])
+		if value > 0:
+			flags = []
+			if value & 0x01: flags.append("rightToLeft")
+			if value & 0x02: flags.append("ignoreBaseGlyphs")
+			if value & 0x04: flags.append("ignoreLigatures")
+			if value & 0x08: flags.append("ignoreMarks")
+			if value & 0x10: flags.append("useMarkFilteringSet")
+			if value & 0xff00: flags.append("markAttachmentType")
```
justvanrossum: Would it make sense to add the value to the string? Say, `markAttachmentType[1]`

Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -1569,6 +1569,19 @@ def xmlWrite(self, xmlWriter, font, value, name, attrs):
...
+		xmlWriter.simpletag(name, attrs + [("value", value)])
+		if value > 0:
+			flags = []
+			if value & 0x01: flags.append("rightToLeft")
+			if value & 0x02: flags.append("ignoreBaseGlyphs")
+			if value & 0x04: flags.append("ignoreLigatures")
+			if value & 0x08: flags.append("ignoreMarks")
+			if value & 0x10: flags.append("useMarkFilteringSet")
+			if value & 0xff00: flags.append("markAttachmentType[%i]" % (value >> 8))
+			xmlWriter.comment(" ".join(flags))
```
anthrotype: i know it's unlikely but if value & 0x00E0 (reserved) then this will output an empty comment. How about, either add an `if flags` guard (or write `"reserved"` in the comment)?

simoncozens: Like it, because if we have an `if flags` guard then we don't need the `if value` test.

anthrotype:
thank you, I think it's good to go. Can I merge?

simoncozens:
Yes please!

PR #1859, opened 2020-03-20
update to unicode 13.0.0

justvanrossum:


setup.py
```diff
@@ -52,8 +52,8 @@ def doraise_py_compile(file, cfile=None, dfile=None, doraise=False):
 		# the unicodedata2 extension module doesn't work on PyPy.
 		# Python 3.8 already has Unicode 12.1, so the backport is not needed.
```
dscorbett: This comment should be updated.

PR #1860, opened 2020-03-20
designspaceLib: allow to construct descriptors from keyword args

anthrotype:
... like data classes.

For now, I resisted the temptation to rewrite the whole thing use attrs or dataclasses, as I am time-constrained. I just wanted to be able to initialize axis/source/instance descriptors using their default constructor, which is more idiomatic than creating an incomplete object and then mutating its attributes afterwards.

I kept the same defaults as before (basically all None or False). The defaults are mostly meaningless, which is another pain point when using dsLib -- you don't know what piece of data is actually _required_ by what consumer.
We shall figure that out one day.

### DIFF
--- a/Lib/fontTools/designspaceLib/__init__.py
+++ b/Lib/fontTools/designspaceLib/__init__.py
@@ -1188,14 +1188,34 @@ class DesignSpaceDocument(LogMixin, AsDictMixin):
     def addSource(self, sourceDescriptor):
         self.sources.append(sourceDescriptor)
 
+    def addSourceDescriptor(self, **kwargs):
+        source = self.writerClass.sourceDescriptorClass(**kwargs)
+        self.addSource(source)
+        return source
+
     def addInstance(self, instanceDescriptor):
         self.instances.append(instanceDescriptor)
+
+    def addInstanceDescriptor(self, **kwargs):
+        instance = self.writerClass.instanceDescriptorClass(**kwargs)
+        self.addInstance(instance)
+        return instance
 
     def addAxis(self, axisDescriptor):
         self.axes.append(axisDescriptor)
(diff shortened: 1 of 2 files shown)


justvanrossum:
FWIW, when programmatically creating DesignSpace objects, I find I frequently use helper functions like

```python
def addAxis(doc, name, tag, minimum, default, maximum):
    axis = doc.newAxisDescriptor()
    axis.name = name
    axis.tag = tag
    axis.minimum = minimum
    axis.default = default
    axis.maximum = maximum
    doc.addAxis(axis)
```

I usually don't need to descriptor objects themselves, but do miss a convenient way to add the to the ds document. Your proposed change makes things a little easier, but requires importing the right classes in addition to the DesignSpaceDocument class.

Perhaps we can consider adding methods of the following form:

```python
	def addAxisDescriptor(self, name, tag, minimum, default, maximum):
	    axis = self.newAxisDescriptor()
	    axis.name = name
	    axis.tag = tag
	    axis.minimum = minimum
	    axis.default = default
	    axis.maximum = maximum
	    self.addAxis(axis)
```

anthrotype:
yeah, I like that. I'm on it

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -100,14 +100,32 @@ class SourceDescriptor(SimpleDescriptor):
               'mutedGlyphNames',
               'familyName', 'styleName']
 
-    def __init__(self):
-        self.filename = None
+    def __init__(
+        self,
+        *,
```
madig: :+1:

justvanrossum:
Awesome, thanks!

PR #1861, opened 2020-03-23
CPAL.Color.fromRGBA

jansindl3r:
there wasn't option to initialize Color using RGBA, but instead BGRA or HEX. User inputs are mostly RGBA, before I had to use uncomfortably and unreadably `Color(*color[:3][::-1], color[3])` or switch it aside.

### DIFF
--- a/Lib/fontTools/ttLib/tables/C_P_A_L_.py
+++ b/Lib/fontTools/ttLib/tables/C_P_A_L_.py
@@ -82,8 +82,8 @@ class table_C_P_A_L_(DefaultTable.DefaultTable):
 		if self.version >= 1:
 			offsetToFirstColorRecord += 12
 		header = struct.pack(">HHHHL", self.version,
-                                     self.numPaletteEntries, len(self.palettes),
-                                     numColorRecords, offsetToFirstColorRecord)
+									 self.numPaletteEntries, len(self.palettes),
+									 numColorRecords, offsetToFirstColorRecord)
 		if self.version == 0:
 			dataList = [header, colorRecordIndices, colorRecords]
 		else:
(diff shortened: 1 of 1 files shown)


anthrotype:
i am ok to add a classmethod `fromRGBA`, but you don't need to remove the namedtuple base class. Please put that back and remove the custom `__init__`

jansindl3r:
done

Lib/fontTools/ttLib/tables/C_P_A_L_.py
```diff
@@ -236,12 +236,12 @@ def fromXML(self, name, attrs, content, ttFont):
...
 
-	def hex(self):
-		return "#%02X%02X%02X%02X" % (self.red, self.green, self.blue, self.alpha)
-
 	def __repr__(self):
 		return self.hex()
 
+	def hex(self):
+		return "#%02X%02X%02X%02X" % (self.red, self.green, self.blue, self.alpha)
+
```
anthrotype: please undo this change (moving `hex` method) as it is just diff noise, thanks

jansindl3r: ye, I am learning this culture of contributing. I saw it as very small improvement in legibility

justvanrossum: It changed the object type to no longer be a tuple subclass, so it would not be backwards compatible.

anthrotype: that change was undone, it's still a namedtuple

justvanrossum: I know, I was just in an explainy mood...

anthrotype: oh ok, thank you! :)

PR #1866, opened 2020-03-30
Pass BASE table through subsetter by default.

punchcutter:


anthrotype:
can you expand a little bit about the rationale for this change?

punchcutter:
I can't see any reason why the BASE table would be ignored by default. I was subsetting a font with BASE table and with no explanation the BASE table was just dropped. Even with `--passthrough-tables` it still gets dropped because it's in the `_drop_tables_default` list. I can understand default dropping some of those other tables, but why BASE?

anthrotype:
ok yes, dropping by default may be too blunt.

> Even with --passthrough-tables it still gets dropped

hm that's not good, indeed. Can we just fix that?

Ideally the subsetter would trim the BASE script coverage to match the subset. But I have to refresh my memory about BASE table to see how that would work.

punchcutter:
Yes, I guess ideally it would trim to match the subset if necessary.

anthrotype:
I think here we should simply remove BASE from the `_drop_tables_deafault` list, so that it respects `--no-subset-tables+=BASE` or `--passthrough-tables`, but it is still dropped by default as subsetter does not know how to subset.

anthrotype:
Even though it's cumbersome, this should currently work to prevent subsetter to drop BASE table:

`--drop-tables-=BASE --no-subset-tables+=BASE`

behdad:
I'm happy to pass it through by default, since that does produce a correct font. @punchcutter can you rebase this plese?

RickyDaMa:
This would be very helpful for some of the projects I'm helping to maintain. I'll take a look at rebasing this next week perhaps

RickyDaMa:
#3949 was merged so this PR can be closed now. Thanks @punchcutter for the original PR, my rebase retained your authorship 🚀

benkiel:
Closing, addressed in https://github.com/fonttools/fonttools/pull/3949, thanks @RickyDaMa

PR #1270, opened 2018-06-07
WIP: fold ufoLib2 into fontTools.ufoLib

anthrotype:
I imported the https://github.com/fonttools/ufoLib2 repo with git subtree, then moved the files and adjusted the imports. 

#1095

PR #1868, opened 2020-03-31
merge cu2qu into fontTools.cu2qu

anthrotype:
Rod (@rsheeter) and I would like to contribute the code from https://github.com/googlefonts/cu2qu to FontTools, so that we can make it more readily available under `fontTools.cu2qu` and `fontTools.pens.cu2quPen`.

Only the core cu2qu module and the cu2qu.pens are copied over in this PR. The `cu2qu.ufo` module and the `cu2qu` command-line tool will continue to live in a separate `cu2qu` package, as they depend on defcon. 
Besids, the separate `cu2qu` project also comes with an optional native module (using Cython), whereas this new `fontTools.cu2qu` will stay pure-python like the rest of fontTools.

I have merged the two repos so that we keep the git history. I created a branch in cu2qu, where I moved all the files to be copied to a new subfolder, deleted the rest of the files, then fetched the branch from my local fontTools repo, did `git merge --allow-unrelated-histories` and finally moved the files to destination.


dscorbett:
Since cu2qu uses the Apache 2.0 license, doesn’t each of these files need an Apache 2.0 header to clarify that they don’t use the MIT license like the rest of fonttools, and doesn’t the top-level file LICENSE need to mention this?

anthrotype:
@dscorbett thanks, you're right. I have added the header to the files that were missing it, and appended the text from Apache 2.0 license (as found in the cu2qu repository) to the `LICENSE.external` file.

anthrotype:
I decided to also copy over the `cu2qu.ufo` module (containing the key function `fonts_to_quadratic` for converting multiple UFOs and keep interpolatability), since it does not import defcon (nor ufoLib2, fontParts, etc) directly, but can work with any defcon-like Font objects.
So it's ok to add cu2qu.ufo module to fonttools without adding additional explicit dependencies.
The dependency if you like is injected implicitly by the client.
I also added the seldom used cu2qu.cli command-line tool (now accessible as `fonttools cu2qu`). This one has a concrete dependency on ufoLib2, but one may chose to ignore it.
The ufo_test.py and cli_test.py are skipped if ufoLib2 is not installed.

anthrotype:
so that means the whole of cu2qu project has been copied over now, not just the core cu2qu module and pens.
The only difference from the original is the lack of cython support (though one could still manually run the `cythonize` tool directly on `Lib/fontTools/cu2qu/cu2qu.py`).
I will see if I can add an option to setup.py to optionally compile the native module.
But I don't think I will publish native platform-specific wheels for fonttools just because of cu2qu.
Better to keep the fonttools official distributions pure-python.

behdad:
So why can't we just use fontTools.ufoLib for everything in there?

For the record, the reason previously I kept this module outside fontTools was that it implements a specific heuristic as opposed to something followed from specs (which is more the rest of fontTools).

But I also think it's a very good heuristic targeted for TrueType glyf table.  So I'm happy you are moving it.  I think it's fair to say it has proved itself.  I don't think we got a bug about the conversion, *ever*.

anthrotype:
> So why can't we just use fontTools.ufoLib for everything in there?

because fontTools.ufoLib is a bit too low level (only exposes the reader/writer classes), it is nicer to work with ufoLib2.Font. Maybe one day I will take up again #1270...

twardoch:
This is great!

miguelsousa:
> I don't think we got a bug about the conversion, ever.

Sorry to be the bearer of bad news, but we just ran into some conversion results that are noticeably off. See https://github.com/googlefonts/cu2qu/issues/190

behdad:
> Sorry to be the bearer of bad news, but we just ran into some conversion results that are noticeably off. See googlefonts/cu2qu#190

Commented there. Doesn't look like a curve conversion error to me.

PR #1876, opened 2020-04-14
[feaLib] Raise exception when GSUB statement doesn't match a rule.

tomarcher101:
Raise an exception when a substitute statement doesn't match any of the GSUB lookup rules. Add tests that trigger and test said exception.

### DIFF
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -827,7 +827,13 @@ class Parser(object):
                 location)
 
         # GSUB lookup type 6: Chaining contextual substitution.
-        assert len(new) == 0, new
+        if len(new) != 0:
+            raise FeatureLibError(
+                'This is not a valid substitution. If attempting contextual '
+                'substitution, remember to mark the input sequence with a "\'"',
+                location
+            )
+
         rule = self.ast.ChainContextSubstStatement(
             old_prefix, old, old_suffix, lookups, location=location)
         return rule
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -826,14 +826,15 @@ class Parser(object):
                 'is not supported',
                 location)
 
-        # GSUB lookup type 6: Chaining contextual substitution.
+        # If there are remaining glyphs to parse, this is an invalid GSUB statement
         if len(new) != 0:
             raise FeatureLibError(
-                'This is not a valid substitution. If attempting contextual '
-                'substitution, remember to mark the input sequence with a "\'"',
+                'This is an invalid GSUB statement. Make sure your statement '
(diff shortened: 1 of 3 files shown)


Tests/feaLib/data/GSUB_error.fea
```diff
@@ -0,0 +1,13 @@
+# Trigger a parser error in the funciton parse_subsitutie_ in order to improve the error message.
```
belluzj, line 1:
```suggestion
# Trigger a parser error in the function parse_substitute_ in order to improve the error message.
```

Lib/fontTools/feaLib/parser.py
```diff
@@ -827,7 +827,13 @@ def parse_substitute_(self):
                 location)
 
         # GSUB lookup type 6: Chaining contextual substitution.
-        assert len(new) == 0, new
+        if len(new) != 0:
+            raise FeatureLibError(
+                'This is not a valid substitution. If attempting contextual '
+                'subsitution, remember to mark the input sequence with a "\'"',
```
madig, line 833:
```suggestion
                'substitution, remember to mark the input sequence with a "\'"',
```

Lib/fontTools/feaLib/parser.py
```diff
@@ -827,7 +827,13 @@ def parse_substitute_(self):
                 location)
 
         # GSUB lookup type 6: Chaining contextual substitution.
-        assert len(new) == 0, new
+        if len(new) != 0:
```
anthrotype, line 830: this negative condition is a bit obscure. I had to look at the code above to understand what that means. Basically, if we haven't handled the sequence of glyphs after by/from yet by the time we reach the end of the parse_substitute_ method, it means there is something wrong with the sub rule, it's unknown or invalid. It doesn't necessarily means that one wanted to write a contextual substitution and forgot the apostrophe, but I can see how this may be frequent.
Maybe add a comment.

tomarcher101: thanks for the advice! I'll make the error message a bit less specific (remove the reference to contextual substitution).

Lib/fontTools/feaLib/parser.py
```diff
@@ -826,8 +826,15 @@ def parse_substitute_(self):
                 'is not supported',
                 location)
 
+        # If there are remaining glyphs to parse, this is an invalid GSUB statement
+        if len(new) != 0:
+            raise FeatureLibError(
+                'This is an invalid GSUB statement. Make sure your statement '
+                'follows the format of one of the GSUB rule.',
```
anthrotype, line 833: I still think the message is still quite verbose and the second sentence doesn't add much info to the fact that the statement is invalid. I'd simply write "Invalid substitution statement" (then the parser will add the location within the file and a user can debug the issue).

PR #1878, opened 2020-04-17
[feaLib] Fix AttributeError: 'NoneType' object has no attribute 'getAlternateGlyphs'

anthrotype:
this was originally reported in https://github.com/googlefonts/fontmake/issues/648 by @m4rc1e and @vv-monsalve

If you have FEA like the one below, with `aalt` feature referencing another feature that in turn contains some chain contextual substitution lookups, one may get an `AttributeError: 'NoneType' object has no attribute 'getAlternateGlyphs'`: 

```
lookup CNTXT_LIGS {
    sub f i by f_i;
    sub c t by c_t;
} CNTXT_LIGS;

lookup CNTXT_SUB {
    sub n by n.end;
    sub s by s.end;
} CNTXT_SUB;

feature calt {
    sub [a e i o u] f' lookup CNTXT_LIGS i' n' lookup CNTXT_SUB;
    sub [a e i o u] c' lookup CNTXT_LIGS t' s' lookup CNTXT_SUB;
} calt;

feature aalt {
    feature calt;
} aalt;
```

I copied the snippet above from the [FEA spec](http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#5fi-specifying-a-chain-sub-rule-and-marking-sub-runs), and simply added an `aalt` feature block referencing the `calt` feature that has the chain contextual lookups.

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -1311,9 +1311,10 @@ class ChainContextSubstBuilder(LookupBuilder):
             if lookups == self.SUBTABLE_BREAK_:
                 continue
             for lookup in lookups:
-                alts = lookup.getAlternateGlyphs()
-                for glyph, replacements in alts.items():
-                    result.setdefault(glyph, set()).update(replacements)
+                if lookup is not None:
+                    alts = lookup.getAlternateGlyphs()
+                    for glyph, replacements in alts.items():
+                        result.setdefault(glyph, set()).update(replacements)
         return result
 
     def find_chainable_single_subst(self, glyphs):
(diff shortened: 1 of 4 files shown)


anthrotype:
This the full test failure with the reproducer: https://travis-ci.org/github/fonttools/fonttools/jobs/676294250#L613-L632

The fix is simply to check for `None`-ness before attempting to call `lookup.getAlternateGlyphs()`.

That's because the list of lookups as parsed by `feaLib.Parser.get_glyph_pattern_` method may contain `None` values, for all the `'` marked glyphs in the input sequence that are not followed by a `lookup` reference, see:

https://github.com/fonttools/fonttools/blob/a6e050ab092755f66be9595e96c5185c6d18a6ca/Lib/fontTools/feaLib/parser.py#L407-L421

Issue #1881, opened 2020-04-20
[subset] FeatureVariations subsetting is wrong


anthrotype:
originally reported by @behdad at https://github.com/harfbuzz/harfbuzz/issues/2334#issue-602270174
> We cannot drop FeatureVariationsRecords that have empty substitutions. That changes the behavior. A FeatureVariationsRecords stops the search when the conditions match; so even if there's no susbstitutions we cannot drop the record as that will keep the search going.
> The only optimizations we can do are 1. dropping from the end of the array any records with no susbtitutions, and 2. dropping records earlier with no susbstitution only after analysing the conditions and proving that the dropping does not change functionality.
> Both of those are unnecessary right now. We should fix the functionality.
> This seems to be wrong in FontTools as well:
> https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/subset/__init__.py#L1325

behdad:
I'll try submitting a minimal fix and a simple pruning but won't be able to contribute tests.  I hope there are tests for those code path already.

anthrotype:
thanks Behdad, i'll fix up the tests as needed

anthrotype:
To test this dropping-from-end-of-array of FeatureVariationRecords with no substitutions, we need create a VF where FeatureTableSubstitutions contain more than one SubstitutionRecords referencing different FeatureIndex; then we'd drop one feature while keeping another and assert that the thus emptied FeatureVariationRecords at the end of the FeatureVariations are dropped, and that empty records that precede non-empty ones are kept.
The current featureVars API can only create FeatureVariations for a one feature at a time (e.g. rvrn). Making a test font manually is not fun.. I will try again when I get the chance. I'll keep this open in the meantime as reminder.

anthrotype:
@behdad when the subsetter has dropped all the existing FeatureVariationRecords, should it not also remove the emptied FeatureVariations table and downgrade the GSUB/GPOS table version to 0x00010000 (inside `prune_post_subset`)?

behdad:
> @behdad when the subsetter has dropped all the existing FeatureVariationRecords, should it not also remove the emptied FeatureVariations table and downgrade the GSUB/GPOS table version to 0x00010000 (inside `prune_post_subset`)?

Yes.  I thought that's implemented.

behdad:
[...]

PR #1882, opened 2020-04-20
[subset] FeatureVariations subsetting is wrong

behdad:
Fixes https://github.com/fonttools/fonttools/issues/1881


anthrotype:
i'm going to add a test case

PR #1887, opened 2020-04-21
Eliminate PDF documentation builds on Read The Docs

chrissimpkins:
The [PDF builds on RTD fail with the current docstrings/configuration](https://readthedocs.org/projects/fonttools/builds/10881463/).  This PR transitions to distribution of [static HTML site zip archives](https://fonttools.readthedocs.io/_/downloads/en/latest/htmlzip/) and [ePub](https://fonttools.readthedocs.io/_/downloads/en/latest/epub/) builds for those who want to download pre-built content.  We can troubleshoot this if PDF documentation downloads are useful to anyone.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -14,7 +14,9 @@ sphinx:
   fail_on_warning: false
 
 # Optionally build your docs in additional formats such as PDF and ePub
-formats: all
+formats:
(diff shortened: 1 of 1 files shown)


chrissimpkins:
confirmed that the docs build completes without errors after these changes

PR #1891, opened 2020-04-22
[Sphinx docs] indent line breaks in lists

chrissimpkins:
Sphinx doesn't like the list item line break formatting that is used in the README.rst file (#1888).  This is a Sphinx doc formatting PR for list items on the home page of the RTD documentation.

### DIFF
--- a/Doc/source/index.rst
+++ b/Doc/source/index.rst
@@ -152,7 +152,9 @@ For example:
 This command will install fonttools, as well as the optional dependencies that
 are required to unlock the extra features named "ufo", etc.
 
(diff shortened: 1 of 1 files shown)


dscorbett:
Instead of removing line breaks, you can just indent the wrapped lines. For example, this:
```reStructuredText
* foo
bar
```
should be this:
```reStructuredText
* foo
  bar
```

chrissimpkins:
> Instead of removing line breaks, you can just indent the wrapped lines. 

Tyvm David!

b67c7f5de310c3f06b99fd314e436b2fb9b12903

Issue #1896, opened 2020-04-26
[feaLib] assertion error mixing pos and sub contextual chaining rules


simoncozens:
Because OpenType is... well, it is what it is - you can create a contextual chaining lookup either with a "pos" statement or a "sub" statement. In particular, you can create a contextual chaining lookup with a "position" statement even if the lookup you chain to is a substitution rule, and vice versa. The following feature code:

```
lookup dummy1 { pos uni0061 50; } dummy1;

feature kern {
    sub uni0061' lookup dummy1 uni0062;
} kern;
```

may not be entirely _sensible_, but it should cause a warning, not a crash:

```
  File "/usr/local/lib/python3.7/site-packages/fontTools/ttLib/ttFont.py", line 172, in save
    writer_reordersTables = self._save(tmp)
  File "/usr/local/lib/python3.7/site-packages/fontTools/ttLib/ttFont.py", line 211, in _save
    self._writeTable(tag, writer, done, tableCache)
  File "/usr/local/lib/python3.7/site-packages/fontTools/ttLib/ttFont.py", line 632, in _writeTable
    tabledata = self.getTableData(tag)
  File "/usr/local/lib/python3.7/site-packages/fontTools/ttLib/ttFont.py", line 650, in getTableData
    return self.tables[tag].compile(self)
  File "/usr/local/lib/python3.7/site-packages/fontTools/ttLib/tables/otBase.py", line 69, in compile
    self.table.compile(writer, font)
  File "/usr/local/lib/python3.7/site-packages/fontTools/ttLib/tables/otBase.py", line 728, in compile
    conv.write(writer, font, table, value)
  File "/usr/local/lib/python3.7/site-packages/fontTools/ttLib/tables/otConverters.py", line 557, in write
    value.compile(subWriter, font)
  File "/usr/local/lib/python3.7/site-packages/fontTools/ttLib/tables/otBase.py", line 691, in compile
    conv.write(writer, font, table, value, i)
  File "/usr/local/lib/python3.7/site-packages/fontTools/ttLib/tables/otConverters.py", line 557, in write
    value.compile(subWriter, font)
  File "/usr/local/lib/python3.7/site-packages/fontTools/ttLib/tables/otBase.py", line 691, in compile
    conv.write(writer, font, table, value, i)
  File "/usr/local/lib/python3.7/site-packages/fontTools/ttLib/tables/otConverters.py", line 557, in write
    value.compile(subWriter, font)
[...]

PR #1897, opened 2020-04-27
[feaLib] Check lookup index in chaining lookups

khaledhosny:
Raise if `lookup_index` is None, which would happen only of the lookup referenced belongs to the wrong table. The error message needs to be less cryptic, though.

Fixes https://github.com/fonttools/fonttools/issues/1896

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -1263,6 +1263,9 @@ class ChainContextPosBuilder(LookupBuilder):
             st.PosLookupRecord = []
             for sequenceIndex, l in enumerate(lookups):
                 if l is not None:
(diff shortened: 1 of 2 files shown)


anthrotype:
> The error message needs to be less cryptic, though.

Indeed. How about "The lookup referenced belongs to the wrong table"

khaledhosny:
> > The error message needs to be less cryptic, though.
> 
> Indeed. How about "The lookup referenced belongs to the wrong table"

I considered that, but felt too-specific when the check is generic (there might be other cases now or in the feature that cause `lookup_index` to be None). What about “Missing index of the specified lookup, might be a positioning lookup.” (and vice versa)?

anthrotype:
Sgtm

PR #1903, opened 2020-04-28
subset: remove FeatureVariations, downgrade GSUB/GPOS to version 1 when empty

anthrotype:
https://github.com/fonttools/fonttools/issues/1881#issuecomment-620719882

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -1631,11 +1631,11 @@ def prune_post_subset(self, font, options):
 	#if table.ScriptList and not table.ScriptList.ScriptRecord:
 	#	table.ScriptList = None
 
-	if not table.FeatureList and hasattr(table, 'FeatureVariations'):
-		table.FeatureVariations = None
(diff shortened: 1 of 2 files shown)


Lib/fontTools/subset/__init__.py
```diff
@@ -1631,11 +1631,11 @@ def prune_post_subset(self, font, options):
 	#if table.ScriptList and not table.ScriptList.ScriptRecord:
 	#	table.ScriptList = None
 
-	if not table.FeatureList and hasattr(table, 'FeatureVariations'):
-		table.FeatureVariations = None
+	if hasattr(table, 'FeatureVariations'):
+		if not (table.FeatureList and table.FeatureVariations.FeatureVariationRecord):
```
behdad, line 1635: The `table.FeatureList` check should be unnecessary as if there's no features left there shouldn't be any non-empty substitutions left so we should have trimmed down FeatureVariationRecord completely.  But then again, no harm doing it the way you are.

PR #1905, opened 2020-04-29
[feaLib] Support multiple lookups per glyph position

simoncozens:
This is an implementation of the suggestion at adobe-type-tools/afdko#1119 to support the syntax:

```
sub a' lookup lu_1 lookup lu_2;
```

I know it's generally not a great idea to add your own non-standard extensions to a commonly used public spec, but there appears to be support for this concept on the Adobe side. My plan is to get this reference implementation working here, see what I can do with the makeotf code and get the spec changed; somebody has to go first.

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -546,8 +546,9 @@ class ChainContextPosStatement(Statement):
                 res += " ".join(g.asFea() for g in self.prefix) + " "
             for i, g in enumerate(self.glyphs):
                 res += g.asFea() + "'"
-                if self.lookups[i] is not None:
-                    res += " lookup " + self.lookups[i].name
+                if self.lookups[i]:
+                    for lu in self.lookups[i]:
+                        res += " lookup " + lu.name
                 if i < len(self.glyphs) - 1:
                     res += " "
             if len(self.suffix):
@@ -578,8 +579,9 @@ class ChainContextSubstStatement(Statement):
                 res += " ".join(g.asFea() for g in self.prefix) + " "
             for i, g in enumerate(self.glyphs):
                 res += g.asFea() + "'"
-                if self.lookups[i] is not None:
-                    res += " lookup " + self.lookups[i].name
+                if self.lookups[i]:
+                    for lu in self.lookups[i]:
+                        res += " lookup " + lu.name
                 if i < len(self.glyphs) - 1:
                     res += " "
             if len(self.suffix):
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -797,9 +797,10 @@ class Builder(object):
         If an input name is None, it gets mapped to a None LookupBuilder.
         """
         lookup_builders = []
-        for lookup in lookups:
-            if lookup is not None:
-                lookup_builders.append(self.named_lookups_.get(lookup.name))
+        for lookuplist in lookups:
+            if lookuplist is not None:
+                for lookup in lookuplist:
+                    lookup_builders.append(self.named_lookups_.get(lookup.name))
             else:
                 lookup_builders.append(None)
         return lookup_builders
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -203,9 +203,11 @@ class Builder(object):
                 raise FeatureLibError("Feature %s has not been defined" % name,
                                       location)
             for script, lang, feature, lookups in feature:
-                for lookup in lookups:
-                    for glyph, alts in lookup.getAlternateGlyphs().items():
-                        alternates.setdefault(glyph, set()).update(alts)
+                for lookuplist in lookups:
+                    if type(lookuplist) != list: lookuplist = [ lookuplist ]
+                    for lookup in lookuplist:
+                        for glyph, alts in lookup.getAlternateGlyphs().items():
+                            alternates.setdefault(glyph, set()).update(alts)
         single = {glyph: list(repl)[0] for glyph, repl in alternates.items()
                   if len(repl) == 1}
         # TODO: Figure out the glyph alternate ordering used by makeotf.
@@ -799,8 +801,8 @@ class Builder(object):
         lookup_builders = []
         for lookuplist in lookups:
             if lookuplist is not None:
-                for lookup in lookuplist:
-                    lookup_builders.append(self.named_lookups_.get(lookup.name))
+                lookup_builders.append([ self.named_lookups_.get(l.name)
+                    for l in lookuplist])
             else:
                 lookup_builders.append(None)
         return lookup_builders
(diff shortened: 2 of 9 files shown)


Lib/fontTools/feaLib/ast.py
```diff
@@ -546,8 +546,9 @@ def asFea(self, indent=""):
                 res += " ".join(g.asFea() for g in self.prefix) + " "
             for i, g in enumerate(self.glyphs):
                 res += g.asFea() + "'"
-                if self.lookups[i] is not None:
-                    res += " lookup " + self.lookups[i].name
+                if self.lookups[i]:
+                    for lu in self.lookups[i]:
```
khaledhosny, line 550: The `if` check is superfluous, the loop will do nothing if the list is empty.

Lib/fontTools/feaLib/builder.py
```diff
@@ -203,9 +203,11 @@ def build_feature_aalt_(self):
                 raise FeatureLibError("Feature %s has not been defined" % name,
                                       location)
             for script, lang, feature, lookups in feature:
-                for lookup in lookups:
-                    for glyph, alts in lookup.getAlternateGlyphs().items():
-                        alternates.setdefault(glyph, set()).update(alts)
+                for lookuplist in lookups:
+                    if type(lookuplist) != list: lookuplist = [ lookuplist ]
```
khaledhosny, line 207: `isinstance()`?

Lib/fontTools/feaLib/ast.py
simoncozens: No, if there's no lookup at that position, then there is an actual None in there that will not iterate.

simoncozens:
Note that this is now implemented in AFDKO.

anthrotype:
Implemented as in "merged" into master? Was the FEA spec also updated?

anthrotype:
I see that's the case in https://github.com/adobe-type-tools/afdko/pull/1132, ok.
I'll take a look at this soon, thanks

Lib/fontTools/feaLib/parser.py
```diff
@@ -404,8 +404,10 @@ def parse_glyph_pattern_(self, vertical):
             else:
                 values.append(None)
 
-            lookup = None
-            if self.next_token_ == "lookup":
+            lookuplist = None
+            while self.next_token_ == "lookup":
+                if not lookuplist:
```
khaledhosny, line 409:
```suggestion
                if lookuplist is None:
```

Lib/fontTools/feaLib/builder.py
```diff
@@ -1310,30 +1317,36 @@ def build(self):
...
+            if lookuplist == self.SUBTABLE_BREAK_:
                 continue
-            for lookup in lookups:
-                if lookup is not None:
-                    alts = lookup.getAlternateGlyphs()
-                    for glyph, replacements in alts.items():
-                        result.setdefault(glyph, set()).update(replacements)
+            for lookups in lookuplist:
+                if not isinstance(lookups, list):
+                    lookups = [ lookups ]
```
khaledhosny, line 1344:
```suggestion
                    lookups = [lookups]
```

Lib/fontTools/feaLib/builder.py
```diff
@@ -203,9 +203,12 @@ def build_feature_aalt_(self):
                 raise FeatureLibError("Feature %s has not been defined" % name,
                                       location)
             for script, lang, feature, lookups in feature:
-                for lookup in lookups:
-                    for glyph, alts in lookup.getAlternateGlyphs().items():
-                        alternates.setdefault(glyph, set()).update(alts)
+                for lookuplist in lookups:
+                    if not isinstance(lookuplist, list):
+                        lookuplist = [ lookuplist ]
```
khaledhosny, line 208:
```suggestion
                        lookuplist = [lookuplist]
```

Lib/fontTools/feaLib/builder.py
```diff
@@ -797,9 +800,10 @@ def find_lookup_builders_(self, lookups):
         If an input name is None, it gets mapped to a None LookupBuilder.
         """
         lookup_builders = []
-        for lookup in lookups:
-            if lookup is not None:
-                lookup_builders.append(self.named_lookups_.get(lookup.name))
+        for lookuplist in lookups:
+            if lookuplist is not None:
+                lookup_builders.append([ self.named_lookups_.get(l.name)
```
khaledhosny, line 805:
```suggestion
                lookup_builders.append([self.named_lookups_.get(l.name)
```

Lib/fontTools/feaLib/builder.py
```diff
@@ -1259,18 +1263,21 @@ def build(self):
...
-                        raise FeatureLibError('Missing index of the specified '
-                            'lookup, might be a substitution lookup',
-                            self.location)
-                    rec = otTables.PosLookupRecord()
-                    rec.SequenceIndex = sequenceIndex
-                    rec.LookupListIndex = l.lookup_index
-                    st.PosLookupRecord.append(rec)
+            for sequenceIndex, ll in enumerate(lookups):
+                if ll is not None:
+                    if type(ll) is not list: ll = [ll]
```
khaledhosny, line 1270: `isinstance()`?

khaledhosny: May  be also a new line after the colon and better variable names than `l` and `ll` (they remind me of my peronal code, `a`, `aa`, `aaa`, `aaaa`).

Lib/fontTools/feaLib/builder.py
```diff
@@ -1310,30 +1317,36 @@ def build(self):
...
-                        raise FeatureLibError('Missing index of the specified '
-                            'lookup, might be a positioning lookup',
-                            self.location)
-                    rec = otTables.SubstLookupRecord()
-                    rec.SequenceIndex = sequenceIndex
-                    rec.LookupListIndex = l.lookup_index
-                    st.SubstLookupRecord.append(rec)
+            for sequenceIndex, ll in enumerate(lookups):
+                if ll is not None:
+                    if type(ll) is not list: ll = [ll]
```
khaledhosny, line 1324: Same as above.

Lib/fontTools/feaLib/builder.py
```diff
@@ -1310,30 +1317,36 @@ def build(self):
...
+                    for l in ll:
+                        st.SubstCount = st.SubstCount + 1
```
khaledhosny, line 1326:
```suggestion
                        st.SubstCount += 1
```

Lib/fontTools/feaLib/builder.py
```diff
@@ -1259,18 +1263,21 @@ def build(self):
...
-                            self.location)
-                    rec = otTables.PosLookupRecord()
-                    rec.SequenceIndex = sequenceIndex
-                    rec.LookupListIndex = l.lookup_index
-                    st.PosLookupRecord.append(rec)
+            for sequenceIndex, ll in enumerate(lookups):
+                if ll is not None:
+                    if type(ll) is not list: ll = [ll]
+                    for l in ll:
+                        st.PosCount = st.PosCount + 1
```
khaledhosny, line 1272:
```suggestion
                        st.PosCount += 1
```

Lib/fontTools/feaLib/builder.py
```diff
@@ -1259,18 +1263,23 @@ def build(self):
...
-                            'lookup, might be a substitution lookup',
-                            self.location)
-                    rec = otTables.PosLookupRecord()
-                    rec.SequenceIndex = sequenceIndex
-                    rec.LookupListIndex = l.lookup_index
-                    st.PosLookupRecord.append(rec)
+            for sequenceIndex, lookupList in enumerate(lookups):
+                if lookupList is not None:
+                    if not isinstance(lookupList, list):
+                        # Can happen with synthesised lookups
```
anthrotype, line 1271: what are "synthesised lookups"?

Lib/fontTools/feaLib/builder.py
```diff
@@ -1310,30 +1319,38 @@ def build(self):
...
-                if l is not None:
-                    if l.lookup_index is None:
-                        raise FeatureLibError('Missing index of the specified '
-                            'lookup, might be a positioning lookup',
-                            self.location)
-                    rec = otTables.SubstLookupRecord()
-                    rec.SequenceIndex = sequenceIndex
-                    rec.LookupListIndex = l.lookup_index
-                    st.SubstLookupRecord.append(rec)
+            for sequenceIndex, lookupList in enumerate(lookups):
```
anthrotype, line 1324: minor: a find the use of `lookupList` (camelCase) and `lookuplist` (lowercase) here and there a bit distracting

Lib/fontTools/feaLib/builder.py
```diff
@@ -1310,30 +1319,38 @@ def build(self):
...
+                if lookupList is not None:
+                    if not isinstance(lookupList, list):
+                        # Can happen with synthesised lookups
+                        lookupList = [ lookupList ]
```
anthrotype, line 1328: i'm reluctant to run `black` on existing modules as it creates noise in the git blame/log, but we can at least try to be close to the black (or PEP8) style for the new lines added to existing modules. Sure, whitespace is innocuous, but I still find it distracting.

Lib/fontTools/feaLib/builder.py
simoncozens: An inline chaining substitution (`sub foo bar' by bar.alt`). But now that I look at it, I wonder if this can be mitigated by returning an array in `get_chained_lookup_`.

anthrotype: if possible, I would avoid so many isinstance checks and just enforce that that is a list of lists at the source

PR #1913, opened 2020-05-03
[doc] Add help options to fonttools CLI

simoncozens:
Gets each __main__ package to describe what they do, and feeds that to a --help command.


Lib/fontTools/cu2qu/__main__.py
```diff
@@ -1,6 +1,7 @@
 import sys
 from .cli import main
 
+description = "Convert a UFO font with cubic curves to quadratic curves"
```
khaledhosny, line 4: Shouldn’t it be `__doc__` or a similarly named variable?

simoncozens: Yeah, using docstrings is clearly the right approach. OK, rewrite incoming...

Lib/fontTools/cu2qu/__main__.py
```diff
@@ -1,6 +1,7 @@
+"""Convert a UFO font with cubic curves to quadratic curves"""
+
 import sys
 from .cli import main
 
-
```
khaledhosny, line 4: I’m sure @anthrotype will tell you that there should be two spaces after import statements based on some PEPsomething 😇

simoncozens: One more time then...

Lib/fontTools/help/__main__.py
```diff
@@ -0,0 +1,35 @@
...
+
+def show_help_list():
+  path = fontTools.__path__[0]
+  for pkg in find_packages(path):
+      qualifiedPkg = "fontTools."+pkg
+      description = get_description(qualifiedPkg)
+      if description:
+        print("fontools %-10s %s" % (pkg, description))
+        pkgpath = path + '/' + qualifiedPkg.replace('.', '/')
+        if (sys.version_info.major == 3 and sys.version_info.minor < 6):
```
anthrotype, line 24: we only support python >= 3.6 now, so we don't need this if branch

Lib/fontTools/help/__main__.py
```diff
@@ -0,0 +1,35 @@
...
+import sys
+from setuptools import find_packages
+from pkgutil import iter_modules
+import fontTools
+import importlib
+
+
+def get_description(pkg):
+  try:
+    return __import__(pkg+".__main__",globals(),locals(),["__doc__"]).__doc__
```
anthrotype, line 12: can you run this through [black](https://github.com/psf/black) so we get consistent formatting? I like to do this at least for newly added modules

Lib/fontTools/help/__main__.py
```diff
@@ -0,0 +1,35 @@
...
+            for _, name, ispkg in iter_modules([pkgpath]):
+                if get_description(pkg+ '.' + name):
+                    modules.add(pkg + '.' + name)
+        else:
+            for info in iter_modules([pkgpath]):
+                if get_description(pkg+ '.' + info.name):
+                    modules.add(pkg + '.' + info.name)
+
+if __name__ == '__main__':
+  print("fonttools v%s\n" % fontTools.__version__)
```
anthrotype, line 34: argparse prints to stderr, so for consistency if we call print() in similar scenarios we should do `print(..., file=sys.stderr)`

Lib/fontTools/__main__.py
```diff
@@ -20,6 +18,10 @@ def main(args=None):
 	# can be added.  Should we just try importing the fonttools
 	# module first and try without if it fails?
 
+	if len(sys.argv) < 2:
+		sys.argv.append("help")
+	if sys.argv[1] == "-h" or sys.argv[1] == "--help":
+		sys.argv[1] = "help"
```
anthrotype, line 24: why making a `help` package with `__init__.py` and `__main__.py`? Can't we do with a single `help.py` module instead? I don't think we plan to add additional sub-modules to this.

simoncozens: This is leveraging the fact that the `fonttools` binary calls `verb.__main__`.

anthrotype: No, It runs any module as if it is called `__main__`, but works on standalone modules too. Try

simoncozens: `ImportError: No module named fontTools.help.__main__; 'fontTools.help' is a package and cannot be directly executed`

anthrotype: Exactly! Don't have a package folder, just a fontTools/help.py module

simoncozens: Got it. Here we go.

anthrotype:
sorry Simon, I merged this accidentally. Then realized that there were two remaining issues.
1) this is only searching for sub-packages in the top level fontTools directory that have a `__main__` submodule; as such, it is ignoring other `runnable` modules that are not themselves packages (e.g. ttx.py, or merge.py), or that are not at the top-level (e.g. fontTools.feaLib.builder, ttLib.woff2, possibly more)
2) it depends on setuptools to `find_packages` which adds a non-desirable runtime dependency on setuptools which I would like to avoid. We should find another way to discover sub-packages. I am sure there is someting in the stdlib for that.
Feel free to reopen the PR, we can continue discussing in there. I like the idea of having a main --help that summarizes the available tools under `fonttools` CLI. Thanks!

simoncozens:
OK. I wonder if we actually want to pursue the idea of autodiscovering commands because, for example, `fontTools.cu2qu.cu2qu` is a runnable module, but it's just testing/benchmarking and we don't necessarily want to expose it to the user. We certainly don't want to expose its module docstring to the user (it will have a docstring very soon...). Come to think of it, this is why I first started with the idea of a specific "description" variable to identify subcommands, rather than using the module docstring.

What do you think of the idea of searching all packages for a `cli_description` or similar variable, and exposing them if it is found? Any other ideas welcome.

anthrotype:
let's get the list of the modules that we want to expose first.
adding a cli_description could work, but I am thinking that maybe hard-coding those descriptions in the help.py would not be so bad.

simoncozens:
Well, the question was (a) how do we know what’s runnable? (Some modules only run if they are called as __main__. Checking whether there’s a “main” function is fragile and I don’t think it covers all cases) and also (b) even if we somehow know which is runnable and which isn’t, how do we distinguish between a runnable module that’s for internal use (such as the cu2qu benchmarks) and one which we want to expose for end users (eg cu2qu itself)? The presence of a specific variable holding a command-line description could be a useful indicator that this file is something to expose to end users. And keeping the description near the code which implements it, not hidden away in a separate help file, is better style anyway.

I think I’ve convinced myself that a special variable export is the way to do this...

anthrotype:
Ok, you convinced me as well!

simoncozens:
-> #1920

PR #1920, opened 2020-05-05
[doc] Add help options to fonttools CLI

simoncozens:
Take 2: Use special import variable to denote CLI-exposed command.

### DIFF
--- a/Lib/fontTools/help.py
+++ b/Lib/fontTools/help.py
@@ -1,0 +1,49 @@
+import pkgutil
+import sys
+import fontTools
+import importlib
+import os
+from pathlib import Path
+
+cli_description = "Show this help"
+
+
+def _discover_path_importables(pkg_pth, pkg_name):
+    """Yield all importables under a given path and package."""
+    for dir_path, _d, file_names in os.walk(pkg_pth):
+        pkg_dir_path = Path(dir_path)
+
+        if pkg_dir_path.parts[-1] == "__pycache__":
+            continue
+
+        if all(Path(_).suffix != ".py" for _ in file_names):
+            continue
+
+        rel_pt = pkg_dir_path.relative_to(pkg_pth)
+        pkg_pref = ".".join((pkg_name,) + rel_pt.parts)
+        yield from (
+            pkg_path
+            for _, pkg_path, _ in pkgutil.walk_packages(
+                (str(pkg_dir_path),), prefix=f"{pkg_pref}.",
+            )
+        )
+
+
+def show_help_list():
+    path = fontTools.__path__
+    for pkg in sorted(
+        set(_discover_path_importables(fontTools.__path__[0], "fontTools"))
+    ):
+        try:
+            description = __import__(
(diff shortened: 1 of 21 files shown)


anthrotype:
I don't know if I like very much these `cli_description` global variables. They feel out of place a bit.
I have another idea. How about we add these one-line cli descriptions to the docstring of the main() function? The presence of a main() function is what make the module runnable (ok, the name `main()` is a convention but we stick to that); then, the presence of a docstring inside main() is a marker that distinguishes _any_ runnable modules from those that we want to publicly advertise in the fonttools help CLI.

simoncozens:
Yeah, that's clearly the right approach. Nice. Will work on it.

Lib/fontTools/help.py
```diff
@@ -0,0 +1,51 @@
+import pkgutil
+import sys
+import fontTools
+import importlib
+import os
+from pathlib import Path
+
+def _discover_path_importables(pkg_pth, pkg_name):
+    """Yield all importables under a given path and package."""
```
anthrotype, line 9: I don't think you need this `_discover_path_importables ` function, you can just call `pkgutil.walk_packages` directly (it works recursively)

```python
sorted(mod.name for mod in pkgutil.walk_packages([fontTools.__path__[0]], prefix="fontTools."))
```

Lib/fontTools/help.py
```diff
@@ -0,0 +1,51 @@
...
+    descriptions = {}
+    for pkg in sorted(
+        set(_discover_path_importables(fontTools.__path__[0], "fontTools"))
+    ):
+        try:
+            description = __import__(
+                pkg, globals(), locals(), ["main"]
+            ).main.__doc__
+            if description:
+                pkg = pkg[10:].replace(".__main__", "")
```
anthrotype, line 41: what's 10? is it `len("fontTools.")`? I'd just remove it, you know it's not going to affect the result of this specific replace

Lib/fontTools/help.py
```diff
@@ -0,0 +1,51 @@
...
+                descriptions[pkg] = description
+        except Exception as e:
```
anthrotype, line 43: this exception handling too broad and all-catching, risks hiding our own bugs. I would only catch ImportError and AttributeError, and even these two I would catch them separately, first trying to import the module, then trying to get is main.__doc__

Lib/fontTools/help.py
simoncozens: It will affect the result. The package name is fully qualified (`fontTools.feaLib.__main__`) and we want to turn it into a verb `feaLib` for display in the list e.g.

```
fonttools feaLib Add features from a feature file (.fea) into a OTF font
```

We could change the first "." to a space but that's more obscure. :-)

anthrotype: ok I see, thanks for the clarification. I'd still replace that magic constant with `len("fontTools.")` for the sake of readability

PR #1921, opened 2020-05-06
[ttLib.name] Add nameTable.findMultilingualName() method

justvanrossum:
It is usually a good idea to reuse name table records for the non-core nameIDs, such as those for `fvar` and `STAT` names. Finding an existing name record matching a single string is not very hard, but it becomes a mess real quick for multilingual names.

The added `nameTable.findMultilingualName(names)` method makes this easier:

```python
names = dict(en="Width", nl="Breedte")
nameID = nameTable.findMultilingualName(names)
if nameID is None:
    nameID = nameTable.addMultilingualName(names)
```

I considered making this part of `addMultilingualName()` itself, but thought that perhaps it's better to keep this separate.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_n_a_m_e.py
+++ b/Lib/fontTools/ttLib/tables/_n_a_m_e.py
@@ -250,7 +250,8 @@ class table__n_a_m_e(DefaultTable.DefaultTable):
 		names that otherwise cannot get encoded at all.
 
 		'nameID' is the name ID to be used, or None to let the library
-		pick an unused name ID.
+		find an existing set of name records that match, or pick an
+		unused name ID.
 
 		If 'windows' is True, a platformID=3 name record will be added.
 		If 'mac' is True, a platformID=1 name record will be added.
--- a/Lib/fontTools/ttLib/tables/_n_a_m_e.py
+++ b/Lib/fontTools/ttLib/tables/_n_a_m_e.py
@@ -184,6 +184,57 @@ class table__n_a_m_e(DefaultTable.DefaultTable):
 			raise ValueError("nameID must be less than 32768")
 		return nameID
 
+	def findMultilingualName(self, names, windows=True, mac=True):
+		"""Return the name ID of an existing multilingual name matching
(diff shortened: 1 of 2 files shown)


justvanrossum:
If people agree that nameID sharing should always be done (for non-standard name IDs), then perhaps it would be better to make `addMultilingualName()` call `findMultilingualName()` by default, if the passed nameID is None.

This would turn on name ID sharing for all client code using `addMultilingualName()` without clients having to do anything.

khaledhosny:
I wouldn’t make it the default since the strings are will be shared internally in the `name` table anyway, so this only shares the name id’s and saves the few bytes of the name records. So the gain is not much compared to any potential issues.

justvanrossum:
I realize the gain at the binary level is small, yet I still feel it's in general a good idea to keep the number of name records small.

It's currently pretty silly how `fvar` and `STAT` have their own distinct name table entries, even if it's _likely_ they will contain some identical strings.

I'm curious what kind of potential issues could arise.

If we agree it should not be done be default, I can imagine adding a flag to `addMultilingualName()` that enables this behavior, without the need to explicitly call `findMultilingualName()`.

khaledhosny:
You are probably right, may be I’m little to worried.

anthrotype:
> If people agree that nameID sharing should always be done (for non-standard name IDs), then perhaps it would be better to make addMultilingualName() call findMultilingualName() by default, if the passed nameID is None.

I like that. I don't think it there is any use for having distinct nameIDs for identical strings+ encoding triplet in the user-defined range of nameIDs. Let's keep the distinct find method, but have add method use that when user doesn't request a specific nameID

justvanrossum:
> I like that. I don't think it there is any use for having distinct nameIDs for identical strings+ encoding triplet in the user-defined range of nameIDs. Let's keep the distinct find method, but have add method use that when user doesn't request a specific nameID

Ok, I just implemented that. Surprisingly, I didn't have to fix any expected results anywhere...

justvanrossum:
@anthrotype is this ok to merge, with the latest changes?

PR #1922, opened 2020-05-06
[WIP] [otlLib.builder] Add function to build STAT table from high-level description

justvanrossum:
Proposal to add a function named `buildStatTable()` that conveniently builds a `STAT` table from a high-level data structure.

Question:
- is `otlLib.builder` the best place for this? I considered `varLib.builder` but `STAT` is not limited to variable fonts

TODO:
- [ ] Add support for `AxisValue` format 4
- [x] Document the input data structure
- [ ] Rebase and fix, once #1921 is accepted


benkiel:
Have you looked at Dalton Maags statmake?

justvanrossum:
> Have you looked at Dalton Maags statmake?

No, I did not know that existed... Will have a look, thanks!

anthrotype:
> Have you looked at Dalton Maags statmake?

Yes @madig wrote that (that's why I pinged him among the reviewers). Ideally statmake (even ufo2ft) would become a client of this new API.

@khaledhosny has been working on adding support for STAT in afdko's makeotf https://github.com/adobe-type-tools/afdko/pull/1127 so he may also have some insights

anthrotype:
I guess that once Adobe feature file syntax support defining STAT (https://github.com/adobe-type-tools/afdko/issues/176), feaLib should follow suit and maybe we won't need any statmake any more.

justvanrossum:
I had a very quick look at statmake:
- I find the building of the individual low level tables more complicated than needed, and not fitting with the style of fonttools
- It contains even higher level functionality like deducing STAT info from `fvar`, something that I deliberately wanted to leave out of this, as I can see there may be different approaches for that

So, while there's some overlap, it's not really solving the same thing. Indeed, statmake could become a client of this API.

justvanrossum:
One thing I'm not 100% sure about is the choice in my data structure to have values be part of an axis definition. That makes sense for AxisValue formats 1, 2 and 3, but less so for format 4.

For 1, 2 and 3 it's a fairly obvious win as you don't need to mention the axis tag for each value:

```python
    axisData = [
        dict(
            tag="wght",
            name="Weight",
            ordering=0,  # optional
            values=[
                dict(value=100, name='Thin'),
                dict(value=300, name='Light'),
                dict(value=400, name='Regular', flags=0x2),
                dict(value=900, name='Black'),
            ],
        )
    ]
```

madig:
statmake is indeed intended to be a higher-level and opinionated tool to maintain STAT information and I'd love to get rid of the bare table generation and filling-in.

madig:
> I guess that once Adobe feature file syntax support defining STAT ([adobe-type-tools/afdko#176](https://github.com/adobe-type-tools/afdko/issues/176)), feaLib should follow suit and maybe we won't need any statmake any more.

madig:
(Blargh, sorry, misclicked)

> I guess that once Adobe feature file syntax support defining STAT ([adobe-type-tools/afdko#176](https://github.com/adobe-type-tools/afdko/issues/176)), feaLib should follow suit and maybe we won't need any statmake any more.

Different discussion, but there's some leg work to be done there. How do you write a single STAT definition and have it tailored to your different VF subsets? How do you include it in your VF feature file but not in your instances?

madig:
> One thing I'm not 100% sure about is the choice in my data structure to have values be part of an axis definition

You can make a format 4 definition, that's what I did in statmake: https://github.com/daltonmaag/statmake/blob/master/tests/data/Test.stylespace#L157-L201

justvanrossum:
> You can make a "locations" definition

Ah, to keep format 4 completely separate. I'll consider that, as an additional arg to the build function.

(I'm pessimistic about .fea offering anything helpful except in the most trivial of cases.)

Lorp:
I find it helpful to think of Format 1 as a special case of Format 4 (i.e. an array of axes containing just one item), and the implementation in [Samsa](https://github.com/Lorp/samsa) partly reflects that. Every STAT value specifies location on a _subset_ of axes in the designspace of the type family. It probably makes sense in a UI to group values by these subsets. In a high-level description I would also recommend grouping by subset, remembering that _if the subset is a single axis_ then you have additional options of range (Format 2) and linked value (Format 3). I considered presenting all values in a consistent manner, so each subset (in any format) with its own dropdown, but in the end only gave the non-Format 4 values their own dropdown, and dumped all Format 4 values in the “multi-axis” dropdown.

BTW Samsa’s heuristic for inserting Format 4 names during composition is to prepend them to the string in the order they appear in the STAT table. This is not ideal, and it might make sense to derive orderings from the ordering of the axes in the subset, as in a spreadsheet function to sort by multiple columns.

justvanrossum:
Thanks Laurence, that is an interesting view (format 1 as a special case of 4).

I now see two directions I could take:

1. don't specify axis values under axes at all, but make it a separate list
   - this is @lorp's view: see format 1 as a special case of 4
2. list axis value formats 1, 2 and 3 under axis definitions, have a separate list for axis value format 4
   - this is @madig's view, and makes format 4 a special case

Option 1 has the advantage that it most closely parallels the actual STAT data structure. But it's also lower level, requiring some redundancies for the (presumably common) non-format 4 cases (specifying the axis tag for every value).

Option 2 separates out format 4 as something exceptional. Which is a valid view, too, in my interpretation of https://twitter.com/pgconstable/status/1258169386075971590?s=20. Format 4 requiring a table version bump also supports this view.

I'll think about this more.

justvanrossum:
Another advantage of option 1: it gives the user full control over the axis value order.

But given that that order has no significance (https://twitter.com/pgconstable/status/1258168277806669826?s=20) I'm not sure if this needs to a be a requirement at all.

madig:
There might be _certain software_ that relies on ordering even when it shouldn't. Like how _some other software_ crashed when the axis record ordering did not match the fvar ordering.

justvanrossum:
> There might be certain software that relies on ordering even when it shouldn't. Like how some other software crashed when the axis record ordering did not match the fvar ordering.

Is the axis value order 100% user-controllable with `statmake`?

madig:
That particular crash was about the "ordering" value that is settable in your code, so if your axes are wght, wdth, slnt but the ordering value is 1, 2, 0 respectively. Boom!

I don't even remember anymore, I think the order is taken from the order in the Stylespace file, but I have never tested if I can actually influence the order.

justvanrossum:
> That particular crash was about the "ordering" value that is settable in your code, so if your axes are wght, wdth, slnt but the ordering value is 1, 2, 0 respectively. Boom!

That part of the spec left me flabbergasted: why say the axis order is arbitrary and then introduce a field that specifies an order that is not necessarily the array order?

justvanrossum:
Oh dear, of course I messed up the rebase :( Does this need fixing and if so, how?

anthrotype:
git fetch origin && git rebase origin/master

anthrotype:
you may need to git push -f after the rebase

justvanrossum:
No, this leaves me in a bigger mess. No, origin is my fork, upstream is fonttools/fonttools.

anthrotype:
well this branch to be rebased on the upstream master branch somehow. Alternatively you could have done a `git merge upstream/master` instead of a rebase, but it's probably too late for that.
If you can't fix it, you can always start a fresh new branch from upstream master, then cherry-pick the commits you need..

justvanrossum:
Yeah, I'm ready to give up and make a new PR.

anthrotype:
yeah, no worries

justvanrossum:
Continuing here: #1926

Lib/fontTools/otlLib/builder.py
```diff
@@ -657,3 +658,146 @@ def build(self):
...
+
+    If the value dict contains a 'nominalValue' item, an AxisValue
+    record Format 2 is built. Optionally it may contain 'rangeMinValue'
+    and 'rangeMaxValue' items. These map to -Infinity and +Infinity
+    respectively if omitted.
+
+    If the value dict contains an 'axisValues' item, an AxisValue
+    record Format 4 is built. It should be a list of value dicts,
+    that each contain two items: 'tag' for the contributing axis, and
+    'value' for the associated value.
```
anthrotype, line 723: it may be worth to also add an example here for the axisValues list (format 4)

anthrotype: oh sorry I was reviewing the wrong PR

PR #1923, opened 2020-05-06
Fix a bunch of lgtm errors

simoncozens:


Lib/fontTools/cu2qu/errors.py
```diff
@@ -19,7 +19,7 @@ class Error(Exception):
 class ApproxNotFoundError(Error):
     def __init__(self, curve):
         message = "no approximation found: %s" % curve
-        super(Error, self).__init__(message)
+        super(ApproxNotFoundError, self).__init__(message)
```
anthrotype, line 22: or just this, since we're py3 only now
```suggestion
        super().__init__(message)
```

Lib/fontTools/misc/loggingTools.py
```diff
@@ -406,16 +406,16 @@ class ChannelsFilter(logging.Filter):
 	>>> logging.getLogger('C.DE').debug('neither this one!')
 	"""
 
-	def __init__(self, *names):
+	def __init__(self, *names): # lgtm[py/missing-call-to-init]
```
anthrotype, line 409: i don't know if I want to add linter-specific comments like this.

Lib/fontTools/misc/plistlib.py
```diff
@@ -397,8 +397,6 @@ def totree(
 ):
     if use_builtin_types is None:
         use_builtin_types = USE_BUILTIN_TYPES
-    else:
-        use_builtin_types = use_builtin_types
```
anthrotype, line 401: nice find

Lib/fontTools/ttLib/sfnt.py
```diff
@@ -553,8 +553,7 @@ def __init__(self, reader=None):
 				reader.file.seek(reader.metaOffset)
 				rawData = reader.file.read(reader.metaLength)
 				assert len(rawData) == reader.metaLength
-				import zlib
-				data = zlib.decompress(rawData)
```
anthrotype, line 557: what's the problem here?

Lib/fontTools/ttLib/woff2.py
```diff
@@ -1168,26 +1168,8 @@ def __init__(self, reader=None, data=None, transformedTables=None):
...
 					"'glyf' and 'loca' must be transformed (or not) together"
 				)
-
-		self.majorVersion = None
-		self.minorVersion = None
-		self.metaData = None
-		self.privData = None
+		super(WOFF2FlavorData, self).__init__(reader=reader)
 		if reader:
-			self.majorVersion = reader.majorVersion
```
anthrotype, line 1177: I don't think removing this whole block is the correct thing to do here

anthrotype:
I actually didn't mean to click "approve" on this PR, just "comment" (I don't know how to take the approval away)

anthrotype:
In general, I prefer more tailored PRs that address specific issues over generic "fix a bunch of errors" that touch many different files

Lib/fontTools/ttLib/woff2.py
simoncozens: I think you're missing the dependency between the two things. WOFF2FlavorData has almost the exact same initializer (with some additions) as its superclass, except for the fact that the decompression method is different. If you split out the decompression part to a method (which gets overridden by WOFF2FlavorData), then you can remove a lot of repeated code by calling super.__init__.

simoncozens:
Noted, I'll split this.

PR #1926, opened 2020-05-07
[otlLib.builder] Add function to build STAT table from high-level description

justvanrossum:
Continuation of #1922 after a rebase debacle. Still WIP.


anthrotype:
FYI there is a way to mark a PR as "Draft", see the link under the Reviewers section at the top right (it says "Still in progress? Convert to draft")

justvanrossum:
For now I went for the additional "locations" argument, loosely following @madig's model. I'm not 100% convinced yet, but it's definitely better than what I had before.

Recapping:

- This treats AxisValue Format 4 as a separate kind of axis values, to be specified in the `locations` argument of `buildStatTable()`. Regular AxisValues (format 1, 2 and 3) are specified as subitems of axis dicts.

- The alternative would be to treat _all_ AxisValue items as potentially multi-axis (making format 1 a special case of format 4, as per @Lorp's view). This would mean we need to supply _all_ axis values separate from the axis dicts, leading to some redundancies for the common case, in which a location uses only one axis.

I'm still open for the latter route. While it is a little lower level for the common case, it would be relatively easy to translate the higher level structure to the lower one.

Lib/fontTools/otlLib/builder.py
```diff
@@ -657,3 +658,177 @@ def build(self):
...
+
+
+def _addName(nameTable, value):
+    if isinstance(value, int):
+        # Already a nameID
+        return value
+    if isinstance(value, str):
+        names = dict(en=value)
+    else:
+        assert isinstance(value, dict)
```
anthrotype, line 831: assertions are optimized away, this may better be a TypeError exception

Lib/fontTools/otlLib/builder.py
```diff
@@ -657,3 +658,177 @@ def build(self):
...
+    ttFont["STAT"] = ttLib.newTable("STAT")
+    statTable = ttFont["STAT"].table = ot.STAT()
+    if not locations:
+        statTable.Version = 0x00010001
+        locations = ()
+    else:
+        # We'll be adding Format 4 AxisValue records, which
+        # requires a higher table version
+        statTable.Version = 0x00010002
+    nameTable = ttFont["name"]
```
anthrotype, line 754: since we're documenting things, it may be a good idea to also mention in the docstring that the ttFont must already have a name table (even though it seems logical)

justvanrossum:
I think this is ready to merge. I've changed the varLib STAT building code with a call to the new function, which was a nice exercise in code deletion.

davelab6:
Does this make statmake obsolete?

justvanrossum:
> Does this make statmake obsolete?

No: statmake is higher level and has ways to come up with the contents of STAT. In @madig's words (from https://github.com/fonttools/fonttools/pull/1922#issuecomment-624860807):

> statmake is indeed intended to be a higher-level and opinionated tool to maintain STAT information (because you can't deduce it automatically from the Designspace in all cases) and I'd love to get rid of the bare table generation and filling-in.

This API is only concerned with the compiling user-supplied content to STAT table data ready to compile to binary. statmake could become a client of this API.

davelab6:
Thanks for explaining!

PR #1929, opened 2020-05-07
Use inheritance to avoid repeated code in WOFF2FlavorData initializer

simoncozens:


anthrotype:
Ah I see, thanks for catching this

PR #1932, opened 2020-05-07
[misc.psCharStrings] Use inheritance to avoid repeated code in initializer

simoncozens:


anthrotype:
hm.. although T2CharString initializer also sets other attributes that are not used by T1CharString. It's probably benign. But I'm tempted to revert for the sake of least surprises.

simoncozens:
T2CharString has the global_subrs attribute, and we have declared T1CharString to be a subclass of T2CharString. This commit doesn't change that reality...

justvanrossum:
I'm not necessarily against this change, but:

> T2CharString has the global_subrs attribute, and we have declared T1CharString to be a subclass of T2CharString. This commit doesn't change that reality...

That's only a "theoretical reality", given how inheritance *should* work. In this practical reality `T1CharString` instances never actually *got* these attributes, at least not before this change. It's definitely not a clean subclass of `T2CharString`, and perhaps it's worthwhile to clean this up at some point.

simoncozens:
Yeah. I'd kind of question that class hierarchy anyway. Saying "a type 1 charstring is a kind of a type 2 charstring" just feels weird. Make them both inherit from an abstract charstring class.

PR #1933, opened 2020-05-07
[docs] Document afmLib

simoncozens:


Lib/fontTools/afmLib.py
```diff
@@ -1,8 +1,50 @@
...
+	>>> from fontTools.afmLib import AFM
+	>>> f = AFM("Tests/afmLib/data/TestAFM.afm")
+	>>>
+	>>> # Accessing a pair gets you the kern value
+	>>> f[("V","A")]
+	-60
+	>>>
+	>>> # Accessing a glyph name gets you metrics
+	>>> f["A"]
+	(65, 668, (8, -25, 660, 666))
```
anthrotype, line 19: would be nice if this were a NamedTuple

justvanrossum: I don't disagree, but this should be very low priority, given the legacy status of AFM in general.

PR #1934, opened 2020-05-07
[docs] Document agl.py

simoncozens:


Lib/fontTools/agl.py
```diff
@@ -3,6 +3,28 @@
...
+Interface to the Adobe Glyph List
+
+This module exists to convert glyph names from the Adobe Glyph List
+to their Unicode equivalents. Example usage:
+
+	>>> from fontTools.agl import toUnicode
+	>>> toUnicode("nahiragana")
+	'な'
+
+It also contains two dictionaries, ``UV2AGL`` and ``AGL2UV`` which maps from
```
dscorbett, line 16: There should be a comma after <code>&#x60;&#x60;AGL2UV&#x60;&#x60;</code> and “maps” should be “map”.

PR #1935, opened 2020-05-08
[docs] Source documentation for cffLib

simoncozens:


anthrotype:
It would be nice if somebody from Adobe can take a look at this (cc @cjchapman )

cjchapman:
@simoncozens Thanks for doing this!

@anthrotype I'd like for @josh-hadley & @punchcutter to review it.

Lib/fontTools/cffLib/specializer.py
```diff
@@ -1,6 +1,17 @@
 # -*- coding: utf-8 -*-
 
-"""T2CharString operator specializer and generalizer."""
+"""T2CharString operator specializer and generalizer.
+
+PostScript glyph drawing operations can be expressed in multiple different
```
josh-hadley, line 5: How about something like, "The Type2 Charstring Format defines several operators which combine multiple drawing commands into one, typically with a smaller representation."

Also it might be nice to have a brief explanation of why/how specialization and generalization are used.

Lib/fontTools/cffLib/__init__.py
```diff
@@ -608,16 +684,46 @@ def produceItem(self, index, data, file, offset):
...
 class GlobalSubrsIndex(Index):
+	"""This index contains all the global subroutines in the font. A global
+	subroutine is a set of ``CharString`` data which is accessible to any
+	glyph in the font, and are used to store repeated instructions - for
+	example, components may be encoded as global subroutines, but so could
+	hinting instructions.
+
+	Remember that when interpreting a ``callgsubr`` instruction (or indeed
+	a ``callsubr`` instruction) that you will need to add the "subroutine
+	number bias" to number given:
```
josh-hadley, line 707: Maybe mention that the bias # depends on the number of subrs/gsubrs defined (it's not always 107).

Lib/fontTools/cffLib/__init__.py
```diff
@@ -1,4 +1,15 @@
-"""cffLib.py -- read/write tools for Adobe CFF fonts."""
+"""cffLib: read/write Adobe CFF fonts
```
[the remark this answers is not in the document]
simoncozens: I'm working my way through - they'll all have it soon.

Lib/fontTools/cffLib/__init__.py
```diff
@@ -28,8 +39,37 @@
 
 
 class CFFFontSet(object):
+	"""A CFF font "file" can contain more than one font, although this is
+	extremely rare (and fontTools can't handle such files anyway...).
```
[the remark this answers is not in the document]
khaledhosny: It is been a while since I looked into this, but I think FonTools can handle “bare” CFF files with multiple fonts, but probably not in the CFF table since this is not allowed by the [spec](https://docs.microsoft.com/en-us/typography/opentype/spec/cff)

Lib/fontTools/cffLib/specializer.py
simoncozens: I'm not sure this is what this module does - I might be wrong about this, but as I understand it, it doesn't really combine multiple drawing commands into one, but substitutes e.g. four-argument `rlineto` with three-argument `hlineto`.

josh-hadley: Ah, right, I remembered multiple operator merging being discussed, but that was never implemented in specializer. Carry on :-)

PR #1937, opened 2020-05-08
[docs] Document cu2qu library

simoncozens:
* Reorganise the documentation so that everything is in one place and users are more clearly pointed to the modules which are likely to be useful for their purposes. (I still think it’s worth having at least a brief reference to ``cu2qu.cli`` in there, as a way of reminding users that there is a command-line implementation.)
* Explain a little bit about the methodology.
* Add docstrings for all the methods I understand.

### DIFF
--- a/Lib/fontTools/cu2qu/cu2qu.py
+++ b/Lib/fontTools/cu2qu/cu2qu.py
@@ -46,7 +46,15 @@ else:
 @cython.returns(cython.double)
 @cython.locals(v1=cython.complex, v2=cython.complex)
 def dot(v1, v2):
-    """Return the dot product of two vectors."""
+    """Return the dot product of two vectors.
+
+    Args:
+        v1 (complex): First vector.
+        v2 (complex): Second vector.
+
+    Returns:
+        double: Dot product.
+    """
     return (v1 * v2.conjugate()).real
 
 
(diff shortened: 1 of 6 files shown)


Lib/fontTools/cu2qu/cu2qu.py
```diff
@@ -77,6 +77,9 @@ def calc_cubic_parameters(p0, p1, p2, p3):
 @cython.cfunc
 @cython.locals(p0=cython.complex, p1=cython.complex, p2=cython.complex, p3=cython.complex)
 def split_cubic_into_n_iter(p0, p1, p2, p3, n):
+    """Given four vectors representing control points of a cubic curve,
```
anthrotype, line 80: are bezier control "points" really "vectors"? I'd call them just "points" (aka locations in bidimensional space relative to origin). "vector" implies a direction and magnitude, a displacement, a subtraction between two points, an arrow..
At least in computational geometry.. Or do you mean "vector" here as an "array" of coordinates?

justvanrossum: +1 for "points".

simoncozens: Yeah, I meant they're not scalars. I think I would prefer to call them "complex numbers" because that clarifies input expectations and we already say that they're "points" later in the sentence.

anthrotype: if we go with #1947, then in the first line (the compact description) you'd call them points and in the Args section you'd annotate them with their complex type.

PR #1938, opened 2020-05-08
[ttLib.table._m_e_t_a] if data happens to be ascii, emit comment in TTX

justvanrossum:
I work on a project where an ascii string is stored in a private `meta` key. For us it would be practical if we could see the ascii contents as a comment string in the TTX output.

This PR checks whether the data could be interpreted as ascii (and does not contain special characters < `0x20`), and emits a comment with the contents decoded from ascii.

Maybe this is a bit too specific, so I won't be heartbroken if this gets rejected.


anthrotype:
could we decode as UTF-8 (superset of ASCII)?

justvanrossum:
We could, but there would be a slightly higher chance of a nonsense comment, if that data just _happens_ to be valid utf-8.

PR #1939, opened 2020-05-08
[docs] Reorganise designspaceLib documentation

simoncozens:
designspaceLib actually has really thorough documentation already! Because of this, I’ve only done light-touch editing (mostly formatting nits) where I’ve seen obvious problems.

The bigger issue with it is that it’s a bit disorganised, and the references to MutatorMath aren’t necessarily helpful. I’ve tried to improve the organising by providing signposting to end-users to portions of the documentation that can help them achieve what they’re trying to do, reorganising it into one page with links so they don’t have to hunt around for things.

### DIFF
--- a/Doc/source/designspaceLib/index.rst
+++ b/Doc/source/designspaceLib/index.rst
@@ -1,16 +1,24 @@
-##############
-designspaceLib
-##############
+#######################################################
+designspaceLib: Read, write, and edit designspace files
+#######################################################
 
-MutatorMath started out with its own reader and writer for designspaces.
-Since then the use of designspace has broadened and it would be useful
-to have a reader and writer that are independent of a specific system.
+Implements support for reading and manipulating ``designspace`` files.
+Allows the users to define sources, axes, rules and instances.
 
-.. toctree::
-   :maxdepth: 1
+This documentation covers:
 
-   readme
-   scripting
+-  `Reading a designspace file in Python <#python-api>`_
+-  `The XML structure of a designspace file <#document-xml-structure>`_
+-  `How to create a designspace file in code <#designspacedocument-scripting>`_
+-  `API documentation  <#designspacelib-api>`_
+
+.. include:: readme.rst
+.. include:: scripting.rst
+
+fontTools.designspaceLib
+########################
+
+.. _designspacelib-api:
 
 .. automodule:: fontTools.designspaceLib
    :inherited-members:
--- a/Doc/source/designspaceLib/readme.rst
+++ b/Doc/source/designspaceLib/readme.rst
@@ -1,23 +1,13 @@
-#################################
-DesignSpaceDocument Specification
-#################################
+.. _python-api:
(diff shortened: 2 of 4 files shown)


anthrotype:
I'd also say that MutatorMath nowadays is no longer being used in the same way it was when these docs where originally written by @LettError . I believe its successor is meant to be [ufoProcessor](https://github.com/LettError/ufoProcessor) -- which uses fontTools.designspaceLib to parse designspace files, and only uses the "math" part of of MutatorMath as an alternative to varLib for blending masters into instances.
There's even another successor to MutatorMath in fontamake.instantiator, which similarly uses designspaceLib for parsing the documents and varLib for interpolation (fontmake itself no longer uses MutatorMath).

We should see whether those references to MutatorMath still make sense or wheter they should be updated or removed.

simoncozens:
Yeah, hang on, this documentation is weird. e.g. in the description of attributes for `SourceDescriptor` it says:

-  ``mutedGlyphNames``: list. Glyphnames that need to be muted in the
   instances. MutatorMath.
-  ``familyName``: string. Family name of this source. Though this data
   can be extracted from the font, it can be efficient to have it right
   here. Varlib.

I don't know what is meant by adding `MutatorMath` and `Varlib` to those attributes, but designspaceLib seems to support both.

Similarly in the description of `<features copy="1" />` it says this is "MutatorMath only". Again, designspaceLib supports the `copyFeatures` attribute.

Is this saying anything useful, (like, does *designspaceLib* support it but *not* varLib?) or can it all be removed?

LettError:
Feel free to remove it from the documentation. Please don't remove it from the code as designspacelib still needs to support some functionality in places outside of fonttools and independent from variable fonts.

davelab6:
If there's code in designspacelib to support some functionality, that kind
of usage should be documented :)

benkiel:
@simoncozens The tags of `varlib` and `mutatormath` let the user know which bit can be used by which math doing interpolations. It's really useful, even if the `mutatormath` bits are not used by fonttools or fontmake when making variable fonts. See the Recusive project where I use them to store all the PS names for the generated static instances. Please don't delete this documentation.

simoncozens:
OK, well, then we need to make that clear. The point is for this not just to be a reminder for people who already understand how it works but to be a guide for those who don't.

anthrotype:
Originally designspace document was designed for generating static UFO instances only, then it was adopted for specifying variable fonts. designspaceLib reads all those attributes, not all may be used when making static instances vs variable fonts. 
Maybe instead of juxtaposing mutatormath vs varLib (and ufoProcessor, fontmake, etc.), it might be useful to make explicit what attributes are intended to be used for producing statics vs variables. Then it's up to specific implementations to say what are actually used or supported.

belluzj:
I included the relevant changes from here into #2436

khaledhosny:
Closing per the last comment.

PR #1940, opened 2020-05-08
[docs] encodings library

simoncozens:
The MacRoman and StandardEncoding parts aren’t really used anywhere (and certainly aren’t user-facing as far as I know), but the ``encodings.codecs`` part might be useful to some people, so explain that and bring it front and center.


benkiel:
Other tools use MacRoman

PR #1941, opened 2020-05-08
[docs] Document feaLib

simoncozens:
Quite a big change, but it’s a big library:

* Rearrange docs by user intention, highlighting the things you can do with each component.
* Remove reference to lexer and error modules from documentation tree, since they’re not user-facing.
* I’ve added docstrings to the parser even though we only provide access to the user-facing part of the API in the main documentation, just to clarify what some of the more obscure methods do and provide links to the spec.
* AST *is* user-facing if you’re building your own feature files in code, so all classes are documented with the user in mind.

### DIFF
--- a/Doc/source/feaLib/index.rst
+++ b/Doc/source/feaLib/index.rst
@@ -1,17 +1,40 @@
-######
-feaLib
-######
+#########################################
+feaLib: Read/write OpenType feature files
+#########################################
 
-.. toctree::
-   :maxdepth: 1
+fontTools' ``feaLib`` allows for the creation and parsing of Adobe
+Font Development Kit for OpenType feature (``.fea``) files. The syntax
+of these files is described `here <http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html>`_.
 
-   ast
-   builder
-   error
-   lexer
-   parser
+The :class:`fontTools.feaLib.parser.Parser` class can be used to parse files
+into an abstract syntax tree, and from there the
+:class:`fontTools.feaLib.builder.Builder` class can add features to an existing
+font file. You can inspect the parsed syntax tree, walk the tree and do clever
+things with it, and also generate your own feature files programmatically, by
+using the classes in the :mod:`fontTools.feaLib.ast` module.
 
-.. automodule:: fontTools.feaLib
-   :inherited-members:
+Parsing
+-------
+
+.. autoclass:: fontTools.feaLib.parser.Parser
+   :members: parse
+   :member-order: bysource
+
+Building
+---------
+
+.. automodule:: fontTools.feaLib.builder
+   :members: addOpenTypeFeatures, addOpenTypeFeaturesFromString
+
+Generation/Interrogation
+------------------------
+
+.. _`glyph-containing object`:
+.. _`glyph-containing objects`:
+
+In the below, a **glyph-containing object** is an object of one of the following
+classes: :class:`GlyphName`, :class:`GlyphClass`, :class:`GlyphClassName`.
+
+.. automodule:: fontTools.feaLib.ast
+   :member-order: bysource
    :members:
-   :undoc-members:
--- a/Doc/source/feaLib/index.rst
+++ b/Doc/source/feaLib/index.rst
@@ -4,7 +4,7 @@ feaLib: Read/write OpenType feature files
 
 fontTools' ``feaLib`` allows for the creation and parsing of Adobe
 Font Development Kit for OpenType feature (``.fea``) files. The syntax
-of these files is described `here <http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html>`_.
+of these files is described `here <https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html>`_.
 
 The :class:`fontTools.feaLib.parser.Parser` class can be used to parse files
 into an abstract syntax tree, and from there the
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -593,7 +593,7 @@ class AnchorDefinition(Statement):
 
 
 class AttachStatement(Statement):
-    """A GDEF table ``Attach`` statement."""
+    """A ``GDEF`` table ``Attach`` statement."""
     def __init__(self, glyphs, contourPoints, location=None):
         Statement.__init__(self, location)
         self.glyphs = glyphs #: A `glyph-containing object`_
@@ -872,7 +872,7 @@ class FontRevisionStatement(Statement):
 
 
 class LigatureCaretByIndexStatement(Statement):
-    """A GDEF table ``LigatureCaretByIndex`` statement. ``glyphs`` should be
+    """A ``GDEF`` table ``LigatureCaretByIndex`` statement. ``glyphs`` should be
     a `glyph-containing object`_, and ``carets`` should be a list of integers."""
     def __init__(self, glyphs, carets, location=None):
         Statement.__init__(self, location)
@@ -889,7 +889,7 @@ class LigatureCaretByIndexStatement(Statement):
 
 
 class LigatureCaretByPosStatement(Statement):
-    """A GDEF table ``LigatureCaretByPos`` statement. ``glyphs`` should be
+    """A ``GDEF`` table ``LigatureCaretByPos`` statement. ``glyphs`` should be
     a `glyph-containing object`_, and ``carets`` should be a list of integers."""
     def __init__(self, glyphs, carets, location=None):
         Statement.__init__(self, location)
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -210,7 +210,7 @@ class GlyphClass(Expression):
         self.curr = len(self.glyphs)
 
     def add_class(self, gc):
-        """Adds glyphs from the given :class:`GlyphClassName` object to the
+        """Add glyphs from the given :class:`GlyphClassName` object to the
         class."""
         if self.curr < len(self.glyphs):
             self.original.extend(self.glyphs[self.curr:])
@@ -382,7 +382,7 @@ class TableBlock(Block):
 
 
 class GlyphClassDefinition(Statement):
-    """Example: ``@UPPERCASE = [A-Z]`;``."""
+    """Example: ``@UPPERCASE = [A-Z];``."""
     def __init__(self, name, glyphs, location=None):
         Statement.__init__(self, location)
         self.name = name #: class name as a string, without initial ``@``
@@ -609,7 +609,7 @@ class AttachStatement(Statement):
 class ChainContextPosStatement(Statement):
     """A chained contextual positioning statement.
 
-    ``prefix``, ``glyph``, and ``suffix`` should be lists of
+    ``prefix``, ``glyphs``, and ``suffix`` should be lists of
     `glyph-containing objects`_ .
 
     ``lookups`` should be a list of :class:`LookupBlock` statements, with
@@ -821,7 +821,7 @@ class LanguageStatement(Statement):
         Statement.__init__(self, location)
         assert(len(language) == 4)
         self.language = language #: A four-character language tag
-        self.include_default = include_default #: If false, "exclude_deflt"
+        self.include_default = include_default #: If false, "exclude_dflt"
         self.required = required
 
     def build(self, builder):
@@ -856,7 +856,7 @@ class LanguageSystemStatement(Statement):
 
 class FontRevisionStatement(Statement):
     """A ``head`` table ``FontRevision`` statement. ``revision`` should be a
-    decimal number with three significant decimal places."""
+    number, and will be formatted to three significant decimal places."""
     def __init__(self, revision, location=None):
         Statement.__init__(self, location)
         self.revision = revision
@@ -998,7 +998,7 @@ class LookupReferenceStatement(Statement):
 
 
 class MarkBasePosStatement(Statement):
-    """A mark-to-base positioning rule. The ``base`` must be a
+    """A mark-to-base positioning rule. The ``base`` should be a
     `glyph-containing object`_. The ``marks`` should be a list of
     (:class:`Anchor`, :class:`MarkClass`) tuples."""
     def __init__(self, base, marks, location=None):
@@ -1134,6 +1134,8 @@ class PairPosStatement(Statement):
     ``glyphs1`` and ``glyphs2`` should be `glyph-containing objects`_.
     ``valuerecord1`` should be a :class:`ValueRecord` object;
     ``valuerecord2`` should be either a :class:`ValueRecord` object or ``None``.
+    If ``enumerated`` is true, then this is expressed as an
+    `enumerated pair <http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#6.b.ii>`_.
     """
     def __init__(self, glyphs1, valuerecord1, glyphs2, valuerecord2,
                  enumerated=False, location=None):
@@ -1283,9 +1285,8 @@ class SinglePosStatement(Statement):
     """A single position statement. ``prefix`` and ``suffix`` should be
     lists of `glyph-containing objects`_.
 
-    ``pos`` should be a list of (`glyph-containing object`_,
-    :class:`ValueRecord`) tuples. There should probably be only one entry in
-    that list."""
+    ``pos`` should be a one-element list containing a (`glyph-containing object`_,
+    :class:`ValueRecord`) tuple."""
 
     def __init__(self, pos, prefix, suffix, forceChain, location=None):
         Statement.__init__(self, location)
@@ -1317,7 +1318,7 @@ class SinglePosStatement(Statement):
 
 
 class SubtableStatement(Statement):
-    """Represents a subtable break"""
+    """Represents a subtable break."""
     def __init__(self, location=None):
         Statement.__init__(self, location)
 
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -1135,7 +1135,7 @@ class PairPosStatement(Statement):
     ``valuerecord1`` should be a :class:`ValueRecord` object;
     ``valuerecord2`` should be either a :class:`ValueRecord` object or ``None``.
     If ``enumerated`` is true, then this is expressed as an
-    `enumerated pair <http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#6.b.ii>`_.
+    `enumerated pair <https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#6.b.ii>`_.
     """
     def __init__(self, glyphs1, valuerecord1, glyphs2, valuerecord2,
                  enumerated=False, location=None):
@@ -1434,7 +1434,7 @@ def simplify_name_attributes(pid, eid, lid):
 
 
 class NameRecord(Statement):
-    """Represents a name record. (`Section 9.e. <http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#9.e>`_)"""
+    """Represents a name record. (`Section 9.e. <https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#9.e>`_)"""
     def __init__(self, nameID, platformID, platEncID, langID, string,
                  location=None):
         Statement.__init__(self, location)
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -7,33 +7,36 @@ import itertools
 SHIFT = " " * 4
 
 __all__ = [
+    'Element',
+    'FeatureFile',
+    'Comment',
+    'GlyphName',
+    'GlyphClass',
+    'GlyphClassName',
+    'MarkClassName',
+    'AnonymousBlock',
+    'Block',
+    'FeatureBlock',
+    'NestedBlock',
+    'LookupBlock',
+    'GlyphClassDefinition',
+    'GlyphClassDefStatement',
+    'MarkClass',
+    'MarkClassDefinition',
     'AlternateSubstStatement',
     'Anchor',
     'AnchorDefinition',
-    'AnonymousBlock',
     'AttachStatement',
     'BaseAxis',
-    'Block',
-    'BytesIO',
     'CVParametersNameStatement',
     'ChainContextPosStatement',
     'ChainContextSubstStatement',
     'CharacterStatement',
-    'Comment',
     'CursivePosStatement',
-    'Element',
     'Expression',
-    'FeatureBlock',
-    'FeatureFile',
-    'FeatureLibError',
     'FeatureNameStatement',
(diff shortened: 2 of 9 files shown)


Lib/fontTools/feaLib/ast.py
```diff
@@ -182,26 +181,37 @@ def asFea(self, indent=""):
...
+        initial and final IDs, and ``glyphs`` is the full list of
+        :class:`GlyphName` objects in the range."""
         if self.curr < len(self.glyphs):
             self.original.extend(self.glyphs[self.curr:])
         self.original.append(("\\{}".format(start), "\\{}".format(end)))
         self.glyphs.extend(glyphs)
         self.curr = len(self.glyphs)
 
     def add_class(self, gc):
+        """Adds glyphs from the given :class:`GlyphClassName` object to the
```
dscorbett, line 213: Some of these docstrings are conjugated like “Add” and some like “Adds”. It should be consistent.

Lib/fontTools/feaLib/ast.py
```diff
@@ -353,11 +382,11 @@ def asFea(self, indent=""):
 
 
 class GlyphClassDefinition(Statement):
-    """Example: @UPPERCASE = [A-Z];"""
+    """Example: ``@UPPERCASE = [A-Z]`;``."""
```
dscorbett, line 385: There is an extra backquote before the semicolon.

Lib/fontTools/feaLib/ast.py
```diff
@@ -527,12 +607,23 @@ def asFea(self, indent=""):
 
 
 class ChainContextPosStatement(Statement):
+    """A chained contextual positioning statement.
+
+    ``prefix``, ``glyph``, and ``suffix`` should be lists of
```
dscorbett, line 612: The parameter is `glyphs`, not `glyph`.

Lib/fontTools/feaLib/ast.py
```diff
@@ -694,15 +815,17 @@ def asFea(self, indent=""):
...
 class LanguageStatement(Statement):
+    """A ``language`` statement within a feature."""
     def __init__(self, language, include_default=True, required=False,
                  location=None):
         Statement.__init__(self, location)
         assert(len(language) == 4)
-        self.language = language
-        self.include_default = include_default
+        self.language = language #: A four-character language tag
+        self.include_default = include_default #: If false, "exclude_deflt"
```
dscorbett, line 824: `exclude_deflt` should be `exclude_dflt`.

Lib/fontTools/feaLib/ast.py
```diff
@@ -718,18 +841,22 @@ def asFea(self, indent=""):
...
+        """Calls the builder object's ``add_language_system`` callback."""
         builder.add_language_system(self.location, self.script, self.language)
 
     def asFea(self, indent=""):
         return "languagesystem {} {};".format(self.script, self.language.strip())
 
 
 class FontRevisionStatement(Statement):
+    """A ``head`` table ``FontRevision`` statement. ``revision`` should be a
+    decimal number with three significant decimal places."""
```
dscorbett, line 859: It’s not clear whether `revision` is a string or a number.

Lib/fontTools/feaLib/ast.py
```diff
@@ -836,23 +982,31 @@ def asFea(self, indent=""):
...
+        """Calls the builder object's ``add_lookup_call`` callback."""
         builder.add_lookup_call(self.lookup.name)
 
     def asFea(self, indent=""):
         return "lookup {};".format(self.lookup.name)
 
 
 class MarkBasePosStatement(Statement):
+    """A mark-to-base positioning rule. The ``base`` must be a
+    `glyph-containing object`_. The ``marks`` should be a list of
```
dscorbett, line 1002: Mixing “must” and “should” implies a meaningful difference. Is the type of `marks` less important than the type of `base`?

Lib/fontTools/feaLib/ast.py
```diff
@@ -936,6 +1129,12 @@ def asFea(self, indent=""):
...
 
 class PairPosStatement(Statement):
+    """A pair positioning statement.
+
+    ``glyphs1`` and ``glyphs2`` should be `glyph-containing objects`_.
+    ``valuerecord1`` should be a :class:`ValueRecord` object;
+    ``valuerecord2`` should be either a :class:`ValueRecord` object or ``None``.
+    """
     def __init__(self, glyphs1, valuerecord1, glyphs2, valuerecord2,
                  enumerated=False, location=None):
```
dscorbett, line 1139: Why isn’t `enumerated` documented?

Lib/fontTools/feaLib/ast.py
```diff
@@ -1043,24 +1266,34 @@ def asFea(self, indent=""):
...
         return "script {};".format(self.script.strip())
 
 
 class SinglePosStatement(Statement):
+    """A single position statement. ``prefix`` and ``suffix`` should be
+    lists of `glyph-containing objects`_.
+
+    ``pos`` should be a list of (`glyph-containing object`_,
+    :class:`ValueRecord`) tuples. There should probably be only one entry in
+    that list."""
```
dscorbett, line 1288: This should probably be more clear. What happens if there isn’t only one entry? Is it a violation of the API? Is it supported but just a silly thing to do?

Lib/fontTools/feaLib/ast.py
```diff
@@ -1084,17 +1317,20 @@ def asFea(self, indent=""):
 
 
 class SubtableStatement(Statement):
+    """Represents a subtable break"""
```
dscorbett, line 1320: Most of the other docstrings end with periods.

Lib/fontTools/feaLib/builder.py
```diff
@@ -17,15 +17,38 @@
...
 
 
 def addOpenTypeFeaturesFromString(font, features, filename=None, tables=None):
+    """Add features from a string to a font. Note that this replaces any
+    features currently present.
+
+    Args:
+        font (feaLib.ttLib.TTFont): The font object.
+        features: A string containing feature code.
+        filename: The directory containing 'filename' is used as the root of
```
dscorbett, line 42: `filename` should be formatted as code.

Lib/fontTools/feaLib/ast.py
simoncozens: I'll clarify this, but separately we should add an assertion at this point, because currently a user could produce invalid output. Clearly it's not a Single Positioning rule if you have more than one entry in the list:

```
>>> SinglePosStatement([ ( GlyphName("a"), ValueRecord(1) ), ( GlyphName("b"), ValueRecord(2) ), ( GlyphName("c"), ValueRecord(3) )], [], [], False).asFea()
'pos a <1 0 0 0> b <2 0 0 0> c <3 0 0 0>;'
```

Doc/source/feaLib/index.rst
```diff
@@ -1,17 +1,40 @@
...
-######
+#########################################
+feaLib: Read/write OpenType feature files
+#########################################
 
-.. toctree::
-   :maxdepth: 1
+fontTools' ``feaLib`` allows for the creation and parsing of Adobe
+Font Development Kit for OpenType feature (``.fea``) files. The syntax
+of these files is described `here <http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html>`_.
```
dscorbett, line 7: All these HTTP links should use HTTPS wherever possible.

simoncozens:
> A lot of functions in ast.py are now documented to accept lists. Do they need to be lists, or is any sequence acceptable? If the argument is a list (or other mutable sequence), is mutating it after passing it to the function supported?

Good question. I think they can be any iterable, and can be mutated after passing:

```
x = [ ( GlyphName("a"), ValueRecord(1) ) ]
foo = SinglePosStatement(x, [],[],False)
x.append( (GlyphName("b"),ValueRecord(2)))
foo.asFea()
# 'pos a <1 0 0 0> b <2 0 0 0>;'
```

I'm not a Python native speaker so I'm not sure how best to put this! Should we say sequence or iterable?

dscorbett:
Mutability is tricky. The documentation should document the guaranteed API, not the current implementation.

[“Iterable”](https://docs.python.org/3/glossary.html#term-iterable) is more broad than [“sequence”](https://docs.python.org/3/glossary.html#term-sequence) in Python. For example, iterators are iterables but can only be iterated over once.

The current implementation seems to work for any [sized](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sized) iterable. That would still be too broad to document: it includes iterables without a guaranteed iteration order, such as sets and (in Python 3.6) dictionaries. Passing one to an `Element` will seem to work but may produce unpredictable results, and as such should not be supported.

I know that these sequences _can_ be mutated after passing, at least in the current implementation; I am wondering whether support was meant to be guaranteed as part of the API. It is currently unclear whether (e.g.) refactoring one of the constructors from `self.glyphs = glyphs` to `self.glyphs = list(glyphs)` (assuming there was some reason to do that) would be backwards-compatible.

Some of these sequences really must specifically be lists. For example, `GlyphClass.append` assumes that `self.glyphs` has an `append` method.

Some constructors are documented to require one-element lists. Is it supported to add or remove elements from such a list after calling the constructor? Why doesn’t the constructor allow a list of any length, as long as it contains only one element by the time `build` is called?

Lib/fontTools/feaLib/ast.py
```diff
@@ -353,11 +382,11 @@ def asFea(self, indent=""):
...
+    """Example: ``@UPPERCASE = [A-Z];``."""
     def __init__(self, name, glyphs, location=None):
         Statement.__init__(self, location)
-        self.name = name
-        self.glyphs = glyphs
+        self.name = name #: class name as a string, without initial ``@``
+        self.glyphs = glyphs #: a :class:`GlyphClass` object
 
     def glyphSet(self):
         return tuple(self.glyphs.glyphSet())
```
dscorbett, line 392: This should be documented.

Lib/fontTools/feaLib/ast.py
```diff
@@ -421,6 +462,9 @@ def addDefinition(self, definition):
             self.glyphs[glyph] = definition
 
     def glyphSet(self):
+        """The set of glyphs contained in this mark class. In the example above,
```
dscorbett, line 465: It’s a set in the general sense but a tuple in the Python sense. Or maybe that is an implementation detail. What kind of collection does this method, in principle, return?

Lib/fontTools/feaLib/ast.py
```diff
@@ -117,14 +109,17 @@ def asFea(g):
...
 
     def __init__(self, location=None):
+        #: location of this element - tuple of ``(filename, line, column)``
         self.location = location
 
     def build(self, builder):
         pass
 
     def asFea(self, indent=""):
+        """Returns this element as a string of feature code."""
```
dscorbett, line 122: `indent` should be documented.

Lib/fontTools/feaLib/ast.py
```diff
@@ -210,21 +220,24 @@ def add_class(self, gc):
 
 
 class GlyphClassName(Expression):
-    """A glyph class name, such as @FRENCH_MARKS."""
+    """A glyph class name, such as ``@FRENCH_MARKS``. This must be instantiated
+    with a :class:`GlyphClassDefinition` object."""
     def __init__(self, glyphclass, location=None):
         Expression.__init__(self, location)
         assert isinstance(glyphclass, GlyphClassDefinition)
```
dscorbett, line 227: Assertions can be disabled, so they shouldn’t be used to catch violations of documented type requirements. It would be better to use something like:
```python3
        if not isinstance(glyphclass, GlyphClassDefinition):
            raise TypeError
```

Lib/fontTools/feaLib/ast.py
```diff
@@ -238,9 +251,12 @@ def asFea(self, indent=""):
 
 
 class AnonymousBlock(Statement):
+    """An anonymous data block"""
```
dscorbett, line 254: Most of the other docstrings end with periods.

Lib/fontTools/feaLib/ast.py
```diff
@@ -320,6 +347,7 @@ def asFea(self, indent=""):
 
 
 class LookupBlock(Block):
+    """A named lookup, containing ``statements``."""
```
dscorbett, line 350: `statements` is a field of the `Block` class, so it seems redundant to call it out here specifically.

Lib/fontTools/feaLib/ast.py
```diff
@@ -718,18 +841,22 @@ def asFea(self, indent=""):
...
     def build(self, builder):
+        """Calls the builder object's ``add_language_system`` callback."""
         builder.add_language_system(self.location, self.script, self.language)
 
     def asFea(self, indent=""):
         return "languagesystem {} {};".format(self.script, self.language.strip())
 
 
 class FontRevisionStatement(Statement):
+    """A ``head`` table ``FontRevision`` statement. ``revision`` should be a
```
dscorbett, line 858: Are table names marked up as code or not? The documentation is inconsistent.

Lib/fontTools/feaLib/ast.py
```diff
@@ -477,6 +548,11 @@ def asFea(self, indent=""):
 
 
 class Anchor(Expression):
+    """An ``Anchor`` element, used inside a ``pos`` rule.
+
+    If a ``name`` is given, this will be used in preference to the coordinates.
+    Other values should be integer.
```
dscorbett, line 554: “integer” should be “integers”. Also, the other values can be `None` instead of integers, and the device tables should not be integers.

Lib/fontTools/feaLib/ast.py
```diff
@@ -500,6 +576,7 @@ def asFea(self, indent=""):
 
 
 class AnchorDefinition(Statement):
+    """A named anchor definition. (2.e.viii). ``name`` should be a string."""
```
dscorbett, line 579: It would be helpful for all these spec references to be links.

Lib/fontTools/feaLib/ast.py
```diff
@@ -43,34 +46,23 @@
...
-    'OrderedDict',
     'PairPosStatement',
-    'Py23Error',
     'ReverseChainSingleSubstStatement',
     'ScriptStatement',
-    'SimpleNamespace',
     'SinglePosStatement',
     'SingleSubstStatement',
     'SizeParameters',
     'Statement',
```
dscorbett, line 63: If `Statement` and `Expression` are in `__all__` they are presumably part of the API and should be documented.

Lib/fontTools/feaLib/ast.py
```diff
@@ -117,14 +109,17 @@ def asFea(g):
 
 
 class Element(object):
+    """A base class representing "something" in a feature file."""
 
     def __init__(self, location=None):
+        #: location of this element - tuple of ``(filename, line, column)``
         self.location = location
 
     def build(self, builder):
```
dscorbett, line 118: What does this method do?

Lib/fontTools/feaLib/ast.py
```diff
@@ -117,14 +109,17 @@ def asFea(g):
 
 
 class Element(object):
+    """A base class representing "something" in a feature file."""
 
     def __init__(self, location=None):
+        #: location of this element - tuple of ``(filename, line, column)``
```
dscorbett, line 115: Does this format mean that the field `location` of `Element` is documented but that the parameter `location` of `Element.__init__` is not documented? Or does Sphinx figure it out and use the same docstring for both?

For this particular field, the docstring is omitting the possibility that the location is `None`.

Lib/fontTools/feaLib/ast.py
```diff
@@ -770,6 +903,14 @@ def asFea(self, indent=""):
 
 
 class LigatureSubstStatement(Statement):
+    """A chained contextual substitution statement.
+
+    ``prefix``, ``glyphs``, and ``suffix`` should be lists of
+    `glyph-containing objects`_; ``replacement`` should be a single
+    `glyph-containing object`_.
+
+    If ``forceChain`` is True, this is expressed as a chaining rule
```
dscorbett, line 912: I think `True` should be `true`: any true value works, not just `True`. Ditto in `MultipleSubstStatement`.

Lib/fontTools/feaLib/ast.py
```diff
@@ -1306,20 +1553,24 @@ def __init__(self, character, tag, location=None):
...
+    """An axis definition, being either a ``VertAxis.BaseTagList/BaseScriptList``
+    pair or a ``HorizAxis.BaseTagList/BaseScriptList`` pair."""
     def __init__(self, bases, scripts, vertical, location=None):
         Statement.__init__(self, location)
-        self.bases = bases
-        self.scripts = scripts
-        self.vertical = vertical
+        self.bases = bases #: A list of baseline tag names as strings 
+        self.scripts = scripts #: A list of script record tuplets (script tag, default baseline tag, base coordinate)
+        self.vertical = vertical #: Boolean; VertAxis if True, HorizAxis if False
```
dscorbett, line 1570: It needn’t be a Boolean: any truth value will do.

simoncozens:
Deciding on the actual guaranteed API is going to have to be someone else's call. My opinion would be to guarantee it for lists, and if you get it working for other kinds of sequence then lucky you but that's undefined behaviour. I also think it is quite useful to be able to mutate the lists after instantiating them.

Questions about e.g. whether to add assertions to ensure only one element when build (and asFea) are called are worth thinking about - but perhaps another time, because this PR is about documentation. :-)

Lib/fontTools/feaLib/ast.py
simoncozens: `#:` is the way you do Sphinx documentation for attributes.

Lib/fontTools/feaLib/ast.py
simoncozens: The `assert` isn't relevant to this documentation PR.

Lib/fontTools/feaLib/ast.py
dscorbett: But there is no `#:` here?

simoncozens: Oh, sorry! - you're talking about the `glyphSet` method? (I was just concentrating on the added lines.) OK, will fix.

Lib/fontTools/feaLib/ast.py
simoncozens: It needs to be ordered. I'll bring this into line with the other classes which have `glyphSet` methods.

Lib/fontTools/feaLib/ast.py
simoncozens: I think I'm going to keep this. The idea is to distinguish it from a lookup reference (`lookup foo;`) which you could also describe as "a named lookup".

justvanrossum:
"sequence" implies indexing and having a length. If the code calls `len(x)` on it or uses indexing `x[n]`, then it's a sequence. If all it does is feed it to (say) `set(x)`, then iterable it is.

It should be rare that code insists on a concrete type, although I would still say "list" if the code mutates the object. "mutable sequence" would be more correct in that case, but that's a bit of a mouthful.

anthrotype:
hopefully all these concerns and uncertainties about the type of arguments and variables should be addressed if/when/once we add support for type hints and static type-checking #1725

Lib/fontTools/feaLib/parser.py
```diff
@@ -12,21 +12,33 @@
...
+
+            from fontTools.feaLib.parser import Parser
+            parser = Parser(file, font.getReverseGlyphMap())
+            parsetree = parser.parse()
+
+    Note: the ``glyphNames`` iterable serves a double role to help distinguish
+    glyph names from ranges in the presence of hyphens and to ensure that glyph
+    names referenced in a feature file are actually part of a font's glyph set.
+    If the iterable is left empty, no glyph name in glyph set checking takes
+    place.
```
anthrotype, line 29: ... and all glyph tokens containing hyphen are treated as literal glyph names, not as ranges. Adding space around the hyphen can also help disambiguate these.

Lib/fontTools/feaLib/parser.py
```diff
@@ -96,16 +111,18 @@ def parse(self):
         return self.doc_
 
     def parse_anchor_(self):
```
anthrotype, line 113: I think the meaning of the `_` at the end was to signal that the method is private (although usually the convention is to put it at the beginning). We don't need to document private methods. How would a client go on and use these in a meaningful way? They simply call `parse` and let the Parser do its job, rather than call specific parse_something_ methods.
Maybe @brawer can confirm this.

anthrotype: @simoncozens maybe you can turn those into inline comments `#` instead of docstrings `"""` as they may still be useful for developers looking at the code (not to the users reading at docs)

Lib/fontTools/feaLib/ast.py
anthrotype: the base implementation does nothing, but subclasses may override to do things with builder

Lib/fontTools/feaLib/ast.py
anthrotype: `#:` is indeed a special comment in sphinx

Lib/fontTools/feaLib/ast.py
```diff
@@ -559,12 +655,22 @@ def asFea(self, indent=""):
 
 
 class ChainContextSubstStatement(Statement):
+    """A chained contextual substitution statement.
+
+    ``prefix``, ``glyphs``, and ``suffix`` should be lists of
+    `glyph-containing objects`_ .
+
+    ``lookups`` should be a list of :class:`LookupBlock` statements, with
```
anthrotype, line 663: > a list of :class:`LookupBlock` statements

I think after #1905 this is now _list of lists_ of LookupBlock statements, no?

Lib/fontTools/feaLib/ast.py
```diff
@@ -694,15 +820,17 @@ def asFea(self, indent=""):
...
 
 class LanguageStatement(Statement):
+    """A ``language`` statement within a feature."""
     def __init__(self, language, include_default=True, required=False,
                  location=None):
         Statement.__init__(self, location)
         assert(len(language) == 4)
-        self.language = language
-        self.include_default = include_default
+        self.language = language #: A four-character language tag
```
anthrotype, line 828: here and elsewhere, please separate inline comments with two spaces

```suggestion
        self.language = language  #: A four-character language tag
```

Lib/fontTools/feaLib/ast.py
anthrotype: yeah, but that should stay an implementation detail. If we were to add type hints to this it would be a `bool`

Lib/fontTools/feaLib/ast.py
anthrotype: that's what `bool`s are for. We don't want to support setting `vertical="yes"`, even though it happens to work

Lib/fontTools/feaLib/parser.py
simoncozens: Many of the private methods had docstrings before I started!
https://github.com/fonttools/fonttools/blob/089f24da6b30cce4e24775a84a4497815ae72d28/Lib/fontTools/feaLib/parser.py#L1670-L1684

If we set the public API either through the `__all__` variable or in Sphinx, so that only those methods appear in the docs, I don't think it's a problem that private methods have docstrings. In fact, it helps with doc coverage testing to have a docstring rather than a comment.

Issue #1879, opened 2020-04-19
[pyftsubset] `AttributeError: InputCoverage` when subsetting any font with GSUB contextual substitution lookup (type = 5), subtable format 3 (coverage-based)


be5invis:
It looks like that `pyftsubset` will try to read property `InputCoverage` of a `fontTools.ttLib.tables.otTables.ContextSubst` object, which does not exist, in function `closure_glyphs` for `otTables.ContextSubst`, injected by `subset/__init.py__` L 904.
For `ContextSubst` the attribute seems to be `Coverage`.

Original: https://github.com/be5invis/Iosevka/issues/502

anthrotype:
@behdad I have this patch that fixes this but am not 100% sure, could you please take a look?

```diff
diff --git a/Lib/fontTools/subset/__init__.py b/Lib/fontTools/subset/__init__.py
index 0119a92b..fc3dd2c5 100644
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -845,8 +845,10 @@ def __subset_classify_context(self):
                                        (r.Backtrack, r.Input, r.LookAhead) = d
                                        (r.BacktrackGlyphCount,r.InputGlyphCount,r.LookAheadGlyphCount,) = (len(d[0]),len(d[1])+1,len(d[2]))
                        elif Format == 3:
-                               Coverage = lambda r: r.Coverage[0]
-                               ChainCoverage = lambda r: r.InputCoverage[0]
+                               def Coverage(r, index=0):
+                                       return r.Coverage[index]
+                               def ChainCoverage(r, index=0):
+                                       return r.InputCoverage[index]
                                ContextData = None
                                ChainContextData = None
                                SetContextData = None
@@ -894,6 +896,8 @@ def __subset_classify_context(self):
                                self.ClassDef = 'InputClassDef' if Chain else 'ClassDef'
                                self.ClassDefIndex = 1 if Chain else 0
                                self.Input = 'Input' if Chain else 'Class'
+                       elif Format == 3:
+                               self.Input = 'InputCoverage' if Chain else 'Coverage'

        if self.Format not in [1, 2, 3]:
                return None     # Don't shoot the messenger; let it go
@@ -987,11 +991,11 @@ def closure_glyphs(self, s, cur_glyphs):
                                if seqi == 0:
                                        pos_glyphs = frozenset(cur_glyphs)
                                else:
[...]

PR #1944, opened 2020-05-11
subset: fix AttributeError while subsetting Context{Subst,Pos} Format 3

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/1879

In Format-3 ChainContext subtables, the array of input coverages is under `InputCoverage` attribute, whereas in non-Chain Context Format 3 subtables it is called simply `Coverage`.

### DIFF
--- a/LICENSE.external
+++ b/LICENSE.external
@@ -26,6 +26,10 @@ XITS font project
 
   This Font Software is licensed under the SIL Open Font License, Version 1.1.
 
+Iosevka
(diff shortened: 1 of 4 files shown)


behdad:
Thanks.  I'm checking.  I'm surprised this happened.  Looks good generally but I like to double-check details.

anthrotype:
@behdad  I believe this is the correct fix, so I'll merge this and include it in the next release, you can always review or fix later.

PR #1948, opened 2020-05-12
Ensure all fonttools CLI tools have help documentation

simoncozens:
Another slightly long PR, sorry; but each commit is self-contained, and some of them are one-liners.

We have a number of command line tools which are somewhat opaque. (varLib.models in particular was very confusing.) This ensures that they all use argparse to have a consistent interface, and all have `--help` documentation which at least details their parameters, and hopefully therefore gives more of a clue about what they do. Those which use logging have had a command-line logging parameter added.

### DIFF
--- a/Lib/fontTools/varLib/interpolatable.py
+++ b/Lib/fontTools/varLib/interpolatable.py
@@ -159,17 +159,25 @@ def test(glyphsets, glyphs=None, names=None):
 
 def main(args):
 	"""Test for interpolatability issues between fonts"""
-	filenames = args
+	import argparse
+	parser = argparse.ArgumentParser(
+		"fonttools varLib.interpolatable",
+		description=main.__doc__,
+	)
+	parser.add_argument('inputs', metavar='FILE', type=str, nargs='+',
+		help="Input TTF files")
+
+	args = parser.parse_args(args)
 	glyphs = None
 	#glyphs = ['uni08DB', 'uniFD76']
 	#glyphs = ['uni08DE', 'uni0034']
 	#glyphs = ['uni08DE', 'uni0034', 'uni0751', 'uni0753', 'uni0754', 'uni08A4', 'uni08A4.fina', 'uni08A5.fina']
 
 	from os.path import basename
-	names = [basename(filename).rsplit('.', 1)[0] for filename in filenames]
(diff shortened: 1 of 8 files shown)


Lib/fontTools/varLib/interpolatable.py
```diff
@@ -159,17 +159,25 @@ def test(glyphsets, glyphs=None, names=None):
 
 def main(args):
```
anthrotype, line 160: let's make sure all main() signatures look like this:

```suggestion
def main(args=None):
```

to have argparse parse_args automatically get sys.argv when None, and being able to call main([...]) with a list of string arguments.

in `if __name__ == "__main__"` below you then only need to call `main()` with no args

Lib/fontTools/varLib/interpolate_layout.py
```diff
@@ -60,26 +60,41 @@ def interpolate_layout(designspace, loc, master_finder=lambda s:s, mapped=False)
 def main(args=None):
 	"""Interpolate GDEF/GPOS/GSUB tables for a point on a designspace"""
 	from fontTools import configLogger
-
+	import argparse
 	import sys
+
 	if args is None:
 		args = sys.argv[1:]
```
anthrotype, line 67: we don't need this if block anymore, because when you pass `None` to parse_args it automatically does exactly that

Lib/fontTools/varLib/models.py
```diff
@@ -425,24 +425,29 @@ def piecewiseLinearMap(v, mapping):
...
+		"fonttools varLib.models",
+		description=main.__doc__,
+	)
+	parser.add_argument('-l', '--loglevel', metavar='LEVEL', default="INFO",
+		help="Logging level (defaults to INFO)")
 
-	# TODO: allow user to configure logging via command-line options
-	configLogger(level="INFO")
+	group = parser.add_mutually_exclusive_group(required=True)
+	group.add_argument('--designspacefile',metavar="DESIGNSPACE",type=str)
```
anthrotype, line 438: `--designspacefile` is too verbose, `--designspace` is enough.
maybe also add `-d` short option (and `-l` for the `--locations`)

anthrotype:
some tests are failing, pls fix thanks

Issue #1942, opened 2020-05-11
Duplicate glyph names: Disambiguated names are invalid


jenskutilek:
The [addition of hash characters](https://github.com/fonttools/fonttools/issues/1119#issuecomment-347512821) to disambiguate glyph names just bit me again, as the resulting glyph names are invalid.

In my case this is a problem because I'm using an input file in a text format to add information to a font, and there I need to use the glyph names that FontTools is assigning. So I need to use `Delta#1`, but the input file parser treats the `#` as a comment marker, thus breaking my input file.

Could we reconsider to change the logic to produce valid glyph names, as [Just suggested](https://github.com/fonttools/fonttools/issues/1119#issuecomment-347512821)? Maybe for compilation, we could still support the `#` notation as not to break existing files.

https://github.com/fonttools/fonttools/blob/1d6dd50d5246e0ee295eb5d2d363c68973be3a43/Lib/fontTools/ttLib/tables/_p_o_s_t.py#L112-L118

anthrotype:
> the resulting glyph names are invalid

what do you mean by invalid?

> the input file parser treats the # as a comment marker

and I suppose you can't change that?

anthrotype:
fontTools does _not_ compile invalid postscript names with `#`, it keeps duplicate glyph names in post table as they were when roundtripping. But since fontTools needs glyph names to be unique it has to internally disambiguate between them by adding the `#<digit>` suffix.
see my comments in the other thread
https://github.com/fonttools/fonttools/issues/1119#issuecomment-347486107
https://github.com/fonttools/fonttools/issues/1119#issuecomment-347514627

jenskutilek:
Yes, I have read the comments. I know the names containing a `#` are not written back into the font. But they are used while the font is decompiled, so every script that deals with the font through `TTFont` and insists on validity of glyph names will have a problem.

I can change the input format parsing, but it's a bit of a hack, as it requires a certain formatting of comments (I can't change the comment marker that is being used, only under which circumstances it denotes a comment, e.g. only at the start of a line or surrounded by whitespace).

```
# This is a comment
# The line below does not contain a comment
Delta#1
   # But this again is a comment
  #Whataboutthis?
```

justvanrossum:
Doesn't the fact that disambiguation is need point to a problem in the source font?

I'm not opposed to a change here, but I do think that a font containing duplicate glyph names in the post table is problematic in other ways, too.

jenskutilek:
[...]

PR #1949, opened 2020-05-13
[docs] Document fontTools.merge

simoncozens:


Doc/source/merge.rst
```diff
@@ -1,8 +1,10 @@
-#####
-merge
-#####
+####################################
+merge: Merge multiple fonts into one
+####################################
 
-.. automodule:: fontTools.merge
+``fontTools.merge`` provides both a library and a command line interface
+(``pyftmerge``) for merging multiple fonts together.
```
anthrotype, line 6: I would like eventually to get rid of the `pyft*` scripts and encourage people to use the `fonttools {subset,merge,...}` entry point. Maybe just mention `fonttools merge` here. We can decide to deprecate them for good

Lib/fontTools/merge.py
```diff
@@ -941,6 +941,38 @@ def __delitem__(self, k):
...
+	This class merges multiple files into a single OpenType font, taking into
+	account complexities such as OpenType layout (``GSUB``/``GPOS``) tables and
+	cross-font metrics (e.g. ``hhea.ascent`` is set to the maximum value across
+	all the fonts).
+
+	If multiple glyphs map to the same Unicode value, subsequent glyphs are
+	renamed and a lookup in the ``locl`` feature will be created to disambiguate
+	them. For example, if the arguments are an Arabic font and a Latin font and
+	both contain a set of parentheses, the Latin glyphs will be renamed to
+	``parenleft#1`` and ``parenright#1``, and the equivalent of the following
```
anthrotype, line 955: The addition of hash characters to disambiguate glyph names in the merger reminds me of https://github.com/fonttools/fonttools/issues/1942

Unlike with that other issue (where a font contains duplicates in post table that fonttools needs to deduplicate internally and temporarily), in the case of the merger, we _are_ writing invalid glyph names (containing an invalid character in postscript `#`) in the merged font's `post` table.

Let's open a separate issue for this; maybe we can us a dot `.` here.

/cc @justvanrossum @jenskutilek

Lib/fontTools/merge.py
```diff
@@ -941,6 +941,38 @@ def __delitem__(self, k):
...
+	feature code will be added:
+
+	.. code:: none
+
+		feature locl {
+			script latn;
+			language dflt;
+			sub [parenleft parenright] by [parenleft#1 parenright#1];
```
anthrotype, line 963: this snippet of FEA code is problematic given the `#` indicates inline comments (but I guess it's ok for the time being while we address that)

simoncozens: Thought of that. The `.. code:: none` turns off syntax highlighting for this block.

Lib/fontTools/merge.py
khaledhosny: Note that there is also code that checks if the duplicate glyphs are the same, so there are more criteria considered than code point: https://github.com/fonttools/fonttools/blob/cfb261cf0279c352538f06f191c2df751477a177/Lib/fontTools/merge.py#L368-L377

Lib/fontTools/merge.py
khaledhosny: I don’t find the code snippet to be adding much value IMHO, something like ”with a lookup mapping ``parenleft`` and ``parenright`` to ``parenleft#1`` and ``parenright#1``, respectively” would be enough.

khaledhosny: Also, if there is already a `locl` feature (with a script other than `DFLT`) the lookup will be inserted in it, so this needs to be mentioned. Also a `GSUB` table is required, merging a font with duplicate glyphs and no `GSUB` table raises an error.

simoncozens: I thought the choice of script/language for the locl feature was worth noting. We can say that the user can switch between the two glyphs using locl - but it's important to know what settings you need to use to get to which glyphset.

khaledhosny: I guess I have aversion to using FEA as _the_ OpenType language, and prefer tool-neutral description as we shouldn’t assume everyone reading the docs can readily understand this syntax.

simoncozens: OK, that's fair; I also agree that FEA is not the only language for feature code. (And certainly not the best ;-) ) Will rewrite it.

PR #1956, opened 2020-05-14
[docs] fontTools.misc.* part 1

simoncozens:
Stopping here before it gets unwieldy.


Lib/fontTools/misc/arrayTools.py
```diff
@@ -21,71 +24,170 @@ def calcBounds(array):
...
-    """Round and cast to 16 bit integer."""
+    """Round a list of floats to 16-bit signed integers.
+
+    Args:
+        array: List of float values.
+
+    Returns:
+        A list of rounded integers.
+    """
     return [int(math.floor(i+0.5)) for i in array]
```
anthrotype, line 112: this is basically `fontTools.misc.fixedTools.otRound` in a list comprehension.
This `asInt16` function does not seem to be used anywhere internally. I wonder if we can get rid of it.
The name is also misleading in that it's not really casting or truncating the integers to make sure they fit a signed 16-bit integer.

Lib/fontTools/misc/bezierTools.py
```diff
@@ -437,10 +571,19 @@ def _splitCubicAtT(a, b, c, d, *ts):
...
-    is neither guaranteed to be sorted nor to contain unique values!
+    """Solve a quadratic equation.
+
+    Solves *a*x*x + b*x + c = 0* where a, b and c are real.
+
+    Args:
+        a: coefficient of *x²*
+        b: coefficient of *x*
+        c: constant term
+        sqrt: square root function to be used
```
anthrotype, line 582: I believe the fact `sqrt` is among the parameters is just an optimization (though it is doubtful it actually makes a difference in real world).
There used to be a recommendation about that in itertools documentation for python 2.7
https://docs.python.org/2/library/itertools.html

> many of the above recipes can be optimized by replacing global lookups with local variables defined as default values

But it seems that comment has been removed in more recent versions of the python documentation. Maybe it no longer applies. In any case, you don't need to mention those here.
I noticed a similar thing with `min` and `max` functions elsewhere in fonttools (I don't exactly recall where).
If you like you could rename the parameter `_sqrt=sqrt` to make it clear that is somewhat private.
Or you can profile and decide to remove it. Sure, not in this same PR.

Lib/fontTools/misc/bezierTools.py
```diff
@@ -462,25 +605,37 @@ def solveQuadratic(a, b, c,
...
-    [-2.911392, 1.0, 1.0]
-    >>> solveCubic(1.0, -4.5, 6.75, -3.375)
-    [1.5, 1.5, 1.5]
-    >>> solveCubic(-12.0, 18.0, -9.0, 1.50023651123)
-    [0.5, 0.5, 0.5]
-    >>> solveCubic(
-    ...     9.0, 0.0, 0.0, -7.62939453125e-05
-    ... ) == [-0.0, -0.0, -0.0]
-    True
+    """Solve a quadratic equation.
```
anthrotype, line 608: copypasta, I think you meant cubic here

Lib/fontTools/misc/bezierTools.py
```diff
@@ -462,25 +605,37 @@ def solveQuadratic(a, b, c,
...
+
+    Solves *a*x*x*x + b*x*x + c*x + d = 0* where a, b, c and d are real.
+
+    Args:
+        a: coefficient of *x³*
+        b: coefficient of *x²*
+        c: coefficient of *x*
+        d: constant term
+        sqrt: square root function to be used
```
anthrotype, line 617: this argument is not present in solveCubic (yet more arguments to get rid of it in solveQuadratic?)

Doc/source/misc/cliTools.rst
```diff
@@ -1,6 +1,6 @@
-########
-cliTools
-########
+###################################################################
+cliTools: Utilities for command-line interfaces and console scripts
```
anthrotype, line 2: I am not sure if we should advertise this module outside of fontTools, I think it's mostly for internal use. Oh well, I guess it doesn't harm.

Lib/fontTools/misc/eexec.py
```diff
@@ -1,5 +1,15 @@
...
+PostScript Type 1 fonts make use of two types of encryption: charstring
+encryption and ``eexec`` encryption. Charstring encryption is used for
+the charstrings themselves, while ``eexec`` is used to encrypt larger
+sections of the font program, such as the ``Private`` and ``CharStrings``
+dictionaries. Despite the different names, the algorithm is the same,
+although ``eexec`` encryption uses a fixed initial key R=55665.
+
+The algorithm uses cipher feedback, meaning that the ciphertext is used
+to modify the key. Because of this, the routines in this module return
+the new key at the end of the operation.
```
anthrotype, line 11: thanks for taking the trouble of documenting this!

Lib/fontTools/misc/bezierTools.py
justvanrossum: Yeah, in the old days this really made a performance difference (global lookups are much slower than locals), and this arg should be scrapped.

Lib/fontTools/misc/bezierTools.py
```diff
@@ -29,7 +29,19 @@
 
 
 def calcCubicArcLength(pt1, pt2, pt3, pt4, tolerance=0.005):
-    """Return the arc length for a cubic bezier segment."""
+    """Calculates the arc length for a cubic Bezier segment.
+
+    Whereas :func:`approximateCubicArcLength` approximates the length, this
+    function calculates it by "measuring", recursively dividing the curve
+    until the divided segments are shorter than ``tolerance``.
```
justvanrossum, line 36: I don't think this is correct: it stops subdividing once the curve is sufficiently flat, ie. the distance (pt1, pt4) is sufficiently close to the sum of the distances of (pt1, pt2), (pt2, pt3) and (pt3, pt4).

behdad: Right.

Doc/source/misc/cliTools.rst
behdad: `makeOutputFileName` *is* used outside I believe.

Lib/fontTools/misc/bezierTools.py
behdad: Oh this `sqrt=sqrt` is an old optimization hack. Wasn't really intended to allow replacing `sqrt`.

Lib/fontTools/misc/arrayTools.py
behdad: Yes, please kill it. Was added by @justvanrossum in 2000: deb3b632f; unused.

PR #1957, opened 2020-05-14
[varLib] Allow feature variations to be active across the entire space

madig:
Original use case: A VF containing weight and italic axes with a subordinate VF containing only a weight axis of the italic forms. Some Cyrillic glyphs need to be swapped out for special italic forms for some languages like Macedonian. The solution we settled on requires that the special forms are substituted across the entire weight axis range, which varLib filtered out before.

### DIFF
--- a/Lib/fontTools/varLib/featureVars.py
+++ b/Lib/fontTools/varLib/featureVars.py
@@ -136,7 +136,11 @@ def overlayFeatureVariations(conditionalSubstitutions):
                     remainder = hashdict(remainder)
                     newMap[remainder] = newMap.get(remainder, 0) | rank
         boxMap = newMap
-    del boxMap[hashdict()]
+    # Delete initMapInit from boxMap. Need to test for `0` explicitly because there
+    # could be an actual user-defined rule that applies to the entire space, which
+    # would be `1`.
+    if boxMap[hashdict()] == 0:
+        del boxMap[hashdict()]
 
     # Generate output
     items = []
--- a/Lib/fontTools/varLib/featureVars.py
+++ b/Lib/fontTools/varLib/featureVars.py
@@ -82,6 +82,7 @@ def overlayFeatureVariations(conditionalSubstitutions):
     ...     ([{"wght": (0.5, 1.0)}], {"dollar": "dollar.rvrn"}),
     ...     ([{"wght": (0.5, 1.0)}], {"dollar": "dollar.rvrn"}),
     ...     ([{"wdth": (0.5, 1.0)}], {"cent": "cent.rvrn"}),
+    ...     ([{"wght": (0.5, 1.0), "wdth": (-1, 1.0)}], {"dollar": "dollar.rvrn"}),
     ... ]
     >>> from pprint import pprint
     >>> pprint(overlayFeatureVariations(condSubst))
@@ -136,16 +137,14 @@ def overlayFeatureVariations(conditionalSubstitutions):
                     remainder = hashdict(remainder)
                     newMap[remainder] = newMap.get(remainder, 0) | rank
         boxMap = newMap
-    # Delete initMapInit from boxMap. Need to test for `0` explicitly because there
-    # could be an actual user-defined rule that applies to the entire space, which
-    # would be `1`.
-    if boxMap[hashdict()] == 0:
-        del boxMap[hashdict()]
 
     # Generate output
     items = []
     for box,rank in sorted(boxMap.items(),
                            key=(lambda BoxAndRank: -popCount(BoxAndRank[1]))):
+        # Skip the initial entire-space box only if no layers apply.
+        if rank == 0:
+            continue
         substsList = []
         i = 0
         while rank:
--- a/Lib/fontTools/varLib/featureVars.py
+++ b/Lib/fontTools/varLib/featureVars.py
@@ -142,7 +142,7 @@ def overlayFeatureVariations(conditionalSubstitutions):
     items = []
     for box,rank in sorted(boxMap.items(),
                            key=(lambda BoxAndRank: -popCount(BoxAndRank[1]))):
-        # Skip the initial entire-space box only if no layers apply.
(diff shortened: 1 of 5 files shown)


madig:
@justvanrossum please also have a look, as I'm not quite sure my fix is a real fix.

justvanrossum:
I think I understand the use case. @behdad wrote this incarnation of the code and should be better able to judge whether this is the correct fix or not.

behdad:
This does look correct.  I'm not fully happy about the comment.  I didn't confirm if it will always be 1 in non-zero cases.  You can change it to "Remove the initial entire-space box only if no layers apply."

behdad:
Can you add a test in the docstring as well?

behdad:
Actually that is not good enough.  We should inspect the last element and drop it if the rank is zero.  The box can be different if something cut it off.  Let me see if I can fix that quick.

behdad:
> We should inspect the last element and drop it if the rank is zero.

Not exactly that.  Let me play with this.

behdad:
I'm having a hard time getting my setup to run doctests properly so I can't fully get to the bottom of this.  But here's my suggested patch instead, if you can test:
```
diff --git a/Lib/fontTools/varLib/featureVars.py b/Lib/fontTools/varLib/featureVars.py
index dab9a0b1..8d4d2689 100644
--- a/Lib/fontTools/varLib/featureVars.py
+++ b/Lib/fontTools/varLib/featureVars.py
@@ -136,12 +136,12 @@ def overlayFeatureVariations(conditionalSubstitutions):
                     remainder = hashdict(remainder)
                     newMap[remainder] = newMap.get(remainder, 0) | rank
         boxMap = newMap
-    del boxMap[hashdict()]
 
     # Generate output
     items = []
     for box,rank in sorted(boxMap.items(),
                            key=(lambda BoxAndRank: -popCount(BoxAndRank[1]))):
+        if rank == 0: continue
         substsList = []
         i = 0
         while rank:
```

What to test is something like: `([{"wght": (0.5, 1.0), "wdth": (-1, 1.0)}], {"dollar": "dollar.rvrn"})` as well.  Not sure if the old code was handling that fine.  I have a hunch that it wasn't.

madig:
With your fix:

```
>>> overlayFeatureVariations(
>>>    [([{"wght": (0.5, 1.0), "wdth": (-1, 1.0)}], {"dollar": "dollar.rvrn"})]
>>> )
[({"wght": (0.5, 1.0)}, [{"dollar": "dollar.rvrn"}])]
```

😮

anthrotype:
> I'm having a hard time getting my setup to run doctests properly

the docstrings are discovered and run automatically by pytest. If you just want to run the docstrings for the `featureVars` module you can do `pytest Lib/fontTools/varLib/featureVars.py`.
The recommended way to run the whole test suite is to use `tox`.

madig:
Tox runs the docstrings, bare pytest doesn't.

anthrotype:
> Tox runs the docstrings, bare pytest doesn't.

that's not true. It's pytest that's running all the tests, tox is simply invoking pytest (via coverage -m pytest).
Here's the docstring test in featureVars, run directly via pytest. If you see anything missing, it's probably because your docstrings aren't propertly formatted.

```
$ pytest Lib/fontTools/varLib/featureVars.py -vv
============================== test session starts ===============================
platform darwin -- Python 3.8.1, pytest-5.4.1, py-1.8.1, pluggy-0.13.1 -- /Users/clupo/Github/fonttools/.venv/bin/python3.8
cachedir: .pytest_cache
rootdir: /Users/clupo/Github/fonttools, inifile: setup.cfg
collected 2 items

Lib/fontTools/varLib/featureVars.py::fontTools.varLib.featureVars.cleanupBox PASSED [ 50%]
Lib/fontTools/varLib/featureVars.py::fontTools.varLib.featureVars.overlayFeatureVariations PASSED [100%]

================================ warnings summary ================================
Lib/fontTools/misc/py23.py:11
  /Users/clupo/Github/fonttools/Lib/fontTools/misc/py23.py:11: DeprecationWarning: The py23 module has been deprecated and will be removed in the next release. Please update your code.
    warnings.warn(

-- Docs: https://docs.pytest.org/en/latest/warnings.html
========================== 2 passed, 1 warning in 0.13s ==========================
```

anthrotype:
OK I see what you meant now about the doctests not being run when invoking `pytest` without explicit test paths.
I fixed that in https://github.com/fonttools/fonttools/commit/3d705b29de71cb19119ecf61cb750b1ba27c9522

Lib/fontTools/varLib/featureVars.py
```diff
@@ -136,12 +137,14 @@ def overlayFeatureVariations(conditionalSubstitutions):
                     remainder = hashdict(remainder)
                     newMap[remainder] = newMap.get(remainder, 0) | rank
         boxMap = newMap
-    del boxMap[hashdict()]
 
     # Generate output
     items = []
     for box,rank in sorted(boxMap.items(),
                            key=(lambda BoxAndRank: -popCount(BoxAndRank[1]))):
+        # Skip the initial entire-space box only if no layers apply.
```
behdad, line 145: The comment is not accurate.  The empty entire-space might have been broken into smaller pieces.

During the overlay there's a requirement that the boxes must cover the entire space.  During output-generation, we skip any box that doesn't have any substitution.  That's what we are doing here.

The rest looks good.  Thanks!

madig: So the comment should be "Skip any box that doesn't have any substitution"? :o

behdad: > During the overlay there's a requirement that the boxes must cover the entire space. During output-generation, we skip any box that doesn't have any substitution. That's what we are doing here.

Okay I debugged that, and the way overlayBoxes works currently that can't happen.  But the change is correct and good regardless.  Adjust the comment and we are good.

PR #1961, opened 2020-05-15
[feaLib.ast] Restore compatibility with 4.9.0

dscorbett:
Version 4.10.0 broke `feaLib.ast.ChainContextPosStatement` and `ChainContextSubstStatement`. Previously each of these classes had a `lookups` field that was a list of lookups. #1905 changed it to a list of lists of lookups. For compatibility, it should treat a lookup within the outer list as if it were a list of a single entry.

In 4.9.0, and once again with my fix, this script:
```python3
#!/usr/bin/env python3

from fontTools.feaLib.ast import (
    ChainContextSubstStatement,
    GlyphName,
    LookupBlock,
)

print(ChainContextSubstStatement(
    [],
    [GlyphName('x')],
    [],
    [LookupBlock('dummy')],
))
```
printed this:
```
sub x' lookup dummy;
```
In 4.10.0, it prints this:
```
  File "./bug.py", line 13, in <module>
    [LookupBlock('dummy')],
  File "/usr/local/lib/python3.7/site-packages/fontTools/feaLib/ast.py", line 128, in __str__
    return self.asFea()
  File "/usr/local/lib/python3.7/site-packages/fontTools/feaLib/ast.py", line 690, in asFea
    for lu in self.lookups[i]:
TypeError: 'LookupBlock' object is not iterable
```

Please merge this, or something similar, and release 4.10.1.

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -617,16 +617,24 @@ class ChainContextPosStatement(Statement):
     ``prefix``, ``glyphs``, and ``suffix`` should be lists of
     `glyph-containing objects`_ .
 
-    ``lookups`` should be a list of lists containing :class:`LookupBlock`
-    statements. The length of the outer list should equal to the length of
-    ``glyphs``; the inner lists can be of variable length. Where there is no
-    chaining lookup at the given glyph position, the entry in ``lookups``
-    should be ``None``."""
+    ``lookups`` should be a list of elements representing what lookups
+    to apply at each glyph position. Each element should be a
+    :class:`LookupBlock` to apply a single chaining lookup at the given
+    position, a list of :class:`LookupBlock`\ s to apply multiple
+    lookups, or ``None`` to apply no lookup. The length of the outer
+    list should equal the length of ``glyphs``; the inner lists can be
+    of variable length."""
 
     def __init__(self, prefix, glyphs, suffix, lookups, location=None):
         Statement.__init__(self, location)
         self.prefix, self.glyphs, self.suffix = prefix, glyphs, suffix
-        self.lookups = lookups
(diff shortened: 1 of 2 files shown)


dscorbett:
@simoncozens, what do you think of the rewording of the docstring?

simoncozens:
Feels a bit funny to say that it's a list of lists, but then in a later sentence say it doesn't have to be. I suggest rewriting the paragraph to say it's a list made up elements which are either lists or bare lookups.

dscorbett:
I’ve updated the documentation.

The fix I implemented is to mutate the lookup list passed to the `ChainContextPosStatement`/`ChainContextSubstStatement` constructor. An alternative is to create a temporary one-element list whenever necessary, without storing it in a field, like this:
https://github.com/fonttools/fonttools/blob/db26cf804e3380b5e9caed3a33d640b8e8a54835/Lib/fontTools/feaLib/builder.py#L229-L231
Any preference?

anthrotype:
@dscorbett it shouldn't matter in most circimstaces, but to avoid surprises, it's probably safer to avoid mutating the input list.

dscorbett:
I agree. I’ll make that change soon.

anthrotype:
actually, I prefer if the ChainContext{Subst,Pos}Statement's constructor normalizes a bare list of lookups to list of lists of lookups. Only the ast classes should change to accommodate the old and new behavior, not also the builder.

anthrotype:
what I mean is inside `__init__` you make sure self.lookups is always a list of lists of lookups, then later other code (in asFea() or builder) can assume that will be the case.

dscorbett:
Okay, so basically what I had at first, except mutating a copy of the input list instead of the input list?

anthrotype:
yes that's what I meant, thank you!

PR #1963, opened 2020-05-16
[ttLib.name] Fix findMultilingualName()

justvanrossum:
When reading from binary, `name.string` may be an encoded bytes sequence. Previously, `findMultilingualName()` would fail to match these with the requested input.

Calling `toUnicode()` before we compare to the requested string fixes this.


justvanrossum:
(will look into the failing tests now)

Issue #1962, opened 2020-05-16
Make TTFont and other objects pickleable by implementing __reduce__


behdad:
For lazy objects it should `ensureDecompiled()`; for our list-like types, for ttFont, etc.

behdad:
Or `__getstate__` for some should be enough.

behdad:
Implementing `__reduce__` also makes `copy()` and `deepcopy()` to work as well.

anthrotype:
I think the problem is only the SFNTReader which holds a reference to an unpickled file object when TTFont is loaded with lazy=False. Normally, with lazy=None (default) or with lazy=False, a TTFont can be pickled because we only store a copy of the input file's data in an in-memory BytesIO.

We already implemented a custom `__deepcopy__` for SFNTReader to handle the lazy=True case:

https://github.com/fonttools/fonttools/blob/db26cf804e3380b5e9caed3a33d640b8e8a54835/Lib/fontTools/ttLib/sfnt.py#L125-L147

To make TTFont (and all children objects) pickleable, we only need to make the SFNTReader pickleable.
The rest of the lazy objects should just work.

PR #1967, opened 2020-05-18
sfnt: add __getstate__ and __setstate__ to SFNTReader to make it pickelable

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/1962

### DIFF
--- a/Lib/fontTools/ttLib/sfnt.py
+++ b/Lib/fontTools/ttLib/sfnt.py
@@ -12,7 +12,8 @@ classes, since whenever to number of tables changes or whenever
 a table's length chages you need to rewrite the whole file anyway.
 """
 
-from fontTools.misc.py23 import *
+from io import BytesIO
+from fontTools.misc.py23 import Tag
 from fontTools.misc import sstruct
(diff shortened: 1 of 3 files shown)


anthrotype:
it turns out we only need `__getstate__` for SFNTReader to make it both pickle-able and deepcopy-able

anthrotype:
to keep it lazy and avoid the extra copy from external file to in-memory bytes stream, I think it's preferable that we only serialize the lazy external file name and position (in `__getstate__`) then after unpickling we re-open the same file at the saved position (in custom `__setstate__`).
We no longer need a custom `__deepcopy__` if we implement the pickle protocol, so I removed that.

anthrotype:
I believe it is safe that both the original lazy=True TTFont and its pickled copy point to the same filename (_not_ the same file object, each has its own file object, but both read from the same named file), since the SFNTReader.file is opened in read-only mode (`rb`).
Previously the custom `SFNTReader.__deepcopy__` would always copy all the file's content to a new BytesIO. Whereas now the lazy copy will still share the external refererence to the filesystem.

PR #1970, opened 2020-05-22
Remove duplicate glyphs from kerning groups.

tomarcher101:
Add check that removes duplicate glyphs from kerning groups. This change is in relation to [rule 5](https://github.com/unified-font-object/ufo-spec/blob/gh-pages/versions/ufo3/groups.plist.md#5-glyphs-should-not-appear-more-than-once-in-a-single-kerning-group) of the kerning group rules found on the [Unified Font Object specification](https://github.com/unified-font-object/ufo-spec).

> 5. Glyphs should not appear more than once in a single kerning group.
Any appearances of a glyph after the first appearance of that glyph in a group must be ignored by authoring tools. Duplicate glyphs in a kerning group provide no additional information to authoring tools and, thus, may be safely ignored silently.

### DIFF
--- a/Lib/fontTools/ufoLib/__init__.py
+++ b/Lib/fontTools/ufoLib/__init__.py
@@ -462,7 +462,7 @@ class UFOReader(_UFOBaseIO):
 		groups = self._getPlist(GROUPS_FILENAME, {})
 		# remove any duplicate glyphs in a kerning group
 		for groupName, glyphList in groups.items():
-			if groupName.startswith('public.kern1.') or groupName.startswith('public.kern2.'):
+			if groupName.startswith(('public.kern1.', 'public.kern2.')):
 				groups[groupName] = list(OrderedDict.fromkeys(glyphList))
 		return groups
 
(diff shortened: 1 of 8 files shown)


Lib/fontTools/ufoLib/__init__.py
```diff
@@ -459,7 +459,12 @@ def readMetaInfo(self, validate=None):
 	# groups.plist
 
 	def _readGroups(self):
-		return self._getPlist(GROUPS_FILENAME, {})
+		groups = self._getPlist(GROUPS_FILENAME, {})
+		# remove any duplicate glyphs in a kerning group
+		for groupName, glyphList in groups.items():
+			if groupName.startswith('public.kern1.') or groupName.startswith('public.kern2.'):
```
justvanrossum, line 465: I have recently learned that you can spell multiple startswith options with a tuple of strings:
`if groupName.startswith(('public.kern1.', 'public.kern2.')):`

tomarcher101: Oh nice! I'll update it to use a tuple. Thanks!

PR #1973, opened 2020-05-27
[feaLib] Explicitly look outside UFO for includes

madig:
ufo2ft reads a UFO's feature file into a StringIO and sets the `name` attribute to the UFO, not the feature file within: https://github.com/googlefonts/ufo2ft/blob/53dd7c6/Lib/ufo2ft/featureCompiler.py#L34-L40. `IncludingLexer` then uses that attribute to determine where to look for includes.

I want to change ufo2ft to make it point to the actual feature file and change `IncludingLexer` to take an explicit directory to use for includes, so that users get a better error message if something goes wrong. For backwards compatibility, `IncludingLexer` continue handling the `name` attribute as above.

### DIFF
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -39,7 +39,7 @@ class Parser(object):
     CV_FEATURE_TAGS = {"cv%02d" % i for i in range(1, 99+1)}
 
     def __init__(self, featurefile, glyphNames=(), followIncludes=True,
-                 include_dir=None, **kwargs):
+                 includeDir=None, **kwargs):
 
         if "glyphMap" in kwargs:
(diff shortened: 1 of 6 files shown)


Lib/fontTools/feaLib/parser.py
```diff
@@ -39,7 +39,7 @@ class Parser(object):
     CV_FEATURE_TAGS = {"cv%02d" % i for i in range(1, 99+1)}
 
     def __init__(self, featurefile, glyphNames=(), followIncludes=True,
-                 **kwargs):
+                 include_dir=None, **kwargs):
```
anthrotype, line 42: the rest of the parameters use camelCase, so it should be `includeDir` here. I know..

PR #1981, opened 2020-06-01
fontTools.misc.* documentation, part 2

simoncozens:


Lib/fontTools/misc/fixedTools.py
```diff
@@ -130,40 +203,74 @@ def strToFixed(string, precisionBits):
...
 
 def ensureVersionIsLong(value):
-	"""Ensure a table version is an unsigned long (unsigned short major,
-	unsigned short minor) instead of a float."""
+	"""Ensure a table version is an unsigned long.
+
+	OpenType table version numbers are expressed as a single unsigned long
+	comprising of an unsigned short major version and unsigned short minor
+	version. This function detects if the value to be used a version number
+	looks too small, and converts it to fixed-point if so.
```
dscorbett, line 266: What counts as too small? This should be documented because `0x0000_0064` is apparently too small, though someone might intend that to represent version 0.100.

How does it convert to fixed-point? This should be documented because someone might expect version 1.1 to be `0x0001_0001` or `0x0001_199A` or `0x0001_1000`.

simoncozens:
Comments addressed; can I get a 👍👎?

Lib/fontTools/misc/plistlib.py
```diff
@@ -395,6 +395,19 @@ def totree(
     pretty_print=True,
     indent_level=1,
 ):
+    """Convert a value derived from a plist into an XML tree.
+
+    Args:
+        value: Any kind of value to be serialized to XML.
+        sort_keys: Whether keys of dictionaries should be sorted.
+        skipkeys: Keys which should be skipped when enumerating.
```
dscorbett, line 403: This makes it sound like `skipkeys` is a collection, but it actually seems to be a boolean.

Lib/fontTools/misc/plistlib.py
```diff
@@ -395,6 +395,19 @@ def totree(
...
+        use_builtin_types: Whether to use standard plist types or
+            to wrap values in a Data block. Deprecated.
```
dscorbett, line 405: `Data` should be formatted as a Python identifier.

This description is vague. It makes it sound like `use_builtin_types=False` will wrap _all_ values in `Data` blocks. The comment above `USE_BUILTIN_TYPES` gives more information. Also, this should explain what the default value of `None` means.

Lib/fontTools/misc/plistlib.py
```diff
@@ -410,6 +423,16 @@ def totree(
...
 def fromtree(tree, use_builtin_types=None, dict_type=dict):
+    """Converts an XML tree to a plist structure.
+
+    Args:
+        tree: An ``etree`` ``Element``.
+        use_builtin_types: Whether to use standard plist types or
+            to wrap values in a Data block. Deprecated.
+        dict_type: What type to use for dictionaries.
+
+    Returns: A dictionary which can be serialized as a plist.
```
dscorbett, line 434: This doesn’t necessarily return a dictionary. The return type is whatever is appropriate for the plist.

Lib/fontTools/misc/intTools.py
```diff
@@ -1,12 +1,19 @@
-"""Misc integer tools."""
-
 from fontTools.misc.py23 import *
 
 __all__ = ['popCount']
 
 
 def popCount(v):
-    """Return number of 1 bits in an integer."""
+    """Return number of 1 bits (population count) of an integer.
```
dscorbett, line 7: This should explain what it does for a negative integer.

Lib/fontTools/misc/loggingTools.py
```diff
@@ -469,10 +477,12 @@ def assertRegex(self, regexp, msg=None):
...
 class LogMixin(object):
 	""" Mixin class that adds logging functionality to another class.
-	You can define a new class that subclasses from LogMixin as well as
+
+	You can define a new class that subclasses from ``LogMixin`` as well as
 	other base classes through multiple inheritance.
-	All instances of that class will have a 'log' property that returns
-	a logging.Logger named after their respective <module>.<class>.
+	All instances of that class will have a ``log`` property that returns
+	a logging.Logger named after their respective ``<module>.<class>``.
```
dscorbett, line 484: `logging.Logger` should be formatted as code.

Lib/fontTools/misc/plistlib.py
```diff
@@ -410,6 +423,16 @@ def totree(
 
 
 def fromtree(tree, use_builtin_types=None, dict_type=dict):
+    """Converts an XML tree to a plist structure.
```
dscorbett, line 426: The inflections of the docstrings in this PR are inconsistent. Is it “Convert” or “Converts”?

Lib/fontTools/misc/plistlib.py
anthrotype: Right. Didn't we say we would use Google docstring style? That has the type of each parameter in parentheses. We should use that wherever makes sense.

Lib/fontTools/misc/plistlib.py
anthrotype: yes, only binary data is wrapped

Lib/fontTools/misc/plistlib.py
anthrotype: > The return type is whatever is appropriate for the plist.

correct. Maybe just say "a plist object"

dscorbett: Other docstrings say “An object (usually a dictionary)”.

anthrotype: sounds good

Lib/fontTools/misc/fixedTools.py
```diff
@@ -21,78 +38,126 @@
...
-	True
-	>>> math.isclose(f2, -0.61883544921875)
-	True
+	"""Converts a float to a fixed-point number and back again.
+
+	By converting the float to fixed, rounding it, and converting it back
+	to float again, this returns a floating point values which is exactly
+	representable in fixed-point format.
+
+	Note: this **is** equivalent to ``fixedToFloat(floatToFixed(value))``.
```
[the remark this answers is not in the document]
dscorbett: Why was this marked resolved? `fixedToFloat(floatToFixed(value))` is not valid use of this API.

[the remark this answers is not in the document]
anthrotype: True, the precisionBits parameter is required in both. I think it was meant to convey the idea of the roundtrip operation, more than literal code, but you're right.

Lib/fontTools/misc/plistlib.py
simoncozens: We are using Google docstring style, but none of the examples on http://google.github.io/styleguide/pyguide.html have types in the `Args:` block, which is why I didn't add any. I'll add them where I can. And it looks like `skipkeys` does something different to what I thought, so I'll fix that.

Lib/fontTools/misc/plistlib.py
```diff
@@ -400,13 +405,25 @@ def totree(
...
         sort_keys: Whether keys of dictionaries should be sorted.
-        skipkeys: Keys which should be skipped when enumerating.
-        use_builtin_types: Whether to use standard plist types or
-            to wrap values in a Data block. Deprecated.
-        pretty_print: Whether to indent the output.
+        skipkeys (bool): Whether to silently skip non-string dictionary
+            keys.
+        use_builtin_types (bool): If true, byte strings will be
+            encoded in Base-64 and wrapped in a ``data`` tag; if
+            false, they will be either stored as strings or an
```
anthrotype, line 412:
```suggestion
            false, they will be either stored as ASCII strings or an
```

Lib/fontTools/misc/plistlib.py
```diff
@@ -400,13 +405,25 @@ def totree(
...
+            exception raised if they cannot be represented. Defaults
```
anthrotype, line 413:
```suggestion
            exception raised if they cannot be decoded as such. Defaults
```

Lib/fontTools/misc/plistlib.py
```diff
@@ -400,13 +405,25 @@ def totree(
...
         indent_level (int): Level of indentation when serializing.
 
     Returns: an ``etree`` ``Element`` object.
+
+    Raises:
+        ``TypeError``
+            if non-string dictionary keys are serialized
+            and ``skipkeys`` is false.
+        ``ValueError``
+            if non-representable binary data is present
```
anthrotype, line 425:
```suggestion
            if non-ASCII binary data is present
```

Lib/fontTools/misc/plistlib.py
```diff
@@ -513,13 +533,24 @@ def dump(
...
-        skipkeys: Keys which should be skipped when enumerating.
-        use_builtin_types: Whether to use standard plist types or
-            to wrap values in a Data block. Deprecated.
-        pretty_print: Whether to indent the output.
+        sort_keys (bool): Whether keys of dictionaries should be sorted.
+        skipkeys (bool): Whether to silently skip non-string dictionary
+            keys.
+        use_builtin_types (bool): If true, byte strings will be
+            encoded in Base-64 and wrapped in a ``data`` tag; if
+            false, they will be either stored as strings or an
```
anthrotype, line 541:
```suggestion
            false, they will be either stored as ASCII strings or an
```

same suggestions as above about "represented" -> "decided as ASCII"

PR #1983, opened 2020-06-03
[voltLib] Support writing back ast as VOLT data

khaledhosny:
Adds `__str__()` method to ast classes that writes back VOLT data. Tries to replicate VOLT syntax idiosyncrasies as much as possible for better round-trip conversion.

### DIFF
--- a/Lib/fontTools/voltLib/ast.py
+++ b/Lib/fontTools/voltLib/ast.py
@@ -1,4 +1,14 @@
 from fontTools.voltLib.error import VoltLibError
+from typing import NamedTuple
+
+
+class Pos(NamedTuple):
+    adv: int
+    dx: int
+    dy: int
+    adv_adjust_by: dict
+    dx_adjust_by: dict
+    dy_adjust_by: dict
 
 
 class Element(object):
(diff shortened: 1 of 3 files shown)


Tests/voltLib/parser_test.py
```diff
@@ -473,6 +473,21 @@ def test_lookup_name_starts_with_letter(self):
                 'END_SUBSTITUTION\n'
             ).statements
 
+    def test_lookup_commnets(self):
```
anthrotype, line 476: typo

```suggestion
    def test_lookup_comments(self):
```

Lib/fontTools/voltLib/ast.py
```diff
@@ -83,6 +114,9 @@ def __init__(self, glyph, location=None):
     def glyphSet(self):
         return (self.glyph,)
 
+    def __str__(self):
+        return f' GLYPH "{self.glyph}"'
```
anthrotype, line 118: just curious, why start this and others with empty space? Is it some kind of in-built indentation?

moyogo:
LGTM.
Just wondering why you added spaces making double space.

Lib/fontTools/voltLib/ast.py
khaledhosny: That is to match VOLT spacing, there are different ways to do it but this seemed to be more consistent.

khaledhosny:
> Just wondering why you added spaces making double space.

To match VOLT spacing. I have a punch of VTP files and if I round them through this they show no changes (except for empty `IN_CONTEXT` but I left this out because it might break backward compatibility).

PR #1984, opened 2020-06-03
Don’t calculate whole sets of unicode codepoints

liZe:
`_getUnicodeRangeSets` used to calculate sets containing lots of numbers, only to get intersections between a set and ranges. Creating and manipulating a lot of big sets is both slow and memory consuming.

The function has been replaced by `_getUnicodeRanges`, returning ranges instead of sets. Ranges are both really lightweight and really fast.

Tests on intersectUnicodeRanges are now 3 times faster and save about 130 MB (!) of RAM.

Before: 0.28 seconds, 148 MB RAM
```
time python fonttools/Lib/fontTools/ttLib/tables/O_S_2f_2.py
0.23user 0.05system 0:00.28elapsed 100%CPU (0avgtext+0avgdata 147760maxresident)k
```

After: 0.08 seconds (**-70%**), 14 MB RAM (**-90%**)
```
time python fonttools/Lib/fontTools/ttLib/tables/O_S_2f_2.py
0.08user 0.00system 0:00.08elapsed 100%CPU (0avgtext+0avgdata 13940maxresident)k
```

### DIFF
--- a/Lib/fontTools/ttLib/tables/O_S_2f_2.py
+++ b/Lib/fontTools/ttLib/tables/O_S_2f_2.py
@@ -2,6 +2,7 @@ from fontTools.misc.py23 import *
 from fontTools.misc import sstruct
 from fontTools.misc.textTools import safeEval, num2binary, binary2num
 from fontTools.ttLib.tables import DefaultTable
+import bisect
 import logging
 
 
@@ -475,22 +476,19 @@ OS2_UNICODE_RANGES = (
 )
 
 
-_unicodeRangeSets = []
+_unicodeStarts = []
+_unicodeValues = [None]
 
-def _getUnicodeRangeSets():
-	# build the sets of codepoints for each unicode range bit, and cache result
-	if not _unicodeRangeSets:
-		for bit, blocks in enumerate(OS2_UNICODE_RANGES):
-			rangeset = set()
-			for _, (start, stop) in blocks:
-				rangeset.update(set(range(start, stop+1)))
-			if bit == 57:
-				# The spec says that bit 57 ("Non Plane 0") implies that there's
-				# at least one codepoint beyond the BMP; so I also include all
-				# the non-BMP codepoints here
-				rangeset.update(set(range(0x10000, 0x110000)))
-			_unicodeRangeSets.append(rangeset)
-	return _unicodeRangeSets
+def _getUnicodeRanges():
+	# build the ranges of codepoints for each unicode range bit, and cache result
+	if not _unicodeStarts:
+		unicodeRanges = [
+			(start, (stop, bit)) for bit, blocks in enumerate(OS2_UNICODE_RANGES)
+			for _, (start, stop) in blocks]
+		for start, (stop, bit) in sorted(unicodeRanges):
+			_unicodeStarts.append(start)
+			_unicodeValues.append((stop, bit))
+	return _unicodeStarts, _unicodeValues
 
 
 def intersectUnicodeRanges(unicodes, inverse=False):
@@ -504,15 +502,22 @@ def intersectUnicodeRanges(unicodes, inverse=False):
 	>>> intersectUnicodeRanges([0x0410, 0x1F000]) == {9, 57, 122}
 	True
 	>>> intersectUnicodeRanges([0x0410, 0x1F000], inverse=True) == (
-	...     set(range(123)) - {9, 57, 122})
+	...     set(range(len(OS2_UNICODE_RANGES))) - {9, 57, 122})
 	True
 	"""
 	unicodes = set(unicodes)
-	uniranges = _getUnicodeRangeSets()
-	bits = set([
-		bit for bit, unirange in enumerate(uniranges)
-		if not unirange.isdisjoint(unicodes) ^ inverse])
-	return bits
+	unicodestarts, unicodevalues = _getUnicodeRanges()
+	bits = set()
+	for code in unicodes:
+		stop, bit = unicodevalues[bisect.bisect(unicodestarts, code)]
+		if code <= stop:
+			bits.add(bit)
+	# The spec says that bit 57 ("Non Plane 0") implies that there's
+	# at least one codepoint beyond the BMP; so I also include all
+	# the non-BMP codepoints here
+	if any(0x10000 <= code < 0x110000 for code in unicodes):
+		bits.add(57)
+	return set(range(len(OS2_UNICODE_RANGES))) - bits if inverse else bits
 
 
 if __name__ == "__main__":


anthrotype:
thanks. You only timed the module loading time. It would be interesting to benchmark also the time it takes to run the `intersectUnicodeRanges` function on random sets of unicode codepoints between the two set vs ranges implementations.
If it turns out that building up the sets on module import makes it faster to check for intersection, despite a slower import time and increased memory usage, it may be a reasonable trade off.

anthrotype:
As I was expecting, running `intersectUnicodeRanges` with your approach is much slower. 
This is the benchmark code I used for testing:

```python
import timeit

num_executions = 1000
total_time = timeit.timeit(
    "bits = intersectUnicodeRanges(random.choices(range(0x110000), k=1000))",
    setup="import random; from fontTools.ttLib.tables.O_S_2f_2 import intersectUnicodeRanges",
    number=num_executions,
)

print(total_time / num_executions)
```

With the current set implementation, on my MBP, I get:

```
0.0004779480810000223
```

With your range and for-loop based implementation, I get:

```
0.015389551754999956
```

That is, almost 30 times slower.
So I guess, it all depends on how many times one runs the `intersectUnicodeRanges` function per session, and on the size of the font in terms of number of unicode characters to test for intersection.

I am still not 100% convinced that your solution would be preferable.

liZe:
> You only timed the module loading time.

That’s not exactly true, because the module launches 3 unit tests…

> It would be interesting to benchmark also the time it takes to run the `intersectUnicodeRanges` function on random sets of unicode codepoints between the two set vs ranges implementations.

… but you’re right: having some speed improvement for the set generation + 3 `intersectUnicodeRanges` is not the same as only testing `intersectUnicodeRanges` speed.

> So I guess, it all depends on how many times one runs the `intersectUnicodeRanges` function per session, and on the size of the font in terms of number of unicode characters to test for intersection.

:+1: 

> I am still not 100% convinced that your solution would be preferable.

Storing 130 MB in RAM during the whole execution time of the program is really bad for my use cases, but I get your point. I’ll try to improve my solution.

liZe:
I’ve changed the implementation, here are the results using the script you provided.

Before: 0.82 ms / loop, 147 MB RAM
```python
time python speed.py
0.0008236442100023851
0.90user 0.07system 0:00.98elapsed 99%CPU (0avgtext+0avgdata 146996maxresident)k
```

After: 0.85 ms / loop, 13 MB RAM
```python
ime ~/.local/share/virtualenvs/weasyprint/bin/python speed.py
0.0008471214559976943
0.90user 0.01system 0:00.92elapsed 100%CPU (0avgtext+0avgdata 13296maxresident)k
```

Lib/fontTools/ttLib/tables/O_S_2f_2.py
```diff
@@ -475,22 +476,26 @@ def recalcUnicodeRanges(self, ttFont, pruneOnly=False):
...
+				_unicodeValues.insert(index, bit)
+				index = bisect.bisect_left(_unicodeSteps, stop + 1)
+				_unicodeSteps.insert(index, stop + 1)
+				_unicodeValues.insert(index, None)
+		for i, (step, value) in enumerate(zip(_unicodeSteps, _unicodeValues)):
+			if i and _unicodeSteps[i] == _unicodeSteps[i-1]:
+				_unicodeSteps.pop(i - 1)
+				_unicodeValues.pop(i - 1)
+		_unicodeValues.insert(0, None)
+	return _unicodeSteps, _unicodeValues
```
justvanrossum, line 498: This seems a bit complicated. The ranges don't overlap, so you could make a list like this:
```python
values = []
for bit, val in enumerate(OS2_UNICODE_RANGES):
    for name, (rng_start, rng_end) in val:
        values.append((rng_start, rng_end, bit))
values.sort()
```
And either do a bisect on that directly, or unpack it in three separate lists, and bisect on the range start values.

But yeah, I was just thinking that bisecting may give the best compromise between loading speed, memory usage and lookup performance, so I'm glad you did this!

anthrotype:
Nice! `bisect` did the trick :)
+1 on simplifying the building of the steps list like Just suggested.

justvanrossum:
Another approach would be to special-case the non-BMP codepoints, as that makes up more than 81% of the current set-based data set.

Lib/fontTools/ttLib/tables/O_S_2f_2.py
```diff
@@ -508,11 +513,17 @@ def intersectUnicodeRanges(unicodes, inverse=False):
...
-		bit for bit, unirange in enumerate(uniranges)
-		if not unirange.isdisjoint(unicodes) ^ inverse])
-	return bits
+	unicodesteps, unicodevalues = _getUnicodeRanges()
+	bits = set(unicodevalues[bisect.bisect(unicodesteps, code)] for code in unicodes)
+	for code in unicodes:
+		# The spec says that bit 57 ("Non Plane 0") implies that there's
+		# at least one codepoint beyond the BMP; so I also include all
+		# the non-BMP codepoints here
+		if 0x10000 <= code <= 0x110000:
```
dscorbett, line 522: This should be `0x10000 <= code < 0x110000`.

Lib/fontTools/ttLib/tables/O_S_2f_2.py
liZe: > This seems a bit complicated. The ranges don't overlap, so you could make a list like this:

Of course, you’re right.

> But yeah, I was just thinking that bisecting may give the best compromise between loading speed, memory usage and lookup performance, so I'm glad you did this!

:smile:

Lib/fontTools/ttLib/tables/O_S_2f_2.py
liZe: :+1:

liZe:
New results are equivalent, with less code.
```
time python speed.py
0.000836255980990245
0.90user 0.00system 0:00.91elapsed 100%CPU (0avgtext+0avgdata 13360maxresident)k
```

I’m sure that it could be even better, but I think that it’s good enough for now :wink:. Thanks a lot for the quick review.

Lib/fontTools/ttLib/tables/O_S_2f_2.py
```diff
@@ -475,22 +476,19 @@ def recalcUnicodeRanges(self, ttFont, pruneOnly=False):
...
-				# The spec says that bit 57 ("Non Plane 0") implies that there's
-				# at least one codepoint beyond the BMP; so I also include all
-				# the non-BMP codepoints here
-				rangeset.update(set(range(0x10000, 0x110000)))
-			_unicodeRangeSets.append(rangeset)
-	return _unicodeRangeSets
+def _getUnicodeRanges():
+	# build the ranges of codepoints for each unicode range bit, and cache result
+	if not _unicodeStarts:
+		unicodeRanges = {
```
justvanrossum, line 485: A list of tuples would be good enough. Using a dict only makes it more complicated, and possibly slower. A list can be sorted in-place, unlike a dict.

Lib/fontTools/ttLib/tables/O_S_2f_2.py
```diff
@@ -508,11 +506,20 @@ def intersectUnicodeRanges(unicodes, inverse=False):
...
+		stop, bit = unicodevalues[bisect.bisect(unicodestarts, code)]
+		if code <= stop:
+			bits.add(bit)
+	for code in unicodes:
+		# The spec says that bit 57 ("Non Plane 0") implies that there's
+		# at least one codepoint beyond the BMP; so I also include all
+		# the non-BMP codepoints here
+		if 0x10000 <= code < 0x110000:
+			bits.add(57)
+			break
```
justvanrossum, line 521: Suggestion:
```python
# The spec says ... etc.
if any(0x10000 <= code < 0x110000 for code in unicodes):
    bits.add(57)
```

justvanrossum: But actually, there's no reason to do this test during lookup at all: non-BMP bit 57 should be incorporated into the `_unicodeStarts` and `_unicodeValues` lists.

Lib/fontTools/ttLib/tables/O_S_2f_2.py
```diff
@@ -508,11 +506,20 @@ def intersectUnicodeRanges(unicodes, inverse=False):
...
+	return set(range(123)) - bits if inverse else bits
```
justvanrossum, line 522: I'm a little uneasy with the 123 literal. How about:
```python
   return set(range(len(OS2_UNICODE_RANGES))) - bits if inverse else bits
```

Lib/fontTools/ttLib/tables/O_S_2f_2.py
justvanrossum:
```python
unicodeRanges = [
    (start, stop, bit) for bit, blocks in enumerate(OS2_UNICODE_RANGES)
    for _, (start, stop) in val]
# The spec says ... etc.
unicodeRanges.append((0x10000, 0x10FFFF, 57))
unicodeRanges.sort()
for start, stop, bit in unicodeRanges:
    ... etc.
```

Lib/fontTools/ttLib/tables/O_S_2f_2.py
```diff
@@ -508,11 +506,20 @@ def intersectUnicodeRanges(unicodes, inverse=False):
...
 	unicodes = set(unicodes)
-	uniranges = _getUnicodeRangeSets()
-	bits = set([
-		bit for bit, unirange in enumerate(uniranges)
-		if not unirange.isdisjoint(unicodes) ^ inverse])
-	return bits
+	unicodestarts, unicodevalues = _getUnicodeRanges()
+	bits = set()
+	for code in unicodes:
+		stop, bit = unicodevalues[bisect.bisect(unicodestarts, code)]
```
justvanrossum, line 512: This will raise `TypeError` if `code` is negative (`unicodevalues[0]` being `None`). Normally this is of course not possible – I'm just wondering how defensive this code needs to be.

Lib/fontTools/ttLib/tables/O_S_2f_2.py
liZe: > A list of tuples would be good enough. Using a dict only makes it more complicated, and possibly slower. A list can be sorted in-place, unlike a dict.

You’re right, it’s faster.

> ```python
> unicodeRanges.append((0x10000, 0x10FFFF, 57))
> ```

I don’t think it’s possible, because `(0x10000, 0x10FFFF)` overlaps with other ranges. Did I miss something?

Lib/fontTools/ttLib/tables/O_S_2f_2.py
liZe: > Suggestion: […]

Of course…

Lib/fontTools/ttLib/tables/O_S_2f_2.py
liZe: Then I’ll change the test too :wink:

Lib/fontTools/ttLib/tables/O_S_2f_2.py
liZe: > But actually, there's no reason to do this test during lookup at all: non-BMP bit 57 should be incorporated into the `_unicodeStarts` and `_unicodeValues` lists.

See above.

Lib/fontTools/ttLib/tables/O_S_2f_2.py
liZe: The real question here is: what do we want to do when this happens. If we want to silently ignore negative values, the fastest solution I can think of is to replace `None` by `(None, None)` in `unicodevalues`, and to remove `None` from `bits`. Or we can just ignore them in the loop if we prefer readability (it’s about 3% slower with the 1000 values of the testing script).

anthrotype: I think it's acceptable to raise TypeError on negative codepoint, which is invalid and unlikely in normal circumstances.

Lib/fontTools/ttLib/tables/O_S_2f_2.py
justvanrossum: > I don’t think it’s possible, because (0x10000, 0x10FFFF) overlaps other ranges. Did I miss something?

Oops, I think you're right, sorry.

Lib/fontTools/ttLib/tables/O_S_2f_2.py
liZe: > I think it's acceptable to raise TypeError on negative codepoint, which is invalid and unlikely in normal circumstances.

Then OK, let’s keep this.

Lib/fontTools/ttLib/tables/O_S_2f_2.py
anthrotype: > A list of tuples would be good enough

Right. @liZe will you change that?

liZe: > Right. @liZe will you change that?

Done!

Issue #1985, opened 2020-06-07
[ttLib] STAT AxisNameID must be > 255


justvanrossum:
This is odd. The [spec](https://docs.microsoft.com/en-us/typography/opentype/spec/stat#axis-records) says that the AxisNameID must be greater than 255. fonttools currently reuses nameID's wherever possible, but this can lead to an AxisNameID to be less than 256, and OTS warns about that: https://github.com/khaledhosny/ots/issues/210

justvanrossum:
Fixed in 90c7c7fae1e886f54a04306611f882b3db4c7509

justvanrossum:
Btw. the [spec](https://docs.microsoft.com/en-us/typography/opentype/spec/stat#axis-records) has indeed been updated:

> The axisNameID field provides a name ID that can be used to obtain strings from the 'name' table that can be used to refer to the axis in application user interfaces. A defined name ID can be used if strings for that name ID have the appropriate string for an axis name. Otherwise, name IDs must be greater than 255 and less than 32768.

@anthrotype, I think we should indeed revert 90c7c7f. Would that better be a separate PR, or would it be an idea to to it here while we're at it?

justvanrossum:
ot-sanitizer complained about it, so reverting that may break ots checks. I'll ask over there.

justvanrossum:
It seems ots still warns about it, so I think it's better to postpone fixing it after ots gets updated.

PR #1986, opened 2020-06-07
Fix for #1985: ensure that the AxisNameID in the STAT table is > 255

justvanrossum:
Fix for #1985: ensure that the AxisNameID in the STAT table is not less than 256. This needed an additional argument to the addMultiLingualName() name table method.

### DIFF
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -784,7 +784,7 @@ def _buildAxisRecords(axes, nameTable):
     for axisRecordIndex, axisDict in enumerate(axes):
         axis = ot.AxisRecord()
         axis.AxisTag = axisDict["tag"]
-        axis.AxisNameID = _addName(nameTable, axisDict["name"])
+        axis.AxisNameID = _addName(nameTable, axisDict["name"], 256)
         axis.AxisOrdering = axisDict.get("ordering", axisRecordIndex)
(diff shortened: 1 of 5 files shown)


anthrotype:
Hm. If it's not a hard FAIL, maybe we can just live with the OTS warning and try to relax the STAT table spec to also allow AxisNameIDs <= 255 like AxisValues valueNameIDs are allowed to be?

justvanrossum:
I need to apply the same logic to `_add_fvar` in `varLib`.

justvanrossum:
I can't see a quick way to add an fvar test for this, though.

anthrotype:
> I can't see a quick way to add an fvar test for this, though.

it's ok like that

justvanrossum:
As mentioned in https://github.com/MicrosoftDocs/typography-issues/issues/433#issuecomment-640757610 there's another [restriction for named instances](https://docs.microsoft.com/en-us/typography/opentype/spec/fvar#instancerecord):

> The subfamilyNameID field provides a name ID that can be used to obtain strings from the 'name' table that can be treated as equivalent to name ID 17 (typographic subfamily) strings for the given instance. Values of 2 or 17 can be used; otherwise, values must be greater than 255 and less than 32768. The values 2 or 17 should only be used if the named instance corresponds to the font’s default instance.

I think the chance that this would accidentally be violated is tiny, and not worth checking for explicitly.

PR #1991, opened 2020-06-10
[feaLib] Refactor chained contextual builders

simoncozens:
This is just a simple refactoring which eliminates common code from the chained contextual sub/pos builders. The purpose is to lay the groundwork for generating other formats (format 1 and 2 lookups) on a subtable basis.

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -1299,6 +1299,10 @@ class ChainContextualBuilder(LookupBuilder):
                         self.addLookupRecord_(st, sequenceIndex, l.lookup_index)
         return self.buildLookup_(subtables)
 
+    def add_subtable_break(self, location):
+        self.rules.append((self.SUBTABLE_BREAK_, self.SUBTABLE_BREAK_,
+                           self.SUBTABLE_BREAK_, [self.SUBTABLE_BREAK_]))
+
 
 class ChainContextPosBuilder(ChainContextualBuilder):
     def __init__(self, font, location):
@@ -1333,11 +1337,6 @@ class ChainContextPosBuilder(ChainContextualBuilder):
                 res = lookup
         return res
 
-    def add_subtable_break(self, location):
-        self.rules.append((self.SUBTABLE_BREAK_, self.SUBTABLE_BREAK_,
-                           self.SUBTABLE_BREAK_, [self.SUBTABLE_BREAK_]))
-
-
 class ChainContextSubstBuilder(ChainContextualBuilder):
     def __init__(self, font, location):
         LookupBuilder.__init__(self, font, location, 'GSUB', 6)
@@ -1359,7 +1358,7 @@ class ChainContextSubstBuilder(ChainContextualBuilder):
     def getAlternateGlyphs(self):
         result = {}
         for (_, _, _, lookuplist) in self.rules:
-            if lookuplist == self.SUBTABLE_BREAK_:
+            if lookuplist[0] == self.SUBTABLE_BREAK_:
                 continue
             for lookups in lookuplist:
                 if not isinstance(lookups, list):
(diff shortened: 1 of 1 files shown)


Lib/fontTools/feaLib/builder.py
```diff
@@ -1284,32 +1280,52 @@ def build(self):
...
                 if lookupList is not None:
                     if not isinstance(lookupList, list):
                         # Can happen with synthesised lookups
                         lookupList = [ lookupList ]
                     for l in lookupList:
-                        st.PosCount += 1
                         if l.lookup_index is None:
                             raise FeatureLibError('Missing index of the specified '
-                                'lookup, might be a substitution lookup',
+                                'lookup, might be a %s lookup' % self.other,
```
anthrotype, line 1297: how about using f-strings `f"{self.other}"`

Lib/fontTools/feaLib/builder.py
```diff
@@ -1321,55 +1337,28 @@ def find_chainable_single_pos(self, lookups, glyphs, value):
...
+        rec.LookupListIndex = lookupIndex
+        st.SubstLookupRecord.append(rec)
+        st.SubstCount += 1
 
     def getAlternateGlyphs(self):
         result = {}
-        for (_, _, _, lookuplist) in self.substitutions:
-            if lookuplist == self.SUBTABLE_BREAK_:
+        for (_, _, _, lookuplist) in self.rules:
+            if lookuplist[0] == self.SUBTABLE_BREAK_:
```
anthrotype, line 1361: ah! what's this change? Was it incorrect before?

Lib/fontTools/feaLib/builder.py
```diff
@@ -1384,18 +1373,18 @@ def getAlternateGlyphs(self):
     def find_chainable_single_subst(self, glyphs):
         """Helper for add_single_subst_chained_()"""
         res = None
-        for _, _, _, substitutions in self.substitutions[::-1]:
-            if substitutions == self.SUBTABLE_BREAK_:
+        for _, _, _, rules in self.rules[::-1]:
+            if len(rules) > 0 and rules[0] == self.SUBTABLE_BREAK_:
```
anthrotype, line 1377: same here.. Did we break this subtable break logic when we switched to a list of lookups?

simoncozens: No, nothing to do with that. It's just that we call the "lookup list" attribute `.rules` in the  positioning lookup builder and `.substitutions` in the substitution lookup builder. This makes it harder to share code in a superclass. So I renamed `.substitutions` to `.rules`.

simoncozens: Sorry, that's something different.

simoncozens: But something similar! Previously, when we wanted a subtable break in a *pos* rule, we did this:

        self.rules.append((self.SUBTABLE_BREAK_, self.SUBTABLE_BREAK_,
                           self.SUBTABLE_BREAK_, [self.SUBTABLE_BREAK_]))

and when we had a subtable break in a *sub* rule we did this:
        self.rules.append((self.SUBTABLE_BREAK_, self.SUBTABLE_BREAK_,
                           self.SUBTABLE_BREAK_, self.SUBTABLE_BREAK_))

There was no reason for them to be different, so I made them the same.

anthrotype: i'm a bit confused.. The snippet above looks weird to me (with the last item of the tuple being a 1-item list). Were we really doing that?

simoncozens: https://github.com/fonttools/fonttools/blob/6b4430a8a39774851b707f10d6b403017451893c/Lib/fontTools/feaLib/builder.py#L1326

simoncozens: Happy to go the other way and make the old-*sub* style the default, but I thought it didn't really matter and we expect the lookups list to be a list anyway.

anthrotype: i don't see immediately see why it has to be like that..

But my question was about this change (ignore the `substituions` being renamed to `rules`):

```diff
- if substitutions == self.SUBTABLE_BREAK_:
+ if len(rules) > 0 and rules[0] == self.SUBTABLE_BREAK_:
```

that's very different, and I am wondering where that comes from

simoncozens: This is what I'm talking about. When we had a manual subtable break, ChainContextPosBuilder would push `[self.SUBTABLE_BREAK_]` onto the lookup list, but ChainContextSubstBuilder would push `self.SUBTABLE_BREAK_` onto the lookup list. I fixed this inconsistency (and refactored `add_subtable_break` to a superclass), by making both builders push `[self.SUBTABLE_BREAK_]`.

That meant that code in ChainContextSubstBuilder  which expected to test against `self.SUBTABLE_BREAK_` needed to be updated.

simoncozens:
Closing for now. #1995 should go in first and this PR redone on top of it.

PR #1992, opened 2020-06-10
Contextual pos/sub format optimizer

simoncozens:
(First four commits are from #1991.)

This is a WIP but I would appreciate a couple of people to look over what I have so far, to check if the approach makes sense.

Chaining contextual lookups (GPOS6/GSUB8) can be expressed in three different ways:
* Format 1: Simple glyph contexts
* Format 2: Class-based contexts
* Format 3: Coverage-based contexts

Format 3 is the most flexible and we use it for everything. But formats 1 and 2 can be significantly more compact. Sometimes format 1 is just more compact than a corresponding format 3 lookup, but both format 1 and format 2 can also express *multiple rules* within a single lookup. For example, this code:

```
feature test {
pos A B' 50;
pos C B' 20;
pos D B' 20;
pos E B' 10;
pos B' 50 A;
pos B' 20 C;
pos B' 20 D;
pos B' 10 E;
pos C' F G;
pos C' X Y Z;
pos J K L C' X Y Z;
} test;
```

takes 382 bytes when expressed as 11 Format 3 lookups; but all the rules can be merged into a single Format 1 lookup, taking 188 bytes.

This branch runs over the subtables of a contextual sub/pos lookup and attempts to convert them to the other formats. (It only does format 1 for now, but I have code for creating format 2 rules in #1960.) It then tries to merge consecutive format 1 / format 2 lookups, and determines the most compact representation for the subtable list.

If the approach looks good (and I have a couple of questions, like how to handle `binary_len`), then I'll tidy this up, fix the failing tests, and add format 2 support.

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -1389,11 +1389,12 @@ class ChainContextSubstBuilder(ChainContextualBuilder):
             st.ChainSubRuleSet.append(ruleset)
         return ruleset
 
-    def newChainRule_(self, ruleset):
+    def newChainRule_(self, ruleset, rule = None):
         if not hasattr(ruleset, "ChainSubRuleCount"):
             ruleset.ChainSubRuleCount = 0
             ruleset.ChainSubRule = []
-        rule = otTables.ChainSubRule()
+        if not rule:
+            rule = otTables.ChainSubRule()
         ruleset.ChainSubRuleCount += 1
         ruleset.ChainSubRule.append(rule)
         return rule
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -1299,6 +1299,10 @@ class ChainContextualBuilder(LookupBuilder):
                         self.addLookupRecord_(st, sequenceIndex, l.lookup_index)
         return self.buildLookup_(subtables)
 
+    def add_subtable_break(self, location):
+        self.rules.append((self.SUBTABLE_BREAK_, self.SUBTABLE_BREAK_,
+                           self.SUBTABLE_BREAK_, [self.SUBTABLE_BREAK_]))
+
 
 class ChainContextPosBuilder(ChainContextualBuilder):
     def __init__(self, font, location):
@@ -1333,11 +1337,6 @@ class ChainContextPosBuilder(ChainContextualBuilder):
                 res = lookup
         return res
 
-    def add_subtable_break(self, location):
-        self.rules.append((self.SUBTABLE_BREAK_, self.SUBTABLE_BREAK_,
-                           self.SUBTABLE_BREAK_, [self.SUBTABLE_BREAK_]))
-
-
 class ChainContextSubstBuilder(ChainContextualBuilder):
     def __init__(self, font, location):
         LookupBuilder.__init__(self, font, location, 'GSUB', 6)
@@ -1359,7 +1358,7 @@ class ChainContextSubstBuilder(ChainContextualBuilder):
     def getAlternateGlyphs(self):
         result = {}
         for (_, _, _, lookuplist) in self.rules:
-            if lookuplist == self.SUBTABLE_BREAK_:
+            if lookuplist[0] == self.SUBTABLE_BREAK_:
                 continue
             for lookups in lookuplist:
                 if not isinstance(lookups, list):
@@ -1375,23 +1374,13 @@ class ChainContextSubstBuilder(ChainContextualBuilder):
         """Helper for add_single_subst_chained_()"""
         res = None
         for _, _, _, rules in self.rules[::-1]:
-            if rules == self.SUBTABLE_BREAK_:
+            if len(rules) > 0 and rules[0] == self.SUBTABLE_BREAK_:
                 return res
             for sub in rules:
                 if (isinstance(sub, SingleSubstBuilder) and
                         not any(g in glyphs for g in sub.mapping.keys())):
                     res = sub
         return res
-
-    def add_subtable_break(self, location):
-        self.rules.append(
-            (
-                self.SUBTABLE_BREAK_,
-                self.SUBTABLE_BREAK_,
-                self.SUBTABLE_BREAK_,
-                self.SUBTABLE_BREAK_,
-            )
-        )
 
     @property
     def other(self):
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -1294,7 +1294,7 @@ class ChainContextualBuilder(LookupBuilder):
                     for l in lookupList:
                         if l.lookup_index is None:
                             raise FeatureLibError('Missing index of the specified '
-                                'lookup, might be a %s lookup' % self.other,
+                                f'lookup, might be a {self.other} lookup',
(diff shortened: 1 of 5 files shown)


Lib/fontTools/feaLib/builder.py
```diff
@@ -1270,46 +1270,91 @@ def add_subtable_break(self, location):
...
                 if lookupList is not None:
                     if not isinstance(lookupList, list):
                         # Can happen with synthesised lookups
                         lookupList = [ lookupList ]
                     for l in lookupList:
-                        st.PosCount += 1
                         if l.lookup_index is None:
                             raise FeatureLibError('Missing index of the specified '
-                                'lookup, might be a substitution lookup',
+                                f'lookup, might be a {self.other} lookup',
```
khaledhosny, line 1301: `other` is an odd property to have. I’d be inclined to have a `kind` property and rephrase the messge to be `might not be a {self.kind} lookup` (that is arguably a better way to phrase it than my existing phrasing).

Lib/fontTools/feaLib/builder.py
```diff
@@ -1270,46 +1270,91 @@ def add_subtable_break(self, location):
...
+            st.PosLookupRecord = []
+        rec = otTables.PosLookupRecord()
+        rec.SequenceIndex = sequenceIndex
+        rec.LookupListIndex = lookupIndex
+        st.PosLookupRecord.append(rec)
+        st.PosCount += 1
+
+    @property
+    def other(self):
+        return "substitution"
```
khaledhosny, line 1356: I’d move this right below `__init__` so it is more visible.

Lib/fontTools/feaLib/builder.py
```diff
@@ -1270,46 +1270,91 @@ def add_subtable_break(self, location):
...
+        LookupBuilder.__init__(self, font, location, 'GPOS', 8)
+        self.rules = []  # (prefix, input, suffix, lookups)
+
+    def newSubtable_(self):
+        st = otTables.ChainContextPos()
+        st.PosCount = 0
+        st.PosLookupRecord = []
+        return st
+
+    def newChainRuleSet_(self, st, index=None):
```
khaledhosny, line 1322: Nitpick, but `st` is a really obscure variable name (I keep reading it statement), now might be a good time to spell it `subtable`.

Lib/fontTools/feaLib/builder.py
```diff
@@ -1270,46 +1270,91 @@ def add_subtable_break(self, location):
...
+        if not hasattr(st, "ChainPosRuleSetCount"):
+            st.ChainPosRuleSetCount = 0
+            st.ChainPosRuleSet = []
+        ruleset = otTables.ChainPosRuleSet()
+        st.ChainPosRuleSetCount += 1
+        if index:
```
khaledhosny, line 1328:
```suggestion
        if index is not None:
```
In case `index` can be zero (or `if index is None` and invert the condition).

Lib/fontTools/feaLib/builder.py
```diff
@@ -1270,46 +1270,91 @@ def add_subtable_break(self, location):
...
+            st.ChainPosRuleSet.insert(index, ruleset)
+        else:
+            st.ChainPosRuleSet.append(ruleset)
+        return ruleset
+
+    def newChainRule_(self, ruleset, rule = None):
```
khaledhosny, line 1334:
```suggestion
    def newChainRule_(self, ruleset, rule=None):
```

Lib/fontTools/feaLib/builder.py
```diff
@@ -1270,46 +1270,91 @@ def add_subtable_break(self, location):
...
+        if not hasattr(ruleset, "ChainPosRuleCount"):
+            ruleset.ChainPosRuleCount = 0
+            ruleset.ChainPosRule = []
+        if not rule:
```
khaledhosny, line 1338:
```suggestion
        if rule is None:
```

Lib/fontTools/feaLib/builder.py
```diff
@@ -1321,55 +1366,53 @@ def find_chainable_single_pos(self, lookups, glyphs, value):
...
+            st.ChainSubRuleSet = []
+        ruleset = otTables.ChainSubRuleSet()
+        st.ChainSubRuleSetCount += 1
+        if index:
+            st.ChainSubRuleSet.insert(index, ruleset)
+        else:
+            st.ChainSubRuleSet.append(ruleset)
+        return ruleset
+
+    def newChainRule_(self, ruleset, rule = None):
```
khaledhosny, line 1392:
```suggestion
    def newChainRule_(self, ruleset, rule=None):
```

Lib/fontTools/otlLib/builder.py
```diff
@@ -20,7 +20,7 @@ def buildCoverage(glyphs, glyphMap):
 LOOKUP_FLAG_USE_MARK_FILTERING_SET = 0x0010
 
 
-def buildLookup(subtables, flags=0, markFilterSet=None):
+def buildLookup(subtables, flags=0, markFilterSet=None, optimize=None, builder=None):
```
khaledhosny, line 23: Hmm, I’m not really sure it is a good idea (layering wise) to pass feaLib builder to otlLib builder to use it, if anything it should be feaLib using otlLib API not the other way around.

Lib/fontTools/otlLib/builder.py
```diff
@@ -46,6 +48,162 @@ def buildLookup(subtables, flags=0, markFilterSet=None):
              "LOOKUP_FLAG_USE_MARK_FILTERING_SET; flags=0x%04x" % flags)
     return self
 
+def optimizeChainingFormats(subtables, builder):
```
khaledhosny, line 51: This assume builder is not None, but call site above does not check that.

Lib/fontTools/otlLib/builder.py
```diff
@@ -46,6 +48,162 @@ def buildLookup(subtables, flags=0, markFilterSet=None):
...
+
+    def binary_len(st):
```
khaledhosny, line 53: This file seems to be otherwise consistent in its use of camlCase function names.

Lib/fontTools/otlLib/builder.py
simoncozens: I did think about this, but then it seemed like a much bigger change. feaLib.builder currently contains several general classes for building OT lookup objects, and those classes actually have very little to do with FEA - they store a location attribute but that's it. Arguably they should be moved to otlLib where they would be available for other code which builds OT lookups - people writing their own feature compilers, and of course, things like this optimizer.

How does that idea sound? @brawer?

anthrotype: > Arguably they should be moved to otlLib where they would be available for other code which builds OT lookups

indeed, that was the plan. We should complete that.

simoncozens: OK. I'm going to do *that*, then I'm going to do #1991, then I'm going to do this.

brawer:
> general classes for building OT lookup objects [...] should be moved to otlLib where they would be available for other code which builds OT lookups - people writing their own feature compilers, and of course, things like this optimizer. How does that idea sound? @brawer?

Sounds very reasonable to me. Thanks for doing this refactoring!

PR #1995, opened 2020-06-15
Move builders to otlib

simoncozens:
As discussed in #1992, this moves the lookup-builder classes from feaLib to otlLib. It also spins off `ClassDefBuilder` to its own file, and documents otlLib. Note this does not yet contain any of the refactorings from #1991.

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -5,6 +5,7 @@ from fontTools.feaLib.error import FeatureLibError
 from fontTools.feaLib.parser import Parser
 from fontTools.feaLib.ast import FeatureFile
 from fontTools.otlLib import builder as otl
+from fontTools.otlLib.lookupBuilders import *
 from fontTools.otlLib.maxContextCalc import maxCtxFont
 from fontTools.ttLib import newTable, getTableModule
 from fontTools.ttLib.tables import otBase, otTables
@@ -1046,14 +1047,18 @@ class Builder(object):
     def add_class_pair_pos(self, location, glyphclass1, value1,
                            glyphclass2, value2):
         lookup = self.get_lookup_(location, PairPosBuilder)
-        lookup.addClassPair(location, glyphclass1, value1, glyphclass2, value2)
+        otValue1 = makeOpenTypeValueRecord(value1, pairPosContext=True)
+        otValue2 = makeOpenTypeValueRecord(value2, pairPosContext=True)
+        lookup.addClassPair(location, glyphclass1, otValue1, glyphclass2, otValue2)
 
     def add_subtable_break(self, location):
         self.cur_lookup_.add_subtable_break(location)
 
     def add_specific_pair_pos(self, location, glyph1, value1, glyph2, value2):
         lookup = self.get_lookup_(location, PairPosBuilder)
-        lookup.addGlyphPair(location, glyph1, value1, glyph2, value2)
+        otValue1 = makeOpenTypeValueRecord(value1, pairPosContext = True)
+        otValue2 = makeOpenTypeValueRecord(value2, pairPosContext = True)
+        lookup.addGlyphPair(location, glyph1, otValue1, glyph2, otValue2)
 
     def add_single_pos(self, location, prefix, suffix, pos, forceChain):
         if prefix or suffix or forceChain:
@@ -1061,8 +1066,9 @@ class Builder(object):
         else:
             lookup = self.get_lookup_(location, SinglePosBuilder)
             for glyphs, value in pos:
+                otValue = makeOpenTypeValueRecord(value, pairPosContext = False)
                 for glyph in glyphs:
-                    lookup.add_pos(location, glyph, value)
+                    lookup.add_pos(location, glyph, otValue)
 
     def add_single_pos_chained_(self, location, prefix, suffix, pos):
         # https://github.com/fonttools/fonttools/issues/514
@@ -1075,13 +1081,13 @@ class Builder(object):
             if value is None:
                 subs.append(None)
                 continue
-            otValue, _ = makeOpenTypeValueRecord(value, pairPosContext=False)
+            otValue = makeOpenTypeValueRecord(value, pairPosContext=False)
             sub = chain.find_chainable_single_pos(targets, glyphs, otValue)
             if sub is None:
                 sub = self.get_chained_lookup_(location, SinglePosBuilder)
                 targets.append(sub)
             for glyph in glyphs:
-                sub.add_pos(location, glyph, value)
+                sub.add_pos(location, glyph, otValue)
             subs.append(sub)
         assert len(pos) == len(subs), (pos, subs)
         chain.rules.append(
@@ -1132,7 +1138,6 @@ class Builder(object):
 
 
 def makeOpenTypeAnchor(anchor):
-    """ast.Anchor --> otTables.Anchor"""
     if anchor is None:
         return None
     deviceX, deviceY = None, None
@@ -1154,7 +1159,7 @@ _VALUEREC_ATTRS = {
 def makeOpenTypeValueRecord(v, pairPosContext):
     """ast.ValueRecord --> (otBase.ValueRecord, int ValueFormat)"""
     if not v:
-        return None, 0
+        return None
 
     vr = {}
     for astName, (otName, isDevice) in _VALUEREC_ATTRS.items():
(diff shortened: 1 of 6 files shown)


Lib/fontTools/otlLib/builder.py
```diff
@@ -324,6 +700,44 @@ def _getValueFormat(f, values, i):
...
+    required.
+
+    Example::
+
+        pairs = {}
+
+        pairs[(
+            [ "K", "X" ],
+            [ "W", "V" ]
+        )] = ( buildValue(xAdvance=+5), buildValue() )
```
dscorbett, line 724: `buildValue` takes a dictionary, not keyword arguments.

Lib/fontTools/otlLib/builder.py
```diff
@@ -21,6 +50,34 @@ def buildCoverage(glyphs, glyphMap):
...
+    the following constants to express lookup flags:
+
+    - ``LOOKUP_FLAG_RIGHT_TO_LEFT``
+    - ``LOOKUP_FLAG_IGNORE_BASE_GLYPHS``
+    - ``LOOKUP_FLAG_IGNORE_LIGATURES``
+    - ``LOOKUP_FLAG_IGNORE_MARKS``
+    - ``LOOKUP_FLAG_USE_MARK_FILTERING_SET``
+
+    Args:
+        subtables (list): A list of layout subtable objects. (e.g.
```
dscorbett, line 69: `subtables` can also be `None`, not just a list.

Lib/fontTools/otlLib/builder.py
```diff
@@ -195,6 +418,35 @@ def buildDevice(deltas):
 
 
 def buildLigatureArray(ligs, numMarkClasses, glyphMap):
+    """Builds a LigatureArray subtable.
+
+    As part of building a mark-to-ligature lookup, you will need to define
+    the set of anchors (for each mark class) on each component of the ligature
+    where marks can be attached. For example, for an Arabic divine name ligature
+    (lam lam heh), you may want to specify mark attachment positioning for
+    superior marks (fatah, etc.) and inferior marks (kasra, etc.) on each glyph
```
dscorbett, line 427: “fatah” should be “fatha”.

Lib/fontTools/otlLib/builder.py
```diff
@@ -207,15 +459,41 @@ def buildLigatureArray(ligs, numMarkClasses, glyphMap):
...
+    """Builds a mark array subtable.
+
+    As part of building mark-to-* positioning rules, you will need to define
+    a ``MarkArray`` subtable, which "defines the class and the anchor point
+    for a mark glyph." This function builds the mark array subtable.
+
+    Example::
+
+        mark = {
+            "acute": (0: buildAnchor(300,712)),
```
dscorbett, line 481: The colon in the tuple should be a comma.

Lib/fontTools/otlLib/builder.py
```diff
@@ -227,11 +505,40 @@ def buildMarkArray(marks, glyphMap):
...
+        bases = {"a": {0: a3, 1: a5}, "b": {0: a4, 1: a5}}
+        markbaseposes = buildMarkBasePos(marks, bases, font.getReverseGlyphMap())
+
+    Args:
+        marks (dict): A mapping specifying anchors for glyphs; the keys being
+            glyph names, and the values being a tuple of mark class number and
+            an ``otTables.Anchor`` object representing the mark's attachment
+            point. (See :func:`buildMarkArray`.)
+        bases (dict): A mapping specifying anchors for glyphs; the keys being
+            glyph names, and the values being dictionaries mapping mark class ID
```
dscorbett, line 534: Sometimes you say “mapping” and sometimes “dictionary”. Why?

Lib/fontTools/otlLib/builder.py
```diff
@@ -265,11 +586,50 @@ def buildMarkBasePosSubtable(marks, bases, glyphMap):
...
+        markligposes = buildMarkLigPos(marks, ligs,
+            font.getReverseGlyphMap())
+
+    Args:
+        marks (dict): A mapping specifying anchors for glyphs; the keys being
+            glyph names, and the values being a tuple of mark class number and
+            an ``otTables.Anchor`` object representing the mark's attachment
+            point. (See :func:`buildMarkArray`.)
+        ligs (dict): A mapping of ligature names to an array of arrays: for
+            each component glyph in the ligature, an array of anchors.
```
dscorbett, line 625: The example above shows that values of `ligs` are arrays of mappings, not arrays of arrays.

“Array” isn’t a Python thing: it should be “sequence” or “list” depending on what you mean.

Lib/fontTools/otlLib/builder.py
```diff
@@ -522,7 +1075,20 @@ def buildValue(value):
 # GDEF
 
 def buildAttachList(attachPoints, glyphMap):
-    """{"glyphName": [4, 23]} --> otTables.AttachList, or None"""
+    """Builds an AttachList subtable.
+
+    A GDEF table may contain an Attachment Point List table (``AttachList``)
```
dscorbett, line 1080: The formatting of table names is inconsistent throughout the documentation: sometimes it is like “AttachList” and sometimes like “`AttachList`”. FWIW the OpenType spec formats them like “AttachList”.

Lib/fontTools/otlLib/builder.py
```diff
@@ -544,23 +1111,53 @@ def buildAttachPoint(points):
...
+    the ligature to represent an insertion point. The insertion positions may
+    be specified either by coordinate or by contour point.
+
+    Example::
+
+        coords = {
+            "f_f_i": [300, 600] # f|fi cursor at 300 units, ff|i cursor at 600.
+        }
+        points = {
+            "c_t": [28] # c|t cursor appears at co-ordinate of contour point 28.
```
dscorbett, line 1146: Everywhere else, this is spelled “coordinate”.

Lib/fontTools/otlLib/builder.py
```diff
@@ -544,23 +1111,53 @@ def buildAttachPoint(points):
...
+        }
+        ligcaretlist = buildLigCaretList(coords, points, font.getReverseGlyphMap())
+
+    Args:
+        coords: A mapping between glyph names and a list of X-coordinates for
```
dscorbett, line 1151: The spec says “Caret coordinates are either X or Y values, depending upon the text direction.” Does this function only support X values?

Lib/fontTools/otlLib/lookupBuilders.py
```diff
@@ -0,0 +1,988 @@
...
+    Users are expected to manually add alternate glyph substitutions to
+    the ``alternates`` attribute after the object has been initialized,
+    e.g.::
+
+        builder.alternates["A"] = ["A.alt1", "A.alt2"]
+
+    Attributes:
+        font (``fontTools.TTLib.TTFont``): A font object.
+        location: A string or tuple representing the location in the original
+            source which produced this lookup.
```
dscorbett, line 110: If it’s a string, does it need to be in any particular format? If it’s a tuple, what does each element represent?

Location arguments in `feaLib` can also be `None`; is that true here too?

Lib/fontTools/otlLib/lookupBuilders.py
```diff
@@ -0,0 +1,988 @@
...
+        alternates: An ordered dictionary of alternates, mapping glyph names
+            to a list of names of alternates.
+        lookupflag: (int) The lookup's flag
```
dscorbett, line 113: The parenthesized type should precede the colon. This occurs throughout the PR.

Lib/fontTools/otlLib/lookupBuilders.py
```diff
@@ -0,0 +1,988 @@
...
+        prefix = [ ["a", "e", "n"] ]
+        suffix = []
+        mapping = { "d": "d.alt" }
+        builder.substitutions.append( (prefix, suffix, mapping) )
+
+    Attributes:
+        font (``fontTools.TTLib.TTFont``): A font object.
+        location: A string or tuple representing the location in the original
+            source which produced this lookup.
+        substitutions: A three-element triple consisting of a prefix sequence,
```
dscorbett, line 676: “three-element triple” is redundant.

Lib/fontTools/otlLib/lookupBuilders.py
```diff
@@ -0,0 +1,988 @@
...
+
+    def getAlternateGlyphs(self):
+        return {glyph: set([repl]) for glyph, repl in self.mapping.items()}
+
+    def add_subtable_break(self, location):
+        self.mapping[(self.SUBTABLE_BREAK_, location)] = self.SUBTABLE_BREAK_
+
+
+class ClassPairPosSubtableBuilder(object):
+    """Builds class-based Pair Positioning (GPOS2, format 2) subtables.
```
dscorbett, line 763: Other similar docstrings leave out the comma, i.e. “GPOS2 format 2”.

Lib/fontTools/otlLib/lookupBuilders.py
```diff
@@ -0,0 +1,988 @@
...
+
+    Note that this does *not* build a GPOS2 ``otTables.Lookup`` directly,
+    but builds a list of ``otTables.PairPos`` subtables. It is used by the
+    :class:`PairPosBuilder` below.
+
+    Attributes:
+        builder (PairPosBuilder): A pair positioning lookup builder.
+        valueFormat1: Unused.
+        valueFormat2: Unused.
```
dscorbett, line 772: `valueFormat1` and `valueFormat2` are parameters, not attributes. If you try to access attributes by those names you’ll get an `AttributeError`.

Lib/fontTools/otlLib/lookupBuilders.py
simoncozens: Once we move all this to otlLib, we are no longer in FEA-land, and this code is open for any developers of feature compilers to use. If your feature compiler is graphical, there may not even be an underlying text file to parse, but there is still a "location" which produced the rule. So I want to keep `location` as generic as possible.

dscorbett: Sounds good. Do there need to be any restrictions on the type of `location` at all?

Lib/fontTools/otlLib/builder.py
simoncozens: So the `buildLigatureArray` example is wrong, which led me astray. I need to fix both the docs *and* that example.

Lib/fontTools/otlLib/lookupBuilders.py
simoncozens: Yes, because it gets stringified if an OTLLibError is raised. I made OTLLibError cope with strings and iterables.

Lib/fontTools/otlLib/builder.py
```diff
@@ -195,6 +419,36 @@ def buildDevice(deltas):
...
+            "lam-lam-heh": [
+                { 0: superiorAnchor1, 1: inferiorAnchor1 }, # attach points for lam1
+                { 0: superiorAnchor2, 1: inferiorAnchor2 }, # attach points for lam2
+                { 0: superiorAnchor3, 1: inferiorAnchor3 }, # attach points for heh
+            ]
+        }, 2, font.getReverseGlyphMap())
+
+    Args:
+        ligs (dict): A mapping of ligature names to an array of dictionaries:
+            for each component glyph in the ligature, an dictionary mapping
```
dscorbett, line 443: “an dictionary” should be “a dictionary”, here and elsewhere.

anthrotype:
would it be possible to split this PR up in two, one with the code refactorings and another with the documentation fixes? A smaller, more targeted PRs are easier to review.

simoncozens:
OK. Here's the first part.

Lib/fontTools/feaLib/builder.py
```diff
@@ -5,6 +5,7 @@
 from fontTools.feaLib.parser import Parser
 from fontTools.feaLib.ast import FeatureFile
 from fontTools.otlLib import builder as otl
+from fontTools.otlLib.lookupBuilders import *
```
anthrotype, line 8: I am not a fan of star imports. Ideally I would like to see the imported symbols explicitly mentioned here.

Lib/fontTools/feaLib/builder.py
```diff
@@ -1046,23 +1047,28 @@ def add_mark_mark_pos(self, location, baseMarks, marks):
...
+        otValue2 = makeOpenTypeValueRecord(value2, pairPosContext=True)
+        lookup.addClassPair(location, glyphclass1, otValue1, glyphclass2, otValue2)
 
     def add_subtable_break(self, location):
         self.cur_lookup_.add_subtable_break(location)
 
     def add_specific_pair_pos(self, location, glyph1, value1, glyph2, value2):
         lookup = self.get_lookup_(location, PairPosBuilder)
-        lookup.addGlyphPair(location, glyph1, value1, glyph2, value2)
+        otValue1 = makeOpenTypeValueRecord(value1, pairPosContext = True)
```
anthrotype, line 1059: pls, no spaces around equal signs in function parameters. I would say "run black autoformatter" but actually I only want it to reformat the modified lines (whereas black reformats everything in the file).

Lib/fontTools/feaLib/builder.py
```diff
@@ -1154,7 +1159,7 @@ def makeOpenTypeAnchor(anchor):
 def makeOpenTypeValueRecord(v, pairPosContext):
     """ast.ValueRecord --> (otBase.ValueRecord, int ValueFormat)"""
```
anthrotype, line 1160: the docstring doesn't match the return value (but you probably know that, and deliberately split the doc fixes to a different PRs). Ideally each PR is coherent within itself, so it's easier to roll it back in case.

Lib/fontTools/otlLib/classDefBuilder.py
```diff
@@ -0,0 +1,60 @@
...
+    def build(self):
+        glyphClasses = {}
+        for classID, glyphs in enumerate(self.classes()):
+            if classID == 0:
+                continue
+            for glyph in glyphs:
+                glyphClasses[glyph] = classID
+        classDef = ot.ClassDef()
+        classDef.classDefs = glyphClasses
+        return classDef
```
anthrotype, line 60: i like to have newline at end of file (my text editor enforces that).

Lib/fontTools/otlLib/builder.py
```diff
@@ -3,6 +3,9 @@
 from fontTools import ttLib
 from fontTools.ttLib.tables import otTables as ot
 from fontTools.ttLib.tables.otBase import ValueRecord, valueRecordFormatDict
+from fontTools.ttLib.tables import otBase, otTables
+from fontTools.otlLib.error import OTLLibError
+from fontTools.otlLib.classDefBuilder import ClassDefBuilder
```
anthrotype, line 8: why the proliferation of submodules? can't ClassDefBuilder live inside otlLib.builder? It's only one class.

Tests/feaLib/builder_test.py
```diff
@@ -259,7 +260,7 @@ def test_singleSubst_multipleSubstitutionsForSameGlyph(self):
 
     def test_singlePos_redefinition(self):
         self.assertRaisesRegex(
-            FeatureLibError,
+            OTLLibError,
```
anthrotype, line 263: I find it problematic that we are changing the class of the raised exceptions from FeatureLibError to OTLLibError (which are unrelated to each other). Now all clients (e.g. ufo2ft) will have to catch both, instead of just FeatureLibError. I think feaLib should catch OTLLibError and re-raise it as FeatureLibError to not break current API.

Lib/fontTools/otlLib/builder.py
simoncozens: Why *not* a proliferation of submodules? ClassDefBuilder is currently buried in the middle of a large and very mixed collection of functions. It's a sufficiently self-contained bit of functionality that moving it to its own file helps with code organisation.

simoncozens:
I'm not sure I can do this. You want

anthrotype:
@simoncozens why closing?

anthrotype:
Ah, I just saw https://github.com/fonttools/fonttools/pull/1999, thank you

PR #1999, opened 2020-06-16
Move feature builders to otlLib, take 2

simoncozens:
This time done in the smallest possible steps. Not yet complete.


simoncozens:
At this point, tests (should) pass. Future commits are about tidying up.

simoncozens:
Right. If these tests pass (which I think they should), then I think I have done all I want to do in this PR. Documentation and refactoring to follow if this gets in.

Lib/fontTools/otlLib/error.py
```diff
@@ -0,0 +1,14 @@
+
+
+class OpentypeLibError(Exception):
```
anthrotype, line 3: I suggest "OpenTypeLibError" with the capital "T"

Lib/fontTools/feaLib/builder.py
```diff
@@ -576,7 +577,11 @@ def buildLookups_(self, tag):
                 continue
             lookup.lookup_index = len(lookups)
             lookups.append(lookup)
-        return [l.build() for l in lookups]
+        try:
+            otLookups = [l.build() for l in lookups]
+        except OpentypeLibError as e:
+            raise FeatureLibError("Failed to build",()) from e
```
anthrotype, line 583: I think `location=None` is better than an empty () tuple here. Also, space after comma.

```suggestion
            raise FeatureLibError("Failed to build", location=None) from e
```

Lib/fontTools/feaLib/error.py
```diff
@@ -6,6 +6,8 @@ def __init__(self, message, location):
         self.location = location
 
     def __str__(self):
+        if self.__cause__:
+            return self.__cause__.__str__()
```
anthrotype, line 10: Hm I don't like that, this will lose/ignore the location. We need to maintain that when we re-raise from OpenTypeLibError. I suggest you do `raise FeatureLibError(str(e), location) from e` when the message that is specific to a FeatureLibError is unimportant and you want to reuse the text from the underlying OpenTypeLibError that's being handled.

Lib/fontTools/otlLib/builder.py
```diff
@@ -589,16 +587,15 @@ def build(self):
                 if builder is not None:
                     builder.addSubtableBreak()
                 continue
-            val1, valFormat1 = makeOpenTypeValueRecord(
-                value1, pairPosContext=True)
-            val2, valFormat2 = makeOpenTypeValueRecord(
-                value2, pairPosContext=True)
+            valFormat1, valFormat2 = 0, 0
+            if value1: valFormat1 = value1.getFormat()
```
anthrotype, line 568: i don't like if statements in one line, even if very short. I prefer consistency.

Lib/fontTools/feaLib/error.py
anthrotype: I just noticed that this OpenTypeLibError is basically identical to the original FeatureLibError and also has a location attribute. Similarly the lookup builders that were moved to otlLib also still keep a reference to their location in the feature file. I would argue that the whole concept of `location` (line and column inside a text file) is alien to otlLib.builder, and only belongs to feaLib proper. I would prefer if otlLib lookup builders were agnostic of location, and the responsibility of providing contextual info was done by feaLib, when catching OpenTypeLibError.

simoncozens: I understand this, but I think it's going to be very difficult in practice. The whole point of having a "builder" class is that you parse rules and add them to a builder object, and then you build all the lookups at the end. If error only arises when the lookup is built, and the builder object isn't keeping track of the location that gave rise to the error, *what is*?

I would say that the concept of location *is* important to a builder, but `location` doesn't just mean a line-and-column-in-a-text-file. We still need to know where a lookup originally "came from" even if the user of otlLib is a graphical font editor. So what I tried to do with my previous PR was to turn `location` into a generic userdata attribute and leave its interpretation up to the user class.

anthrotype: > turn location into a generic userdata attribute and leave its interpretation up to the user class.

I like this idea actually.
If the otlLib classes do not manipulate this passed-in `location` but simply use as prefix for the exception error messages, the caller that uses otlLib builders can pass in whatever they want as location, it only needs to be something that pretty-print itself nicely when `str(location)`.
So instead of passing in a tuple as location parameter from feaLib to otlLib, feaLib should pre-format the tuple as a colon-separated str, and then otlLib just interpolate that into the error messages. Let's do that.

simoncozens:
Go again!

simoncozens:
@anthrotype, any objections to a merge? Or anyone else we should check with first?

Lib/fontTools/feaLib/error.py
anthrotype: we do we need to do `if self.__cause__: return self.__cause__.__str__()`? I don't like that we mess with python built-ibn exception chaning mechanism. If we want FeatureLibError to print the same message as the previous exception that it is handling, then we do that e.g. `raise FeatureLibError(str(e), location) from e`

anthrotype:
> any objections to a merge?

I think I'm fine with the merge. Only want reassurance that all the places where a possible OpenTypeLibError can be raised, the feaLib module is re-raising the appropriate FeatureLibError and we don't let OpenTypeLibError "bleed" outside. I don't want clients to need to change to handle OpenTypeLibError themselves.

Lib/fontTools/feaLib/builder.py
```diff
@@ -560,7 +583,11 @@ def buildLookups_(self, tag):
                 continue
             lookup.lookup_index = len(lookups)
             lookups.append(lookup)
-        return [l.build() for l in lookups]
+        try:
+            otLookups = [l.build() for l in lookups]
+        except OpenTypeLibError as e:
+            raise FeatureLibError("Failed to build", location=None) from e
```
anthrotype, line 589:
```suggestion
            raise FeatureLibError(f"Failed to build {tag} lookups", location=None) from e
```

Lib/fontTools/feaLib/builder.py
```diff
@@ -1046,23 +1073,31 @@ def add_mark_mark_pos(self, location, baseMarks, marks):
...
         else:
             lookup = self.get_lookup_(location, SinglePosBuilder)
             for glyphs, value in pos:
+                otValueRecord = makeOpenTypeValueRecord(value, pairPosContext=False)
                 for glyph in glyphs:
-                    lookup.add_pos(location, glyph, value)
+                    try:
+                        lookup.add_pos(location, glyph, otValueRecord)
+                    except OpenTypeLibError as e:
+                        raise FeatureLibError("Failed to build", location) from e
```
anthrotype, line 1100: again here, "Failed to build" sounds quite generic, may be better to add more context.
Or is it that you are relying on that `if self.__context__: ...` clause in `FeatureLibError.__str__` method, to basically ignore this generic message and use the one from the underlying exception?
I don't like this approach, as I pointed out earlier.

Lib/fontTools/otlLib/builder.py
```diff
@@ -47,6 +53,568 @@ def buildLookup(subtables, flags=0, markFilterSet=None):
...
+        key = (glyph1, glyph2)
+        oldValue = self.glyphPairs.get(key, None)
+        if oldValue is not None:
+            # the Feature File spec explicitly allows specific pairs generated
+            # by an 'enum' rule to be overridden by preceding single pairs
+            otherLoc = self.locations[key]
+            log.debug(
+                'Already defined position for pair %s %s at %s:%d:%d; '
+                'choosing the first value',
+                glyph1, glyph2, otherLoc[0], otherLoc[1], otherLoc[2])
```
anthrotype, line 545: here an otlLib builder is treating location as a tuple and formatting it as file:line:column. It should not be doing that, as you made it an opaque string prefix in your latest commit. Pls make sure this and other places where this might occur are fixed accordingly (I haven't checked thoroughly)

Lib/fontTools/feaLib/builder.py
```diff
@@ -180,8 +197,14 @@ def build(self, tables=None):
...
 
+    def format_location_(self, location):
+        if not location:
+            return None
+        path, line, column = location
+        return f"{path}:{line}:{column}"
+
     def get_chained_lookup_(self, location, builder_class):
-        result = builder_class(self.font, location)
+        result = builder_class(self.font, self.format_location_(location))
```
anthrotype, line 207: now that I look at this, I am thinking it is kind of wasteful to pre-format the location tuple as `file:line:col` string, considering it won't be used most of the time unless an error occurs..
On the other hand, we wouldn't want to hard-code the meaning of "location"  to a text file with lines/columns for otlLib builders.
How about the `OpenTypeLibError.__str__()` checks if `location` is a tuple, then it does `":".join(str(s) for s in location)` (without assuming it's going to be three strings); if `location` is None, it skips it; and finally, if `location` is a str already, it uses it as the prefix (the latter may or may not be necessary, as one can always pass a tuple with a single str item).

simoncozens: I'm going to take a slightly different but equivalent approach. We have established that users (e.g. feaLib.builder) are responsible for catching OpenTypeLibError exceptions and re-throwing them in ways that make sense to their end-users. So ordinarily the formatting the message (including the location) should be done in the class that uses otlLib. If the end-user sees an OpenTypeLibError message, that's a bug where the calling class failed to catch the exception. What I'm going to do in OpenTypeLibError is just have it try and stringify the `location` as a "best effort" and prefix it, and if it can't stringify it, just return the message without the exception.

simoncozens: Except that won't work because of this:

```
        log.warning(OpenTypeLibError(
            'unsupported "subtable" statement for lookup type',
            location
        ))
```

Aaargh.

simoncozens: I'm going to use a location-formatter lambda attribute, and feel bad about myself.

simoncozens:
In the end I used a hybrid of your approach and mine, which I think will work for other (non-feaLib) clients as well as feaLib. If you could have another quick look over, that would be great.

simoncozens:
This has turned into a mess again. I am learning. Will close and reopen when I have solved the string serialization issue.

simoncozens:
OK, I think I have a neat solution but it might "[multiply extra entities unnecessarily](https://en.wikipedia.org/wiki/Occam%27s_razor)". I once had a boss who taught me that if you don't know the best way to implement a class hierarchy, implement *another* class and it'll show you which bits should go where. So I made another otlLib-client called `MockBuilder` which uses "locations" which are two-element tuples of integers, and now it all makes a bit more sense.

The basic problem we have is that "location" information needs to be stringified, sometimes in the caller (as part of an exception being thrown back from otlLib) but also sometimes in otlLib itself (as part of a warning). Something smells funny about that, because we are saying that locations are abstract to otlLib, but otlLib is having to handle them, because it has no way to pass warning information back to the caller. That's sucky but I don't know of a way around it - you can't "raise a warning".

So we have a location that we need to stringify but we don't necessarily know how we should stringify it. But if we specified that `location` parameters had to be *objects*, those objects could take care of their own stringification. This also will deal with the problem where you have *two* locations that you need to handle - e.g. `location1: Already defined position for pair A V at location2`.

Issue #2001, opened 2020-06-20
[ufolib] glyph names A and a_


typemytype:
when a ufo has glyphs `A` and `a_` it should write both .glif files.

I understand there are case insensitive OSs but overwriting an existing file is not good. The glyphs/contents.plist is also wrong, due to the same issue.

~~this used to work...~~ (edited)
```
from fontTools.ufoLib import UFOWriter

p = 'dummyWriterOnly.ufo'
writer = UFOWriter(path=p)
glyphSet = writer.getGlyphSet(layerName="foreground", defaultLayer=True)

glyphSet.writeGlyph("a", None, None)
glyphSet.writeGlyph("A", None, None)
glyphSet.writeGlyph("a_", None, None)
glyphSet.writeContents()

writer.writeLayerContents(["foreground"])
```

typemytype:
this didnt work either with the old seperate ufoLib pacakge

typemytype:
narrowing down the issue to:

```python
from fontTools.ufoLib.filenames import userNameToFileName

names = [
    "a", "a_", "A", "A_", "i_j"
    ]

for name in names:
    # lower as the file name must be case insensitive 
    print(userNameToFileName(name).lower())
```

ouput:

```console
a
a_
a_
a__
i_j
```

* `a_` is wrong must be `a__`
* `A_` is wrong must be `A___`
* `i_j` is wrong must be `i__j`

typesupply:
I think this could be because the case-insensitive names are not being sent to the file name generator [here](https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/ufoLib/glifLib.py#L455). This might solve it:

```python
fileName = self.glyphNameToFileName(glyphName, self._existingFileNames.values())
```

typemytype:
Thanks Tal!

a glyph with the name `a_` is considered as a edge case, so let the heavy fallback chip in...

anthrotype:
Fixed bt #2002

PR #2002, opened 2020-06-20
make sure a list of possible case-insensitive file names is send

typemytype:
not a dict where the key is the case sensitive file name

fixing #2001


anthrotype:
Thanks for the fix. I'll add a test after merging

PR #2004, opened 2020-06-25
Move feature builders to otlLib, take 3

simoncozens:
See notes at the end of #1999 - this uses a self-stringifying "location object" to avoid otlLib having to know how to report location information.


dscorbett:
`fontTools.feaLib.ast.Element.__init__`’s `location` parameter is documented as a tuple.

simoncozens:
That's true! So...?

simoncozens:
I mean, is the implication that we should change the docs, or we need to change __init__ to consume a tuple, or we can't make anything better because backwards compatibility yadda yadda yadda, or...?

dscorbett:
`__init__` can now take a `FeatureLibLocation`, so at the very least the documentation shouldn’t limit it to a tuple. For backwards compatibility, though, it still needs to be able to take a tuple. Since the only thing a location is used for is being converted to a string, a location can really be of almost any type. I think the best solution is to just say that `location` is a “location”, with a link to an explanation of what locations are and how they are used, maybe in the `otlLib` module documentation.

simoncozens:
Thank you, that's a helpful suggestion. I've updated the doc for `Element`, made `Element.__init__` convert to a location if it isn't one already, and added a test. Documentation for how Locations work will be added in the PR after next, which documents otlLib.

Lib/fontTools/feaLib/location.py
```diff
@@ -0,0 +1,9 @@
+class FeatureLibLocation(object):
+    """A location in a feature file"""
+    def __init__(self, file, line, column):
+        self.file = file
+        self.line = line
+        self.column = column
+
+    def __str__(self):
+    		return "%s:%i:%i" % (self.file, self.line, self.column)
```
anthrotype, line 9:
```suggestion
    		return f"{self.file}:{self.line}:{self.column}"
```

Lib/fontTools/feaLib/location.py
```diff
@@ -0,0 +1,9 @@
+class FeatureLibLocation(object):
```
anthrotype, line 1: you could use a namedtuple with a custom `__str__` method, e.g. like this https://github.com/fonttools/fonttools/blob/80dac9c0f4304b70d2a8103b5962439d842e0ab9/Lib/fontTools/voltLib/ast.py#L2-L11

Lib/fontTools/feaLib/ast.py
```diff
@@ -460,8 +460,8 @@ def addDefinition(self, definition):
                 if otherLoc is None:
                     end = ""
                 else:
-                    end = " at %s:%d:%d" % (
-                        otherLoc[0], otherLoc[1], otherLoc[2])
+                    end = " at %s" % (
+                        otherLoc)
```
anthrotype, line 467: make it one line? And turn it into an f-string if you like

Lib/fontTools/feaLib/builder.py
```diff
@@ -795,8 +795,8 @@ def getMarkAttachClass_(self, location, glyphs):
                 _, loc = self.markAttach_[glyph]
                 raise FeatureLibError(
                     "Glyph %s already has been assigned "
-                    "a MarkAttachmentType at %s:%d:%d" % (
-                        glyph, loc[0], loc[1], loc[2]),
+                    "a MarkAttachmentType at %s" % (
+                        glyph, loc),
```
anthrotype, line 799: this can go the the previous line as well

Lib/fontTools/feaLib/builder.py
```diff
@@ -963,7 +963,7 @@ def add_multiple_subst(self, location,
                     'Removing duplicate multiple substitution from glyph'
                     ' "%s" to %s%s',
                     glyph, replacements,
-                    ' at {}:{}:{}'.format(*location) if location else '',
+                    ' at {}'.format(location) if location else '',
```
anthrotype, line 966:
```suggestion
                    f' at {location}' if location else '',
```

Tests/otlLib/mock_builder_test.py
```diff
@@ -0,0 +1,90 @@
...
+        font = makeTTFont()
+
+        logger = logging.getLogger("fontTools.otlLib.builder")
+
+        with CapturingLogHandler(logger, "INFO") as captor:
+            builder = SinglePosBuilder(font, location)
+            builder.add_subtable_break(MockBuilderLocation((5, "beta")))
+            builder.build()
+
+        captor.assertRegex('5:beta: unsupported "subtable" statement for lookup type')
```
anthrotype, line 77: for new test modules we should use pytest-style tests instead of unittest-style ones.
The class doesn't have to subclass unittest.TestCase, the name only needs to end with "Test" for pytest to discover it. Test methods must start with "test_" to be discovered as such.
Then you use normal built-in `assert` statements instead of the self.assertBlahBlah() methods of unittest and pytest does the rest.

For matching exception messages using regex, you can do

```python
with pytest.raises(Exception, match="your pattern"):
    # code that raises the error
```

Tests/otlLib/mock_builder_test.py
```diff
@@ -0,0 +1,90 @@
...
+    SinglePosBuilder,
+)
+from fontTools.otlLib.error import OpenTypeLibError
+from fontTools.ttLib import TTFont
+from fontTools.misc.loggingTools import CapturingLogHandler
+import logging
+import unittest
+
+
+def makeTTFont():
```
anthrotype, line 24: in pytest, this would be a ["fixture"](https://docs.pytest.org/en/stable/fixture.html) function

```python
@pytest.fixture
def ttfont():
    ...
```

then the test methods that need this would add a parameter with the name of that fixture, e.g.:

```python
def test_something(ttfont):
    assert "a" in ttfont.getGlyphOrder()
```

Lib/fontTools/feaLib/ast.py
```diff
@@ -112,7 +113,9 @@ class Element(object):
     """A base class representing "something" in a feature file."""
 
     def __init__(self, location=None):
-        #: location of this element - tuple of ``(filename, line, column)``
+        #: location of this element as a `FeatureLibLocation` object.
```
anthrotype, line 116: I think this special comment `#:` should be right before the `self.something = ...` assignement it is referring to, for sphinx to treat it as such, so i'd put it right after the if statement. Or maybe it works already?

PR #2005, opened 2020-06-26
Add HashPointPen from psautohint

jenskutilek:
Adapted from https://github.com/adobe-type-tools/psautohint/blob/08b346865710ed3c172f1eb581d6ef243b203f99/python/psautohint/ufoFont.py#L800-L838

The pen generates a hash from a glyph's outlines and components and can be used to check if a glyph has changed.

### DIFF
--- a/Lib/fontTools/pens/hashPointPen.py
+++ b/Lib/fontTools/pens/hashPointPen.py
@@ -1,0 +1,52 @@
+# Modified from https://github.com/adobe-type-tools/psautohint/blob/08b346865710ed3c172f1eb581d6ef243b203f99/python/psautohint/ufoFont.py#L800-L838
+import hashlib
+from fontTools.pens.pointPen import AbstractPointPen
+
+
+class HashPointPen(AbstractPointPen):
+    DEFAULT_TRANSFORM = (1, 0, 0, 1, 0, 0)
+
+    def __init__(self, glyph, glyphSet=None):
+        self.glyphset = glyphSet
+        self.width = round(getattr(glyph, "width", 1000), 9)
+        self.data = ["w%s" % self.width]
+
+    def getHash(self):
+        data = "".join(self.data)
+        if len(data) >= 128:
+            data = hashlib.sha512(data.encode("ascii")).hexdigest()
+        return data
+
+    def beginPath(self, identifier=None, **kwargs):
+        pass
+
+    def endPath(self):
+        pass
+
+    def addPoint(
+        self,
+        pt,
+        segmentType=None,
+        smooth=False,
+        name=None,
+        identifier=None,
+        **kwargs
+    ):
+        if segmentType is None:
+            pt_type = ""
+        else:
+            pt_type = segmentType[0]
+        self.data.append("%s%s%s" % (pt_type, repr(pt[0]), repr(pt[1])))
+
+    def addComponent(
+        self, baseGlyphName, transformation, identifier=None, **kwargs
+    ):
+        self.data.append("base:%s" % baseGlyphName)
+
+        for i, v in enumerate(transformation):
+            if transformation[i] != self.DEFAULT_TRANSFORM[i]:
+                self.data.append(str(round(v, 9)))
+
+        self.data.append("w%s" % self.width)
+        glyph = self.glyphset[baseGlyphName]
+        glyph.drawPoints(self)
--- a/Lib/fontTools/pens/hashPointPen.py
+++ b/Lib/fontTools/pens/hashPointPen.py
@@ -1,6 +1,7 @@
 # Modified from https://github.com/adobe-type-tools/psautohint/blob/08b346865710ed3c172f1eb581d6ef243b203f99/python/psautohint/ufoFont.py#L800-L838
 import hashlib
 from fontTools.pens.pointPen import AbstractPointPen
+from fontTools.pens.transformPen import TransformPointPen
 
 
 class HashPointPen(AbstractPointPen):
@@ -41,12 +42,5 @@ class HashPointPen(AbstractPointPen):
     def addComponent(
         self, baseGlyphName, transformation, identifier=None, **kwargs
     ):
-        self.data.append("base:%s" % baseGlyphName)
-
-        for i, v in enumerate(transformation):
-            if transformation[i] != self.DEFAULT_TRANSFORM[i]:
-                self.data.append(str(round(v, 9)))
-
-        self.data.append("w%s" % self.width)
-        glyph = self.glyphset[baseGlyphName]
-        glyph.drawPoints(self)
+        tpen = TransformPointPen(self, transformation)
+        self.glyphset[baseGlyphName].drawPoints(tpen)
--- a/Lib/fontTools/pens/hashPointPen.py
+++ b/Lib/fontTools/pens/hashPointPen.py
@@ -37,7 +37,7 @@ class HashPointPen(AbstractPointPen):
             pt_type = ""
         else:
             pt_type = segmentType[0]
-        self.data.append("%s%s%s" % (pt_type, repr(pt[0]), repr(pt[1])))
+        self.data.append(f"{pt_type}{pt[0]}{pt[1]:+}")
 
     def addComponent(
         self, baseGlyphName, transformation, identifier=None, **kwargs
--- a/Lib/fontTools/pens/hashPointPen.py
+++ b/Lib/fontTools/pens/hashPointPen.py
@@ -55,6 +55,6 @@ class HashPointPen(AbstractPointPen):
     ):
         tr = "".join([f"{t:+}" for t in transformation])
         self.data.append(f"[({tr})")
-        tpen = TransformPointPen(self, transformation)
+        tpen = TransformPointPen(self, Identity)
         self.glyphset[baseGlyphName].drawPoints(tpen)
         self.data.append("]")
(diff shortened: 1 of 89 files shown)


Lib/fontTools/pens/hashPointPen.py
```diff
@@ -0,0 +1,46 @@
...
+        smooth=False,
+        name=None,
+        identifier=None,
+        **kwargs
+    ):
+        if segmentType is None:
+            pt_type = ""
+        else:
+            pt_type = segmentType[0]
+        self.data.append("%s%s%s" % (pt_type, repr(pt[0]), repr(pt[1])))
```
justvanrossum, line 40: Perhaps use an f-string here?
```python
        self.data.append(f"{pt_type}{pt[0]}{pt[1]}")
```
Btw. this is not robust against a change of, say, `(1, 11)` to `(11, 1)`, as both will be represented as `111`.

jenskutilek: Ha, nice catch!

Lib/fontTools/pens/hashPointPen.py
```diff
@@ -0,0 +1,46 @@
+# Modified from https://github.com/adobe-type-tools/psautohint/blob/08b346865710ed3c172f1eb581d6ef243b203f99/python/psautohint/ufoFont.py#L800-L838
+import hashlib
+from fontTools.pens.pointPen import AbstractPointPen
+from fontTools.pens.transformPen import TransformPointPen
+
+
+class HashPointPen(AbstractPointPen):
+    DEFAULT_TRANSFORM = (1, 0, 0, 1, 0, 0)
```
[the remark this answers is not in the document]
jenskutilek: Actually the DEFAULT_TRANSFORM is not needed, I've removed it. It was a remnant from the Adobe code.

Lib/fontTools/pens/hashPointPen.py
```diff
@@ -0,0 +1,57 @@
...
+from fontTools.pens.pointPen import AbstractPointPen
+from fontTools.pens.transformPen import TransformPointPen
+
+
+class HashPointPen(AbstractPointPen):
+    """
+    This pen can be used to check if a glyph's contents (outlines plus
+    components) has changed.
+
+    Components are decomposed, because that is what happens in TrueType
```
anthrotype, line 13: IIUC that means a composite glyph with one component would hash the same as the base glyph it references. I find that a bit strange for a pen that is supposed to check if a glyph's contents have changed.

Lib/fontTools/pens/hashPointPen.py
```diff
@@ -0,0 +1,57 @@
...
+import hashlib
+
+from fontTools.pens.pointPen import AbstractPointPen
+from fontTools.pens.transformPen import TransformPointPen
+
+
+class HashPointPen(AbstractPointPen):
+    """
+    This pen can be used to check if a glyph's contents (outlines plus
+    components) has changed.
```
anthrotype, line 11:
```suggestion
    components) have changed.
```

Lib/fontTools/pens/hashPointPen.py
```diff
@@ -0,0 +1,57 @@
...
+    Components are decomposed, because that is what happens in TrueType
+    hinting.
+    """
+
+    def __init__(self, glyph, glyphSet=None):
+        self.glyphset = glyphSet
+        self.width = round(getattr(glyph, "width", 0), 9)
+        self.data = ["w%s" % self.width]
+
+    def getHash(self):
```
anthrotype, line 22: what's the use of `getHash()` method if we also have a `hash` property?

Lib/fontTools/pens/hashPointPen.py
jenskutilek: Hm... the original hash pen from Adobe just added the base glyph name and transformation to the hash. But for detecting outline changes, especially for TrueType hinting, we need to make sure the component's outlines also haven't changed. On the other hand, a base glyph may have been renamed, so it may be the same base glyph even if the component references a different name.

Maybe adding a generic "component marker" (just to say "this path comes from a component" without specifying the base glyph name) to the hash and still adding the (transformed) component points verbatim would solve this?

Then again, only the base glyph's hinting may have changed, which may or may not break any additional TT instructions in the composite glyph, which can't be detected by this hashing mechanism and maybe is out of scope.

Perhaps it's best to be on the safe side and include the component names as well as the transformations in the hash, WDYT?

Lib/fontTools/pens/hashPointPen.py
jenskutilek: A false sense of backwards compatibility I guess; it was called getHash in the Adobe code. But I'll get rid of it, there's no need for it.

Lib/fontTools/pens/hashPointPen.py
anthrotype: how about recording the _original_ base glyphs' outlines followed by the respective transform? The name of the base glyph is irrelevant as you said, it's the content that this hash pen is recording.

Lib/fontTools/pens/hashPointPen.py
```diff
@@ -0,0 +1,60 @@
...
+        else:
+            pt_type = segmentType[0]
+        self.data.append(f"{pt_type}{pt[0]:g}{pt[1]:+g}")
+
+    def addComponent(
+        self, baseGlyphName, transformation, identifier=None, **kwargs
+    ):
+        tr = "".join([f"{t:+}" for t in transformation])
+        self.data.append(f"[({tr})")
+        tpen = TransformPointPen(self, Identity)
```
anthrotype, line 58: this transform pen is doing nothing

anthrotype: you can just pass `self` to the drawPoints below

jenskutilek: Stupid hack, fix is forthcoming

Lib/fontTools/pens/hashPointPen.py
```diff
@@ -0,0 +1,60 @@
...
+            pt_type = "o"  # offcurve
+        else:
+            pt_type = segmentType[0]
+        self.data.append(f"{pt_type}{pt[0]:g}{pt[1]:+g}")
+
+    def addComponent(
+        self, baseGlyphName, transformation, identifier=None, **kwargs
+    ):
+        tr = "".join([f"{t:+}" for t in transformation])
+        self.data.append(f"[({tr})")
```
anthrotype, line 56: I was rather thinking that you dump the original base glyph coordinagtes first, and then the transform tuple (you do the transform first). If you have nested composite glyphs, you'd have the base glyph coordinates recursively followed by a chain of transform tuples.

jenskutilek: You think it's good to keep the hash parseable/human-readable? I was briefly tempted to do so, but anything longer than 128 chars will be turned into SHA512 anyway, so ...

anthrotype: hm, you're right about that.
BTW, have you got a test with nested components? if not you could add one for completeness.

Lib/fontTools/pens/hashPointPen.py
```diff
@@ -0,0 +1,58 @@
...
+from fontTools.misc.transform import Identity
+from fontTools.pens.pointPen import AbstractPointPen
+
+
+class HashPointPen(AbstractPointPen):
+    """
+    This pen can be used to check if a glyph's contents (outlines plus
+    components) have changed.
+
+    Components are decomposed, because that is what happens in TrueType
```
anthrotype, line 13: the docstring here no longer reflects what the pen is doing with components, could you please update this as well?

jenskutilek: Sure!

Lib/fontTools/pens/hashPointPen.py
```diff
@@ -0,0 +1,58 @@
...
+    This pen can be used to check if a glyph's contents (outlines plus
+    components) have changed.
+
+    Components are added as the original outline plus each composite's
+    transformation.
+    """
+
+    DEFAULT_TRANSFORM = Identity
+
+    def __init__(self, glyph, glyphSet=None):
```
anthrotype, line 19: I find a bit strange that we pass glyph in the constructor (just to get its width) but then we require one to call glyph.drawPoints(pen).
I think we should only pass the width to the pen constructor, not the entire glyph.

anthrotype:
I suggest you add some examples of how to use the pen in the class docstring

jenskutilek:
I will resume working on this beginning of September.

Lib/fontTools/pens/hashPointPen.py
jenskutilek: I have changed it.

jenskutilek:
I think this is ready to be merged if no further issues arise. WDYT, @anthrotype?

Lib/fontTools/pens/hashPointPen.py
```diff
@@ -0,0 +1,76 @@
...
+    > stored_hash = ttdata.get("id", None)  # The hash is stored in the "id" key
+    > if stored_hash is None or stored_hash != hash_pen.hash:
+    >    logger.error(f"Glyph hash mismatch, glyph '{name}' will have no instructions in font.")
+    > else:
+    >    # The hash values are identical, the outline has not changed.
+    >    # Compile the hinting code ...
+    >    pass
+    """
+
+    DEFAULT_TRANSFORM = Identity
```
anthrotype, line 36: this seems unused (as well as the respective import at the top)

anthrotype (approved):

PR #2009, opened 2020-07-03
otlLib documentation

simoncozens:


Lib/fontTools/otlLib/builder.py
```diff
@@ -471,6 +839,15 @@ def add_subtable_break(self, location):
 
 
 class ClassPairPosSubtableBuilder(object):
+    """Builds class-based Pair Positioning (GPOS2 format 2) subtables.
+
+    Note that this does *not* build a GPOS2 ``otTables.Lookup`` directly,
+    but builds a list of ``otTables.PairPos`` subtables. It is used by the
+    :class:`PairPosBuilder` below.
+
+    Attributes:
```
anthrotype, line 848: should we also document the other two init parameters `valueFormat1` and `valueFormat2` here?

Lib/fontTools/otlLib/builder.py
```diff
@@ -668,6 +1204,20 @@ def buildLigatureSubstSubtable(mapping):
...
 
 def buildAnchor(x, y, point=None, deviceX=None, deviceY=None):
+    """Builds an Anchor table.
+
+    This determines the appropriate anchor format based on the passed parameters.
+
+    Args:
+        x (int): X coordinate.
+        y (int): Y coordinate.
+        point (int): Index of glyph contour point, if provided.
```
anthrotype, line 1214: I don't know if we want to use the typing annotation syntax in the docstrings or not (ideally yes, but perhaps can be done later?), e.g.

```suggestion
        point (Optional[int]): Index of glyph contour point, if provided.
```

If we do, we'll have to change this and many other similar places.

Lib/fontTools/otlLib/builder.py
```diff
@@ -695,14 +1270,29 @@ def buildBaseArray(bases, numMarkClasses, glyphMap):
 
 
 def buildBaseRecord(anchors):
-    """[ot.Anchor, ot.Anchor, ...] --> ot.BaseRecord"""
+    # [otTables.Anchor, otTables.Anchor, ...] --> otTables.BaseRecord
+    # (Not docstringed because it's only needed when building BaseArray,
+    # at which point you should be using the above.)
```
anthrotype, line 1275: this comment is not necessary, we just stick to the implicit rule that if something hasn't a `"""` docstring it isn't in the publicly documented API

Lib/fontTools/otlLib/builder.py
```diff
@@ -765,15 +1423,41 @@ def buildLigatureArray(ligs, numMarkClasses, glyphMap):
 
 
 def buildLigatureAttach(components):
-    """[[Anchor, Anchor], [Anchor, Anchor, Anchor]] --> LigatureAttach"""
+    # [[Anchor, Anchor], [Anchor, Anchor, Anchor]] --> LigatureAttach
+    # (Not docstringed because it's only needed when building LigatureArray,
+    # at which point you should be using the above.)
```
anthrotype, line 1428: unnecessary comment, like above

Lib/fontTools/otlLib/builder.py
```diff
@@ -863,14 +1643,16 @@ def buildMarkRecord(classID, anchor):
 
 
 def buildMark2Record(anchors):
-    """[ot.Anchor, ot.Anchor, ...] --> ot.Mark2Record"""
+    # [otTables.Anchor, otTables.Anchor, ...] --> otTables.Mark2Record
+    # (Not docstringed because it's only needed when building MarkMarkPos,
+    # at which point you should be using otl.lookupBuilders.MarkMarkPosBuilder.)
```
anthrotype, line 1648: pls remove comment

Lib/fontTools/otlLib/builder.py
```diff
@@ -854,6 +1633,7 @@ def buildMarkLigPosSubtable(marks, ligs, glyphMap):
 
 
 def buildMarkRecord(classID, anchor):
+    # Only used by buildMarkArray
```
anthrotype, line 1636: is this comment necessary?

Lib/fontTools/otlLib/builder.py
```diff
@@ -129,13 +184,39 @@ def build_subst_subtables(self, mapping, klass):
         return subtables
 
     def add_subtable_break(self, location):
+        """Add an explicit subtable break.
+
+        Args:
+            location: A string or tuple representing the location in the
```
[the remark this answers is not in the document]
simoncozens: I don't like calling something  "optional" if you have to provide it. :-) I'll reword this.

Lib/fontTools/otlLib/builder.py
simoncozens: I can't see that they're actually used. I'm going to remove them instead...

anthrotype: Interesting. I wonder why they were there in the first place..

simoncozens:
We will rethink the typing information on the docs another time.

Are we happy with the change to ClassPairPosSubtableBuilder? If so I'll merge this.

anthrotype:
> Are we happy with the change to ClassPairPosSubtableBuilder? If so I'll merge this.

it's a pretty niche class, if those attributes weren't really used internally so they can safely be dropped indeed.

PR #2010, opened 2020-07-03
Test that passing a markFilterSet sets the appropriate lookup flag

simoncozens:
See discussion in 3b50a5. Don't add the USE_MARK_FILTERING_SET flag explicitly, but still check it is set on export.


anthrotype:
I would keep both tests, one with explicit flag, another without, and assert they do the same.

behdad:
@simoncozens can you address review please?

PR #2013, opened 2020-07-07
Fix documentation for MultipleSubstStatement (fixes #2012)

### LINKED ISSUE #2012: [feaLib] MultipleSubstStatement docstring mistake
Here `glyph` seems to actually expect a single `glyph-containing object` instead of a list?

https://github.com/fonttools/fonttools/blob/73631c9642235e55bbf31eab1fc2e9f156b62ec2/Lib/fontTools/feaLib/ast.py#L1119-L1156

simoncozens:
This fixes the docstring; yes, it does it using a Google-style docstring, and yes, this is currently inconsistent, but half the docstrings I've written so far are Python and the other half are Google-style so we're inconsistent already, and better docs doesn't hurt. Going through all the docstrings and converting them all to Google-style is on my todo list. Consider this just getting one in early. :-)

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -1119,11 +1119,14 @@ class MarkMarkPosStatement(Statement):
 class MultipleSubstStatement(Statement):
     """A multiple substitution statement.
 
-    ``prefix``, ``glyph``, ``suffix`` and ``replacement`` should be lists of
-    `glyph-containing objects`_.
-
-    If ``forceChain`` is True, this is expressed as a chaining rule
-    (e.g. ``sub f' i' by f_i``) even when no context is given."""
+    Args:
+        prefix: a list of `glyph-containing objects`_.
+        glyph: a single glyph-containing object.
+        suffix: a list of glyph-containing objects.
+        replacement: a list of glyph-containing objects.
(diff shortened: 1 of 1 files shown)


Lib/fontTools/feaLib/ast.py
```diff
@@ -1119,11 +1119,14 @@ def asFea(self, indent=""):
...
     """A multiple substitution statement.
 
-    ``prefix``, ``glyph``, ``suffix`` and ``replacement`` should be lists of
-    `glyph-containing objects`_.
-
-    If ``forceChain`` is True, this is expressed as a chaining rule
-    (e.g. ``sub f' i' by f_i``) even when no context is given."""
+    Args:
+        prefix: a list of `glyph-containing objects`_.
+        glyph: a single glyph-containing object.
```
anthrotype, line 1124: In the line above you have wrapped `glyph-containing object` as inline code inside backticks, should you do it consistently for the following lines as well?

simoncozens: This was deliberate. The first line is a link to explain what a glyph-containing object is; after that, the reader should know and doesn't need the explanation again.

anthrotype: ok thanks!

Issue #1856, opened 2020-03-13
feaLib should build simple contextual lookups when no backtrack/lookahead


anthrotype:
The Adobe FEA spec does not distinguish between contextual positioning/substitution lookups (GPOS 7, GSUB 5), on the one hand, and _chaining_ contextual pos/subst lookups (GPOS 8, GSUB 6), on the other. The two types of lookups are essentially the same except that the latter also can define optional backtrack/lookahead sequences.

FeaLib builder currently always encodes contextual pos/subst rules with Format 3 (Coverage-based) ChainContex{Pos,Subst} lookup subtables.

https://github.com/fonttools/fonttools/blob/42dcfddfb9687de106e5deb59cea015280dd507f/Lib/fontTools/feaLib/builder.py#L1246

https://github.com/fonttools/fonttools/blob/42dcfddfb9687de106e5deb59cea015280dd507f/Lib/fontTools/feaLib/builder.py#L1293

Ideally it could use the more efficient types of non-chaining contextual lookups when a contextual rule contains no backtrack nor lookahend sequences.

http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#5e-gsub-lookuptype-5-contextual-substitution
http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#6g-gpos-lookuptype-7-contextual-positioning

I haven't had the chance to check, does makeotf do anything like that?

simoncozens:
I'll work on this, but just for info, makeotf does not do this - it treats all contextuals as chaining contextuals.

PR #2016, opened 2020-07-07
Generate GSUB5/GPOS7 lookups (See #1856)

simoncozens:
Done in fairly small steps. Note that the ClassContextualRuleset object which this introduces is also going to be used when choosing the appropriate format representation of contextual lookups, so this is also the start of the contextual optimizer.


simoncozens:
I now have another couple of commits on top of this which add Format 1 subtables where possible, but will save that for another PR as and when this one gets merged.

Lib/fontTools/otlLib/builder.py
```diff
@@ -244,11 +251,70 @@ def add_subtable_break(self, location):
...
+
+    def addRule(self, prefix, glyphs, suffix, lookups):
+        self.rules.append((prefix, glyphs, suffix, lookups))
+
+    @property
+    def hasPrefixOrSuffix(self):
+        # Do we have any prefixes/suffixes? If this is False for all
+        # rulesets, we can express the whole lookup as GPOS5/GSUB7.
+        for (prefix, glyphs, suffix, lookups) in self.rules:
+            if len(prefix) > 0 or len(suffix) > 0: return True
```
anthrotype, line 266: if statement split over two lines, please

Lib/fontTools/otlLib/builder.py
```diff
@@ -244,11 +251,70 @@ def add_subtable_break(self, location):
...
+        return False
+
+    @property
+    def hasAnyGlyphClasses(self):
+        # Do we use glyph classes anywhere in the rules? If this is False
+        # we can express this subtable as a Format 1.
+        for (prefix, glyphs, suffix, lookups) in self.rules:
+            for coverage in (prefix, glyphs, suffix):
+                if any([len(x) > 1 for x in coverage]):
```
anthrotype, line 275:
```suggestion
                if any(len(x) > 1 for x in coverage):
```

Lib/fontTools/otlLib/builder.py
```diff
@@ -244,11 +251,70 @@ def add_subtable_break(self, location):
...
+                    return True
+        return False
+
+    def format2Classdefs(self):
```
anthrotype, line 279:
```suggestion
    def format2ClassDefs(self):
```

Lib/fontTools/otlLib/builder.py
```diff
@@ -244,11 +251,70 @@ def add_subtable_break(self, location):
...
+        PREFIX, GLYPHS, SUFFIX = 0,1,2
+        classdefbuilders = []
```
anthrotype, line 281:
```suggestion
        classDefBuilders = []
```

I think we should stick to one style for the entire module. If the rest uses camelCase, then new code should follow.

Lib/fontTools/otlLib/builder.py
```diff
@@ -244,11 +251,70 @@ def add_subtable_break(self, location):
...
+
 class ChainContextualBuilder(LookupBuilder):
     def equals(self, other):
         return (LookupBuilder.equals(self, other) and
                 self.rules == other.rules)
 
+    def rulesets(self):
+        # Return a list of ChainContextRuleset objects, taking explicit
+        # subtable breaks into account
+        ruleset = [ ChainContextualRuleset() ]
```
anthrotype, line 309: there's no reason to add spaces to square brackets, it's just distracting. I know the rest of the codebase is not consistent, but for at least new code let's stick to black formatting so we don't waste time over style.
```suggestion
        ruleset = [ChainContextualRuleset()]
```

Lib/fontTools/otlLib/builder.py
```diff
@@ -244,11 +251,70 @@ def add_subtable_break(self, location):
...
+        for (prefix, glyphs, suffix, lookups) in self.rules:
+            if prefix == self.SUBTABLE_BREAK_:
+                ruleset.append(ChainContextualRuleset() )
```
anthrotype, line 312:
```suggestion
                ruleset.append(ChainContextualRuleset())
```

Lib/fontTools/otlLib/builder.py
```diff
@@ -257,35 +323,46 @@ def build(self):
...
         subtables = []
-        for (prefix, glyphs, suffix, lookups) in self.rules:
-            if prefix == self.SUBTABLE_BREAK_:
-                continue
-            st = self.newSubtable_()
-            subtables.append(st)
-            st.Format = 3
+        chaining = False
+        rulesets = self.rulesets()
+        chaining = any([ruleset.hasPrefixOrSuffix for ruleset in rulesets])
```
anthrotype, line 328:
```suggestion
        chaining = any(ruleset.hasPrefixOrSuffix for ruleset in rulesets)
```

Lib/fontTools/otlLib/builder.py
```diff
@@ -257,35 +323,46 @@ def build(self):
...
-                        self.addLookupRecordToSubtable_(st, rec)
-        return self.buildLookup_(subtables)
+        else:
+            self.setCoverage_(glyphs, st)
+
+        for sequenceIndex, lookupList in enumerate(lookups):
+            if lookupList is not None:
+                if not isinstance(lookupList, list):
+                    # Can happen with synthesised lookups
+                    lookupList = [ lookupList ]
```
anthrotype, line 351:
```suggestion
                    lookupList = [lookupList]
```

Tests/otlLib/builder_test.py
```diff
@@ -1392,6 +1392,63 @@ def test_stat_infinities():
     assert struct.pack(">l", posInf) == b"\x7f\xff\xff\xff"
 
 
+class ChainContextualRulesetTest(object):
+    def test_makeRulesets(self):
+        font = ttLib.TTFont()
+        font.setGlyphOrder(["a","b","c","d","A","B","C","D","E"])
+        sb = builder.ChainContextSubstBuilder(font, None)
+        prefix = [ ["a"], ["b"] ]
```
anthrotype, line 1400:
```suggestion
        prefix = [["a"], ["b"]]
```
and all the following lines as well

Tests/otlLib/builder_test.py
```diff
@@ -1392,6 +1392,63 @@ def test_stat_infinities():
...
+        prefix = [ ["a"], ["d"] ]
+        sb.rules.append((prefix, input_, suffix, lookups))
+
+        sb.add_subtable_break(None)
+
+        # Second subtable has some glyph classes
+        prefix = [ ["A"] ]
+        input_ = [ ["E"] ]
+        sb.rules.append((prefix, input_, suffix, lookups))
+        input_ = [ ["C","D"] ]
```
anthrotype, line 1415:
```suggestion
        input_ = [["C", "D"]]
```

Tests/otlLib/builder_test.py
```diff
@@ -1392,6 +1392,63 @@ def test_stat_infinities():
...
+        sb.rules.append((prefix, input_, suffix, lookups))
+
+        sb.add_subtable_break(None)
+
+        # Third subtable has no pre/post context
+        prefix = []
+        suffix = []
+        sb.rules.append((prefix, input_, suffix, lookups))
+        input_ = [ ["C","D"] ]
+        sb.rules.append((prefix, input_, suffix, lookups))
```
anthrotype, line 1428: it's hard to track the state of the prefix, input_, suffix local variables as you mutate them inside the test. maybe spell out the content of each every time you append a new rule (even if it means repeating the same string literals) so it's clear what rules contain what.

Lib/fontTools/otlLib/builder.py
```diff
@@ -172,6 +172,13 @@ def setInputCoverage_(self, glyphs, subtable):
             coverage = buildCoverage(g, self.glyphMap)
             subtable.InputCoverage.append(coverage)
 
+    def setCoverage_(self, glyphs, subtable):
+        subtable.GlyphCount = len(glyphs)
+        subtable.Coverage = []
```
khaledhosny, line 177:
```suggestion
        subtable.Coverage = [buildCoverage(g, self.glyphMap) for g in glyphs]
```

simoncozens: Probably better to be consistent with existing methods (setBacktrackCoverage_/setLookAheadCoverage_/setInputCoverage_). Happy to fix them all in a separate PR.

khaledhosny: Sounds good (Also it would be nice if all these separate `*Count` members dropped or ignored, since they are redundant. But that is totally off-topic for this change).

PR #2020, opened 2020-07-11
Build format 1 subtables for contextual / chaining contextual lookups where appropriate

simoncozens:


Tests/feaLib/builder_test.py
```diff
@@ -51,7 +51,7 @@ def makeTTFont():
 
 class BuilderTest(unittest.TestCase):
     # Feature files in data/*.fea; output gets compared to data/*.ttx.
-    TEST_FEATURE_FILES = """
+    TEST_FEATURE_FILES = os.getenv("TEST_FEATURE_FILES","""
```
anthrotype, line 54: You don't need to read from an environment variable to select a specific test to run. With pytest you can pass the `-k` option whtere the argument are test names (or parts of their name). E.g.

```
$ pytest -k MultipleLookupsPerGlyph2
```

Or you can be more specific and only select one test file and restrict to the test named such:

```
$ pytest Tests/feaLib/builder_test.py -k MultipleLookupsPerGlyph2
```

If you're using `tox` to run test suite, you can also pass parameters to pytest. Anything after `--` is passed on to pytest:

```
$ tox -e py37-cov -- Tests/feaLib/builder_test.py -k MultipleLookupsPerGlyph2
```

This is standard tox and pytest, documented in their respecitve documentation. I prefer this over custom use of environment vars.

Lib/fontTools/otlLib/builder.py
```diff
@@ -328,12 +328,42 @@ def build(self):
...
         # buildLookup_
         return self.buildLookup_(subtables)
 
+    def buildFormat1Subtable(self, ruleset, chaining=True):
+        st = self.newSubtable_(chaining=chaining)
+        st.Format = 1
+        rulesetObjs = OrderedDict()
+        for rule in ruleset.rules:
+            (prefix, glyphs, suffix, lookups) = rule
+            firstGlyph = list(glyphs[0])[0] # odict_keys can't be indexed, needs coercion
```
anthrotype, line 347: the fact odict_keys can't be indexed is obvious -- what is not is whether the `rule` tuple should contain odict_keys as items. Was that an explicit/documented decision? For consistency it would be better that the type of this ruleset rules be  and always the same (be it list, tuple, or whatever). So this coercion to list could be done when the rule tuple is first created, instead of down the line.

Besides, these rules ideally could be turned into a named tuple and their items accessed with attribute lookup (e.g. `rule.prefix`) instead of unpacking the tuple to get to its items by position.

Lib/fontTools/otlLib/builder.py
```diff
@@ -328,12 +328,42 @@ def build(self):
...
+            if not firstGlyph in rulesetObjs:
+                rulesetObjs[firstGlyph] = self.newRuleSet_(st, chaining=chaining)
+            rulesetObj = rulesetObjs[firstGlyph]
+            ruleObj = self.newRule_(rulesetObj, chaining=chaining)
+            if chaining:
+                ruleObj.BacktrackGlyphCount = len(prefix)
+                ruleObj.InputGlyphCount = len(glyphs)-1
+                ruleObj.LookAheadGlyphCount = len(suffix)
+                ruleObj.Backtrack = [list(x)[0] for x in reversed(prefix)]
+                ruleObj.Input     = [list(x)[0] for x in glyphs[1:]]
```
anthrotype, line 357: we don't use whitespace for expressive purpose, we should stick to black style.

Lib/fontTools/otlLib/builder.py
```diff
@@ -328,12 +328,42 @@ def build(self):
...
+        for rule in ruleset.rules:
+            (prefix, glyphs, suffix, lookups) = rule
+            firstGlyph = list(glyphs[0])[0] # odict_keys can't be indexed, needs coercion
+            if not firstGlyph in rulesetObjs:
+                rulesetObjs[firstGlyph] = self.newRuleSet_(st, chaining=chaining)
+            rulesetObj = rulesetObjs[firstGlyph]
+            ruleObj = self.newRule_(rulesetObj, chaining=chaining)
+            if chaining:
+                ruleObj.BacktrackGlyphCount = len(prefix)
+                ruleObj.InputGlyphCount = len(glyphs)-1
```
anthrotype, line 354:
```suggestion
                ruleObj.InputGlyphCount = len(glyphs) - 1
```

Lib/fontTools/otlLib/builder.py
```diff
@@ -328,12 +328,42 @@ def build(self):
         rulesets = self.rulesets()
         chaining = any(ruleset.hasPrefixOrSuffix for ruleset in rulesets)
         for ruleset in rulesets:
-            for rule in ruleset.rules:
-                subtables.append(self.buildFormat3Subtable(rule, chaining))
+            if not ruleset.hasAnyGlyphClasses and len(ruleset.rules) > 2:
+                # Experimentally, format 1 is only more efficient for >2 rules
```
anthrotype, line 332: I would be curious to understand why that's the case.

Tests/feaLib/builder_test.py
simoncozens: Oh great - I'll take this commit out.

simoncozens:
It needs redoing now we have named tuples for rules.

PR #2022, opened 2020-07-15
Reformat feaLib and otlLib with black

simoncozens:
Since I'm working on these things a lot, it's going to be a lot easier if they're compliant with our preferred style already, then I can just use the autoformatter on save.


simoncozens:
I don't know why appveyor couldn't merge this, and I don't know how to ping it to check again.

PR #2026, opened 2020-07-21
Use namedtuple to represent contextual rules

simoncozens:
As mentioned in #2020:

> Besides, these rules ideally could be turned into a named tuple and their items accessed with attribute lookup (e.g. rule.prefix) instead of unpacking the tuple to get to its items by position.

(Note: This sits on top of #2025. Merge that one first, and this will magically become readable.)

### DIFF
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -269,15 +269,15 @@ class ChainContextualRuleset:
     def __init__(self):
         self.rules = []
 
-    def addRule(self, prefix, glyphs, suffix, lookups):
-        self.rules.append((prefix, glyphs, suffix, lookups))
+    def addRule(self, rule):
+        self.rules.append(rule)
 
     @property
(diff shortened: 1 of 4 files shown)


Lib/fontTools/otlLib/builder.py
```diff
@@ -257,6 +257,14 @@ def add_subtable_break(self, location):
         self.alternates[(self.SUBTABLE_BREAK_, location)] = self.SUBTABLE_BREAK_
 
 
+class ChainContextualRule(
+    namedtuple("ChainContextualRule", ["prefix", "glyphs", "suffix", "lookups"])
+):
+    @property
+    def is_subtable_break(self):
```
anthrotype, line 264: much nicer

anthrotype:
(actually we don't need to merge/rebase master, this is ready for merge already)

simoncozens:
I just rebased it. :-)

anthrotype:
> I just rebased it. :-)

you're fast!

PR #2027, opened 2020-07-22
Document mtiLib library

simoncozens:


Lib/fontTools/mtiLib/__init__.py
```diff
@@ -1146,12 +1146,33 @@ def expectendswith(self, s):
...
 
 def main(args=None, font=None):
-	"""Convert a FontDame OTL file to TTX XML"""
+	"""Convert a FontDame OTL file to TTX XML.
+
+	Writes XML output to stdout.
+
+	Args:
+		args: Command line arguments (``--font``, ``--table``, input files).
+		font (TTFont): If passed, used for sorting glyph sets.
```
anthrotype, line 1174: I don't think we should document this parameter, it's only used/useful by the Tests/mtiLib/mti_test.py to pass in a mock TTFont. A `main()` function is only meant to be used as CLI, and must have one optional `args` (again only used by tests to override sys.args), and should not be considered part of a library's API.

Maybe even rename the parameter `_font` to make it clearer it's private, or alternatively use `**kwargs` and pop a font from it (`kwargs.pop("font", None)`) from inside main().

Lib/fontTools/mtiLib/__init__.py
```diff
@@ -1146,12 +1146,33 @@ def expectendswith(self, s):
 		return line
 
 def build(f, font, tableTag=None):
+	"""Parse a Monotype font layout file.
+
+	A font object must be passed, but this may be a "dummy" font; it is only
+	used for sorting glyph sets when making coverage tables.
```
anthrotype, line 1152: The TTFont object is also used to store the OT layout tables built from the FontDame source

Lib/fontTools/mtiLib/__init__.py
```diff
@@ -1146,12 +1146,33 @@ def expectendswith(self, s):
 		return line
 
 def build(f, font, tableTag=None):
+	"""Parse a Monotype font layout file.
```
anthrotype, line 1149: it doesn't just parse but it also builds OT layout tables from it (GPOS, GSUB, GDEF) as it parses along.

anthrotype (approved):

Issue #2029, opened 2020-07-23
Is there a way to decompose components of a glyph in a TTF?


arrowtype:
I am attempting to decompose components in glyphs within a TTF, but having trouble.

Is there a way to do this? 

It seems that there is nothing direct, and that I may need to use pens. I am pretty new to pens, and not sure I understand them completely.

I _think_ I would need to find a way to record points of each component to a `pen` object, and then draw those points into the ttGlyph with the `.drawPoints()` method. However, it doesn’t seem that the [decomposingRecordingPen](https://github.com/fonttools/fonttools/blob/43599ee3ddad3e20ea0807f2a7b3fbd984b93125/Lib/fontTools/pens/recordingPen.py#L70) works with TTFont objects. I also can’t figure out whether I am using the ttGlyphPen correctly (can it even record points, or is it just a way to store points to later draw?).

```python
def decomposeGlyph(font, glyphName):

    components = (font['glyf'][glyphName].getComponentNames(font['glyf']))

    ## BASIC IDEA (not yet working)
    # for component in glyph
        # record tt points of component base
        # draw points to glyph
    
    glyphset = font.getGlyphSet()

    for comp in components:
        glyph = glyphset[comp]
        pen = ttGlyphPen.TTGlyphPen(glyph)
        # font['glyf'][glyphName].drawPoints(pen, font['glyf'])
        font['glyf'][glyphName].draw(pen, font['glyf'])
```

It may or may not be a related goal, but I am also hoping to remove overlap in the glyphs in a TTF. If this is easy to point out, I would love to get a head start on it, but if it’s a totally separate thing, feel free to skip this part and I will look into it separately.

Thank you so much for any tips or insights!

justvanrossum:
This snippet works:

```python
from fontTools.ttLib import TTFont
from fontTools.pens.recordingPen import DecomposingRecordingPen
from fontTools.pens.ttGlyphPen import TTGlyphPen


src = "font.ttf"
dst = "font-decomposed.ttf"

f = TTFont(src)

glyfTable = f["glyf"]
gs = f.getGlyphSet()
for glyphName in gs.keys():
    if not glyfTable[glyphName].isComposite():
        continue
    dcPen = DecomposingRecordingPen(gs)
    gs[glyphName].draw(dcPen)
    ttPen = TTGlyphPen(None)
    dcPen.replay(ttPen)
    glyfTable[glyphName] = ttPen.glyph()

f.save(dst)
```

[...]

PR #2030, opened 2020-07-23
[Snippets] add snippet that shows how to decompose glyphs in a TTF

justvanrossum:
Snippet based on https://github.com/fonttools/fonttools/issues/2029#issuecomment-662816227

### DIFF
--- a/Snippets/decompose-ttf.py
+++ b/Snippets/decompose-ttf.py
@@ -1,3 +1,9 @@
+#! /usr/bin/env python3
+
+# Example script to decompose the composite glyphs in a TTF into
+# non-composite outlines.
+
+
 import sys
 from fontTools.ttLib import TTFont
 from fontTools.pens.recordingPen import DecomposingRecordingPen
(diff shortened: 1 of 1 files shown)


anthrotype:
i'm going to add code to remove overlaps using skia-pathops to this snippet

arrowtype:
> Curious why would one want to decompose a TTF in the first place? Or maybe it's just to show how the decomposing pen works

Partly, it’s a way to hack “code ligatures” to make their positioning work in strict monospace-only environments (https://github.com/arrowtype/recursive/pull/354). I think I _could_ alternatively set their transformation positions different, instead. But, I also ended up being very curious to understand how a decomposing pen might work in a TTF, as I’m still connecting the dots for how to manipulate fonts in various formats. Thanks so much for the help here!

arrowtype:
This works beautifully in [my script](https://github.com/arrowtype/recursive/blob/92fc0a7aa209841dc9ba77c0c30b50a67fca7d2d/src/build-scripts/make-release/dlig2calt.py). Thanks again!

PR #2034, opened 2020-07-27
Validate required attributes and their types

simoncozens:
I generated a (bad) designspace file in an app I've been working on, and was greeted by this from fontmake:

```
fontmake: Error: In 'foo.designspace': Reading Designspace failed: float() argument must be a string or a number, not 'NoneType'
```

Well, thanks. This patch validates the attributes of designspace file elements and provides a more helpful error message.

(Note that I use %-formats instead of f-strings to match existing code style.)

### DIFF
--- a/Lib/fontTools/designspaceLib/__init__.py
+++ b/Lib/fontTools/designspaceLib/__init__.py
@@ -919,7 +919,7 @@ class BaseDocReader(LogMixin):
                 self.log.warning("Location with undefined axis: \"%s\".", dimName)
                 continue
             xValue = yValue = None
-            self.validateAttributes(dimensionElement, { "xvalue": float })
+            self.validateAttributes(dimensionElement, {"xvalue": float})
             xValue = float(dimensionElement.attrib.get('xvalue'))
             try:
                 yValue = dimensionElement.attrib.get('yvalue')
--- a/Lib/fontTools/designspaceLib/__init__.py
+++ b/Lib/fontTools/designspaceLib/__init__.py
@@ -735,6 +735,21 @@ class BaseDocReader(LogMixin):
         self.readInstances()
         self.readLib()
 
+    def validateAttributes(self, element, attributeTypeMap):
+        for attributeName, attributeType in attributeTypeMap.items():
+            attrib = element.attrib.get(attributeName)
+            if attrib is None:
+                raise DesignSpaceDocumentError(
+                    "<%s> element missing required attribute '%s'"
+                    % (element.tag, attributeName))
+            try:
+                attributeType(attrib)
+            except Exception as e:
+                raise DesignSpaceDocumentError(
+                    "<%s> element attribute '%s' could not be converted "
+                    "to type %s"
+                    % (element.tag, attributeName, str(attributeType)))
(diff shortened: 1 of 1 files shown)


Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -735,6 +735,21 @@ def read(self):
...
+    def validateAttributes(self, element, attributeTypeMap):
+        for attributeName, attributeType in attributeTypeMap.items():
+            attrib = element.attrib.get(attributeName)
+            if attrib is None:
+                raise DesignSpaceDocumentError(
+                    "<%s> element missing required attribute '%s'"
+                    % (element.tag, attributeName))
+            try:
+                attributeType(attrib)
+            except Exception as e:
```
anthrotype, line 747: i think you should only catch TypeError and ValueError here, not blind Exception

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -735,6 +735,21 @@ def read(self):
         self.readInstances()
         self.readLib()
 
+    def validateAttributes(self, element, attributeTypeMap):
+        for attributeName, attributeType in attributeTypeMap.items():
+            attrib = element.attrib.get(attributeName)
+            if attrib is None:
+                raise DesignSpaceDocumentError(
+                    "<%s> element missing required attribute '%s'"
```
anthrotype, line 743: the required-ness of attributes is a pain point in designspace spec and lib (I commented on that somewhere, forgot where exactly). I believe the code was written deliberately to allow most of the attributes to be optional (you can tell by the many `.get(something)` that don't handle `None`-ness). So basically clients are supposed to check for that and handle missing data.
For example, here you made both axis.name and axis.tag required; I think varLib works even if one of the two is missing (attempting to derive one from the other).
Before we enforce required-ness with an exception like you did here, we should review thoroughly all the piece of data that we think should be required, make that explicit in the spec and adjust the code accordingly (while avoiding possible breakages in other tools).

simoncozens: An XML schema for designspace would be super helpful...

simoncozens: I don't know how entangled designspace and UFO are (the use of `lib` and `info` suggest that they are, and I was quite surprised when I realised that all the filenames could only refer to .ufo files, not .ttfs), but if so maybe that's something to bring up at Friday's UFO meeting.

anthrotype: with varLib as consumer, source filenames can also be .ttf or even .ttx files.

> maybe that's something to bring up at Friday's UFO meeting

there's a lot on the table already but sure, you can try

simoncozens:
I think I want to actually fix this in a different way, with a more generic designspace validator module. There are many more tests we can do on designspace files/objects to make sure that they make sense semantically, not just that they are syntactically valid - for example, checking for repeated axes, values out of range, multiple masters/instances at the same positions, etc.

benkiel:
@simoncozens should this be closed then?

PR #2035, opened 2020-08-02
Remove import of py23 fron subset/__init__.py

jkseppan:
This import is causing an unsightly DeprecationWarning.

I checked manually: the only names being used from py23 are
open, range and zip, which are defined to be the same as the
corresponding Python 3 builtins, so this should cause no
visible change except for suppressing the warning.


anthrotype:
Thanks

PR #2037, opened 2020-08-07
[feaLib] Allow anonymous glyphclass in LookupFlags

khaledhosny:
Almost everywhere else when a glyph class name is accepted, anonymous glyph classes are also accepted. This is rather inconsistent and inconvenient.


khaledhosny:
Corresponding AFDKO PR https://github.com/adobe-type-tools/afdko/pull/1206

PR #2039, opened 2020-08-12
[feaLib] Add support for STAT table

punchcutter:
Add STAT table support as discussed in https://github.com/adobe-type-tools/afdko/issues/176

One issue not yet updated in afdko is discussed here: https://github.com/adobe-type-tools/afdko/issues/1197
This change is already reflected in the feaLib implementation.


Tests/feaLib/builder_test.py
```diff
@@ -453,6 +453,140 @@ def test_chain_pos_refrences_GSUB_looup(self):
...
+            '   nameid 256 "Roman"; '
+            '} name;'
+            'table STAT {'
+            '    ElidedFallbackName { name "Roman"; };'
+            '    DesignAxis opsz 0 { name "Optical Size"; };'
+            '    DesignAxis opsz 1 { name "Optical Size"; };'
+            '} STAT;')
+
+    def test_STAT_design_axis_multilingual_name(self):
+        print(self.build(
```
moyogo, line 484: This should assert() the table is correct instead of print().

Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -332,6 +334,21 @@ def xmlWrite(self, xmlWriter, font, value, name, attrs):
 		xmlWriter.newline()
 
 
+class STATFlags(UShort):
```
anthrotype, line 337: Maybe a generic NamedFlags converter that could be defined in otConverters and used anywhere there are some bit flags that we want to be accompanied by a comment with the flag names. Perhaps add a TODO for later.

Lib/fontTools/feaLib/builder.py
```diff
@@ -494,6 +500,176 @@ def checkattr(table, attrs):
...
+            names_, locations_, flags_ = record
+            if sorted(names_) == sorted(names) and sorted(locations_) == \
+                    sorted(locations) and sorted(flags_) == sorted(flags):
+                raise FeatureLibError(
+                    'Duplicate AxisValueRecords are not allowed.',
+                    location,
+                )
+        self.stat_["AxisValueRecords"].append((names, locations, flags))
+
+    def build_STAT(self):
```
khaledhosny, line 533: We recently got `otlLib.builder.buildStatTable()`, can’t it be used here?

Lib/fontTools/feaLib/ast.py
```diff
@@ -77,6 +81,15 @@ def deviceToString(device):
         return "<device %s>" % ", ".join("%d %d" % t for t in device)
 
 
+def statNameToString(name_rec):
+    platformID, platEncID, langID, string = name_rec
+    if platformID == 3 and platEncID == 1 and langID == 0x0409:
+        return f'name "{string}";'
+    string = string.encode('unicode-escape').decode().replace('\\u', '\\').upper()
+    langID = '0x%04x' % langID
+    return f'name {platformID} {platEncID} {langID} "{string}";'
```
khaledhosny, line 90: There are many name entries elsewhere in this module (`NameRecord`, `FeatureNameStatement`, `CVParametersNameStatement`) and they don’t do any escaping. I suggest you try to match their code for now, and open a separate issue if you think it should do escaping. It is better to be consistent here, IMO.

Lib/fontTools/feaLib/ast.py
```diff
@@ -1861,3 +1874,103 @@ def asFea(self, indent=""):
...
         return "{} {};".format(keywords[self.key], self.value)
+
+
+class STATDesignAxis(Statement):
+    """STAT DesignAxis
+
+    Args:
+        tag: a 4 letter axis tag
+        order: an int to specify Axis ordering
+        names: a list of name record tuples, eg (3, 1, 1, "Bold")
```
khaledhosny, line 1885: Instead of using a tuple for name record, it might be better to subclass `FeatureNameStatement` and override its `build` method.

Lib/fontTools/feaLib/ast.py
```diff
@@ -1861,3 +1874,103 @@ def asFea(self, indent=""):
...
+
+    def asFea(self, indent=""):
+        res = f'DesignAxis {self.tag} {self.axisOrder} {{ '
+        for name_rec in self.names:
+            res += statNameToString(name_rec)
+        res += f' }};'
+        return res
+
+
+class ElidedFallbackName(Statement):
```
khaledhosny, line 1905: Same as above.

Tests/feaLib/builder_test.py
punchcutter: Argh, I thought I removed that whole thing. The "test" is in another place.

Lib/fontTools/feaLib/builder.py
```diff
@@ -1261,6 +1437,11 @@ def add_hhea_field(self, key, value):
     def add_vhea_field(self, key, value):
         self.vhea_[key] = value
 
+    def add_vmtx_field(self, key, glyph, value):
```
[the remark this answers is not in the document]
punchcutter: Nothing, but I had started doing some vmtx stuff on the same branch and failed to remove this part.

Lib/fontTools/feaLib/builder.py
moyogo: It could be used but the logic of `otLib.build.buildStatTable()` is rather different from what FEA syntax does, which is closer to what AxisValueRecords do, and it also doesn’t support languageIDs.
Basically `otLib.build.buildStatTable()` was written to be easy to use, but is not easy to use with FEA syntax.

anthrotype: Thanks Khaled for reminding me of that. Yes, we should definitely try to reuse that code in here too, see whether it's possible to modify it to accommodate both.

Lib/fontTools/feaLib/ast.py
anthrotype: agreed

Lib/fontTools/feaLib/ast.py
```diff
@@ -77,6 +81,15 @@ def deviceToString(device):
         return "<device %s>" % ", ".join("%d %d" % t for t in device)
 
 
+def statNameToString(name_rec):
+    platformID, platEncID, langID, string = name_rec
+    if platformID == 3 and platEncID == 1 and langID == 0x0409:
+        return f'name "{string}";'
+    string = string.encode('unicode-escape').decode().replace('\\u', '\\').upper()
```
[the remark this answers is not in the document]
punchcutter: We don't actually need any of this. At first I was trying to make sure tests and round-tripping were working to some degree, but feaLib can't actually do that currently so I dropped all of this.

Lib/fontTools/feaLib/ast.py
```diff
@@ -1861,3 +1874,103 @@ def asFea(self, indent=""):
...
+        self.location = location
+
+    def build(self, builder):
+        builder.addDesignAxis(self.location, self.tag, self.axisOrder, self.names)
+
+    def asFea(self, indent=""):
+        res = f'DesignAxis {self.tag} {self.axisOrder} {{ '
+        for name_rec in self.names:
+            res += statNameToString(name_rec)
+        res += f' }};'
```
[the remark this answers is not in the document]
punchcutter: Tell that to Felipe. He likes to keep the f in fontbakery.

Lib/fontTools/feaLib/ast.py
```diff
@@ -1861,3 +1874,103 @@ def asFea(self, indent=""):
...
+        self.names = names
+        self.values = values
+        self.flags = flags
+        self.location = location
+
+    def build(self, builder):
+        builder.addAxisValueRecord(self.location, self.names, self.values, self.flags)
+
+    def asFea(self, indent=""):
+        res = f''
```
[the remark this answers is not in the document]
punchcutter: Ok, I prefer single quotes as you can see, but I'm fine doing double to try to be more consistent.

Lib/fontTools/feaLib/parser.py
```diff
@@ -1283,6 +1284,150 @@ def parse_table_OS_2_(self, table):
...
+            log.warning(
+                f'Unregistered axis tag {axis_tag} should be uppercase.'
+            )
+        axis_order = self.expect_number_()
+        axis_names = self.expect_stat_name_()
+        return axis_tag, axis_order, axis_names
+
+    def parse_STAT_axis_value_(self):
+        assert self.is_cur_keyword_('AxisValue')
+        block = self.ast.NestedBlock(
```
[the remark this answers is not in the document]
punchcutter: We don't actually need it at all.

Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -332,6 +334,21 @@ def xmlWrite(self, xmlWriter, font, value, name, attrs):
...
+	def xmlWrite(self, xmlWriter, font, value, name, attrs):
+		xmlWriter.simpletag(name, attrs + [("value", value)])
+		if value:
+			xmlWriter.write("  ")
+			stat_flags = {
+				1: 'OlderSiblingFontAttribute',
+				2: 'ElidableAxisValueName',
+				3: 'OlderSiblingFontAttribute ElidableAxisValueName',
```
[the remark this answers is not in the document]
punchcutter: Updated all of this to do the same as lookup flags.

Lib/fontTools/ttLib/tables/otConverters.py
punchcutter: Updated to do the same as lookup flags.

Lib/fontTools/feaLib/ast.py
punchcutter: Done. Dropped all this and used FeatureNameStatement.

Lib/fontTools/feaLib/ast.py
punchcutter: Actually, there's no reason to subclass FeatureNameStatement. These can just be FeatureNameStatement objects because they behave the same.

khaledhosny: I was concerned that its `build()` method might not be appropriate for `STAT` table, but if it works then all good.

Lib/fontTools/feaLib/parser.py
```diff
@@ -1283,6 +1284,150 @@ def parse_table_OS_2_(self, table):
...
+    def parse_STAT_location(self):
+        values = []
+        loc_tag = self.expect_tag_()
+        if len(loc_tag.strip()) != 4:
+            raise FeatureLibError(f'Axis tag {self.cur_token_} must be 4 '
+                                  f'characters', self.cur_token_location_)
+
+        while self.next_token_ != ';':
+            if self.next_token_type_ in [Lexer.NUMBER, Lexer.FLOAT]:
+                value = self.expect_stat_values_()
```
khaledhosny, line 1345: I‘m not sure why do you need `expect_stat_values_()` at all, since you are already checking the token type, you can simplify this to `self.advance_lexer_()` and read the value from `self.cur_token_`.

punchcutter: I made it separate originally because we also had the hyphen for ranges and it seemed worth it to move to a separate `expect_`, but you're right, there's no need now.

Lib/fontTools/feaLib/ast.py
```diff
@@ -77,6 +81,15 @@ def deviceToString(device):
         return "<device %s>" % ", ".join("%d %d" % t for t in device)
 
 
+def statNameToString(name_rec):
+    platformID, platEncID, langID, string = name_rec
+    if platformID == 3 and platEncID == 1 and langID == 0x0409:
+        return f'name "{string}";'
```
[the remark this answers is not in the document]
punchcutter: redid all of this to match the rest of feaLib

Lib/fontTools/feaLib/builder.py
```diff
@@ -494,6 +500,176 @@ def checkattr(table, attrs):
         if version >= 5:
             checkattr(table, ("usLowerOpticalPointSize", "usUpperOpticalPointSize"))
 
+    def set_ElidedFallbackName(self, names):
```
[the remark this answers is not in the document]
punchcutter: updated everything (I think)

Lib/fontTools/feaLib/ast.py
punchcutter: Actually, you're right. The `build()` method for FeatureNameStatement does some stuff we don't want to do with STAT names so I subclassed NameRecord to get what we need.

Tests/feaLib/builder_test.py
```diff
@@ -141,13 +142,20 @@ def check_feature_file(self, name):
...
+        # TODO: Do better matching than this. Whitespace is mostly irrelevant
+        # except if it's inside a string. Name ID numbers can be hex, but are
+        # converted to ints so the diff fails. Name strings are allowed to be
+        # escaped in the source, but they become unescaped and diff fails.
         font = makeTTFont()
         fname = (name + ".fea") if '.' not in name else name
         p = parser(self.getpath(fname), glyphNames=font.getGlyphOrder())
         doc = p.parse()
         actual = self.normal_fea(doc.asFea().split("\n"))
+        actual = " ".join(actual).replace("\n", "").split(";")
```
anthrotype, line 154: what are you trying to do here? Why split at semicolons?
Doesn't the `normal_fea` method already return list of lines stripped of the ending newline character?

Tests/feaLib/runit.py
```diff
@@ -0,0 +1,60 @@
...
+
+def build(featureFile, tables=None):
+    font = makeTTFont()
+    addOpenTypeFeatures(font, featureFile, tables=tables)
+    return font
+
+
+if __name__ == "__main__":
+    feafile = sys.argv[1]
+    build(feafile)
```
anthrotype, line 59: what is this `runit.py` script?

Lib/fontTools/feaLib/builder.py
anthrotype: @punchcutter have you tried to see if there's anything in otlLib that you can reuse in here?
https://github.com/fonttools/fonttools/blob/bec19252c5b424a88e1dfeb423fb2c5068fb36e9/Lib/fontTools/otlLib/builder.py#L2399-L2590

Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -52,6 +52,8 @@ def buildConverters(tableSpec, tableNamespace):
 			converterClass = FeatureParams
 		elif name in ("CIDGlyphMapping", "GlyphCIDMapping"):
 			converterClass = StructWithLength
+		elif name == "Flags":
+			converterClass = STATFlags
```
anthrotype, line 56: remove this `elif` block, add a `STATFlags: STATFlags` to the `converterMapping` below, then change the entry for the STAT Flags in `otData.py` module so that the type of the converter is `STATFlags` instead of `uint16`

Tests/feaLib/builder_test.py
punchcutter: Like I put in the TODO above this way of comparing files has all kinds of problems so the comparison only works for some things and assumes all lines are exactly the same. Splitting on semicolons makes more sense than newlines because semicolons have meaning, but newlines can be arbitrary. Also, spaces, including newlines, are only relevant within a string so to truly compare fea we should also ignore all spaces except in strings. That's why I only did this minor adjustment and added a TODO. This can be much better, but isn't necessary right now.

Tests/feaLib/runit.py
punchcutter: Oops. Not really sure where that came from. Deleted.

Lib/fontTools/ttLib/tables/otConverters.py
punchcutter: Sounds good.

Lib/fontTools/feaLib/builder.py
punchcutter: The base build code was taken from there since it was already there. Of course it's possible to merge the two, but I wanted to get this working in feaLib first before messing with that.

Lib/fontTools/feaLib/builder.py
```diff
@@ -500,156 +499,160 @@ def checkattr(table, attrs):
...
+            if token in self.stat_:
+                raise FeatureLibError(
+                    f"{token} is already set.",
+                    location,
+                )
         self.stat_["ElidedFallbackName"] = names
 
-    def set_ElidedFallbackNameID(self, value):
+    def setElidedFallbackNameID(self, value, location):
+        for token in ("ElidedFallbackName", "ElidedFallbackNameID"):
```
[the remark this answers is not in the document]
punchcutter: Ok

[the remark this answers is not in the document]
punchcutter: Actually, makes sense to consolidate these since we can check if the value is an int or list and assign it with that.

Lib/fontTools/feaLib/ast.py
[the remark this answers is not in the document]
felipesanches: @punchcutter that was not exactly my point though. I agree with @anthrotype's suggestion here.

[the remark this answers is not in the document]
felipesanches: In FontBakery codebase we use the f-string prefix on all lines of a multi-line string for the sake of keeping the code's visual alignment "beautiful" in cases like this:

```python
print(f'This is and example in which'
      f' the value of xyz is {xyz} and'
      f' blah blah blah whatever...')
```

but this is not the case here, where the string in `res += ' };'` is isolated in its own statement. There's no need for visual alignment here, so it is completely unrelated to the code-style adopted in fontbakery

Lib/fontTools/feaLib/builder.py
```diff
@@ -508,16 +508,10 @@ def setElidedFallbackName(self, names, location):
...
-            if token in self.stat_:
-                raise FeatureLibError(
-                    f"{token} is already set.",
-                    location,
-                )
-        self.stat_["ElidedFallbackNameID"] = value
+        if isinstance(value, int):
+            self.stat_["ElidedFallbackNameID"] = value
+        if isinstance(value, list):
+            self.stat_["ElidedFallbackName"] = value
```
anthrotype, line 514: even if unlikely it's always worth adding a final else block for self-protection  
```
else:
    raise AssertionError(value)
```

Lib/fontTools/feaLib/builder.py
```diff
@@ -500,156 +499,160 @@ def checkattr(table, attrs):
...
+        self.stat_["AxisValueRecords"].append(axisValueRecord)
 
     def build_STAT(self):
         if not self.stat_:
             return
         self.font["STAT"] = newTable("STAT")
         table = self.font["STAT"].table = otTables.STAT()
         table.Version = 0x00010001
-        name_table = self.font.get("name")
+        nameTable = self.font["name"]
```
anthrotype, line 552: in bare-bone fonts (e.g. unit tests) the name table may not be present, here's what we do elsewhere in feaLib.builder

https://github.com/fonttools/fonttools/blob/a41005077dc36f27798e64d659934aa3dcdf016a/Lib/fontTools/feaLib/builder.py#L391-L394

Tests/feaLib/builder_test.py
anthrotype: > Splitting on semicolons makes more sense than newlines because semicolons have meaning, but newlines can be arbitrary.

but the arbitrary newlines are the ones that feaLib asFea() itself puts, so at least they should be consistent and predictable. 
Anyway I'm not convinced this is necessary for the present PR. I would prefer if this was addressed separately if at all.

anthrotype (commented):
Apologies if I asked this already, have you tried to see if we can reuse/import code from the buildStatTable in otlLib?

Since this is quite a lot of code that touches on several places, i'd like someone else (e.g. @justvanrossum @simoncozens @khaledhosny @dscorbett) to also give their LGTM

Lib/fontTools/feaLib/builder.py
anthrotype: yes I agree

Lib/fontTools/feaLib/builder.py
justvanrossum: Also, that second if should be an elif.

Lib/fontTools/feaLib/builder.py
justvanrossum: I also agree, and would be happy to review changes or refactorings to `fontTools.otlLib.builder.buildStatTable()` to make this possible.

punchcutter: Ok, my initial thought was to get this in ASAP so we'd have it for some existing projects that are already set up with STAT in the feature file. And then after that definitely do a refactoring at some point. Since this is already quite a lot of new stuff I felt like it might be good to leave the otlLib code alone for a first round, but either way is fine.

Tests/feaLib/builder_test.py
punchcutter: Ok, fine to drop it from this PR.

Lib/fontTools/feaLib/builder.py
punchcutter: Replaced

Lib/fontTools/feaLib/builder.py
punchcutter: Updated

Tests/feaLib/data/STAT_test.fea
```diff
@@ -9,10 +9,14 @@ table STAT {
                          name 3 1 1041 "ローマン";
                          };
 
-    DesignAxis opsz 0 { name "Optical Size"; };
-    DesignAxis wdth 1 { name "Width"; };
-    DesignAxis wght 2 { name "Weight"; };
-    DesignAxis ital 3 { name "Italic"; };  # here comment
+    DesignAxis opsz 0 { name "Optical Size";
+    };
```
anthrotype, line 13: why this?

punchcutter: Why what? New line? I just updated now so it's consistent with the rest. In practice people can do whatever they want with the whitespace, but here we expect newlines for each name entry since there can be multiple.

Lib/fontTools/feaLib/ast.py
```diff
@@ -62,6 +66,9 @@
     "SingleSubstStatement",
     "SizeParameters",
     "Statement",
+    "STATAxisValueRecord",
+    "STATDesignAxis",
```
anthrotype, line 70: you'll notice the convention in feaLib.ast module is to name the classes with an ending "*Statement" when they consist of a "statement" (i.e. a line of FEA ending with semicolon). It may be good to keep that convention for consistency. All the new classes defined here are statements in this sense. Mind renaming them, even though the names becomes a bit long?

Lib/fontTools/feaLib/ast.py
```diff
@@ -1882,3 +1900,123 @@ def asFea(self, indent=""):
...
+            for i in range(len(flags)):
+                if self.flags & curr != 0:
+                    flagStrings.append(flags[i])
+                curr = curr << 1
+            res += f"flag {' '.join(flagStrings)};\n"
+        res += "};"
+        return res
+
+
+class AxisValueLocation(NamedTuple):
```
anthrotype, line 2013: any reason why this is a namedtuple and not an ast.Statement like the others?
I think you turn it into a statement object and add asFea() method as well, so STATAxisValueRecord can use that in its asFea

anthrotype (commented):
Thank you for your first contribution Kamile!
I added some inline review comments below, please take a look.

After all reviewers have added their LGTM, do you mind also running the modules you touched through the [black](https://github.com/psf/black) auto-formatter, to normalize minor whitespace issues? Not before we're ready to merge though, otherwise we can't distinguish what changed since the last review pass.

Lib/fontTools/feaLib/builder.py
```diff
@@ -510,6 +515,114 @@ def checkattr(table, attrs):
...
+
+        axis = self.stat_.get("DesignAxes")
+        if not axis:
+            raise FeatureLibError('DesignAxes not defined', None)
+        axisValueRecords = self.stat_.get("AxisValueRecords")
+        axisValues = {}
+        format4_locations = None
+        for tag in axis:
+            axisValues[tag.tag] = []
+        if axisValueRecords:
```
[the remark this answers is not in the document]
kaydeearts: `axisValueRecords` may return as None instead of an empty list, so I will specify this to be `if axisValueRecords is not None:`. Without the check, there is a _NoneType Object is not iterable_ error

kaydeearts:
@anthrotype Thank you for your review, I've gone through all of your comments & updated the code based on them.

Lib/fontTools/otlLib/builder.py
```diff
@@ -2801,14 +2801,10 @@ def _addName(nameTable, value, minNameID=0):
...
-        for nameRecord in value:
-            if all(hasattr(nameRecord, attr) for attr in
-                   ["string", "nameID", "platformID", "platEncID", "langID"]):
-                nameTable.setName(nameRecord.string,
-                                  nameID,nameRecord.platformID,
-                                  nameRecord.platEncID,nameRecord.langID)
-        return nameID
+        names = {}
+        for name in value:
+            langCode = _WINDOWS_LANGUAGES.get(name.langID)
```
anthrotype, line 2806: hm.. here you're assuming that the namerecord is a windows-platform one and you're only using its langID, i don't think you can make this assumption. 
it kind of was more correct before this last edit... You were actually right, we don't want to use `addMultilingualName` when we already have a list of NameRecords.

What we want to do here is basically append the name records, it turns out that the name table has an `addName` method that does exactly what you want: i.e. adds a string with given triplet encoding and assigns it to an unused nameID slot in the user range > 255.
Just use that instead of the combo _findUnusedNameID + setName.

Also, when _addName gets a list, make sure that all the values are instances of `_n_a_m_e.NameRecord` (raise TypeError if they aren't). Document that in the buildStatTable docstring that `elidedFallbackName` can also be a list of namerecords (in addition to a int or a dict).

Tests/feaLib/data/STAT_test.ttx
```diff
@@ -8,9 +8,15 @@
     <namerecord nameID="256" platformID="3" platEncID="1" langID="0x409">
       Roman
     </namerecord>
+    <namerecord nameID="256" platformID="1" platEncID="0" langID="0x0" unicode="True">
+      Roman
```
anthrotype, line 12: these duplicated entries are an unintentional byproduct of using addMultilingualName early on (which has an option mac=True), when given a list of elidedFallbackName records. We only want to add the records that we are requested, we don't want to automatically add "mac" names unless explicitly requested. Using the simple addName (as I commented above) should do what we want and remove these duplicate entries.

Lib/fontTools/otlLib/builder.py
kaydeearts: @anthrotype I tried using `addName` where it iterates through `value` and is called for each name. However, I realized it gives a new nameID for each name every time, which creates a problem for an ElidedFallBackName with multiple names. For example:

`table STAT {
    ElidedFallbackName {
        name "Roman";
        name 3 1 1041 "ローマン";
    };`

I was going to suggest this, which still uses `addMultilingualName` but specifies which platform is being requested, and keeps the nameID the same for multiple names in ElidedFallbackName:

`        names = {}
        mac = False
        windows = False
        for name in value:
            if name.platformID == 1:
                langCode = _MAC_LANGUAGES.get(name.langID)
                mac = True
            else:
                langCode = _WINDOWS_LANGUAGES.get(name.langID)
                windows = True
            names[langCode] = name.string
        return nameTable.addMultilingualName(names, windows=windows, mac=mac, minNameID=minNameID)`

anthrotype: I see. Ok, then I suggest we revert to using setName like you were doing originally. It's probably fine that you use `_findUnusedNameID` even if underscore-prefixed and intended to be private.

anthrotype:
Thanks, I think this is looking good now. Maybe @justvanrossum would like to take another pass before we run black autoformatter, squash the commits and merge?

justvanrossum (approved):
I didn't manage a proper review, but overall this looks good to me.

anthrotype:
Thanks Just for the quick review.

I just run the `black` command on the feaLib modules that were already blackened in https://github.com/fonttools/fonttools/pull/2039/commits/6b87f5588c9e9cdfce9125e6bead91ba55b38476.

I am going to merge now. Thanks again to @punchcutter who initiated this PR as well as @kaydeearts for completing it! 👍

PR #2041, opened 2020-08-12
Better document rule processing order

### LINKED ISSUE #2040: [otvar] Rule conditionsets conflict with OT layout (shaping)
Considering [this `.designspace`](https://github.com/ctrlcctrlv/Noto-Sans-Tagalog/blob/f73914a8fa0620059a15fc101310e51f8b359033/NotoSansTagalog.designspace#L11) excerpt:

```xml
    <rule name="ra_crossing">
      <conditionset>
        <condition name="weight" minimum="625" maximum="700"/>
      </conditionset>
      <sub name="uni170D" with="uni170D.avar"/>
    </rule>
```

And [this OTLayout feature](https://github.com/ctrlcctrlv/Noto-Sans-Tagalog/blob/f73914a8fa0620059a15fc101310e51f8b359033/features.fea#L50):

```fea
# Ra to archaic Ra
feature hist {
  lookup hist_1 {
    sub uni170D by uni171F;
  } hist_1;
} hist;
```

Why, when `hist` is active, does a `wght` &ge; 625 cause `uni170D`, shaped to `uni171F`, to become `uni170D.avar`?

To confirm that this bug exists, open https://ctrlcctrlv.github.io/baybayin/ , type <kbd>ra</kbd> as input, tick "Archaic _ra_" checkbox, and increase weight above 625.

ctrlcctrlv:
Close #2040

### DIFF
--- a/Doc/source/designspaceLib/readme.rst
+++ b/Doc/source/designspaceLib/readme.rst
@@ -81,7 +81,7 @@ Methods
    location. Returns None if there isn't one.
 -  ``normalizeLocation(aLocation)``: return a dict with normalized axis values.
 -  ``normalize()``: normalize the geometry of this designspace: scale all the
-  locations of all masters and instances to the ``-1 - 0 - 1`` value.
+   locations of all masters and instances to the ``-1 - 0 - 1`` value.
 -  ``loadSourceFonts()``: Ensure SourceDescriptor.font attributes are loaded,
    and return list of fonts.
 -  ``tostring(encoding=None)``: Returns the designspace as a string. Default 
@@ -297,6 +297,8 @@ RuleDescriptor object
 -  Each condition is a dict with ``name``, ``minimum`` and ``maximum`` keys.
 -  ``subs``: list of substitutions
 -  Each substitution is stored as tuples of glyphnames, e.g. ("a", "a.alt").
+-  Note: By default, rules are applied *before* text shaping/OpenType layout. See 
+   `5.0 rules element`_.
 
 Evaluating rules
 ----------------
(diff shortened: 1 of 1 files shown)


Doc/source/designspaceLib/readme.rst
```diff
@@ -297,6 +297,8 @@ RuleDescriptor object
 -  Each condition is a dict with ``name``, ``minimum`` and ``maximum`` keys.
 -  ``subs``: list of substitutions
 -  Each substitution is stored as tuples of glyphnames, e.g. ("a", "a.alt").
+-  Note: By default, rules are applied *before* text shaping/OpenType layout. See 
+   `5.0 rules element`_.
```
ctrlcctrlv, line 301: Sorry, I tried hard but reStructuredText is very hard to use. I can't figure out how to make this link go to &sect; 5.0 rules element but say "the `<rules>` Attributes section".

Doc/source/designspaceLib/readme.rst
```diff
@@ -297,6 +297,8 @@ RuleDescriptor object
 -  Each condition is a dict with ``name``, ``minimum`` and ``maximum`` keys.
 -  ``subs``: list of substitutions
 -  Each substitution is stored as tuples of glyphnames, e.g. ("a", "a.alt").
+-  Note: By default, rules are applied *before* text shaping/OpenType layout. See
```
anthrotype, line 300: It's not applied "before text shaping", it is still part of that process, only it occurs at the beginning of it, before the rest of GSUB substitions.

https://docs.microsoft.com/en-us/typography/opentype/spec/features_pt#-tag-rvrn

> should be processed early in GSUB processing, before application of the localized forms feature or features related to shaping of complex scripts or discretionary typographic effects.

Maybe just say that by default they are encoded using `rvrn` (which is applied before all others), unless processing="last" is set. In the latter case, it uses `rclt` (just an arbitrary GSUB feature really that is always on basically) which is applied last because we add the rules lookup at the end of the lookup list and rclt follows the specified order.

Doc/source/designspaceLib/readme.rst
anthrotype: it's ok, I don't know either :)

PR #2049, opened 2020-08-20
[varLib.mutator] Prune nameIDs from STAT table

khaledhosny:
`STAT` table might contain names not in `fvar` table (e.g. `ital` axis in variable font with no italic variations).

### DIFF
--- a/Lib/fontTools/varLib/mutator.py
+++ b/Lib/fontTools/varLib/mutator.py
@@ -360,6 +360,12 @@ def instantiateVariableFont(varfont, location, inplace=False, overlap=True):
 		for i in fvar.instances:
 			exclude.add(i.subfamilyNameID)
 			exclude.add(i.postscriptNameID)
+		if 'STAT' in varfont:
+			stat = varfont['STAT'].table
+			if stat.DesignAxisRecord:
+				exclude.update(a.AxisNameID for a in stat.DesignAxisRecord.Axis)
+			if stat.AxisValueArray:
+				exclude.update(v.ValueNameID for v in stat.AxisValueArray.AxisValue)
 		if 'ltag' in varfont:
 			# Drop the whole 'ltag' table if all its language tags are referenced by
 			# name records to be pruned.


anthrotype:
varLib.mutator is supposedly superseded by varLib.instancer. I'm ok to backport low-hanging things from the latter to the former, but any new behavior should go to instancer first. Does varLib.instancer do the right thing in this case?

khaledhosny:
Never used `instancer` to be honest, and I was a bit confused about the relation of the two. If `mutator` is deprecated (or does superseded here mean something different?) then this needs to be communicated. Now I need to figure out how to use `instancer`.

khaledhosny:
`instancer` does not drop the `STAT` table, so this change is not relevant to it (and as it happens, I want to drop `STAT` table).

anthrotype:
It should probably be communicated better, my fault. Some of that is in the release notes, and the module docstring of varLib.instancer.
By superseded I mean that it has all the features of mutator (it can make full instances from VFs) plus more (it can also make partial instances, ie. VFs with less axes or restricted min/max ranges).
It works the same way as mutator; only difference is that in mutator when an axis is not specified, it will drop it at its default coordinate; whereas, instancer assumes that you want to keep the omitted axis in the partial VF. If you specify a coordinate for all the fvar axes, both tool should produce a functionally identical static instance.

anthrotype:
At some point, in https://github.com/fonttools/fonttools/commit/3c6ddb0ef8e669e8c1cab81632b5f94841235d66#diff-1426d15a6d51869dfb3a5552b1c0dc49, we decided that varLib.instancer should never drop STAT DesignAxes, but only prune out-of-range AxisValues. The commit message tries to explain the reason.

If you think dropping the STAT table as a whole (and pruning the name table accordingly) is desirable, then maybe we could an option with a list of table to drop.

In general I think it's not advisable that we continue to maintain two very similar tools, so I propose that we deprecate and eventually drop mutator in favour of instancer, after carefully re-reviewing any features from the old tool that may be missing from the new one. Definitely we should try to avoid that they drift futher away from one another.

punchcutter:
Why would you drop the STAT table? Or shouldn't that be optional? Also, keeping something like `ital` when it's not in the fvar makes sense because it's something that applies to the family (whether static or variable), not just the single instance.

khaledhosny:
For the fonts I’m instantiating (seperate variable upright and italic with only weight axis), `STAT` table does not provide much value and `varLib.mutator` is already dropping the table, so that is suitable for me. The variable fonts has `STAT` table with `ital` axis, but no corresponding `fvar` axis. `varLib.mutator` prunes name IDs used in `fvar`, but it was keeping behind the name IDs used only from `STAT` table, this change fixes this.

anthrotype:
exactly because it is arguable whether the table provides much value for specific cases that the instancer tool doesn't make that decision on behalf of the user. The change is fairly minimal, so I'd be ok to LGTM this one. I'm only concerned that we will end up maintaining the two tools with overlapping features, but slightly divergent behaviour such as in this case. That's all.
Also, it should be trivial to write a special purpose script that does just that (drops STAT and prunes name).

khaledhosny:
OK, I switched to `varLib.instancer` and dropped the `STAT` table myself.

anthrotype:
Noted. I hope I didn't sound too dismissive.. If there's demand for dropping the STAT table in instancer, I'm open to PRs that add options along those lines.

punchcutter:
I think it's a good option to have in instancer to drop or keep. The only place I know for sure that the STAT will make a difference in the static font is on Windows where the STAT will be used to construct the menu name.

PR #2054, opened 2020-09-03
[varLib] Test for interpolatability of paths

simoncozens:
Today I was surprised to run `fontTools varLib.interpolatable` on a font, get a pass, and then find that `fontTools varLib` could not in fact interpolate the font. Here are some more checks in `interpolatable`.

* Checks node count of each path
* Checks operation type of each path node

### DIFF
--- a/Lib/fontTools/varLib/interpolatable.py
+++ b/Lib/fontTools/varLib/interpolatable.py
@@ -104,6 +104,7 @@ def test(glyphsets, glyphs=None, names=None):
 
 		try:
 			allVectors = []
+			allNodeTypes = []
 			for glyphset,name in zip(glyphsets, names):
 				#print('.', end='')
 				glyph = glyphset[glyph_name]
@@ -114,8 +115,11 @@ def test(glyphsets, glyphs=None, names=None):
 				del perContourPen
 
 				contourVectors = []
+				nodeTypes = []
+				allNodeTypes.append(nodeTypes)
 				allVectors.append(contourVectors)
 				for contour in contourPens:
+					nodeTypes.append(tuple([ instruction[0] for instruction in contour.value ]))
 					stats = StatisticsPen(glyphset=glyphset)
 					contour.replay(stats)
 					size = abs(stats.area) ** .5 * .5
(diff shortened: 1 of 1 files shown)


Lib/fontTools/varLib/interpolatable.py
```diff
@@ -114,8 +115,11 @@ def test(glyphsets, glyphs=None, names=None):
 				del perContourPen
 
 				contourVectors = []
+				nodeTypes = []
+				allNodeTypes.append(nodeTypes)
 				allVectors.append(contourVectors)
 				for contour in contourPens:
+					nodeTypes.append(tuple([ instruction[0] for instruction in contour.value ]))
```
anthrotype, line 122:
```suggestion
					nodeTypes.append(tuple(instruction[0] for instruction in contour.value))
```

Lib/fontTools/varLib/interpolatable.py
```diff
@@ -131,6 +135,23 @@ def test(glyphsets, glyphs=None, names=None):
 					#print(vector)
 
 			# Check each master against the next one in the list.
+			for i,(m0,m1) in enumerate(zip(allNodeTypes[:-1],allNodeTypes[1:])):
```
anthrotype, line 138:
```suggestion
			for i, (m0, m1) in enumerate(zip(allNodeTypes[:-1], allNodeTypes[1:])):
```

Lib/fontTools/varLib/interpolatable.py
```diff
@@ -131,6 +135,23 @@ def test(glyphsets, glyphs=None, names=None):
...
+				if len(m0) != len(m1):
+					print('%s: %s+%s: Glyphs not compatible (wrong number of paths %i+%i)!!!!!' % (glyph_name, names[i], names[i+1], len(m0), len(m1)))
+				if m0 == m1:
+					continue
+				for pathIx, (nodes1, nodes2) in enumerate(zip(m0,m1)):
```
anthrotype, line 143:
```suggestion
				for pathIx, (nodes1, nodes2) in enumerate(zip(m0, m1)):
```

anthrotype (approved):
thanks!

(then some day we'll turn all those print() into proper logging messages)

PR #2055, opened 2020-09-07
Add registry for custom table packer/unpacker classes

justvanrossum:
Rationale:
- Currently, all sfnt table packer/unpacker classes must live in `fontTools.ttLib.tables`
- Sometimes a project can benefit from private table data, and would like to use the standard TTFont table API, without needing an explicit step to pack/unpack
- Proposals for new OpenType tables would benefit from this, as experimentation can then easily be done without forking fonttools.

Implementation:
```python
def registerCustomTableClass(tag, moduleName, className=None):
    """Register a custom packer/unpacker class for a table.
    The 'moduleName' must be an importable module. If no 'className'
    is given, it is derived from the tag, for example it will be
    table_C_U_S_T_ for a 'CUST' tag.
    """
    ...
```

I considered taking the simpler route by having the register function take a class object, but that would prevent lazy importing.

### DIFF
--- a/Tests/ttLib/ttFont_test.py
+++ b/Tests/ttLib/ttFont_test.py
@@ -1,0 +1,42 @@
+import io
+from fontTools.ttLib import TTFont, newTable, registerCustomTableClass
+from fontTools.ttLib.tables.DefaultTable import DefaultTable
+
+
+class CustomTableClass(DefaultTable):
+
+    def decompile(self, data, ttFont):
+        self.numbers = list(data)
+
+    def compile(self, ttFont):
+        return bytes(self.numbers)
+
+    # not testing XML read/write
+
+
+table_C_U_S_T_ = CustomTableClass  # alias for testing
+
+
+TABLETAG = "CUST"
+
+
+def test_registerCustomTableClass():
+    font = TTFont()
+    font[TABLETAG] = newTable(TABLETAG)
+    font[TABLETAG].data = b"\x00\x01\xff"
+    f = io.BytesIO()
+    font.save(f)
+    f.seek(0)
+    assert font[TABLETAG].data == b"\x00\x01\xff"
+    registerCustomTableClass(TABLETAG, "ttFont_test", "CustomTableClass")
+    font = TTFont(f)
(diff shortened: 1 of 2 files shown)


Tests/ttLib/ttFont_test.py
```diff
@@ -0,0 +1,44 @@
...
+
+def test_registerCustomTableClass():
+    font = TTFont()
+    font[TABLETAG] = newTable(TABLETAG)
+    font[TABLETAG].data = b"\x00\x01\xff"
+    f = io.BytesIO()
+    font.save(f)
+    f.seek(0)
+    assert font[TABLETAG].data == b"\x00\x01\xff"
+    registerCustomTableClass(TABLETAG, "ttFont_test", "CustomTableClass")
```
anthrotype, line 31: maybe wrap the block between register/unregister in a try... finally clause.
Or if we find ourselves wanting to be able to temporarily register/unregister custom classes outside of self-testing, we could turn the registerCustomTableClass into a context manager that automatically deregisters itself upon exit the `with` block. E.g. you wrap it with contextlib.contextmanager decorator, do the global registering then do a bare `yield` and finally the unregistering. If one doesn't care about the final part, one can call the function as usual and ignore the return value

anthrotype (approved):
that's very useful, thanks Just!

Tests/ttLib/ttFont_test.py
justvanrossum: I literally only added unregister because it messed up the tests in unpredictable ways :) I think normal use cases wouldn't want to unregister ever, so a context manager seems overkill to me.

That said, I'll add the try/except to the tests. I considered and then was too lazy. But you're right, it's better.

justvanrossum: (Btw. simply calling a `contextlib.contextmanager`-wrapped function doesn't execute the pre-yield bit, so for the suggester behavior a context manager class would be needed, with the registry being done in `__init__` instead of `__enter__`.)

anthrotype: ah ok thanks for the checking. Probably not worth making it a full context manager class.

PR #2058, opened 2020-09-10
Disable checksum checking by default

### LINKED ISSUE #2057: Disable checksum-checking
In sfnt.py, the `SFNTReader` constructor defaults to `checkChecksums=1`. I suggest we default to off.

alerque:
Closes #2057.

Also disabled by default for the WOFF2 reader, largely because that's how the API is documented. I have no idea if this makes sense for WOFF2 the same way it does for SFNT.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,6 @@ dmypy.json
 
 # Cython sources (e.g. cu2qu)
 Lib/**/*.c
+
+# Ctags
+tags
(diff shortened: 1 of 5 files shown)


justvanrossum (approved):

anthrotype (commented):
Let's change to bool type (True/False) everywhere as it is more pythonic. Existing scripts that use the int values will continue to work because in python 0/1 have practically the same meaning

alerque:
@anthrotype It's not a boolean value though, the options are 0, 1, or 2. See my commit message for 7e3c4ac.

PR #2063, opened 2020-09-22
clean code and add docstring

matrixgram:


anthrotype:
hello @matrixgram 
what are you trying to do with this PR?

matrixgram (commented):
Hello @anthrotype.
I use autopep8 to reformatting this file and I think change it better:
1. unused variable in def main  -> args
2. add doc strings

And I see added a to-do plan as a comment in def main.
Maybe better if create a to-do list file and write in that.

anthrotype:
Thank you. We plan to use black as autoformatter instead of autopep8. I'll take care of it at some point soon.

PR #2066, opened 2020-09-23
Expose zopfli and zopfli compression level support in ttLib.TTFont.save method

### LINKED ISSUE #2064: Can I convert to woff with zopfli at TTFont?
**Related Topic**
#491

**Describe**
I'm trying to convert the font using `TTFont`.
It works fine.
```python

font = TTFont( SOURCE_PATH )
font.flavor = "woff"
font.save( TARGET_PATH )
```

But..
There is no way to use `zopfli` in document [ttFont Doc](https://rsms.me/fonttools-docs/ttLib/ttFont.html).
What can I do?

chrissimpkins:
Closes #2064 

This is an initial stab at addressing Cosimo's comment in https://github.com/fonttools/fonttools/issues/2064#issuecomment-696866956.

The changes here modify the approach to build zopfli compressed woff format files with `ttLib.TTFont` from:

```python
from fontTools.ttLib import sfnt, TTFont

tt = TTFont(filepath)
tt.flavor = "woff"
sfnt.USE_ZOPFLI = True
tt.save(filepath)
```

to:

```python

tt = TTFont(filepath)
tt.flavor = "woff"
tt.save(filepath, use_zopfli=True)
```

ttx was updated with these changes.

### DIFF
--- a/Lib/fontTools/ttLib/sfnt.py
+++ b/Lib/fontTools/ttLib/sfnt.py
@@ -14,6 +14,7 @@ a table's length chages you need to rewrite the whole file anyway.
 
 from io import BytesIO
 from types import SimpleNamespace
+import warnings
 from fontTools.misc.py23 import Tag
 from fontTools.misc import sstruct
 from fontTools.ttLib import TTLibError
@@ -152,6 +153,12 @@ class SFNTReader(object):
 # default compression level for WOFF 1.0 tables and metadata
 ZLIB_COMPRESSION_LEVEL = 6
 
+# if set to True, use zopfli instead of zlib for compressing WOFF 1.0.
+# The Python bindings are available at https://pypi.python.org/pypi/zopfli
+# Note: this global variable is deprecated, please use the `use_zopfli`
+# parameter instead
+USE_ZOPFLI = None
+
 # mapping between zlib's compression levels and zopfli's 'numiterations'.
 # Use lower values for files over several MB in size or it will be too slow
 ZOPFLI_LEVELS = {
@@ -168,7 +175,7 @@ ZOPFLI_LEVELS = {
 }
 
 
-def compress(data, use_zopfli=False, level=ZLIB_COMPRESSION_LEVEL):
+def compress(data, level=ZLIB_COMPRESSION_LEVEL, use_zopfli=False):
 	""" Compress 'data' to Zlib format. If 'use_zopfli' argument is True,
 	zopfli is used instead of the zlib module.
 	The compression 'level' must be between 0 and 9. 1 gives best speed,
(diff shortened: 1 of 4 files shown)


Lib/fontTools/ttLib/sfnt.py
```diff
@@ -172,16 +168,16 @@ def __setstate__(self, state):
 }
 
 
-def compress(data, level=ZLIB_COMPRESSION_LEVEL):
-	""" Compress 'data' to Zlib format. If 'USE_ZOPFLI' variable is True,
+def compress(data, use_zopfli=False, level=ZLIB_COMPRESSION_LEVEL):
```
dscorbett, line 171: Putting the new parameter before an existing parameter is not backwards-compatible. Previously, `compress(data, 9)` would compress with the best compression level of the module selected by `USE_ZOPFLI`. With this change, the same code will compress with the default compression level of Zopfli.

It is also backwards-incompatible because the current documentation says that setting `USE_ZOPFLI` has an effect, which with this change is no longer true. It would be better to keep `USE_ZOPFLI` but deprecate it and say that the new `use_zopfli` parameter takes precedence if set.

chrissimpkins: Thanks David!  

I attempted to address these issues in 60b47cd244679bb2c015e4f199af6a2b74dde482

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -152,11 +152,14 @@ def close(self):
 		if self.reader is not None:
 			self.reader.close()
 
-	def save(self, file, reorderTables=True):
+	def save(self, file, reorderTables=True, use_zopfli=False, compression_level=ZLIB_COMPRESSION_LEVEL):
```
anthrotype, line 155: the other parameter names are camelCase, these should be too for consistency

Lib/fontTools/ttLib/sfnt.py
```diff
@@ -209,17 +226,20 @@ def __new__(cls, *args, **kwargs):
...
 		self.flavor = flavor
 		self.flavorData = flavorData
+		self.use_zopfli = use_zopfli
+		self.compression_level = compression_level
 
 		if self.flavor == "woff":
 			self.directoryFormat = woffDirectoryFormat
 			self.directorySize = woffDirectorySize
 			self.DirectoryEntry = WOFFDirectoryEntry
+			self.DirectoryEntry.use_zopfli = use_zopfli
```
anthrotype, line 242: setting a class variable is as bad as a global one

Lib/fontTools/ttLib/woff2.py
```diff
@@ -163,7 +163,7 @@ class WOFF2Writer(SFNTWriter):
 	flavor = "woff2"
 
 	def __init__(self, file, numTables, sfntVersion="\000\001\000\000",
-		         flavor=None, flavorData=None):
+		         flavor=None, flavorData=None, use_zopfli=False, compression_level=0):
```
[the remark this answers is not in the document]
anthrotype: me neither...

anthrotype (commented):
I didn't do this originally also because before this I wanted to extract the WOFF related parts into separate WOFFReader/Writer class (like in woff2), out of SFNTReader.
I'm ok in general that TTFont.save() gains new parameters like useZopfli and compressionLevel, but not so much that the generic SFNTReader constructor is changed with those.
Also note that TTFont instantiates SFTNWriter again in reorderFontTables function, the compression options should passed along in there as well otherwise if you reorder you'd lose the zopfli compression.

chrissimpkins:
> I'm ok in general that TTFont.save() gains new parameters like useZopfli and compressionLevel, but not so much that the generic SFNTReader constructor is changed with those.

Thanks for having a look Cosimo.  Should we bag this in favor of the refactor that it needs when someone has an opportunity to do it?

behdad:
Can we use zopfli level=0 to disable compression? Let's resurrect this or close it.

PR #2068, opened 2020-09-23
add module to remove overlaps from TTF with skia-pathops

anthrotype:
this may help for VF generated instances containing overlaps on old macOS (10.11 El Capitan and lower) that seem to ignore the OVERLAP_SIMPLE/OVERLAP_COMPOUND flags,  see https://github.com/google/fonts/issues/2602

Note I haven't tested this, feel free to use and improve on it.

EDIT (30/09/2020):

The PR has changed from a snippet to a full-blown new module called `fontTools.ttLib.removeOverlaps`. This exports a `removeOverlaps` function that takes a TTFont and an optional list of glyphNames to modify (by default it does them all).
Also, the `fonttools varLib.instancer` tool has a new `--remove-overlaps` command-line option that uses the new module to remove overlaps of full VF instances.
Finally, in the `instantiateVariableFont` function in `varLib.instancer` module, the `overlap` option is no longer a `bool` (True/False) value, but is now an enumeration (IntEnum) called `OverlapMode`. The default is the same as before, ie. KEEP_AND_SET_FLAGS. The client can set it to `OverlapMode.REMOVE` to request that overlaps be removed from a full instance (does nothing for partial VFs). The old True/False `overlap` values continue to work as before (they are casted to enum automatically).

Note that the new feature requires the [skia-pathops](https://github.com/fonttools/fonttools) package. I added a new extra requirement called `pathops` (so one can do `pip install fonttools[pathops]`). Alternatively, one can `pip install skia-pathops` separately as usual.

### DIFF
--- a/Lib/fontTools/ttLib/removeOverlaps.py
+++ b/Lib/fontTools/ttLib/removeOverlaps.py
@@ -1,37 +1,29 @@
-#! /usr/bin/env python3
+""" Simplify TrueType glyphs by merging overlapping contours/components.
 
-# Example script to remove overlaps in TTF using skia-pathops.
-# Overlapping components will be decomposed.
+Requires https://github.com/fonttools/skia-pathops
+"""
 
-
-import sys
 from typing import Iterable, Optional, Mapping
+
 from fontTools.ttLib import ttFont
 from fontTools.ttLib.tables import _g_l_y_f
 from fontTools.pens.recordingPen import DecomposingRecordingPen
 from fontTools.pens.ttGlyphPen import TTGlyphPen
 
-try:
-    import pathops
-except ImportError:
-    sys.exit(
-        "This script requires the skia-pathops module. "
-        "`pip install skia-pathops` and then retry."
-    )
+import pathops
+
 
 _TTGlyphMapping = Mapping[str, ttFont._TTGlyph]
 
 
-def skpath_from_simple_glyph(glyphName: str, glyphSet: _TTGlyphMapping) -> pathops.Path:
+def skPathFromSimpleGlyph(glyphName: str, glyphSet: _TTGlyphMapping) -> pathops.Path:
     path = pathops.Path()
     pathPen = path.getPen()
     glyphSet[glyphName].draw(pathPen)
     return path
 
 
-def skpath_from_composite_glyph(
-    glyphName: str, glyphSet: _TTGlyphMapping
-) -> pathops.Path:
+def skPathFromCompositeGlyph(glyphName: str, glyphSet: _TTGlyphMapping) -> pathops.Path:
     # record TTGlyph outlines without components
     dcPen = DecomposingRecordingPen(glyphSet)
     glyphSet[glyphName].draw(dcPen)
@@ -42,7 +34,7 @@ def skpath_from_composite_glyph(
     return path
 
 
-def simple_glyph_from_skpath(path: pathops.Path) -> _g_l_y_f.Glyph:
+def ttfGlyphFromSkPath(path: pathops.Path) -> _g_l_y_f.Glyph:
     # Skia paths have no 'components', no need for glyphSet
     ttPen = TTGlyphPen(glyphSet=None)
     path.draw(ttPen)
@@ -53,21 +45,37 @@ def simple_glyph_from_skpath(path: pathops.Path) -> _g_l_y_f.Glyph:
     return glyph
 
 
-def remove_overlaps(
+def removeOverlaps(
     font: ttFont.TTFont, glyphNames: Optional[Iterable[str]] = None
 ) -> None:
-    if glyphNames is None:
-        glyphNames = font.getGlyphOrder()
+    """ Simplify glyphs in TTFont by merging overlapping contours.
 
-    glyfTable = font["glyf"]
+    Overlapping components are first decomposed to simple contours, then merged.
+
+    Currently this only works with TrueType fonts with 'glyf' table.
+    Raises NotImplementedError if 'glyf' table is absent.
+
+    Args:
+        font: input TTFont object, modified in place.
+        glyphNames: optional iterable of glyph names (str) to remove overlaps from.
+            By default, all glyphs in the font are processed.
+    """
+    try:
+        glyfTable = font["glyf"]
+    except KeyError:
+        raise NotImplementedError("removeOverlaps currently only works with TTFs")
+
     hmtxTable = font["hmtx"]
     glyphSet = font.getGlyphSet()
 
+    if glyphNames is None:
+        glyphNames = font.getGlyphOrder()
+
     for glyphName in glyphNames:
         if glyfTable[glyphName].isComposite():
-            path = skpath_from_composite_glyph(glyphName, glyphSet)
+            path = skPathFromCompositeGlyph(glyphName, glyphSet)
         else:
-            path = skpath_from_simple_glyph(glyphName, glyphSet)
+            path = skPathFromSimpleGlyph(glyphName, glyphSet)
 
         # duplicate path
         path2 = pathops.Path(path)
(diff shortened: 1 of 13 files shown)


anthrotype:
I'm thinking of adding such remove_overlaps feature to the main library, as it could be useful to have as an option for the varLib.instancer -- but I am not sure where to best place it.
It could be a method of TTFont, e.g. `TTFont.removeOverlaps()`, however this currently only works for TTF-flavored opentype fonts, not for CFF or CFF2 or others.
Also, it requires skia-pathops, and I wouldn't want to add a native dependency for a feature that is not "core" so to speak.
So I am thinking of adding a new `removeOverlaps` submodule inside ttLib -- which would fail with `ImportError` if `pathops` module can't be imported.
And to use it, one would do `from fontTools.ttLib.removeOverlaps import removeOverlaps` then call it on a TTFont instance.

anthrotype:
Before I can merge this, I need to figure out a way to have skia-pathops keep the glyph's original path direction to mimize the changes. It's currently always setting it to counter-clockwise.
https://github.com/fonttools/fonttools/pull/2068

Lib/fontTools/ttLib/removeOverlaps.py
```diff
@@ -0,0 +1,117 @@
...
+            path = skPathFromSimpleGlyph(glyphName, glyphSet)
+
+        # duplicate path
+        path2 = pathops.Path(path)
+
+        # remove overlaps
+        path2.simplify()
+
+        # replace TTGlyph if simplified copy is different
+        if path2 != path:
```
chrissimpkins, line 87: It might be worth adding reporting to indicate if any paths were modified during the script run.

chrissimpkins:
> Before I can merge this, I need to figure out a way to have skia-pathops keep the glyph's original path direction to mimize the changes. It's currently always setting it to counter-clockwise.
> #2068

Is this coming from the default `fix_winding = True` parameter setting in `simplify()`?

https://github.com/fonttools/skia-pathops/blob/6581155e5151624ae9756392db025810933b74d6/src/python/pathops/_pathops.pyx#L1393

https://github.com/fonttools/skia-pathops/blob/6581155e5151624ae9756392db025810933b74d6/src/python/pathops/_pathops.pyx#L1119

chrissimpkins:
> > Before I can merge this, I need to figure out a way to have skia-pathops keep the glyph's original path direction to mimize the changes. It's currently always setting it to counter-clockwise.
> > #2068
> 
> Is this coming from the default `fix_winding = True` parameter setting in `simplify()`?
> 
> https://github.com/fonttools/skia-pathops/blob/6581155e5151624ae9756392db025810933b74d6/src/python/pathops/_pathops.pyx#L1393
> 
> https://github.com/fonttools/skia-pathops/blob/6581155e5151624ae9756392db025810933b74d6/src/python/pathops/_pathops.pyx#L1119

I tested this.  It seems to be happening on the skia side.

chrissimpkins:
This approach seems to address the path direction:

```python
        # duplicate path
        path2 = pathops.Path(path)
        pre_clockwise = path2.clockwise

        # remove overlaps
        path2.simplify()

        post_clockwise = path2.clockwise

        if pre_clockwise != post_clockwise:
            path2.reverse()
```

### Pre-simplify



```
path.fillType = FillType.WINDING
path.moveTo(15, 0)
path.lineTo(287, 716)
path.lineTo(383, 716)
path.lineTo(655, 0)
path.lineTo(562, 0)
path.lineTo(365, 535)
path.lineTo(337, 612)
path.lineTo(333, 612)
path.lineTo(305, 535)
path.lineTo(108, 0)
path.close()
path.moveTo(164, 197)
path.lineTo(193, 277)
path.lineTo(477, 277)
path.lineTo(506, 197)
path.close()
```

### Post-simplify, current



```
path.fillType = FillType.WINDING
path.moveTo(15, 0)
path.lineTo(108, 0)
path.lineTo(180.54, 197)
path.lineTo(489.46, 197)
path.lineTo(562, 0)
path.lineTo(655, 0)
path.lineTo(383, 716)
path.lineTo(287, 716)
path.close()
path.moveTo(460.002, 277)
path.lineTo(209.998, 277)
path.lineTo(305, 535)
path.lineTo(333, 612)
path.lineTo(337, 612)
path.lineTo(365, 535)
path.lineTo(460.002, 277)
path.close()
```

### Post-simplify, new



```
path.fillType = FillType.WINDING
path.moveTo(287, 716)
path.lineTo(383, 716)
path.lineTo(655, 0)
path.lineTo(562, 0)
path.lineTo(489.46, 197)
path.lineTo(180.54, 197)
path.lineTo(108, 0)
path.lineTo(15, 0)
path.close()
path.moveTo(460.002, 277)
path.lineTo(365, 535)
path.lineTo(337, 612)
path.lineTo(333, 612)
path.lineTo(305, 535)
path.lineTo(209.998, 277)
path.lineTo(460.002, 277)
path.close()
```

anthrotype:
Thanks Chris for taking a look. I decided to add a new `clockwise=False` option to the pathops methods to control this. 
https://github.com/fonttools/skia-pathops/pull/31
Once I release that, I'll change this PR so that the simplified path remembers the original path's direction, e.g. `path2 = pathops.simplify(path, clockwise=path.clockwise)`

Lib/fontTools/ttLib/removeOverlaps.py
anthrotype: maybe we can print the modified glyph names in a logging.debug message, I don't want to pollute the output with too much extra info. Or maybe just an overall stat like the number of simple vs composite glyphs modified in total.

chrissimpkins: Yeah I agree that a dump of all glyphs would be overwhelming.  A summary that it did / did not change something or the total number of glyphs that were changed may be sufficient.  Alternatively, we could add a check to determine if there are any glyphs with overlaps. That will likely be the info that someone needs pre/post script execution.  If not appropriate here, perhaps as a fontbakery check.

anthrotype: checking for overlaps isn't trivial and and implies a lot of the work of removing the overlaps themselves. that's why here I first do the overlap removal and then check if the simplified path has changed compared to the original. And also that's why I need to minimize the changes to only those strictly needed by the overlap removal process.
At least for composites, I could optimise this by first checking that the bboxes of each component _don't_ overlap in which case I should avoid the overlap-removal. Maybe later.

Lib/fontTools/ttLib/removeOverlaps.py
```diff
@@ -0,0 +1,153 @@
...
+        glyfTable = font["glyf"]
+    except KeyError:
+        raise NotImplementedError("removeOverlaps currently only works with TTFs")
+
+    hmtxTable = font["hmtx"]
+    # wraps the underlying glyf Glyphs, takes care of interfacing with drawing pens
+    glyphSet = font.getGlyphSet()
+
+    if glyphNames is None:
+        glyphNames = font.getGlyphOrder()
```
chrissimpkins, line 97: Maybe filter out glyphs that do not have contours?

anthrotype: good idea, thanks

anthrotype:
I think this is now ready for merge.

The PR has changed from a snippet to a full-blown new module called `fontTools.ttLib.removeOverlaps`. This exports a `removeOverlaps` function that takes a TTFont and an optional list of glyphNames to modify (by default it does them all).
Also, the `fonttools varLib.instancer` tool has a new `--remove-overlaps` command-line option that uses the new module to remove overlaps of full VF instances.
Finally, in the `instantiateVariableFont` function in `varLib.instancer` module, the `overlap` option is no longer a `bool` (True/False) value, but is now an enumeration (IntEnum) called `OverlapMode`. The default is the same as before, ie. KEEP_AND_SET_FLAGS. The client can set it to `OverlapMode.REMOVE` to request that overlaps be removed from a full instance (does nothing for partial VFs). The old True/False `overlap` values continue to work as before (they are casted to enum automatically).

Note that the new feature requires the [skia-pathops](https://github.com/fonttools/fonttools) package. I added a new extra requirement called `pathops` (so one can do `pip install fonttools[pathops]`). Alternatively, one can `pip install skia-pathops` separately as usual.

anthrotype:
And the new module also has a command-line interface `fonttools ttLib.removeOverlaps INPUT.ttf OUTPUT.ttf`

ftCLI:
On the other side, setting `clockwise=True` would correct the contour's winding in case it's wrong. Would be possible to add an option like `--fix-winding` to removeOverlaps and consequently set clockwise to True? Are there downsides that I didn't consider?

anthrotype:
preferably you should open a new issue. I don't understand what you mean

PR #2078, opened 2020-10-06
Instancer: update name table

m4rc1e:
**Warning, still very much a WIP. I need to do further tests and investigate whether we actually need WWS name records etc.**

Currently, `varLib.instancer` will only prune unused name records when instantiating a variable font. This PR will allow users to optionally update the name table by using the STAT table axisValues.

@twardoch I'm wondering whether you'd be interesting in reviewing this PR and its output? your knowledge of application font handling is far superior to mine. I'm currently debating whether I should keep the WWS entries for static instances?

TODO:
 - Write tests and seek out more hardcore edge cases
 - ~~Update uniqueID name records~~
- ~~Ensure multilingual name records actually work. Atm they're simply overwritten.~~
- ~~Support format 4 axisValues~~
- Tidyup/refactor once we're confident the output works across all platforms etc

### DIFF
--- a/Lib/fontTools/varLib/instancer.py
+++ b/Lib/fontTools/varLib/instancer.py
@@ -1024,10 +1024,12 @@ def instantiateSTAT(varfont, axisLimits):
         return  # STAT table empty, nothing to do
 
     log.info("Instantiating STAT table")
-    _instantiateSTAT(stat, axisLimits)
+    newAxisValueTables = axisValuesFromAxisLimits(stat, axisLimits)
+    stat.AxisValueArray.AxisValue = newAxisValueTables
+    stat.AxisValueCount = len(stat.AxisValueArray.AxisValue)
 
 
-def _instantiateSTAT(stat, axisLimits):
+def axisValuesFromAxisLimits(stat, axisLimits):
     location, axisRanges = splitAxisLocationAndRanges(axisLimits, rangeType=AxisRange)
 
     def isAxisValueOutsideLimits(axisTag, axisValue):
@@ -1068,9 +1070,7 @@ def _instantiateSTAT(stat, axisLimits):
         else:
             log.warn("Unknown AxisValue table format (%s); ignored", axisValueFormat)
         newAxisValueTables.append(axisValueTable)
-
-    stat.AxisValueArray.AxisValue = newAxisValueTables
-    stat.AxisValueCount = len(stat.AxisValueArray.AxisValue)
+    return newAxisValueTables
 
 
 def getVariationNameIDs(varfont):
@@ -1130,15 +1130,20 @@ def updateNameTable(varfont, axisLimits):
     #   AxisValues.
     # - Create a dictionary which contains the new default locations for each
     #   axis.
-    # - Duplicate the existing stat table and instantiate it using the dict
-    #   we created in the last step. We should now have a new STAT table
-    #   which only contains AxisValues for the default axis locations.
-    # - Remove any AxisValues which have the Elidable_AXIS_VALUE_NAME set.
-    # - Remove and sort AxisValues so format 4 AxisValues take presedence.
+    # - Create a new list of AxisValues whose Values match the new default
+    #   locations.
+    # - Remove any AxisValues from the list which have the
+    #   Elidable_AXIS_VALUE_NAME set.
+    # - Remove and sort AxisValues in the list so format 4 AxisValues take
+    #   presedence.
 
     #   2. Updating a name table's style and family names from a list of
     #   AxisValues:
-    # - Sort AxisValues into two groups, one for RIBBI, the other for non-RIBBI
+    # - Sort AxisValues into two groups. For the first group, the names must
+    #   any of the following ["Regular", "Italic", "Bold", "Bold Italic"].
+    #   This group of names is often referred to as "RIBBI" names. For the
+    #   other group, names must be non-RIBBI e.g "Medium Italic", "Condensed"
+    #   etc.
     # - Repeat the next steps for each name table record platform:
     #   - Create new subFamily name and Typographic subFamily from the above
     #     groups
@@ -1163,28 +1168,21 @@ def updateNameTable(varfont, axisLimits):
         ):
             defaultAxisCoords[axisTag] = val
 
-    # To get the required Axis Values for the zero origin, we can simply
-    # duplicate the STAT table and instantiate it using the axis coords we
-    # created in the previous step.
-    stat_new = deepcopy(stat)
-    _instantiateSTAT(stat_new, defaultAxisCoords)
-    checkMissingAxisValues(stat_new, defaultAxisCoords.keys())
+    axisValueTables = axisValuesFromAxisLimits(stat, defaultAxisCoords)
+    checkMissingAxisValues(stat, axisValueTables, defaultAxisCoords.keys())
 
-    axisValueTables = stat_new.AxisValueArray.AxisValue
     # Remove axis Values which have ELIDABLE_AXIS_VALUE_NAME flag set.
     # Axis Values which have this flag enabled won't be visible in
     # application font menus.
     axisValueTables = [
         v for v in axisValueTables if not v.Flags & ELIDABLE_AXIS_VALUE_NAME
     ]
-    stat_new.AxisValueArray.AxisValue = axisValueTables
-    axisValueTables = _sortAxisValues(stat_new)
+    axisValueTables = _sortAxisValues(axisValueTables)
     _updateNameRecords(varfont, axisValueTables)
 
 
-def checkMissingAxisValues(stat, axisTags):
+def checkMissingAxisValues(stat, axisValues, axisTags):
     seen = set()
-    axisValues = stat.AxisValueArray.AxisValue
     designAxes = stat.DesignAxisRecord.Axis
     for val in axisValues:
         if val.Format == 4:
@@ -1201,11 +1199,9 @@ def checkMissingAxisValues(stat, axisTags):
         raise ValueError(f"Cannot find Axis Values for axes [{missing}]")
 
 
-def _sortAxisValues(stat):
+def _sortAxisValues(axisValues):
     # Sort and remove duplicates ensuring that format 4 Axis Values
     # are dominant
-    axisValues = stat.AxisValueArray.AxisValue
-    designAxes = stat.DesignAxisRecord.Axis
     results = []
     seenAxes = set()
     # Sort format 4 axes so the tables with the most AxisValueRecords
--- a/Lib/fontTools/varLib/instancer.py
+++ b/Lib/fontTools/varLib/instancer.py
@@ -1131,7 +1131,7 @@ def updateNameTable(varfont, axisLimits):
     # If we're instantiating a partial font, we will populate the unpinned
     # axes with their default axis values.
     fvarDefaults = {a.axisTag: a.defaultValue for a in fvar.axes}
-    axisCoords = axisLimits
+    axisCoords = deepcopy(axisLimits)
     for axisTag, val in fvarDefaults.items():
         if axisTag not in axisCoords or isinstance(axisCoords[axisTag], tuple):
             axisCoords[axisTag] = val
--- a/Lib/fontTools/varLib/instancer.py
+++ b/Lib/fontTools/varLib/instancer.py
@@ -1187,6 +1187,7 @@ def instantiateVariableFont(
     inplace=False,
     optimize=True,
     overlap=OverlapMode.KEEP_AND_SET_FLAGS,
+    update_nametable=False
 ):
     """Instantiate variable font, either fully or partially.
 
@@ -1272,6 +1273,10 @@ def instantiateVariableFont(
                 log.info("Removing overlaps from glyf table")
                 removeOverlaps(varfont)
 
+    if update_nametable:
+        log.info("Updating nametable")
+        updateNameTable(varfont, axisLimits)
+
     varLib.set_default_weight_width_slant(
         varfont,
         location={
@@ -1282,6 +1287,90 @@ def instantiateVariableFont(
     )
 
     return varfont
+
+
+def updateNameTable(varfont, axisLimits):
+    nametable = varfont["name"]
+    if "STAT" not in varfont:
+        raise ValueError("Cannot update name table since there is no STAT table.")
+    stat = varfont['STAT']
+    axisRecords = stat.table.DesignAxisRecord.Axis
+    axisValues = stat.table.AxisValueArray.AxisValue
+
+    axisOrder = {a.AxisOrdering: a.AxisTag for a in axisRecords}
+    axisValueNames = []
+    for axisValue in axisValues:
+        axisTag = axisOrder[axisValue.AxisIndex]
+        axisValueName = nametable.getName(
+            axisValue.ValueNameID,
+            3,
+            1,
+            0x409
+        ).toUnicode()
+        # Ignore axisValue if it has ELIDABLE_AXIS_VALUE_NAME flag enabled.
+        # Enabling this flag will hide the axisValue in application font menus.
+        if axisValue.Flags == 2:
+            continue
+
+        if axisValue.Format in (1, 3):
+            # Add axisValue if it's used to link to another variable font
+            if axisTag not in axisLimits and axisValue.Value == 1.0:
+                axisValueNames.append(axisValueName)
+
+            # Add axisValue if its value is in the axisLimits and the user has
+            # pinned the axis
+            elif isinstance(axisLimits[axisTag], float) and axisValue.Value == axisLimits[axisTag]:
+                axisValueNames.append(axisValueName)
+
+        if axisValue.Format == 2:
+            if isinstance(axisLimits[axisTag], float) and axisLimits[axisTag] >= axisValue.RangeMinValue \
+                and axisLimits[axisTag] <= axisValue.RangeMaxValue:
+                    axisValueNames.append(axisValueName)
+
+        # TODO Format 4
+        if axisValue.Format == 4:
+            raise NotImplementedError("Cannot update nametable using format 4 axisValues")
+        
+    _updateNameRecords(nametable, axisValueNames)
+
+
+def _updateNameRecords(nametable, axisValueNames):
+    # Update nametable based on the discovered axisValueNames
+    # using the R/I/B/BI model.
+    ribbiStyles = frozenset(["Regular", "Italic", "Bold", "Bold Italic"])
+    ribbiParticles = [n for n in axisValueNames if n in ribbiStyles]
+    nonRibbiParticles = [n for n in axisValueNames if n not in ribbiStyles]
+
+    currentFamilyName = nametable.getName(16, 3, 1, 0x409) or \
+            nametable.getName(1, 3, 1, 0x409)
+    currentFamilyName = currentFamilyName.toUnicode()
+
+    currentStyleName = nametable.getName(17, 3, 1, 0x409) or \
+            nametable.getName(2, 3, 1, 0x409)
+    currentStyleName = currentStyleName.toUnicode()
+
+    nameIDs = {1: currentFamilyName, 2: " ".join(ribbiParticles) or "Regular"}
+    # Include Typographic family and subfamily names if there are
+    # nonRibbiParticles
+    if nonRibbiParticles:
+        nameIDs[1] = f"{currentFamilyName} {' '.join(nonRibbiParticles)}"
+        nameIDs[16] = currentFamilyName
+        nameIDs[17] = " ".join(nonRibbiParticles + ribbiParticles)
+
+    newFamilyName = nameIDs.get(16) or nameIDs.get(1)
+    newStyleName = nameIDs.get(17) or nameIDs.get(2)
+    # Update full font name
+    nameIDs[4] = f"{newFamilyName} {newStyleName}"
+    # Update postscript name
+    nameIDs[6] = f"{newFamilyName.replace(' ', '')}-{newStyleName.replace(' ', '')}"
+    # Update uniqueID
+    # TODO
+    # versionRecord = nametable.getName(5, 3, 1, 0x409)
+
+    existingPlatforms = set((r.platformID, r.platEncID, r.langID) for r in nametable.names)
+    for nameID, string in nameIDs.items():
+        for plat in existingPlatforms:
+            nametable.setName(string, nameID, *plat)
 
 
 def splitAxisLocationAndRanges(axisLimits, rangeType=AxisRange):
@@ -1377,6 +1466,12 @@ def parseArgs(args):
         help="Merge overlapping contours and components (only applicable "
         "when generating a full instance). Requires skia-pathops",
     )
+    parser.add_argument(
+        "--update-nametable",
+        action="store_true",
+        help="Update the instantiated font's nametable using the STAT "
+        "table Axis Values"
+    )
     loggingGroup = parser.add_mutually_exclusive_group(required=False)
     loggingGroup.add_argument(
         "-v", "--verbose", action="store_true", help="Run more verbosely."
@@ -1428,6 +1523,7 @@ def main(args=None):
         inplace=True,
         optimize=options.optimize,
         overlap=options.overlap,
+        update_nametable=options.update_nametable,
     )
 
     outfile = (
--- a/Lib/fontTools/varLib/instancer.py
+++ b/Lib/fontTools/varLib/instancer.py
@@ -1315,31 +1315,35 @@ def updateNameTable(varfont, axisLimits):
         if axisValue.Format in (1, 3):
             # Add axisValue if it's used to link to another variable font
             if axisTag not in axisLimits and axisValue.Value == 1.0:
-                axisValueNames.append(axisValueName)
+                axisValueNames.append((axisValueName, axisTag))
 
             # Add axisValue if its value is in the axisLimits and the user has
             # pinned the axis
             elif isinstance(axisLimits[axisTag], float) and axisValue.Value == axisLimits[axisTag]:
-                axisValueNames.append(axisValueName)
+                axisValueNames.append((axisValueName, axisTag))
 
         if axisValue.Format == 2:
             if isinstance(axisLimits[axisTag], float) and axisLimits[axisTag] >= axisValue.RangeMinValue \
                 and axisLimits[axisTag] <= axisValue.RangeMaxValue:
-                    axisValueNames.append(axisValueName)
+                    axisValueNames.append((axisValueName, axisTag))
 
         # TODO Format 4
         if axisValue.Format == 4:
             raise NotImplementedError("Cannot update nametable using format 4 axisValues")
         
-    _updateNameRecords(nametable, axisValueNames)
+    _updateNameRecords(varfont, nametable, axisValueNames)
 
 
-def _updateNameRecords(nametable, axisValueNames):
+def _updateNameRecords(varfont, nametable, axisValueNames):
     # Update nametable based on the discovered axisValueNames
-    # using the R/I/B/BI model.
+    # using the R/I/B/BI and WWS models.
     ribbiStyles = frozenset(["Regular", "Italic", "Bold", "Bold Italic"])
-    ribbiParticles = [n for n in axisValueNames if n in ribbiStyles]
-    nonRibbiParticles = [n for n in axisValueNames if n not in ribbiStyles]
+    wwsAxes = frozenset(["wght", "wdth", "ital"])
+
+    ribbiParticles = [n for n, _ in axisValueNames if n in ribbiStyles]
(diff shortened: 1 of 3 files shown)


twardoch (commented):
There is the [Adobe spec for generating nameID 6](https://wwwimages2.adobe.com/content/dam/acom/en/devnet/font/pdfs/5902.AdobePSNameGeneration.pdf) referenced from https://docs.microsoft.com/en-us/typography/opentype/spec/otvaroverview — which ideally should be taken into account

twardoch:
@PeterConstable Your input would be highly appreciated — this is about `name` entries based on `STAT` particle names when building static instances

m4rc1e:
Thanks Adam. I'll add some test cases so we can check these first before we review the code.

Lib/fontTools/varLib/instancer.py
```diff
@@ -1284,6 +1298,158 @@ def instantiateVariableFont(
...
+    axisOrder = {a.AxisOrdering: a.AxisTag for a in axisRecords}
+    pinnedAxes = set(k for k, v in axisLimits.items() if isinstance(v, (float, int)))
+
+    results, seen_axes = [], set()
+    for axisValue in axisValues:
+
+        # Ignore axisValue if it has ELIDABLE_AXIS_VALUE_NAME flag enabled.
+        # Enabling this flag will hide the axisValue in application font menus.
+        # TODO this is too greedy! we need to retain wght axisValues
+        if axisValue.Flags == 2:
```
PeterConstable, line 1319: Can you define constants for the flags and then use that?
`if axisValue.Flags == ELIDABLE_AXIS_VALUE_NAME:`

Lib/fontTools/varLib/instancer.py
```diff
@@ -1284,6 +1298,158 @@ def instantiateVariableFont(
...
+            for rec in axisValue.AxisValueRecord:
+                axisTag = axisOrder[rec.AxisIndex]
+                if axisTag not in pinnedAxes:
+                    continue
+                if rec.Value == axisLimits[axisTag]:
+                    seen_axes.add(rec.AxisIndex)
+                    results.append((rec.AxisIndex, axisValue))
+
+        elif axisValue.Format in (1, 3):
+            axisTag = axisOrder[axisValue.AxisIndex]
```
PeterConstable: You've created `axisOrder` as a set, unordered but keyed by the `AxisRecord.axisOrdering` field. Yet here you're looking up a tag in `axisOrder` by using `AxisValueFormat1.axisIndex`? Unless I'm missing something, that doesn't seem right. I would think you'd index into `axisRecords` to get the tag.

`axisTag = axisRecords[axisValue.AxisIndex].AxisTag`

m4rc1e: Hey Peter. I'm still drafting the PR so I'm moving things around quite a bit

Atm, it may be best to just review the unit tests.

Lib/fontTools/varLib/instancer.py
m4rc1e: Will do thank you!

Lib/fontTools/varLib/instancer.py
```diff
@@ -1284,6 +1298,158 @@ def instantiateVariableFont(
...
+            # Add axisValue if it's used to link to another variable font
+            if axisTag not in axisLimits and axisValue.Value == 1.0:
```
PeterConstable: It seems to be assumed that the axisLimits parameter passed in must have an element for every axis in this font. If so, perhaps clarify that with comment/doc string back at the start of this function defn.

Also, it's not clear why `axisValue.Value == 1.0` is a condition. 

Let's assume the likely case for two distinct but related variable fonts of Roman and Italic counterparts. The italic font should have an axis value table for 'ital' with value 1.0; but then the Roman font should have an axis value table for 'ital' with value 0. Wouldn't you want both to be treated the same (module whether the ELIDABLE flag is set)?

But generalizing, there's nothing saying that only Roman and Italic could be paired in separate variable fonts in that way. As a hypothetical example, one could (for whatever reason) create variable fonts with wght and wdth axes, but have discrete optical size variants within the family. Or, create variable fonts with wdth and opsz axes but have discrete weight variants Regular, Bold, Black.

So, `1.0` is a very specific case being singled out. If there's a particular reason for that, it should be clarified in a comment.

Lib/fontTools/varLib/instancer.py
```diff
@@ -1284,6 +1298,158 @@ def instantiateVariableFont(
...
+
+        elif axisValue.Format == 2:
+            axisTag = axisOrder[axisValue.AxisIndex]
+            if axisTag not in pinnedAxes:
+                continue
+            if axisLimits[axisTag] >= axisValue.RangeMinValue \
+                and axisLimits[axisTag] <= axisValue.RangeMaxValue:
+                    seen_axes.add(axisValue.AxisIndex)
+                    results.append((axisValue.AxisIndex, axisValue))
+    return [v for k, v in sorted(results)]
```
PeterConstable: To be clear, this is sorting by `axisValue.AxisIndex`, which is an index into the `axisRecords` array; and not sorting by axis ordering values. Is that what's intended? (If so, probably worth calling out in a comment.)

anthrotype:
Thanks Marc for working on this!

> update the name table by using the STAT table axisValues.

I haven't reviewed this yet. I just have more like a design sort of question. 

Should we not use the `fvar` named instances style names first and foremost when the requested instance's location matches exactly one of the predefined `fvar` named instances?

I know that the OT spec says that STAT is supposed to be a _superset_ of fvar and the two tables should be in sync (i.e. the ordering of the STAT design axes and the AxisValue names should match the implicit ordering and content of the named instances' style name particles); in this sense, by only using STAT like you seem to do in this PR would still produce the same effect as trying to use the fvar first -- at least for fonts where STAT and fvar are truly in agreement with one another.

But I believe there are many fonts out there with half-empty STAT tables (varLib still does that). So how about we first see if there's already an fvar namedinstance available at requested location, and use that, otherwise check in the STAT table?

m4rc1e:
> So how about we first see if there's already an fvar namedinstance available at requested location, and use that, otherwise check in the STAT table?

That's the plan :-)

I'm just tackling the STAT now because the fonts we serve on Google Fonts have a tiny amount of fvar instances so drilling into the STAT is required.

m4rc1e:
Using an fvar instance to update the name table is a bit tricker than expected, especially if you want to use the R/I/B/BI model. If we have an instance named 'Semi Bold Italic', one would expect the following nameIDs:

```
NameID 1: FamilyName Semi Bold
NameID 2: Italic
NameID 16: FamilyName
NameID 17: Semi Bold
```

The above records cannot be created by simply using the space as a delimiter to tokenize the instance name. If we did this we'd get `['Semi', 'Bold', 'Italic']`, when the desired result is `["Semi Bold", "Italic"]`. One way to get the individual tokens for each axis would be to compare the fvar instances against each other. I need to think about this problem a bit more. In the meantime, I'll carry on working on the STAT approach.

anthrotype:
Fair enough.. "to compose strings that may be required for use in legacy applications" (e.g. RBBI model) is indeed one of the stated goals of the STAT table.

m4rc1e:
My plan is still to have both. I think I'll open the PR up for review once I'm happy with the STAT approach and then add the fvar later on.

Tests/varLib/instancer_test.py
```diff
@@ -1903,6 +1903,167 @@ def test_normalizeAxisLimits_missing_from_fvar(varfont):
...
+        (r.nameID, r.platformID, r.platEncID, r.langID): r.toUnicode()
+        for r in nametable.names
+    }
+
+
+def test_updateNameTable_with_registered_axes(varfont):
+    # Regular
+    instancer.updateNameTable(varfont, {"wght": 400})
+    names = _get_name_records(varfont)
+    assert names[(1, 3, 1, 0x409)] == "Test Variable Font"
```
m4rc1e, line 1918: for NameIDs use the enum

m4rc1e: Don't unset the other flags as well.

Tests/varLib/instancer_test.py
```diff
@@ -1903,6 +1903,167 @@ def test_normalizeAxisLimits_missing_from_fvar(varfont):
...
+
+def test_updateNameTable_missing_stat(varfont):
+    del varfont["STAT"]
+    with pytest.raises(ValueError, match="Cannot update name table since there is no STAT table."):
+        instancer.updateNameTable(varfont, {"wght": 400})
+
+
+def test_updateNameTable_vf_with_italic_attribute(varfont):
+    font_link_axisValue = varfont["STAT"].table.AxisValueArray.AxisValue[4]
+    font_link_axisValue.Flags = 0
```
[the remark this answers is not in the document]
m4rc1e: Don't unset the other flags as well.

Lib/fontTools/varLib/instancer.py
```diff
@@ -1099,6 +1115,223 @@ def pruningUnusedNames(varfont):
             del varfont["ltag"]
 
 
+def updateNameTable(varfont, axisLimits):
+    """Update an instatiated variable font's name table using the Axis
+    Value Tables from the STAT table.
+
+    To establish which Axis Value Tables are needed, we first remove all
```
m4rc1e, line 1122: remove = ignore

Lib/fontTools/varLib/instancer.py
```diff
@@ -1099,6 +1115,223 @@ def pruningUnusedNames(varfont):
...
+
+    # The updated name table must reflect the new 'zero origin' of the font.
+    # If a user is instantiating a partial font, we will populate the
+    # unpinned axes with their default axis values.
+    fvarDefaults = {a.axisTag: a.defaultValue for a in fvar.axes}
+    axisCoords = axisLimits
+    for axisTag, val in fvarDefaults.items():
+        if axisTag not in axisCoords:
+            axisCoords[axisTag] = val
+        elif isinstance(axisCoords[axisTag], tuple):
```
m4rc1e, line 1143: use or instead elif

Lib/fontTools/varLib/instancer.py
```diff
@@ -1099,6 +1115,223 @@ def pruningUnusedNames(varfont):
...
+            axisCoords[axisTag] = val
+
+    stat_new = deepcopy(stat).table
```
m4rc1e, line 1146: Write a comment for this.

PeterConstable:
> Using an fvar instance to update the name table is a bit tricker than expected, especially if you want to use the R/I/B/BI model. If we have an instance named 'Semi Bold Italic', one would expect the following nameIDs:
> 
> ```
> NameID 1: FamilyName Semi Bold
> NameID 2: Italic
> NameID 16: FamilyName
> NameID 17: Semi Bold
> ```
> 
> The above records cannot be created by simply using the space as a delimiter to tokenize the instance name. If we did this we'd get `['Semi', 'Bold', 'Italic']`, when the desired result is `["Semi Bold", "Italic"]`. One way to get the individual tokens for each axis would be to compare the fvar instances against each other. I need to think about this problem a bit more. In the meantime, I'll carry on working on the STAT approach.

Shouldn't Name ID 17 be "Semi Bold Italic"?

Lib/fontTools/varLib/instancer.py
```diff
@@ -1099,6 +1115,223 @@ def pruningUnusedNames(varfont):
...
+    format4 = sorted(
+        [v for v in axisValueTables if v.Format == 4],
+        key=lambda v: len(v.AxisValueRecord),
+        reverse=True,
+    )
+    nonFormat4 = [v for v in axisValueTables if v not in format4]
+
+    for val in format4:
+        axisIndexes = set(r.AxisIndex for r in val.AxisValueRecord)
+        if bool(seenAxes & axisIndexes) == False:
```
m4rc1e, line 1199: don't need bool...I think

Lib/fontTools/varLib/instancer.py
```diff
@@ -1099,6 +1115,223 @@ def pruningUnusedNames(varfont):
...
+            seenAxes |= axisIndexes
+            results.append((tuple(axisIndexes), val))
```
m4rc1e, line 1201: tuples need to be sorted

m4rc1e: Perhaps you only want to store the min axisIndex

Lib/fontTools/varLib/instancer.py
m4rc1e: We need to keep bool since `bool(set() & set()) == False` returns True, yet `set() & set() == False` returns False.

justvanrossum: `if not seenAxes & axisIndices:` would be idiomatic.

justvanrossum: But if you insist on the `bool()`, then: `if not bool(seenAxes & axisIndices):` would be best.

m4rc1e: Thanks!

m4rc1e:
@PeterConstable you're right. Apologies my example was incorrect.

@twardoch I'm pretty much done. I still haven't implemented your ps name suggestion but I'll add it to this PR. In the meantime, can you confirm that my test cases work as expected? Also, if you can think of any others, I'll happily add them.

Lib/fontTools/varLib/instancer.py
m4rc1e: Thanks for the in depth comments Peter. In the end, I just used an existing function within this module (instantiateSTAT) since my code was pretty much the same.

m4rc1e:
@anthrotype I think this is ready for a first review. I still haven't implemented Adobe postscript naming. I will probably do this when I make another PR to update the fsSelection/macStyle bits.

davelab6:
@m4rc1e what is the status and next steps for this? :)

m4rc1e:
@anthrotype Feels ready now.

I think the failing test is a network error.

anthrotype (commented):
I haven't finished reviewing this, but wanted to flush a bunch of review comments below lest they get lost.

In general, it'd be helpful if you could provide a brief general outline of the algorithm. Thank you!

Lib/fontTools/varLib/instancer.py
```diff
@@ -1099,6 +1117,293 @@ def pruningUnusedNames(varfont):
...
+    axisCoords = deepcopy(axisLimits)
+    for axisTag, val in fvarDefaults.items():
+        if axisTag not in axisCoords or isinstance(axisCoords[axisTag], tuple):
+            axisCoords[axisTag] = val
+
+    # To get the required Axis Values for the zero origin, we can simply
+    # duplicate the STAT table and instantiate it using the axis coords we
+    # created in the previous step.
+    stat_new = deepcopy(stat)
+    _instantiateSTAT(stat_new, axisCoords)
```
anthrotype, line 1146: It's not entirely clear to me why we have to call instantiateSTAT on a copy of the actual table here. Could we avoid it by running the updateNameTable after the normal instantiateSTAT has run?

m4rc1e: To construct the name table records, we need a single AxisValue each axis. If the user instantiates a partial font, there could potentially be AxisValues for a single axis. By reinstantiating a duplicate STAT table using the new axisCoords, we will get a single AxisValue for each axis.

I haven't looked at this code in a while and this took me a fair bit of time to work out. I may refactor this.

Lib/fontTools/varLib/instancer.py
```diff
@@ -1123,6 +1123,28 @@ def updateNameTable(varfont, axisLimits):
...
+    # This task can be split into two parts:
+
+    #   1. Collecting and sorting the relevant AxisValues:
+    # - First check the variable font has a STAT table and it contains
+    #   AxisValues.
+    # - Create a dictionary which contains the new default locations for each
+    #   axis.
+    # - Duplicate the existing stat table and instantiate it using the dict
+    #   we created in the last step. We should now have a new STAT table
+    #   which only contains AxisValues for the default axis locations.
```
anthrotype, line 1135: I think you should be able to get this list of AxisValues for the new default locations without having to make a duplicate STAT table and running `_instantiateSTAT` on it.

Lib/fontTools/varLib/instancer.py
```diff
@@ -1123,6 +1123,28 @@ def updateNameTable(varfont, axisLimits):
...
+    # - Remove any AxisValues which have the Elidable_AXIS_VALUE_NAME set.
+    # - Remove and sort AxisValues so format 4 AxisValues take presedence.
+
+    #   2. Updating a name table's style and family names from a list of
+    #   AxisValues:
+    # - Sort AxisValues into two groups, one for RIBBI, the other for non-RIBBI
```
anthrotype, line 1141: pls define the meaning of "RIBBI", we can't assume it's obvious

Lib/fontTools/varLib/instancer.py
anthrotype: you basically want to get the relevant axis values out from the STAT table, it's a matter of iterating over them and storing them in a list, no need for the copy deepcopy and re-running of instantiateSTAT. If you see there's logic duplicated between your loop and the instantiateSTAT itself, you can factor out the relevant portion and reuse accordingly

m4rc1e: > If you see there's logic duplicated between your loop and the instantiateSTAT itself, you can factor out the relevant portion and reuse accordingly

SGTM

m4rc1e: Yep, the implementation is now much cleaner. Thanks!

Lib/fontTools/varLib/instancer.py
```diff
@@ -1099,6 +1117,313 @@ def pruningUnusedNames(varfont):
...
+    if not currentRecord:
+        return None
+
+    def isSubString(string1, string2):
+        if string2 in string1:
+            return True
+        return False
+
+    # Check if full name and postscript name are a substring of currentRecord
+    for nameID in (4, 6):
```
anthrotype, line 1402: what's 4 or 6? Didn't we define an enum for these?

Lib/fontTools/varLib/instancer.py
```diff
@@ -1099,6 +1117,313 @@ def pruningUnusedNames(varfont):
...
+                nameRecord.toUnicode(), nameIDs[nameRecord.nameID]
+            )
+    # Create a new string since we couldn't find any substrings.
+    fontVersion = _fontVersion(varfont, platform)
+    vendor = varfont["OS/2"].achVendID.strip()
+    psName = nameIDs[NameID.POSTSCRIPT_NAME]
+    return f"{fontVersion};{vendor};{psName}"
+
+
+def _fontVersion(font, platform=(3, 1, 0x409)):
```
anthrotype, line 1417: this special Windows English (platformID,platEncID,langID) triplet returns a lot, can we name a constant and reuse perhaps?

Lib/fontTools/varLib/instancer.py
```diff
@@ -1099,6 +1117,313 @@ def pruningUnusedNames(varfont):
...
+        nameRecord = nametable.getName(nameID, *platform)
+        if not nameRecord:
+            continue
+        if isSubString(currentRecord.toUnicode(), nameRecord.toUnicode()):
+            return currentRecord.toUnicode().replace(
+                nameRecord.toUnicode(), nameIDs[nameRecord.nameID]
+            )
+    # Create a new string since we couldn't find any substrings.
+    fontVersion = _fontVersion(varfont, platform)
+    vendor = varfont["OS/2"].achVendID.strip()
```
anthrotype, line 1412: achVendID could potentially contain non-text garbage like control characters and stuff like that, maybe it's not safe to store as is inside the unique ID string. I think the OT spec allows only "printable ASCII characters", maybe you can filter the achVendID accordingly before inserting in the uniqueID string

Tests/varLib/instancer_test.py
```diff
@@ -1916,6 +1917,212 @@ def test_normalizeAxisLimits_missing_from_fvar(varfont):
...
+    instancer.updateNameTable(varfont, {"wght": 400})
+    names = _get_name_records(varfont)
+    assert names[(1, 3, 1, 0x409)] == "Test Variable Font"
+    assert names[(2, 3, 1, 0x0409)] == "Regular"
+    assert names[(3, 3, 1, 0x0409)] == "2.001;GOOG;TestVariableFont-Regular"
+    assert names[(6, 3, 1, 0x409)] == "TestVariableFont-Regular"
+    assert (16, 3, 1, 0x409) not in names
+    assert (17, 3, 1, 0x409) not in names
+
+    # Black
```
anthrotype, line 1939: splitting into separate unit test methods would be preferable. You could even parametrize the test input/expected output using `pytest.mark.parameterize` decorator, so you only write the test code once and only change the inputs and expected results. But that's optional.

Tests/varLib/instancer_test.py
```diff
@@ -1916,6 +1917,212 @@ def test_normalizeAxisLimits_missing_from_fvar(varfont):
...
+
+    buildStatTable(varfont, axes)
+    instancer.updateNameTable(varfont, {"wdth": 75, "wght": 400})
+    names = _get_name_records(varfont)
+    assert names[(17, 3, 1, 0x409)] == "Condensed Regular"
+
+
+def test_updateNameTable_with_multilingual_names(varfont):
+    name = varfont["name"]
+    name.setName("Test Variable Font", 1, 3, 1, 0x405)
```
anthrotype, line 2003: curious what language is 0x405, too lazy to check (a comment maybe)?

Tests/varLib/instancer_test.py
```diff
@@ -1916,6 +1917,212 @@ def test_normalizeAxisLimits_missing_from_fvar(varfont):
...
+    assert names[(16, 3, 1, 0x405)] == "Test Variable Font"
+    assert names[(17, 3, 1, 0x405)] == "Negreta Zhuštěné"
+
+
+def test_updateNametable_partial(varfont):
+    instancer.updateNameTable(varfont, {"wdth": 79, "wght": (400, 900)})
+    names = _get_name_records(varfont)
+    assert names[(1, 3, 1, 0x409)] == "Test Variable Font Condensed"
+    assert names[(2, 3, 1, 0x409)] == "Regular"
+    assert (3, 3, 1, 0x405) not in names
```
anthrotype, line 2042: what's the use of this assertion in this particular test? you're only checking english names and the `varfont` doesn't contain any 0x405-language nameIDs anyway.

Tests/varLib/instancer_test.py
```diff
@@ -1916,6 +1917,212 @@ def test_normalizeAxisLimits_missing_from_fvar(varfont):
...
+    assert names[(16, 3, 1, 0x409)] == "Test Variable Font"
+    assert names[(17, 3, 1, 0x409)] == "Condensed"
+
+
+def test_updateNameTable_missing_axisValues(varfont):
+    with pytest.raises(ValueError, match="Cannot find Axis Values for axes \[wght\]"):
```
anthrotype, line 2048: maybe the error message should specify for which input location it cannot find axis values for an axis. Just saying "Cannot find Axis Values for axes [...]" is too broad, seems to imply there are none for that axis. It might be there are some, only that the ones requested are not there

Lib/fontTools/varLib/instancer.py
m4rc1e: Good spot. According to MS Spec "All vendor IDs use the Tag data type, which is equivalent to a four-character string composed of a limited set of ASCII characters."

Tests/varLib/instancer_test.py
m4rc1e: Czech. I'll add a comment

Lib/fontTools/varLib/instancer.py
```diff
@@ -1099,6 +1117,313 @@ def pruningUnusedNames(varfont):
...
+    # - Sort AxisValues into two groups. For the first group, the names must
+    #   any of the following ["Regular", "Italic", "Bold", "Bold Italic"].
+    #   This group of names is often referred to as "RIBBI" names. For the
+    #   other group, names must be non-RIBBI e.g "Medium Italic", "Condensed"
+    #   etc.
+    # - Repeat the next steps for each name table record platform:
+    #   - Create new subFamily name and Typographic subFamily from the above
+    #     groups
+    #   - Update nameIDs 1, 2, 3, 4, 5, 6, 16, 17 using the new name created
+    #     in the last step.
```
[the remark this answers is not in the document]
m4rc1e: Do you think it's too verbose to have an example which runs through both tasks? I don't mind covering every step in an example.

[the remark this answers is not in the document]
anthrotype: Both sounds good, thanks

Lib/fontTools/varLib/instancer.py
m4rc1e: I'm now stripping out non-ASCII characters. If I check the registered [MS Vendors](https://docs.microsoft.com/en-us/typography/vendors/), I can see some contain punctuation.

Tests/varLib/instancer_test.py
m4rc1e: Thanks. I'll parameterize my tests.

Lib/fontTools/varLib/instancer.py
```diff
@@ -1099,6 +1117,379 @@ def pruningUnusedNames(varfont):
...
+
+def nameIDIsRibbi(nametable, nameID):
+    engNameRecords = any(
+        r
+        for r in nametable.names
+        if (r.platformID, r.platEncID, r.langID) == (3, 1, 0x409)
+    )
+    if not engNameRecords:
+        raise ValueError(
+            f"Canot determine if there are RIBBI Axis Value Tables "
```
anthrotype, line 1358:
```suggestion
            f"Cannot determine if there are RIBBI Axis Value Tables "
```

Lib/fontTools/varLib/instancer.py
```diff
@@ -1099,6 +1117,379 @@ def pruningUnusedNames(varfont):
...
+    currentFamilyName = nametable.getName(
+        NameID.TYPOGRAPHIC_FAMILY_NAME, *platform
+    ) or nametable.getName(NameID.FAMILY_NAME, *platform)
+
+    currentStyleName = nametable.getName(
+        NameID.TYPOGRAPHIC_SUBFAMILY_NAME, *platform
+    ) or nametable.getName(NameID.SUBFAMILY_NAME, *platform)
+
+    currentFamilyName = currentFamilyName.toUnicode()
+    currentStyleName = currentStyleName.toUnicode()
```
anthrotype, line 1393: if neither typo nor legacy family/style name ids are present, calling `toUnicode()` on a None value will raise TypeError. I think getName returns None if the requested name id is missing. You should check and handle that case. I know it's unlikely that the basic two faminy/style names are missing from a valid name table, but you never know what you're thrown at

Lib/fontTools/varLib/instancer.py
```diff
@@ -1219,6 +1611,11 @@ def instantiateVariableFont(
             contours and components, you can pass OverlapMode.REMOVE. Note that this
             requires the skia-pathops package (available to pip install).
             The overlap parameter only has effect when generating full static instances.
+        updateFontNames (bool): if True, update the instantiated font's nametable using
```
anthrotype, line 1614:
```suggestion
        updateFontNames (bool): if True, update the instantiated font's name table using
```

Lib/fontTools/varLib/instancer.py
```diff
@@ -1219,6 +1611,11 @@ def instantiateVariableFont(
...
+            the Axis Value Tables from the STAT table. The name table will be updated so
+            it conforms to the R/I/B/BI model. If the STAT table is missing or
+            an Axis Value table is missing for a given axis coordinate, an Error will be
```
anthrotype, line 1617:
```suggestion
            an Axis Value table is missing for a given axis coordinate, a ValueError will be
```

Lib/fontTools/varLib/instancer.py
```diff
@@ -1099,6 +1117,379 @@ def pruningUnusedNames(varfont):
...
+    for axisTag, val in fvarDefaults.items():
+        if axisTag not in defaultAxisCoords or isinstance(
+            defaultAxisCoords[axisTag], AxisRange
+        ):
+            defaultAxisCoords[axisTag] = val
+
+    axisValueTables = axisValuesFromAxisLimits(stat, defaultAxisCoords)
+    checkAxisValuesExist(stat, axisValueTables, defaultAxisCoords)
+
+    # Remove axis Values which have ELIDABLE_AXIS_VALUE_NAME flag set.
```
anthrotype, line 1236: saying "remove" implies we're modifying the axis values, whereas we're just filtering them, including some excluding others

```suggestion
    # Ignore axis Values which have ELIDABLE_AXIS_VALUE_NAME flag set.
```

anthrotype:
let's continue together on https://github.com/fonttools/fonttools/pull/2189

Issue #2079, opened 2020-10-07
Rules order. Browsers vs Adobe Apps


juandelperal:
We are having trouble with the consistency of how rules work across different software.
Maybe this is an Adobe bug, but I wanted to know how this could be addressed.

The desired behavior is 
`oslash` -> `oslash.bold`
`oslash.sc` -> `oslash.sc.bold`

Using rules with `first` works well in Illustrator but replaces `oslash.sc` by `oslash.bold` in Chrome
Using rules with `last` works well in Chrome but doesn't do anything in Illustrator.

<img width="1284" alt="screenshot" src="https://user-images.githubusercontent.com/2414972/95331974-0c140300-08ab-11eb-92d9-72c2db871960.png">

I posted the files here:
https://github.com/juandelperal/FonttoolsRulesTest

You can take a look at the design space here
https://github.com/huertatipografica/Alegreya/blob/78b3afbdde40327e4e57a0427242951198fb82d8/sources/Alegreya.designspace

justvanrossum:
It is because Adobe doesn't support 'rclt' in the apps you've tested with (support for this feature is neccesary to implement "last" behavior).

I'm not sure, but perhaps switching to the Adobe "World-Ready Composer" can make a difference.

juandelperal:
<img width="915" alt="Screen Shot 2020-10-07 at 14 56 23" src="https://user-images.githubusercontent.com/2414972/95333709-6910b880-08ad-11eb-8a7e-7d177f04efb1.png">

justvanrossum:
A workaround is to build with "first", and add a `oslash.bold -> oslash.sc.bold` substitution to your smcp feature.

juandelperal:
Thanks Just!
I can confirm that Illustrator doesn't make the substitutions with any of these options:

![image](https://user-images.githubusercontent.com/2414972/95334763-cf4a0b00-08ae-11eb-9f7f-e637834800b2.png)



OK, will give it a try to your workaround. Thanks

juandelperal:
InDesign Word-Ready paragraph composer doesn't support `rclt` either. 
As this font and others I am working have many users using in Adobe apps, is there a chance that we can use last and specify a different feature tag instead of `rclt` for this?

justvanrossum:
[...]

PR #2080, opened 2020-10-07
[varLib] Add designspace lib key for custom feavar feature tag

justvanrossum:
Add designspace lib key to specify a custom feature tag for feature variations, as discussed in #2079 

This looks for the custom tag when "processing = last". Alternatively it could override the value for "processing" altogether.

I used `com.github.fonttools.varLib.featureVarsFeatureTag` for the lib key, as it's really `varLib` that's defining it.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -44,7 +44,8 @@ from .errors import VarLibError, VarLibValidationError
 log = logging.getLogger("fontTools.varLib")
 
 # This is a lib key for the designspace document. The value should be
-# an OpenType feature tag, to be used as the FeatureVariations faature.
+# an OpenType feature tag, to be used as the FeatureVariations feature.
+# If present, the DesignSpace <rules processing="..."> flag is ignored.
 FEAVAR_FEATURETAG_LIB_KEY = "com.github.fonttools.varLib.featureVarsFeatureTag"
 
 #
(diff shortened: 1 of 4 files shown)


Lib/fontTools/varLib/__init__.py
```diff
@@ -917,7 +918,11 @@ def build(designspace, master_finder=lambda s:s, exclude=[], optimize=True):
 	if 'cvar' not in exclude and 'glyf' in vf:
 		_merge_TTHinting(vf, model, master_fonts)
 	if 'GSUB' not in exclude and ds.rules:
-		_add_GSUB_feature_variations(vf, ds.axes, ds.internal_axis_supports, ds.rules, ds.rulesProcessingLast)
+		if ds.rulesProcessingLast:
```
anthrotype, line 921: how about the `processing="last"` is implicitly set whenever the lib key is present? It can't be otherwise, can it?

anthrotype: I mean, if one uses the lib key to set the custom tag, then `processing` attribute is irrelevant, since what it really matters is whether it is "rvrn" (which will be processed first) or any other feature tag (which will be processed last or following normal lookup order)

justvanrossum: Yeah, thanks, that makes sense. I'll change it.

anthrotype (approved):

juandelperal:
Thanks Just and Cosimo :)

PR #2082, opened 2020-10-08
[varLib] Interpret empty non-default glyphs as 'missing'

justvanrossum:
Interpret empty non-default glyphs as 'missing', if the default glyph is not empty. Part of discussion over at googlefonts/ufo2ft/issues/408

If this idea is considered useful and not harmful, it probably needs a CFF2 counterpart.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -216,6 +216,7 @@ def _add_gvar(font, masterModel, master_ttfs, tolerance=0.5, optimize=True):
 	assert "gvar" not in font
 	gvar = font["gvar"] = newTable('gvar')
 	glyf = font['glyf']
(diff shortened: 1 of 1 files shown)


anthrotype:
👍  to this. I think it's a useful thing to have as it allow to change only the components' offsets without requiring to also fill the base glyphs' outlines in the sparse masters even if these don't change.

justvanrossum:
Components are indeed the main motivation. From that angle, we don't need to bother with CFF2.

justvanrossum:
Does this need more discussion and review?

The change is fairly innocent, but it can hide errors, for example when a glyph is empty by mistake. cu2qu would catch that, but it's not always part of the pipeline. (That's also the reason why my ufo2ft filter is `pre=False`, so it runs *after* cu2qu...)

anthrotype (approved):

PR #2092, opened 2020-10-20
update colorLib to latest COLRv1 draft

anthrotype:
This syncs otData and colorLib implementation with the latest COLRv1 draft spec.

* each layers in LayerV1List is now DAGs of Paints
* added four new Paint formats: PaintGlyph, PaintColorGlyph, PaintTransform and PaintComposite
* PaintRadialGradient no longer has Affine2x2 field, but can optionally be wrapped by an Affine2x3 (like other Paints)


Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -525,17 +526,21 @@ def write(self, writer, font, tableDict, value, repeatIndex=None):
 
 class Table(Struct):
 
-	longOffset = False
 	staticSize = 2
 
 	def readOffset(self, reader):
 		return reader.readUShort()
```
rsheeter, line 532: Suggest either making readOffset support all sizes - and not implementing it on subclasses - or making the subclasses implement writeNullOffset.

I prefer the subclass implements and the parent should assert staticSize == 2 in both readOffset and writeNullOffset

Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -1688,7 +1753,8 @@ def _buildClasses():
 		if m:
 			# XxxFormatN subtable, we only add the "base" table
 			name = m.group(1)
-			baseClass = FormatSwitchingBaseTable
+			formatType = table[0][0]  # can be 'uint16' or 'uint8'
```
rsheeter, line 1756: Can we use named access or add a comment to explain what table[0] and table[0][0] are meant to match?

rsheeter:
Generally lgtm, couple of nits plus shoudl rename PaintColrGlyph => PaintColrSlice

rsheeter (approved):

Lib/fontTools/colorLib/builder.py
```diff
@@ -487,115 +610,50 @@ def buildPaintRadialGradient(
     return self
 
 
-def buildPaintGlyph(glyph: str, paint: _PaintInput) -> ot.Paint:
+def buildPaintGlyph(layerCollector: LayerCollector, glyph: str, paint: _PaintInput) -> ot.Paint:
```
anthrotype, line 613: I wonder if we can make this layerCollector an optional keyword argument that defaults to None, and when None these buildPaintSomething would just make a new LayerCollector and pass it down. E.g. if one only want to build a specific paint table that doesn't need to refer to the global LayerV1List.

rsheeter: Would it be too disruptive to propose we have a class akin to ColrBuilder and hang the various buildSomething method on it? - akin to FontBuilder. It's a model that makes a lot of sense for a series of related and stateful calls, which is now true here.

Lib/fontTools/colorLib/builder.py
```diff
@@ -430,6 +432,127 @@ def _to_color_line(obj):
...
+    2: lambda _, kwargs: buildPaintLinearGradient(**kwargs),
+    3: lambda _, kwargs: buildPaintRadialGradient(**kwargs),
+    4: lambda builder, kwargs: buildPaintGlyph(builder, **kwargs),
+    5: lambda _, kwargs: buildPaintColrGlyph(**kwargs),
+    6: lambda builder, kwargs: buildPaintTransform(builder, **kwargs),
+    7: lambda builder, kwargs: buildPaintComposite(builder, **kwargs),
+}
+
+
+def _tuple(obj) -> Tuple[Any, ...]:
```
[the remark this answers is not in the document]
rsheeter: renamed _as_tuple

Lib/fontTools/colorLib/builder.py
```diff
@@ -430,6 +432,127 @@ def _to_color_line(obj):
...
+        elif hasattr(value, "__dict__"):
+            return _tuple(value)
+        elif isinstance(value, collections.abc.MutableSequence):
+            return tuple(_tuple_safe(e) for e in value)
+        return value
+    return tuple((k, _tuple_safe(v)) for k, v in obj.__dict__.items())
+
+
+class LayerCollector:
+    Slices: List[ot.Paint]
```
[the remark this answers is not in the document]
rsheeter: It was indeed to be consistent with otTables lol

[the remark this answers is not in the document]
rsheeter: Fixed

Lib/fontTools/colorLib/builder.py
anthrotype: > have a class akin to ColrBuilder

yeah, I like that.

Lib/fontTools/colorLib/builder.py
```diff
@@ -430,6 +432,127 @@ def _to_color_line(obj):
...
+            return self.buildColrLayers(paint)
+        elif isinstance(paint, collections.abc.Mapping):
+            kwargs = dict(paint)
+            fmt = kwargs.pop("format")
+            try:
+                return _PAINT_BUILDERS[fmt](self, kwargs)
+            except KeyError:
+                raise NotImplementedError(fmt)
+        raise TypeError(
+            f"expected int, Mapping or ot.Paint, found {type(paint).__name__}: {paint!r}"
```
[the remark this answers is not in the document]
rsheeter: Removed expected; seems inevitable it will drift

Lib/fontTools/colorLib/builder.py
rsheeter: Updated

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -2,1666 +2,5463 @@
 from fontTools.misc.py23 import *
 
 otData = [
-
```
anthrotype, line 5: I see you have run `black` on otData.py.. I would prefer to keep the changes to a minimum and revert this. The rule (enforced on a best effort basis) is to only run `black` on newly added modules (e.g. colorLib.builder), but keep whatever is the current style for older existing modules.

rsheeter: Oops, will undo

rsheeter (approved):

PR #2094, opened 2020-10-29
svgLib: handle arc commands with bool flags not separated by space/comma

anthrotype:
Some SVG authoring tool write arc commands without any space or comma around the boolean 'large-arc' and 'sweep' flags, leading our svg path parser to choke.
E.g. `A1.2 1.2 0 0122 6.2` is currently tokenized as `["A", "1.2", "1.2", "0", "0122", "6.2"]`, whereas it should instead be tokenized as `["A", "1.2", "1.2", "0", "0", "1", "22", "6.2"]`

This PR makes the path parser smarter by special-casing arc command so that we only consume one character ('0' or '1') for these special boolean flags.

### DIFF
--- a/Lib/fontTools/svgLib/path/parser.py
+++ b/Lib/fontTools/svgLib/path/parser.py
@@ -13,18 +13,81 @@ import re
 
 
 COMMANDS = set('MmZzLlHhVvCcSsQqTtAa')
+ARC_COMMANDS = set("Aa")
 UPPERCASE = set('MZLHVCSQTA')
 
 COMMAND_RE = re.compile("([MmZzLlHhVvCcSsQqTtAa])")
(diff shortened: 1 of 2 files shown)


Lib/fontTools/svgLib/path/parser.py
```diff
@@ -13,18 +13,81 @@
...
+    arc_cmd = None
     for x in COMMAND_RE.split(pathdef):
         if x in COMMANDS:
+            arc_cmd = x if x in ARC_COMMANDS else None
             yield x
-        for token in FLOAT_RE.findall(x):
-            yield token
+            continue
+
+        if arc_cmd:
```
rsheeter, line 40: optional, if you are going to branch for arcs you could just write a regex specifically for arc (e.g. flag should be 0/1, has *optional* whitespace, etc)

rsheeter (approved):

PR #2100, opened 2020-11-11
Switch to GitHub Actions

madig:
Let's see if this works! Travis is going to wind down their open source program.


madig:
Ide thought: before I some day fix the Windows issue about relative paths, should pypy3 and cython be tested on more platforms? I already find it wasteful to test 3.6-3.9 on all platforms...

anthrotype:
cython and pypy just on linux.
macos and windows only on latest (stable) python: i.e. py3.8, whatever homebrew is the default.
for linux only, all the python3s

moyogo:
Isn’t py3.9 the default on homebrew now?

madig:
Not sure Homebrew is used on macOS but 3.9 is available. Once we sort out the Windows relpath issues we'll see if everything works :)

benkiel:
Thanks for this, watching closely to learn and then change all the RoboTools repositories...

madig:
@anthrotype so, I think this can go in after you set up the secret env vars.

anthrotype:
> this can go in after you set up the secret env vars.

added PYPI_USERNAME and PYPI_PASSWPRD as repo secrets in the settings. I hope that works.

madig:
Only one way to find out.

PR #2101, opened 2020-11-16
[otlLib] Build format 1 and format 2 contextual lookups

simoncozens:
This builds on existing work in the otlLib builder module. Currently we build all contextual/chained contextual lookups as format 3 subtables. However, there are often more efficient representations. Format 1 (simple glyph contexts) and format 2 (class based contexts) may save disk space.

In this PR, we build all the possible representations and compare their byte size, adding the most compact representation to the font. I've tried to break it down into manageable pieces.

Fun fact: I'm working on a font with lots of contextual rules; this patch took the final font size from 236Kb to 144Kb.

### DIFF
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -454,7 +454,6 @@ class ChainContextualBuilder(LookupBuilder):
         for _ in inClasses:
             classSet = self.newRuleSet_(format=2, chaining=chaining)
             classSet.populateDefaults()
-            classSets.append(classSet)
 
         coverage = set()
         classRuleAttr = self.ruleAttr_(format=2, chaining=chaining)
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -455,7 +455,6 @@ class ChainContextualBuilder(LookupBuilder):
         classSets = []
         for _ in inClasses:
             classSet = self.newRuleSet_(format=2, chaining=chaining)
-            classSet.populateDefaults()
             classSets.append(classSet)
 
         coverage = set()
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -567,20 +567,24 @@ class ChainContextualBuilder(LookupBuilder):
     # The following functions generate the attribute names and subtables according
     # to this naming convention.
     def ruleSetAttr_(self, format=1, chaining=True):
-        if format == 2:
-            formatType = "Class"
-        elif format == 1:
+        if format == 1:
             formatType = "Rule"
+        elif format == 2:
+            formatType = "Class"
+        else:
+            raise AssertionError(formatType)
         subtablename = f"{self.subtable_type[0:3]}{formatType}Set"  # Sub, not Subst.
         if chaining:
             subtablename = "Chain" + subtablename
         return subtablename
 
     def ruleAttr_(self, format=1, chaining=True):
-        if format == 2:
-            formatType = "Class"
-        elif format == 1:
+        if format == 1:
             formatType = ""
+        elif format == 2:
+            formatType = "Class"
+        else:
+            raise AssertionError(formatType)
         subtablename = f"{self.subtable_type[0:3]}{formatType}Rule"  # Sub, not Subst.
         if chaining:
             subtablename = "Chain" + subtablename
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -343,6 +343,8 @@ class ChainContextualBuilder(LookupBuilder):
             w["LookupType"] = CountReference(
                 {"LookupType": st.LookupType}, "LookupType"
             )
+            # We need to make a copy here because compiling
+            # modifies the subtable (finalizing formats etc.)
             copy.deepcopy(st).compile(w, self.font)
             size += len(w.getAllData())
         return size
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -400,6 +400,50 @@ class ChainContextualBuilder(LookupBuilder):
         setattr(st, f"{self.subtable_type}LookupRecord", [])
         return st
 
+    # Format 1 and format 2 GSUB5/GSUB6/GPOS7/GPOS8 rulesets and rules form a family:
+    #
+    #       format 1 ruleset      format 1 rule      format 2 ruleset      format 2 rule
+    # GSUB5 SubRuleSet            SubRule            SubClassSet           SubClassRule
+    # GSUB6 ChainSubRuleSet       ChainSubRule       ChainSubClassSet      ChainSubClassRule
+    # GPOS7 PosRuleSet            PosRule            PosClassSet           PosClassRule
+    # GPOS8 ChainPosRuleSet       ChainPosRule       ChainPosClassSet      ChainPosClassRule
+    #
+    # The following functions generate the attribute names and subtables according
+    # to this naming convention.
+    def ruleSetAttr_(self, format=1, chaining=True):
+        if format == 2:
+            formatType = "Class"
+        elif format == 1:
+            formatType = "Rule"
+        subtablename = f"{self.subtable_type[0:3]}{formatType}Set"  # Sub, not Subst.
+        if chaining:
+            subtablename = "Chain" + subtablename
+        return subtablename
+
+    def ruleAttr_(self, format=1, chaining=True):
+        if format == 2:
+            formatType = "Class"
+        elif format == 1:
+            formatType = ""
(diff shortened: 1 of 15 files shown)


alerque:
Is the most compact representation always the most efficient to parse and apply? Obviously the latter question is likely somewhat implementation specific...

simoncozens:
Yes, I think so, because formats 1 and 2 can merge multiple lookups into a single subtable, reducing processing time. To put numbers on that:

Without patch: file size 452k, 3800 substitution subtables, 100,000 shaping iterations in 5.05s
With patch: file size 248k, 178 substitution subtables, 100,000 shaping iterations in 3.67s

Lib/fontTools/otlLib/builder.py
```diff
@@ -348,6 +348,76 @@ def build(self):
...
+            ) = [c.build() for c in classdefs]
+        else:
+            st.ClassDef = classdefs[1].build()
+
+        inClasses = classdefs[1].classes()
+
+        classSets = []
+        for _ in inClasses:
+            classSet = self.newRuleSet_(format=2, chaining=chaining)
+            classSet.populateDefaults()
```
anthrotype, line 456: you've already called populateDefaults inside the `newRuleSet_` method so this second call is unneeded

Lib/fontTools/otlLib/builder.py
```diff
@@ -348,6 +348,76 @@ def build(self):
...
+        coverage = set()
+        classRuleAttr = self.ruleAttr_(format=2, chaining=chaining)
+
+        for rule in ruleset.rules:
+            rule_table = self.newRule_(format=2, chaining=chaining)
+            if chaining:
+                rule_table.BacktrackGlyphCount = len(rule.prefix)
+                rule_table.LookAheadGlyphCount = len(rule.suffix)
+                rule_table.Backtrack = [
+                    st.BacktrackClassDef.classDefs[list(x)[0]]
```
anthrotype, line 468: it's unclear to me why you need to cast x to a `list(x)[0]`; what type of values does `rule.prefix` contain? Mind that if x is a set, the order is random so getting the first item may give different results each time you initialize python.

Lib/fontTools/otlLib/builder.py
```diff
@@ -348,6 +348,76 @@ def build(self):
...
+        for _ in inClasses:
+            classSet = self.newRuleSet_(format=2, chaining=chaining)
+            classSet.populateDefaults()
+            classSets.append(classSet)
+
+        coverage = set()
+        classRuleAttr = self.ruleAttr_(format=2, chaining=chaining)
+
+        for rule in ruleset.rules:
+            rule_table = self.newRule_(format=2, chaining=chaining)
```
anthrotype, line 463: i'm a bit distracted by the mixing of camelCase and snake_case for variable names within the same block of code, but whatever.

Lib/fontTools/otlLib/builder.py
```diff
@@ -348,6 +348,52 @@ def build(self):
...
+                rule_table.InputGlyphCount = len(rule.glyphs)
+            else:
+                rule_table.GlyphCount = len(rule.glyphs)
+
+            rule_table.Input = [list(x)[0] for x in rule.glyphs[1:]]
+
+            self.buildLookupList(rule, rule_table)
+
+            firstGlyph = list(rule.glyphs[0])[0]
+            if not firstGlyph in rulesets_by_first_glyph:
```
anthrotype, line 417: "if item **not** in collection" is the correct way, not "if **not** item in collection"

Lib/fontTools/otlLib/builder.py
```diff
@@ -330,6 +336,17 @@ def rulesets(self):
...
         return [x for x in ruleset if len(x.rules) > 0]
 
+    def getCompiledSize_(self, subtables):
+        size = 0
+        for st in subtables:
+            w = OTTableWriter()
+            w["LookupType"] = CountReference(
+                {"LookupType": st.LookupType}, "LookupType"
+            )
+            copy.deepcopy(st).compile(w, self.font)
```
anthrotype, line 346: i suppose the copy is necessary because the OTTableWriter modifies the subtable somehow? A comment would help clarify why you need this.

anthrotype (commented):
Thanks Simon for working on this! I left some comments.
Overall it's looking good. Do we have tests that explicitly cover each format and chaining/non-chaining variant of contextual sub/pos lookups?

Lib/fontTools/otlLib/builder.py
simoncozens: I'll add a comment

simoncozens:
I've addressed all the review comments in the commits above, but the varLib test failures are concerning and I will need to think about them.

simoncozens:
Oh dear, that should teach me to code before having coffee. The old varLib failures were expected because I hadn't updated the tests expectations. Then I was getting *different* failures because I yanked the wrong line of code out. Now I'm back to where I was, and have fixed the varLib expectations. Should be fine now.

Lib/fontTools/otlLib/builder.py
```diff
@@ -393,38 +395,38 @@ def buildFormat1Subtable(self, ruleset, chaining=True):
         st.Format = 1
         st.populateDefaults()
         coverage = set()
-        rulesets_by_first_glyph = {}
+        rulesetsByFirstGlyph = {}
         ruleAttr = self.ruleAttr_(format=1, chaining=chaining)
 
         for rule in ruleset.rules:
-            rule_table = self.newRule_(format=1, chaining=chaining)
+            ruleAsSubtable = self.newRule_(format=1, chaining=chaining)
```
anthrotype, line 402: how about just ruleTable? ruleAsSubtable is a bit verbose

Lib/fontTools/otlLib/builder.py
```diff
@@ -453,45 +455,47 @@ def buildFormat2Subtable(self, ruleset, classdefs, chaining=True):
...
         for rule in ruleset.rules:
-            rule_table = self.newRule_(format=2, chaining=chaining)
+            ruleAsSubtable = self.newRule_(format=2, chaining=chaining)
             if chaining:
-                rule_table.BacktrackGlyphCount = len(rule.prefix)
-                rule_table.LookAheadGlyphCount = len(rule.suffix)
-                rule_table.Backtrack = [
+                ruleAsSubtable.BacktrackGlyphCount = len(rule.prefix)
+                ruleAsSubtable.LookAheadGlyphCount = len(rule.suffix)
+                # The glyphs in the rule may be list, tuple, odict_keys...
```
anthrotype, line 468: all the types you mentioned keep their order, so the comment that follows about order not being important is misplaced. Only if these were unordered set or frozenset it would be relevant.
I find it strange that we are so inconsistent we collection types. Can't we make sure we store these rules using the same type, instead of casting them ex-post?

anthrotype (commented):
thanks. Do we test all combinations of pos/sub, chain/non-chain, formats available?
would be nice to have individual tests for each one, instead of repurposing existing tests which were written under different circumstances.

Lib/fontTools/otlLib/builder.py
simoncozens: It's not a table of rules.

Lib/fontTools/otlLib/builder.py
anthrotype: IIUC, the reason you do `list(x)[0]` is to be able to get the first item of a collection which can sometimes be list, tuple or odict_keys.
I wonder: is there a good reason why the type used for the rule.prefix/glyphs/suffix can be either of these? I don't think there is, it's just happens to be that way. I suggest that we enforce one of these types and stick with it in a consistent way.
Do we want an immutable sequence that keeps order and that we can index into, then we should use tuple only.
I don't like this overly relaxed approach to typing, that induces one to do extra copies because anything goes.

anthrotype: A more efficient way to get the first item of a collection would be to do `next(iter(x))`, so you don't have to construct a whole list just to get the first item, and would work for odict_keys or (randomly ordered) sets too. In anycase, I reiterate my preference for sticking to one type only for the same kind of data, in this case a rule's glyph names.

anthrotype:
thanks for adding explicit tests for all formats.

Lib/fontTools/otlLib/builder.py
simoncozens: I agree that we should tidy up what goes in the rules. Can I suggest, though, that we make an issue for that and address it in a separate PR? It's going to require changes to the parts of feaLib.builder and feaLib.ast which are creating these ChainContextualRules, and that's not directly related to what I'm doing here.

anthrotype: sgtm

simoncozens: Thanks. Are you happy with this PR on that basis, then?

anthrotype (approved):
LGTM

Issue #1581, opened 2019-04-19
Support higher-order interpolation


rsheeter:
Allow multiple axes with the same tag. All but one must be hidden. This permits higher-order interpolation without requiring spec changes. This is the FontTools edition of harfbuzz/harfbuzz#1673.

twardoch:
Well, it doesn't "permit" it, currently the behavior of engines is undefined (some accidentally perform HOI). There is nothing nowhere about "all axes but one must be hidden", that's anecdotal and by no means supported by any spec. 

But there is some early consensus that multiple same-tagged axes are permissible.

bungeman:
I ran into this issue while writing https://bungeman.github.io/hoi.html . In the font itself axes are referred to by index, and in fvar axes have-a tag.

The idea is that at the API level if a user sets a variation value for a given axis tag then all axes with that tag should be set with that value (though if the axes have different min and max values that value may be clamped to different values). This is more or less what all modern implementations do.

Hiding axes isn't really necessary here, though APIs and UIs will need to be aware of the difference between an axis tag and an axis index. Generally one would probably want to mark all but one axis with a given tag as hidden for sanity and backward compatibility. If the API or UI is based on axis tag and shows two sliders both associated with the same axis tag which should it use to set the value? However, there shouldn't be any limitation in FontTools around this since it should be general purpose where it can be.

I propose a <AxisID> (like <GlyphID>) which could be used in an <fvar> <Axis> tag. If present the axis id would be the content of <AxisID>, if not the the axis id would be the content of <AxisTag>. Then the axis id would be used to specify an axis in all the places the axis attribute is used. When converting a font file to ttx the first axis with a given tag would just be given an id which matches the tag, subsequent axes with the same tag would be given an id like 'axis#1'. This is nicely backward compatible and shouldn't be too difficult to implement.

justvanrossum:
I was thinking along the same lines, except that I think we don't need a formal definition for `<AxisID>`.

- upon writing `fvar`, if `len(axisTag) > 4`, strip the rest. (Perhaps requiring that `axisTag[4] == "#", but nothing more.)
- upon reading `fvar`, append `#N` to the tags that exist multiple times, where N increments for each additional duplicate.

justvanrossum:
> https://bungeman.github.io/hoi.html

BTW., excellent write-up! Thanks for sharing.

bungeman:
[...]

PR #2104, opened 2020-11-18
Allow duplicate fvar axis tags

justvanrossum:
This implements #1581. Needs tests, though.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_f_v_a_r.py
+++ b/Lib/fontTools/ttLib/tables/_f_v_a_r.py
@@ -80,9 +80,15 @@ class table__f_v_a_r(DefaultTable.DefaultTable):
             raise TTLibError("unsupported 'fvar' version %04x" % header["version"])
         pos = header["offsetToData"]
         axisSize = header["axisSize"]
+        axisTagsSeen = {}
         for _ in range(header["axisCount"]):
             axis = Axis()
             axis.decompile(data[pos:pos+axisSize])
+            if axis.axisTag in axisTagsSeen:
+                axisTagsSeen[axis.axisTag] += 1
+                axis.axisTag = f"{axis.axisTag}#{axisTagsSeen[axis.axisTag]}"
+            else:
+                axisTagsSeen[axis.axisTag] = 0
             self.axes.append(axis)
             pos += axisSize
         instanceSize = header["instanceSize"]
@@ -119,7 +125,12 @@ class Axis(object):
         self.maxValue = 1.0
 
     def compile(self):
-        return sstruct.pack(FVAR_AXIS_FORMAT, self)
+        if len(self.axisTag) > 4:
+            obj = dict(self.__dict__)
+            obj["axisTag"] = self.axisTag[:4]
+        else:
+            obj = self
+        return sstruct.pack(FVAR_AXIS_FORMAT, obj)
 
     def decompile(self, data):
         sstruct.unpack2(FVAR_AXIS_FORMAT, data, self)


anthrotype:
I think the ttLib/tables/TupleVariation module also needs changing, as the tuple.axes is still a dictionary with unique keys. Then also the instancer module that uses that, and probably more..

justvanrossum:
> I think the ttLib/tables/TupleVariation module also needs changing, as the tuple.axes is still a dictionary with unique keys. Then also the instancer module that uses that, and probably more..

Everyone presumably gets their axisTags from fvar, so I was thinking that it should be enough to disambiguate it at the source. It works for https://bungeman.github.io/hoi.html, as the gvar tuple coordinates indeed pick up the disambiguated tags.

For `.designspace` files we may choose to disambiguate when compiling, or we should just ask users to disambiguate with a `#N` tag extension.

I haven't really looked at the possible consequences for instancer.

anthrotype:
Ah now I see, thanks. The change looked suspiciosly small.
I wonder if it would make more sense instead to represent fvar axes as a mapping keyed by axisTag where values are groups of axis records that share the same axisTag. That would definitely require more changes in the codebase where the expectation is that fvar axes are a flat dictionary of {axisTag:AxisRecord}, but it may be easier to work with than parsing strings.

Alternatively, another idea that springs to mind is adding `@property` getters to fvar AxisRecord class to obtain an `axis.baseTag` (4-letter common tag) and `axis.id` (digit after the `#`).

justvanrossum:
Not sure. In my implementation the bit after `#` is completely irrelevant, as long as it is uniqueifying the tag. There's really nothing to parse: just strip after the 4th char.

(I think it would be better if my implementation insisted on `#` as `tag[4]`, to avoid accidental wrong lengths.

anthrotype:
> better if my implementation insisted on # as tag[4]

yes, it should error out if that's not the case.

So basically with this patch, whenever one wants to get the true fvar axis 4-letter tag, one should always extract it as the `axis.axisTag[:4]`, because the one in axisTag may "lie" and also contain the disambiguation suffix for duplicate axes.

anthrotype:
and the presence of `#` as fifth character in an axis.axisTag string would signal that this axis record is a non-primary (presumably hidden) one.

justvanrossum:
> and the presence of # as fifth character in an axis.axisTag string would signal that this axis record is a non-primary (presumably hidden) one.

With my patch that is not necessarily the case, and I'm not sure it needs to be. The hidden flag should be leading.

This patch is purely about disambiguating duplicates, and I think that's an ok scope. (But as always, I'm happy to listen and let you change my mind.)

anthrotype:
> The hidden flag should be leading

of course, the hidden flag is already explicitly encoded. I just meant that any AxisRecord whose len(axisTag) > 4 would automatically be considered as one of the duplicated ones, as axisTag as encoded in fvar is always 4-char long. The compile method truncates whatever follows axisTag[:4] so as to get to the _true_ axisTag.
Other code elsewhere would have to do the same, and follow the new convention.
The only circumstance where len(axisTag) may be > 4 and the fifth character _not_ be `#` is if the axisTag string was set by some other user code.

justvanrossum:
What I currently do is that the first occurance of a tag (in fvar.axes order) keeps the plain tag, and only duplicates get a `#N` extension. I would perhaps be better if *all* occurances of a tag (that occurs more than once) get an extension. I don't care much either way – what I did was the easiest to implement.

bungeman:
There is no reason for '#' to imply a hidden axis. One would hope that, as much as possible, the xml format should try to be a higher level view of the font data. Even if the specification were updated to say that there should (or even must) only be one non-hidden axis with a given tag the xml format should still allow the representation and creation of fonts which have multiple non-hidden axes with the same tag since there is no good reason not to. Please keep the axis tag, axis flags, and axis id separate concepts and don't mix them together (especially in a low level tool like this). If there had been no initial confusion over conflating just the axis tag and axis id then this whole issue just wouldn't exist.

I would much prefer `<AxisTag>` stay how it has been (just be the axis tag) and instead `<Axis>` get a `name` attribute (like glyphs) which is an identifier (in the way the index of an axis in the fvar table is its identifier in the font file). This is, of course, a bit more work on the code side, but a far more orthogonal change which doesn't conflate multiple distinct ideas. It would be nice if the xml format were a little more consistent to begin with: xPOS tables do references by `index`, the names in the name table are referred to by their `nameID`, and even glyphs are referred to by their `name` even though they also have an `id` which is merely informative.

justvanrossum:
It's not just about the XML format, it's also about how axis tags are used internally. There are a lot of places that assume axis tags are unique, such as in location dicts, where axis tags are used as keys. Rewriting all that while keeping the public API compatible is more work than I think is worth it, or is at least more work than I am willing to put in.

If anyone has a better idea and has the capacity to implement it soon-ish, I'd be all ears.

behdad:
Should this be merged? I think it's harmless enough to merge.

behdad:
> Should this be merged? I think it's harmless enough to merge.

Well, it will trip the instancer and similar modules. But probably not more than they currently do.

justvanrossum:
This doesn't necessarily supersede #2113, although I like this better as it's less intrusive. But the arguments for #2113 aren't wrong, either.

Lorp:
> @justvanrossum: Everyone presumably gets their axisTags from fvar, so I was thinking that it should be enough to disambiguate it at the source. …

It’s not mentioned above, but in case it’s useful, here is a reminder that STAT indexes axes by tag, not id.

PR #2107, opened 2020-11-19
Added QuartzPen

verbosus:
FontTools supports drawing on macOS via the [`CocoaPen`](https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/pens/cocoaPen.py), which transforms a path into an `NSBezierPath`. However, for lower-level drawing on macOS, animation with CoreAnimation, and other use-cases, it’s useful to have outlines represented as `CGPath` instead of `NSBezierPath`. RoboFab used to have a [`QuartzPen`](https://github.com/robotools/robofab/blob/master/Lib/robofab/pens/quartzPen.py) but it’s no longer functional under Python 3. This PR adds a `QuartzPen` back to FontTools, together with some unit tests that only run when PyObjC is installed.

### DIFF
--- a/Lib/fontTools/pens/quartzPen.py
+++ b/Lib/fontTools/pens/quartzPen.py
@@ -1,0 +1,46 @@
+from fontTools.misc.py23 import *
+from fontTools.pens.basePen import BasePen
+
+from Quartz.CoreGraphics import CGPathCreateMutable, CGPathMoveToPoint
+from Quartz.CoreGraphics import CGPathAddLineToPoint, CGPathAddCurveToPoint
+from Quartz.CoreGraphics import CGPathAddQuadCurveToPoint, CGPathCloseSubpath
+	
+
+__all__ = ["QuartzPen"]
+
+
+class QuartzPen(BasePen):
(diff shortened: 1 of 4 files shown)


anthrotype (approved):
Grazie Antonio! LGTM

verbosus:
Wow, Travis really does take forever. A few minor improvements:

- I added tests for `CocoaPen` too, since it didn’t have any;
- I removed an unused import in the `QuartzPen` tests;

I think it would be nice if the Travis osx build environment installed `pyobjc` by default, so the tests actually run there, but I don’t know how to do that.

anthrotype:
> it would be nice if the Travis osx build environment installed pyobjc by default

sure, but maybe better to do that after we hafe completed the migration to Github Actions (#2100)

anthrotype:
can you please try to rebase off master? I just merged GH Actions PR

verbosus:
Done, I’ve rebased the PR off of ae07ce6e554f5b4a2d6e7a6524f5a1307b307bcd and squashed all my commits in one.

PR #2110, opened 2020-11-20
Optionally install PyObjC if on macOS test env

### LINKED ISSUE #2108: Add PyObjC library to the macOS test environment
As per [this discussion](https://github.com/fonttools/fonttools/pull/2107#issuecomment-730500161) it would be handy if we  could add the `pyobjc` library to the pip installed packages, _on the macOS test environment only_. That would ensure all the tests, includng the platform-specific ones like the `CocoaPen` and `QuartzPen` tests are always excercised. Cosimo [suggests](https://github.com/fonttools/fonttools/pull/2107#issuecomment-730549189) we tackle this after the Travis → Github Actions (#2100) migration is complete.

verbosus:
Trying my luck with GHA conditional syntax (is there a way to see what a workflow change would do without pushing to github?) in order to fix #2108. The updated workflow should `pip install pyobjc` only on macOS.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,6 +43,9 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install packages
       run: pip install tox coverage
+    - name: (Optionally install PyObjC package if on macOS platform)
+      if: ${{ runner.os == "macOS" }}
+      run: pip install pyobjc
     - name: Run Tox
       run: tox -e py-cov
     - name: Run Tox without lxml
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,9 +43,6 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install packages
       run: pip install tox coverage
-    - name: (Optionally install PyObjC package if on macOS platform)
-      if: ${{ runner.os == "macOS" }}
-      run: pip install pyobjc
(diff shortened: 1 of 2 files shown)


verbosus:
OK, so if the PR gets approved then the GHA runner will actually execute and we’ll get to see if the changes where any good? 🤪

anthrotype:
the GH action workflow has not run because it says "invalid workflow file".. There seems to be a syntax error somewhere

https://github.com/fonttools/fonttools/actions/runs/374300397

anthrotype:
whatever the syntax error in the yaml config, I think the best way to pip install some test dependency is to update the `requirements.txt` file

https://github.com/fonttools/fonttools/blob/master/requirements.txt

and add the following conditional requirement:

`pyobjc; sys_platform == 'darwin'"`

chrissimpkins:
> the GH action workflow has not run because it says "invalid workflow file".. There seems to be a syntax error somewhere
> 
> 
> 
> https://github.com/fonttools/fonttools/actions/runs/374300397

It looks like you need single quotes on string literals in expressions.  🤷‍♂ 

https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#literals

verbosus:
OK, I reverted the change in the GHA workflow and updated requirements.txt as per Cosimo’s request (I pinned the version to the latest one which works here.)

PR #2112, opened 2020-11-20
[subset] Allow LangSys tags in --layout-scripts

khaledhosny:
Allow specifying which LangSys tags to accept for each script, using `script.lang` form. For example:

    --layout-scripts=arab.dflt,arab.URD,latn

To keep `DefaultLangSys` and `URD` language for `arab` script, and all languages for `latn` script.


anthrotype:
thanks Khaled

PR #2113, opened 2020-11-23
Distinguish between axis tag and id.

bungeman:
Inside an OpenType font an axis is referred to by its index. In ttx
the axis tag was used as an id but the axis tag is not required to be
unique. Explicitly create and use an axis id which is separate from the
axis tag. This allows for proper handling of fonts with multiple axes
which specify the same tag.

Note that this explicitly does not touch the non-fvar axis tags in the
STAT table, which are a separate set of tags.

Implements #1581.

### DIFF
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -599,8 +599,8 @@ class FontBuilder(object):
 
         assert "fvar" in self.font, "fvar must to be set up first"
         assert "CFF2" in self.font, "CFF2 must to be set up first"
-        axisTags = [a.axisTag for a in self.font["fvar"].axes]
-        varRegionList = buildVarRegionList(regions, axisTags)
+        axisIds = [axis.axisId for axis in self.font["fvar"].axes]
+        varRegionList = buildVarRegionList(regions, axisIds)
         varData = buildVarData(list(range(len(regions))), None, optimize=False)
         varStore = buildVarStore(varRegionList, [varData])
         vstore = VarStoreData(otVarStore=varStore)
@@ -830,10 +830,17 @@ def addFvar(font, axes, instances):
 
     fvar = newTable('fvar')
     nameTable = font['name']
+    axisTagsSeen = {}
 
     for tag, minValue, defaultValue, maxValue, name in axes:
         axis = Axis()
         axis.axisTag = Tag(tag)
+        if axis.axisTag in axisTagsSeen:
+            axisTagsSeen[axis.axisTag] += 1
+            axis.axisId = f"{axis.axisTag}#{axisTagsSeen[axis.axisTag]}"
+        else:
+            axisTagsSeen[axis.axisTag] = 0
+            axis.axisId = axis.axisTag
         axis.minValue, axis.defaultValue, axis.maxValue = minValue, defaultValue, maxValue
         axis.axisNameID = nameTable.addName(tounicode(name))
         fvar.axes.append(axis)
--- a/Lib/fontTools/ttLib/tables/TupleVariation.py
+++ b/Lib/fontTools/ttLib/tables/TupleVariation.py
@@ -62,20 +62,20 @@ class TupleVariation(object):
 		"""
 		return any(c is not None for c in self.coordinates)
 
-	def toXML(self, writer, axisTags):
+	def toXML(self, writer, axisIds):
 		writer.begintag("tuple")
 		writer.newline()
-		for axis in axisTags:
-			value = self.axes.get(axis)
+		for axisId in axisIds:
+			value = self.axes.get(axisId)
 			if value is not None:
 				minValue, value, maxValue = value
 				defaultMinValue = min(value, 0.0)  # -0.3 --> -0.3; 0.7 --> 0.0
 				defaultMaxValue = max(value, 0.0)  # -0.3 -->  0.0; 0.7 --> 0.7
 				if minValue == defaultMinValue and maxValue == defaultMaxValue:
-					writer.simpletag("coord", axis=axis, value=fl2str(value, 14))
+					writer.simpletag("coord", axis=axisId, value=fl2str(value, 14))
 				else:
 					attrs = [
-						("axis", axis),
+						("axis", axisId),
 						("min", fl2str(minValue, 14)),
 						("value", fl2str(value, 14)),
 						("max", fl2str(maxValue, 14)),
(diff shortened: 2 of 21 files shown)


justvanrossum:
Thanks! It's good to see this approach implemented.

I haven't formed a proper opinion yet (and I find it hard to assess backwards compatibility, although it looks good so far).

I'm a little uncomfortable with the added complexity, which seems to burden the common case.

In the TTX format there is now an "axis name" which maps to an "axis ID", which I find a little confusing.

#2104 is admittedly a bit of a hack, just like our glyph name disambiguation scheme is a bit of a hack. However, duplicate glyph names in the 'post' table are more likely a mistake than a feature, whereas duplicate axis tags are now becoming a feature rather than a mistake. So perhaps the solution should indeed be better than "a bit of a hack". 

(On the other hand, I find the simplicity of #2104 appealing, despite it being a bit of a hack.)

Curious to learn what other people think.

bungeman:
> In the TTX format there is now an "axis name" which maps to an "axis ID", which I find a little confusing.

Yeah, that seems strange. The issue is that elsewhere the `id` attribute is informative but unused with `name` being the actual id; internally `axisNameID`  already exists which makes using `axisName` a bit confusing. I would be fine with the attribute being `id` instead of `name` here if we think that makes more sense.

Part of the difficulty with this is that really the axes don't have ids, the ids are owned by the `fvar` table and the order in which the axes are laid out there. It would be nice if the object model for creating an `Axis` were more like `my_axis = font["fvar"].new_axis()` or something like that so that the `new_axis` could set the `axisId` and so that it was more obvious that the `Axis` was tied to that particular table. Either that or something like `font["fvar"].add_axis(my_axis)` which could actually append a copy of `my_axis` with a made up `axisId` if `my_axis.id is None` and check for dupes instead of just appending to `font["fvar"].axis` directly.

Before this change the code assumed the `axisTag` would all be unique and silently did odd things if not. In this change `table__f_v_a_r` and `Axis` conspire to check that all created `Axis` have `axisId`. However, the tests which create an `axis` directly with `Axis()` also needed to be updated to maintain the setting of the `axisId`. It would be nice if it were enforced more directly instead of implicitly.

This extends to the generation of `axisId`s in `addFvar` (`fontbuilder`) and `_add_fvar` (DesignSpace) which continue to work the way they used to, except that if the same tag is used in multiple axes that is handled by creating new unique `axisId`. In theory these could be extended to allow for an `axisId` key to be allowed, but then the `axisId`s would need to be checked for uniqueness (the code prior to this change did not check the `tag`s for uniqueness). I suppose this is an actual difference with #2104, in this change the user would specify multiple axes with the same tag by just doing so; with #2104 the user would need to manage the axis name mangling (and maybe the code in `addFvar` and `_add_fvar` would need to be updated to not allow duplicate extended tags). It seems `fontbuilder` is experimental and can change, so maybe it should. It seems designspaceLib pretty much has its own view of axes and only ever dips into _add_fvar when serializing.

justvanrossum:
> I suppose this is an actual difference with #2104

Yes, and I think this is an improvement. So even if the approach of #2104 were to be chosen, it needs to be extended to `_add_fvar()`.

anthrotype:
> really the axes don't have ids, the ids are owned by the fvar table and the order in which the axes are laid out there.

That's the key point here. The _index_ of an axis is what uniquely identifies them, not their tag. That's how axes are referenced everywhere in the other OT tables. Before we were simply assuming the tag were also always unique, whereas now we need to change that assumption and instead use the position of the axis record within the fvar axes array as the unique identifier.

In places where we really want to have a dictionary of axes with unique keys, instead of a list of axis records, we could create such dict by e.g. `{f"axis.axisTag#index" for index, axis in enumerate(fvar.axes)}`.

I don't think fvr `Axis` object should keep record of an `axisId` which doesn't exist in the spec, and is essentially just its position within the container axes list. That could always be derived. E.g. if you have an axis object and want to know what is its index, you'd do `fvar.axes.index(axis)` in linear time (yes, you'd also need a pointer to the parent axes list, the object itself woudn't know).

THB I haven't thought about all the implications of this change yet, but wanted to drop my initial 2 cents.

anthrotype:
The `AxisDescriptor` object in `fontTools.designspaceLib` has a required `name` attribute that is guaranteed (or intended) to be unique since it is used as the key for the masters/instances locations (dicts). This is distinct for the `tag` attribute, also present in dsLib AxisDescriptor (tag was added later when designspace was repurposed for building VFs, originally it was designed for interpolating UFOs). This `name` attribute is for internal use by designspace, and is logically distinct from the axis `labelname` elements that are the user-visible localizable names (and that get encoded in the VF name table). 
The `name` is completely arbitrary (usually a lowercase name for the axis that is not restricted to 4 characters). It doesn't have any relation with the order of the designspace axes, which has its intrinsic order (i.e. the order in which axis child elements appear in the axes parent element). Conceptualy designspace axes are like an ordered dictionary of axis descriptors keyed by axis.name.

https://fonttools.readthedocs.io/en/latest/designspaceLib/readme.html#id6

I'm starting to warm up to the idea that fvar Axis object could have such a "name" attribute as well, that is guaranteed to be unique (with `<tag>#<index>` disambiguating duplicated tags), and used that as the unique key to quickly look up the ordered dict of axes.

One way to avoid having to store the `name` attribute onto the Axis object (where it arguable doesn't belong) could be turning the whole `fvar.axes` into an OrderedDict keyed by this unique name, with the values being Axis objects (in turn containing possibly duplicated axisTag, etc.). That could indeed break a lot existing code that assumes fvar.axes is a plain list of Axis.

bungeman:
> really the axes don't have ids, the ids are owned by the fvar table and the order in which the axes are laid out there.

Just to add to that, in the OpenType format axes kinda do and kinda don't have ids, which works because an actual axis definition doesn't exist separate from its fvar entry. So in the format itself one can think of either the axis definition having an id which is its index or you can think of the fvar table having a mapping from index to axis. These are both equivalent in the font file itself where there is no create, update, or delete of axes (only read). Part of the issue in the tests, or with the idea of just being able to create random axes like `axis = Axis()`, is allowing an `Axis` to exist which isn't associated with an fvar table and in some sense should not be referred to until it is part of one and only one fvar table (and its index isn't going to change).

This is somewhat at odds a bit with the XML format, where currently axes are referred to by an `axis` attribute. If there were a real 1:1 mapping here between file format and XML, the `axis` attribute value would just be a non-negative number which is the index of the axis in the fvar table (not unlike xPOS referring to everything by index). It is nice to lift this a little and allow an enumeration of axes (give the indexes unique names). Using this "every problem can be solved by another layer of indirection", the fvar table could own the list of name to index mappings. But the easiest way to represent that in the XML would be to just give each `<Axis>` a `name` attribute. However, it isn't necessary for the `Axis` type to actually cary that name, the fvar table could just have the mapping. Of course that means `fvar.axes` needs to be smarter, leading to it not being a plain list.

As a somewhat silly idea... just model things the way the user API would. In the XML or API any `axis` attribute which is a non-negative integer could refer to the axis with that index in the fvar table. Any `axis` attribute otherwise would be applied to any and all axes with the given tag (potentially more than one). Internally always refer to the axes by index. This would change all of the `for axisId in axisIds` into `for axisIndex in range(ttFont["fvar"].axes.size())` and mean not passing around all the `axisIds` at all. This has the downside of doing away with the nice naming of the axes, forcing the ordering of the `<Axes>` to be meaningful. This suggestion is more the alternate view that the axes do have implicit ids, it's just that currently the `axis` attributes don't use that id at the moment (necessitating a new one that looks like the one they currently use).

bungeman:
Since I just noticed it, in the `<STAT>` table XML serialization the `<DesignAxisRecord>` has its own `<Axis>` with an `index` attribute. So it would at least be consistent to instead give the `<fvar>` `<Axis>` an `index` attribute and then later have any `axis` attributes which refer to that axis specify it by an index. For backwards compatibility (and for interesting effect) any `axis` attribute which looks like a four byte tag could apply to all axes with that tag.

behdad:
It's too bad this work is stable now. I think we should resurrect it.

behdad:
Or does https://github.com/fonttools/fonttools/pull/2104/files supersede this?

Issue #2111, opened 2020-11-20
[ufoLib] GlyphSet should raise if passed path contains no contents.plist


madig:
Pass `GlyphSet` a `path` that contains no `contents.plist` and it will return an empty glyph set. The [spec however states](https://unifiedfontobject.org/versions/ufo3/glyphs/):

> The directory may contain zero or more GLIF files. It must contain a contents.plist file defining the file name to glyph name mapping even if no GLIF files are present.

So it should explode. The code seems to intentionally let it slide though: https://github.com/fonttools/fonttools/blob/4.17.0/Lib/fontTools/ufoLib/glifLib.py#L192-L193. Original commit: https://github.com/fonttools/fonttools/commit/ba16039873654959dbc31f26d0318b34d79a8caf -- is this still relevant? @typesupply 

Found via https://github.com/source-foundry/ufolint/pull/82.

madig:
Cosimo says it's best not to change this directly:

```
you see, the problem is GlyphSet class represent a R/W filesystem
it's used both for reading an existing path (where contents is expected to be present)
and for writing to a new (previously empty) directory
imagine that one wants to create a new GlyphSet in an empty directory, it'd be stupid to raise an error if contents.plist is missing, of course it is
on the other hand, if one attempts to read a glyph set from a directory that does not contain a contents.plist, it may be a good idea to raise error if there isn't one
so I can see why Tal did it that way

let's then add a new method to GlyphSet, something like "expectContentsFile" that raises if it's not there
you should not reach to the underlying filesystem
from the client library
you can't make assumptions about the underlying filesystem
definitely not use os module
add a new method that raises if not present
or make that _havePreviousFile a public property
but also have UFOReader.getGlyphSet() raise if contents is not there
so that we also fix ufoLib2/defcon
```

typesupply:
Cosimo read my mind! 😄

madig:
Argh. Adding `expectContentsFile` to `GlyphSet` is straightforward, but adding a check to `UFO*.getGlyphSet` is not, because it seems to be the canonical way of instantiating a fresh new glyph set, e.g. in https://github.com/fonttools/fonttools/blob/4.17.1/Tests/ufoLib/UFO3_test.py#L4574-L4577 and https://github.com/fonttools/fonttools/blob/4.17.1/Tests/ufoLib/UFO3_test.py#L3945-L3948.

anthrotype:
[...]

PR #2114, opened 2020-11-23
Add expectContentsFile parameter to GlyphSet

madig:
For use when reading existing UFOs, to comply with the specification stating that a contents.plist file must exist in a glyph set.

Avoid unconditionally checking for the file because `UFO(Reader|Writer).getGlyphSet` and therefore indirectly `GlyphSet` are used to create new fresh glyph sets.

Closes https://github.com/fonttools/fonttools/issues/2111.


madig:
The test suites for defcon, fontTools and ufoLib2 pass with this change :)

anthrotype (approved):
LGTM

PR #2117, opened 2020-12-02
Switch from brotlipy to brotlicffi for PyPy support

sethmlarson:
brotlipy has been renamed to brotlicffi to no longer conflict with the import namespace `brotli` and now guarantees the same API as Google's C bindings.

This example project shows how to use the `brotlicffi` library: https://github.com/python-hyper/brotlicffi/tree/master/example


anthrotype:
Thank you!

PR #2119, opened 2020-12-04
Cache tuples

rsheeter:
Makes a huge difference when processing complex svgs to COLR. On my macbook:

```
time nanoemoji ~/oss/noto-emoji/third_party/waved-flags/svg/emoji_u1f1f7_1f1f8.svg
  before: real  1m57.618s
  after: real  1m14.734s
```

Also shows in profile, see https://github.com/googlefonts/nanoemoji/issues/218

### DIFF
--- a/Lib/fontTools/colorLib/builder.py
+++ b/Lib/fontTools/colorLib/builder.py
@@ -474,8 +474,8 @@ class LayerV1ListBuilder:
                 return tuple(_tuple_safe(e) for e in value)
             return value
 
-        # Cache the tuples for individual Paint instead of the a whole sequence
-        #  because the seq could be a transient slice
+        # Cache the tuples for individual Paint instead of the whole sequence
+        # because the seq could be a transient slice
         result = self.tuples.get(id(paint), None)
         if result is None:
             result = _tuple_safe(paint)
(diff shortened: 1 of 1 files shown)


anthrotype (approved):

anthrotype:
hm actually there's something fishy going on, see test failure for 3.9 (while the other pass)
https://github.com/fonttools/fonttools/runs/1502516748?check_suite_focus=true#step:6:438

Lib/fontTools/colorLib/builder.py
```diff
@@ -465,11 +451,35 @@ class LayerV1ListBuilder:
...
         self.reusePool = {}
+        self.tuples = {}
+
+    def _paint_tuple(self, paint: ot.Paint):
+        # start simple, who even cares about cyclic graphs or interesting field types
+        def _tuple_safe(value):
+            if isinstance(value, enum.Enum):
+                return value
+            elif hasattr(value, "__dict__"):
+                return tuple((k, _tuple_safe(v)) for k, v in value.__dict__.items())
```
anthrotype, line 468: maybe we need to sort the `__dict__.items()`?

anthrotype:
Relying on the`id(object)` for the tuple cache is error prone, because id() is unique only thoughout the object's lifetime; if the latter is garbage-collected, the same `id` (which in cpython means the object memory address) can be reused by a new object
Either we need to keep all the paints alive (right now we're garbaging some paints after we find a reuse), or we should try to use more unique identifiers (e.g. uuid4)

See https://stackoverflow.com/a/52181584

we wouldn't have this problem if paints were immutable and hashable.

rsheeter:
> id() is unique only thoughout the object's lifetime

Good catch; I think we can safely keep them all alive

PR #2137, opened 2020-12-24
codecs: handle errors different from 'strict' for extended mac encodings

### LINKED ISSUE #2132: `name` table decoding error
While using the following code:
```
for record in font['name'].names:
    if record.nameID == 1:
        print(record)
```
It gave an error for some of the Chinese fonts I have:
```
  File "C:\Python39\lib\site-packages\fontTools\encodings\codecs.py", line 54, in decode
    assert errors == 'strict'
AssertionError

The above exception was the direct cause of the following exception:

  File "d:\Desktop\coding\fontname\main.py", line 75, in <module>
    print(record)
  File "C:\Python39\lib\site-packages\fontTools\ttLib\tables\_n_a_m_e.py", line 414, in __str__
    return self.toStr(errors='backslashreplace')
  File "C:\Python39\lib\site-packages\fontTools\ttLib\tables\_n_a_m_e.py", line 460, in toUnicode
    string = tounicode(string, encoding=encoding, errors=errors)
  File "C:\Python39\lib\site-packages\fontTools\misc\py23.py", line 82, in tounicode
    return s.decode(encoding, errors)
AssertionError: decoding with 'x_mac_simp_chinese_ttx' codec failed (AssertionError: )
```
It seems to be stuck on this section (based on `ttx`):
```
    <namerecord nameID="1" platformID="3" platEncID="1" langID="0x804">
      汉仪彩云体简
    </namerecord>
```

anthrotype:
Fixes #2132

### DIFF
--- a/Lib/fontTools/encodings/codecs.py
+++ b/Lib/fontTools/encodings/codecs.py
@@ -16,43 +16,29 @@ class ExtendCodec(codecs.Codec):
 		self.info = codecs.CodecInfo(name=self.name, encode=self.encode, decode=self.decode)
 		codecs.register_error(name, self.error)
 
-	def encode(self, input, errors='strict'):
-		assert errors == 'strict'
-		#return codecs.encode(input, self.base_encoding, self.name), len(input)
-
-		# The above line could totally be all we needed, relying on the error
-		# handling to replace the unencodable Unicode characters with our extended
-		# byte sequences.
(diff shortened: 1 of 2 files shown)


justvanrossum (approved):
I'm not sure I understand enough of what's going on to make my review trustworthy, but it looks good to me.

PR #2146, opened 2021-01-13
Subsetting: prune redundant format 12 cmaps

madig:
Behdad was prescient at https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/subset/__init__.py#L2228-L2231.

My idea: 

1. Delete a `platformID="0" platEncID="4" format="12" language="X"` (Unicode platform, Unicode full repertoire) cmap if there is a `platformID="0" platEncID="3" language="X"` (Unicode BMP only) cmap that contains the same Unicode set after subsetting.
2. Delete a `platformID="3" platEncID="10" format="12" language="X"` (Windows platform, Unicode full repertoire) cmap if there is a `platformID="3" platEncID="1" language="X"` (Unicode BMP) cmap that contains the same Unicode set after subsetting.

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -2207,7 +2207,17 @@ def prune_pre_subset(self, font, options):
 @_add_method(ttLib.getTableClass('cmap'))
 def subset_glyphs(self, s):
 	s.glyphs = None # We use s.glyphs_requested and s.unicodes_requested only
+
+	tables_format12_bmp = []
+	table_plat0_enc3 = {}  # Unicode platform, Unicode BMP only, keyed by language
+	table_plat3_enc1 = {}  # Windows platform, Unicode BMP, keyed by language
+
 	for t in self.tables:
+		if t.platformID == 0 and t.platEncID == 3:
+			table_plat0_enc3[t.language] = t
+		if t.platformID == 3 and t.platEncID == 1:
+			table_plat3_enc1[t.language] = t
+
 		if t.format == 14:
 			# TODO(behdad) We drop all the default-UVS mappings
 			# for glyphs_requested.  So it's the caller's responsibility to make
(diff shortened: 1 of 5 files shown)


Lib/fontTools/subset/__init__.py
```diff
@@ -2207,28 +2207,58 @@ def prune_pre_subset(self, font, options):
...
+
 		if t.format == 14:
 			# TODO(behdad) We drop all the default-UVS mappings
 			# for glyphs_requested.  So it's the caller's responsibility to make
 			# sure those are included.
 			t.uvsDict = {v:[(u,g) for u,g in l
-					      if g in s.glyphs_requested or u in s.unicodes_requested]
-				     for v,l in t.uvsDict.items()}
+						  if g in s.glyphs_requested or u in s.unicodes_requested]
+					 for v,l in t.uvsDict.items()}
```
anthrotype: is the whitespace change only cosmetic? (one day we'll blacken everything so I wouldn't bother)

Lib/fontTools/subset/__init__.py
```diff
@@ -2207,28 +2207,58 @@ def prune_pre_subset(self, font, options):
 @_add_method(ttLib.getTableClass('cmap'))
 def subset_glyphs(self, s):
 	s.glyphs = None # We use s.glyphs_requested and s.unicodes_requested only
+
+	tables_format12 = []
+	table_plat0_enc3 = {}
+	table_plat3_enc1 = {}
```
anthrotype, line 2213: pls add comment describing what `plat` and `enc` number means

Lib/fontTools/subset/__init__.py
```diff
@@ -2207,28 +2207,58 @@ def prune_pre_subset(self, font, options):
...
+		):
+			t.cmap.clear()
+
 	self.tables = [t for t in self.tables
 			 if (t.cmap if t.format != 14 else t.uvsDict)]
 	self.numSubTables = len(self.tables)
-	# TODO(behdad) Convert formats when needed.
-	# In particular, if we have a format=12 without non-BMP
-	# characters, either drop format=12 one or convert it
-	# to format=4 if there's not one.
```
anthrotype, line 2231: please keep the TODO about `convert it to format=4 if there's not one`

Lib/fontTools/subset/__init__.py
```diff
@@ -2207,28 +2207,58 @@ def prune_pre_subset(self, font, options):
...
 			t.cmap = {u:g for u,g in t.cmap.items()
-				      if g in s.glyphs_requested}
+					  if g in s.glyphs_requested}
+
+	# Fomat 12 tables are redundant if they contain just the same BMP codepoints
+	# their little BMP-only encoding siblings contain.
+	for t in tables_format12:
+		if (
+			t.platformID == 0
+			and t.platEncID == 4
```
anthrotype, line 2245: comment what platEncID 4 and 10 mean

Lib/fontTools/subset/__init__.py
```diff
@@ -2219,16 +2229,36 @@ def subset_glyphs(self, s):
...
 				      if g in s.glyphs_requested}
+
+	# Fomat 12 tables are redundant if they contain just the same BMP codepoints
+	# their little BMP-only encoding siblings contain.
+	for t in tables_format12:
+		if (
+			t.platformID == 0
+			and t.platEncID == 4
+			and t.language in table_plat0_enc3
+			and set(table_plat0_enc3[t.language].cmap) == set(t.cmap)
```
anthrotype, line 2247: why cast to set? can't you compare the dicts directly? or their .keys()?

Lib/fontTools/subset/__init__.py
madig: Removed, that was unintentional.

Lib/fontTools/subset/__init__.py
madig: Works :o

anthrotype (approved):
LGTM

PR #2148, opened 2021-01-14
COLRv1: avoid abrupt change after rounding c0 when too near c1's perimeter

anthrotype:
Fixes encoding a tricky radial gradient in Noto Emoji's 🦟 (U+1F99F) https://github.com/googlefonts/picosvg/issues/158
For more info also see discussion at https://github.com/googlefonts/colr-gradients-spec/issues/204

### DIFF
--- a/Lib/fontTools/colorLib/builder.py
+++ b/Lib/fontTools/colorLib/builder.py
@@ -34,7 +34,7 @@ from fontTools.ttLib.tables.otTables import (
     VariableInt,
 )
 from .errors import ColorLibError
-from .geometry import nudge_start_circle_almost_inside
+from .geometry import round_start_circle_stable_containment
 
 
 # TODO move type aliases to colorLib.types?
@@ -534,7 +534,7 @@ class LayerV1ListBuilder:
         r1 = _to_variable_value(r1)
 
         # avoid abrupt change after rounding when c0 is near c1's perimeter
-        c = nudge_start_circle_almost_inside(
+        c = round_start_circle_stable_containment(
             (x0.value, y0.value), r0.value, (x1.value, y1.value), r1.value
         )
         x0, y0 = x0._replace(value=c.centre[0]), y0._replace(value=c.centre[1])
--- a/Lib/fontTools/colorLib/builder.py
+++ b/Lib/fontTools/colorLib/builder.py
@@ -34,6 +34,7 @@ from fontTools.ttLib.tables.otTables import (
     VariableInt,
 )
 from .errors import ColorLibError
+from .geometry import nudge_start_circle_almost_inside
 
 
 # TODO move type aliases to colorLib.types?
@@ -328,9 +329,9 @@ def _split_color_glyphs_by_version(
 
 def _to_variable_value(
     value: _ScalarInput,
-    minValue: _Number,
-    maxValue: _Number,
-    cls: Type[VariableValue],
+    cls: Type[VariableValue] = VariableFloat,
+    minValue: Optional[_Number] = None,
+    maxValue: Optional[_Number] = None,
 ) -> VariableValue:
     if not isinstance(value, cls):
         try:
@@ -339,9 +340,9 @@ def _to_variable_value(
             value = cls(value)
         else:
             value = cls._make(it)
-    if value.value < minValue:
+    if minValue is not None and value.value < minValue:
         raise OverflowError(f"{cls.__name__}: {value.value} < {minValue}")
-    if value.value > maxValue:
+    if maxValue is not None and value.value > maxValue:
         raise OverflowError(f"{cls.__name__}: {value.value} < {maxValue}")
     return value
 
@@ -526,7 +527,20 @@ class LayerV1ListBuilder:
         ot_paint.Format = int(ot.Paint.Format.PaintRadialGradient)
         ot_paint.ColorLine = _to_color_line(colorLine)
 
-        for i, (x, y), r in [(0, c0, r0), (1, c1, r1)]:
+        # normalize input types (which may or may not specify a varIdx)
+        x0, y0 = _to_variable_value(c0[0]), _to_variable_value(c0[1])
+        r0 = _to_variable_value(r0)
+        x1, y1 = _to_variable_value(c1[0]), _to_variable_value(c1[1])
+        r1 = _to_variable_value(r1)
+
+        # avoid abrupt change after rounding when c0 is near c1's perimeter
+        c0x, c0y = nudge_start_circle_almost_inside(
+            (x0.value, y0.value), r0.value, (x1.value, y1.value), r1.value
+        )
+        x0, y0 = x0._replace(value=c0x), y0._replace(value=c0y)
+
+        for i, (x, y, r) in enumerate(((x0, y0, r0), (x1, y1, r1))):
+            # rounding happens here as floats are converted to integers
             setattr(ot_paint, f"x{i}", _to_variable_int16(x))
             setattr(ot_paint, f"y{i}", _to_variable_int16(y))
             setattr(ot_paint, f"r{i}", _to_variable_uint16(r))
--- a/Lib/fontTools/colorLib/builder.py
+++ b/Lib/fontTools/colorLib/builder.py
@@ -534,10 +534,11 @@ class LayerV1ListBuilder:
         r1 = _to_variable_value(r1)
 
         # avoid abrupt change after rounding when c0 is near c1's perimeter
-        c0x, c0y = nudge_start_circle_almost_inside(
+        c = nudge_start_circle_almost_inside(
             (x0.value, y0.value), r0.value, (x1.value, y1.value), r1.value
         )
-        x0, y0 = x0._replace(value=c0x), y0._replace(value=c0y)
+        x0, y0 = x0._replace(value=c.centre[0]), y0._replace(value=c.centre[1])
+        r0 = r0._replace(value=c.radius)
 
         for i, (x, y, r) in enumerate(((x0, y0, r0), (x1, y1, r1))):
             # rounding happens here as floats are converted to integers
--- a/Lib/fontTools/colorLib/geometry.py
+++ b/Lib/fontTools/colorLib/geometry.py
@@ -4,12 +4,8 @@ from math import copysign, cos, hypot, pi
 from fontTools.misc.fixedTools import otRound
 
 
-def _vector_between_points(a, b):
-    return (b[0] - a[0], b[1] - a[1])
-
-
-def _distance_between_points(a, b):
-    return hypot(*_vector_between_points(a, b))
+def _vector_between(origin, target):
+    return (target[0] - origin[0], target[1] - origin[1])
 
 
 def _round_point(pt):
@@ -33,7 +29,7 @@ _NEARLY_ZERO = 1 / (1 << 12)  # 0.000244140625
 
 
 def _is_circle_inside_circle(c0, r0, c1, r1):
-    dist = r0 + _distance_between_points(c0, c1)
+    dist = r0 + hypot(*_vector_between(c0, c1))
     return abs(r1 - dist) <= _NEARLY_ZERO or r1 > dist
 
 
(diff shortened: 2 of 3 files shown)


Lib/fontTools/colorLib/geometry.py
```diff
@@ -0,0 +1,110 @@
...
+from math import copysign, cos, sqrt, pi
+from fontTools.misc.fixedTools import otRound
+
+
+def _vector_between_points(a, b):
+    return (b[0] - a[0], b[1] - a[1])
+
+
+def _vector_length(vec):
+    return sqrt(vec[0] * vec[0] + vec[1] * vec[1])
```
justvanrossum, line 12: You could consider `math.hypot()`.

anthrotype: good idea thanks

Lib/fontTools/colorLib/geometry.py
```diff
@@ -1,19 +1,15 @@
 """Helpers for manipulating 2D points and vectors in COLR table."""
 
-from math import copysign, cos, sqrt, pi
+from math import copysign, cos, hypot, pi
 from fontTools.misc.fixedTools import otRound
 
 
 def _vector_between_points(a, b):
```
rsheeter, line 7: suggest descriptive names; vector_between(origin, target) or w/e

Lib/fontTools/colorLib/geometry.py
```diff
@@ -1,19 +1,15 @@
...
     return (b[0] - a[0], b[1] - a[1])
 
 
-def _vector_length(vec):
-    return sqrt(vec[0] * vec[0] + vec[1] * vec[1])
-
-
 def _distance_between_points(a, b):
-    return _vector_length(_vector_between_points(a, b))
+    return hypot(*_vector_between_points(a, b))
```
rsheeter, line 12: given hypot I wonder if we even need this function?

Lib/fontTools/colorLib/builder.py
```diff
@@ -526,7 +527,20 @@ def buildPaintRadialGradient(
...
         ot_paint.ColorLine = _to_color_line(colorLine)
 
-        for i, (x, y), r in [(0, c0, r0), (1, c1, r1)]:
+        # normalize input types (which may or may not specify a varIdx)
+        x0, y0 = _to_variable_value(c0[0]), _to_variable_value(c0[1])
+        r0 = _to_variable_value(r0)
+        x1, y1 = _to_variable_value(c1[0]), _to_variable_value(c1[1])
+        r1 = _to_variable_value(r1)
+
+        # avoid abrupt change after rounding when c0 is near c1's perimeter
```
rsheeter, line 536: If acceptable in FT # <link to gh issue> would be nice

anthrotype: pardon?

Lib/fontTools/colorLib/geometry.py
anthrotype: done

Lib/fontTools/colorLib/geometry.py
anthrotype: i removed it

Lib/fontTools/colorLib/geometry.py
```diff
@@ -0,0 +1,106 @@
...
+        return None
+    return (vec[0] / length, vec[1] / length)
+
+
+# This is the same tolerance used by Skia's SkTwoPointConicalGradient.cpp to detect
+# when a radial gradient's focal point lies on the end circle.
+_NEARLY_ZERO = 1 / (1 << 12)  # 0.000244140625
+
+
+def _is_circle_inside_circle(c0, r0, c1, r1):
```
rsheeter, line 31: looks like you have to read the code to know which one is which, consider descriptive names? - c_inner, r_inner, etc might make this more obvious

Lib/fontTools/colorLib/geometry.py
```diff
@@ -0,0 +1,106 @@
...
+    https://github.com/googlefonts/colr-gradients-spec/issues/204
+    https://github.com/googlefonts/picosvg/issues/158
+    """
+    inside_before_round = _is_circle_inside_circle(c0, r0, c1, r1)
+    rc0, rr0 = _round_circle(c0, r0)
+    rc1, rr1 = _round_circle(c1, r1)
+    inside_after_round = _is_circle_inside_circle(rc0, rr0, rc1, rr1)
+
+    if inside_before_round != inside_after_round:
+        # at most 2 iterations ought to be enough to converge
```
rsheeter, line 83: doesnt' explain why

Lib/fontTools/colorLib/geometry.py
anthrotype: we have a sort of convention elsewhere in the code that c0 is the start circle and c1 is the end circle. It's not really inner or outer.

anthrotype: renamed the parameter like you suggested

rsheeter: I still find the name/parameters a bit unclear even with that convention. "given a start and an end circle tell me if circle is inside circle" would lead me to ask you to clarify if you mean is start inside end, is end inside start, or if it's fine as long as either one is inside the other.

Lib/fontTools/colorLib/geometry.py
```diff
@@ -0,0 +1,102 @@
...
+        for _ in range(2):
+            if rc0 == rc1:  # nowhere to nudge along a zero vector, bail out
+                break
```
rsheeter, line 86: if we know rounding broke us *and* that we can't fix it shouldn't we raise an error?

anthrotype: this is a limit case that we can't fix by simply moving around c0..

When c0 and c1 are on top of one another (or very close) _and_ the r0 and r1 are also almost equal.

We'd need another degree of freedom, e.g. changing r0

Lib/fontTools/colorLib/geometry.py
anthrotype: rounding can only move each by at most -/+0.5. All of x0, y0, r0, x1, y1, r1 roundings can contribute to changing the position of c0 relative to c1. In the worst case, {x0, y0, r0} and {x1, y1, r1} would round in opposite directions and the maximum error along either axies would be 0.5*2 + 0.5*2 == 2... Did I convice you? I guess not.

Lib/fontTools/colorLib/geometry.py
anthrotype: E.g. take this

```
c0 = (0.5000001, 0)
r0 = 0.500001
c1 = (0.499999, 0)
r1 = 0.49999999
```

Given our `_NEARLY_ZERO` tolerance, we consider c0 inside c1. They are literally on top of one another.
there's no way by controlling only where c0 rounds to, to restore the original inside-ness. The r1 is now 0, and r0 is 1 so they can't fit anymore.
I think we should not bother fixing these cases. They don't look like valid radial gradients that one would create in the first place.

Lib/fontTools/colorLib/builder.py
anthrotype: oh now I read your comment (the markdown formatting somehow broke it). Yes, I did add a link, its in the docstring of `nudge_start_circle_almost_inside`

Lib/fontTools/colorLib/geometry.py
```diff
@@ -0,0 +1,102 @@
...
+            if inside_before_round == inside_after_round:
+                break
+        else:  # ... or it's a bug
+            raise AssertionError(
+                f"Nudging circle <c0={c0}, r0={r0}> "
+                f"{'inside' if inside_before_round else 'outside'} "
+                f"<c1={c1}, r1={r1}> failed after two attempts!"
+            )
+
+    return rc0
```
rsheeter, line 102: Just playing, would something like this be equivalent?

```
def round_start_stable_containment(c0, r0, c1, r1):
  # round start such that it's containment in end remains the same
  # if necessary move start to ensure this
  c1_round, r1_round = _round_cricle(c1, r1)
  step = _vector_toward(c0, c1, 1.0)
  for _ in range(max_attempts):
    c_candidate, r_candidate = _round_circle(c0, r0)
    if (_is_c0_inside_c1(c0, r0, c1, r1)
        == _is_c0_inside_c1(c_candidate, r_candidate, c1_round, r1_round))
        return (c_candidate, r_candidate)
    c0, r0 = c_candidate + step, r_candidate
    
  # error out, we returned the answer if there was one
```

Lib/fontTools/colorLib/geometry.py
anthrotype: added explanation

Lib/fontTools/colorLib/geometry.py
anthrotype: I now handle the case of concentrical circles like this by updating r0

anthrotype:
@rsheeter Please take another look at the latest commits. I don't think I can simplify more than this.

Lib/fontTools/colorLib/geometry.py
```diff
@@ -0,0 +1,139 @@
...
+    The rounding of circle coordinates to integers may cause an abrupt change
+    if the start circle c0 is so close to the end circle c1's perimiter that
+    it ends up falling outside (or inside) as a result of the rounding.
+    To keep the gradient unchanged, we nudge it in the right direction.
+
+    See:
+    https://github.com/googlefonts/colr-gradients-spec/issues/204
+    https://github.com/googlefonts/picosvg/issues/158
+    """
+    start_circle, end_circle = Circle(c0, r0), Circle(c1, r1)
```
rsheeter, line 93: optional, I don't think you necessarily need _circle; method is still clear without

Lib/fontTools/colorLib/geometry.py
```diff
@@ -0,0 +1,139 @@
...
+    for _ in range(max_attempts):
+        inside_after_round = round_start.inside(round_end)
+        if inside_before_round == inside_after_round:
+            break
+        if round_start.concentric(round_end):
+            # can't move c0 towards c1 (they are the same), so we change the radius
+            if inside_after_round:
+                round_start.radius += 1.0
+            else:
+                round_start.radius -= 1.0
```
rsheeter, line 125: I don't think you should change direction in the loop; you never want to move one step forward and one step back. Compute step outside and in the loop move by step.

Applies to both radius adjustment and vector nudges.

anthrotype: but I'm not changing direction

rsheeter: protect against radius becoming <= 0?

anthrotype: it's ok if it's 0. Less than 0 not ok, I'll protect against that

rsheeter: If you make the step a function your loop is something like:

if inside_before_round == round_start.inside(round_end)
  break # solved!
step() # step doesn't change direction, it always bumps radius or nudges in a given direction that was determined pre-loop

rsheeter: Re changing direction, your loop has the ability to add one to radius one spin and subtract one the next. I don't think that's ever desired; you should always make the same adjustment each spin.

anthrotype: I can't determine pre-loop because I'm checking `inside_after_round = round_start.inside(round_end)` _inside_ the loop.
It's impossible to flip direction! The loop stops as soon as `inside_before_round == inside_after_round`; it continues only if they are `!=`. It can't nudge in one direction, still be `inside_before_round != inside_after_round` and subsequently nudge in the opposite direction.

anthrotype: I changed it to determine the direction to nudge outside of the loop, and placed an assert against negative radius (which I can't even reproduce empirically). I hope that addesses all the remaining issues.

Lib/fontTools/colorLib/geometry.py
```diff
@@ -0,0 +1,144 @@
...
+    # x and y axes moves it towards c1 by hypot(-1.0, -1.0) = 1.41421. Two of these
+    # moves cover twice that distance, which is enough to restore containment.
+    max_attempts = 2
+    for _ in range(max_attempts):
+        if round_start.concentric(round_end):
+            # can't move c0 towards c1 (they are the same), so we change the radius
+            round_start.radius += radius_delta
+            assert round_start.radius >= 0
+        else:
+            round_start.nudge_towards(direction)
```
rsheeter, line 134: Optional, I feel like move(direction, amount) might be nicer? - move would have no rounding or anything, it would just move in direction by amount

rsheeter (approved):

PR #2154, opened 2021-01-19
[varLib] featureVars.addFeatureVariationsRaw: Raise legible error message when script is missing a dflt langsys

madig:
I.e. when you forgot to add `languagesystem grek dflt;` to your feature file.

Before: 

> fontmake: Error: In 'MyFont.designspace': Generating fonts from Designspace failed: 'NoneType' object has no attribute 'FeatureIndex'

After:

> fontmake: Error: In 'MyFont.designspace': Generating fonts from Designspace failed: Feature variations require that the script 'grek' defines a default language system.

### DIFF
--- a/Lib/fontTools/varLib/featureVars.py
+++ b/Lib/fontTools/varLib/featureVars.py
@@ -10,7 +10,7 @@ from fontTools.ttLib.tables import otTables as ot
 from fontTools.otlLib.builder import buildLookup, buildSingleSubstSubtable
 from collections import OrderedDict
 
-from .errors import VarLibValidationError
+from .errors import VarLibError, VarLibValidationError
 
 
 def addFeatureVariations(font, conditionalSubstitutions, featureTag='rvrn'):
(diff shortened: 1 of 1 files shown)


justvanrossum (approved):

anthrotype (approved):

Lib/fontTools/varLib/featureVars.py
```diff
@@ -298,6 +298,11 @@ def addFeatureVariationsRaw(font, conditionalSubstitutions, featureTag='rvrn'):
         varFeatureIndex = gsub.FeatureList.FeatureRecord.index(varFeature)
 
         for scriptRecord in gsub.ScriptList.ScriptRecord:
+            if scriptRecord.Script.DefaultLangSys is None:
```
anthrotype, line 301: @madig @justvanrossum I wonder, instead of raising an error, we could simply skip the missing DefaultLangSys here?
I don't see the precise reason why it should be required.

madig: Well you can use it to force people to use `dflt`, which is good practice anyway I think?

Issue #2120, opened 2020-12-06
`ClassDefBuilder.canAdd` is inconsistent with `ClassDefBuilder.add`


dscorbett:
If `ClassDefBuilder.canAdd` returns true for a certain glyph class, it is supposed to be safe to call `ClassDefBuilder.add` on that glyph class. However, `add` asserts that each glyph is not in `glyphs_` before adding it to `glyphs_`. Therefore, if a glyph class contains the same glyph twice, `canAdd` may return true but `add` will throw an error.

I noticed this because of the optimization in #2101.

anthrotype:
Fixed by #2155

dscorbett:
@anthrotype, I don’t think this has been fixed. #2155 improved the error message but did not address this bug.

anthrotype:
oh ok. Then I didn't understand what the error is about :)
I'll try again when I get the chance

PR #2155, opened 2021-01-20
Raise an exception when ClassDefBuilder attempts to add a duplicate glyph.

tomarcher101:
There is currently an assertion in ClassDefBuild.add that stops a glyph that is already present in a class from being added again. However, there is no error reported, and thus it can be quite difficult to debug.

This PR changes this assertion into a raised exception, while also adding the related tests.

### DIFF
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -2575,7 +2575,7 @@ class ClassDefBuilder(object):
         self.classes_.add(glyphs)
         for glyph in glyphs:
             if glyph in self.glyphs_:
-                raise ValueError(f"Glyph {glyph} is already present in class.")
+                raise OpenTypeLibError(f"Glyph {glyph} is already present in class.", None)
             self.glyphs_[glyph] = glyphs
 
     def classes(self):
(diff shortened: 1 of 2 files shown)


Lib/fontTools/otlLib/builder.py
```diff
@@ -2574,7 +2574,8 @@ def add(self, glyphs):
             return
         self.classes_.add(glyphs)
         for glyph in glyphs:
-            assert glyph not in self.glyphs_
+            if glyph in self.glyphs_:
+                raise ValueError(f"Glyph {glyph} is already present in class.")
```
anthrotype, line 2578: there's an `OpenTypeLibError` custom exception in `fontTools.otlLib.error` module, you should use that instead

anthrotype:
hm, this reminded me of https://github.com/fonttools/fonttools/issues/2120

@dscorbett can you take a look?

Lib/fontTools/otlLib/builder.py
tomarcher101: Done. However, I'm unsure if there's anything useful I can pass in for the location?

anthrotype: set location to None, it will not be printed

anthrotype:
It's unclear to me when exactly this can occur. Could you provide a feature file snippet that reproduces this error?

dscorbett:
I don’t remember the exact circumstances where I saw this, but I think it was in a contextual substitution rule that used a class containing the same glyph multiple times.

PR #2161, opened 2021-01-28
WOFF2: An off-curve should stay off even with flags

madig:
This fixes outline corruption of glyphs that start with an off-curve and have e.g. the overlap bit set, previously they'd be turned into on-curves.

### DIFF
--- a/Tests/ttLib/woff2_test.py
+++ b/Tests/ttLib/woff2_test.py
@@ -1,6 +1,7 @@
 from fontTools.misc.py23 import *
 from fontTools import ttLib
 from fontTools.ttLib import woff2
+from fontTools.ttLib.tables import _g_l_y_f
 from fontTools.ttLib.woff2 import (
 	WOFF2Reader, woff2DirectorySize, woff2DirectoryFormat,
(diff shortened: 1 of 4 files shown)


Tests/ttLib/woff2_test.py
```diff
@@ -1220,6 +1221,20 @@ def test_roundtrip_no_glyf_and_loca_tables(self):
 		assert tmp.getvalue() == tmp2.getvalue()
 		assert ttFont.flavor == "woff2"
 
+	def test_roundtrip_overlap_bit(self):
```
anthrotype, line 1224: we don't actually "roundtrip" the overlap bit, but we drop it. 
```suggestion
	def test_roundtrip_off_curve_despite_overlap_bit(self):
```

anthrotype (approved):
🥇

madig:
Thanks for the fix 😬

anthrotype:
thank your for the bug! 🐛

justvanrossum:
Thanks!

(Same bug as #1771, #1772, #1774...)

anthrotype:
it did look familiar indeed :)

madig:
Any other bug we have to fix? :D

PR #2164, opened 2021-01-30
Pen maintenance

madig:
1. Remove the unnecessary `py23` import
2. Turn `AbstractPen` and `AbstractPointPen` into true abstract base classes and add typing
3. Add the missing `curve` method to `TTGlyphPen`

The ABC is recognized even for classes that do not subclass the abstract base classes (NullPen is freestanding):
```
>>> from fontTools.pens.basePen import AbstractPen, NullPen

>>> issubclass(NullPen, AbstractPen)
True

>>> n = NullPen()
>>> isinstance(n, AbstractPen)
True
```


justvanrossum (commented):
Thanks! Looks good to me. But please await Cosimo's review!

madig:
(BTW: I followed https://realpython.com/python-interface/ to learn about ABCs)

justvanrossum:
Minor: While `NullPen` is a good test case for `__subclasshook__`, shouldn't it subclass `AbstractPen` anyway?

madig:
I suppose it should :grimacing:

Lib/fontTools/pens/basePen.py
```diff
@@ -36,26 +36,50 @@
...
 			"decomposeSuperBezierSegment", "decomposeQuadraticSegment"]
 
 
-class AbstractPen(object):
-
-	def moveTo(self, pt):
+class AbstractPen(abc.ABC):
+	@classmethod
+	def __subclasshook__(cls, subclass: Any) -> bool:
+		return (
```
anthrotype, line 51: in the `__subclasshook__` you also need to add a check like `if cls is not AbstractPen: return NotImplemented`, to ensure concrete subclasses of AbstractPen will use the regular `issubclass` logic instead of the custom checks.
You don't want that issubclass return True even with concrete pens.

See
https://stackoverflow.com/questions/35858285/why-check-if-cls-is-the-class-in-subclasshook

Cf collections.abc or the example in the official docs does:
https://docs.python.org/3/library/abc.html#abc.ABCMeta.__subclasshook__

anthrotype:
some tests would be nice

anthrotype (approved):

PR #2166, opened 2021-02-01
fontBuilder: pass glyphMap to buildCOLR to sort base records by GID 

anthrotype:
COLR Base glyph records must be sorted by glyph index. The `colorLib.builder.buildCOLR` function has an optional `glyphMap` parameter that maps from glyph names to glyph indices (as returned from `TTFont.getReverseGlyphMap()`).
FontBuilder knows all that so it should pass it on to colorLib to ensure base records are sorted by GID.

### DIFF
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -770,7 +770,7 @@ class FontBuilder(object):
             self.font, conditionalSubstitutions, featureTag=featureTag
         )
 
-    def setupCOLR(self, colorLayers):
+    def setupCOLR(self, colorLayers, version=None, varStore=None):
         """Build new COLR table using color layers dictionary.
 
         Cf. `fontTools.colorLib.builder.buildCOLR`.
@@ -778,7 +778,9 @@ class FontBuilder(object):
         from fontTools.colorLib.builder import buildCOLR
 
         glyphMap = self.font.getReverseGlyphMap()
-        self.font["COLR"] = buildCOLR(colorLayers, glyphMap=glyphMap)
+        self.font["COLR"] = buildCOLR(
+            colorLayers, version=version, glyphMap=glyphMap, varStore=varStore
+        )
 
     def setupCPAL(
         self,
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -777,7 +777,8 @@ class FontBuilder(object):
         """
         from fontTools.colorLib.builder import buildCOLR
 
-        self.font["COLR"] = buildCOLR(colorLayers)
+        glyphMap = self.font.getReverseGlyphMap()
+        self.font["COLR"] = buildCOLR(colorLayers, glyphMap=glyphMap)
 
     def setupCPAL(
         self,


justvanrossum:
Why does `buildCOLR()` need GIDs to begin with? Isn't converting to GIDs and sorting typically handled by the table compiler?

anthrotype:
it doesn't use glyph indexes directly, but it has to create base glyph records and the spec wants these sorted by glyph id

justvanrossum:
But that sounds then like the sorting is done in the wrong place: why can't the table compiler deal with that?

anthrotype:
actually sorry. it looks like we are already doing that... 
https://github.com/fonttools/fonttools/blob/8612319487c298f7feee3c8653ca34d9d8f8e607/Lib/fontTools/ttLib/tables/otTables.py#L1232-L1249

anthrotype:
basically, unless one passes glyphMap, the table returned from buildCOLR will have the base records ordered in whatever order they are passed in in the source colorGlyphs dictionary (which maintains order in py3.6+).
But only upon compiling the table, they will be reordered following the TTFont glyph order.
I suppose that's acceptable? Still, since fontbuilder already knows the glyphMap it can well pass it through (to avoid the client having to save and reload the TTFont just to ensure the base records are sorted to the final order).

justvanrossum:
> (to avoid the client having to save and reload the TTFont just to ensure the base records are sorted to the final order)

I don't understand this bit. Can you elaborate?

If the order at that level has no further semantic meaning, and the low level compiler correctly sorts anyway, why bother with a glyph map in `buildCOLR()` at all?

anthrotype:
the COLRv0 table has a ColorLayers dictionary where order doesn't matter. The COLRv1 table is implemented as otData-style low-level table and the BaseGlyphV1List is not a dictionary, it is a list of BaseGlyphV1Records with an inner ordering (which the spec mandates to be ordered by GID).

justvanrossum:
> and the BaseGlyphV1List is not a dictionary

Perhaps it should be? Coverage tables are also stored sorted by GID, but the Python API still presents and accepts glyph names.

justvanrossum:
I don't think there's a single otData-based subtable that depends on GIDs in any way, as seen from the Python side.

anthrotype:
feaLib builder similarly takes a glyphMap because it needs one to create those Coverage table sorted by GID..

anthrotype:
we do store only glyphnames in COLRv1, not glyph indexes. but the base records need to be stored sorted by GID so we need a glyphMap

anthrotype:
I know this current hybrid situation (COLRv0 with a hand-coded nicer Python API vs COLRv1 with a dynamically generated low level API) is not ideal, but it helped iterating faster as new paint formats are being specified in the draft spec.. Yes, I agree with you ideally we would have a fancier interface on top of that.

anthrotype:
anyway this isn't really a problem, as compiled fonts will have them sorted correctly in the end. It's only a small annoyance when using fontBuilder to create fonts for unit testing, that the COLR table just built (but not yet compiled) may not have them ordered by glyph order until it does get compiled. Not a big deal.

justvanrossum:
Ah, I didn't realize that indeed Coverage (and some other tables) do require correct sorting on input. (Coverage actually checks and warns if it's wrong. Not sure why it doesn't just always sort.)

It doesn't need a fully hand coded API for this detail to be done right: many tables have (small) bits of additional manual logic on top of the dynamic otTables-based generated classes, such as Coverage and SingleSubst. SingleSubst is also a good example of how to handle this.

anthrotype:
ok let's keep this open until I find the time to do that.

justvanrossum:
Or: do the merge and make an issue to improve the `BaseGlyphV1List` API, with a reminder to remove the workaround?

anthrotype:
> do the merge and make an issue to improve the BaseGlyphV1List API, with a reminder to remove the workaround?

sounds good

PR #2170, opened 2021-02-03
[feaLib] Allow deleting glyphs

simoncozens:
Support "sub A by NULL;" as described in https://github.com/adobe-type-tools/afdko/issues/1234 and implemented in AFDKO as https://github.com/adobe-type-tools/afdko/pull/1251

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -188,6 +188,21 @@ class Comment(Element):
         return self.text
 
 
+class NullGlyph(Expression):
+    """The NULL glyph, used in glyph deletion substitutions."""
+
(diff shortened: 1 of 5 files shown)


anthrotype (approved):
LGTM, thanks Simon!

anthrotype:
(btw, that `parse_substitute_` method is 166 lines long.. one day maybe we should split it up into more manageable chunks)

ctrlcctrlv:
# :eyes: 

dude…you can delete glyphs in OT Layout?

cool…wow. This makes a lot of hacks I've considered but then gave up on possible.

justvanrossum:
See also https://github.com/MicrosoftDocs/typography-issues/issues/673

PR #2171, opened 2021-02-03
COLRv1: add functions to un-build COLR otTables to raw dicts

anthrotype:
This adds an `colorLib.unbuilder.unbuildColrV1` which does the inverse of `colorLib.builder.buildColrV1`.
That is, it takes a `LayerV1List` and `BaseGlypV1List` and returns a map of base glyphs to raw data structures (list, dict, float, str, etc.).
Useful not only for debugging purpose, but also for implementing COLRv1 subsetting (where we need to drop whole chunks of paints which may be reused by multiple glyphs..).


rsheeter:
Thinking "aloud" this gets me to wondering whether we can have a model that avoids hand-coding buildX and unbuildX. Do we know enough from otData to write a generic implementation? - for example, we hand-code things like:

```python
    def buildPaintRotate(
        self,
        paint: _PaintInput,
        angle: _ScalarInput,
        centerX: _ScalarInput,
        centerY: _ScalarInput,
    ) -> ot.Paint:
        ot_paint = ot.Paint()
        ot_paint.Format = int(ot.Paint.Format.PaintRotate)
        ot_paint.Paint = self.buildPaint(paint)
        ot_paint.angle = _to_variable_f16dot16_float(angle)
        ot_paint.centerX = _to_variable_f16dot16_float(centerX)
        ot_paint.centerY = _to_variable_f16dot16_float(centerY)
        return ot_paint
```

Isn't how to do most of this is clear from otData? Can we avoid much of the hand-coding?

rsheeter:
Generalizing could be a separate effort though I fear it's exactly the type of thing it's easy to never actually do. Still, I don't think it makes sense to hold up this PR on the idea.

rsheeter (approved):

anthrotype:
> Do we know enough from otData to write a generic implementation?

yes and no. We do know about the struct field names and the types, however sometimes we diverge: e.g. we fold the ColorIndex (with its paletteIndex and alpha) into the parent PaintSolid or ColorStop, rather than having it nested; or for the linear/radial gradient we group the x/y coordinates into point tuples. But yeah, I see your point. 🤔

anthrotype:
the generalizing idea requires more thoughs. I'll merge this as is for now

rsheeter:
> sometimes we diverge

Is it possible we shouldn't?

PR #2175, opened 2021-02-05
COLRv1: fix lazy loading by moving Paint.Format enum outside of Paint class nested scope

anthrotype:
When a `TTFont` is loaded with `lazy=True`, the otTables are only decompiled upon request by `BaseTable.__getattr__` triggered when the requested attribute is not found in the instance's `__dict__`.

Since the `Paint.Format` enum class was defined inside the `Paint` class, every `Paint` instance, even when loaded lazily, would have a 'Format' attribute and the magic decompile-on-missing-attribute would not trigger, because the class attribute was being returned and thus `__getattr__` would not be executed, and the lazy Paint instance was not correctly decompiled on demand.

For this reason (and to not add further special cases), it's better to simply move this `Paint.Format` enum class outside the `Paint` class to the module level scope, and rename it `PaintFormat`.

### DIFF
--- a/Lib/fontTools/colorLib/builder.py
+++ b/Lib/fontTools/colorLib/builder.py
@@ -492,7 +492,7 @@ class LayerV1ListBuilder:
         self, paletteIndex: int, alpha: _ScalarInput = _DEFAULT_ALPHA
     ) -> ot.Paint:
         ot_paint = ot.Paint()
-        ot_paint.Format = int(ot.Paint.Format.PaintSolid)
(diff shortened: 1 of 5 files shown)


anthrotype:
To reproduce the problem, just open a `TTFont("COLRv1.ttf", lazy=True)`, get one of the Paints, and as first thing get `paint.Format`. The class attribute (enum class) is returned, but we actually want the instance's "Format" attribute of the same name (the IntEnum value).

rsheeter (approved):
Good catch, that seems like a nasty gotcha

PR #2182, opened 2021-02-10
Update otData for static vs variable formats of COLR v1

rsheeter:
See https://github.com/googlefonts/colr-gradients-spec/issues/233 and associated https://github.com/googlefonts/colr-gradients-spec/pull/249.

This extends #2181; until that lands please prefer to compare against that over master (https://github.com/fonttools/fonttools/compare/colr_to_from_dicts...colrv1_static)


Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -59,14 +59,20 @@ def buildConverters(tableSpec, tableNamespace):
...
 			tableClass = tableNamespace.get(tp)
 		else:
 			tableClass = tableNamespace.get(tableName)
-		if tableClass is not None:
-			conv = converterClass(name, repeat, aux, tableClass=tableClass)
-		else:
-			conv = converterClass(name, repeat, aux)
+
+		if not conv.tableClass:
+			conv.tableClass = tableClass
```
anthrotype, line 74: this `_buildConverters` method is one of the most obscure pieces of code in the whole library, so I'm always weary when we make any changes. Let me see I understand what you're trying to do here.
I see that you want to keep the tableClass from the "template" converters which set it in the converter constructor via partial(). E.g. a converter typed "LOffset24(VarColorLine)" assigned to a table field named "ColorLine" is a problem because "ColorLine" is also the name of a tableClass, and if you weren't doing that, the non-var ColorLine table class would be returned by `tableNamespace.get(tableName)`. Right?
The optional-ness of `tableClass` (the fact that it is allowed to be None, with the use of dict.get()) always makes me nervous that an unexpected "TypeError: NoneType.." is down the line. I think all Struct coverters really have to have a not-None tableClass at some point to be functional (ideally already after their default init constructor).
Also I have no idea why "MortChain" & co. are special-cased here (but that's a separate issue).
I just want to make sure that we are setting a not-None tableClass for the converters that require one, and that we are not inadvertently setting a not-None tableClass for the converters that do _not_ need/use one (like simple values, enums or namedtuples). Whatever makes this code clearer is a bonus.

anthrotype (approved):
It must not have been fun to update all those tests (especially the binary data ones), so thanks for sparing me the chore! 😅 

This LGTM overall, I only have one comment about the change to otConverters.py

Lib/fontTools/ttLib/tables/otConverters.py
rsheeter: My theory is that:

1. The only way that `conv = converterClass(name, repeat, aux)` *has* a tableClass is from partial
1. partial is only used for "template" types in the `converterMapping` in `otConverters.py`
1. If a "template" type exists that's the highest priority: the user explicitly told us what type to use! 
   * Its incredibly weird and probably not intentional that `OffsetTo(Type) not-a-type-name` gives a different type than `OffsetTo(Type) SomeOtherTypeName`
1. It's safe to read `.tableClass` (vs get/has attr) because all converters are `BaseConverter` derivatives and tableClass is declared there

Does that seem valid?

anthrotype: sgtm

anthrotype (approved):

PR #2185, opened 2021-02-11
fontBuilder improvements

simoncozens:
This adds a few improvements to fontBuilder:

* Adds documentation to the setupFvar method
* Allows setupFvar to optionally take designspaceLib.AxisDescriptor objects
* Adds a new setupAvar method
* Supports localised names for axes and named instances

### DIFF
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -943,7 +943,6 @@ def addFvar(font, axes, instances):
         inst = NamedInstance()
         inst.subfamilyNameID = nameTable.addMultilingualName(name, ttFont=font)
         if psname is not None:
-            psname = psname
             inst.postscriptNameID = nameTable.addName(psname)
         inst.coordinates = coordinates
         fvar.instances.append(inst)
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -634,8 +634,9 @@ class FontBuilder(object):
             axes (list): See below.
             instances (list): See below.
 
-        The axes should be a list of iterables; each axis should be supplied
-        in the format ```tupletag, minValue, defaultValue, maxValue, name``.
+        ``axes`` should be a list as axes, with each axis either supplied as
+        a py:class:`.designspaceLib.AxisDescriptor` object, or a tuple in the
+        format ```tupletag, minValue, defaultValue, maxValue, name``.
         The ``name`` is either a string, or a dict, mapping language codes
         to strings, to allow localized name table entries.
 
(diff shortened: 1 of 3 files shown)


Lib/fontTools/fontBuilder.py
```diff
@@ -838,7 +854,10 @@ def addFvar(font, axes, instances):
         axis = Axis()
         axis.axisTag = Tag(tag)
         axis.minValue, axis.defaultValue, axis.maxValue = minValue, defaultValue, maxValue
-        axis.axisNameID = nameTable.addName(tounicode(name))
+        if isinstance(name, basestring):
```
justvanrossum, line 857: Probably best to use `str`.

Ideally, the py23 import should be eliminated completely (ugh, there are two of them...). I'm fairly sure the `Tag` cast is as superfluous as `tounicode`.

simoncozens:
Actually I'm wondering whether to rework this so that fontBuilder takes AxisDescriptor objects for `addFvar`. That would look neater and make adding avar support much easier.

justvanrossum:
> Actually I'm wondering whether to rework this so that fontBuilder takes AxisDescriptor objects for addFvar.

If, then optionally as a convenience. Tying this in to designspaceLib is an unnecessary coupling.

> That would look neater

How would that look neater?

simoncozens:
> If, then optionally as a convenience. Tying this in to designspaceLib is an unnecessary coupling.

Thing is, `AxisDescriptor` isn't really `.designspace`-specific. It's just an object to represent an axis definition, which is exactly what we want in this situation. Here's my thinking:

* I need to add `addAvar` to fontMake.
* fontTools already has a function to add an avar table to a font already, so probably better to reuse that than rewrite it.
* `varLib._add_avar` takes `AxisDescriptor` objects. That makes sense; `AxisDescriptor` is how fontTools represents axes. So if I use varLibs add-an-avar-table implementation, we're already coupled to designspaceLib. (But this is all part of fontTools, so big deal, right?)
* If we're going to do have to create `AxisDescriptor` objects *anyway*, we should support using`AxisDescriptor`s rather than random tuples in the `addFvar` function.

anthrotype:
> I need to add addAvar to fontMake.

you mean fontBuilder?

simoncozens:
Right.

simoncozens:
Of course I hadn't reckoned on the fact that fontTools already has two different classes to represent the concept of an axis, and they're not attribute-compatible. Aaaaargh.

justvanrossum:
> It's just an object to represent an axis definition, which is exactly what we want in this situation. Here's my thinking:

Got it, sounds reasonable upon further reflection. And may allow for more code sharing between varLib and fontBuilder, so that's good, too.

Lib/fontTools/fontBuilder.py
```diff
@@ -826,32 +893,59 @@ def buildCmapSubTable(cmapping, format, platformID, platEncID):
...
+
+        if isinstance(name, str):
+            name = dict(en=name)
 
         inst = NamedInstance()
-        inst.subfamilyNameID = nameTable.addName(name)
+        inst.subfamilyNameID = nameTable.addMultilingualName(name, ttFont=font)
         if psname is not None:
-            psname = tounicode(psname)
+            psname = psname
```
anthrotype, line 946: this line was probably left in involutarily

anthrotype (approved):

anthrotype:
what's the status of this? Is it ready to merge? I'd like to cut a new release later today

simoncozens:
This is far as I'm going to push it in this PR. More changes another day.

anthrotype:
Thanks. Would you mind also writing a couple of short sentences in NEWS.rst about the fontBuilder changes in this PR?

anthrotype:
@simoncozens actually, nevermind, I see you have edited the first PR message (was "WIP" originally) with a detailed list of changes. I can use that one for drafting the changelong, thanks!

PR #2189, opened 2021-02-15
Instancer: update name table (take 2)

anthrotype:
I moved @m4rc1e branch to upstream repo (named `instancer-name`) and moved some files around.

This supersedes https://github.com/fonttools/fonttools/pull/2078


Lib/fontTools/varLib/instancer/names.py
```diff
@@ -83,6 +83,7 @@ def updateNameTable(varfont, axisLimits):
     # AxisValues found in step 1. The updated name table will conform to the
     # RIBBI naming model, more info can be found here:
```
anthrotype, line 84: nit: you could insert that `(Regular, Italic, Bold, Bold Italic)` right after you first mention the acronym

anthrotype: or even better, in the docstring where you talk about the `R/I/B/BI naming model`. docstrings are user visible, more than inline comments which are for devs

anthrotype: actually the whole comment and example is now short enough that it could be moved to the docstring itself

m4rc1e: Thanks!

anthrotype:
@m4rc1e do you this is ready to merge?

m4rc1e:
Just finished testing on Roboto, Cabin and Open Sans. It's behaving as expected but the following could be improved:
- Drop name records which have a nameID 25 if we're instantiating a static font
- What about updating the filename of instantiated fonts?

I'll implement dropping nameID 25 now. For updating the filename, we can tackle this when we set the bits correctly, or if the heuristic is too complex, leave it be.

Lib/fontTools/varLib/instancer/names.py
```diff
@@ -314,6 +314,9 @@ def _updateNameTableStyleRecords(
         assert string, nameID
         nametable.setName(string, nameID, *platform)
 
+    if "fvar" not in varfont:
+        nametable.removeNames(nameID=25)
```
anthrotype, line 318: Don't we have symbolic names for these nameID numbers?

anthrotype:
LGTM (it seems Github doesn't let PR authors approve their own PRs 😅..)

PR #2192, opened 2021-02-18
Add intersections and point-at-time functions to bezierTools

simoncozens:
Adds the following functions to bezierTools:

* quadraticPointAtT
* cubicPointAtT
* linePointAtT

And the generic function which takes a list of tuples and dispatches to the right point-at-time function

* segmentPointAtT

As well as intersection functions:

* lineLineIntersections
* curveLineIntersections
* curveCurveIntersections

And the generic function which takes two lists of tuples and dispatches to the right intersections function:

* segmentSegmentIntersections

Implementations taken from my beziers.py library.

### DIFF
--- a/Lib/fontTools/misc/bezierTools.py
+++ b/Lib/fontTools/misc/bezierTools.py
@@ -879,32 +879,36 @@ def lineLineIntersections(s1, e1, s2, e2):
         >>> (intersection.t1, intersection.t2)
         (0.45069111555824454, 0.5408153767394238)
     """
-    ax, ay = s1
-    bx, by = e1
-    cx, cy = s2
-    dx, dy = e2
-    if math.isclose(cx, dx) and math.isclose(ax, bx):
+    s1x, s1y = s1
+    e1x, e1y = e1
+    s2x, s2y = s2
+    e2x, e2y = e2
+    if (
+        math.isclose(s2x, e2x) and math.isclose(s1x, e1x) and not math.isclose(s1x, s2x)
+    ):  # Parallel vertical
         return []
-    if math.isclose(cy, dy) and math.isclose(ay, by):
+    if (
+        math.isclose(s2y, e2y) and math.isclose(s1y, e1y) and not math.isclose(s1y, s2y)
+    ):  # Parallel horizontal
         return []
-    if math.isclose(cx, dx) and math.isclose(cy, dy):
+    if math.isclose(s2x, e2x) and math.isclose(s2y, e2y):  # Line segment is tiny
         return []
-    if math.isclose(ax, bx) and math.isclose(ay, by):
+    if math.isclose(s1x, e1x) and math.isclose(s1y, e1y):  # Line segment is tiny
         return []
-    if math.isclose(bx, ax):
-        x = ax
-        slope34 = (dy - xy) / (dx - cx)
-        y = slope34 * (x - cx) + cy
+    if math.isclose(e1x, s1x):
+        x = s1x
+        slope34 = (e2y - s2y) / (e2x - s2x)
+        y = slope34 * (x - s2x) + s2y
         pt = (x, y)
         return [
             Intersection(
                 pt=pt, t1=_line_t_of_pt(s1, e1, pt), t2=_line_t_of_pt(s2, e2, pt)
             )
         ]
-    if math.isclose(cx, dx):
-        x = cx
-        slope12 = (by - ay) / (bx - ax)
-        y = slope12 * (x - ax) + ay
+    if math.isclose(s2x, e2x):
+        x = s2x
+        slope12 = (e1y - s1y) / (e1x - s1x)
+        y = slope12 * (x - s1x) + s1y
         pt = (x, y)
         return [
             Intersection(
@@ -912,12 +916,12 @@ def lineLineIntersections(s1, e1, s2, e2):
             )
         ]
 
-    slope12 = (by - ay) / (bx - ax)
-    slope34 = (dy - cy) / (dx - cx)
+    slope12 = (e1y - s1y) / (e1x - s1x)
+    slope34 = (e2y - s2y) / (e2x - s2x)
     if math.isclose(slope12, slope34):
         return []
-    x = (slope12 * ax - ay - slope34 * cx + cy) / (slope12 - slope34)
-    y = slope12 * (x - ax) + ay
+    x = (slope12 * s1x - s1y - slope34 * s2x + s2y) / (slope12 - slope34)
+    y = slope12 * (x - s1x) + s1y
     pt = (x, y)
     if _both_points_are_on_same_side_of_origin(
         pt, e1, s1
@@ -931,6 +935,9 @@ def lineLineIntersections(s1, e1, s2, e2):
 
 
 def _alignment_transformation(segment):
+    # Returns a transformation which aligns a segment horizontally at the
+    # origin. Apply this transformation to curves and root-find to find
+    # intersections with the segment.
     start = segment[0]
     end = segment[-1]
     m = Offset(-start[0], -start[1])
@@ -942,13 +949,11 @@ def _alignment_transformation(segment):
 def _curve_line_intersections_t(curve, line):
     aligned_curve = _alignment_transformation(line).transformPoints(curve)
     if len(curve) == 3:
-        a, b, c = calcCubicParameters(*aligned_curve)
-        intersections = solveQuadratic(a[0], b[0], c[0])
-        intersections.extend(solveQuadratic(a[1], b[1], c[1]))
+        a, b, c = calcQuadraticParameters(*aligned_curve)
+        intersections = solveQuadratic(a[1], b[1], c[1])
     elif len(curve) == 4:
         a, b, c, d = calcCubicParameters(*aligned_curve)
-        intersections = solveCubic(a[0], b[0], c[0], d[0])
-        intersections.extend(solveCubic(a[1], b[1], c[1], d[1]))
+        intersections = solveCubic(a[1], b[1], c[1], d[1])
     else:
         raise ValueError("Unknown curve degree")
     return sorted([i for i in intersections if 0.0 <= i <= 1])
@@ -1000,7 +1005,7 @@ def _split_segment_at_t(c, t):
     if len(c) == 2:
         s, e = c
         midpoint = linePointAtT(s, e, t)
-        return [ ( s,  midpoint), (midpoint, e) ]
+        return [(s, midpoint), (midpoint, e)]
     if len(c) == 3:
         return splitQuadraticAtT(*c, t)
     elif len(c) == 4:
--- a/Lib/fontTools/misc/bezierTools.py
+++ b/Lib/fontTools/misc/bezierTools.py
@@ -2,9 +2,13 @@
 """fontTools.misc.bezierTools.py -- tools for working with Bezier path segments.
 """
 
-from fontTools.misc.arrayTools import calcBounds
+from fontTools.misc.arrayTools import calcBounds, sectRect, rectArea
+from fontTools.misc.transform import Offset, Identity
 from fontTools.misc.py23 import *
 import math
+from collections import namedtuple
+
+Intersection = namedtuple("Intersection", ["pt", "t1", "t2"])
 
 
 __all__ = [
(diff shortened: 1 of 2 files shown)


Lib/fontTools/misc/bezierTools.py
```diff
@@ -753,10 +765,376 @@ def calcCubicPoints(a, b, c, d):
...
+        pt = pointFinder(*curve, t)
+        intersections.append(Intersection(pt=pt, t1=t, t2=_line_t_of_pt(*line, pt)))
+    return intersections
+
+
+def _curve_bounds(c):
+    if len(c) == 3:
+        return calcQuadraticBounds(*c)
+    elif len(c) == 4:
+        return calcCubicBounds(*c)
```
anthrotype, line 995: else raise AssertionError or ValueError maybe instead of returning None; you don't seem to be handline the None case below

Lib/fontTools/misc/bezierTools.py
```diff
@@ -753,10 +765,376 @@ def calcCubicPoints(a, b, c, d):
...
+
+
+def _split_curve_at_t(c, t):
+    if len(c) == 3:
+        return splitQuadraticAtT(*c, t)
+    elif len(c) == 4:
+        return splitCubicAtT(*c, t)
```
anthrotype, line 1006: same here, add an else raise AssertionError(len(c)) instead of letting the None slip out of the if statement

Lib/fontTools/misc/bezierTools.py
```diff
@@ -753,10 +765,376 @@ def calcCubicPoints(a, b, c, d):
...
+    )
+    found.extend(
+        _curve_curve_intersections_t(
+            c12, c22, precision, range1=c12_range, range2=c22_range
+        )
+    )
+    unique_key = lambda ts: int(ts[0] / precision)
+    seen = set()
+    return [
+        seen.add(unique_key(ts)) or ts for ts in found if unique_key(ts) not in seen
```
anthrotype, line 1064: what about this `or` in here? set.add() will always return None.. Are you doing this to be able to update the set while also doing the list comprehension? Looks a bit hacky to me, i'd prefer a more explicit for loop.

Lib/fontTools/misc/bezierTools.py
```diff
@@ -753,10 +765,376 @@ def calcCubicPoints(a, b, c, d):
...
+        a, b, c = calcCubicParameters(*aligned_curve)
+        intersections = solveQuadratic(a[0], b[0], c[0])
+        intersections.extend(solveQuadratic(a[1], b[1], c[1]))
+    elif len(curve) == 4:
+        a, b, c, d = calcCubicParameters(*aligned_curve)
+        intersections = solveCubic(a[0], b[0], c[0], d[0])
+        intersections.extend(solveCubic(a[1], b[1], c[1], d[1]))
+    else:
+        raise ValueError("Unknown curve degree")
+    return sorted([i for i in intersections if (0.0 <= i <= 1)])
```
anthrotype, line 954:
```suggestion
    return sorted(i for i in intersections if 0.0 <= i <= 1)
```

anthrotype:
Thanks Simon for working on this! Pardon that isolated comment, I was doing a line-by-line review but that one was posted before the rest. I'll continue with the review now

anthrotype (commented):
This is very useful and I think it's good to have it in fonttools. Are you planning to add some tests as well? That'd be nice.

simoncozens:
> This is very useful and I think it's good to have it in fonttools. Are you planning to add some tests as well? That'd be nice.

This module seems to use doctests, so that is what I have included.

Lib/fontTools/misc/bezierTools.py
```diff
@@ -753,10 +765,380 @@ def calcCubicPoints(a, b, c, d):
...
+    px, py = pt
+    if not math.isclose(sx, ex):
+        return (px - sx) / (ex - sx)
+    if not math.isclose(sy, ey):
+        return (py - sy) / (ey - sy)
+    # Line is a point!
+    return -1
+
+
+def _both_points_are_on_same_side_of_origin(a, b, c):
```
anthrotype, line 853: the name of this function or its parameters aren't clear to me.. which one of a, b or c is the "origin"? Would it be "c" I guess? Maybe name it accordingly

anthrotype:
yeah, but the doctests are more as examples and probably won't cover all the edge cases or error handling etc. Since curve-curve intersection is especially tricky, it may be worth having some explicit pytest-style tests

Lib/fontTools/misc/bezierTools.py
```diff
@@ -753,10 +765,389 @@ def calcCubicPoints(a, b, c, d):
...
+    ) and _both_points_are_on_same_side_of_origin(pt, s2, e2):
+        return [
+            Intersection(
+                pt=pt, t1=_line_t_of_pt(s1, e1, pt), t2=_line_t_of_pt(s2, e2, pt)
+            )
+        ]
+    return []
+
+
+def _alignment_transformation(segment):
```
anthrotype, line 933: maybe make explicit (in the name of the function and/or in comments) that this is returning a transformation to translate/rotate the curve and the intersecting line such that the latter becomes colinear with the horizontal x axis

Lib/fontTools/misc/bezierTools.py
```diff
@@ -753,10 +765,389 @@ def calcCubicPoints(a, b, c, d):
...
+    angle = math.atan2(endpt[1], endpt[0])
+    return m.reverseTransform(Identity.rotate(-angle))
+
+
+def _curve_line_intersections_t(curve, line):
+    aligned_curve = _alignment_transformation(line).transformPoints(curve)
+    if len(curve) == 3:
+        a, b, c = calcCubicParameters(*aligned_curve)
+        intersections = solveQuadratic(a[0], b[0], c[0])
+        intersections.extend(solveQuadratic(a[1], b[1], c[1]))
```
anthrotype, line 947: Hm, are you sure it's correct to solve for _both_ x=0 and y=0 like you seem to be doing here? Should you not only solve for y=0, since with your transformation you have rotated the curve and the intersecting line to coincide with the horizontal x axis, thus you want to know the t values for which y=0?

See the similar function inside bezierTools called `splitCubic`: this does basically the same thing as yours, minus the previous rotation. If you do your transform, then call splitCubic with `where` parameter set to 0 and `isHorizontal` set to True, meaning `where` is interpreted as a Y coordinate, then you see it's only calling `solveCubic` once (using the y cubic parameters), not twice.

Lib/fontTools/misc/bezierTools.py
```diff
@@ -753,10 +765,389 @@ def calcCubicPoints(a, b, c, d):
...
+        >>> intersection = a[0]
+        >>> intersection.pt
+        (374.44882952482897, 313.73458370177315)
+        >>> (intersection.t1, intersection.t2)
+        (0.45069111555824454, 0.5408153767394238)
+    """
+    ax, ay = s1
+    bx, by = e1
+    cx, cy = s2
+    dx, dy = e2
```
anthrotype, line 885: it's a bit hard to follow along with this cryptic variable names. I get that "s" means start and "e" means end, but a, b, c, d don't mean much. Maybe rename to `s1x, s1y`, `e1x, e1y` etc. or make a Point named tuple with x and y parameters so you can do `s1.x`, `s1.y` etc.

Lib/fontTools/misc/bezierTools.py
```diff
@@ -753,10 +765,389 @@ def calcCubicPoints(a, b, c, d):
...
+    if math.isclose(cx, dx) and math.isclose(ax, bx):
+        return []
+    if math.isclose(cy, dy) and math.isclose(ay, by):
+        return []
+    if math.isclose(cx, dx) and math.isclose(cy, dy):
+        return []
+    if math.isclose(ax, bx) and math.isclose(ay, by):
+        return []
```
anthrotype, line 893: maybe for these special cases where you return [], you could add comments saying the line segments are parallel or are the same, etc.

simoncozens:
> yeah, but the doctests are more as examples and probably won't cover all the edge cases or error handling etc. Since curve-curve intersection is especially tricky, it may be worth having some explicit pytest-style tests

I agree that a comprehensive test suite would definitely be a good idea. I also have to think of the best use of my time... :-/

Lib/fontTools/misc/bezierTools.py
```diff
@@ -753,10 +765,394 @@ def calcCubicPoints(a, b, c, d):
...
+def _alignment_transformation(segment):
+    # Returns a transformation which aligns a segment horizontally at the
+    # origin. Apply this transformation to curves and root-find to find
+    # intersections with the segment.
+    start = segment[0]
+    end = segment[-1]
+    m = Offset(-start[0], -start[1])
+    endpt = m.transformPoint(end)
+    angle = math.atan2(endpt[1], endpt[0])
+    return m.reverseTransform(Identity.rotate(-angle))
```
justvanrossum, line 946: `return Identity.rotate(-angle).translate(-start[0], -start[1])` should do the same, slightly more compactly.

Lib/fontTools/misc/bezierTools.py
```diff
@@ -753,10 +765,394 @@ def calcCubicPoints(a, b, c, d):
...
+            c11, c22, precision, range1=c11_range, range2=c22_range
+        )
+    )
+    found.extend(
+        _curve_curve_intersections_t(
+            c12, c22, precision, range1=c12_range, range2=c22_range
+        )
+    )
+
+    unique_key = lambda ts: int(ts[0] / precision)
```
justvanrossum, line 1069: Edge case maybe? If the second curve loops, and the first curve goes exactly through the looping intersection point, does this still do the right thing? There would be two distinct ts[1] values for one ts[0].

Lib/fontTools/misc/bezierTools.py
```diff
@@ -753,10 +765,394 @@ def calcCubicPoints(a, b, c, d):
...
+    seen = set()
+    unique_values = []
+
+    for ts in found:
+        if unique_key(ts) in seen:
+            continue
+        seen.add(unique_key(ts))
```
justvanrossum, line 1076: Avoid calling `unique_key(ts)` twice.

justvanrossum:
These are very useful additions! Thanks.

(I'm curious how it behaves with "difficult" curve-curve intersections, for example intersecting at a shallow angle: these have a tendency to degenerate)

anthrotype:
btw, the tests didn't trigger because we had renamed master branch to main, and workflow was set up to only run tests for master branch or pr against that https://github.com/fonttools/fonttools/pull/2194, if you merge main into your pr branch then they should trigger

anthrotype:
> I agree that a comprehensive test suite would definitely be a good idea. I also have to think of the best use of my time... :-/

I totally understand. I'd be ok to merge this as is and refine/add tests at a later stage.
I see you are testing this code, even though indirectly, in https://github.com/googlefonts/glyphsLib/pull/663

If you could address the questions/comments that @justvanrossum posed earlier, and he also agrees to, then we can proceed with the merge.

Lib/fontTools/misc/bezierTools.py
simoncozens: I'm not sure about this. To work out the angle, you need to first transform the end point relative to the start point. So *something like* the first three lines are needed anyway.

justvanrossum: I may have messed up, let me try again.

Lib/fontTools/misc/bezierTools.py
simoncozens: I'll fix this, but that's the edgiest of edge cases and if a curve goes precisely through the intersection of loop I think there should be some debate as to whether that should actually count as one or intersections...

Lib/fontTools/misc/bezierTools.py
justvanrossum: You're right, my suggestion wasn't complete. The following `v2` is equivalent and clearer in its intention:
```python
import math
from fontTools.misc.transform import Offset, Identity

def v1(segment):
    start = segment[0]
    end = segment[-1]
    m = Offset(-start[0], -start[1])
    endpt = m.transformPoint(end)
    angle = math.atan2(endpt[1], endpt[0])
    return m.reverseTransform(Identity.rotate(-angle))

def v2(segment):
    start = segment[0]
    end = segment[-1]
    angle = math.atan2(end[1] - start[1], end[0] - start[0])
    return Identity.rotate(-angle).translate(-start[0], -start[1])


segment = [(100, 200), (300, 410)]
print(v1(segment))
print(v2(segment))
```

anthrotype: I also think Just approach looks clearer, you mind updating it?

anthrotype (approved):

simoncozens:
I'll just wait for an OK from Just.

justvanrossum (approved):

PR #2199, opened 2021-02-26
Revert "Merge pull request #2164 from fonttools/pen-maintenance"

### LINKED ISSUE #2198: Breaking change with AbstractPen
I understand why and I probably need to embrace the typing future

but :)

This https://github.com/fonttools/fonttools/commit/f6a62e0db127c01f4d23f3398d70df21bbc6dccc PR is a breaking change.

```python
from fontTools.pens.basePen import AbstractPen

class D(AbstractPen):
    
    def moveTo(self, pt):
        print(pt)
    def lineTo(self, pt):
        print(pt)
    def curveToOne(self, *pts):
        print(pts)
    def closePath(self):
        print("close")
        
d = D()
```

results in:

```console
  File "<untitled>", line 15, in <module>
TypeError: Can't instantiate abstract class D with abstract methods addComponent, curveTo, endPath, qCurveTo
```

Maybe this had to be discussed more to envision the impact of such a change. Pens and pointPens are the core of lots of tools. 

For example a pen only dealing with components didnt need the requirement to implement all the extra `NotImplemented callbacks (like moveTo, lineTo, ... ) 

I dont know enough about `abc` to make it less strict  :) or maybe optionally at init of such a class to not raise an error. 

again, I dont have an issue with changing my code, but getting a breaking change is huge

anthrotype:
This reverts commit f6a62e0db127c01f4d23f3398d70df21bbc6dccc, reversing
changes made to 4fb22eed1b6149cbcf681dfaad9fe9f887d04e03.

Fixes #2198


justvanrossum:
Most of the work @madig did was good and useful, it seems sad to revert all of it. Perhaps we should only revert the ABC changes. Unless you think it's better to first revert all, and then selectively redo the rest?

justvanrossum:
Yeah, let me quickly make an alternate PR that only reverts the ABC changes.

anthrotype:
@justvanrossum that'd be nice, thank you

PR #2201, opened 2021-02-27
[misc.arrayTools] move Vector to its own submodule, and rewrite as a tuple subclass

justvanrossum:
I felt our n-dimensional Vector class could be improved by making it a subclass of tuple, and deserves its own module.

This change makes Vector instances immutable. It appears this was the intention, as it did support `__getitem__` but not `__setitem__`.

The internal `values` attribute is gone. I added a `values` read-only property for backward compatibility. Setting `.values` is no longer supported (due to Vector's new immutablity).

I added a two methods:
- `vector.length()`: an alias for `abs(vector)`
- `vector.normalized()`: returns the norm vector

I added some DeprecationWarnings.

### DIFF
--- a/Lib/fontTools/misc/arrayTools.py
+++ b/Lib/fontTools/misc/arrayTools.py
@@ -4,9 +4,10 @@ so on.
 
 from fontTools.misc.py23 import *
 from fontTools.misc.fixedTools import otRound
-from numbers import Number
+from fontTools.misc.vector import Vector as _Vector
 import math
-import operator
+import warnings
+
 
 def calcBounds(array):
     """Calculate the bounding rectangle of a 2D points array.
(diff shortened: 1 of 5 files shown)


anthrotype (approved):
Thanks! 

When I implement Vector class, I like to call what you named`length()` => `norm()` and what yuo named `normalized()` => `unit()`.

But it probably doesn't really matter.

I wonder about the use of `abs()` as a shortcut to computing the norm/magnitude/length/whathaveyou of the vector. Is that common? I guess it can be handy, but also may surprise a bit (it did surprise me).

justvanrossum:
> When I implement Vector class, I like to call what you named length() => norm()

It seems that "length", "norm" and "magnitude" are all synonyms here. I prefer "length" as it's a word that anyone understands. I just had to google "norm", as I wasn't even sure...

> and what you named normalized() => unit().

Right, that makes sense, too. I prefer "normalized" as it implies an action.

> I wonder about the use of abs() as a shortcut to computing the norm/magnitude/length/whathaveyou of the vector. Is that common? I guess it can be handy, but also may surprise a bit (it did surprise me).

It surprised me, too, but I took it from the original Vector, so it better stay for b/w compatibility... It makes sense given the `|vec|` mathematical notation, but this was the first time I saw it used in Python via `__abs__`.

anthrotype:
ok, sgtm! thanks

justvanrossum:
Now it would be nice to use it more in fonttools itself. Many things in `bezierTools` could benefit from it, like `calcCubicParameters()`.

(Back in the day fonttools used numpy to make these kinds of operations easier to write. I'm very glad Behdad got rid of that dependency, but I've occationally missed the convenience.)

PR #2202, opened 2021-02-28
Experiment with hypothesis

madig:
Wherein I use [Hypothesis](https://hypothesis.readthedocs.io) to add property based testing to the suite, a.k.a blasting your API with random data and checking if invariants hold. Think of a monkey throwing shit at your API to see what sticks.

As a first step, this PR utilizes stateful testing to give pens a good thrashing to shake out unhandled crashers. I know that there are invariants to consider while using the API, and I sketched out a more intelligent test at https://gist.github.com/madig/08d3d07aad47e45e96ba3d4fac23ded4. I however need something better than a recording pen to compare against, because some pens do some processing like removing empty paths and turning single points into a `move` with `smooth=False` or buffer paths and retire them after endPath, so a dumb recording pen ends up with a different command stream if you addComponent inbetween.

As an initial finding, there's some inconsistency in how pens handle identifier and kwarg args.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,5 @@ Lib/**/*.c
 
 # Ctags
 tags
+
+.hypothesis
--- a/Lib/fontTools/pens/basePen.py
+++ b/Lib/fontTools/pens/basePen.py
@@ -40,8 +40,12 @@ from typing import Any, Tuple
 
 from fontTools.misc.loggingTools import LogMixin
 
-__all__ =  ["AbstractPen", "NullPen", "BasePen",
+__all__ =  ["AbstractPen", "NullPen", "BasePen", "PenError",
 			"decomposeSuperBezierSegment", "decomposeQuadraticSegment"]
+
+
+class PenError(Exception):
+	"""Represents an error during penning."""
 
 
 class AbstractPen:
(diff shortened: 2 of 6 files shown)


behdad:
Thanks @madig. Such fuzzing is something I've always want to set up in our various projects. For HarfBuzz I've wanted to do that as well. One library I think might be relevant is https://github.com/silentbicycle/theft

madig:
Yeah, I too am intrigued by fuzzing. So far, I have found that running a fuzz test takes some time, unlike your typical unit tests :thinking: not sure how to proceed here. The tests aren't the most interesting, they just look for crashes or roundtrips preserving data. There's a more interesting test at https://github.com/madig/HTLetterspacer/blob/reduce-better/tests/test_segmentation.py, where I test a custom implementation of `Contour.segments`. Basically, what we need to make effective use of fuzzing is properties we can test :)

madig:
Here's an article on using Hypothesis: https://increment.com/testing/in-praise-of-property-based-testing/

madig:
Here's an issue stumbling over an AssertionError: https://github.com/googlefonts/fontmake/issues/764

The error message could probably be improved by catching and bubbling up the PenError I made in this PR.

madig:
Added rough sketch for testing TTglyphPen and TTGlyphPointPen for equal behavior. Getting some failures, but need to wrap my brain around whether those are actual failures or me holding it wrong.

PR #2205, opened 2021-03-02
Add TTGlyphPointPen

jenskutilek:
See https://github.com/googlefonts/ufo2ft/issues/469: The TTGlyphPen may change the point indices. This adds a PointPen-based TTGlyphPointPen to avoid changes in the outline.

I'm not sure what to put in the tests; how strict must path handling be? E.g. What happens if `beginPath()` is called twice in a row, or `addPoint()` is not called on a path; or should the base PointPen take care of those edge cases?


jenskutilek:
I added some assertions to the pen so that beginPath can not be called twice, etc. which is more strict than the normal PointPen. I also adapted the tests from TTGlyphPen.

justvanrossum:
Quick comments:

- consider adding it to the ttGlyphPen module (the recording pens also live in the same place)
- can both tt pens be refactored to share most code?
- could the segment pen be implemented in terms of the point pen?

Tests/pens/ttGlyphPointPen_test.py
```diff
@@ -0,0 +1,324 @@
...
+import unittest
+import struct
+
+from fontTools import ttLib
+from fontTools.misc.testTools import TestCase
+from fontTools.pens.ttGlyphPointPen import TTGlyphPointPen, MAX_F2DOT14
+from fontTools.ttLib.tables import ttProgram
+
+
+class TTGlyphPointPenTest(TestCase):
```
anthrotype, line 11: for new test modules we should try to use the pytest API rather then unittest. Among the nice things it provides an easy way to parametrize repetitive tests. You probably copied form the existing ttGlyphPen_test.py which still uses unittest.

anthrotype (commented):
like Just said, we should avoid duplication between the TTGlyphPen and this. Please move to the same module and refactor to reuse as much as you can.

Maybe try also reimplementing the segment-oriented pen in terms of the point pen like Just suggested earlier, e.g. by having TTGlyphPen use the SegmentToPointPen internally delegating to an instance of TTGlyphPointPen

justvanrossum:
Or perhaps they can have a common base class.

madig:
Maybe I can fuzz this in #2202 🤔 If you want to be anal about error handling (I would), check the changes I did to pens in that PR.

Tests/pens/ttGlyphPointPen_test.py
jenskutilek: Thanks, I changed the tests for both TTGlyphPen and TTGlyphPointPen to pytest.

jenskutilek:
I moved the `TTGlyphPointPen` class into pens.ttGlyphPen, and also implemented both pens with a common base class, `TTGlyphBasePen`.

I also tried to subclass `TTGlyphPen` from `SegmentToPointPen` and use `TTGlyphPointPen` in there, but it turned out it gets more complex, so I kept the common subclass approach.

jenskutilek:
> Maybe I can fuzz this in #2202 🤔 If you want to be anal about error handling (I would), check the changes I did to pens in that PR.

Looks useful, I could add the improved error handling after your PR is merged.

Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -47,36 +55,127 @@ def __init__(self, glyphSet, handleOverflowingTransforms=True):
         self.handleOverflowingTransforms = handleOverflowingTransforms
         self.init()
 
-    def init(self):
+    def init(self) -> None:
         self.points = []
         self.endPts = []
         self.types = []
         self.components = []
+        self.currentPath = None
```
anthrotype, line 63: self.currentPath is not used by TTGlyphBasePen nor by TTGlyphPen, but only by TTGlyphPointPen. It would make more sense to have it defined only in the latter's `__init__` method, after the `super().__init__()` has been called.

Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -103,72 +202,79 @@ def closePath(self):
...
 
-        return glyph
+        self.currentPath.append(pt)
+        self.points.append(pt)
+
+    def glyph(self, componentFlags: int = 0x4) -> Glyph:
+        """
+        Returns a :py:class:`~._g_l_y_f.Glyph` object representing the glyph.
+        """
+        assert self.currentPath is None
```
anthrotype, line 279: `assert self.currentPath is None` could be the content of `TTGlyphPointPen._isClosed`, the point-pen version of `_isClosed` method of `TTGlyphPen`. The base class could define an "abstract" `_isClosed` that raises NotImplementedError, so the subclasses must override it with specific implementations. Then you would no longer need to have separate `_glyph` and `glyph` methods, with the latter only difference in the way to check that the countor is closed.

Tests/pens/ttGlyphPointPen_test.py
```diff
@@ -1,24 +1,25 @@
...
 from fontTools import ttLib
-from fontTools.misc.testTools import TestCase
-from fontTools.pens.ttGlyphPointPen import TTGlyphPointPen, MAX_F2DOT14
-from fontTools.ttLib.tables import ttProgram
+from fontTools.pens.ttGlyphPen import TTGlyphPointPen, MAX_F2DOT14
 
 
-class TTGlyphPointPenTest(TestCase):
-
+class TTGlyphPointPenTest:
```
anthrotype, line 9: you should move this test class in the same test module as ttGlyphPen_test.py, since we decided to also move the class itself to the same Lib module. The test directory has the same layout as the Lib one.

Tests/pens/ttGlyphPointPen_test.py
```diff
@@ -112,16 +112,16 @@ def test_glyph_decomposes(self):
         pen.endPath()
         plainGlyph = pen.glyph()
 
-        self.assertEqual(plainGlyph, compositeGlyph)
+        assert plainGlyph == compositeGlyph
 
     # def test_remove_extra_move_points(self):
```
anthrotype, line 117: why are these left in but commented out?

anthrotype (commented):
thanks Jens, I left some more comments.

Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -1,30 +1,38 @@
...
 
     This pen can be used to construct or modify glyphs in a TrueType format
     font. After using the pen to draw, use the ``.glyph()`` method to retrieve
     a :py:class:`~._g_l_y_f.Glyph` object representing the glyph.
     """
 
-    def __init__(self, glyphSet, handleOverflowingTransforms=True):
-        """Construct a new pen.
+    def __init__(
+        self, glyphSet: _TTGlyphSet, handleOverflowingTransforms: bool = True
```
justvanrossum, line 29: This can be any kind of glyph set. I don't think it makes sense to use this (private) class as a type declaration.

Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -1,30 +1,38 @@
...
+    ) -> None:
+        """
+        Construct a new pen.
 
         Args:
-            glyphSet (ttLib._TTGlyphSet): A glyphset object, used to resolve components.
+            glyphSet (ttLib.ttFont._TTGlyphSet): A glyphset object, used to resolve components.
```
justvanrossum, line 35: Hm, so _TTGlyphSet was mentioned already. Doesn't make sense to me, though.

Lib/fontTools/pens/ttGlyphPen.py
justvanrossum: About `self.currentPath`: why is it even a list? It's effectively a three-state flag: 
1. no current path (not a valid contour)
2. there is a current path but no points have been added (not a valid contour)
3. there is a current path with at least one point (a valid contour)

Idea: call it `_currentContourStartIndex` or something, set to `None` upon reset, and set to `len(self.points)` when starting a new contour. Then you can compare with `len(self.points)` to see if there are any points.

Lib/fontTools/pens/ttGlyphPen.py
anthrotype: true, maybe leave it as Dict[str, Any] until we define some Drawable protocol that we can annotate this with, like https://github.com/fonttools/ufoLib2/blob/master/src/ufoLib2/typing.py

Tests/pens/ttGlyphPen_test.py
```diff
@@ -290,6 +290,310 @@ def test_scaled_component_bounds(self):
         self.assertGlyphBoundsEqual(uni0302_uni0300, (-278, 745, 148, 1025))
 
 
+class TTGlyphPointPenTest:
+    def runEndToEnd(self, filename):
```
anthrotype, line 294: this `runEndToEnd` is basically identical to the one in `runEndToEnd` (except for the pen class) maybe have a base class that defines that and have both subclass from it, with the pen class defined as a class attribute

Tests/pens/ttGlyphPen_test.py
```diff
@@ -290,6 +290,310 @@ def test_scaled_component_bounds(self):
...
+
+    # def test_keep_duplicate_end_point(self):
+    #     pen = TTGlyphPointPen(None)
+    #     pen.moveTo((0, 0))
+    #     pen.lineTo((100, 0))
+    #     pen.qCurveTo((100, 50), (50, 100), (0, 0))
+    #     pen.lineTo((0, 0))  # the duplicate point is not removed
+    #     pen.closePath()
+    #     assert len(pen.points) == 5
+    #     assert pen.points[0] == (0, 0)
```
anthrotype, line 431: remove commented out code or explain with a comment why it is left in

Tests/pens/ttGlyphPen_test.py
```diff
@@ -290,6 +290,310 @@ def test_scaled_component_bounds(self):
...
+
+            if hasattr(oldGlyph, "program"):
+                newGlyph.program = oldGlyph.program
+
+            assert oldGlyph.compile(glyfTable) == newGlyph.compile(glyfTable)
+
+    def test_e2e_linesAndSimpleComponents(self):
+        self.runEndToEnd("TestTTF-Regular.ttx")
+
+    def test_e2e_curvesAndComponentTransforms(self):
```
anthrotype, line 323: these methods are intentical for TTGlyphPenTest and TTGlyphPointPenTest classes, use the same trick that I suggested for runEndToEnd above, and define them in the base class then subclass from them, the test will be run with both using different pen classes 

Only make sure you don't name the base class with `*Test` suffix otherwise pytest thinks it is a concrete test class (e.g. like `TTGlyphPenTest`). Call it "TTGlyphPenTestBase" maybe

Tests/pens/ttGlyphPen_test.py
```diff
@@ -290,6 +290,310 @@ def test_scaled_component_bounds(self):
...
+            pen.addPoint((0, 0), "curve")
+
+    def test_beginPath_beginPathOnOpenPath(self):
+        pen = TTGlyphPointPen(None)
+        pen.beginPath()
+        pen.addPoint((0, 0))
+        with pytest.raises(AssertionError):
+            pen.beginPath()
+
+    def test_closePath_ignoresAnchors(self):
```
anthrotype, line 351: what do you mean here by "anchors"? I only see a beginPath and an endPath (an empty contour)

Lib/fontTools/pens/ttGlyphPen.py
jenskutilek: As TTGlyphPointPen has no `__init__` method yet, I think leaving an unused instance variable probably doesn't hurt, WDYT?

Lib/fontTools/pens/ttGlyphPen.py
jenskutilek: I changed it.

Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -47,36 +54,136 @@ def __init__(self, glyphSet, handleOverflowingTransforms=True):
...
+        Check if the current path is closed.
+        """
+        raise NotImplementedError
+
+    def init(self) -> None:
         self.points = []
         self.endPts = []
         self.types = []
         self.components = []
+        self._currentContourStartIndex = None
```
justvanrossum, line 68: I currently don't see this used as an index, just as a flag meaning "we're in between beginPath and endPath".

Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -47,36 +54,136 @@ def __init__(self, glyphSet, handleOverflowingTransforms=True):
...
+        components = self._buildComponents(componentFlags)
+
+        glyph = Glyph()
+        glyph.coordinates = GlyphCoordinates(self.points)
+        glyph.coordinates.toInt()
+        glyph.endPtsOfContours = self.endPts
+        glyph.flags = array("B", self.types)
+        self.init()
+
+        if components:
```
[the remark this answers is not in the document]
anthrotype: the current decomposing behaviour it's tested here

https://github.com/fonttools/fonttools/blob/31ab3aae0b3061ea106551adc54c124f8afec465/Tests/pens/ttGlyphPen_test.py#L78-L107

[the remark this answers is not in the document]
jenskutilek: Argh, I saw that part but thought it only decomposed because of the overflowing transformation.

Should a warning be logged when a decompose happens?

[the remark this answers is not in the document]
jenskutilek: Silently decomposing should be fine unless you absolutely want to avoid overlaps.

A theoretical TrueType hinting authoring tool — should it even allow to hint a mixed composite — will notice that the glyph has changed when compiling and hopefully throw an error.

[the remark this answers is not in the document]
anthrotype: if you like you can add that assert and make it even clearer in the docstring that this is the expected behavior, but it's a bit too late to warn in my opinion.

Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -103,72 +215,73 @@ def closePath(self):
...
-                continue
-            if (self.points or
-                    (self.handleOverflowingTransforms and overflowing)):
-                # can't have both coordinates and components, so decompose
-                tpen = TransformPen(self, transformation)
-                self.glyphSet[glyphName].draw(tpen)
-                continue
+class TTGlyphPointPen(TTGlyphBasePen, LogMixin, AbstractPointPen):
+    """
+    Pen used for drawing to a TrueType glyph.
```
anthrotype, line 225: this docstring is identical to TTGlyphBasePen's, either you make it more specific (e.g. say it's a point pen, not just a pen), or if you keep the one from the base pen, you delete this one
```suggestion
    Point pen used for drawing to a TrueType glyph.
```

Tests/pens/ttGlyphPen_test.py
```diff
@@ -290,6 +296,260 @@ def test_scaled_component_bounds(self):
...
+        pen.addPoint((50, 0), "line")
+        pen.addPoint((450, 0), "line")
+        pen.addPoint((450, 700), "line")
+        pen.addPoint((50, 700), "line")
+        pen.endPath()
+        glyph = pen.glyph()
+        assert glyph.numberOfContours == 1
+        assert glyph.endPtsOfContours == [3]
+
+    def test_glyph_errorOnCurve(self):
```
anthrotype, line 315:
```suggestion
    def test_addPoint_errorOnCurve(self):
```

Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -103,72 +215,73 @@ def closePath(self):
...
-            glyph.numberOfContours = -1
+    def endPath(self) -> None:
+        """
+        End the current sub path.
+        """
+        # TrueType contours are always "closed"
+        if self._isClosed():
+            raise PenError("Contour is already closed.")
+        if self._currentContourStartIndex == len(self.points):
+            raise PenError("Tried to end an empty contour.")
```
anthrotype, line 261: I wonder if it's too harsh to error out if the contour is empty. We can safely ignore it I guess, with no loss of information

Lib/fontTools/pens/ttGlyphPen.py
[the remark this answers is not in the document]
jenskutilek: I've added a note to the docstring.

Lib/fontTools/pens/ttGlyphPen.py
jenskutilek: I've added setting `_currentContourStartIndex` to the `init()` method of the point pen, which now in turn calls `super().init()`.

jenskutilek: ... and moved the docstrings around.

Lib/fontTools/pens/ttGlyphPen.py
jenskutilek: I agree beginning and immediately ending a path is "transparent", but then I wonder in which case a script may do it ... better to err on the safe side? Likewise, we could relax the checks about `endPath` being called again before a new `beginPath`, because `beginPath` may assume ending the previous path implicitly ... I think I'd leave it like it is.

Lib/fontTools/pens/ttGlyphPen.py
jenskutilek: I moved it to the subclass.

jenskutilek:
I think this is ready to be merged now. Any more comments @anthrotype @justvanrossum?

justvanrossum (approved):
LGTM, thanks!

anthrotype (approved):
LGTM, thank you Jens!

anthrotype:
I just released https://github.com/fonttools/fonttools/releases/tag/4.24.0 with the new point pen

Issue #2207, opened 2021-03-03
[varLib] varLib.models.VariationModel assumes all locations are full-range


simoncozens:
VariationModel seems to be based on the idea that all "master" locations apply across the full range of the axes:

```
	def _computeMasterSupports(self, axisPoints):
		...
		# Compute min/max across each axis, use it as total range.
```

but it suggests that one day this will change and it will be possible to specify location that only apply to certain areas of the design space:

```
		# TODO Take this as input from outside?
```

It would be helpful, particularly in the case of variable layout, if locations were able to specify their box min/max points explicitly.

This would require some upheaval - a new interface design for `VariationModel.__init__(locations)` at least.

justvanrossum:
I have been wondering whether the behavior could be better even without explicit input.

The top image is the current bahavior with a single intermediate master, but the bottom image would be much better, as the first half of the axis range would then only have one participating delta, and the exact middle position would be the sum of two integers (exact), instead of the sum of two integers plus an interpolated value (needs to be rounded).

![image](https://user-images.githubusercontent.com/4246121/109787894-ff53cb00-7c0e-11eb-8f01-ee06480731f2.png)

justvanrossum:
In my ideal world, on a single axis with several intermediates, there would only ever be two deltas participating at a time: no more than two "tents" ever overlap. Currently, all deltas always participate, so is more work for a renderer.

(I'm not sure of the implictions for multiple axes.)

simoncozens:
I think VariationModel already does what you suggest in the second diagram, thanks to [this code here](https://github.com/fonttools/fonttools/blob/07a5e2de62b5806517accffa088c8dc8ac1ec95a/Lib/fontTools/varLib/models.py#L317-L347). (But I have to admit that I *barely* understand all this, so you may be talking about something else.)

```
>>> v = VariationModel([ { "wght": 0}, { "wght": +0.5 }, { "wght": +1}])
>>> v.supports
[{}, {'wght': (0, 0.5, 1)}, {'wght': (0.5, 1, 1)}]
```

But this is somewhat orthogonal to the need for explicit region support.

justvanrossum:
I barely understand either, and you may be right. And indeed let's not distract from your actual issue. (I'll try to more precisely find what I'm unhappy about and will post a new issue if/when I manage.)

simoncozens:
[...]

PR #2208, opened 2021-03-03
Refactor VariationModel

simoncozens:
See #2207

- Separate computing delta weights
- Split off box-splitting/support code to separate function


simoncozens:
After chatting with Behdad I need to fix some terminology issues with the variable naming.

behdad:
Sure let's merge this. I'll try to reshuffle things around into more useful structure.

behdad:
Ah, thanks for fixing the terminology.

PR #2209, opened 2021-03-03
Add definitions for varLib-specific terms

simoncozens:
varLib uses a lot of technical terms in sometimes unfamiliar ways, which can only be understood correctly if you're Behdad. ;-) After a conversation with him this morning where he explained the system to me, I've tried to document my understanding of what those terms mean.

### DIFF
--- a/Doc/source/varLib/index.rst
+++ b/Doc/source/varLib/index.rst
@@ -1,17 +1,18 @@
-##########################
-OpenType Variation Support
-##########################
+##################################
+varLib: OpenType Variation Support
+##################################
 
-The `fontTools.varLib` package contains a number of classes and routines
+The ``fontTools.varLib`` package contains a number of classes and routines
 for handling, building and interpolating variable font data. These routines
 rely on a common set of concepts, many of which are equivalent to concepts
-in the OpenType Specification, but some of which are unique to `varLib`.
+in the OpenType Specification, but some of which are unique to ``varLib``.
 
 Terminology
 -----------
 
 axis
-   "A designer-determined variable in a font face design that can be used to derive multiple, variant designs within a family." (OpenType Specification)
+   "A designer-determined variable in a font face design that can be used to
+   derive multiple, variant designs within a family." (OpenType Specification)
    An axis has a minimum value, a maximum value and a default value.
 
 designspace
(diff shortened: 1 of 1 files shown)


Doc/source/varLib/index.rst
```diff
@@ -1,6 +1,93 @@
...
+location
+   A point in the designspace, specified as a set of coordinates on one or
+   more axes. In the context of ``varLib``, a location is a dictionary with
+   the keys being the axis tags and the values being the coordinates on the
+   respective axis. A ``varLib`` location dictionary may be "sparse", in the
+   sense that axes defined in the font may be omitted from the location's
+   coordinates, in which case the default value of the axis is assumed.
+   For example, given a font having a ``wght`` axis ranging from 200-1000
+   with default 400, and a ``wdth`` axis ranging 100-300 with default 150,
+   the location ``{"wdth": 200}`` represents the point ``wght=400,wdth=200``.
```
anthrotype, line 39: not sure if relevant for this discussion (probably not), but in the case of the varLib.instancer, location dicts where some axes are omitted don't imply default coordinates, but they identify a partial instance where the omitted axes are not "pinned" to some specific coordinate, but left to vary.

anthrotype (approved):
thanks Simon (and Behdad) for this! It's good to have more explicit definitions for these terms.

Doc/source/varLib/index.rst
```diff
@@ -1,6 +1,93 @@
...
+The `fontTools.varLib` package contains a number of classes and routines
+for handling, building and interpolating variable font data. These routines
+rely on a common set of concepts, many of which are equivalent to concepts
+in the OpenType Specification, but some of which are unique to `varLib`.
+
+Terminology
+-----------
+
+axis
+   "A designer-determined variable in a font face design that can be used to derive multiple, variant designs within a family." (OpenType Specification)
```
simoncozens, line 14: Oops wrapping

Doc/source/varLib/index.rst
```diff
@@ -1,6 +1,94 @@
...
+
+support
+   While designers tend to think in terms of masters - that is, a precise
+   location having a particular value - OpenType Variations specifies the
+   variation of scalars in terms of deltas which are themselves composed of
+   the combined contributions of a set of triangular regions, each having
+   a contribution value of 0 at its minimum value, rising linearly to its
+   full contribution at the *peak* and falling linearly to zero from the
+   peak to the maximum value. The OpenType Specification calls these "regions",
+   while ``varLib`` calls them "supports" (a mathematical term used in real
```
behdad, line 69: Just a note: we also call the objects region. When they are used as the support of a function, we call them support. "region" is the type of the variable; "support" is their functionality. :)

As in, when you write "float area" in C...

Issue #2213, opened 2021-03-04
[varLib.models] Make VariationModel optionally do rounding


behdad:
Forking off from https://github.com/fonttools/fonttools/issues/2207#issuecomment-790199221

This allows for rounding error from one set of deltas to be accounted for when computing subsequent deltas, reducing the total error from accumulating unbounded as number of masters grows, to always being limited to 0.5.

We should be careful to update all relevant callsites and remove the rounding that is done at callsite boundaries, specially whenever `model.getDeltas` is called and results rounded, can be simplified.

PR #2214, opened 2021-03-04
[varLib] Move rounding to VariationModel, to avoid error accumulation from multiple deltas

behdad:
Fixes https://github.com/fonttools/fonttools/issues/2213


Lib/fontTools/misc/fixedTools.py
```diff
@@ -18,14 +18,14 @@
...
 from fontTools.misc.py23 import *
+from .roundTools import otRound
 import math
 import logging
 
 log = logging.getLogger(__name__)
 
 __all__ = [
 	"MAX_F2DOT14",
-	"otRound",
```
justvanrossum, line 28: The original `otRound` func is still present in this module.

Lib/fontTools/misc/roundTools.py
```diff
@@ -0,0 +1,59 @@
+"""
+Various round-to-integer helpers.
+"""
+
+from fontTools.misc.py23 import *
```
justvanrossum, line 5: Would be fantastic if importing py23 could be avoided here.

justvanrossum (approved):
Looks good to me.

Lib/fontTools/varLib/__init__.py
```diff
@@ -262,7 +262,7 @@ def _add_gvar(font, masterModel, master_ttfs, tolerance=0.5, optimize=True):
 		endPts = control.endPts
 
 		for i,(delta,support) in enumerate(zip(deltas[1:], supports[1:])):
-			if all(abs(v) <= tolerance for v in delta.array) and not isComposite:
+			if all(v == 0 for v in delta.array) and not isComposite:
```
anthrotype, line 265: the `tolerance` parameter of `_add_gvar` seems to be no longer used, did you leave it in intentionally (e.g. not to break other code)? In that case, default it to `None` and do `warnings.warn` if set by caller.

Tests/varLib/data/test_results/TestSparseCFF2VF.ttx
```diff
@@ -290,7 +290,7 @@
         </CharString>
         <CharString name="cid06449" fdSelectIndex="1">
           2 vsindex
-          -60 30 203 30 -9 9 67 7 -7 14 -14 30 -20 20 80 30 59 30 121 30 18 93 -30 30 -30 108 -23 0 -26 67 2 76 -98 -2 -111 42 0 47 -13 0 -14 13 0 14 -33 0 -37 11 0 13 -11 0 -13 8 0 9 -7 0 -8 53 0 60 -32 0 -36 32 0 36 -52 0 -59 57 1 65 -33 0 -38 53 0 60 -83 -1 -93 54 0 60 -6 -19 -24 33 19 55 -76 -1 -86 76 1 86 -76 -1 -86 59 1 67 26 blend
+          -60 30 203 30 -9 9 67 7 -7 14 -14 30 -20 20 80 30 59 30 121 30 18 93 -30 30 -30 108 -23 0 -26 67 2 76 -98 -2 -111 42 0 47 -13 0 -14 13 0 14 -33 0 -37 11 0 13 -11 0 -13 8 0 9 -7 -1 -8 53 0 60 -32 0 -36 32 0 36 -52 0 -59 57 1 65 -33 0 -38 53 0 60 -83 -1 -93 54 0 60 -6 -19 -24 33 19 55 -76 -1 -86 76 1 86 -76 -1 -86 59 1 67 26 blend
```
anthrotype, line 293: not immediately obvious to me if this diff was intentional in the expected test result

anthrotype (approved):
LGTM (good to see you back!)

Would be nice to also add a test that confirms that we are indeed avoiding error accumulation from multiple deltas

Lib/fontTools/misc/roundTools.py
behdad: Oh, didn't know we are avoiding imports in new modules. Cool. Done.

Lib/fontTools/varLib/__init__.py
behdad: The `tolerance` argument seems to be still used in the IUP optimizer. So I left it as is.

Tests/varLib/data/test_results/TestSparseCFF2VF.ttx
behdad: Right. This changed one 0 delta to -1. I assumed it's a result of the error accumulation changes I've made.

That said, there's a tweak I want to make to the rounding still; I'll comment below.

behdad:
Thanks for the reviews!

One thing that has been bothering me since last night is this: if a delta vector consists of numbers up to +0.5, then ideally we should (and we were) eliminating that delta vector. But now because of `otRound()`, we are not. Moreover, `otRound` is not symmetric, so -0.5 behaves differently.

For these reasons, and since these are deltas, not absolute values, I have convinced myself that using the Python3 `round()` builtin, that does banker's rounding, is desirable. Am going to do that. Please object if you will.

behdad:
Note that GlyphCoordinate deltas (ie., `gvar` deltas) are still rounding using `otRound` because they are done in `GlyphCoordinates.__round__()`.

anthrotype:
yeah, I also am under the impression that otRound (round half towards +Inf) only makes sense for absolute values (e.g. XY coordinates), rather than relative values (ie. deltas) or distances. I think I agree with you that a more balanced round function would make more sense for deltas. I'm a bit concerned that this would create binary diffs in most of the VF out there. FWIW does the OT spec have an opinion on this?

anthrotype:
(on the other hand, this PR is precisely intended to produce a better rounding overall so binary diffs are expected and desired)

behdad:
> I'm a bit concerned that this would create binary diffs in most of the VF out there.

Humm. Right. We should definitely consider that.

> FWIW does the OT spec have an opinion on this?

Not that I know of.

> (on the other hand, this PR is precisely intended to produce a better rounding overall so binary diffs are expected and desired)

Fair point indeed.

Lib/fontTools/varLib/__init__.py
anthrotype: Ah, I missed that one (that chunk of code was hidden), thanks

Tests/varLib/data/test_results/TestSparseCFF2VF.ttx
behdad: Good news. Rounding deltas with round3 instead of otRound indeed reverted this change.

behdad:
Okay I think I'm happy with this now.

> Would be nice to also add a test that confirms that we are indeed avoiding error accumulation from multiple deltas

Indeed. Let me see what I can do...

behdad:
> Let me see what I can do...

I'm failing to think of how to test this in a way that doesn't make assumptions about the internals of the Model...

anthrotype (approved):

behdad:
> I'm failing to think of how to test this in a way that doesn't make assumptions about the internals of the Model...

Okay here's the test: https://github.com/fonttools/fonttools/pull/2217

PR #2217, opened 2021-03-05
Test var model error

behdad:
From the commit message:

    [varLib.models] Add test for modeling rounding error
    
    Tests https://github.com/fonttools/fonttools/pull/2214
    
    If you flip demo to True, it does a slower test and demos the new error as well
    as the error the old code was producing (ie. rounding deltas post-modeling).
    
    Indeed, the new error is always capped by 0.5 as expected, whereas the old one
    was unbounded. Here's the worst-case error of the bad code:
    
    ...
    240     0.42    4.8
    ...
    
    240 is just the line number. 0.42 is new error. 4.8 is old error.

### DIFF
--- a/Tests/varLib/models_test.py
+++ b/Tests/varLib/models_test.py
@@ -33,41 +33,39 @@ def test_supportScalar():
     assert supportScalar({'wght':.2}, {'wght':(0,2,3)}) == 0.1
     assert supportScalar({'wght':2.5}, {'wght':(0,2,4)}) == 0.75
 
-def test_modeling_error(*, demo=False):
-    # https://github.com/fonttools/fonttools/issues/2213
-    if demo:
-        numLocations=127
-        numSamples=509
-    else:
-        numLocations=31
-        numSamples=251
 
-    locations = [{'axis':float(i)/numLocations} for i in range(numLocations)]
+@pytest.mark.parametrize(
+    "numLocations, numSamples", [
+        (127, 509),
+        (31, 251),
+    ]
+)
+def test_modeling_error(numLocations, numSamples):
+    # https://github.com/fonttools/fonttools/issues/2213
+    locations = [{'axis': float(i)/numLocations} for i in range(numLocations)]
     masterValues = [100. if i else 0. for i in range(numLocations)]
 
     model = VariationModel(locations)
 
     for i in range(numSamples):
-        loc = {'axis':float(i)/numSamples}
+        loc = {'axis': float(i)/numSamples}
         scalars = model.getScalars(loc)
 
         deltas_float = model.getDeltas(masterValues)
-        deltas_round   = model.getDeltas(masterValues, round=round)
+        deltas_round = model.getDeltas(masterValues, round=round)
 
         expected = model.interpolateFromDeltasAndScalars(deltas_float, scalars)
-        actual   = model.interpolateFromDeltasAndScalars(deltas_round, scalars)
+        actual = model.interpolateFromDeltasAndScalars(deltas_round, scalars)
 
         err = abs(actual - expected)
         assert err <= .5, (i, err)
 
-        if demo:
-            deltas_late_round   = [round(d) for d in deltas_float]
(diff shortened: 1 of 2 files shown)


justvanrossum:
So, this test verifies that with the new delta rounding behavior the maximum error is indeed 0.5, right? Just like you predicted, and that makes sense.

FWIW, if the end result is also rounded (often the case), the maximum error becomes 1, no matter which rounding variant I use:

```python
import math
from fontTools.misc.roundTools import otRound
from fontTools.varLib.models import VariationModel

def roundAwayFromZero(value):
    if value >= 0:
    	return int(math.floor(value + 0.5))
    else:
    	return int(math.ceil(value - 0.5))

locations = [{}, {"a": 0.2}, {"a": 0.6}, {"a": 1.0}]
model = VariationModel(locations)
masterValues = [2, 101, 55, 200]
loc = locations[2]

for deltaRound in [round, otRound, roundAwayFromZero]:
    for valueRound in [round, otRound, roundAwayFromZero]:
        deltas = model.getDeltas(masterValues, round=deltaRound)
        value = model.interpolateFromDeltas(loc, deltas)
        assert abs(value - masterValues[2]) <= 0.5
        print("rounded value error:", abs(valueRound(value) - masterValues[2]))
```

The output:
```
rounded value error: 1
rounded value error: 1
rounded value error: 1
rounded value error: 1
rounded value error: 1
rounded value error: 1
rounded value error: 1
rounded value error: 1
rounded value error: 1
```

I'm not saying this is bad, but is some justification for #2218

behdad:
> I'm not saying this is bad, but is some justification for #2218

Yeah I'm not saying your zero-error request is invalid :).

I'll check why this produces rounded error 1 all the time. I didn't expect that.

justvanrossum:
> I'll check why this produces rounded error 1 all the time. I didn't expect that.

It doesn't, just for this carefully picked set of master values :)

behdad:
> > I'll check why this produces rounded error 1 all the time. I didn't expect that.
> 
> It doesn't, just for this carefully picked set of master values :)

Ah yes. Got it.

Lib/fontTools/varLib/models.py
```diff
@@ -379,9 +379,9 @@ def getDeltas(self, masterValues, *, round=noRound):
 			out.append(round(delta))
 		return out
 
-	def getDeltasAndSupports(self, items, *, round=noRound):
+	def getDeltasAndSupports(self, items, **kwargs):
```
anthrotype, line 382: i don't like too much residual `**kwargs` at the end of function signatures, they make the API less clear. I think it was better with the keyword-only `round` parameter

anthrotype (commented):
Hope you don't mind I took the liberty to turn your test_modeling_error into an actual pytest-style test, so it fits with the rest of the test module and suite.

LGTM, but see my other comment on kwargs

anthrotype:
Btw, I kept the slow test to be still run by default. It takes about 1 second, so I think it's ok-ish. But if we think this is too long, we could modify the default pytest config (in setup.cfg) and have it automatically skip all the tests marked as "slow", unless one deliberately selects them with the `-m slow` CLI option.

Lib/fontTools/varLib/models.py
behdad: I don't like them either. I did when I was changing the default round and realized that the default was hardcoded in 4 places... Up to you.

behdad:
> Btw, I kept the slow test to be still run by default. It takes about 1 second, so I think it's ok-ish. But if we think this is too long, we could modify the default pytest config (in setup.cfg) and have it automatically skip all the tests marked as "slow", unless one deliberately selects them with the `-m slow` CLI option.

Thanks. The last 5 lines of the test don't test anything, and do print hundreds of lines. So I'd comment those out or outright remove.

justvanrossum (commented):
LGTM

anthrotype (approved):

PR #2218, opened 2021-03-06
[varLib] Add optional 'simple' supports/region builder

justvanrossum:
Add an optional 'simple' supports/region builder for 'simple' master configurations, addressing some comments in #2207.

With 'simple' I mean:
- There are masters at all corners of the designspace
- Intermediates on an axis must be present at all other axis master locations, too. For example, in a `wght`/`wdth` system, if there is a `wdth` intermediate for narrow, the same intermediate must exist for wide.

The resulting model is equivalent with the one build by the "generic" solver, but it will behave better with respect to rounding at the peak positions: intermediate masters always get matched exactly, instead of being (maximally) 0.5 unit off. (This is especially important if 100% metrics compatibility is desired at the master locations: this is currently not possible with the general solver.)

There will be fewer deltas participating at any one point in the designspace, so this will be more efficient at runtime.

However, as Behdad pointed out, the deltas to be stored will likely be larger, so may be stored less efficiently.

### DIFF
--- a/Lib/fontTools/varLib/models.py
+++ b/Lib/fontTools/varLib/models.py
@@ -197,12 +197,13 @@ class VariationModel(object):
 	  7: 0.6666666666666667}]
 	"""
 
-	def __init__(self, locations, axisOrder=None):
+	def __init__(self, locations, axisOrder=None, preferSimpleRegions=False):
 		if len(set(tuple(sorted(l.items())) for l in locations)) != len(locations):
 			raise VariationModelError("Locations must be unique.")
 
 		self.origLocations = locations
 		self.axisOrder = axisOrder if axisOrder is not None else []
+		self.preferSimpleRegions = preferSimpleRegions
 
 		locations = [{k:v for k,v in loc.items() if v != 0.} for loc in locations]
 		keyFunc = self.getMasterLocationsSortKeyFunc(locations, axisOrder=self.axisOrder)
@@ -212,7 +213,11 @@ class VariationModel(object):
 		self.mapping = [self.locations.index(l) for l in locations]
 		self.reverseMapping = [locations.index(l) for l in self.locations]
 
-		self._computeMasterSupports(keyFunc.axisPoints)
+		if preferSimpleRegions and self._isSimpleModel(locations, keyFunc.axisPoints):
+			self._computeMasterSupportsSimple(keyFunc.axisPoints)
+		else:
+			self._computeMasterSupports(keyFunc.axisPoints)
+		self._computeDeltaWeights()
 		self._subModels = {}
 
 	def getSubModel(self, items):
@@ -221,7 +226,11 @@ class VariationModel(object):
 		key = tuple(v is not None for v in items)
 		subModel = self._subModels.get(key)
 		if subModel is None:
-			subModel = VariationModel(subList(key, self.origLocations), self.axisOrder)
+			subModel = VariationModel(
+				subList(key, self.origLocations),
+				self.axisOrder,
+				self.preferSimpleRegions,
+			)
 			self._subModels[key] = subModel
 		return subModel, subList(key, items)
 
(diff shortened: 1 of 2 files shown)


justvanrossum:
A followup could be to add a lib key to the designspace document that turns on the new behavior. But I'd first like to see if there's some support for this PR in general.

behdad:
Thanks Just.

I'm certainly open this this line of work.

I'm against options that makes designers having to make decisions without knowing exactly what they are choosing. How about we add a `tolerance=.5` argument to `VariationModel`? That satisfies your needs but is automatic.

We can also relax the conditions under which that's attainable. If we cannot *guarantee* such a model, the object creation would fail.

behdad:
With a guarantee that `tolerance=.5` would never fail.

justvanrossum:
Without knowing the master values, what does that tolerance mean, though?

justvanrossum:
My thinking was:
- Your general algorithm is pretty fantastic, and should probably always be used by default, as it imposes no restrictions on the master configuration
- In some fringe cases (100% master precision requirement, with a simple master configuration), it is better to use the "simple" algorithm
- Those cases are truly expert use, so having to set an option explicitly isn't going to be a problem

Alternatively, I could see us always use the simple algorithm when the master configuration allows it. 

Advantage:
- Zero config

Disadvantages:
- No control for the designer
- Possibly larger binaries
-  Not backward compatible for post-processing client code that expects output from the general algorithm

justvanrossum:
Third option: always use the simple algorithm when the master configuration allows it, with a flag to optionally turn it off: opt-out vs opt-in.

justvanrossum:
I just realized that the "simple" region configuration is also possible in a system where some "corner masters" are absent. Will think of how to improve the algorithm to accomodate that.

Trivial example: `[{}, {'a': 1.0}, {'b': 0.3}, {'b': 0.6}, {'b': 1.0}]`

justvanrossum:
The last commit broadened what can be considered a "simple" model:

	A model is "simple" when any two axes that actively cooperate
	(ie. have masters at non-zero positions for both axes) have a
	complete set of intermediates. For example, if axes "a" and "b"
	cooperate, and axis "a" has masters at 0, 0.5 and 1 from the
	neutral master, it must have those intermediates at all values
	for axis "b" as well.

	This property allows us to use a simpler algorithm to compute
	the master supports, which will have fewer overlapping regions.

I'm not convinced of how well that is worded, suggestions welcome.

justvanrossum:
To clarify for the casual passer-by: the alternative "simple" region algorithm proposed here is 100% equavalent to the existing (more general) algorithm: the resulting interpolated shapes will be the same at all positions within the design space. (There's even a unit test that verifies that.)

There will be tiny differences due to different rounding, and that's where the "simple" algorithm is better: it guarantees that the  shapes and metrics at all master positions match the input masters _exactly_, where the general algorithm may be (maximally) 0.5 units off (which practically may be one unit, depending on rendering environment).

behdad:
So I'm okay merging this for now, on the condition that we can remove it and reconcile the ideas into the main model soon. I've since solved the modeling theoretical problems I was dealing with. I'll comment here or in another issue.

justvanrossum:
I'm also ok with leaving this unmerged for now, for reference, if you're confident you can recreate the functionality within the main model. Whichever you find more convenient.

khaledhosny:
What is the status of this pull request, do we still want to keep it unmerged?

justvanrossum:
Merging it is probably better than the status quo. But ugh, whitespace merge conflict :(

behdad:
I'll take a look in the next ten days to see if I can adjust the default algorithm to generate more simple regions.

behdad:
Hi @justvanrossum. Based on your code here, I modified the generic solver. It's in https://github.com/fonttools/fonttools/pull/2717

PR #2220, opened 2021-03-10
[otlLib] ligatureSubstBuilder.inferGlyphClasses() adds GSUB lookup type 4 glyphs to GDEF ligatures class

### LINKED ISSUE #2210: [feaLib] generate GDEF classes do not populate ligature class with lookup type 4 glyphs
makeOTF fills in the GDEF ligature class (if not defined in feature code) with the resulting glyphs from GSUB LookupType 4 substitutions. It doesn't appear that feaLib does the same ([from this far too long winded issue I started and closed over at fontmake](https://github.com/googlefonts/fontmake/issues/746))

moyogo:
Fixes #2210 and updates tests with GDEF by adding GSUB lookup type 4 glyphs to GDEF "ligatures" class.
See https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#9b-gdef-table.

Note: This doesn't add components to the GDEF "components" class.

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -776,12 +776,15 @@ class Builder(object):
                     )
                 )
                 classes.update(lookup.inferGlyphClasses())
-            markClassesUsed = 3 in classes.values()
-            for markClass in self.parseTree.markClasses.values():
-                for markClassDef in markClass.definitions:
-                    for glyph in markClassDef.glyphSet():
-                        classes[glyph] = 3
-            if not markClassesUsed and not ignoreClassFlagsUsed:
+            # Only infer glyph classes if mark classes are used or if ignore
+            # flags are used.
+            if 3 in classes.values() or ignoreClassFlagsUsed:
+                # Prefer mark glyph class to other glyph class
+                for markClass in self.parseTree.markClasses.values():
+                    for markClassDef in markClass.definitions:
+                        for glyph in markClassDef.glyphSet():
+                            classes[glyph] = 3
+            else:
                 classes.clear()
         if classes:
             result = otTables.GlyphClassDef()


anthrotype:
I don't know if it's a good idea to create a GDEF by default every time you have a ligature substitution.. That would modify a lot of fonts out there (a lot of tests changed in this PR). Even the FEA spec says that the GDEF classes should be inferred only when:

> If any mark class has been defined, or if any of the lookup flags for skipping glyphs of a certain class have been seen...

Lib/fontTools/feaLib/builder.py
```diff
@@ -764,12 +764,25 @@ def buildGDEFGlyphClassDef_(self):
...
+                    )
+                )
                 classes.update(lookup.inferGlyphClasses())
+            markClassesUsed = 3 in classes.values()
             for markClass in self.parseTree.markClasses.values():
                 for markClassDef in markClass.definitions:
                     for glyph in markClassDef.glyphSet():
                         classes[glyph] = 3
+            if not markClassesUsed and not ignoreClassFlagsUsed:
+                classes.clear()
```
anthrotype, line 785: why you clear after the loop? Before you enter the for loop above, you aready know whether markClassesUsed or ignoreClassFlagsUsed

Lib/fontTools/feaLib/builder.py
```diff
@@ -764,12 +764,25 @@ def buildGDEFGlyphClassDef_(self):
...
         else:
             classes = {}
+            ignoreClassFlagsUsed = False
             for lookup in self.lookups_:
+                ignoreClassFlagsUsed = (
+                    ignoreClassFlagsUsed or (
+                        lookup.lookupflag & (
+                            otl.LOOKUP_FLAG_IGNORE_BASE_GLYPHS +
+                            otl.LOOKUP_FLAG_IGNORE_LIGATURES +
+                            otl.LOOKUP_FLAG_IGNORE_MARKS
```
anthrotype, line 774: so if _any_ of the `lookupflags Ignore*` are used, you infer all of them? I would have interpreted the wording of the FEA spec as meaning that, if `ignoreMarks` is used, infer class 3, if `ignoreLigature` is used, infer class 2, etc.

anthrotype: (emphasis mine)

> if any of the lookup flags for skipping glyphs _of a certain class_...

behdad:
LGTM in principle. Thanks.

anthrotype:
@behdad does one need to specify all 4 categories if _any_ of the `lookupflag Ignore*` are used? Or is it sufficient to only list the marks (class 3) when `lookupflag IgnoreMarks` is used, the ligatures (class 2) when `lookupflag IgnoreLigatures` is used, etc. Can one save a few bytes by not listing those classes that are not explicitly ignored via lookupflag?

behdad:
> @behdad does one need to specify all 4 categories if _any_ of the `lookupflag Ignore*` are used? Or is it sufficient to only list the marks (class 3) when `lookupflag IgnoreMarks` is used, the ligatures (class 2) when `lookupflag IgnoreLigatures` is used, etc. Can one save a few bytes by not listing those classes that are not explicitly ignored via lookupflag?

While you can do that, I advise against doing it. It's just more complexity...

The glyph classes other than mark are really badly designed and documented.

One thing is for sure: *If* you add GDEF glyph classes, you MUST generate mark classes. That affects whether or not mark glyph advances are zeroed. Ligature / base, are only really used for skipping. But is not fully unimaginable that an implementation would assume if a glyph has ligature carets, then it must be marked as ligature in GDEF.

To keep things simple, I suggest either you synthesize a full glyph-class table, or leave None.

Lib/fontTools/otlLib/builder.py
```diff
@@ -782,6 +782,14 @@ def __init__(self, font, location):
     def equals(self, other):
         return LookupBuilder.equals(self, other) and self.ligatures == other.ligatures
 
+    def inferGlyphClasses(self):
+        result = {
+            glyph: 2
+            for glyph in self.ligatures.values()
+            if glyph != self.SUBTABLE_BREAK_
+        }
+        return result
```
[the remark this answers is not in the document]
moyogo: The mark ligature will be assigned the ligature class by LigatureSubtBuilder.inferGlyphClasses() but then buildGDEFGlyphClassDef_() will assign it to the mark class if it's in any defined markClasses.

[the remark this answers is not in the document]
anthrotype: sounds good. would be nice if there was an explicit test for that

Lib/fontTools/feaLib/builder.py
```diff
@@ -764,12 +764,28 @@ def buildGDEFGlyphClassDef_(self):
...
```
anthrotype, line 774: minor: you could union these three flags before the for loop in some local (or even global) variable instead of doing it for each lookup, result is always the same

anthrotype (approved):

khaledhosny:
What is the status of this pull request?

PR #2221, opened 2021-03-10
[subset] Improve PairPosFormat2 subsetting

behdad:
This does two things:

1. Intersect subsetter glyphset with the table's Coverage before
   passing to ClassDef1 for subsetting.  Anything that doesn't
   get past Coverage wouldn't ever get to ClassDef1,

2. Never reuse class0 of ClassDef2.  There's unspoken assumption
   that ClassDef2's class0 is never used for actual kerning, since
   that's the unbounded "every other glyph" class.  Previously our
   ClassDef subsetter was reusing class0 if "every other glyph"
   happened to become empty because of the subset glyphset.  Don't
   do that for PairPos's ClassDef2.  As a result of this assumption,
   don't keep a PairPosClass2 subtable if only ClassDef2's class0
   survived subsetting.

Would be good to add tests for both.

Related to https://github.com/harfbuzz/harfbuzz/issues/2703

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -427,13 +427,14 @@ def intersect_class(self, glyphs, klass):
 		     if v == klass and g in glyphs)
 
 @_add_method(otTables.ClassDef)
-def subset(self, glyphs, remap=False):
+def subset(self, glyphs, remap=False, useClass0=True):
 	"""Returns ascending list of remaining classes."""
 	self.classDefs = {g:v for g,v in self.classDefs.items() if g in glyphs}
 	# Note: while class 0 has the special meaning of "not matched",
 	# if no glyph will ever /not match/, we can optimize class 0 out too.
+	# Only do this if allowed.
 	indices = _uniq_sort(
-		 ([0] if any(g not in self.classDefs for g in glyphs) else []) +
+		 ([0] if ((not useClass0) or any(g not in self.classDefs for g in glyphs)) else []) +
 			list(self.classDefs.values()))
 	if remap:
 		self.remap(indices)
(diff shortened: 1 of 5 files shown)


anthrotype (approved):
LGTM

anthrotype:
> Would be good to add tests for both.

it looks like @garretrieger is already adding some test files along with the respective harfbuzz subsetter's changes in https://github.com/harfbuzz/harfbuzz/commit/f6298da947bc45d77d631266f142c8ebf1e5856e

Could you point to the actual file(s) relevant for this issue so we can add it here too?

garretrieger:
The font file we use in the harfbuzz tests is: https://github.com/googlefonts/harfbuzz/blob/gpos_2/test/subset/data/fonts/gpos2_2_font5.otf

and the specific subset text that triggers a situation effected by this change is:

.#

anthrotype:
@garretrieger Hey Garret, I tried using the font file you pointed to from harfbuzz test suite, called "test/subset/data/fonts/gpos2_2_font5.otf". If I subset it using the text string you said, with the command`fonttools subset gpos2_2_font5.otf --text='.#'` I get the same subsetted output whether @behdad's current PR patch is applied or not, in both cases the GPOS table is emptied (no lookups left).
Are you sure this is the right font/string combination?

anthrotype:
@behdad could you help me figure out a good test case for this? someting like "given some kern feature containing pos class rules, and this subsetter --text input, the expected result is ..." I can then turn that into an actual test. Or I just assign it back to you and you finish this.

garretrieger:
Hmm let me check on the result of that test case in harfbuzz.

garretrieger:
Ok, I see the same in the expected files. These test cases
!#
!#%

both produced changes in the fonttools expected files in the PR and have a non-empty GPOS. So I'd try those instead. You can also look at all the cases we test (it's a cross product of the fonts, profiles, and subsets):

https://github.com/harfbuzz/harfbuzz/blob/master/test/subset/data/tests/layout.gpos2.tests

anthrotype:
Thanks Garrett, I tried both "gpos2_1_font7.otf" and "gpos2_2_font5.otf" with the combinations of text strings you posted, and they both result in an empty GPOS whenever I run through the `fonttools subset <fontfile> --text-file=subset.txt`, whether or not @behdad's patch from this current fonttools PR is applied or not. I run the same command with fonttools installed in editable mode, checking out the main branch (where the patch is not applied) and this PR branch, and I don't get any different result. In all cases, the resulting GPOS is empty. I need to figure out a test case that makes the current main branch fail, and this PR branch pass.

anthrotype:
AH! I had to pass `--layout-features+=test` otherwise the "test" feature under which those PairPosFormat2 were located were dropped by default :)

anthrotype:
@behdad PTAL (I can't add you as reviewer as you're the PR author)

PR #2225, opened 2021-03-15
Correct otlLib documentation

simoncozens:
When trying to use otlLib to build some lookups I found the documentation wasn't quite accurate.

### DIFF
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -1037,8 +1037,8 @@ class MarkMarkPosBuilder(LookupBuilder):
             source which produced this lookup.
         marks: An dictionary mapping a glyph name to a two-element
             tuple containing a mark class ID and ``otTables.Anchor`` object.
-        baseMarks: An dictionary mapping a glyph name to a two-element
-            tuple containing a mark class ID and ``otTables.Anchor`` object.
+        baseMarks: An dictionary mapping a glyph name to a dictionary
+            containing one item: a mark class ID and a ``otTables.Anchor`` object.
         lookupflag (int): The lookup's flag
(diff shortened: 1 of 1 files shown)


Lib/fontTools/otlLib/builder.py
```diff
@@ -1029,7 +1029,7 @@ class MarkMarkPosBuilder(LookupBuilder):
         builder.marks["acute"]     = (0, a1)
         builder.marks["grave"]     = (0, a1)
         builder.marks["cedilla"]   = (1, a2)
-        builder.baseMarks["acute"] = (0, a3)
+        builder.baseMarks["acute"] = {0: a3}
```
anthrotype, line 1032: the docstring for `baseMarks` below says it's a tuple

```
        baseMarks: An dictionary mapping a glyph name to a two-element
            tuple containing a mark class ID and ``otTables.Anchor`` object.
```

anthrotype: and what about the other `marks`? are those a 2-tuple? If so, I wonder why they should differ from `baseMarks`

simoncozens: marks *is* a tuple (because you can only have one anchor per mark) but baseMarks is a dictionary because you can have multiple anchors per base. I'll fix the doctoring too.

simoncozens: Wait, sorry, I'm looking at the wrong thing. The interface is slightly weird (you can only have one element in the dictionary, because you can only have one mark on a base mark) but it's parallel to the MarkBasePosBuilder interface, so it makes sense.

anthrotype (approved):

PR #2228, opened 2021-03-18
Support variable scalars in feature file

simoncozens:
This is not for merging, but for demonstration purposes.

Implements the variable scalar format proposed in [this AFDKO discussion](https://github.com/adobe-type-tools/afdko/issues/153#issuecomment-800973369):

```fea
pos A B' <0 (wght=200:-100 wght=900:-150 wght=900,wdth=150:-120) 0 0> C' D;
```


davelab6:
https://twitter.com/simoncozens/status/1373175391133655041 has a visual demo :)

simoncozens:
OK, something odd seems to have happen to `main` since I started writing this, because it passed tests until I rebased. What is happening is that when builder_test does:

```python
        # Check that:
        # 1) tables do compile (only G* tables as long as we have a mock font)
        # 2) dumping after save-reload yields the same TTX dump as before
            if tag in font:
                data = font[tag].compile(font)
                font[tag].decompile(data, font)
```

the first TTX is correct, but after compile/decompile roundtripping, additional comments for `FeatureVariationCount`, `ConditionCount` and `LookupCount` have been added to the TTX:

```diff
--- /Users/simon/others-repos/fonttools/Tests/feaLib/data/variable_conditionset.ttx
+++ /var/folders/js/26vkgd8s5x15c5kwbhq2c2jw0000gn/T/tmpp1m8cmke/tmp1.ttx
@@ -39,10 +39,8 @@
     </LookupList>
     <FeatureVariations>
       <Version value="0x00010000"/>
-      <!-- FeatureVariationCount=1 -->
       <FeatureVariationRecord index="0">
         <ConditionSet>
-          <!-- ConditionCount=1 -->
           <ConditionTable index="0" Format="1">
             <AxisIndex value="0"/>
             <FilterRangeMinValue value="0.625"/>
@@ -54,7 +52,6 @@
           <SubstitutionRecord index="0">
             <FeatureIndex value="0"/>
             <Feature>
-              <!-- LookupCount=1 -->
               <LookupListIndex index="0" value="0"/>
             </Feature>
           </SubstitutionRecord>
```


 Obviously I can't keep both tests happy. I'm not sure what to do about this.

anthrotype:
the meaning of that tests is, if you're building those FeatureVariations and Feature records, you need to make sure the `*Count` fields are set at build time, and not rely on the compiler which will (re)set those for you; that way dumping to TTX straight after constructing the table objects will give the same output as building followed by compile.
The expected variable_conditionset.ttx should have those count fields set, but from that diff it looks like it has not.

anthrotype:
> it passed tests until I rebased

I'm not exactly sure what would have changed in the meantime

simoncozens:
Got it, thank you.

ctrlcctrlv:
Please rebase again when you get a chance @simoncozens. I need a commit later in history.

PR #2229, opened 2021-03-19
[varLib/otlLib] Allow merging into VariationFont without first saving GPOS PairPos2

behdad:
From commit log:
```
    [otlLib] Generate Value1 / Value2 if respective formats are non-zero
    
    Previously otlLib was generating None if the values themselves were
    empty even if the value format was non-empty.  This happened to work
    for compiling to binary since the compiler handles Value=None.
    
    But this was confusing varLib.merger module (as in when building
    variable fonts from such otlLib-built master GSUB/GPOS tables, without
    roundtripping to OTF/TTF binary first), because in varLib.merger,
    a None means "this master doesn't provide that info; skip it"; whereas
    in a PairPos table a None as generated by otlLib simply meant "all
    values are zero", which is different from "this master doesn't
    provide this value".
    
    This fixes that, such that ufo2ft can build variable-font without
    saving masters to binary.
    
    Part of https://github.com/googlefonts/ufo2ft/pull/486
```

A few tests fail because expected ttx changed. I think the expectation should be updated. I haven't checked, but if you build font binary and back, you get the ttx output of what this PR is doing, so the old expectation was wrong IMO.


behdad:
[First commit is unrelated. I just had it sitting here.]

behdad:
Currently `Tests/feaLib/builder_test.py` does this:
```
    def check_feature_file(self, name):
        font = makeTTFont()
        feapath = self.getpath("%s.fea" % name)
        addOpenTypeFeatures(font, feapath)
        # Make sure we can produce binary OpenType tables, not just XML.
            if tag in font:
                font[tag].compile(font)
        ...
```
It should, instead, decompile the saved table and check that the same ttx was produced. That will fail for some things maybe (Coverage / ClassDef formats anyone?), but we should enable and see what breaks and fix those.

madig:
@behdad you mean remove the `compile` loop? I did that locally and all tests still pass.

behdad:
No, I mean we should decompile the compiled binary and compare the ttx again. That would have caught this bug and will catch similar ones.

Thanks for updating the tests.

madig:
Mh, different PR?

anthrotype:
different pr sgtm

anthrotype (approved):

PR #2231, opened 2021-03-20
[py23] Nit: change py23 deprecation warning from "next release" to "future release"

chrissimpkins:
This warning doesn't seem to have teeth since it has been around for 2+ yrs :)


justvanrossum (approved):

chrissimpkins:
Thanks Just.

PR #878, opened 2017-03-08
[varLib.interpolate_layout] Tests for GPOS LookupType 1 and LookupType 2 specific pairs

miguelsousa:
The failures are due to XML round-tripping mismatches: `<Coverage Format="1">` vs. `<Coverage>`.
Cosimo, please do your magic. I believe the TTX result files are correct. Thanks!

Issue #475, opened 2016-01-22
[merge] Exception: Don't know how to merge key Format of class ClassDef


anthrotype:
I think it might be related to this commit: 6851f66d1810a4d3f484f2a82e39960e1a677f42 

```
$ pyftmerge --verbose --timing font1.ttf font2.ttf
Merged 'cmap'.
Took 0.010s to merge 'cmap'
Merged 'GSUB'.
Took 0.000s to merge 'GSUB'
Merged 'post'.
Took 0.000s to merge 'post'
Don't know how to merge 'VDMX'.
Dropped 'VDMX'.
Took 0.002s to merge 'VDMX'
Merged 'GPOS'.
Took 0.000s to merge 'GPOS'
Merged 'maxp'.
Took 0.000s to merge 'maxp'
Merged 'prep'.
Took 0.000s to merge 'prep'
Merged 'fpgm'.
Took 0.000s to merge 'fpgm'
Merged 'hmtx'.
Took 0.000s to merge 'hmtx'
Merged 'glyf'.
Took 0.001s to merge 'glyf'
Merged 'OS/2'.
Took 0.001s to merge 'OS/2'
Merged 'head'.
Took 0.000s to merge 'head'
Merged 'gasp'.
Took 0.000s to merge 'gasp'
Merged 'loca'.
Took 0.000s to merge 'loca'
Merged 'hhea'.
Took 0.000s to merge 'hhea'
Traceback (most recent call last):
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/merge.py", line 106, in mergeObjects
    mergeLogic = logic[key]
KeyError: 'Format'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/merge.py", line 109, in mergeObjects
    mergeLogic = logic['*']
KeyError: '*'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/var/pyenv/versions/fonttools35/bin/pyftmerge", line 9, in <module>
    load_entry_point('fonttools', 'console_scripts', 'pyftmerge')()
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/merge.py", line 943, in main
    font = merger.merge(args)
  File "/Users/cosimolupo/Documents/Github/fonttools/Lib/fontTools/merge.py", line 773, in merge
[...]

Issue #92, opened 2014-01-19
[ttx] Option to preserve PUSHB/PUSHW etc


behdad:
From Werner:

"""
[FontTools-2.4-696-gfea81ee]

I'm not happy that the TrueType instruction dump in ttx no longer
shows the difference between PUSHB and PUSHW.  I agree that this
information isn't relevant to the functionality, however, it's a
valuable information while debugging and comparing dumps.

Similarly, it is quite helpful to see the formats of tables (Coverage,
GlyphClassDef, etc.)

Can you revert that change?  Or maybe add an option to ttx which
increases the verbosity of table dumps in general.
"""

schriftgestalt:
>  I agree that this information isn't relevant to the functionality,

In the case there is a `JMPR` command, the difference is indeed vital to the fictionality of the code. Compiling the instruction again will change the length of the code and thous invalidates the jump address. This will break all TTF generated with Glyphs.

schriftgestalt:
And an option to switch of the optimisation is not sufficient as everybody is using it to import/export the fonts with ttx will get broken fonts and will not know to set that flag. Maybe the other way around, add an option to optimise (and check if there is a jump command).

lemzwerg:
Indeed, I haven't thought of `JMPR`.

On the other hand: Making ttx squeeze a sequence of `PUSHB` and `PUSHW` as much as possible would be a good thing as an optimization, in case it can also properly adjust `JMPR` offsets.

schriftgestalt:
If someone figures out how to fix the jump offsets I’m more than happy with the optimisation.

behdad:
Ok I definitely had not thought about `JMPR` instruction.

Note that the original "optimize at write time" was done before my time.  What I did was change the bytecode to reflect that.  Ie. before, bytecode was being corrupted silently at write time!

I think we should disable this feature and enable it as an optimization option somehow.

Fixing the jump offset is NOT impossible, but a lot more work than this feature entails.  I suppose maybe we can still do this for the burst of push operations at the beginning of each bytecode string but never in the middle?

lemzwerg:
Yes. It shouldn't be too difficult to scan for `JMPR` instructions to find out the negative offset where `PUSHX` optimization has to stop.  Or, as a starter, simply refrain from `PUSHX` optimization as soon as a `JMPR` instruction is found.

justvanrossum:
[...]

PR #2236, opened 2021-03-22
[feaLib] Test that G* table compilation does not change data

madig:
Continuation of https://github.com/fonttools/fonttools/pull/2229.

@behdad


madig:
It's all missing `format` attributes 🤔

madig:
I'm lost. Setting e.g. `self.Format = 1` in `def buildCoverage` in otlLib/builder.py has no effect on e.g. the TTX diff for `BuilderTest.test_FeatureFile_AlternateChained`.

behdad:
This has some history. I'll type in a few hours.

behdad:
> I'm lost. Setting e.g. `self.Format = 1` in `def buildCoverage` in otlLib/builder.py has no effect on e.g. the TTX diff for `BuilderTest.test_FeatureFile_AlternateChained`:

I'm not sure what you mean by this.

But this is why this thing is happening. For GSUB/GPOS types `Coverage`, `ClassDef`, `SingleSubst`, `AlternateSubst`, and `LigatureSubst`, ttLib has code in `otTables.py` to unify their multiple different formats into a unified Python API.

So naturally, the above code should also remove the `.Format` attribute, since it's already handled out. However, it wasn't done that way and the Format was dangling around in the objects loaded from binary. The Format is recomputed at compile time, depending on glyph order and the contents of the object. So the old dangling .Format is ignored. It has no purpose.

So when I forked fonttools, one of the things I did was to remove Format attribute from those structs. That was in 2013:

  https://github.com/fonttools/fonttools/commit/e93e29125f6c88a352e15b4d7732fc94a6402548

That caused a breakage in Compositor. While that was happening, feaLib was also developed. Issues around the .Format attribute popped up there; that's 2015:

  https://github.com/robotools/compositor/issues/7
  https://github.com/fonttools/fonttools/issues/475

In 2016, after I managed to get @justvanrossum's attention and we all started working on the same codebase, this Format change remained a pet peeve of his, and a few others also fought hard for the Format to come back. So I conceded:

  https://github.com/fonttools/fonttools/issues/92#issuecomment-199153825

Since then, it has been a routine pain point, because we can't roundtrip test GSUB/GPOS exactly as you are hitting with this change :(. I've even had to put hacks in just to make it bearable:

https://github.com/fonttools/fonttools/commit/fb0c60cd2ce68df546ea2887946bc7f97bbb6921
https://github.com/fonttools/fonttools/commit/479b52e13dbe6b7395bc652aa92d6481824d67b7
https://github.com/fonttools/fonttools/pull/878

Now, I remember a couple of years ago @justvanrossum commented on an issue saying he has since started to see things more the way I was seeing. So I wonder if it's time to reconsider the issue. Not being able to test that saving to binary and back produces the same objects makes it really hard to have confidence in the objects we build, which indeed makes saving to binary and loading a necessary "normalization" step as was found in ufo2ft.  To fix it, we need to make roundtrip-testing possible first.

HTH

justvanrossum:
> So I wonder if it's time to reconsider the issue

If there's a general consensus that the Format field has no added value and only gets in the way, I will no longer insist on keeping it.

(Out of curiosity: @simoncozens recently improved compilation of certain GPOS subtables tables considerably, by compiling to the most efficient Format for the given data. Would discovering this and fixing that have been harder if the ttx hadn't contained the Format field, or would it not have made a difference? Would love to hear your opinion.)

madig:
Oof, sounds like an epic refactoring if we want to make it possible to test roundtrips?

How to proceed here?

behdad:
> If there's a general consensus that the Format field has no added value and only gets in the way, I will no longer insist on keeping it.

Thanks.  @anthrotype / @madig can you cherry-pick / adapt https://github.com/fonttools/fonttools/commit/e93e29125f6c88a352e15b4d7732fc94a6402548

> (Out of curiosity: @simoncozens recently improved compilation of certain GPOS subtables tables considerably, by compiling to the most efficient Format for the given data. Would discovering this and fixing that have been harder if the ttx hadn't contained the Format field, or would it not have made a difference? Would love to hear your opinion.)

Not really. For the structs that ttLib is presenting a higher-level API (Coverage, ClassDef, SingleSubst, AlternateSubst, and LigatureSubst), the ttLib compiler already picks the optimal format. Simon implemented that for other tables, namely the (Chain)Context{Subst,Pos}. Basically, for Coverage, ClassDef, SingleSubst, AlternateSubst, and LigatureSubst, you implemented otlLib.builder functionality right in the ttLib compiler.

behdad:
> Thanks. @anthrotype / @madig can you cherry-pick / adapt [e93e291](https://github.com/fonttools/fonttools/commit/e93e29125f6c88a352e15b4d7732fc94a6402548)


I'm working on this.

behdad:
> Oof, sounds like an epic refactoring if we want to make it possible to test roundtrips?

Wasn't too bad. Done: https://github.com/fonttools/fonttools/pull/2238

behdad:
I've rebased this now. Let's see test fails. I will fix the ValueRecord failures in https://github.com/fonttools/fonttools/pull/2237/files

behdad:
Humm. I'm not sure why the tests are passing without https://github.com/fonttools/fonttools/pull/2237

anthrotype:
> I'm not sure why the tests are passing without #2237

because we had committed the (wrong) expected test result, which that PR fixes

behdad:
> > I'm not sure why the tests are passing without #2237
> 
> because we had committed the (wrong) expected test result, which that PR fixes

But why doesn't going to binary and back catch that?

anthrotype:
the test_buildPairPosGlyphsSubtable is not affected by compiling to binary, it only tests (via getXML) straight after built

anthrotype (approved):

PR #2237, opened 2021-03-23
[otlLib] Ensure built ValueRecords are padded with 0s like when decompiled

anthrotype:
Follow up from https://github.com/fonttools/fonttools/pull/2229

see discussion at https://github.com/fonttools/fonttools/commit/2089d051265d96db9dfdbcf15af10e1e003ccd82#r48578120

### DIFF
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -2190,8 +2190,8 @@ def buildPairPosGlyphsSubtable(pairs, glyphMap, valueFormat1=None, valueFormat2=
         for glyph2, val1, val2 in sorted(p[glyph], key=lambda x: glyphMap[x[0]]):
             pvr = ot.PairValueRecord()
             pvr.SecondGlyph = glyph2
-            pvr.Value1 = val1 if valueFormat1 else None
-            pvr.Value2 = val2 if valueFormat2 else None
+            pvr.Value1 = ValueRecord(src=val1, valueFormat=valueFormat1) if valueFormat1 else None
+            pvr.Value2 = ValueRecord(src=val2, valueFormat=valueFormat2) if valueFormat2 else None
             ps.PairValueRecord.append(pvr)
         ps.PairValueCount = len(ps.PairValueRecord)
     self.PairSetCount = len(self.PairSet)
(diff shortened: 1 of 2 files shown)


madig:
I can't review this because I don't understand how this all works :O

behdad:
LGTM. You can / should do the same for PairPos as well.

behdad:
> I can't review this because I don't understand how this all works :O

I'd be happy to explain it all to you / others over video. Is not that complicated.

behdad:
So, if there's feature file like this (copying from `GPOS_2b.fea`):
```
@PUNC = [comma semicolon period];

feature kern {
  pos [A] @PUNC 1;
  pos [B C] [comma] 2;
  pos [D E F] [comma] 3;
  pos [D E F] [semicolon period] 4;
  pos [G] @PUNC <5 5 5 5>;
} kern;
```
then *if* we end up putting all in the same subtable, the value records we get in the builder contain just XAdvance for the first items but XPlacement="5" YPlacement="5" XAdvance="5" YAdvance="5" for the last item. However, if it's one subtable, we end up choosing valueFormat to be able to encode all of those, so it will be encoding all four of XPla/YPla/XAdv/YAdv. So we should create new values for all the previous ones to match...

Which is to say, your patch is correct. Just do it for PairPos2 as well. Is a bummer that we have to create new objects for all of these specially since the original objects will be discarded right after calling into the builder... Maybe we can have an `inPlace=False` parameter...

I'm reviewing to code for other places we might need this.

anthrotype:
thanks @behdad, feel free to merge this and add the rest of the changes

anthrotype:
or you can commit to this very same branch if you prefer. I'm off for today

behdad:
> or you can commit to this very same branch if you prefer. I'm off for today

Yeah easier. Thanks.

Here's another place:
https://github.com/fonttools/fonttools/blob/2089d051265d96db9dfdbcf15af10e1e003ccd82/Lib/fontTools/otlLib/builder.py#L2315

behdad:
> Here's another place:

That's all. There's one place for SinglePos and two for PairPos. the rest of GPOS doesn't use values.

behdad:
Okay this is ready to land.

I like to move the shared logic into `ValueRecordFactory` and call them `regularizeValues` and `regularizeValuePairs` perhaps, which would extract the ValueFormat as well as regularize the valuess. And only create values if necessary.

But that can be done separately. I'll create another issue for that and assign to myself.

behdad:
@anthrotype back to you to review. I can't assign you as reviewer. :))

behdad (approved):

anthrotype (commented):
LGTM, thanks behdad!

PR #2238, opened 2021-03-23
Remove .Format from Coverage, ClassDef, SingleSubst, LigatureSubst, AlternateSubst

behdad:
From https://github.com/fonttools/fonttools/commit/fb0c60cd2ce68df546ea2887946bc7f97bbb6921#issuecomment-805132445

Initially ~470 tests failed. I fixed all; expect for 25 mtiLib tests still fail. I don't understand why. Help appreciated.


anthrotype:
in mti_test.py, if you move the order in which you dump to ttx and compile the table (by moving the dump from built table before the table.compile), then you get the expected ttx output (without Formats). I think it's because the preWrite methods called by compile re-add the Format attrs, so if you dump after a compile you still get those..

```diff
diff --git a/Tests/mtiLib/mti_test.py b/Tests/mtiLib/mti_test.py
index 1666dd45..6704751e 100644
--- a/Tests/mtiLib/mti_test.py
+++ b/Tests/mtiLib/mti_test.py
@@ -174,13 +174,6 @@ class MtiTest(unittest.TestCase):
             self.assertEqual(tableTag, table.tableTag)
         tableTag = table.tableTag

-        # Make sure it compiles.
-        blob = table.compile(font)
-
-        # Make sure it decompiles.
-        decompiled = table.__class__()
-        decompiled.decompile(blob, font)
-
         # XML from built object.
         writer = XMLWriter(StringIO(), newlinestr='\n')
         writer.begintag(tableTag); writer.newline()
@@ -188,6 +181,13 @@ class MtiTest(unittest.TestCase):
         writer.endtag(tableTag); writer.newline()
         xml_built = writer.file.getvalue()

+        # Make sure it compiles.
+        blob = table.compile(font)
+
+        # Make sure it decompiles.
+        decompiled = table.__class__()
+        decompiled.decompile(blob, font)
+
         # XML from decompiled object.
         writer = XMLWriter(StringIO(), newlinestr='\n')
         writer.begintag(tableTag); writer.newline()
```

anthrotype:
there's still one mtiLib test failing after I apply the above patch though.. Should you also `del self.Format` in MultipleSubst maybe?

```
Tests/mtiLib/mti_test.py::MtiTest::test_MtiFile_mti/gsubmultiple_GSUB FAILED [100%]

==================================== FAILURES ====================================
___________________ MtiTest.test_MtiFile_mti/gsubmultiple_GSUB ___________________

self = <mti_test.MtiTest testMethod=test_MtiFile_mti/gsubmultiple_GSUB>

>   return lambda self: self.check_mti_file(os.path.join(*name.split('/')), tableTag=tableTag)

Tests/mtiLib/mti_test.py:220:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
Tests/mtiLib/mti_test.py:198: in check_mti_file
    self.expect_ttx(xml_binary,   xml_built, fromfile='decompiled',      tofile='built')
Tests/mtiLib/mti_test.py:154: in expect_ttx
E   AssertionError: TTX output is different from expected
------------------------------ Captured stderr call ------------------------------

--- decompiled
+++ built
@@ -7,7 +7,7 @@
       <LookupType value="2"/>
       <LookupFlag value="0"/>
       <!-- SubTableCount=1 -->
-      <MultipleSubst index="0" Format="1">
+      <MultipleSubst index="0">
         <Substitution in="janyevoweltelugu" out="jaivoweltelugu,nyasubscripttelugu"/>
         <Substitution in="kassevoweltelugu" out="kaivoweltelugu,ssasubscripttelugu"/>
       </MultipleSubst>
```

behdad:
Ah, thanks. Let's keep the test code in same order so we test this route. I'll add a comment as well.

Will fix preWrites and MultipleSubst.

behdad:
Okay! Ready to review / merge. All tests pass. :)

behdad:
> in mti_test.py, if you move the order in which you dump to ttx and compile the table (by moving the dump from built table before the table.compile), then you get the expected ttx output (without Formats). I think it's because the preWrite methods called by compile re-add the Format attrs, so if you dump after a compile you still get those..

So; we have feaLib/mtiLib/otlLib/fontBuilder building Python objects, then there's saving to binary and back, and to XML and back. Would be good if we can centralize testing of all that :))). Right now each test module does it's own.

But yeah, glad mti's had different order so I caught more to fix.

madig:
I like the centralization idea but don't know enough to meaningfully review this PR or centralize anything in ttLib :o

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -654,9 +654,15 @@ def decompile(self, reader, font):
 
 	def compile(self, writer, font):
 		self.ensureDecompiled()
+		# TODO Following hack to be removed by rewriting how FormatSwitching tables
+		# are handled.
+		# https://github.com/fonttools/fonttools/pull/2238#issuecomment-805192631
 		if hasattr(self, 'preWrite'):
+			deleteFormat = not hasattr(self, 'Format')
 			table = self.preWrite(font)
+			deleteFormat = deleteFormat and hasattr(self, 'Format')
```
madig, line 663: What does this double check do?

anthrotype: it means "delete Format attribute if it wasn't there before calling preWrite and it's been added by the preWrite"

madig: So why delete it at the end of the function?

anthrotype (approved):
LGTM

Lib/fontTools/ttLib/tables/otBase.py
anthrotype: it's used when serializing the tables, we only want to delete after compile is completed

behdad: It's just a hack. Ideally compilation shouldn't modify the objects. The `preWrite` methods are modifying by adding .Format. Eventually we will remove that, by rewriting how format-switching table machinery works. But for now, to make it work, I let `preWrite` add .Format, but will delete it at the end of compile when they are not needed anymore.

Another place we modify objects during compilation is when we split large lookups. But that's for another day...

Issue #2232, opened 2021-03-20
py23 module use within the project


chrissimpkins:
I took a quick look into what it would take to part ways with the py23 module imports within the project and it is significant.  The following is a ballpark estimate of matched strings in modules under `Lib/fontTools/` that does not account for a py23 module import statement in the match.  This may be an over count for the strings that are Python builtins/stdlib.

- `basestring`: 43 matches across 19 files
- `bytechr`: 65 matches across 16 files
- `byteord`: 112 matches across 32 files
- `BytesIO`: 35 matches across 19 files
- `bytesjoin`: 89 matches across 32 files
- `open`: 277 matches across 49 files
- `Py23Error`: 3 matches in 1 file
- `range`: 897 matches across 109 files
- `RecursionError`: 4 matches across 2 files
- `round`: 302 matches across 50 files
- `SimpleNamespace`: 18 matches across 4 files
- `StringIO`: 9 matches across 7 files
- `strjoin`: 24 matches across 15 files
- `Tag`: 744 matches across 47 files
- `tobytes`: 70 matches across 37 files
- `tostr`: 31 matches across 18 files
- `tounicode`: 33 matches across 13 files
- `UnicodeIO`: 5 matches across 3 files
- `xrange`: 3 matches in 1 file
- `zip`: 176 matches across 39 files

Any interest in beginning to chip away at this?

Related #2231 

---

TODO:

- [x] replace `basestring` with `str` #2241 
- [x] replace `UnicodeIO` with `StringIO` and import from `io` module
- [x] replace `tounicode` with `tostr`
- [x] import `BytesIO` from `io` module
- [x] import `SimpleNamespace` from `types` module
- [x] move `py23.bytechr` to new module
- [x] move `py23.byteord` to new module
- [x] move `py23.bytesjoin` to new module
- [x] move `py23.strjoin` to new module
- [x] move `py23.tobytes` to new module
- [x] move `py23.tostr` to new module
- [x] move `py23.Tag` to new module
- [x] remove `misc.py23` import statements

khaledhosny:
Some of these are easy to fix. `basestring` should be unconditionally replace with `str`, `UnicodeIO` with `StringIO` and both should be imported from `io` module together with `BytesIO`. `SimpleNamespace` should be imported from `types` module. 

`open`, `round`, `range`, `zip` need no change, `py23` is already using the Python 3 builtin, so removing the import is all needed.

[...]

PR #2241, opened 2021-03-27
py23 transition: Convert `basestring` to `str`

chrissimpkins:
Tracking in https://github.com/fonttools/fonttools/issues/2232

This PR replaces `basestring` with `str` across all library and test modules *except* misc.py23.  I tested this locally in cPy3.9 with the py23 module basestring definition removed and the test suite passes

### DIFF
--- a/Lib/fontTools/misc/psCharStrings.py
+++ b/Lib/fontTools/misc/psCharStrings.py
@@ -997,7 +997,7 @@ class T2CharString(object):
 			# If present, remove return and endchar operators.
 			if program and program[-1] in ("return", "endchar"):
 				program = program[:-1]
-		elif program and not isinstance(program[-1], basestring):
+		elif program and not isinstance(program[-1], str):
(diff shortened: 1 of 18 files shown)


justvanrossum (approved):

Lib/fontTools/misc/psCharStrings.py
```diff
@@ -1043,8 +1043,7 @@ def setBytecode(self, bytecode):
 		self.program = None
 
 	def getToken(self, index,
-			len=len, byteord=byteord, basestring=basestring,
-			isinstance=isinstance):
+			len=len, byteord=byteord, isinstance=isinstance):
```
khaledhosny, line 1046: Do any callers pass the `basestring`, would this change be an API break?

chrissimpkins: Good catch.  My fault.  I added it back in #2242

PR #2242, opened 2021-03-27
Replace basestring parameter in T2CharString.getToken

chrissimpkins:
Addresses https://github.com/fonttools/fonttools/pull/2241#pullrequestreview-622658587

### DIFF
--- a/Lib/fontTools/misc/psCharStrings.py
+++ b/Lib/fontTools/misc/psCharStrings.py
@@ -1043,7 +1043,8 @@ class T2CharString(object):
 		self.program = None
 
 	def getToken(self, index,
-			len=len, byteord=byteord, isinstance=isinstance):
+			len=len, byteord=byteord, basestring=str,
+			isinstance=isinstance):
 		if self.bytecode is not None:
 			if index >= len(self.bytecode):
 				return None, 0, 0


Lib/fontTools/misc/psCharStrings.py
```diff
@@ -1043,7 +1043,8 @@ def setBytecode(self, bytecode):
 		self.program = None
 
 	def getToken(self, index,
-			len=len, byteord=byteord, isinstance=isinstance):
+			len=len, byteord=byteord, basestring=str,
+			isinstance=isinstance):
```
khaledhosny, line 1047: You need to pass `basestring` to the `isinstance()` call below, otherwise code that depend on passing a different class will have it's behavior changed.

justvanrossum (commented):
I think these can all go: they're not part of an API, it's an old optimization (local lookups being faster than global lookups). So your original change was fine, and this PR should be abandoned.

chrissimpkins:
> I think these can all go: they're not part of an API, it's an old optimization (local lookups being faster than global lookups). So your original change was fine, and this PR should be abandoned.

Closing.  Thanks all!

behdad:
Agree with Just. The optimization still produced measurable results in 2016 when I last measured. Would be good to see if that's still the case.

justvanrossum:
If the optimization is still desirable, then the correct thing to do is add `str=str` to the `getToken()` arg list.

behdad:
Correct.

PR #2243, opened 2021-03-27
Misc py23 cleanups

justvanrossum:
Tracking in #2232

I started to do a few, ended up doing a lot, if not all. Hope this doesn't get in the way of your plans, @chrissimpkins!

- Replaced all `from ...py23 import *` with explicit name imports, or removed completely when possible.
- Replaced `tounicode()` with `tostr()`
- Changed all BytesIO ans StringIO imports to `from io import ...`, replaced all UnicodeIO with StringIO.
- Replaced all `unichr()` with `chr()`

I checked meticulously with flake8 for undefined names, so even untested code should still work.


chrissimpkins:
This is great! Thanks Just!

behdad:
This is great. Thank you Just!

The remaining bits should probably be moved to textTools module..

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -1099,10 +1100,10 @@ def fromstring(cls, string, readerClass=None, writerClass=None):
         return self
 
     def tostring(self, encoding=None):
-        if encoding is unicode or (
+        if encoding is str or (
```
khaledhosny, line 1103: This is such a weird line, I checked it before and I still don’t understand how it was supposed to work. Did callers actually pass Python 2 `unicode` function as the encoding?

Lib/fontTools/misc/xmlWriter.py
```diff
@@ -34,8 +34,8 @@ def __init__(self, fileOrPath, indentwhite=INDENT, idlefunc=None, encoding="utf_
 			self.totype = tobytes
 		except TypeError:
 			# This better not fail.
-			self.file.write(tounicode(''))
-			self.totype = tounicode
+			self.file.write(tostr(''))
```
khaledhosny, line 37: `tostr()` is probably redundant here.

Lib/fontTools/designspaceLib/__init__.py
behdad: Looks like that's how it was written.

Lib/fontTools/misc/xmlWriter.py
behdad: Correct.

Lib/fontTools/designspaceLib/__init__.py
justvanrossum: Shall I just remove the `encoding is str` bit? It was already weird for Python 2, and now it's even weirder.

Btw. the same thing is written in `fontTools.misc.etree`:

https://github.com/fonttools/fonttools/blob/a363bd75a19553ff5449a05fb7725d913232c36a/Lib/fontTools/misc/etree.py#L153

This should be changed in the same way. Any objections to removing it?

justvanrossum:
> The remaining bits should probably be moved to textTools module..

Agreed. Best in a different PR, though.

Lib/fontTools/designspaceLib/__init__.py
justvanrossum: That one in etree is weirder than I thought: it mimics the lxml etree API when we don't have lxml, but lxml's `ElementTree.write` doesn't support such a value for the "encoding" argument.

The case in designspaceLib may be caused by looking at the fontTools.misc.etree api, because that's where that call will eventually go to. Perhaps @anthrotype remembers? He implemented `DesignSpaceDocument().tostring()`.

Either way, I strongly suggest to get rid of the `encoding is unicode`/`encoding is str` cases.

justvanrossum: Even more: an encoding value of `"unicode"` does not work for `DesignSpaceDocument().tostring()` when lxml is installed. So either some fixing needs to be done, or that option needs to be removed.

khaledhosny: I think we should merge this PR as it does not change how this code was working, and address this issue in another PR.

Lib/fontTools/ttLib/ttCollection.py
```diff
@@ -104,7 +104,7 @@ def __getitem__(self, item):
 		return self.fonts[item]
 
 	def __setitem__(self, item, value):
-		self.fonts[item] = values
```
anthrotype, line 107: ouch! :)

Tests/pens/pointPen_test.py
```diff
@@ -256,10 +256,10 @@ def test_quad(self):
         pen.closePath()
         self.assertEqual("beginPath() addPoint((10, 10), segmentType='line') "
                          "addPoint((10, 20)) addPoint((20, 20)) "
-                         "addPoint((20, 10), segmentType=qcurve) endPath()",
+                         "addPoint((20, 10), segmentType='qcurve') endPath()",
                          repr(tpen))
 
-    def test_quad(self):
+    def test_quad2(self):
```
anthrotype, line 262: good catch

Lib/fontTools/designspaceLib/__init__.py
anthrotype: I believe I added this to mimic the lxml etree API, see "Serialising to Unicode strings" paragraph in https://lxml.de/parsing.html

or corresponding code in lxml/etree.pyx:
 https://github.com/lxml/lxml/blob/71667f9ac7694216ee8e793192bcd0993a0cdc66/src/lxml/etree.pyx#L3413-L3418

but Just said it doesn't actually work, need to investigate why. let's file a separate issue/PR for that

anthrotype (approved):
Thanks Just! It must have not been fun :)

You said you used flake8, we could probably add it to the `tox -e lint` step.

justvanrossum:
> You said you used flake8, we could probably add it to the tox -e lint step.

I specifically used it to catch NameErrors and `from xxx import *` warnings. It has shitloads of other things to complain about, so until we address those, perhaps not run it as part of any CI...

madig:
> It has shitloads of other things to complain about, so until we address those, perhaps not run it as part of any CI...

You can do it gradually by disabling all warnings and then gradually enabling vetted ones.

justvanrossum:
> You can do it gradually by disabling all warnings and then gradually enabling vetted ones.

Sounds good. Wouldn't it be better to start doing this after black has done its work?

madig:
I'm running flake8 right now and it works even without black if you disable style checks :) PR incoming.

PR #2246, opened 2021-03-30
Add flake8 to lint run

madig:
Love me some lints.

This highlights an annoyance: static analyzers stumble over otTables.py, which injects various classes into `globals()`. This seeps into all files that use these classes elsewhere. Maybe there's a way to declare these upfront so analyzers know what they are, rather than ignoring the error outright...


madig:
Plan: Take `ttLib.tables.otTables._buildClasses` and make it generate Python code instead, to get rid of magical dynamic class generation.

behdad:
> Plan: Take `ttLib.tables.otTables._buildClasses` and make it generate Python code instead, to get rid of magical dynamic class generation.

I suggest going a different direction: turn `otData.py` into Python code using metaclasses.

So, for example, the current:
```
        ('Script', [
                ('Offset', 'DefaultLangSys', None, None, 'Offset to DefaultLangSys table-from beginning of Script table-may be NULL'),
                ('uint16', 'LangSysCount', None, None, 'Number of LangSysRecords for this script-excluding the DefaultLangSys'),
                ('struct', 'LangSysRecord', 'LangSysCount', 0, 'Array of LangSysRecords-listed alphabetically by LangSysTag'),
        ]),
```
will become:
```
class Script : OTTable
    DefaultLangSys = Property(Offset32To(DefaultLangSys))
    LangSysCount = Property(uint16)
    LangSysRecord = Property(ArrayOf(LangSysRecord, "LangSysCount"))
```

But admittedly, it's a much bigger change. So maybe go ahead and generate code and we'll take it next step later.

PR #2249, opened 2021-03-30
Add missing CFF2 dict operators

punchcutter:
`LanguageGroup` and `ExpansionFactor` were missing from the CFF2 dict operators. Found while building CFF2 variable fonts and seeing that the `LanguageGroup` settings were disappearing.


behdad:
> `LanguageGroup` and `ExpansionFactor` were missing from the CFF2 dict operators.

What do these two do? I can't find any explanation for them.

Also, is it the case that default values are written out to XML for CFF properties? I expected otherwise.

punchcutter:
There's some explanation in https://adobe-type-tools.github.io/font-tech-notes/pdfs/5015.Type1_Supp.pdf

I wasn't sure about the default values being written for these. I just copied as-is from the CFF `privateDictOperators` which have default values that are not `None`.

anthrotype:
it seems that defaults are only used in cffLib while decompiling the table if an operator is missing, but are not taken into account when dumping toXML. This applies more generally, not specifically to CFF2. We may want to file a separate issue about that.

PR #2254, opened 2021-03-31
Update NEWS.rst

madig:
@justvanrossum Is there anything you want to get into the next release immediately?


behdad:
You might want to be more verbose about:
> - [ttLib] Remove .Format from Coverage, ClassDef, SingleSubst, LigatureSubst, AlternateSubst (#2238)
to note that other projects' saved TTX output will change.

Also, MultipleSubst got the same treatment, please add that.

anthrotype (approved):

PR #2261, opened 2021-04-07
Faster glyph coordinates

behdad:
This optimizes and simplifies GlyphCoordinates.

I see 30% speedup in total time of loading master ttfs, building gvar, and saving.

### DIFF
--- a/Lib/fontTools/misc/py23.py
+++ b/Lib/fontTools/misc/py23.py
@@ -71,7 +71,7 @@ def strjoin(iterable, joiner=""):
 
 
 def tobytes(s, encoding="ascii", errors="strict"):
-    if not isinstance(s, bytes):
+    if isinstance(s, str):
         return s.encode(encoding, errors)
     else:
         return s
--- a/Lib/fontTools/misc/py23.py
+++ b/Lib/fontTools/misc/py23.py
@@ -74,7 +74,7 @@ def tobytes(s, encoding="ascii", errors="strict"):
     if isinstance(s, str):
         return s.encode(encoding, errors)
     else:
-        return s
+        return bytes(s)
 
 
 def tounicode(s, encoding="ascii", errors="strict"):
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -2203,7 +2203,7 @@ def subset_glyphs(self, s):
 def remapComponentsFast(self, glyphidmap):
 	if not self.data or struct.unpack(">h", self.data[:2])[0] >= 0:
 		return	# Not composite
-	data = array.array("B", self.data)
+	data = self.data = bytearray(self.data)
 	i = 10
 	more = 1
 	while more:
@@ -2222,8 +2222,6 @@ def remapComponentsFast(self, glyphidmap):
 		elif flags & 0x0040: i += 4	# WE_HAVE_AN_X_AND_Y_SCALE
 		elif flags & 0x0080: i += 8	# WE_HAVE_A_TWO_BY_TWO
 		more = flags & 0x0020	# MORE_COMPONENTS
-
-	self.data = data.tobytes()
 
 @_add_method(ttLib.getTableClass('glyf'))
 def closure_glyphs(self, s):
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -2247,7 +2247,7 @@ def prune_pre_subset(self, font, options):
 		g = self[self.glyphOrder[0]]
 		# Yay, easy!
 		g.__dict__.clear()
-		g.data = ""
+		g.data = b''
 	return True
 
 @_add_method(ttLib.getTableClass('glyf'))
@@ -2262,7 +2262,7 @@ def subset_glyphs(self, s):
 	Glyph = ttLib.getTableModule('glyf').Glyph
 	for g in s.glyphs_emptied:
 		self.glyphs[g] = Glyph()
-		self.glyphs[g].data = ''
+		self.glyphs[g].data = b''
 	self.glyphOrder = [g for g in self.glyphOrder if g in s.glyphs or g in s.glyphs_emptied]
 	# Don't drop empty 'glyf' tables, otherwise 'loca' doesn't get subset.
 	return True
--- a/Lib/fontTools/ttLib/tables/TupleVariation.py
+++ b/Lib/fontTools/ttLib/tables/TupleVariation.py
@@ -633,15 +633,17 @@ def compileTupleVariationStore(variations, pointCount,
 			pointSetCount[usedPoints] += 1
 			pointSitu.append(usedPoints)
 
+		compiledPoints = {pointSet:TupleVariation.compilePoints(pointSet, n)
+				  for pointSet in pointSetCount}
+
 		# Find point-set which saves most bytes.
 		def key(pn):
 			pointSet = pn[0]
 			count = pn[1]
-			data = TupleVariation.compilePoints(pointSet, n)
-			return len(data) * (count - 1)
+			return len(compiledPoints[pointSet]) * (count - 1)
 		sharedPoints = max(pointSetCount.items(), key=key)[0]
 
-		data.append(TupleVariation.compilePoints(sharedPoints, n))
+		data.append(compiledPoints[sharedPoints])
 		tupleVariationCount |= TUPLES_SHARE_POINT_NUMBERS
 
 		# Only allowed to reuse one set.
--- a/Lib/fontTools/ttLib/tables/TupleVariation.py
+++ b/Lib/fontTools/ttLib/tables/TupleVariation.py
@@ -48,6 +48,8 @@ class TupleVariation(object):
 		return self.coordinates == other.coordinates and self.axes == other.axes
 
 	def getUsedPoints(self):
+		if None not in self.coordinates:
+			return frozenset()
 		return frozenset([i for i,p in enumerate(self.coordinates) if p is not None])
 
 	def hasImpact(self):
@@ -129,7 +131,7 @@ class TupleVariation(object):
 		auxData = []
 
 		if pointData is None:
-			points = self.getUsedPoints() if None in self.coordinates else frozenset()
+			points = self.getUsedPoints()
 			pointData = self.compilePoints(points)
 
 		coord = self.compileCoord(axisTags)
@@ -636,7 +638,7 @@ def compileTupleVariationStore(variations, pointCount,
 	# Collect and count all pointSets
 	pointSetCount = defaultdict(int)
 	for v in variations:
-		points = v.getUsedPoints() if None in v.coordinates else frozenset()
+		points = v.getUsedPoints()
 		pointSetCount[points] += 1
 		pointDatas.append(points)
 
--- a/Lib/fontTools/ttLib/tables/TupleVariation.py
+++ b/Lib/fontTools/ttLib/tables/TupleVariation.py
@@ -130,7 +130,7 @@ class TupleVariation(object):
 	def compile(self, axisTags, sharedCoordIndices={}, sharedPoints=None):
 		tupleData = []
 
-		assert all(tag in axisTags for tag in self.axes.keys()), ("Unknown axis tag found.", self.axes.keys(), axisTags)
+		assert set(self.axes.keys()) <= set(axisTags), ("Unknown axis tag found.", self.axes.keys(), axisTags)
 
 		coord = self.compileCoord(axisTags)
 		flags = sharedCoordIndices.get(coord)
--- a/Lib/fontTools/ttLib/tables/TupleVariation.py
+++ b/Lib/fontTools/ttLib/tables/TupleVariation.py
@@ -38,7 +38,7 @@ class TupleVariation(object):
 
 	def __init__(self, axes, coordinates):
 		self.axes = axes.copy()
-		self.coordinates = coordinates[:]
+		self.coordinates = list(coordinates)
 
 	def __repr__(self):
 		axes = ",".join(sorted(["%s=%s" % (name, value) for (name, value) in self.axes.items()]))
--- a/Lib/fontTools/ttLib/tables/TupleVariation.py
+++ b/Lib/fontTools/ttLib/tables/TupleVariation.py
@@ -158,9 +158,13 @@ class TupleVariation(object):
 
 	def compileCoord(self, axisTags):
 		result = bytearray()
+		axes = self.axes
 		for axis in axisTags:
-			_minValue, value, _maxValue = self.axes.get(axis, (0.0, 0.0, 0.0))
-			result.extend(struct.pack(">h", fl2fi(value, 14)))
+			triple = axes.get(axis)
+			if triple is None:
+				result.extend(b'\0\0')
+			else:
+				result.extend(struct.pack(">h", fl2fi(triple[1], 14)))
 		return bytes(result)
 
 	def compileIntermediateCoord(self, axisTags):
--- a/Lib/fontTools/ttLib/tables/TupleVariation.py
+++ b/Lib/fontTools/ttLib/tables/TupleVariation.py
@@ -157,11 +157,11 @@ class TupleVariation(object):
 		return (b''.join(tupleData), auxData, usesSharedPoints)
 
 	def compileCoord(self, axisTags):
-		result = []
+		result = bytearray()
 		for axis in axisTags:
 			_minValue, value, _maxValue = self.axes.get(axis, (0.0, 0.0, 0.0))
-			result.append(struct.pack(">h", fl2fi(value, 14)))
-		return b''.join(result)
+			result.extend(struct.pack(">h", fl2fi(value, 14)))
+		return bytes(result)
 
 	def compileIntermediateCoord(self, axisTags):
 		needed = False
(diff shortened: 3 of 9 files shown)


madig:
How did you benchmark? I see no definitive speedup for fontmake :(

I did `py-spy record --format speedscope --output /tmp/fontmake.speedscope -s -n -- fontmake --verbose WARNING -m NotoSans-MM.designspace -o variable --output-dir /tmp`.

behdad:
> How did you benchmark? I see no definitive speedup for fontmake :(

That is indeed because gvar generation is not a big chunk of that...  My testing is with:
```
$ time fonttools varLib --disable-iup master_ttf_interpolatable/NotoSans-MM.designspace -x GSUB -x GPOS -x GDEF
```
and I get 10s before, 7s after.

madig:
Hm, I guess it's a start and it removes some lines in aggregate :)

behdad:
> and I get 10s before, 7s after.

I'm down from 10s before to 6s after.

behdad:
This is ready for review I think.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -1443,9 +1441,6 @@ def __init__(self, iterable=[]):
 	def array(self):
 		return self._a
 
-	def isFloat(self):
```
anthrotype, line 1443: we may as well keep it since it wasn't prefixed with an underscore and tools in the wild may be accessing that method for whatever reason. bonus point if we add a deprecation warning as well

Lib/fontTools/misc/py23.py
```diff
@@ -71,7 +71,7 @@ def strjoin(iterable, joiner=""):
 
 
 def tobytes(s, encoding="ascii", errors="strict"):
```
anthrotype, line 73: i'd expected a method called `tobytes` to always return a `bytes` object, not sometimes a bytes other times a bytearray. Maybe you should add another `if isinstance(s, bytearray): bytes(s)`

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -380,8 +380,7 @@ def encodeDeltaRunAsBytes_(deltas, offset, bytearr):
 			runLength += 1
 		assert runLength >= 1 and runLength <= 64
 		bytearr.append(runLength - 1)
-		for i in range(offset, pos):
-			bytearr.extend(struct.pack('b', deltas[i]))
+		bytearr.extend(array.array('b', deltas[offset:pos]))
```
anthrotype: doesn't `bytearray.extend(deltas[offset:pos])` work already?

anthrotype:
Thanks Behdad for working on this!

Lib/fontTools/subset/__init__.py
```diff
@@ -2459,7 +2459,7 @@ class UnknownOptionError(OptionError): pass
 				'EBSC', 'SVG', 'PCLT', 'LTSH']
 	_drop_tables_default += ['Feat', 'Glat', 'Gloc', 'Silf', 'Sill']  # Graphite
 	_no_subset_tables_default = ['avar', 'fvar',
-				     'gasp', 'head', 'hhea', 'maxp',
+				     'gasp', 'head', 'hhea', 'maxp', 'meta',
```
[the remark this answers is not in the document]
behdad: Oh. That wasn't unintended. Thanks for catching.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -1705,12 +1705,18 @@ def __imul__(self, other):
...
 		"""
-		if isinstance(other, Number):
-			other = (other, other)
 		if isinstance(other, tuple):
 			assert len(other) ==  2
 			self.scale(other)
 			return self
+		if isinstance(other, Number):
+			if other == 1.0: return self
+			if other == int(other): other = int(other)
```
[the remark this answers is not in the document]
behdad: This was subsequently removed since I switched it to always keep float. And int-float multiplication is same speed as float-float.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -1430,11 +1430,13 @@ def __ne__(self, other):
 		result = self.__eq__(other)
 		return result if result is NotImplemented else not result
 
+def _i(v):
+	return int(v) if int(v) == v else v
```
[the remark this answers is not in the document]
behdad: Ah didn't know about that. Great. Thanks. Is twice faster indeed.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
behdad: The type was rather internal that's why I have no problem removing it. Any code using it needs to be updated probably. But sure, adding back with warning.

Is ugly that we'll keep forever.

Lib/fontTools/ttLib/tables/TupleVariation.py
behdad: No, because deltas are signed byte. Bytearray accepts unsigned byte only.

anthrotype: gotcha

Lib/fontTools/ttLib/tables/_g_l_y_f.py
anthrotype: actually, it's fine for me if you want to remove it. I doubt anybody uses that

Lib/fontTools/misc/py23.py
behdad: I generally agree. But I don't see the point of an extra copy.  If we were to do that though, we can just always make the default path return `bytes(s)`, which like Tuple type, is a no-op if input is bytes already.

So, my changes in this branch make some internal data types to be left as bytearray instead of bytes after, eg. subsetting. I'm treating the two types interchangable for readonly use purposes. I feel like that's how they were designed to be.

And I removed tobytes() invocations where I could audit that we never get a str there. I think my branch doesn't depend on this change per se. I made it because it looks correct; since the `.encode()` method is a property of `str` only. There has been an assumption that what's passed to this function is either `bytes` or `str`. And under that assumption this change is still correct.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -1705,12 +1705,18 @@ def __imul__(self, other):
...
 		GlyphCoordinates([(2, 4)])
 		"""
-		if isinstance(other, Number):
-			other = (other, other)
 		if isinstance(other, tuple):
 			assert len(other) ==  2
 			self.scale(other)
 			return self
+		if isinstance(other, Number):
+			if other == 1.0: return self
```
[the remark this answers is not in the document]
behdad: Fixed it.

Lib/fontTools/misc/py23.py
behdad: Okay, I made it always return bytes.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -1442,12 +1442,12 @@ def array(self):
 		return self._a
 
 	def isFloat(self):
-		return self._a.typecode == 'd'
+		return self._a.typecode == 'f'
```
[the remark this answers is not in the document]
behdad: Humm. Do you remember how that issue was brought up to your attention? With otRound you can't observe that difference anymore. Is just a waste to use doubles there but I'm fine with switching back.

behdad:
Thanks @anthrotype! I addressed all issues except for the GlyphCoordinates float vs double typecode. Let me know what you decide and I'll adjust.

Lib/fontTools/misc/py23.py
anthrotype: > I'm treating the two types interchangable for readonly use purposes.

that's ok.

> I made it because it looks correct; since the .encode() method is a property of str only. 

true. Then for the same logic, we should change tounicode as well..

> make the default path return bytes(s)

yeah I feel that would be better, but i'm concerned for the extra copy.
Imagine later we decide to add type hints to these functions, what would the type of the return value be? Depends on the type of the input. I guess with `typing.overload` one could describe that. Though it's just weid that if one passes a ramdon object that isn't `str` to tobytes the latter returns it back as it was without complaining.

behdad: I've made it return `bytes(s)` for now.

If typing, we should make it a generic that passes bytes-like objects intact, where bytes-like is defined as implementing the buffer protocol with item type of unsigned byte.

behdad: > Then for the same logic, we should change tounicode as well..

But both `bytes` and `bytearray` provide `.decode()`.

behdad:
Down from 10s before to 5s after. I hope I can stop now. :))

madig:
Compiling a Noto Sans LGC VF now takes 40s instead of 45s on my machine :)

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -599,12 +600,11 @@ def decompileSharedTuples(axisTags, sharedTupleCount, data, offset):
 
 
 def compileSharedTuples(axisTags, variations):
-	coordCount = {}
+	coordCount = defaultdict(int)
```
anthrotype, line 611: there's also `collections.Counter`, not sure if it's any faster https://docs.python.org/3/library/collections.html#counter-objects

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -48,11 +48,7 @@ def __eq__(self, other):
 		return self.coordinates == other.coordinates and self.axes == other.axes
 
 	def getUsedPoints(self):
-		result = set()
-		for i, point in enumerate(self.coordinates):
-			if point is not None:
-				result.add(i)
-		return result
+		return set([i for i,p in enumerate(self.coordinates) if p is not None])
```
anthrotype: you could remove the inner list comprehension `set(i for i, p ...)` works the same, I'd expect it to also be a bit faster

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -48,7 +48,7 @@ def __eq__(self, other):
 		return self.coordinates == other.coordinates and self.axes == other.axes
 
 	def getUsedPoints(self):
-		return set([i for i,p in enumerate(self.coordinates) if p is not None])
+		return frozenset([i for i,p in enumerate(self.coordinates) if p is not None])
```
anthrotype, line 54: same for `frozenset(i for i, p ...)` it also takes a generator expression

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -614,67 +614,50 @@ def compileTupleVariationStore(variations, pointCount,
...
-	# 3K for BuffaloGalRegular, and 10K for JamRegular. However, in
-	# the latter case the delta arrays will become larger, but I
-	# haven't yet measured by how much. From gut feeling (which may be
-	# wrong), the optimum is to share some but not all points;
-	# however, then we would need to try all combinations.
-	#
-	# For the time being, we try two variants and then pick the better one:
-	# (a) each tuple supplies its own private set of points;
-	# (b) all tuples refer to a shared set of points, which consists of
-	#     "every control point in the glyph that has explicit deltas".
```
anthrotype, line 632: I see you removed this long block of comments, as I guess no longer applies to the code you modified. I don't fully grasp the consequences of your change. Is it going to change the way TupleVariations gets compiled to binary? Could it affect the size of the table? I should probably refresh my memory by rereading the spec. Would be nice if you added some words (can be shorter that the verbose comment above) about how this is different from before.

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -646,16 +651,12 @@ def key(pn):
...
 		tupleVariationCount |= TUPLES_SHARE_POINT_NUMBERS
 
-		# Only allowed to reuse one set.
-		pointSitu = [sharedPoints if sharedPoints == points else None
-			     for points in pointSitu]
-
-		del pointSetCount
-	else:
-		pointSitu = [None] * len(variations)
+	# NotImplemented means "use shared points"
```
anthrotype, line 661: using NotImplemented like that looks a bit weird to me. I think it would be clearer and need no extra comment if you added an additional `useSharedPoints: bool` parameter to compile method, and use None `if points != sharedPoints`. But take my comment as optional.

anthrotype (approved):

Lib/fontTools/ttLib/tables/_g_l_y_f.py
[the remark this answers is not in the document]
anthrotype: If you flip the sign of `afloat` in the `test_double_precision_float` then it will fail, even if we use otRound as we do now:

https://github.com/fonttools/fonttools/blob/cbb33d8aa9e21f01ba72700d5748d16c09e0342b/Tests/ttLib/tables/_g_l_y_f_test.py#L165-L173

```
E       assert -242 == -243
E        +  where -243 = otRound(-242.50000000000003)
```

Besides, aren't we giving up some precision when we do arithmetic operations on floats as opposed to doubles? Or you think it'd negligible? What's the measurable effect in terms of memory/speed of using`'f'` instead of `'d'`?

Lib/fontTools/ttLib/tables/TupleVariation.py
behdad: Thanks. Used it in one place.

Lib/fontTools/ttLib/tables/TupleVariation.py
behdad: I know. And for years I've measured and the generator version is slower. Is till slightly true today :(.

Lib/fontTools/ttLib/tables/TupleVariation.py
behdad: same. is slower:
```
In [2]: timeit set([_ for _ in l])
2.12 µs ± 6.44 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [3]: timeit set(_ for _ in l)
2.76 µs ± 10.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [4]: timeit set(l)
821 ns ± 2.35 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

Frozenset is very marginally slower than set; but that's because it does hashing.

Lib/fontTools/ttLib/tables/TupleVariation.py
behdad: I'll add a comment.

> Is it going to change the way TupleVariations gets compiled to binary? Could it affect the size of the table?

Fore pretty much all real cases it doesn't make a difference. For very esoteric designspace situations the new code can use the sharedPoints whereas the old code couldn't.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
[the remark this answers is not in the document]
behdad: Okay I switch back to `'d`'. The effect is 8 bytes instead of 4, for every coordinate. Absolutely negligible in Python land. :)

In terms of true precision, `float` is definitely more than enough. It has 24 bits of mantissa, which means that for numbers in the 16k range (which glyph coordinates are constrained to), it has 9 bits of sub-font-unit precision.

The problem with that test is that the number you are starting with (-242.50000000000003) has more bits in it that fits in a float. The question is why does that number matter... So my question, if you remember, is how was that bug diagnosed as a bug to begin with? Where did it come from?

Does Python `float` always be double? Or is there a way to get the internal size so we match it?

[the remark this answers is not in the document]
anthrotype: i believe python `float` is always double precision. I don't remember how the issue came up I'm afraid.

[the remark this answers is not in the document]
behdad: Np. Reverted.

Lib/fontTools/ttLib/tables/TupleVariation.py
behdad: Since the caller already has to compute the points for each TupleVariation to figure out which pointset to share, I made the `compile()` method get compiled `pointData` directly now. So there won't be any `if points != sharedPoints` possible inside since I wanted to avoid calling `getUsedPoints()` again.

I can use another singleton value if you prefer. Or I can use `b''`; still would need a comment because it's going to be a special-case.

behdad: > Or I can use `b''`; still would need a comment because it's going to be a special-case.

Doing that now.

Lib/fontTools/ttLib/tables/TupleVariation.py
behdad: It might make iup-optimized fonts smaller though. Let me test with NotoSans.

behdad: Honestly I'm struggling to find what to comment there. The old code was adhoc and wrong. The new code is exact: it collects all pointSets used and picks the one to share that saves most bytes...

As a headsup, to truly do things properly we should move the IUP optimizer into this compile code. I'll head in that direction...

behdad: It does:
```
1264516 bytes 15 seconds this branch
1270936 bytes 21 seconds main branch
```

behdad: Added a comment.

behdad:
Okay! I believe I've addressed all review comments. Thanks @anthrotype!

anthrotype (approved):
LGTM

PR #2266, opened 2021-04-13
[varLib/glyf] Rework phantom-points calculations

behdad:


behdad:
Calls to `ttFont.__getitem__` go down from over 200,000, to 70k so far.

behdad:
And following timing goes from 5s to 4.6s:
```
$ time fonttools varLib  master_ttf_interpolatable/NotoSans-MM.designspace -x GSUB -x GPOS -x GDEF --disable-iup
```

behdad:
> Calls to `ttFont.__getitem__` go down from over 200,000, to 70k so far.

And now down to just 170.

This is ready for review now.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -254,57 +260,41 @@ def __len__(self):
...
+
 		return [
 			(leftSideX, 0),
 			(rightSideX, 0),
 			(0, topSideY),
 			(0, bottomSideY),
 		]
 
-	def getCoordinatesAndControls(self, glyphName, ttFont, defaultVerticalOrigin=None):
+	def getCoordinatesAndControls(self, glyphName, hMetrics, vMetrics=None):
```
justvanrossum, line 297: I have external code that uses this method. While it's easy enough to adapt (and I'm willing to), this seems to be a public method, and this change is not b/w compatible.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -236,6 +236,12 @@ def has_key(self, glyphName):
 
 	__contains__ = has_key
 
+	def get(self, glyphName):
```
anthrotype, line 239: add a `default=None` parameter maybe, for consistency with dict api

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -254,57 +260,41 @@ def __len__(self):
...
-			# without vmtx, use ascent as vertical origin and UPEM as vertical advance
-			# like HarfBuzz does
-			verticalAdvanceWidth = ttFont["head"].unitsPerEm
-			if "hhea" in ttFont:
-				topSideY = ttFont["hhea"].ascent
-			else:
-				# sparse masters may not contain an hhea table; use the ascent
-				# of the default master as the vertical origin
-				if defaultVerticalOrigin is not None:
-					topSideY = defaultVerticalOrigin
```
anthrotype, line 291: so you decided to remove this `defaultVerticalOrigin` parameter, why? what was wrong with it?

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -323,14 +323,11 @@ def limitTupleVariationAxisRange(var, axisTag, axisRange):
         return [var, newVar]
 
 
-def instantiateGvarGlyph(varfont, glyphname, axisLimits, optimize=True):
-    glyf = varfont["glyf"]
-    coordinates, ctrl = glyf.getCoordinatesAndControls(glyphname, varfont)
+def _instantiateGvarGlyph(glyphname, glyf, gvar, hMetrics, vMetrics, axisLimits, optimize=True):
```
anthrotype, line 326: in this case I don't mind too much that the signature is different (api break), but I guess it would still be nice to export a non `_`-prefixed public `instantiateGvarGlyph` method. Imagine one wants to instantiate different glyphs and different positions for whatever reason

Lib/fontTools/ttLib/tables/_g_l_y_f.py
anthrotype: would it be possible to add a new method and keep this one but with a deprecation warning? I know it's difficult to come up with a good alternative name. Maybe just `getCoordinates`?

behdad: Okay, I'll put the old one back. If doing that I might as well leave the old `getPhantomPoints` alone...

For names.. I don't like just `getCoordinates` given the phantomPoint dance this does. I'm happy to add `_getPhantomPoints` and `_getCoordinatesAndControls` to mark them internal so we can adapt as needed.

behdad: > I have external code that uses this method.

Thanks @justvanrossum. I'm curious how you use it. Do you use the control data as well? The phantom points?

behdad: I suppose I have to leave alone the old `setCoordinates` as well then.

justvanrossum: @behdad: my code is pretty much a copy of this fragment in mutator.py:

https://github.com/fonttools/fonttools/blob/3cfc87be711756baede6059d088cc2a2ebd20605/Lib/fontTools/varLib/mutator.py#L196-L207

I need to instantiate an individual composite glyph. I don't use the controls, and strip the phantom points.

I can't use instancer's `instantiateGvarGlyph()` as this modifies the `glyf` table: all I need at this point are the instantiated component offsets.

A public function doing something like "give me the instantiated coordinates as a GlyphCoordinates array" would be helpful for my case.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
behdad: If the font does not have `vmtx`, we were synthesizing glyph vertical origin (and advance-width). However, this is not necessary. If you look at either mutator or instancer, the deltas for vertical phantom points are simply ignored if font doesn't have `vmtx` table.

At rendering time, if font doesn't have `vmtx`, HarfBuzz, etc synthesize glyph vertical origin, from the intsance's `hhea`, which is already interpolated. Included deltas for vertical origin here will either be ignored (in which case we don't need them), or worse, applied on top of the synthesized vertical origin, which would be wrong.

behdad:
Okay. Thanks everyone. I put back the old names. Didn't quite deprecate them. Am not sure where to go from here.

I think these were added as public API a bit hastily. We can deprecate and remove, while we figure out what to expose as public API now.

behdad:
I think this can be merged if we open a tracker issue for the deprecation / new API part.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
madig: I also have a copy of `instantiateGvarGlyph()` in a tool to instantiate a glyph at a certain point without changing the font :O

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -396,14 +386,55 @@ def setCoordinates(self, glyphName, coord, ttFont):
...
+		if vMetrics is not None:
 			verticalAdvanceWidth = otRound(topSideY - bottomSideY)
 			if verticalAdvanceWidth < 0:  # unlikely but do the same as horizontal
 				verticalAdvanceWidth = 0
 			topSideBearing = otRound(topSideY - glyph.yMax)
-			ttFont["vmtx"].metrics[glyphName] = verticalAdvanceWidth, topSideBearing
+			vMetrics[glyphName] = verticalAdvanceWidth, topSideBearing
+
+
+	# Deprecated
```
madig, line 399: Wait, wait, wait, why are these methods deprecated? Because they shouldn't be used publicly?

Lib/fontTools/ttLib/tables/_g_l_y_f.py
behdad: Understood. Let's add a designed API for instantiation; ideally as methods of an instancer object.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
madig: Oh I see you just said something about that.

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -362,12 +359,22 @@ def instantiateGvarGlyph(varfont, glyphname, axisLimits, optimize=True):
...
             var.optimize(coordinates, endPts, isComposite)
 
+def instantiateGvarGlyph(varfont, glyphname, axisLimits, optimize=True):
+    """Remove?
+    https://github.com/fonttools/fonttools/pull/2266"""
+    gvar = varfont["gvar"]
+    glyf = varfont["glyf"]
+    hMetrics = varfont['hmtx'].metrics
+    vMetrics = getattr(varfont.get('vmtx'), 'metrics', None)
+    _instantiateGvarGlyph(glyphname, glyf, gvar, hMetrics, vMetrics, axisLimits, optimize=True)
```
madig, line 369:
```suggestion
    _instantiateGvarGlyph(glyphname, glyf, gvar, hMetrics, vMetrics, axisLimits, optimize=optimize)
```

Lib/fontTools/ttLib/tables/_g_l_y_f.py
madig: Cosimo said instancer is made for font mutation. Changing that would probably be an involved refactoring.

behdad: As it happens I've been looking into that code. I like to see L4 subsetting happening. To me, it's a waste if we don't use the same code to back GUI tools like FontGoggle and others.

Can you open an issue please, expanding on your use-case, and hopefully we get Just's as well. Then we can design a new module to lay under or replace `mutator` and `instancer`.

madig (approved):

anthrotype (approved):

Truthofmanyis1 (commented):
Lib/fontTools/misc/arrayTools.py

Issue #2272, opened 2021-04-21
[subset] Drops variation Device tables if `--drop-hinting`; ouch!


behdad:
We discovered this in:
https://github.com/harfbuzz/harfbuzz/pull/2957

Basically, code from before variations always dropped Device tables if no-hinting was requested. And we reused the Device table mechanism for variations. We never updated the code!

In the subsetter code just search for `Drop device tables`. There's ~5 places.

anthrotype:
ouch, so basically if you subset an hinted variable font and request to --no-hinting we also strip out all the Device tables, including those with DeltaFormat = 0x8000 that contain indices into ItemVarStore

behdad:
Correct. I'm on it.

behdad:
Is uglier than I like it to be. :)

Issue #471, opened 2016-01-19
[feaLib] Stripping away ValueRecords with ValueFormat zero


brawer:
When building SinglePos tables moved to otlLib, feaLib started stripping off SinglePos entries whose ValueRecord.ValueFormat is zero. However, `makeotf` emits such entries in its output, so we now have a discrepancy to makeotf. Question: Are we doing a valid optimization here, or are they semantically different? What if SinglePos lookup gets chained from another ChainContextPos lookup?

Sample input:

```
feature test {
    pos zero 0;
    pos four 500;
} test;
```

makeotf output:

```
      <!-- LookupCount=1 -->
      <Lookup index="0">
        <!-- LookupType=1 -->
        <LookupFlag value="0"/>
        <!-- SubTableCount=2 -->
        <SinglePos index="0" Format="1">
          <Coverage>
            <Glyph value="zero"/>
          </Coverage>
          <ValueFormat value="0"/>
        </SinglePos>
        <SinglePos index="1" Format="1">
          <Coverage>
            <Glyph value="four"/>
          </Coverage>
          <ValueFormat value="4"/>
          <Value XAdvance="500"/>
        </SinglePos>
      </Lookup>
```

feaLib output:

```
      <!-- LookupCount=1 -->
      <Lookup index="0">
        <!-- LookupType=1 -->
        <LookupFlag value="0"/>
        <!-- SubTableCount=1 -->
        <SinglePos index="0" Format="1">
          <Coverage>
            <Glyph value="four"/>
          </Coverage>
          <ValueFormat value="4"/>
          <Value XAdvance="500"/>
        </SinglePos>
      </Lookup>
```

behdad:
This is an unsafe optimization *when and only when* other subtables might be appended after this one.  In other words, the semantic difference is that, when the entry is available (albeit all zero), it will stop searching forward in the subtables that follow.

It does not have any effect in chaining/context.

behdad:
> This is an unsafe optimization when and only when other subtables might be appended after this one.

For this reason, I feel more comfortable if you don't drop it if it was specifically written in the input.

brawer:
After the last change, there still is ine difference in the XML: The fontTools output contains an empty  `<Value/>` entry while the makeotf output does not have this. Probably an artifact, but I’ll fix it nevertheless.

makeotf:
```
      <!-- LookupCount=1 -->
      <Lookup index="0">
        <!-- LookupType=1 -->
        <LookupFlag value="0"/>
[...]

PR #2275, opened 2021-04-21
[subset] Fix hint-dropping

behdad:
Fixes https://github.com/fonttools/fonttools/issues/2272

The 4 failing tests are adjacent and because of the second commit.  2 of them just need expectations updated. The other two, that are producing empty lookup:
```
+      <Lookup index="0" empty="1"/>
```
need a fix in `feaLib` (I suppose) to prune empty lookups. I couldn't understand why that's not happening.

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -14,6 +14,7 @@ import struct
 import array
 import logging
 from collections import Counter, defaultdict
+from functools import reduce
 from types import MethodType
 
 __usage__ = "pyftsubset font-file [glyph...] [--option=value]..."
@@ -527,6 +528,17 @@ def subset_glyphs(self, s):
 	else:
 		assert 0, "unknown format: %s" % self.Format
 
+@_add_method(otTables.Device)
+def is_hinting(self):
+	return self.DeltaFormat in (1,2,3)
+
+@_add_method(otTables.ValueRecord)
+def prune_hints(self):
+	for name in ['XPlaDevice', 'YPlaDevice', 'XAdvDevice', 'YAdvDevice']:
+		v = getattr(self, name, None)
+		if v is not None and v.is_hinting():
+			delattr(self, name)
+
 @_add_method(otTables.SinglePos)
 def subset_glyphs(self, s):
 	if self.Format == 1:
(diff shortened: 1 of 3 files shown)


behdad:
Adding Simon since he knows feaLib.

behdad:
Fixed two expectations. The two remaining failures is where I like to get others' input on. Thanks.

Lib/fontTools/otlLib/builder.py
```diff
@@ -1359,10 +1359,10 @@ def build(self):
                     builder.addSubtableBreak()
                 continue
             valFormat1, valFormat2 = 0, 0
-            if value1:
-                valFormat1 = value1.getFormat()
-            if value2:
-                valFormat2 = value2.getFormat()
+            valFormat1 = value1.getEffectiveFormat() if value1 is not None else 0
+            valFormat2 = value1.getEffectiveFormat() if value2 is not None else 0
```
garretrieger, line 1363: value2?

behdad: Shhhhhhhhhh!

anthrotype:
> The other two, that are producing empty lookup <Lookup index="0" empty="1"/> need a fix in feaLib (I suppose) to prune empty lookups. I couldn't understand why that's not happening.

That's because now you're skipping adding pairpos when both valueFormat1 and 2 are 0, and you end up with an empty list of subtables for the specific test file containing all zero values; then the feaLib `buildLookups_` which calls the `otlLib.PairPosBuilder.build()` method does not check whether the returned object is None (which happens when there are no subtables), but unconditionally adds it to the lookup list, so you end up with an empty Lookup.

I have this patch below to make feaLib `buildLookups_` check for None after lookup.build(), which makes those two ZeroValue_PairPos tests generate no GSUB at all, since there's no lookups nor features in it.

```diff
diff --git a/Lib/fontTools/feaLib/builder.py b/Lib/fontTools/feaLib/builder.py
index 4a7d9575..8649ed05 100644
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -804,21 +804,24 @@ class Builder(object):
         assert tag in ("GPOS", "GSUB"), tag
         for lookup in self.lookups_:
             lookup.lookup_index = None
-        lookups = []
+        otLookups = []
         for lookup in self.lookups_:
             if lookup.table != tag:
                 continue
-            lookup.lookup_index = len(lookups)
             self.lookup_locations[tag][str(lookup.lookup_index)] = LookupDebugInfo(
                 location=str(lookup.location),
                 name=self.get_lookup_name_(lookup),
                 feature=None,
             )
-            lookups.append(lookup)
-        try:
-            otLookups = [l.build() for l in lookups]
-        except OpenTypeLibError as e:
-            raise FeatureLibError(str(e), e.location) from e
+
+            try:
+                otLookup = lookup.build()
+            except OpenTypeLibError as e:
+                raise FeatureLibError(str(e), e.location) from e
+
+            if otLookup is not None:
+                lookup.lookup_index = len(otLookups)
+                otLookups.append(otLookup)
         return otLookups

     def makeTable(self, tag):
```

But then a bunch of other tests then start to fail as soon as you do that. I'm not sure what's going on, but maybe it was a deliberate to never drop lookups even when they do nothing.

BTW, are we sure it's safe to remove PairPos subtable when it contains only zero values?
This reminded me of a similar optimization where feaLib was removing no-op SinglePos tables with ValueFormat=0 which was deemed unsafe: https://github.com/fonttools/fonttools/issues/471

> This is an unsafe optimization when and only when other subtables might be appended after this one. In other words, the semantic difference is that, when the entry is available (albeit all zero), it will stop searching forward in the subtables that follow.

behdad:
Thanks.

> But then a bunch of other tests then start to fail as soon as you do that. I'm not sure what's going on, but maybe it was a deliberate to never drop lookups even when they do nothing.

I don't believe so. When Sascha wrote feaLib, he just made it match AFDKO, even if it was suboptimal. We should feel free to do what makes sense.

Please advise how to move forward. I can split this into two branches, one of the for fixing hint-dropping, which can be merged, another for improving lookup compiler efficiency.

behdad:
Done. Will send next PR after merging this.

anthrotype (approved):

Issue #2276, opened 2021-04-22
feaLib should not fail in getAlternateGlyphs for empty lookups


kontur:
Hey,

following case: I have two feature files, `uprights.fea` and `italics.fea` being compiled. Both files include a `shared.fea` which contains a lot of shared feature code. However, there is a small difference in uprights and italics. 

To get around it, my `uprights.fea` has this:

```
lookup ONLY_UPRIGHTS {
  sub GlyphOnlyExistsInUprights by GlyphOnlyExistsInUprights.alt;
} ONLY_UPRIGHTS;
```

and in `italics.fea` has this:

```
lookup ONLY_UPRIGHTS {
  # pass
} ONLY_UPRIGHTS;
```

and `shared.fea` uses the lookup.

This will fail with a (partial) stacktrace of:
```
  File "/Users/johannes/.pyenv/versions/3.7.6/envs/mrhandy/lib/python3.7/site-packages/ufo2ft/featureWriters/markFeatureWriter.py", line 859, in _getIndicGlyphs
    gsub = self.compileGSUB()
  File "/Users/johannes/.pyenv/versions/3.7.6/envs/mrhandy/lib/python3.7/site-packages/ufo2ft/featureWriters/baseFeatureWriter.py", line 307, in compileGSUB
    gsub = compileGSUB(self.context.feaFile, glyphOrder)
  File "/Users/johannes/.pyenv/versions/3.7.6/envs/mrhandy/lib/python3.7/site-packages/ufo2ft/util.py", line 227, in compileGSUB
    addOpenTypeFeatures(font, featureFile, tables={"GSUB"})
  File "/Users/johannes/.pyenv/versions/3.7.6/envs/mrhandy/lib/python3.7/site-packages/fontTools/feaLib/builder.py", line 60, in addOpenTypeFeatures
    builder.build(tables=tables, debug=debug)
  File "/Users/johannes/.pyenv/versions/3.7.6/envs/mrhandy/lib/python3.7/site-packages/fontTools/feaLib/builder.py", line 184, in build
    self.build_feature_aalt_()
  File "/Users/johannes/.pyenv/versions/3.7.6/envs/mrhandy/lib/python3.7/site-packages/fontTools/feaLib/builder.py", line 286, in build_feature_aalt_
    for glyph, alts in lookup.getAlternateGlyphs().items():
AttributeError: 'NoneType' object has no attribute 'getAlternateGlyphs'
```

[...]

PR #2277, opened 2021-04-22
fix AttributeError while generating aalt feature with empty lookup

anthrotype:
mean to fix https://github.com/fonttools/fonttools/issues/2276

still trying to reproduce

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -1005,7 +1005,8 @@ class Builder(object):
         assert lookup_name in self.named_lookups_, lookup_name
         self.cur_lookup_ = None
         lookup = self.named_lookups_[lookup_name]
-        self.add_lookup_to_feature_(lookup, self.cur_feature_name_)
+        if lookup is not None:  # skip empty named lookup
+            self.add_lookup_to_feature_(lookup, self.cur_feature_name_)
 
     def set_font_revision(self, location, revision):
         self.fontRevision_ = revision
--- a/Tests/feaLib/builder_test.py
+++ b/Tests/feaLib/builder_test.py
@@ -829,6 +829,15 @@ class BuilderTest(unittest.TestCase):
                 "} test;")
         captor.assertRegex('Already defined position for pair A V at')
 
+    def test_ignore_empty_lookup_block(self):
+        # https://github.com/fonttools/fonttools/pull/2277
+        font = self.build(
+            "lookup EMPTY { ; } EMPTY;"
+            "feature ss01 { lookup EMPTY; } ss01;"
+        )
+        assert "GPOS" not in font
+        assert "GSUB" not in font
+
 
 def generate_feature_file_test(name):
     return lambda self: self.check_feature_file(name)
--- a/Tests/feaLib/builder_test.py
+++ b/Tests/feaLib/builder_test.py
@@ -65,7 +65,7 @@ class BuilderTest(unittest.TestCase):
         spec9a spec9b spec9c1 spec9c2 spec9c3 spec9d spec9e spec9f spec9g
         spec10
         bug453 bug457 bug463 bug501 bug502 bug504 bug505 bug506 bug509
-        bug512 bug514 bug568 bug633 bug1307 bug1459
+        bug512 bug514 bug568 bug633 bug1307 bug1459 bug2276
         name size size2 multiple_feature_blocks omitted_GlyphClassDef
         ZeroValue_SinglePos_horizontal ZeroValue_SinglePos_vertical
         ZeroValue_PairPos_horizontal ZeroValue_PairPos_vertical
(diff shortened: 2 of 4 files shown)


anthrotype:
the CI failure reproduces the `AttributeError: 'NoneType' object has no attribute 'getAlternateGlyphs'` error from #2276 
https://github.com/fonttools/fonttools/pull/2277/checks?check_run_id=2409640286#step:5:388

@kontur does the FEA code I used resemble what you're trying to do? namely this:

```
lookup EMPTY {
    # pass
} EMPTY;
feature ss01 {
    sub a by a.alt1;
    lookup EMPTY;
} ss01;
feature aalt {
    feature ss01;
} aalt;
```

kontur:
@anthrotype yep, that's it. not sure how the aalt plays into it but the EMPTY and ss01 (scmp in my case) looks like what I encountered the problem with. Particularly if you replace "#pass" with something like "sub a by a;" and it does not throw the error then that seems to further confirm it.

anthrotype:
what do you expect a FEA compiler to do with such an empty lookup? Ignore it altogether as if it was not there? Or still add an entry in the LookupList with a NULL offset that points to nothing?

the spec is silent about these
http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#4e-lookup

kontur:
I'd probably expect to drop it, but would there be sense in having in there for some merging of fonts? Not sure. 
Anyway, for me the empty lookup is just a "workflow" way of composing different features for slightly different fonts. For the most part that is better done with `include`s for shared parts or different class definitions, but in this case it seemed more convenient via the lookup.

khaledhosny:
If makeotf accepts this, then we should do what it does (skip, add empty lookup, whatever), if it does not then I think erroring would be the safest action, otherwise adding an empty lookup. Silently ignoring the lookup would not be a good idea IMO.

khaledhosny:
makeotf errors with the feature code above:
```
syntax error at "}" missing { feature script language subtable lookupflag enumerate except ignore substitute reverse position parameters feat_names sizemenuname markClass GCLASS ";" } [test.fea 3]
makeotfexe [FATAL] <Test-Regular> aborting because of errors
```

simoncozens:
makeotf does, however, work with:

```
lookup EMPTY {
   ;
} EMPTY;
```

khaledhosny:
Yes. It will ignore the lookup and the features that reference only it, so:
```fea
lookup EMPTY {
    ;
} EMPTY;
feature ss01 {
    lookup EMPTY;
} ss01;
feature aalt {
    feature ss01;
} aalt;
```

will not build any layout tables.

anthrotype:
thanks for checking. So basically with this PR we're doing the same as makeotf, if I'm not mistaken.

If I change test fea file to add that empty semicolon statement:

```diff
lookup EMPTY {
-    # pass
+    ;
} EMPTY;
```

... the built output doesn't change, but the fea2fea tests fail because the duplicate semicolon is stripped by our parser.

Anyway, I think the current patch is correct, if the expectation is that references to empty named lookups should be ignored.

anthrotype:
> would there be sense in having in there for some merging of fonts?

good question. /cc @behdad

behdad:
> > would there be sense in having in there for some merging of fonts?
> 
> good question. /cc @behdad

My position is that we should always remove empty lookups. Lookup presence has no semantic value. Features do however sometimes. The case I know of currently is that HarfBuzz picks shaper for Khmer based on whether a `pref` feature is present. So in the subsetter, we keep the empty `pref` feature even if it has no lookups.

Merging shouldn't  be a problem.

behdad:
Also I don't think we should do whatever AFDKO does. We should produce the output that makes most sense and is most compact.

anthrotype:
> Merging shouldn't be a problem.

I believe he meant "merging" as in the `varLib.merger` module for making VFs, not just the`fonttools merge` script for static fonts. We currently require that the GPOS tables of each master are aligned such that they all have the same number and type of lookups (or no table at all, for "sparse" masters). But in the case of a truly empty lookup in feature file code (e.g. `lookup EMPTY { ; } EMPTY;`) the latter doesn't contain any pos/sub rules that would assign it a _type_, so it's really undefined. In this case it's safe to presume that such a no-op lookup is not present at all.
Whereas, a lookup that contains only zero values (https://github.com/fonttools/fonttools/pull/2275#issuecomment-824723754) does have a type and could, potentially, be used to make the GPOS masters align with one another to allow merging into a VF. Hm maybe just we need to think of a more granular way to specify "sparseness" for varLib.merger than "no table, or fully aligned table".

behdad:
Right...

At some point we might implement better merging such that only features need to match, not lookups. Note that we already do more analysis within SinglePos and PairPos lookup types, aligning subtables. In theory we can do more.

twardoch:
Yes, better v-merging of GPOS with sparse masters would be wonderful. It's quite a common source of problems.

PR #2278, opened 2021-04-22
[otlLib] Use ValueRecord.getEffectiveFormat() to produce smaller lookups

behdad:
Continuing from https://github.com/fonttools/fonttools/pull/2275


behdad:
@anthrotype if this looks good to you I can resolve conflicts.

behdad:
I figured why we didn't merge, because tests are failing. I'm not going to chase how to fix this as it's not trivial...

PR #2284, opened 2021-04-30
Pen error maintenance

madig:
Extracted from https://github.com/fonttools/fonttools/pull/2202 to not let my commits rot too long.

1. Make RecordingPointPen actually pass on identifiers
2. Introduce PenError to eventually replace all the asserts in pen code, as assert should be used for debugging only.

### DIFF
--- a/Lib/fontTools/pens/pointPen.py
+++ b/Lib/fontTools/pens/pointPen.py
@@ -15,7 +15,7 @@ For instance, whether or not a point is smooth, and its name.
 import math
 from typing import Any, Optional, Tuple
 
-from fontTools.pens.basePen import AbstractPen
+from fontTools.pens.basePen import AbstractPen, PenError
 
 __all__ = [
 	"AbstractPointPen",
@@ -74,7 +74,8 @@ class BasePointToSegmentPen(AbstractPointPen):
 		self.currentPath = None
 
 	def beginPath(self, identifier=None, **kwargs):
-		assert self.currentPath is None
+		if self.currentPath is not None:
+			raise PenError("Path already begun.")
 		self.currentPath = []
 
 	def _flushContour(self, segments):
@@ -106,7 +107,8 @@ class BasePointToSegmentPen(AbstractPointPen):
 		raise NotImplementedError
 
 	def endPath(self):
-		assert self.currentPath is not None
+		if self.currentPath is None:
+			raise PenError("Path not begun.")
 		points = self.currentPath
 		self.currentPath = None
 		if not points:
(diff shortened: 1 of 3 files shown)


Lib/fontTools/pens/pointPen.py
```diff
@@ -260,27 +271,35 @@ def moveTo(self, pt):
...
 
 	def lineTo(self, pt):
-		assert self.contour is not None, "contour missing required initial moveTo"
+		if self.contour is None:
+			raise PenError("Contour missing required initial moveTo")
 		self.contour.append((pt, "line"))
 
 	def curveTo(self, *pts):
-		assert self.contour is not None, "contour missing required initial moveTo"
+		if not pts:
```
justvanrossum, line 279: Should this be TypeError? Likewise qCurveTo.

Other than that: LGTM, thanks!

justvanrossum (approved):

Lib/fontTools/pens/pointPen.py
madig: Hm.

https://docs.python.org/3/library/exceptions.html?highlight=typeerror#TypeError

> Passing arguments of the wrong type (e.g. passing a list when an int is expected) should result in a TypeError, but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a ValueError.

I'll change it to a ValueError but then I need to change my Hypothesis thing :thinking:

justvanrossum: What you quote doesn't seem relevant for the case here.

```python
>>> def foo(a, b, c):
...     ...
... 
>>> foo()
  File "<stdin>", line 1, in <module>
TypeError: foo() missing 3 required positional arguments: 'a', 'b', and 'c'
>>> 
```

justvanrossum: Or:
```python
>>> def foo(pt1, *pts):
...     ...
... 
>>> pts = ()
>>> foo(*pts)
  File "<stdin>", line 1, in <module>
TypeError: foo() missing 1 required positional argument: 'pt1'
>>> 
```

madig: 🤔 🤔 🤔 

```text
In [3]: class A:
   ...:     def curveTo(self, *pts):
   ...:         print(self, repr(pts))
   ...:

In [4]: a = A()

In [5]: a.curveTo(1,2,3)
<__main__.A object at 0x00000245B88029A0> (1, 2, 3)

In [6]: a.curveTo()
<__main__.A object at 0x00000245B88029A0> ()

In [7]: a.curveTo(())
<__main__.A object at 0x00000245B88029A0> ((),)

In [8]: pts = ()
In [9]: a.curveTo(*pts)
<__main__.A object at 0x00000245B88029A0> ()
```

We always get an empty sequence as a fallback, so ValueError seems relevent to me?

justvanrossum: `*pts` represents an arbitrary number of arguments. You want to ensure the method called with at least one argument. Calling a function with the wrong number of argument ordinarily raises TypeError, hence my suggestion.

madig: Oh, I see what you're getting at now. TypeError it is then!

Issue #2279, opened 2021-04-27
[ttLib/varLib] Implement 32 bit VarStore variant, so we can fully support COLRv1


justvanrossum:
Opening this issue to discuss how to approach, and who can do what. We (@BlackFoundry and me) have some capacity to contribute, and would like to see variable COLRv1 working.

This is (at least) about:

- VarStore
- OnlineVarStoreBuilder
- VarStoreInstancer
- VarStore optimizer

Cc. @behdad, @horasio, @JeremieHornus, @rsheeter, @drott, @davelab6

behdad:
Related: https://github.com/fonttools/fonttools/issues/2211

behdad:
For reference, here's the relevant part of the change (from https://github.com/googlefonts/colr-gradients-spec/blob/main/OFF_AMD2_WD.md):

In 7.2.3.4, in the table for the ItemVariationData subtable structure, replace the field name "shortDeltaCount" with "wordDeltaCount", and replace the description of that field with the following:"

A packed field: the high bit is a flag—see details below.

Following the table for the ItemVariationData subtable structure, replace the remainder of 7.2.3.4 (including the table for the DeltaSet record structure) with the following:

The wordDeltaCount field contains a packed value that includes a flag and a “word” delta count. The format of this value is as follows:

Mask | Name | Description
-- | -- | --
0x8000 | LONG_WORDS | Flag indicating that “word” deltas are long (int32)
0x7FFF | WORD_DELTA_COUNT_MASK | Count of “word” deltas

The representation of delta values uses a mix of long types (“words”) and short types. If the LONG_WORDS flag is set, deltas are represented using a mix of int32 and int16 values. If the flag is not set, deltas are presented using a mix of int16 and int8 values. See the description of the DeltaSet record below for additional details.

The count value indicated by WORD_DELTA_COUNT_MASK is a count of the number of deltas that use the long (“word”) representation, and shall be less than or equal to regionIndexCount.

behdad:
Mostly done in https://github.com/fonttools/fonttools/pull/2285

The VarStore optimizer still needs to learn about 32bit encoding. But that branch is already useful and correctly builds 32bit VarStores, just not necessarily fully optimal in size.

behdad:
Instancer, subsetter, etc, already work and need to no change.

davelab6:
@justvanrossum what is needed to close this out?

justvanrossum:
> what is needed to close this out?

Not much, see #2285

[...]

Issue #2267, opened 2021-04-13
Reconsidering use of complex() and array.array() for vectors / points / arrays


behdad:
I've been looking into performance of various ways to represent points / vectors / arrays of coordinates etc with @hamidhma. We found some new information that came as a surprise to me, and might justify taking different directions with a few things.

Opening this issue for discussions around this topic.

Related to:
https://github.com/fonttools/fonttools/issues/1095
https://github.com/fonttools/fonttools/pull/2206 (which was reverted later)

behdad:
Basically, CPython in recent versions has optimized tuple and list so much that any other aggregate type is significantly slower. By significantly I mean factor of 2x or worse up to 5x or more. Even `complex` type is 3x slower than tuple
```
Python 3.9.4 (default, Apr  5 2021, 01:50:46) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.21.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: i = 1

In [2]: j = 2

In [3]: timeit (i, j)
43.7 ns ± 0.2 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [4]: timeit object()
69.4 ns ± 0.266 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [5]: class Point: pass

In [6]: timeit Point()
80.8 ns ± 0.494 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [7]: class Point:
   ...:     def __init__(self, x, y):
   ...:         self.x = x
   ...:         self.y = y
   ...: 

In [8]: timeit Point(i, j)
291 ns ± 1.25 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [9]: timeit complex(i, j)
148 ns ± 0.577 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

Using `complex` is still beneficial in functions that can use half the math operations compared to a `tuple` implementation. But just using it for more efficient implementation is not justified (anymore).

behdad:
All in all, this is very sad, as this makes CPython really a useful assembly language rather than higher-level language with abstractions.

That said, Instagram's Cinder looks like a very promising replacement for CPython:
https://github.com/facebookincubator/cinder

behdad:
This was more of an FYI issue. cc @justvanrossum @anthrotype @madig

madig:
> All in all, this is very sad

[...]

PR #2285, opened 2021-04-30
Improve otBase facilities / towards 32bit VariationStore

behdad:
Working on https://github.com/fonttools/fonttools/issues/2279

Doing some clean-up / optimization while at it. This doesn't do the actual 32bit VarStore yet.

Parts so far can be merged, or I'll keep adding to it.


behdad:
Okay. This is enough to be useful already. Is just the optimizer left.

behdad:
Okay, I believe this is complete. Would be nice to add tests.

@justvanrossum

justvanrossum:
> Okay, I believe this is complete. Would be nice to add tests.

Thanks! I'll try to contribute tests this week. Up to you and @anthrotype whether you want to merge before, or whether you want me to push tests to this PR. Probably better in a separate PR?

behdad:
Thanks Just. I let Cosimo decide.

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -728,13 +728,12 @@ def compile(self, writer, font):
 					# conv.repeat is a propagated count
 					writer[conv.repeat].setValue(countValue)
 				values = value
-				for i, value in enumerate(values):
-					try:
-						conv.write(writer, font, table, value, i)
-					except Exception as e:
-						name = value.__class__.__name__ if value is not None else conv.name
-						e.args = e.args + (name+'['+str(i)+']',)
```
anthrotype, line 707: hm. this is the base table's compile method, so by default we'll lose the index into the array where an exception occurs. I suppose that can sometimes be useful, e.g. which lookup or which subtable failed to be compiled for whatever reason?

Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -338,6 +338,14 @@ def readArray(self, reader, font, tableDict, count):
...
 	def read(self, reader, font, tableDict):
 		return font.getGlyphName(reader.readValue(self.typecode, self.staticSize))
+	def writeArray(self, writer, font, tableDict, values):
+		glyphMap = font.getReverseGlyphMap()
+		try:
+			values = [glyphMap[glyphname] for glyphname in values]
+		except KeyError:
+			# Slower, but will not throw a KeyError on an out-of-range glyph name.
+			values = [font.getGlyphID(glyphname) for glyphname in values]
+		writer.writeArray(self.typecode, values)
```
anthrotype, line 348: is this change related to https://github.com/fonttools/fonttools/issues/1536#issuecomment-472581569 ?

Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -1622,9 +1622,18 @@ def read(self, reader, font, tableDict):
...
 		wordCount = tableDict["NumShorts"]
 
+		# https://github.com/fonttools/fonttools/issues/2279
+		longWords = bool(wordCount & 0x8000)
+		wordCount = wordCount & 0x7FFF
+
+		(readBigArray, readSmallArray) = {
+			False: (reader.readShortArray, reader.readInt8Array),
+			True:  (reader.readLongArray,  reader.readShortArray),
+		}[longWords]
```
anthrotype, line 1632: the condition/key being just a bool, is this really faster/more readable than an `if longWords`..?

Lib/fontTools/varLib/builder.py
```diff
@@ -26,39 +26,40 @@ def buildVarRegionList(supports, axisTags):
...
-	zeroes = set(range(count))
+	bit_lengths = [0] * count
 	for item in items:
-		wides = [i for i in narrows if not (-128 <= item[i] <= 127)]
-		narrows.difference_update(wides)
-		nonzeroes = [i for i in zeroes if item[i]]
-		zeroes.difference_update(nonzeroes)
-		if not narrows and not zeroes:
-			break
+		bl = [(i + (i < -1)).bit_length() for i in item]
```
anthrotype, line 37: is `i + (i < -1)` maybe too clever a trick? you're deliberately using the fact that True/False coerce to integers 1 and 0 in python, but it makes the code slightly harder to understand (to me at least)

Lib/fontTools/varLib/varStore.py
```diff
@@ -5,7 +5,6 @@
 				      buildVarRegion, buildVarData)
 from functools import partial
 from collections import defaultdict
-from array import array
```
anthrotype, line 8: is that because you found out `list` are actually faster than `array.array`?

Lib/fontTools/varLib/varStore.py
```diff
@@ -374,12 +374,11 @@ def _characteristic_overhead(chars):
 		as a VarData."""
 		c = 6
 		while chars:
-			if chars & 3:
+			if chars & 15:
```
anthrotype, line 377: maybe `0b1111` instead of `15` may make it more readable

anthrotype (approved):
Thanks Behdad for working on this!!
LGTM, with some minor comments.

Yeah, we want to add tests before we ship this, it doesn't really matter if you do it in this or a separate PR.

justvanrossum:
Should we worry about how it's now possible to accidentally create a 32-bit VarStore instead of getting an overflow error, for non-COLRv1 use?

anthrotype:
> accidentally create a 32-bit VarStore

how can one do that? got an example?

anthrotype:
I suppose if one passes to varLib master values with very big numbers, it may be the case that delta values exceed the max 16bit integer limits and instead of an overflow error one would now get a 32bit itemVarStore.
I guess it may be desirable sometimes (e.g. backward compatibility reasons) to only use the old varstore format, without 32-bit deltas. Maybe we should have option for that.

justvanrossum:
> > accidentally create a 32-bit VarStore

> how can one do that? got an example?

Even if all master values are between - 32768 and 32767, it's possible to create deltas that are outside that range. This would fail with an overflow before, but will now silently upgrade to 32 bit.

justvanrossum:
In other words, it's currently possible to silently compile a non OT1. 8 compliant font. I feel this will come back to bite us...

justvanrossum:
Tests are ready for review.

justvanrossum:
Hm, I'm running into a problem with 32-bit values when adding a compile/decompile step to my test in Tests/varLib/varStore_test.py. Uncomment the commented-out test data and it starts failing.

anthrotype:
> Uncomment the commented-out test data and it starts failing.

maybe uncomment and let the CI fail, so we can read what error is printed in the test log

justvanrossum:
Found it, but I'm not convinced the solution is correct in all cases. The array.array documentation doesn't provide absolute answers wrt. the item size of `i` and `l` type codes: https://docs.python.org/3/library/array.html

justvanrossum:
> The array.array documentation doesn't provide absolute answers wrt. the item size of i and l type codes 

I see that in many places in fonttools we assume that `array.array("I").itemsize == 4`, so it can't be all wrong...

behdad:
Thank you both. I'll address all review comments.

behdad:
> In other words, it's currently possible to silently compile a non OT1. 8 compliant font. I feel this will come back to bite us...

That's how we always did things. :)

That said, with ALL the other expansion projects I'm planning, I think we will get to a point that to compile fonts you have to tell fonttools what font "profile" you want to adhere to. We'll address this then if needed.

justvanrossum:
I'd love to get this merged. Some of Cosimo's review comments are stylistic in nature, and I could gladly address them myself or just leave them be. But the other comments are beyond me and need some resolution.

anthrotype:
i'm ok if you'd like to merge. My only minor concern is with the loss of array index information from the exception traceback, which I guess can be useful in some circumstances?
Re: `l` vs `i` array.array type size, I suppose it's ok, int should be 4 bytes on 32-bit and 64-bit architectures, and we don't care about 16-bit anymore (even 32-bit is relatively outdated except for windows where I think python.org keeps offering that as the default installer - might be worth testing on a 32-bit windows python to confirm this is correct, I don't know what architecture our github actions windows runner's python are on)

justvanrossum:
> Re: l vs i array.array type size, I suppose it's ok,

Yeah, like I said, we assume "i" is 4 bytes all over the place already.

anthrotype:
> the other comments are beyond me and need some resolution.

which ones?

justvanrossum:
> > the other comments are beyond me and need some resolution.

> which ones?

- https://github.com/fonttools/fonttools/pull/2285#discussion_r625695848
- https://github.com/fonttools/fonttools/pull/2285#discussion_r625698973
- https://github.com/fonttools/fonttools/pull/2285#discussion_r625707717

Lib/fontTools/ttLib/tables/otConverters.py
behdad: Is what I proposed there as well. I didn't remember that. Just came across it when profiling.

Lib/fontTools/ttLib/tables/otBase.py
behdad: The lookup/subtable etc raise proper exception because the index (`i`) is passed down to write objects. This was a hack I introduced to get something resembling the full path down to the failure. I agree it's *sometimes* useful. I can put it back inside `writeArray`. Let me try a few things.

Then again, we were fine without this... I think I added it when I was working on 99proof.

Lib/fontTools/varLib/varStore.py
behdad: Yes. I will comment in https://github.com/fonttools/fonttools/issues/2267
But in general, in CPython, they super-optimized list and tuple. Anything else is lots slower :(.
Even the use in `GlyphCoordinates` I think I will remove soon.

Lib/fontTools/varLib/builder.py
anthrotype: gotcha

behdad:
> Yeah, like I said, we assume "i" is 4 bytes all over the place already.

I will add a big assert in otBase or something.

Lib/fontTools/ttLib/tables/otConverters.py
behdad: I changed my mind. :)

behdad:
> I will add a big assert in otBase or something.

What bothers me more is the mismatch now:
```
        def writeULong(self, value):
                self.items.append(struct.pack(">L", value))
        def writeULongArray(self, values):
                self.writeArray('I', values)
```
I'm going to use "I" for both.

behdad:
Okay! Ready to be merged AFAIC. Thanks all.

justvanrossum (approved):
LGTM, thanks!

justvanrossum:
@anthrotype would you like to give it a final once-over or shall I just merge?

anthrotype (approved):
Lgtm, thanks

PR #2286, opened 2021-05-02
[mutator] Round lsb for CFF2 fonts as well

khaledhosny:
Otherwise head table compilation would fail because of the floating point side bearing. The advance width is already rounded.


behdad:
Lgtm. Thanks.

PR #2288, opened 2021-05-04
removeOverlaps: work around pathops.simplify error

anthrotype:
Sometimes skia-pathops simplify may fail (for unknown reasons which I'm still trying to debug).
It's a good idea to know the name of the offending glyph
https://github.com/google/fonts/issues/3365

EDIT: It's actually even better to **not** fail at all. It turns out that if we round the path's coordinates to integers _before_ passing it on to pathops.simplify, then it works -- at least for the two glyphs in the Noto Ethiopic and Telugu fonts which triggered this issue.

The effect of rounding before removing overlaps is minimal, since we are going to round afterwards anyway once we encode the glyf table; besides, we only try rounding when simplify fails with the unrounded float coordinates (the ones that we get from the instancer).

All in all, I believe the workaround is better than blocking those projects while the Skia bug is fixed upstream (https://bugs.chromium.org/p/skia/issues/detail?id=11958), which may take a while since we have no idea what's going on under the hood.

### DIFF
--- a/Lib/fontTools/ttLib/removeOverlaps.py
+++ b/Lib/fontTools/ttLib/removeOverlaps.py
@@ -81,6 +81,37 @@ def ttfGlyphFromSkPath(path: pathops.Path) -> _g_l_y_f.Glyph:
     return glyph
 
 
+def _simplify(path: pathops.Path, debugGlyphName: str) -> pathops.Path:
+    # skia-pathops has a bug where it sometimes fails to simplify paths when there
+    # are float coordinates and control points are very close to one another.
+    # Rounding coordinates to integers works around the bug.
+    # Since we are going to round glyf coordinates later on anyway, here it is
+    # ok(-ish) to also round before simplify. Better than failing the whole process
+    # for the entire font.
+    # https://bugs.chromium.org/p/skia/issues/detail?id=11958
+    # https://github.com/google/fonts/issues/3365
+    try:
+        path2 = pathops.simplify(path, clockwise=path.clockwise)
+    except pathops.PathOpsError:
+        path2 = pathops.Path()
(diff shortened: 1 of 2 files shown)


Lib/fontTools/ttLib/removeOverlaps.py
```diff
@@ -76,6 +81,37 @@ def ttfGlyphFromSkPath(path: pathops.Path) -> _g_l_y_f.Glyph:
...
+            raise RemoveOverlapsError(
+                f"Failed to remove overlaps from glyph {debugGlyphName!r}"
+            ) from e
+        else:
+            log.debug(
+                "skia-pathops failed to simplify '%s' with float coordinates, "
+                "but succeded using rounded integer coordinates",
+                debugGlyphName,
+            )
+    return path2
```
rsheeter, line 112: Optional, I found this a little hard to read. Spitballing, what about:

```python
try:
  # Normally simplify works fine
  return pathops.simplify(path, clockwise=path.clockwise)
except pathops.PathOpsError:
  pass

# Sometimes simplify fails on a path but works when rounded
path = _round_path(path)

try:
  # Original code did pathops.simplify one time, path.simplify the second, not sure if that was significant
  path = pathops.simplify(path, clockwise=path.clockwise)
  log.debug("skia-pathops failed to simplify etc etc")
  return path
except pathops.PathOpsError:
  # dump and re-raise

assert False, "Unreachable"
```

rsheeter (approved):

Lib/fontTools/ttLib/removeOverlaps.py
anthrotype: Done

Issue #2295, opened 2021-05-07
undefined variable in method _synthesizeVMetrics() of class table__g_l_y_f


Eigi:
While fixing issue #2292 I found another one. Variable "topSideBearing" in method _synthesizeVMetrics() is not defined. I' not sure if topSideBearing is ment to be the same as topSideY.

anthrotype:
ouch, this probably broke when the code was reshuffled in https://github.com/fonttools/fonttools/pull/2266

it should be `topSideBearing = otRound(topSideY - glyph.yMax)`

behdad:
Ugh. I'm really sorry for making this mess. Thanks Eigi & Cosimo.

PR #2296, opened 2021-05-07
Test getPhantomPoints

Eigi:
This new test passes with fontTools version 4.22.0 but raises NameError with version 4.22.2.dev0
see issue #2295


anthrotype:
I cherry-picked your test in #2305 

for the //UNC windows paths, we should fix in a separate PR

PR #2298, opened 2021-05-07
[otTables] Default varIdx should be 0xFFFFFFFF

### LINKED ISSUE #2297: [colorLib/ttLib.tables.otTables] Should default for VariableValue.varIdx be 0xFFFFFFFF?
The default varIdx for a VariableValue object is 0.

When building color glyphs with colorLib, when a field in a Var structure is not a tuple but a simple value, this default varIdx is used.

To me it would make more sense for this default to be 0xFFFFFFFF, meaning "no variations". This would be more convenient when constructing variable COLRv1 glyphs. At the moment I have to set 0xFFFFFFFF explicitly in my color glyph Paint dict.

What do you think, @anthrotype and @behdad?

https://github.com/fonttools/fonttools/blob/9c18e2a07d10d61dd377cca285bbd11dc188d1b0/Lib/fontTools/ttLib/tables/otTables.py#L1259-L1264

justvanrossum:
The default value for varIdx should be 0xFFFFFFFF to indicate 'no variations'. Fixes #2297


Tests/ttLib/tables/C_O_L_R_test.py
```diff
@@ -147,12 +147,12 @@ def test_round_trip_xml(self, font):
     (b"\x00\x00\x08", "Offset to BackdropPaint from beginning of PaintComposite (8)"),
     (b"\x0d", "BaseGlyphV1Record[1].Paint.BackdropPaint.Format (13)"),
     (b"\x00\x00\x34", "Offset to Paint from beginning of PaintVarTransform (52)"),
-    (b"\x00\x01\x00\x00\x00\x00\x00\x00", "Affine2x3.xx.value (1.0)"),
-    (b"\x00\x00\x00\x00\x00\x00\x00\x00", "Affine2x3.xy.value (0.0)"),
-    (b"\x00\x00\x00\x00\x00\x00\x00\x00", "Affine2x3.yx.value (0.0)"),
-    (b"\x00\x01\x00\x00\x00\x00\x00\x00", "Affine2x3.yy.value (1.0)"),
-    (b"\x01\x2c\x00\x00\x00\x00\x00\x00", "Affine2x3.dx.value (300.0)"),
-    (b"\x00\x00\x00\x00\x00\x00\x00\x00", "Affine2x3.dy.value (0.0)"),
+    (b"\x00\x01\x00\x00\xFF\xFF\xFF\xFF", "Affine2x3.xx.value (1.0)"),
```
anthrotype, line 150: nit: I believe the rest of the hex dump is using lowercase letters (like python bytes repr)

anthrotype (approved):
LGTM, thanks!

PR #2300, opened 2021-05-10
VarLibMergeError first parameter must be a Merger object

anthrotype:
a lot of places instead pass in the 'self' parameter, incorrectly assuming that is a Merger.
This is meant to fix 'AttributeError: 'dict' object has no attribute 'ttfs'' as reported in https://github.com/googlefonts/fontmake/issues/766#issuecomment-836576341

there's still some places where I'm not sure what the right thing to do is

### DIFF
--- a/Lib/fontTools/varLib/merger.py
+++ b/Lib/fontTools/varLib/merger.py
@@ -143,7 +143,7 @@ class AligningMerger(Merger):
 def merge(merger, self, lst):
 	if self is None:
 		if not allNone(lst):
-			raise NotANone(self, expected=None, got=lst)
+			raise NotANone(merger, expected=None, got=lst)
 		return
 
 	lst = [l.classDefs for l in lst]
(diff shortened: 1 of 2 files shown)


anthrotype:
ping @simoncozens

Lib/fontTools/varLib/merger.py
```diff
@@ -584,8 +586,10 @@ def _MarkBasePosFormat1_merge(self, lst, merger, Mark='Mark', Base='Base'):
 		# input masters.
 
 		if not allEqual(allClasses):
+			# XXX What's going on here? 'allClasses' is not an Exception. Shouldn it be
+			# raise ShouldBeConstant(merger, expected=allClasses[0], got=allClasses)
```
simoncozens, line 590: I think it probably should be, yes.

simoncozens (commented):
Looks good so far - what are the other areas you were not sure about?

PR #2304, opened 2021-05-12
Small refactor (and reformat) varLib.models

simoncozens:
* axisPoints is not used outside the sort function
* Avoid unnecessary intermediary variables
* Index was never used
* Another unused enumeration
* Reformat with black
* Remove else-after-return

### DIFF
--- a/Lib/fontTools/varLib/models.py
+++ b/Lib/fontTools/varLib/models.py
@@ -344,7 +344,7 @@ class VariationModel(object):
 				maxV[k] = max(v, maxV.get(k, v))
 
 		regions = []
-		for i,loc in enumerate(locations):
+		for loc in locations:
 			region = {}
 			for axis,locV in loc.items():
 				if locV > 0:
--- a/Lib/fontTools/varLib/models.py
+++ b/Lib/fontTools/varLib/models.py
@@ -280,7 +280,7 @@ class VariationModel(object):
 		return new_list
 
 	def _computeMasterSupports(self):
-		supports = []
+		self.supports = []
 		regions = self._locationsToRegions()
 		for i,region in enumerate(regions):
 			locAxes = set(region.keys())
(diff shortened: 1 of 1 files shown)


behdad:
LGTM. Thanks!

anthrotype (approved):

Lib/fontTools/varLib/models.py
```diff
@@ -280,7 +280,7 @@ def reorderMasters(self, master_list, mapping):
 		return new_list
 
 	def _computeMasterSupports(self):
-		supports = []
+		self.supports = []
```
justvanrossum: Although: doing many getattrs in the inner loop will be slower than appending to a local var and then setting the attr. Same for deltaWeights below.

simoncozens: This may be true. This isn't hot code, however; for each table, the model is instantiated and the supports are computed once, then the submodels are used to generate deltas and interpolate.

behdad: Both of those comments are correct indeed. I have mentally switched to always using a local. But am fine both ways.

justvanrossum: > This isn't hot code 

Fair enough.

simoncozens:
I missed one:

```
                if weight == 1:
                    delta -= out[j]
                else:
                    delta -= out[j] * weight
```

Just the last line will do.

justvanrossum:
> Just the last line will do.

But that's a useful optimization, no? Esp. if we're working on arrays of other objects that behave number-like.

behdad:
> But that's a useful optimization, no? Esp. if we're working on arrays or other objects that behave number-like.

Correct. Eg. a `GlyphCoordinates` object.

simoncozens:
Oh, good point. I always mess up the typing of getDeltas.

justvanrossum:
It's one of the great powers of the VariationModel abstraction: it works correctly on anything that correctly behaves like a number. (Or actually, I think objects only need to support `__add__`, `__sub__` and `__mul__` to work well with VariationModel.)

behdad:
> I think objects only need to support `__add__`, `__sub__` and `__mul__` to work well with VariationModel.)

I think it's down to just `__isub__` and `__mul__`.

behdad:
> Oh, good point. I always mess up the typing of getDeltas.

Feel free to annotate it with `Generic` type markers:
https://docs.python.org/3/library/typing.html#typing.Generic

Issue #2299, opened 2021-05-09
function posix(path) in designspaceLib does not work with Windows UNC path


Eigi:
On Windows UNC (Uniform Naming Convention) paths have the following form:  
`\\SERVER\SHARE\FOLDER-NAME\FILE-NAME` or  
`\\IP-ADDRESS\SHARE\FOLDER-NAME\FILE-NAME`  
the possix function translates this to `SERVER/SHARE/FOLDER-NAME/FILE-NAME`. Therefore all path attributes for sources and instances are invalid if a DesignSpaceDocument is loaded via an UNC path.

anthrotype:
@Eigi I see in 84d77e5 you add a `//` prefix, does that still work for UNC paths? I don't have a Windows machine to test with. Should we instead maybe restore the original backslashes `\\` or even just leave the UNC paths untouched?

madig:
Where are the UNC paths coming from?

anthrotype:
I suppose from the user? why not, it's perfectly valid to wanting to store one's sources on some server

madig:
I mean, doesn't the DS format always want relative paths? We had this problem before, with CI tmp dirs mounted on a different drive.

anthrotype:
well, it is possible that both the designspace _and_ the sources are on the same UNC drive..

anthrotype:
ah, I see what you mean. If both were on the same drive, they would be using relative paths and there would not be this problem.. only when they are on different drives that one may incur in absolute paths, including UNC paths 🤔

anthrotype:
> doesn't the DS format always want relative paths?

I'm actually not sure

anthrotype:
The designspace spec says only relative paths are allowed, so that would basically exclude UNC paths

> filename: required, string. A path to the source file, relative to the root path of this document. The path can be at the same level as the document or lower.

Eigi:
Yes, in my case designspace, sources and instances are on the same UNC share.
See my comment in the pull request.

Eigi:
> @Eigi I see in [84d77e5](https://github.com/fonttools/fonttools/commit/84d77e5c3ac35dbea123b4dcbf4197ff81ac0a22) you add a `//` prefix, does that still work for UNC paths? I don't have a Windows machine to test with. Should we instead maybe restore the original backslashes `\\` or even just leave the UNC paths untouched?

Yes it works with forward slashes. Tested in my environment.

anthrotype:
I meged #2306, thank you

PR #2306, opened 2021-05-13
make function posix(path) in designspaceLib work with Windows UNC path

Eigi:
fix for #2299


anthrotype:
can you elaborate a bit on how you use UNC paths in designspace files?
According to the spec, the `filename` attribute should contain only relative paths, and as far as I understand these UNC paths are by definition absolute paths.

Eigi:
Yes, the filename attribute is relative, but the path attribute of the designSpaceDocument itself is absolut. And the path attribute of sources and instances is calculated as absolute paths form the respective filename and the path of the designSpaceDocument. The error occurs if the designSpaceDocument is loaded from a server with an UNC path. As a result the sources can't be loaded even if they are in the same folder as the designSpaceDocument. This is the standard situation in my production environment.

anthrotype:
i see. But are you sure that inserting two leading _forward_ slashes will still make it work as a UNC path? Don't the latter always require two backslashes? /cc @madig

Eigi:
two leading forward slashes work. and we have one forward slash in the mount part of the UNC path anyway, between server and share. `//server/share/folder/filename`.

anthrotype:
ok, let's merge it then

PR #2307, opened 2021-05-13
Add ttLib.TTFont.getTableTags public method for access to full set of table tags in a TTFont object

chrissimpkins:
I was trying to do the following:

```python
for table in [GET A LIST OF AVAILABLE TTFont TABLE TAGS]:
    table_data = tt.getTableData(table)
    ...
```

and couldn't find a consistent approach to get a list of available table tags across TTFont obj's that are instantiated from font binary file paths and TTX XML imports.  These data seem to be buried in the `reader.keys()` or `tables.keys()` instance attribute methods based on whether the TTFont is gen'd from a font file path (reader appears to include full data with all table tags, tables appears to include partial table data in my test on NotoSansDevanagari-Regular.ttf) or XML import (reader instance attribute not available, tables has full list but includes an unnecessary "GlyphOrder" item).  

This PR adds a new `ttLib.TTFont.getTableTags` public method that appears to return a consistent table tag list irrespective of the approach used to gen the TTFont obj.

### DIFF
--- a/Lib/fontTools/ttLib/ttFont.py
+++ b/Lib/fontTools/ttLib/ttFont.py
@@ -656,6 +656,17 @@ class TTFont(object):
 		else:
 			raise KeyError(tag)
 
+	def getTableTags(self):
+		"""Returns a list of available table tags as strings.
+		"""
+		if self.reader:
+			return list(self.reader.keys())
+		else:
+			table_tags = list(self.tables.keys())
+			if "GlyphOrder" in table_tags:
+				table_tags.remove("GlyphOrder")
+			return table_tags
+
 	def getGlyphSet(self, preferCFF=True):
 		"""Return a generic GlyphSet, which is a dict-like object
 		mapping glyph names to glyph objects. The returned glyph objects
--- a/Tests/ttLib/ttFont_test.py
+++ b/Tests/ttLib/ttFont_test.py
@@ -1,4 +1,5 @@
 import io
+from pathlib import Path
 from fontTools.ttLib import TTFont, newTable, registerCustomTableClass, unregisterCustomTableClass
 from fontTools.ttLib.tables.DefaultTable import DefaultTable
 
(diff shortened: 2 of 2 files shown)


behdad:
Why not call it `.keys()`?

chrissimpkins:
> Why not call it `.keys()`?

Seems to be one @

https://github.com/fonttools/fonttools/blob/2aaeba66c10dc3ec58bddcaa1b0a249ab152c9c8/Lib/fontTools/ttLib/ttFont.py#L355

but it doesn't do what I am trying to do here.  It looks like it sorts the tags and keeps/adds "GlyphOrder" in the list.  The sorting is not an issue (though unnecessary for my purposes), but "GlyphOrder" is not an actual table so we need another step to remove it from the list.

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -656,6 +656,17 @@ def getTableData(self, tag):
 		else:
 			raise KeyError(tag)
 
+	def getTableTags(self):
+		"""Returns a list of available table tags as strings.
+		"""
+		if self.reader:
+			return list(self.reader.keys())
```
justvanrossum, line 663: This misses tables that have been added to the font since it was read from disk.

Maybe just call `self.keys()` and delete `GlyphOrder` from the result?

anthrotype: just do `return [tag for tag in self.keys() if tag != "GlyphOrder"]`

Tests/ttLib/ttFont_test.py
```diff
@@ -46,3 +47,54 @@ def test_registerCustomTableClassStandardName():
         assert font[TABLETAG].compile(font) == b"\x04\x05\x06"
     finally:
         unregisterCustomTableClass(TABLETAG)
+
+
+def test_gettabletags_method(tmp_path):
```
anthrotype, line 52: not sure it's worth adding this very verbose test for such a trivial change

Lib/fontTools/ttLib/ttFont.py
chrissimpkins: Np!  Thanks for this input.  I'll change it.

Tests/ttLib/ttFont_test.py
chrissimpkins: Np, happy to remove it.  The behavior was different by reader type and I was confirming that I landed on something that worked for both font reads and ttx reads.

Lib/fontTools/ttLib/ttFont.py
anthrotype: indeed, this getTableTabls looks to me just syntactic sugar given we already have `.keys()` and one can always filter "GlyphOrder" out with a one liner if one is not interested in it. But if you think it's still useful to have, I don't mind adding it

chrissimpkins: > indeed, this getTableTabls looks to me just syntactic sugar given we already have `.keys()` and one can always filter "GlyphOrder" out with a one liner if one is not interested in it. But if you think it's still useful to have, I don't mind adding it

Sgtm.  Now I know the approach.  I'll use this in the future. Thanks all!

PR #2309, opened 2021-05-17
Refactor SingleSubst format determination

simoncozens:
I had to look up what a `for...else...` loop meant. It's really not obvious. This is hopefully a bit easier to follow.

### DIFF
--- a/Lib/fontTools/ttLib/tables/otTables.py
+++ b/Lib/fontTools/ttLib/tables/otTables.py
@@ -752,20 +752,17 @@ class SingleSubst(FormatSwitchingBaseTable):
 		sortableItems = sorted(zip(gidItems, items))
 
 		# figure out format
-		format = 2
-		delta = None
-		for inID, outID in gidItems:
-			if delta is None:
-				delta = (outID - inID) % 65536
-
-			if (inID + delta) % 65536 != outID:
-					break
+		if len(gidItems) == 0:
+			format = 2
 		else:
-			if delta is None:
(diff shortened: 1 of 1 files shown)


anthrotype:
> I had to look up what a for...else... loop meant

anthrotype:
> I had to look up what a for...else... loop meant

I just read the `else` as if it's written `ifnobreak`

justvanrossum:
The rationale of "let's refactor because it's using a Python feature I haven't seen before" is rather dubious.

It's an arbitrary change that perhaps makes it better for you, perhaps not for others. It's completely siubjective, and therefore better left alone.

Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -752,20 +752,17 @@ def preWrite(self, font):
...
-			else:
-				format = 1
+			format = 1
+			items = iter(gidItems)
+			firstInID, firstOutID = next(items)
+			delta = (firstOutID - firstInID) % 65536
+			for inID, outID in items:
+				if (inID + delta) % 65536 != outID:
+					format = 2
+					break
```
anthrotype, line 765:
```python
format = 2 if any((inID + delta) % 65536 != outID for inID, outID in items) else 1
```

anthrotype (approved):

anthrotype:
> perhaps makes it better for you, perhaps not for others

fair point. I actually had to read this several times to convince myself it was doing the same thing as before the patch.

PR #2311, opened 2021-05-19
[feaLib] fixed two instances of 'DeprecationWarning: invalid escape sequence'

justvanrossum:
...although I'm not sure what this backslash-space sequence is supposed to do.

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -700,7 +700,7 @@ class AttachStatement(Statement):
 
 
 class ChainContextPosStatement(Statement):
-    """A chained contextual positioning statement.
+    r"""A chained contextual positioning statement.
 
     ``prefix``, ``glyphs``, and ``suffix`` should be lists of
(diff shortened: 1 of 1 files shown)


dscorbett:
The backslash suppresses the space between the code span and the ⟨s⟩. ``:class:`LookupBlock`\ s`` is rendered ⟨`LookupBlock`s⟩ whereas ``:class:`LookupBlock` s`` is rendered ⟨`LookupBlock` s⟩. The space is necessary because ``:class:`LookupBlock`s`` is rendered literally ⟨:class:\`LookupBlock\`s⟩ plus a warning message.

justvanrossum:
> The backslash suppresses the space between the code span [ ... ]

Thanks.

The alternative fix for the warning is to make these doc strings "raw" strings: `r"Abc \ def"`. Any preference?

dscorbett:
As long as the output documentation looks the same, I have no preference for the type of docstring.

anthrotype:
+1 for ``r""":class:`LookupBlock`\ s"""``

Issue #2312, opened 2021-05-19
Subsetting fails: 'NoneType' object has no attribute 'getEffectiveFormat'


nathan-williams:
The following command fails: `fonttools subset fonts/ofl/NotoSansSiddham-Regular.ttf "*"`. Forgive me if I'm using the tool incorrectly. 

Stack trace:

```
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/bin/fonttools", line 11, in <module>
    sys.exit(main())
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/fontTools/__main__.py", line 30, in main
    runpy.run_module(mod, run_name='__main__')
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 208, in run_module
    return _run_code(code, {}, init_globals, run_name, mod_spec)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/fontTools/subset/__main__.py", line 6, in <module>
    sys.exit(main())
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/fontTools/misc/loggingTools.py", line 372, in wrapper
    return func(*args, **kwds)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/fontTools/subset/__init__.py", line 3094, in main
    subsetter.subset(font)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/fontTools/subset/__init__.py", line 2897, in subset
    self._prune_post_subset(font)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/fontTools/subset/__init__.py", line 2880, in _prune_post_subset
    retain = table.prune_post_subset(font, self.options)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/fontTools/subset/__init__.py", line 1665, in prune_post_subset
    table.LookupList.prune_post_subset(font, options)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/fontTools/subset/__init__.py", line 1264, in prune_post_subset
[...]

PR #2313, opened 2021-05-19
subset: fix AttributeError 'NoneType' object has no attribute 'getEffectiveFormat'

anthrotype:
when SinglePos table `Value` can be None (when ValueFormat == 0).
Attempting to `getEffectiveFormat` from it, leads to AttributeError in SinglePos `prune_post_subset` method.
We should simply skip prune_post_subset when the Value is None.

Adding a failing test case to repro issue #2312, followed by the fix itself

### DIFF
--- a/Tests/subset/data/GPOS_SinglePos_no_value_issue_2312.subset.ttx
+++ b/Tests/subset/data/GPOS_SinglePos_no_value_issue_2312.subset.ttx
@@ -1,0 +1,234 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ttFont sfntVersion="\x00\x01\x00\x00" ttLibVersion="4.24">
+
+  <GlyphOrder>
+    <!-- The 'id' attribute is only for humans; it is ignored when parsed. -->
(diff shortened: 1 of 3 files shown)


anthrotype:
(arguably a zero valued SinglePos adjustment is useless, but still valid so subsetter should not choke on it)

behdad:
First, I suggest you use `if value is not None` instead of `if not value`.

That said, what you wrote is wrong IMO as it will drop the "useless" subtable. But such "useless" subtable might have been there to stop the next subtable from applying... So, your patch can change semantics of the font.

I'd just do a `value.getEffectiveFormat() if value is not None else 0` instead...

behdad:
Hummm.

I now realize that the existing code has the same problem. Ouch!!!

behdad:
Change the ultimate return back to `return True`.  I broke it last month in 3a6ebb8dc.

Drop such subtables needs more involved logic than we currently have implemented.

anthrotype:
> "useless" subtable might have been there to stop the next subtable from applying

good point. Ok I changed to always return True. So wixed two bugs in one ;)

PR #2317, opened 2021-05-20
README: add instructions on how to make a release

anthrotype:
to reduce the bus factor -- or maybe I should call it the 'new baby factor'


justvanrossum (approved):
Thanks!

Issue #2314, opened 2021-05-20
Parsing of 'post' tables version 2.0


eric-muller:
The current code in table__p_o_s_t.unpackPSStrings, used for post version 2.0, assumes that all bytes in the 'post' table after the glyphNameIndex array are glyph names, and decodes all those bytes accordingly. Consider a font where there are 2 glyphs

- the first is name ".notdef" and the second is named "eric". 
- post.glyphNameIndex [0] = 0 and post.glyphNameIndex [1] = 258
- post.stringData [0..4] = {4, 'e', 'r', 'i', 'c'}

Bytes in the post table after post.stringData [4] are not referenced, and arguably could be anything. Of course, it's not a good idea to have such extra bytes, but I ran into a font like that, and it's arguably  valid. The spec issue is that the size of the stringData array is not well defined: if could be "what it takes to exhaust the rest of the post table" (what fonttools does today) or it could be "what it takes to resolve all the indices in glyphNameIndex", which is what many other tools do (spot, aots, DLT OT Master).

Should fonttools be changed to the second interpretation?

anthrotype:
If I understand correctly, the issue here is that `unpackPStrings` is attempting to parse anything that looks like a PS string it can until it uses up the the stringData, but instead you suggest it should only parse up to numGlyphs and ignore the tail (perhaps with a warning)?

https://github.com/fonttools/fonttools/blob/604bfdb2fbfe38d713eb91780330859532fde826/Lib/fontTools/ttLib/tables/_p_o_s_t.py#L255-L263

Sounds reasonable, I'm not opposed to that.
What exactly happens now, you get extra garbage strings in the post.extraNames? Do you get an error?

emuller-amazon:
> until it uses up the stringData
more precisely until the end of the post table

> you suggest it should only parse up to numGlyphs
Up to the number of strings referenced in glyphNameIndex

See pull request <https://github.com/fonttools/fonttools/pull/2318> for details.

> What exactly happens now, you get extra garbage strings in the post.extraNames? Do you get an error?

There is no direct failure in unpackPStrings: data[index+1:index+1+length] will silently stop at len(data) if index+1+length is past the end of the data, and tostr accepts any byte sequence. But when using ttx to decompile, I end up with random bytes in the generated XML, in particular U+0000, which is illegal in XML.

anthrotype:
I see, thanks for the PR and for clarifying that.

behdad:
Let's kill `extraNames` please.

[...]

PR #2318, opened 2021-05-20
Fix parsing of post tables version 2.0 (#2314)

emuller-amazon:


Lib/fontTools/ttLib/tables/_p_o_s_t.py
```diff
@@ -252,14 +259,17 @@ def fromXML(self, name, attrs, content, ttFont):
...
+def unpackPStrings(data, nbStrings):
 	strings = []
 	index = 0
-	dataLen = len(data)
-	while index < dataLen:
+	while nbStrings > 0:
 		length = byteord(data[index])
 		strings.append(tostr(data[index+1:index+1+length], encoding="latin1"))
 		index = index + 1 + length
+		nbStrings = nbStrings - 1
```
anthrotype, line 269:
```suggestion
		nbStrings -= 1
```

anthrotype: Does "nb" in `nbStrings` stand for "NumBer" of strings? I think just `n` or `num` would be clear, as the function name gives context about the thing we are counting with that number.

anthrotype (commented):
thank you, I left some comments. 
I see we have no unit tests for the post table in Tests/ttLib/tables directory. Ideally we should have some.

Lib/fontTools/ttLib/tables/_p_o_s_t.py
anthrotype: apologies for the bikeshedding; thinking some more about nbStrings parameter, maybe `maxStrings` reads better, as it's the _maximum_ number of ps strings we want to be read from the string data.

anthrotype: In addition to extra bytes, is it ever possible that we don't have enough bytes, that is maxStrings is greater than the number of strings actually encoded in the string data? if so, do we get an IndexError at the next iteration following the one where we run out of data?

emuller-amazon: Yes, not enough bytes is a possibility. Should that generate a log.warning (and use "" for missing strings) or a log.error?

Lib/fontTools/ttLib/tables/_p_o_s_t.py
```diff
@@ -85,7 +87,12 @@ def decode_format_2_0(self, data, ttFont):
 		indices.frombytes(data[:2*numGlyphs])
 		if sys.byteorder != "big": indices.byteswap()
 		data = data[2*numGlyphs:]
-		self.extraNames = extraNames = unpackPStrings(data)
+
+		maxIndex = 0
+		for i in range(numGlyphs):
+			if indices[i] > maxIndex: maxIndex = indices [i]
```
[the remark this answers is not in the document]
emuller-amazon: Used Behdad's suggestion.

Lib/fontTools/ttLib/tables/_p_o_s_t.py
```diff
@@ -252,14 +255,34 @@ def fromXML(self, name, attrs, content, ttFont):
...
+		if (dataLen <= index):
+			length = 0
+		else:
+			length = byteord(data[index])
+		index += 1
+
+		if (dataLen <= index + length - 1):
+			name = ""
+		else:
+			name = tostr(data[index:index+length], encoding="latin1")
```
anthrotype, line 276: if you try accessing a slice of a string which is outside its bounds, you already get an empty string (or a portion of the string that interesect the slice), so the `if` statement above is unnecessary, e.g.

```python
>>> s = "abc"
>>> s[2:6]
'c'
>>> s[3:4]
''
>>> s[4:5]
''
```

Lib/fontTools/ttLib/tables/_p_o_s_t.py
```diff
@@ -252,14 +255,34 @@ def fromXML(self, name, attrs, content, ttFont):
...
 	strings = []
 	index = 0
 	dataLen = len(data)
-	while index < dataLen:
-		length = byteord(data[index])
-		strings.append(tostr(data[index+1:index+1+length], encoding="latin1"))
-		index = index + 1 + length
+
+	for _ in range(n):
+		if (dataLen <= index):
```
anthrotype, line 267: `if` statements in python don't need the parentheses, they are no-op

Lib/fontTools/ttLib/tables/_p_o_s_t.py
```diff
@@ -85,7 +87,8 @@ def decode_format_2_0(self, data, ttFont):
 		indices.frombytes(data[:2*numGlyphs])
 		if sys.byteorder != "big": indices.byteswap()
 		data = data[2*numGlyphs:]
-		self.extraNames = extraNames = unpackPStrings(data)
+		maxIndex = max(indices[:numGlyphs])
```
anthrotype, line 90: I think`max(indices)` would do the same here, because the `indices` local variable is an array of USHORT that is unpacked from a bytes string (`data`) which is maximum `2*numGlyphs` long (i.e. `indices.frombytes(data[:2*numGlyphs])`)

Lib/fontTools/ttLib/tables/_p_o_s_t.py
emuller-amazon: The "if" part handles the case where the length of the string is n but we have < n bytes following it. We could use those bytes as the name, but since the string is not well formed, using "" seems better.

Lib/fontTools/ttLib/tables/_p_o_s_t.py
emuller-amazon: Fixed.

Lib/fontTools/ttLib/tables/_p_o_s_t.py
emuller-amazon: Fixed.

PR #2323, opened 2021-05-25
instancer: Fix AttributeError when ValueRecord has XAdvDevice but no XAdvance

anthrotype:
If ValueRecord has XAdvDevice but no XAdvance, the MutatorMerger currently raises AttributeError as it assumes that whenever one of the Device tables are present, the respective non-device values are also present -- but they might be omitted, and should default to 0. 

Adding the failing test case, followed by the fix itself

### DIFF
--- a/Lib/fontTools/varLib/merger.py
+++ b/Lib/fontTools/varLib/merger.py
@@ -993,7 +993,7 @@ def merge(merger, self, lst):
 		varidx = (dev.StartSize << 16) + dev.EndSize
 		delta = otRound(instancer[varidx])
 
-		setattr(self, name, getattr(self, name) + delta)
+		setattr(self, name, getattr(self, name, 0) + delta)
 
 
 #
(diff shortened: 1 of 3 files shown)


anthrotype:
```AttributeError: 'ValueRecord' object has no attribute 'XAdvance'```

https://github.com/fonttools/fonttools/pull/2323/checks?check_run_id=2664751738#step:5:449

anthrotype:
this issue was introduced with https://github.com/fonttools/fonttools/commit/3a6ebb8dc769d68b4c5fc477d3f28b49204c8b0e, where the subsetter calls ValueRecord.getEffectiveFormat() which drops 0 values from PairPos subtables to make them more compact.
It can be reproduced by taking a VF like [Oswald](https://github.com/google/fonts/blob/main/ofl/oswald/Oswald%5Bwght%5D.ttf), running the subsetter on it first and then running the varLib.instancer.

behdad:
Looks good. Thanks!

anthrotype:
@behdad thanks for taking a look!

PR #2326, opened 2021-05-28
Add GPOS compaction code in varLib.merger and otlLib.builder + new `fonttools otlLib.optimize` command

belluzj:
This PR upstreams to fontTools a way to compact binary GPOS table in TTF, by splitting them into subtables in a smart way. The algorithm can yield significant file size savings depending on the font, e.g. sometimes it removes 50% of the GPOS table, while preserving all the kerning information, leading to no change in the font's spacing, and no reflows in documents.

The idea behind this algorithm is that the GPOS subtables can be big sparse matrices, full of zeros, especially on fonts that cover multiple script with kerning between glyphs of the same script but not across scripts. (see the picture below for illustration purposes: it represents an hypothetical class to class kerning lookup, and shows how large regions without any data naturally appear in cross-script kerning or even within a single script, eg. lowercase against uppercase).

image

We can compact them, by instead of having 1 large sparse subtable, split that subtable into N small subtables, 1 per line, and remove zeros from each line, achieving maximum density/no sparseness, at the expense of the overhead of many subtables. That's the "max" mode of the algorithm.

Then, we can also target a middle ground, where we use clustering to group together the tiny subtables that have the same (or almost the same) columns, so that we reduce the number of new subtables while keeping high density of data. That's the "auto" mode of the algorithm, and it's implemented with an off-the-shelf clustering algorithm from scikit-learn. (see the picture below for a potential solution)

image

The above result could be obtained by splitting into 1 subtable per script, but that requires information about which script each glyph belongs to, and that prevents taking advantage of patterns within a script. That's why we propose to use clustering instead: it doesn't need any information about the glyphs, it only looks at patterns inside the kerning data.

After discussing initial results with Behdad, he suggested that we roll our own clustering algorithm, taking inspiration from this VarStore optimizer at https://github.com/fonttools/fonttools/blob/1df4f1eac1fa185e1b838e1203c92d67fcc41067/Lib/fontTools/varLib/varStore.py#L434-L523 

The custom clustering has several benefits:
* no new dependency
* custom stop criteria = implements a trade-off between file size savings and the proliferation of subtables (too many subtables is harmful to WOFF2 compression and potentially to processing time while rendering the font)
* custom cost function = by predicting almost exactly the outcome in terms of number of bytes of each subtable arrangement, the custom clustering takes the optimal decisions each time and outperforms the off-the-shelf clustering. It also knows to do nothing when nothing interesting can be done (where the off-the-shelf algorithm would still split subtables and actually increase the file size).

The custom clustering is implemented by strategies 1 to 9. The number reflects how many new subtables the algorithm will add in order to achieve a 50% reduction in file size. Strategy 9 is actually allowed to create as many subtables as necessary to create the best file size savings.

The goals of this PR are:
* validate the implementation of the algorithm
* provide it in fontTools as a function that can be applied to existing TTFs to reduce their size (there's a new snippet that does that and also reports timing and size savings as a CSV)
* call the algorithm in various places inside fontTools so that newly compiled fonts are already optimized, with "sane" defaults (no subtable explosion)

@behdad @chrissimpkins @madig 

Progress:
- [x] Provide a way to optimize existing TTFs: there's a new Snippet that you can use like so:
    ```bash
    python Snippets/compact_gpos.py YourFont.ttf
    ```
    It will create compacted copies of the input TTF using various compression levels and print a report about the resulting file sizes and runtimes.
- [x] Hook up the optimizer so that both static fonts and variable fonts get optimized when using fontmake
    - otlLib.builder.PairPosBuilder 
    - varLib.merger
- [x] Profile to make faster
    - There might be still some tricks to use like class slots or different datatypes in the hot loop
    - The cost function is very... costly: my intuition is that instead of being precise about e.g. the Class1/2 bytes, we could use approximations that are faster to compute but still yield good decisions about what to merge. For example the version at https://github.com/fonttools/fonttools/pull/2326/commits/c11ff23f64162717c400a314d8e2e1999261591f was quite fast and giving very similar merge decisions to the latest version that is slow because of the Class1/2 computations.
    - The main loop is O(n^2) in calls to the cost function thanks to memoization of the cost of clusters
        - n lines in the subtable = n initial clusters
        - the first iteration evaluates n^2 "potential" merges, of which we select 1 "definitive" merge
        - each further iteration only evaluates new potential merges, that is: we perform up to n definitive merges = up to n further iterations, each of which triggers the evaluation of n new "potential" merges (new cluster against n other clusters) so also n^2
    - Maybe the cost of the main loop could be reduced by having a heuristic that chealy discards potential merges that won't bring any benefit, without evaluating their precise cost. e.g. we could have a "lower bound cost" that has zero instead of the Class1/2 costs, and if the lower bound cost is already not interesting, no need to consider the full cost (I'm going to implement that to see if that works)
- [x] Decide best way to pass the mode parameter to the various functions that use it (see above) => moved to https://github.com/fonttools/fonttools/issues/2349
    - Currently it's an ENV variable
    - Should it be an attribute on ttLib.TTFont instead?
       - Attach tree of key-values to TTFont, a bit like a YAML configuration file, but with inheritance, so that you can set optimize to the top level and override in lower levels
- [x] Do some further cleanup identified by Behdad:  "Maybe now's time to add TTFont API for that. And clean up all the ClassDef/Coverage code you referenced." https://github.com/fonttools/fonttools/pull/2326#discussion_r646782471
- [x] Remove strategies "one", "auto" and "max" and other dead code
- [x] Remove print() -> use the logger or remove completely
- [x] Understand why strategy "one" was yielding size savings despite doing nothing and fix matching bug => change the default otlLib builder to delete columns with all zero kerning => https://github.com/fonttools/fonttools/pull/2278/files
- [x] Make otlLib.optimize.gpos that can be executed
- [x] Figure out best default value for the mode
    - ~~If it was very fast we could have a middle value like "5"~~ It's not very fast
    - Otherwise default to 0 = disabled
- [x] Add tests
- Further testing: can be done in the future, in case we want to turn the optimization on by default:
  - Run on Roboto
  - Publish evaluation of performance impact of more subtables on hb-shape
  - Check performance with pypy

### DIFF
--- a/Lib/fontTools/misc/intTools.py
+++ b/Lib/fontTools/misc/intTools.py
@@ -1,14 +1,25 @@
-__all__ = ['popCount']
+__all__ = ["popCount"]
 
 
 try:
     bit_count = int.bit_count
 except AttributeError:
+
     def bit_count(v):
-        return bin(v).count('1')
+        return bin(v).count("1")
+
 
 """Return number of 1 bits (population count) of the absolute value of an integer.
 
 See https://docs.python.org/3.10/library/stdtypes.html#int.bit_count
 """
 popCount = bit_count
+
+
+def bit_indices(v):
+    """Return list of indices where bits are set, 0 being the index of the least significant bit.
+
+    >>> bit_indices(0b101)
+    [0, 2]
+    """
+    return [i for i, b in enumerate(bin(v)[::-1]) if b == "1"]
--- a/Lib/fontTools/otlLib/optimize/gpos.py
+++ b/Lib/fontTools/otlLib/optimize/gpos.py
@@ -128,7 +128,9 @@ def _getClassRanges(glyphIDs: Iterable[int]):
 
 # Adapted from https://github.com/fonttools/fonttools/blob/f64f0b42f2d1163b2d85194e0979def539f5dca3/Lib/fontTools/ttLib/tables/otTables.py#L960-L989
 def _classDef_bytes(
-    class_data: List[Tuple[List[Tuple[int, int]], int, int]], class_ids: List[int], coverage=False
+    class_data: List[Tuple[List[Tuple[int, int]], int, int]],
+    class_ids: List[int],
+    coverage=False,
 ):
     if not class_ids:
         return 0
@@ -152,6 +154,10 @@ def cluster_pairs_by_class2_coverage_custom_cost(
     pairs: Pairs,
     compression: int = 5,
 ) -> List[Pairs]:
+    if not pairs:
+        # The subtable was actually empty?
+        return [pairs]
+
     # Sorted for reproducibility/determinism
     all_class1 = sorted(set(pair[0] for pair in pairs))
     all_class2 = sorted(set(pair[1] for pair in pairs))
@@ -229,10 +235,13 @@ def cluster_pairs_by_class2_coverage_custom_cost(
         @property
         def cost(self):
             if self._cost is None:
-                # From: https://docs.microsoft.com/en-us/typography/opentype/spec/gpos#pair-adjustment-positioning-format-2-class-pair-adjustment
                 self._cost = (
-                    # uint16	posFormat	Format identifier: format = 2
+                    # 2 bytes to store the offset to this subtable in the Lookup table above
                     2
+                    # Contents of the subtable
+                    # From: https://docs.microsoft.com/en-us/typography/opentype/spec/gpos#pair-adjustment-positioning-format-2-class-pair-adjustment
+                    # uint16	posFormat	Format identifier: format = 2
+                    + 2
                     # Offset16	coverageOffset	Offset to Coverage table, from beginning of PairPos subtable.
                     + 2
                     + self.coverage_bytes
@@ -267,7 +276,9 @@ def cluster_pairs_by_class2_coverage_custom_cost(
                 # uint16	glyphArray[glyphCount]	Array of glyph IDs — in numerical order
                 + sum(len(all_class1[i]) for i in self.indices) * 2
             )
-            ranges = sorted(chain.from_iterable(all_class1_data[i][0] for i in self.indices))
+            ranges = sorted(
+                chain.from_iterable(all_class1_data[i][0] for i in self.indices)
+            )
             merged_range_count = 0
             last = None
             for (start, end) in ranges:
--- a/Lib/fontTools/otlLib/optimize/gpos.py
+++ b/Lib/fontTools/otlLib/optimize/gpos.py
@@ -16,12 +16,19 @@ log = logging.getLogger("fontTools.otlLib.optimize.gpos")
 
 
 def compact(font: TTFont, mode: str) -> TTFont:
-    # Plan:
+    # Ideal plan:
     #  1. Find lookups of Lookup Type 2: Pair Adjustment Positioning Subtable
     #     https://docs.microsoft.com/en-us/typography/opentype/spec/gpos#lookup-type-2-pair-adjustment-positioning-subtable
     #  2. Extract glyph-glyph kerning and class-kerning from all present subtables
     #  3. Regroup into different subtable arrangements
     #  4. Put back into the lookup
+    #
+    # Actual implementation:
+    #  2. Only class kerning is optimized currently
+    #  3. If the input kerning is already in several subtables, the subtables
+    #     are not grouped together first; instead each subtable is treated
+    #     independently, so currently this step is:
+    #     Split existing subtables into more smaller subtables
     gpos = font["GPOS"]
     for lookup in gpos.table.LookupList.Lookup:
         if lookup.LookupType == 2:
--- a/Lib/fontTools/otlLib/optimize/gpos.py
+++ b/Lib/fontTools/otlLib/optimize/gpos.py
@@ -1,5 +1,5 @@
 import logging
-from collections import defaultdict
+from collections import defaultdict, namedtuple
 from functools import reduce
 from itertools import chain
 from math import log2
@@ -156,6 +156,143 @@ def _classDef_bytes(
     return min(format1_bytes, format2_bytes)
 
 
+ClusteringContext = namedtuple(
+    "ClusteringContext",
+    [
+        "lines",
+        "all_class1",
+        "all_class1_data",
+        "all_class2_data",
+        "valueFormat1_bytes",
+        "valueFormat2_bytes",
+    ],
+)
+
+
+class Cluster:
+    # TODO(Python 3.7): Turn this into a dataclass
+    # ctx: ClusteringContext
+    # indices: int
+    # Caches
+    # TODO(Python 3.8): use functools.cached_property instead of the
+    # manually cached properties, and remove the cache fields listed below.
+    # _indices: Optional[List[int]] = None
+    # _column_indices: Optional[List[int]] = None
+    # _cost: Optional[int] = None
+
+    __slots__ = "ctx", "indices_bitmask", "_indices", "_column_indices", "_cost"
+
+    def __init__(self, ctx: ClusteringContext, indices_bitmask: int):
+        self.ctx = ctx
+        self.indices_bitmask = indices_bitmask
+        self._indices = None
+        self._column_indices = None
+        self._cost = None
+
+    @property
+    def indices(self):
+        if self._indices is None:
+            self._indices = bit_indices(self.indices_bitmask)
+        return self._indices
+
+    @property
+    def column_indices(self):
+        if self._column_indices is None:
+            # Indices of columns that have a 1 in at least 1 line
+            #   => binary OR all the lines
+            bitmask = reduce(int.__or__, (self.ctx.lines[i] for i in self.indices))
+            self._column_indices = bit_indices(bitmask)
+        return self._column_indices
+
+    @property
+    def width(self):
+        # Add 1 because Class2=0 cannot be used but needs to be encoded.
+        return len(self.column_indices) + 1
+
+    @property
+    def cost(self):
+        if self._cost is None:
+            self._cost = (
+                # 2 bytes to store the offset to this subtable in the Lookup table above
+                2
+                # Contents of the subtable
+                # From: https://docs.microsoft.com/en-us/typography/opentype/spec/gpos#pair-adjustment-positioning-format-2-class-pair-adjustment
+                # uint16	posFormat	Format identifier: format = 2
+                + 2
+                # Offset16	coverageOffset	Offset to Coverage table, from beginning of PairPos subtable.
+                + 2
+                + self.coverage_bytes
+                # uint16	valueFormat1	ValueRecord definition — for the first glyph of the pair (may be zero).
+                + 2
+                # uint16	valueFormat2	ValueRecord definition — for the second glyph of the pair (may be zero).
+                + 2
+                # Offset16	classDef1Offset	Offset to ClassDef table, from beginning of PairPos subtable — for the first glyph of the pair.
+                + 2
+                + self.classDef1_bytes
+                # Offset16	classDef2Offset	Offset to ClassDef table, from beginning of PairPos subtable — for the second glyph of the pair.
+                + 2
+                + self.classDef2_bytes
+                # uint16	class1Count	Number of classes in classDef1 table — includes Class 0.
+                + 2
+                # uint16	class2Count	Number of classes in classDef2 table — includes Class 0.
+                + 2
+                # Class1Record	class1Records[class1Count]	Array of Class1 records, ordered by classes in classDef1.
+                + (self.ctx.valueFormat1_bytes + self.ctx.valueFormat2_bytes)
+                * len(self.indices)
+                * self.width
+            )
+        return self._cost
+
+    @property
+    def coverage_bytes(self):
+        format1_bytes = (
+            # From https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#coverage-format-1
+            # uint16	coverageFormat	Format identifier — format = 1
+            # uint16	glyphCount	Number of glyphs in the glyph array
+            4
+            # uint16	glyphArray[glyphCount]	Array of glyph IDs — in numerical order
+            + sum(len(self.ctx.all_class1[i]) for i in self.indices) * 2
+        )
+        ranges = sorted(
+            chain.from_iterable(self.ctx.all_class1_data[i][0] for i in self.indices)
+        )
+        merged_range_count = 0
+        last = None
+        for (start, end) in ranges:
+            if last is not None and start != last + 1:
+                merged_range_count += 1
+            last = end
+        format2_bytes = (
+            # From https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#coverage-format-2
+            # uint16	coverageFormat	Format identifier — format = 2
+            # uint16	rangeCount	Number of RangeRecords
+            4
+            # RangeRecord	rangeRecords[rangeCount]	Array of glyph ranges — ordered by startGlyphID.
+            # uint16	startGlyphID	First glyph ID in the range
+            # uint16	endGlyphID	Last glyph ID in the range
+            # uint16	startCoverageIndex	Coverage Index of first glyph ID in range
+            + merged_range_count * 6
+        )
+        return min(format1_bytes, format2_bytes)
+
+    @property
+    def classDef1_bytes(self):
+        # We can skip encoding one of the Class1 definitions, and use
+        # Class1=0 to represent it instead, because Class1 is gated by the
+        # Coverage definition. Use Class1=0 for the highest byte savings.
+        # Going through all options takes too long, pick the biggest class
+        # = what happens in otlLib.builder.ClassDefBuilder.classes()
+        biggest_index = max(self.indices, key=lambda i: len(self.ctx.all_class1[i]))
+        return _classDef_bytes(
+            self.ctx.all_class1_data, [i for i in self.indices if i != biggest_index]
+        )
+
+    @property
+    def classDef2_bytes(self):
+        # All Class2 need to be encoded because we can't use Class2=0
+        return _classDef_bytes(self.ctx.all_class2_data, self.column_indices)
+
+
 def cluster_pairs_by_class2_coverage_custom_cost(
     font: TTFont,
     pairs: Pairs,
@@ -196,134 +333,14 @@ def cluster_pairs_by_class2_coverage_custom_cost(
     valueFormat1_bytes = bit_count(format1) * 2
     valueFormat2_bytes = bit_count(format2) * 2
 
-    # Agglomerative clustering by hand, checking the cost gain of the new
-    # cluster against the previously separate clusters
-    # Start with 1 cluster per line
-    # cluster = set of lines = new subtable
-    # The class is here so it has a closure over the data above (lines, etc.)
-    class Cluster:
-        # TODO(Python 3.7): Turn this into a dataclass
-        # indices: int
-        # Caches
-        # TODO(Python 3.8): use functools.cached_property instead of the
-        # manually cached properties, and remove the cache fields listed below.
-        # _indices: Optional[List[int]] = None
-        # _column_indices: Optional[List[int]] = None
-        # _cost: Optional[int] = None
-
-        __slots__ = "indices_bitmask", "_indices", "_column_indices", "_cost"
-
-        def __init__(self, indices_bitmask: int):
-            self.indices_bitmask = indices_bitmask
-            self._indices = None
-            self._column_indices = None
-            self._cost = None
-
-        @property
-        def indices(self):
-            if self._indices is None:
-                self._indices = bit_indices(self.indices_bitmask)
-            return self._indices
-
-        @property
-        def column_indices(self):
-            if self._column_indices is None:
-                # Indices of columns that have a 1 in at least 1 line
-                #   => binary OR all the lines
-                bitmask = reduce(int.__or__, (lines[i] for i in self.indices))
-                self._column_indices = bit_indices(bitmask)
-            return self._column_indices
-
-        @property
-        def width(self):
-            # Add 1 because Class2=0 cannot be used but needs to be encoded.
-            return len(self.column_indices) + 1
-
-        @property
-        def cost(self):
-            if self._cost is None:
-                self._cost = (
-                    # 2 bytes to store the offset to this subtable in the Lookup table above
-                    2
-                    # Contents of the subtable
-                    # From: https://docs.microsoft.com/en-us/typography/opentype/spec/gpos#pair-adjustment-positioning-format-2-class-pair-adjustment
-                    # uint16	posFormat	Format identifier: format = 2
-                    + 2
-                    # Offset16	coverageOffset	Offset to Coverage table, from beginning of PairPos subtable.
-                    + 2
-                    + self.coverage_bytes
-                    # uint16	valueFormat1	ValueRecord definition — for the first glyph of the pair (may be zero).
-                    + 2
-                    # uint16	valueFormat2	ValueRecord definition — for the second glyph of the pair (may be zero).
-                    + 2
-                    # Offset16	classDef1Offset	Offset to ClassDef table, from beginning of PairPos subtable — for the first glyph of the pair.
-                    + 2
-                    + self.classDef1_bytes
-                    # Offset16	classDef2Offset	Offset to ClassDef table, from beginning of PairPos subtable — for the second glyph of the pair.
-                    + 2
-                    + self.classDef2_bytes
-                    # uint16	class1Count	Number of classes in classDef1 table — includes Class 0.
-                    + 2
-                    # uint16	class2Count	Number of classes in classDef2 table — includes Class 0.
-                    + 2
-                    # Class1Record	class1Records[class1Count]	Array of Class1 records, ordered by classes in classDef1.
-                    + (valueFormat1_bytes + valueFormat2_bytes)
-                    * len(self.indices)
-                    * self.width
-                )
-            return self._cost
-
-        @property
-        def coverage_bytes(self):
-            format1_bytes = (
-                # From https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#coverage-format-1
-                # uint16	coverageFormat	Format identifier — format = 1
-                # uint16	glyphCount	Number of glyphs in the glyph array
-                4
-                # uint16	glyphArray[glyphCount]	Array of glyph IDs — in numerical order
-                + sum(len(all_class1[i]) for i in self.indices) * 2
-            )
-            ranges = sorted(
-                chain.from_iterable(all_class1_data[i][0] for i in self.indices)
-            )
-            merged_range_count = 0
-            last = None
-            for (start, end) in ranges:
-                if last is not None and start != last + 1:
-                    merged_range_count += 1
-                last = end
-            format2_bytes = (
-                # From https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#coverage-format-2
-                # uint16	coverageFormat	Format identifier — format = 2
-                # uint16	rangeCount	Number of RangeRecords
-                4
-                # RangeRecord	rangeRecords[rangeCount]	Array of glyph ranges — ordered by startGlyphID.
-                # uint16	startGlyphID	First glyph ID in the range
-                # uint16	endGlyphID	Last glyph ID in the range
-                # uint16	startCoverageIndex	Coverage Index of first glyph ID in range
-                + merged_range_count * 6
-            )
-            return min(format1_bytes, format2_bytes)
-
-        @property
-        def classDef1_bytes(self):
-            # We can skip encoding one of the Class1 definitions, and use
-            # Class1=0 to represent it instead, because Class1 is gated by the
-            # Coverage definition. Use Class1=0 for the highest byte savings.
-            # Going through all options takes too long, pick the biggest class
-            # = what happens in otlLib.builder.ClassDefBuilder.classes()
-            biggest_index = max(self.indices, key=lambda i: len(all_class1[i]))
-            return _classDef_bytes(
-                all_class1_data, [i for i in self.indices if i != biggest_index]
-            )
-
-        @property
-        def classDef2_bytes(self):
-            # All Class2 need to be encoded because we can't use Class2=0
-            return _classDef_bytes(all_class2_data, self.column_indices)
-
-        def merge(self, other: "Cluster") -> "Cluster":
-            return make_cluster(self.indices_bitmask | other.indices_bitmask)
+    ctx = ClusteringContext(
+        lines,
+        all_class1,
+        all_class1_data,
+        all_class2_data,
+        valueFormat1_bytes,
+        valueFormat2_bytes,
+    )
 
     cluster_cache: Dict[int, Cluster] = {}
 
@@ -331,10 +348,17 @@ def cluster_pairs_by_class2_coverage_custom_cost(
         cluster = cluster_cache.get(indices, None)
         if cluster is not None:
             return cluster
-        cluster = Cluster(indices)
+        cluster = Cluster(ctx, indices)
         cluster_cache[indices] = cluster
         return cluster
 
+    def merge(cluster: Cluster, other: Cluster) -> Cluster:
+        return make_cluster(cluster.indices_bitmask | other.indices_bitmask)
+
+    # Agglomerative clustering by hand, checking the cost gain of the new
+    # cluster against the previously separate clusters
+    # Start with 1 cluster per line
+    # cluster = set of lines = new subtable
     clusters = [make_cluster(1 << i) for i in range(len(lines))]
 
     # Cost of 1 cluster with everything
@@ -349,7 +373,7 @@ def cluster_pairs_by_class2_coverage_custom_cost(
         best_merged = None
         for i, cluster in enumerate(clusters):
             for j, other in enumerate(clusters[i + 1 :]):
-                merged = cluster.merge(other)
+                merged = merge(cluster, other)
                 cost_change = merged.cost - cluster.cost - other.cost
                 if lowest_cost_change is None or cost_change < lowest_cost_change:
                     lowest_cost_change = cost_change
--- a/Lib/fontTools/otlLib/optimize/gpos.py
+++ b/Lib/fontTools/otlLib/optimize/gpos.py
@@ -380,7 +380,7 @@ def cluster_pairs_by_class2_coverage_custom_cost(
         del clusters[best_other_index]
         clusters[best_cluster_index] = best_merged
 
-    # All clusters are in done; turn bitmasks back into the "Pairs" format
+    # All clusters are final; turn bitmasks back into the "Pairs" format
     pairs_by_class1: Dict[Tuple[str, ...], Pairs] = defaultdict(dict)
     for pair, values in pairs.items():
         pairs_by_class1[pair[0]][pair] = values
--- a/Lib/fontTools/otlLib/optimize/gpos.py
+++ b/Lib/fontTools/otlLib/optimize/gpos.py
@@ -9,6 +9,9 @@ from fontTools.misc.intTools import bit_count, bit_indices
 from fontTools.ttLib import TTFont
 from fontTools.ttLib.tables import otBase, otTables
 
+# NOTE: activating this optimization via the environment variable is
+# experimental and may not be supported once an alternative mechanism
+# is in place. See: https://github.com/fonttools/fonttools/issues/2349
 GPOS_COMPACT_MODE_ENV_KEY = "FONTTOOLS_GPOS_COMPACT_MODE"
 GPOS_COMPACT_MODE_DEFAULT = "0"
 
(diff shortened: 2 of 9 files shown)


belluzj:
Move code to otlLib.optimize

belluzj:
Introduce compression number = 1 to 9, default 5 like gzip for example. See code in https://github.com/harfbuzz/packtab/blob/master/Lib/packTab/__init__.py#L690-L691 for meaning of number (1 = accept to add 1 subtable to reduce size by 50%, 5 = accept to add 5 subtables to reduce size by 50%)

chrissimpkins:
@davelab6

behdad:
Thanks @belluzj. I haven't checked the new code. How does it perform?

Lib/fontTools/otlLib/optimize.py
```diff
@@ -0,0 +1,434 @@
...
+        indices: FrozenSet[int]
+        # Caches
+        # NOTE: (Jany) I wanted to use functools.cached_property but it's Python 3.8+
+        _column_indices: Optional[List[int]] = None
+        _cost: Optional[int] = None
+
+        @property
+        def column_indices(self):
+            if self._column_indices is None:
+                bitmask = bin(reduce(or_, (lines[i] for i in self.indices)))[::-1]
```
behdad, line 227: I've always used `int.__or__` in such cases.

belluzj:
@behdad I ran the new implementation against a few Google Fonts that I selected based on how well the "auto" strategy from scikit-learn was performing on them (a few very bad examples such as Fira Sans where the GPOS table size was going up +50%, and a few very good, such as Pridi where the GPOS was reduced by 70%). On this sample, the new strategies, from 1 to 8, (almost) always result in either no change to GPOS size or significant savings. 

Strategies 1 to 8 are implemented so that while splitting a subtable, they will allow the creation of at most 1 to 8 new subtables to justify a 50% size saving. That seems to work well because when no savings can be achieved it keeps the number of subtables low or even just 1 = falls back to strategy "one" which is to not split the subtable.

Strategy 9 doesn't have that stop criteria: it will create as many subtables as needed to get the max size saving. However strategy 9 fails on the "hard" fonts (where no savings can be achieved) => that indicates that my custom cost function is off (underestimating the overhead of extra subtables). If the cost function was representing reality exactly I would expect strategy 9 to always yield the best savings.

https://docs.google.com/spreadsheets/d/1qUjbvmrGeVxL1_yofX8VTN2HPqvHKVnRqovPUOBSfGk/edit?usp=sharing

image

Note: here I'm running the strategies on final fonts, after the fact; I'll evaluate later what the strategies yield when they run as part of the compilation process (with access to subtables before overflow resolution).

About the time that the new algorithm takes to run, it's 3-5 times slower than the previous strategies. See runtimes below on the sample fonts. I haven't profiled anything or done any performance optimizations (apart from memoizing Cluster instances to avoiding recomputing their cost all the time).

```
    Runtimes, in seconds; each column is one font, each line is a strategy
        one         0.23  0.60  0.04  0.60  0.31  0.13  0.51  1.85  1.30  0.08  0.79  0.63  0.36  0.72
        max         0.26  0.56  0.06  0.61  0.31  0.14  0.56  2.01  1.33  0.25  0.88  0.63  0.37  0.85
        auto        1.30  0.60  0.04  0.66  0.35  0.11  0.57  1.99  1.26  0.08  0.82  0.63  0.39  0.78
        1           0.88  1.90  0.15  2.12  1.04  0.30  1.18  2.59  5.24  0.19  2.18  2.22  1.05  2.08
        2           0.89  1.92  0.15  2.18  1.06  0.31  1.24  2.60  5.33  0.15  2.11  2.15  1.03  2.10
        3           0.88  1.87  0.14  2.06  1.13  0.30  1.21  2.53  5.57  0.19  2.15  2.17  1.06  2.11
        4           0.94  1.92  0.17  2.04  1.04  0.30  1.19  2.61  5.27  0.16  2.11  2.10  0.98  2.08
        5           0.85  1.88  0.15  2.07  1.05  0.31  1.17  2.64  5.55  0.18  2.19  2.14  1.00  2.13
        6           0.86  1.85  0.14  2.05  1.01  0.26  1.17  2.64  5.27  0.15  2.09  2.24  0.95  2.16
        7           0.91  2.01  0.14  2.00  1.04  0.31  1.17  2.54  5.32  0.17  2.11  2.10  0.95  2.10
        8           0.84  1.85  0.17  1.99  1.00  0.29  1.13  2.61  5.16  0.15  2.11  1.97  0.94  2.07
        9           0.75  1.72  0.12  1.85  0.98  0.29  1.03  2.39  4.73  0.16  1.74  1.48  0.90  1.73
```

belluzj:
I fixed a bug in the cost function and the results are looking much better already: there's only on Noto Ethiopic that strategy 9 isn't the best but otherwise all the numbered strategies save bytes, with 9 saving the most as expected. To make the cost function really spot-on I would have to compute which glyph ranges are contiguous or not and choose the best format accordingly (maybe I should actually build the subtables with fontTools and measure how many bytes they take? Instead of trying to just count bytes without actually putting together the data)

image

belluzj:
I updated the PR description to explain in more details what this is about :)

behdad:
> To make the cost function really spot-on I would have to compute which glyph ranges are contiguous or not and choose the best format accordingly (maybe I should actually build the subtables with fontTools and measure how many bytes they take

Right. Is not that bad to count number of ranges. We used to have code to that effect in a couple of places. Let me find... Here:
https://github.com/fonttools/fonttools/commit/3d578dd1dec43b727c32f77f6441ba1681f1c50f

Lib/fontTools/otlLib/optimize.py
```diff
@@ -0,0 +1,451 @@
...
+            subtables.append(
+                buildPairPosClassesSubtable(pairs, font.getReverseGlyphMap())
+            )
+    elif mode == "auto":
+        grouped_pairs = cluster_pairs_by_class2_coverage_agglomerative(all_pairs)
+        for pairs in grouped_pairs:
+            subtables.append(
+                buildPairPosClassesSubtable(pairs, font.getReverseGlyphMap())
+            )
+    elif mode in ("1", "2", "3", "4", "5", "6", "7", "8", "9"):
```
behdad, line 101: You can also write `mode in '123456789'`.

Lib/fontTools/otlLib/optimize.py
```diff
@@ -0,0 +1,451 @@
...
+    ungrouped_lines = []
+    for class1, label in zip(all_class1, labels):
+        if label == -1:
+            ungrouped_lines.append(lines[class1])
+        else:
+            grouped_lines[label].update(lines[class1])
+    return [*grouped_lines.values(), *ungrouped_lines]
+
+
+def _bit_count(n):
```
behdad, line 182: Let's move this (and other useful facilities) into `fontTools.misc.bitTools`.

behdad: Of.. `fontTools.misc.intTools`... which... already has a `popCount` which does the same. We should measure them maybe and pick one.

behdad: Filed https://github.com/fonttools/fonttools/issues/2331

belluzj:
I made it about 2x faster by approximating more: computing the exact valueFormats for each cluster was taking a lot of time, and I'm not sure it is useful. Instead I'm now assuming that all clusters will have the same valueFormats as the full subtable. That may not be true when the subtable a mix of positioning rules with different formats... but on the few sample fonts I was looking at, the results in terms of savings are very similar to yesterday's.

The new timings after that change are:
```
    Runtimes **before** the change (computing the actual value formats)
        one         0.23  0.60  0.04  0.60  0.31  0.13  0.51  1.85  1.30  0.08  0.79  0.63  0.36  0.72
        max         0.26  0.56  0.06  0.61  0.31  0.14  0.56  2.01  1.33  0.25  0.88  0.63  0.37  0.85
        auto        1.30  0.60  0.04  0.66  0.35  0.11  0.57  1.99  1.26  0.08  0.82  0.63  0.39  0.78
        1           0.88  1.90  0.15  2.12  1.04  0.30  1.18  2.59  5.24  0.19  2.18  2.22  1.05  2.08
        2           0.89  1.92  0.15  2.18  1.06  0.31  1.24  2.60  5.33  0.15  2.11  2.15  1.03  2.10
        3           0.88  1.87  0.14  2.06  1.13  0.30  1.21  2.53  5.57  0.19  2.15  2.17  1.06  2.11
        4           0.94  1.92  0.17  2.04  1.04  0.30  1.19  2.61  5.27  0.16  2.11  2.10  0.98  2.08
        5           0.85  1.88  0.15  2.07  1.05  0.31  1.17  2.64  5.55  0.18  2.19  2.14  1.00  2.13
        6           0.86  1.85  0.14  2.05  1.01  0.26  1.17  2.64  5.27  0.15  2.09  2.24  0.95  2.16
        7           0.91  2.01  0.14  2.00  1.04  0.31  1.17  2.54  5.32  0.17  2.11  2.10  0.95  2.10
        8           0.84  1.85  0.17  1.99  1.00  0.29  1.13  2.61  5.16  0.15  2.11  1.97  0.94  2.07
        9           0.75  1.72  0.12  1.85  0.98  0.29  1.03  2.39  4.73  0.16  1.74  1.48  0.90  1.73
    Runtimes **after** the change (assuming the same value formats for all subtables)
        one         0.25  0.82  0.04  0.67  0.34  0.16  0.56  2.18  1.38  0.21  0.87  0.75  0.53  0.83
        max         0.25  0.69  0.04  0.71  0.37  0.13  0.66  2.06  1.21  0.09  0.92  0.76  0.42  0.98
        auto        1.62  0.72  0.05  0.66  0.39  0.18  0.61  2.08  1.39  0.10  0.98  0.74  0.40  0.84
        1           0.47  0.97  0.11  1.15  0.51  0.22  0.91  2.19  1.97  0.11  1.18  1.17  0.63  1.25
        2           0.50  0.95  0.08  1.20  0.51  0.23  1.03  2.29  1.92  0.13  1.14  1.14  0.61  1.14
        3           0.48  1.10  0.08  1.14  0.55  0.19  0.91  2.19  1.84  0.11  1.23  1.24  0.56  1.17
        4           0.46  1.03  0.08  1.11  0.47  0.19  0.95  2.24  1.93  0.09  1.25  1.23  0.60  1.18
        5           0.49  0.95  0.08  1.08  0.54  0.23  0.97  2.31  1.76  0.14  1.36  1.11  0.57  1.28
        6           0.46  0.95  0.12  1.18  0.54  0.21  0.89  2.15  1.92  0.11  1.29  1.18  0.57  1.24
        7           0.49  1.01  0.08  1.12  0.44  0.22  0.97  2.25  1.78  0.13  1.27  1.17  0.58  1.25
        8           0.50  0.95  0.08  1.16  0.54  0.19  0.83  2.15  1.94  0.10  1.24  1.09  0.57  1.18
        9           0.42  0.97  0.08  1.06  0.49  0.23  0.88  2.15  2.03  0.21  1.22  0.98  0.56  1.18
```

Lib/fontTools/otlLib/optimize.py
belluzj: Agreed. About this code in general, I think the mode should come from something else than an environment variable, right? Like maybe a command-line option to fontmake, that goes into a new field on the TTFont object for example. What would fit best the existing behaviours in fontTools?

Lib/fontTools/otlLib/optimize.py
```diff
@@ -0,0 +1,454 @@
...
+    pairs: Pairs,
+    compression: int = 5,
+) -> List[Pairs]:
+    # Sorted for reproducibility/determinism
+    all_class1 = sorted(set(pair[0] for pair in pairs))
+    all_class2 = sorted(set(pair[1] for pair in pairs))
+    # Use Python's big ints for binary vectors representing each line
+    lines = [
+        sum(
+            (1 if (class1, class2) in pairs else 0) << i
```
behdad, line 193: Minor optimization: `(1 << i) if (class1, class2) in pairs else 0`

belluzj:
I added another performance optimization, using int bitmasks instead of frozensets to represent which lines are in a cluster. That reduces timings a tiny bit. Then I also implemented the precise ClassDef format computation and that made everything slow again.

```
    Runtimes using frozensets
        one         0.25  0.82  0.04  0.67  0.34  0.16  0.56  2.18  1.38  0.21  0.87  0.75  0.53  0.83
        max         0.25  0.69  0.04  0.71  0.37  0.13  0.66  2.06  1.21  0.09  0.92  0.76  0.42  0.98
        auto        1.62  0.72  0.05  0.66  0.39  0.18  0.61  2.08  1.39  0.10  0.98  0.74  0.40  0.84
        1           0.47  0.97  0.11  1.15  0.51  0.22  0.91  2.19  1.97  0.11  1.18  1.17  0.63  1.25
        2           0.50  0.95  0.08  1.20  0.51  0.23  1.03  2.29  1.92  0.13  1.14  1.14  0.61  1.14
        3           0.48  1.10  0.08  1.14  0.55  0.19  0.91  2.19  1.84  0.11  1.23  1.24  0.56  1.17
        4           0.46  1.03  0.08  1.11  0.47  0.19  0.95  2.24  1.93  0.09  1.25  1.23  0.60  1.18
        5           0.49  0.95  0.08  1.08  0.54  0.23  0.97  2.31  1.76  0.14  1.36  1.11  0.57  1.28
        6           0.46  0.95  0.12  1.18  0.54  0.21  0.89  2.15  1.92  0.11  1.29  1.18  0.57  1.24
        7           0.49  1.01  0.08  1.12  0.44  0.22  0.97  2.25  1.78  0.13  1.27  1.17  0.58  1.25
        8           0.50  0.95  0.08  1.16  0.54  0.19  0.83  2.15  1.94  0.10  1.24  1.09  0.57  1.18
        9           0.42  0.97  0.08  1.06  0.49  0.23  0.88  2.15  2.03  0.21  1.22  0.98  0.56  1.18

    Runtimes after switch frozenset -> int bitmasks
        one         0.25  0.57  0.04  0.60  0.30  0.13  0.49  1.77  1.41  0.06  0.73  0.73  0.40  0.72
        max         0.24  0.61  0.03  0.63  0.32  0.11  0.55  1.96  1.24  0.10  0.79  0.61  0.38  0.78
        auto        4.34  0.66  0.03  0.89  0.34  0.15  0.57  2.24  1.29  0.07  0.82  0.65  0.34  0.84
        1           0.40  0.89  0.10  1.05  0.51  0.20  0.83  2.23  1.63  0.13  1.10  1.03  0.54  1.10
        2           0.46  0.88  0.08  1.07  0.45  0.19  0.87  2.27  1.65  0.09  1.12  1.03  0.55  1.03
        3           0.42  0.88  0.07  1.07  0.59  0.17  0.85  2.33  1.70  0.10  1.13  0.96  0.56  1.06
        4           0.38  0.89  0.08  0.98  0.48  0.17  0.83  2.08  1.74  0.12  1.10  1.00  0.53  1.11
        5           0.43  0.91  0.07  1.10  0.51  0.21  0.80  2.11  1.59  0.10  1.10  1.00  0.52  1.09
        6           0.39  0.90  0.13  1.01  0.50  0.19  0.83  2.08  1.64  0.12  1.05  1.01  0.56  1.10
        7           0.44  0.86  0.07  0.95  0.43  0.19  0.84  2.11  1.66  0.09  1.08  0.95  0.51  1.07
        8           0.43  0.89  0.07  1.06  0.48  0.16  0.81  2.04  1.71  0.13  1.08  0.93  0.53  1.03
        9           0.39  0.90  0.07  0.96  0.45  0.20  0.78  2.07  1.57  0.09  1.08  0.98  0.52  1.06

    Runtimes after computing ClassDef formats
        one         0.28  0.63  0.04  0.61  0.29  0.11  0.55  1.69  1.53  0.07  0.91  0.78  0.50  0.93
        max         0.23  0.59  0.04  0.62  0.32  0.14  0.55  1.84  1.49  0.08  0.92  0.75  0.44  1.01
        auto        1.44  0.63  0.03  0.65  0.33  0.15  0.53  1.87  1.46  0.13  1.16  0.79  0.41  1.04
        1           2.45  5.50  0.43  3.46  3.13  0.45  1.94  2.92  6.95  0.29  6.00  4.49  2.32  5.31
        2           2.45  5.45  0.43  3.54  3.10  0.43  1.94  3.22  6.45  0.32  5.64  4.22  2.44  5.66
        3           2.45  5.52  0.45  3.46  2.74  0.51  1.95  2.95  6.97  0.31  5.56  4.23  2.43  5.76
        4           2.47  5.55  0.43  3.42  2.86  0.46  1.93  4.77  6.48  0.34  5.61  4.11  2.47  5.83
        5           2.42  5.51  0.42  3.41  2.88  0.48  1.87  3.30  6.62  0.30  5.56  4.07  2.42  5.32
        6           2.39  5.54  0.41  3.39  2.89  0.44  1.87  3.61  6.76  0.34  5.68  4.01  2.31  5.42
        7           2.40  5.52  0.46  3.38  2.76  0.51  2.13  3.60  6.69  0.29  6.17  4.09  2.31  5.32
        8           2.42  5.57  0.42  3.42  2.72  0.51  1.92  3.67  6.51  0.30  5.83  4.18  2.53  5.51
        9           2.43  5.37  0.42  3.50  2.77  0.48  1.85  3.94  6.49  0.35  5.62  3.86  2.49  5.47
```

Lib/fontTools/otlLib/optimize.py
```diff
@@ -180,7 +181,42 @@ def cluster_pairs_by_class2_coverage_agglomerative(
     return [*grouped_lines.values(), *ungrouped_lines]
 
 
+# Adapted from https://github.com/fonttools/fonttools/blob/f64f0b42f2d1163b2d85194e0979def539f5dca3/Lib/fontTools/ttLib/tables/otTables.py#L935-L958
+def _getClassRanges(classes: Iterable[Tuple[str]], font: TTFont):
+    getGlyphID = font.getGlyphID
+    items = []
+    for cls, glyphNames in enumerate(classes):
+        for glyphName in glyphNames:
+            items.append((getGlyphID(glyphName), glyphName, cls))
```
behdad, line 190: `getGlyphID` is killing your performance. See this for how to do it faster:

https://github.com/fonttools/fonttools/blob/f64f0b42f2d1163b2d85194e0979def539f5dca3/Lib/fontTools/ttLib/tables/otConverters.py#L334-L352

Also see:
https://github.com/fonttools/fonttools/issues/1536

Maybe now's time to add `TTFont` API for that. And clean up all the ClassDef/Coverage code you referenced.

behdad:
A bit more work but consider mapping glyphnames to indices at entry and work with integers all through.

Lib/fontTools/otlLib/optimize.py
```diff
@@ -0,0 +1,487 @@
...
+    # Agglomerative clustering by hand, checking the cost gain of the new
+    # cluster against the previously separate clusters
+    # Start with 1 cluster per line
+    # cluster = set of lines
+    # The class is here so it has a closure over lines and effective_value_formats
+    class Cluster:
+        # NOTE: (Jany) I wanted to use a dataclass it's Python 3.7+
+        # indices: int
+        # # Caches
+        # # NOTE: (Jany) I wanted to use functools.cached_property but it's Python 3.8+
```
behdad, line 261: How about using `lru_cache`?

belluzj: The nice thing about @cached_property is that it saves typing but in the end I implemented the cached properties myself by using the _indices fields as the cache for the @property indices. I don't think lru_cache would help here. I could remove that comment actually since I have the caching in place?

behdad: Leave the comment. Maybe we should start using a form like:
```
# TODO(Python 3.7): ...
```
such that when we bump up requirements, we can go clean such items.

Lib/fontTools/otlLib/optimize.py
```diff
@@ -0,0 +1,493 @@
...
+        # TODO(Python 3.7): Turn this into a dataclass
+        # indices: int
+        # Caches
+        # TODO(Python 3.8): use functools.cached_property instead of the
+        # manually cached properties, and remove the cache fields listed below.
+        # _indices: Optional[List[int]] = None
+        # _column_indices: Optional[List[int]] = None
+        # _cost: Optional[int] = None
+
+        # FIXME: (Jany) I'm using __slots__ just in case it provides a saving, I haven't profiled that
```
behdad, line 270: That's fine. No need for a FIXME.

Lib/fontTools/otlLib/optimize.py
```diff
@@ -0,0 +1,493 @@
...
+                # uint16	coverageFormat	Format identifier — format = 1
+                2
+                # uint16	glyphCount	Number of glyphs in the glyph array
+                + 2
+                # uint16	glyphArray[glyphCount]	Array of glyph IDs — in numerical order
+                + sum(len(all_class1[i]) for i in self.indices) * 2
+            )
+
+        @property
+        def classDef1_bytes(self):
```
behdad, line 344: For classDef1, you can use the class=0 smartly, assigning it to whichever class that saves most bytes. That can be implemented in your optimizer.

Lib/fontTools/otlLib/optimize.py
```diff
@@ -0,0 +1,493 @@
...
+    # Sort clusters by numbers of entries in the line?
+    # FIXME: (Jany) I get the intuition that lines with fewer 1's will cluster
+    # together and not with lines with many 1's because that would drive up the
+    # sparseness; however this sorting has no impact on the algorithm below? How
+    # to use that intuition to speed up the algorithm?
+    # clusters = sorted(clusters, key=lambda c: c.width)
+
+    # Cost of 1 cluster with everything
+    # `2 ** len - 1` gives a bitmask full of 1's of length `len`
+    cost_before_splitting = make_cluster(2 ** len(lines) - 1).cost
```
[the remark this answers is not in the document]
behdad:
```
In [1]: %timeit 2**123
496 ns ± 0.359 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [2]: %timeit 1<<123
8.52 ns ± 0.0947 ns per loop (mean ± std. dev. of 7 runs, 100000000 loops each)
```

belluzj:
These are the latest runtimes after my changes today:

```
    Runtimes:
        one         0.37  0.93  0.04  0.82  0.40  0.15  0.71  2.14  1.61  0.09  0.92  0.86  0.46  0.90
        max         0.38  0.87  0.04  0.90  0.41  0.21  0.68  2.37  1.43  0.24  0.93  0.71  0.44  1.06
        auto        1.85  0.89  0.05  0.90  0.43  0.21  0.74  2.36  1.52  0.09  1.02  0.74  0.46  1.03
        1           0.68  1.65  0.15  1.64  0.77  0.25  1.16  2.64  2.52  0.13  1.60  1.41  0.82  1.61
        2           0.70  1.48  0.11  1.59  0.75  0.32  1.11  2.62  2.53  0.17  1.58  1.41  0.89  1.63
        3           0.73  1.58  0.11  1.62  0.73  0.32  1.10  2.56  2.59  0.13  1.55  1.39  0.71  1.57
        4           0.67  1.45  0.12  1.58  0.74  0.32  1.12  2.60  2.63  0.15  1.55  1.33  0.76  1.54
        5           0.67  1.51  0.12  1.60  0.70  0.29  1.08  2.57  2.51  0.13  1.50  1.29  0.71  1.49
        6           0.69  1.47  0.15  1.58  0.74  0.26  1.12  2.64  2.45  0.16  1.52  1.30  0.70  1.52
        7           0.72  1.45  0.10  1.52  0.73  0.33  1.12  2.62  2.53  0.12  1.50  1.36  0.70  1.52
        8           0.77  1.46  0.11  1.60  0.71  0.32  1.09  3.03  2.56  0.17  1.68  1.34  0.70  1.52
        9           0.66  1.60  0.11  1.50  0.78  0.25  1.06  2.50  2.55  0.12  1.77  1.31  0.70  1.56
```

I also updated the spreadsheet at: https://docs.google.com/spreadsheets/d/1qUjbvmrGeVxL1_yofX8VTN2HPqvHKVnRqovPUOBSfGk/edit#gid=1294503886

belluzj:
Make otlLib.optimize.gpos that can be executed

belluzj:
Run on Roboto

belluzj:
Check performance with pypy

belluzj:
change the default otlLib builder to delete columns with all zero kerning

belluzj:
Build graph with binary vectors of columns, build spanning tree in that graph, and only consider edges along the spanning when trying to merge?

Lib/fontTools/misc/intTools.py
```diff
@@ -1,14 +1,25 @@
-__all__ = ['popCount']
+__all__ = ["popCount"]
```
behdad, line 1: Just wondering. Any reason you prefer double-quotes over single? :) Since they are same in Python, one can argue that the single wins since doesn't require shifting in QWERTY. :)

madig: It's the default in black :O and since me and Cosimo want to blacken the source _at some point_....

behdad: I see.

behdad: Do it after this lands. I don't think there's any big outstanding PRs.

madig: I need to fiddle usort and black into place and then the blackening and sorting can begin... maybe when I get the time.

behdad: Is usort the new black? (funny my fingers auto type "ushort" instead :)))

Lib/fontTools/otlLib/optimize.py
```diff
@@ -468,8 +468,7 @@ def make_cluster(indices: int) -> Cluster:
             # See https://github.com/harfbuzz/packtab/blob/master/Lib/packTab/__init__.py#L690-L691
             # Given the size reduction we have achieved so far, compute how many
             # new subtables are acceptable.
-            size_reduction_per_subtable = 1 - 0.5 ** (1 / compression)
-            max_new_subtables = size_reduction / size_reduction_per_subtable
+            max_new_subtables = size_reduction * 2 * compression
```
behdad, line 468: I believe both are wrong in implementing the criteria I had in mind.
```patch
diff --git a/Lib/fontTools/otlLib/optimize.py b/Lib/fontTools/otlLib/optimize.py
index 70390182..a5ec06a1 100644
--- a/Lib/fontTools/otlLib/optimize.py
+++ b/Lib/fontTools/otlLib/optimize.py
@@ -2,6 +2,7 @@ from collections import defaultdict
 import enum
 from functools import reduce
 from itertools import chain
+from math import log2
 from sys import stderr
 from typing import Any, DefaultDict, Dict, Iterable, List, Sequence, Set, Tuple
 
@@ -465,7 +466,7 @@ def cluster_pairs_by_class2_coverage_custom_cost(
             # See https://github.com/harfbuzz/packtab/blob/master/Lib/packTab/__init__.py#L690-L691
             # Given the size reduction we have achieved so far, compute how many
             # new subtables are acceptable.
-            max_new_subtables = size_reduction * 2 * compression
+            max_new_subtables = -log2(1 - size_reduction) * compression
             print(
                 f"            len(clusters) = {len(clusters):3d}    size_reduction={size_reduction:5.2f}    max_new_subtables={max_new_subtables}",
                 file=stderr,
```

Doesn't make much of a difference in the one example I tried.

I suppose you did most of your analysis using `size_reduction` of .5, which mades all three of these equivalent...

Consider `size_reduction` of .75 and the differences should become more obvious.

behdad:
Okay I've taken this for a drive now and fully studied the main loop as well. You've done most of the obvious optimizations indeed :).

Alternative would be to create the cost matrix (maybe in numpy this time!) and modify that matrix by removing row/columns and adding one, to avoid redoing the `n^2` cost calls every round. But since you've memoized those all, not a huge deal.

Let me think about further ideas.

Lib/fontTools/otlLib/optimize.py
behdad: Thinking more, I think something linear like what you had is more adequate here than the log one I had in packtab. Again, it shouldn't matter. But a pathological case is if the matrix was pretty sparse, say, 100x100 matrix with only 100 non-zero entries... are what would we want at any compression level? A hundred subtables is not the best idea...

In reality, that should have been made a PairPosFormat1 perhaps....

behdad: I'll think more about it. Am falling towards maybe just using `max_new_subtables = compression`. But that would push a lot of fonts towards 5 subtables with negligible filesize savings. That's exactly why I did the packtab tradeoff.

Let me think more about it. Maybe comparing to size of the fully-clustered is the wrong approach... Humm. I've confused myself.

behdad: Umm. Thinking more, let's keep the log function.

belluzj: Thanks. What is your reasoning behind the log? So that I can write it in a comment next to the formula in the code

belluzj:
Latest runtimes on my machine, a tiny bit better than last time.
```
    Runtimes:
        1           0.51  1.07  0.09  1.10  0.56  0.22  0.85  2.01  2.05  0.11  1.30  1.29  0.56  1.15
        2           0.51  1.04  0.09  1.11  0.55  0.24  0.92  2.13  2.04  0.23  1.40  1.41  0.58  1.22
        3           0.53  1.13  0.09  1.14  0.54  0.24  0.89  2.08  2.06  0.11  1.32  1.17  0.58  1.31
        4           0.57  1.06  0.09  1.11  0.55  0.21  0.89  2.09  2.89  0.13  1.36  1.12  0.57  1.27
        5           0.53  1.07  0.09  1.09  0.52  0.20  0.88  2.14  2.33  0.10  1.35  1.20  0.63  1.23
        6           0.55  1.02  0.10  1.11  0.55  0.23  0.88  2.08  2.01  0.13  1.38  1.09  0.57  1.21
        7           0.55  1.07  0.11  1.12  0.53  0.21  0.89  2.07  2.11  0.10  1.36  1.12  0.56  1.22
        8           0.52  1.06  0.09  1.08  0.53  0.20  0.88  2.18  2.17  0.13  1.36  1.07  0.56  1.23
        9           0.57  1.08  0.09  1.05  0.52  0.20  0.86  2.11  2.22  0.12  1.37  1.07  0.56  1.32
```

belluzj:
Review from Cosimo:
* Config = flat dict, namespaced keys with first module path and then last word is the name of the option, like a lib => let's call it lib eve? Maybe confusing. Config is better
  * From your module, look it up and you're done
  * We could maintain a global config dictionary in fontTools.config that has default values, and if you create a TTFont without bothering to pass any config, then you get the default values.
  * The dict passed by the user at TTFont() time can be merged (dict.update()) into the default config object
  * Maybe deprecate other parameters (TTFont constructor, "optimize" option to varLib, which is used for something else) and replace them by matching parameters in the new TTFont config.
* Make sure that all top-level public-facing interfaces (TTFont constructor but also varLib.build) where you might want to set configuration and actually allow passing configuration options. Add parameter "fonttools config" in ufo2ft that can be passed to fontTools
* Disable by default; since it's a trade-off (it takes some time to run and WOFF2 size my go slightly up) then it should be up to the user to enable it
* Add comment in the code about hooking it deep in fontTools so that it happens before overflow resolution

belluzj:
TODO: move back to using ENV for this; move the TTFont.Config discussion to another issue.

belluzj:
I added a few tests that show the characteristics of the algorithm and of the mode values on various block matrices.

https://github.com/fonttools/fonttools/pull/2326/commits/df2cc1be479e4e5160956a41825e06bf241f08dd#diff-2f44d7efe42b96d7eca7d77f650692ba03645174de2f65f2954f5d8c28a8a559R129-R148

I now consider that the PR is ready for review and merge, so I've removed the WIP. Since this optimization will be off by default / opt-in, I think that the further performance testing can be done in another PR, as a prerequisite to maybe enabling some optimization by default for everyone.

Lib/fontTools/otlLib/optimize/gpos.py
```diff
@@ -0,0 +1,405 @@
...
+log = logging.getLogger("fontTools.otlLib.optimize.gpos")
+
+
+def compact(font: TTFont, mode: str) -> TTFont:
+    # Plan:
+    #  1. Find lookups of Lookup Type 2: Pair Adjustment Positioning Subtable
+    #     https://docs.microsoft.com/en-us/typography/opentype/spec/gpos#lookup-type-2-pair-adjustment-positioning-subtable
+    #  2. Extract glyph-glyph kerning and class-kerning from all present subtables
+    #  3. Regroup into different subtable arrangements
+    #  4. Put back into the lookup
```
moyogo, line 24: Maybe this needs to be clarified. 
In compact_ext_lookup() when it calls compact_pair_pos(), each subtable is treated independently as it runs compact_class_pairs() for each subtable.
For the description, one would expect the extension lookup subtables to be treated and reorganized together instead of individually. Saying `3. Split subtables into more smaller subtables` would be clearer.

Lib/fontTools/otlLib/optimize/gpos.py
```diff
@@ -0,0 +1,405 @@
...
+        format2 |= value[1].getEffectiveFormat() if value[1] else 0
+    valueFormat1_bytes = bit_count(format1) * 2
+    valueFormat2_bytes = bit_count(format2) * 2
+
+    # Agglomerative clustering by hand, checking the cost gain of the new
+    # cluster against the previously separate clusters
+    # Start with 1 cluster per line
+    # cluster = set of lines = new subtable
+    # The class is here so it has a closure over the data above (lines, etc.)
+    class Cluster:
```
anthrotype, line 197: shouldn't this `Cluster` class be defined outside of nested function scope (unless it captures the latter's state)? otherwise I think it's redefined each time the function is called

Snippets/compact_gpos.py
```diff
@@ -0,0 +1,124 @@
+import argparse
```
anthrotype, line 1: who's the potential user for this script? maybe add a docstring summarizing what this script does, what's it for -- for our future selves.

Tests/otlLib/optimize_test.py
```diff
@@ -0,0 +1,172 @@
...
+        ],
+        check=True,
+    )
+    assert output.exists()
+
+
+# Copy-pasted from https://stackoverflow.com/questions/2059482/python-temporarily-modify-the-current-processs-environment
+# TODO: remove when moving to the Config class
+@contextlib.contextmanager
+def set_env(**environ):
```
anthrotype, line 51: I guess this function could be useful elsewhere, maybe move to misc.testTools?

anthrotype (approved):
I had a brief look at this, though I haven't studied in detail the core of this agglomerative clustering 
algorithm. By just looking at the test results, I can see this is a big improvement, so overall I'm LGTM on merging as is, and maybe improving on it later.

Lib/fontTools/otlLib/optimize/gpos.py
belluzj: It does capture the state, I have put a comment just above to say that it refers to `lines` and other arrays such as `all_class1` etc.

Is there a big performance penalty from using such a nested class? I could maybe make the class "top-level" and make each instance have a pointer to a "data" thing with all the arrays that I'm using, if you think it would be beneficial.

anthrotype: oh sorry I had missed that comment..

> make the class "top-level" and make each instance have a pointer to a "data" thing with all the arrays that I'm using

I guess that would be preferable even only from a code-readability standpoint

Tests/otlLib/optimize_test.py
belluzj: I think this is needed only for this PR now because it's passing around an environment variable, but as soon as there's a Config class in fontTools, the environment variable should go and the associated test too.

Maybe we should clarify somewhere that the environment variable is not a stable feature and that it will be removed eventually? So that if projects start using this feature in their pipelines, they know that at some point they will have to turn it on via the Config class instead?

Or is there the argument that if we merge this with the ENV variable, then it will need to work with the ENV variable forever, even once the Config class is introduced?

Snippets/compact_gpos.py
belluzj: Done: https://github.com/fonttools/fonttools/pull/2326/files#diff-931862a8bb3134a124ab6c069eeee5a0f14905f5e91a0d3c10b3f622a6638d97R3-R19

Lib/fontTools/otlLib/optimize/gpos.py
belluzj: Done, I clarified that the plan was the ideal plan and that the actual implementation was as you pointed out

Tests/otlLib/optimize_test.py
anthrotype: yeah you can mark it as experimental if you like. we can remove support once it becomes a burden

Lib/fontTools/otlLib/optimize/gpos.py
belluzj: Done

Tests/otlLib/optimize_test.py
belluzj: How do I mark as experimental?

anthrotype: just add a comment along those lines (and we can also mention that in the NEWS.rst when we update that)

Issue #2252, opened 2021-03-31
[ttLib] font data get unnecessarily copied in TTFont.__init__


justvanrossum:
TTFont.__init__ will copy the input fule to a BytesIO (unless `lazy=True`). This was done so it's possible to overwrite a file: https://github.com/fonttools/fonttools/commit/356c92341140e7e0a0bce1e4c1ad2a61dc24aeb1

However, later a similar thing was done in `TTFont.save()`: https://github.com/fonttools/fonttools/commit/6ba67ab699d62962690a7a523c286ebf0c8b0ae4 which I *think* makes the former redundant: `font.save()` now always goes through a tmp stream, so overwriting the input should not be a problem.

The current behavior is unnecessarily wasteful in most read-only use cases (not everyone knows to pass `lazy=True` if they just need to read the name table), but is problematic in TTCollection (#2019): if you *don't* pass `lazy=True` to TTCollection, it will _copy the entire ttc/otc data for every font that's in the collection_. Therefore, getting rid of the copy in TTFont.__init__ solves #2019, too.

behdad:
I think the write logic is still wrong. IMO it should create a temp file and then rename it onto the final file. OR, delete the file and then create it.  That is, we don't want to reuse the same file-system inode.

I recommend reconsidering mmap (https://github.com/fonttools/fonttools/pull/581) as well.

justvanrossum:
In the meantime I submitted a PR that fixes another long standing write problem: #2253 for #536. That change improves behavior when compilation fails, and is generally better: the current save has an open empty file sitting during compilation, which can be a long time. Other than that it doesn't change anything and should be a non-controversial change.

behdad:
Yes I noticed, and that's good change. Thanks.

Issue #2019, opened 2020-07-09
[ttLib.ttCollection] TTCollection memory usage


justvanrossum:
See https://twitter.com/ken_lunde/status/1281042388132089856

`TTCollection` copies .ttc data as many times as there are fonts in the .ttc, unless `lazy=True` is passed.

The font "SourceHanNotoCJK.ttc" from https://github.com/adobe-fonts/source-han-super-otc/releases/tag/20190603 contains 216 fonts and is 389 MB big, so opening this font will (try to) use at least 88 GB of memory.

behdad:
@justvanrossum looks like it's already implemented that you can pass `shareTables` to the TTCollection constructor.

behdad:
It's even documented.

behdad:
Sharing tables is implemented. It's the extra copy in TTFont that is killing it.

behdad:
This is really stuck on https://github.com/fonttools/fonttools/pull/2330
there's nothing TTCollection-specific about it.

behdad:
Closing as duplicate of #2330

PR #2330, opened 2021-06-03
[ttFont] remove unncessary copy to BytesIO in TTFont.__init__

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/2252 and https://github.com/fonttools/fonttools/issues/2019

As @justvanrossum noted, TTFont.save already saves to a temporary stream which takes care of the problem with overwriting input file while SFNTReader holds a reference to it.

I add a test first, then follow up with the actual change

### DIFF
--- a/Tests/ttLib/ttFont_test.py
+++ b/Tests/ttLib/ttFont_test.py
@@ -1,10 +1,17 @@
 import io
-from fontTools.ttLib import TTFont, newTable, registerCustomTableClass, unregisterCustomTableClass
+import os
+import tempfile
+from fontTools.ttLib import (
+    TTFont,
+    newTable,
+    registerCustomTableClass,
+    unregisterCustomTableClass,
+)
 from fontTools.ttLib.tables.DefaultTable import DefaultTable
+import pytest
 
 
 class CustomTableClass(DefaultTable):
-
     def decompile(self, data, ttFont):
         self.numbers = list(data)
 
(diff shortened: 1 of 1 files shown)


anthrotype:
hm I have a hunch that this won't work on Windows, because IIRC the latter won't allow to overwrite a file that's already open for reading by an existing process (in this case, we ourselves).

justvanrossum:
I worry what happens *after* the save to the original file, as that invalidates the sfnt reader. If a table was not parsed, the data may have been written directly from the reader IIRC. In other words: after a save to the same file, can we successfully do another save, elsewhere or not? Should the reader then be recreated from the new file?

anthrotype:
well, saving the whole font should read all the input file's tables, no?

> invalidates the sfnt reader.

indeed. Maybe we just set self.reader to None after a TTFont.save and close any open resources it still holds?

justvanrossum:
> well, saving the whole font should read all the input file's tables, no?

Read, yes, but parse, no. Open a font, then save, will barely parse anything, and the raw data from the reader goes straight to the writer without being parsed. See `TTFont.getTableData()`.

justvanrossum:
So if there are any unparsed tables, setting reader to None will make the font not saveable anymore.

anthrotype:
you're right..

> Should the reader then be recreated from the new file?

I'll see if I can do that.

behdad:
> TTFont.save already saves to a temporary stream which takes care of the problem with overwriting input file

I highly strongly whatever suggest that TTFont.save be changed to save to a temporary *file* and then *rename* that file to final destination.

That way, the reader will continue to work, because it's reading from a different, now deleted, file, which is still valid since it has an open handle to it.

anthrotype:
windows can't delete an open file

behdad:
> I highly strongly whatever suggest that TTFont.save be changed to save to a temporary _file_ and then _rename_ that file to final destination.

That's the only way to ensure atomicity. That is: that we won't leave a half-written file in case user presses ctrl-C or we hit an exception or just a power cutoff! This reduces problems with build systems, which would otherwise think the partially-written output is current and need not to be rebuilt.

behdad:
> windows can't delete an open file

I don't see a problem if, on windows, output file has to be different.

anthrotype:
If you look at the tests, it seems that overwriting to the same input file does _not_ fail on Windows (the ttFont_test [passed](https://github.com/fonttools/fonttools/runs/2739380344)). The tests that do fail with permission errors are those that try to delete an open file (with os.remove or indirectly with shutil.rmtree which tries to remove all the files in a directory) when another or same process has an open file handle.
I would think that this practice of deleting font files while a live TTFont instance holds a reference to it might be all too common in the wild. If some of our own tests don't bother using context manager or calling TTFont.close() to clean up open file handles, I'd expect many user codebases do the same.
I can go and fix the tests to make sure they close their TTFonts before tearDown removes the directory they were in, but I worry that this may break others who may unknowingly rely on this "feature"..

All in all, I'm not sure whether we should go on with this change right now. Need to think more about all possible consequences of the change. Will finish this at some point. For now, if one is concerned about memory usage one can already load with lazy=True which skips this eager upfront read/copy.

behdad:
Let's finish this please.

PR #2340, opened 2021-06-17
Add more moments to MomentsPen

n8willis:
This adds several autogenerated moments to MomentsPen.

- 3rd and 4th (planar) moments about the x-axis and y-axis
- 2nd-product (planar) moment (x^2)*(y^2)
- 4th polar moment

The 4th moments were used in David Kindersley & Neil Wiseman's LOGOS software, as described in the now-expired patent, as well as in some later publications. The 2nd and 3rd moments are likely just useful for comparison and research.

Presently this is not hooked up to StatisticsPen, and probably would not be generally-useful-enough to be needed there.


behdad:
Thanks. Before we merge... Does this significantly slow down `StatisticsPen`? I don't remember enough to know without checking further. But looks like it might.

n8willis:
> Thanks. Before we merge... Does this significantly slow down `StatisticsPen`? I don't remember enough to know without checking further. But looks like it might.

Good question. Would it be enough to just time it & compare? Honestly, I'm not sure who uses statisticsPen in real workflows, or how they do — to know how big of an impact would be considered important.

behdad:
Just run it as a module on a couple of fonts.

I'm guessing that this is making it, like, 8 times slower.

After we know, I can advise of alternative API.

behdad:
We should probably make the new code into `HigherMomentsPen`.

n8willis:
> Just run it as a module on a couple of fonts.
> 
> I'm guessing that this is making it, like, 8 times slower.

OK. Difference was negligible (like, down in third-decimal territory) when I used the built-in test set, so I fed it a-zA-Z instead.

It's higher, but just by like 30%....

These are bash `time`, user+sys:
```
Georgia_Italic
orig   higher
.152   .232
.152   .230
.163   .232
.162   .229
.157   .232
-
157.2  231.0 avg

Oxygen Sans
orig   higher
.128   .148
.116   .143
.113   .140
.126   .156
.115   .141
-
119.6  145.6 avg

Noto Serif Reg
orig   higher
.154   .198
.159   .200
.157   .197
.150   .196
.150   .201
-
154.0  198.4 avg

=
143.6  191.67 avg
```

That is a little surprising.... There are a lot of `print()`s in there, though. I'll try timestamps inside a Python script, too

behdad:
That's good news. Thanks.

Was there any of those a CFF font? Just want to see we measure the cubic path as well.

I'm fine merging otherwise.

cc @LettError

n8willis:
> Was there any of those a CFF font? Just want to see we measure the cubic path as well.

No, but I'm sure I've got some handy so I'll just check it.

n8willis:
(okay, not as many as I thought. BUT, Cantarell.... not that different in CLI script test)
```
Cantarell R (CFF)
orig   higher
.151   .206
.150   .187
.151   .189
.154   .190
.152   .191
-
151.6  192.6 avg (1.27x)
```

n8willis:
However, it is a substantially bigger cost just on the statisticspen's `draw` .

I did a minimal script that takes a `perf_counter()` immediately before and after a loop that goes through a `draw()` call on a-zA-Z, and it looks like it's closer to 3x impact if you isolate that:
```
perf_counter script iterating over ascii_letters

Georgia_Italic
orig   higher
.0672  .2020
.0655  .2164
.0643  .2033
.0649  .2053
.0655  .2046
-
654.8  2063.2 = 3.15x

Oxygen Sans
orig   higher
.0253  .0826
.0287  .0788
.0253  .0791
.0276  .0804
.0302  .0819
-
274.2  805.6 = 2.94x

Noto Serif Reg
orig   higher
.0452  .1452
.0532  .1531
.0467  .1499
.0435  .1483
.0441  .1546
-
465.4  1502.2 = 3.23x

CantarellR CFF
orig   higher
.0270  .1233
.0324  .1250
.0266  .1244
.0274  .1219
.0281  .1204
-
283.0  1230 = 4.35x
```

behdad:
Right. 3x is closer to what I thought. Let me devise a plan.

behdad:
> * 4th polar moment

From your code:
```
		      ('momentYYYY', y**4),
		      ('momentXXYY', x*x*y*y),
		      ('momentPolar4', 2*x*x*y*y+x**4+y**4),
```

isn't that one just a simple addition of three other moments we have? In that case we should just make it a property and not accumulate it separately. The code generator is not smart enough to reuse at that leve.

behdad:
[Separately: we should also add type annotations to these functions and see what Cython does with them.]

n8willis:
> isn't that one just a simple addition of three other moments we have? In that case we should just make it a property and not accumulate it separately. The code generator is not smart enough to reuse at that leve.

Yeah; that's actually why I put the XXYY in there. Originally I was cross-checking if the autogenerated moments matched what I came up with on paper, but that process broke when I was attempting the 3rd polar.

behdad:
> Yeah; that's actually why I put the XXYY in there. Originally I was cross-checking if the autogenerated moments matched what I came up with on paper, but that process broke when I was attempting the 3rd polar.

Okay I played with this a bit. Couldn't find an easy way to make the optimizer work better.

I suggest you remove the Polar 4, measure the slowdown of the rest, and if better, we merge.

n8willis:
In that approach, just add Polar 4 to statisticsPen? (and the others, maybe?)

If the speed holds up, that's certainly fine by me. If it's got potential usefulness.

behdad:
> In that approach, just add Polar 4 to statisticsPen? (and the others, maybe?)

Yes.

n8willis:
I think I'll tag this for review when it looks ready; I still need to poke around some I guess.

behdad:
Okay. Thanks for doing the work. I don't have bandwidth to determine what should go in moments and what in statistics currently.

n8willis:
Closing to restart based on #2674.

Issue #2343, opened 2021-06-24
Smoothness testing in GuessSmoothPointPen


takaakifuji:
`GuessSmoothPointPen` looks like passing coordinates to `math.atan2(y, x)` in wrong order:
https://github.com/fonttools/fonttools/blob/b3acb8b984aaa78732d869c356ebd6fa126b000b/Lib/fontTools/pens/pointPen.py#L369-L370
Also, I'm a bit concerned about the choice of the hard-coded threshold value:
https://github.com/fonttools/fonttools/blob/b3acb8b984aaa78732d869c356ebd6fa126b000b/Lib/fontTools/pens/pointPen.py#L371-L372
0.05 rad roughly equals to 2.865 deg, and I'm afraid it would cause too many false positives when handling kinks.
```
>>> upm = 1000.0
>>> math.degrees(math.atan2(1.0, upm))
0.05729576041450061
```
Apart from inflection points, I suppose 0.05 deg should be rather appropriate for my cases. The newer version found in [fontPens](https://github.com/robotools/fontPens/blob/master/Lib/fontPens/guessSmoothPointPen.py) allows me to configure the threshold, thus keeping the code in sync might just fit my needs though.

I believe the code itself hasn't been changed for almost 20 years, thus that makes me skeptical if my guess is valid.

justvanrossum:
Well spotted about the y, x atan2 argument order! Luckily it doesn't make a difference for this case.

I agree that a parametrized tolerance would be better, and perhaps we should sync with the fontPens version. PR welcome.

behdad:
I kept reading "Guess" as "Gauss" and kept being baffled...

PR #2344, opened 2021-06-25
Use newer version of GuessSmoothPointPen from fontPens

takaakifuji:
Allows GuessSmoothPointPen to accept a parameterized tolerance for testing. Note that the code is not exactly identical to the fontPens version to keep the diff minimal. As a bonus, `math.atan2()` call is fixed as well. See also #2343.


justvanrossum (approved):
Looks good to me, thanks!

PR #2346, opened 2021-06-28
COLRv1: rename otData structs to remove 'V1' to match updated draft spec

anthrotype:
This is a breaking change (but the COLRv1 API was already marked as unstable and subject to change)

The changes in this PR are meant to match the changes from the COLRv1 draft spec at:
https://github.com/googlefonts/colr-gradients-spec/pull/302


anthrotype:
at least nanoemoji and fontgoggles will have to be updated accordingly before this is released

rsheeter (approved):

PR #2351, opened 2021-07-01
sort shared tuples produced by Counter.most_common()

garretrieger:
In python 3.7 and earlier most_common() sorts first by the item count and then arbitrarily in the event of ties. This can result in two identical instancing operations producing different binary results.


anthrotype:
thanks Garret!

anthrotype:
btw, the non-determinism is only for python < 3.7; from 3.7 and above, Counter is a dict subclass and thus remembers insertion order.

Actually, on second thought, I think it'd be preferable that we keep the shared tuples sorted by the most frequent first (like we currently do), and only in case of ties we sort lexicographically, i.e. `sorted(c.most_common(), key=lambda i: (-i[1], i[0]))`

PR #2359, opened 2021-07-06
[COLRv1] reduced precision transforms, angles as F2Dot14, centers as FWORD, offsetted Affine2x3

anthrotype:
Updates fonttools' COLRv1 implementation to match the current draft spec:

https://github.com/googlefonts/colr-gradients-spec/pull/307
https://github.com/googlefonts/colr-gradients-spec/pull/310
https://github.com/googlefonts/colr-gradients-spec/pull/313
https://github.com/googlefonts/colr-gradients-spec/pull/305


Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -426,6 +428,27 @@ def fromString(value):
...
+		return angle % math.copysign(360, angle)
+	def read(self, reader, font, tableDict):
+		return super().read(reader, font, tableDict) * 180
+	def write(self, writer, font, tableDict, value, repeatIndex=None):
+		value = self.wrapAround(value)
+		super().write(writer, font, tableDict, value / 180, repeatIndex=repeatIndex)
+	@classmethod
+	def fromString(cls, value):
+		# quantize to nearest multiples of minimum fixed-precision angle
+		return otRound(float(value) / cls.factor) * cls.factor
```
behdad, line 446: I would expect that you call something ala `fixedToFloatToFixed` here as well.

Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -426,6 +428,27 @@ def fromString(value):
...
+	factor = 1.0/(1<<14) * 180  # 0.010986328125
+	@staticmethod
+	def wrapAround(angle):
+		# Upon compile/dump, normalize angles that span beyond a full circle.
+		# We want to keep the orientation after modulo: e.g. -415 => -55, not 305
+		return angle % math.copysign(360, angle)
+	def read(self, reader, font, tableDict):
+		return super().read(reader, font, tableDict) * 180
+	def write(self, writer, font, tableDict, value, repeatIndex=None):
+		value = self.wrapAround(value)
```
behdad, line 441: Why do you modify value (by calling `wrapAround`) before writing? I don't understand.

behdad: The reason I'm objecting is that it can hide bugs / change semantics. If you build masters this way then merge them as variable, this `wrapAround` would silently change the semantics of the rotation. Better fail than corrupt.

Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -426,6 +428,27 @@ def fromString(value):
 	def toString(value):
 		return fl2str(value, 14)
 
+class Angle(F2Dot14):
```
behdad, line 431: This `Angle` class is in degrees; that choice is not reflected in the name. I don't have a good suggestion though.

Lib/fontTools/ttLib/tables/otConverters.py
anthrotype: it's common for fonttools to auto-update things on compile. I don't believe it's changing semantics, because wrapAround keeps the sign while doing modulo operation, so e.g. -415 is compiled as -55, not 305.

anthrotype: but I can change it if you prefer and we raise an exception when angle >= 360 or <= -360. We'll have to move the wrapping upstream

Lib/fontTools/ttLib/tables/otConverters.py
anthrotype: what about `DegreeAngle`?

Lib/fontTools/ttLib/tables/otConverters.py
anthrotype: you meant `floatToFixedToFloat` or `strToFixedToFloat`. The code above is basically doing that, but on a value which represents degrees [-360..360), not fixed-point values [-2.0..2.0) with exact `precisionBits` (the second parameter to the above mentioned functions).

in other words, my code above is doing this:

```python
fixedToFloat(floatToFixed(float(value) / 180, 14), 14) * 180
```

I can change to this if it's preferable

Lib/fontTools/ttLib/tables/otConverters.py
behdad: Either that or AngleDegrees. Both are ugly. Maybe just Angle is fine.

Lib/fontTools/ttLib/tables/otConverters.py
behdad: > I don't believe it's changing semantics, because wrapAround keeps the sign while doing modulo operation, so e.g. -415 is compiled as -55, not 305.

I don't see what significance the -55 vs 305 has.

Here's my deal: imagine upstream sources produce a rotation of 300 in one master, and 390 in another master. Going from first master to second, you expect a 90degree rotation counterclockwise.

But if you silently encode 390 as 30, then the rotation will become 270degrees rotation clockwise. That's the semantic change I'm talking about.

Better err than allow this.

behdad: I don't even think you should do any wrapping upstream.

Lib/fontTools/ttLib/tables/otConverters.py
rsheeter: Perhaps the name `Angle` with a comment indicating it's degrees and describing how it maps on read/write with a link to learn more?

Lib/fontTools/ttLib/tables/otConverters.py
rsheeter: If you know you are compiling a static font wrap is safe (though ... then your sources are odd and perhaps you should fix those!). Which I guess means I agree, don't wrap here, error :)

anthrotype: > I don't see what significance the -55 vs 305 has.

i would think one is clockwise, the other counter-clockwise, given the sign (but perhaps that doesn't matter as the end result is the same?)

> imagine upstream sources produce a rotation of 300 in one master, and 390 in another master.. 

if the "sources" are binary TTFs to be merged into a VF (like varLib currently works for the rest of the variable tables) it's technically impossible to encode 390 in the master COLR binary table, it has to be 390 % 360 to fit the limits of F2Dot14-as-angle.

anthrotype: i removed the wrapping of the angle in https://github.com/fonttools/fonttools/pull/2359/commits/60f072414478dba95c743e8f6bbdb5792088c71b

Lib/fontTools/ttLib/tables/otConverters.py
anthrotype: I kept the name `Angle` and added a comment in 60f072414478dba95c743e8f6bbdb5792088c71b

Lib/fontTools/misc/roundTools.py
```diff
@@ -56,3 +56,50 @@ def roundFunc(tolerance, round=otRound):
...
+    if not value:
+        return "0.0"
+
+    value = otRound(value / factor) * factor
+    eps = .5 * factor
+    lo = value - eps
+    hi = value + eps
+    # If the range of valid choices spans an integer, return the integer.
+    if int(lo) != int(hi):
+        return str(float(round(value)))
```
rsheeter, line 93: why do you need float()?

rsheeter (approved):

Lib/fontTools/misc/roundTools.py
anthrotype: this was intentional. I need the return value to be formatted as a float (180.0 instead of 180), this was the convention so far in ttx dumps for fields that are encoded as fixed-point values. It was already like this in the original code (from which this function is factored out) https://github.com/fonttools/fonttools/blob/12d66ded1230c46a0e62afedfe0a16bd99d78105/Lib/fontTools/misc/fixedTools.py#L145

PR #2379, opened 2021-07-28
[colrv1] otData: define ClipList, ClipBox, etc.

anthrotype:
Syncs COLRv1 implementation with latest draft spec, incorporating changes since https://github.com/googlefonts/colr-gradients-spec/pull/329 and following PRs


Lib/fontTools/colorLib/builder.py
```diff
@@ -230,13 +239,39 @@ def buildCOLR(
...
+    n = len(clipBox)
+    clip = ot.ClipBox()
+    if n < 4 or n > 5:
+        raise ValueError(f"Invalid ClipBox: expected 4 or 5 values, found {n}")
+    clip.xMin, clip.yMin, clip.xMax, clip.yMax = intRect(clipBox[:4])
+    if n == 5:
+        clip.Format = 1
+        clip.VarIndexBase = int(clipBox[4])
+    else:
+        clip.Format = 0
```
rsheeter, line 271: Optional, shorter

```
clip.Format = int(n == 5)
if clip.Format == 1:
  clip.VarIndexBase = ...
```

Lib/fontTools/colorLib/builder.py
```diff
@@ -230,13 +239,39 @@ def buildCOLR(
...
+    clipList.clips = {name: buildClipBox(box) for name, box in clipBoxes.items()}
+    return clipList
+
+
+def buildClipBox(clipBox: _ClipBoxInput) -> ot.ClipBox:
+    if isinstance(clipBox, ot.ClipBox):
+        return clipBox
+    n = len(clipBox)
+    clip = ot.ClipBox()
+    if n < 4 or n > 5:
```
rsheeter, line 264: Optional, n not in (4,5) [or {4,5}] reads more easily to me

Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -1297,6 +1298,144 @@ def preWrite(self, font):
...
+					glyph = glyphOrder[glyphID]
+				except IndexError:
+					continue
+				if glyph not in clips:
+					clips[glyph] = copy.copy(rec.ClipBox)
+				else:
+					rangesOverlap = True
+			if rangesOverlap:
+				log.warning(
+					"ClipRecord %i overlap previous records; "
```
rsheeter, line 1332: specifying the gids or names or both might be handy when troubleshooting such a font?

Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -1297,6 +1298,144 @@ def preWrite(self, font):
...
+	def postRead(self, rawTable, font):
+		clips = {}
+		glyphOrder = font.getGlyphOrder()
+		for i, rec in enumerate(rawTable["ClipRecord"]):
+			rangesOverlap = False
+			for glyphID in range(rec.StartGlyphID, rec.EndGlyphID + 1):
+				try:
+					glyph = glyphOrder[glyphID]
+				except IndexError:
+					continue
```
rsheeter, line 1325: Shouldn't we warn about it? - seems sort of equally broken as overlap and we logged that

Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -1297,6 +1298,144 @@ def preWrite(self, font):
...
+					"redefined clip boxes are skipped",
+					i,
+				)
+		self.clips = clips
+
+	def groups(self):
+		glyphsByClip = defaultdict(list)
+		uniqueClips = {}
+		for glyphName, clipBox in self.clips.items():
+			key = hash(clipBox.as_tuple())
```
rsheeter, line 1342: noob question, why is it necessary to call hash here rather than just using the tuple directly as the key?

Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -1297,6 +1298,144 @@ def preWrite(self, font):
...
+	def preWrite(self, font):
+		if not hasattr(self, "clips"):
+			self.clips = {}
+		clipBoxRanges = {}
+		glyphMap = font.getReverseGlyphMap()
+		for glyphs, clipBox in self.groups().items():
+			glyphIDs = sorted(
+				glyphMap[glyphName] for glyphName in glyphs
+				if glyphName in glyphMap
+			)
```
rsheeter, line 1360: aren't we hiding an error here? We don't need to fail or warn?

rsheeter (approved):

Lib/fontTools/ttLib/tables/otTables.py
justvanrossum: I think it's even wrong, as `hash()` is not guaranteed to be unique. `clipBox.as_tuple()` should be suitable as a dict key.

anthrotype: Tuple is hashed by value so there should not be any difference. I thought hash() better conveyed the intent, but I'm ok to change to use a bare tuple (the hash function will be used internally by the dict type).

anthrotype: A tuple of ints at least is hashed by value

rsheeter: Typically hashes can overlap and you have to do a further search among the things with the same hash to establish a full match if there are >1 of them. I initially thought maybe this was some sort of Python trick but from Just comment it sounds like it works the normal way.

justvanrossum: > Tuple is hashed by value

Yes, but that still doesn't guarantee it's unique. The job of hash is to make collisions unlikely, but any collisions will be dealt with by dict or set.

anthrotype: Got it, thanks for the clarification 👍

Lib/fontTools/ttLib/tables/otTables.py
anthrotype: I actually did this on purpose, so that if any of the base glyphs are removed in a subset operation the respective clipBox is also silently dropped. Alternatively I can have it blow up, and require whomever modifies the glyphOrder (renames or subsets) also update the COLR ClipList.

anthrotype: other tables follow a similar pattern, e.g. hmtx.compile() method only loops over the ttFont.getGlyphOrder() (which may have changed since the table was decompiled/created), effectively skipping any glyph metrics whose glyphName is not in the current glyphOrder.

Lib/fontTools/ttLib/tables/otTables.py
anthrotype: i'm warning now

Lib/fontTools/ttLib/tables/otTables.py
anthrotype: i'm now logging the overlapping redefined ranges

behdad:
I might be late to this party... but can we call ClipRect instead of ClipBox? cc @rsheeter

anthrotype:
well, a "box" _is_ still a rectangle, the meaning is the same..

behdad:
> well, a "box" _is_ still a rectangle, the meaning is the same..

To me "box" has higher-dimension connotation. Anyway.

Issue #2363, opened 2021-07-13
Instancer failing to remove overlaps


nathan-williams:
(fonttools v4.25.0)

The following instancing command fails: `fonttools varLib.instancer ./NotoSansEthiopic\[wdth\,wght\].ttf wdth=87.5 wght=800 --remove-overlaps`.

The font I was attempting to instance: [NotoSansEthiopic[wdth,wght].ttf.zip](https://github.com/fonttools/fonttools/files/6812558/NotoSansEthiopic.wdth.wght.ttf.zip).

Here's the stack trace:

```
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/fontTools/ttLib/removeOverlaps.py", line 110, in _simplify
    path = pathops.simplify(path, clockwise=path.clockwise)
  File "src/python/pathops/_pathops.pyx", line 1476, in pathops._pathops.simplify
  File "src/python/pathops/_pathops.pyx", line 1487, in pathops._pathops.simplify
pathops._pathops.PathOpsError: operation did not succeed

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/bin/fonttools", line 8, in <module>
    sys.exit(main())
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/fontTools/__main__.py", line 30, in main
    runpy.run_module(mod, run_name='__main__')
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 208, in run_module
    return _run_code(code, {}, init_globals, run_name, mod_spec)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/fontTools/varLib/instancer/__main__.py", line 5, in <module>
    sys.exit(main())
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/fontTools/varLib/instancer/__init__.py", line 1415, in main
    updateFontNames=options.update_name_table,
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/fontTools/varLib/instancer/__init__.py", line 1251, in instantiateVariableFont
[...]

PR #2382, opened 2021-07-29
instancer: add --ignore-overlap-errors option

anthrotype:
This is to be able to ignore tricky glyphs that sometimes trip up Skia PathOps.Simplify operation.
We have no idea how to fix this upstream (short of having the glyphs redrawn), and perfect is the enemy of good..

Works around https://github.com/fonttools/fonttools/issues/2363 and https://github.com/google/fonts/issues/3365

/cc @nathan-williams


justvanrossum:
(I remember working around this by falling back to booleanOperations at some point...)

anthrotype:
but booleanOperations doesn't support quadratic curves as far as I know

justvanrossum:
I think you're right about that.

PR #2384, opened 2021-07-30
Default to "\n" for newlinestr instead of None

khaledhosny:
If newlinestr is None, os.linesep is used, bu it is the third millennium
and we don’t need or want different line endings per-platform.

### DIFF
--- a/Lib/fontTools/misc/xmlWriter.py
+++ b/Lib/fontTools/misc/xmlWriter.py
@@ -11,7 +11,7 @@ INDENT = "  "
 class XMLWriter(object):
 
 	def __init__(self, fileOrPath, indentwhite=INDENT, idlefunc=None, encoding="utf_8",
-			newlinestr=None):
+			newlinestr="\n"):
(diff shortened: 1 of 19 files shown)


anthrotype:
I'm ok to change the default to be `"\n"` on all platforms, but why deprecate the parameter and discard the value passed in altogether? If one wants to continue using CRLF (or maybe even an empty string could be a valid value in some situations), why should we stop them from doing this? I don't think it costs us anything to continue supporting the option

khaledhosny:
> but why deprecate the parameter and discard the value passed in altogether?

Because it is not something FontTools should have ever been concerned about. If someone wants to use different line endings for whatever reasons, there are no shortage of tools that can convert between line breaks. By disabling this functionality we can later remove the API entirely if we get to break the API. Just like how XMLWriter no longer support any encoding other than UTF-8 (it will raise if any other encoding is used, but line spacing is more minor IMHO so I just warn and ignore).

khaledhosny:
Feel free to discard the last two commits, though, that is why I committed these changes separately.

PR #2390, opened 2021-08-09
[COLRv1] Add missing CompositeMode.PLUS

anthrotype:
Our enumeration of `CompositeMode` was missing one `COMPOSITE_PLUS` operator (12)... oops. 

see 'PaintComposite' paragraph in https://github.com/googlefonts/colr-gradients-spec/blob/main/OFF_AMD2_WD.md

Thanks @qxliu76 for spotting it!


anthrotype:
https://github.com/googlefonts/colr-gradients-spec/issues/208

PR #2398, opened 2021-08-19
Clean up virtual GID handling

behdad:


behdad:
The first commit removes warnings and all. @khaledhosny maybe you can add your font as test case?

I shall speed this up by way of bulk glyphID-to-glyphName API on ttFont. There was an issue for it, I cannot find anymore.

khaledhosny:
> @khaledhosny maybe you can add your font as test case?

I added a test based on that font to this branch.

khaledhosny:
> I shall speed this up by way of bulk glyphID-to-glyphName API on ttFont. There was an issue for it, I cannot find anymore.

https://github.com/fonttools/fonttools/pull/1654?

behdad:
> #1654?

Yes. Thanks. I'll do that next.

khaledhosny:
You forced pushed and lost my tests commit, do you want me to push it back to this branch?

Lib/fontTools/ttLib/tables/_c_m_a_p.py
```diff
@@ -14,15 +14,12 @@
 
 def _make_map(font, chars, gids):
 	assert len(chars) == len(gids)
+	glyphNames = font.getGlyphNameMany(gids)
 	cmap = {}
 	glyphOrder = font.getGlyphOrder()
```
dscorbett, line 19: `glyphOrder` is no longer needed here.

Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -1222,8 +1216,7 @@ def _readTransition(self, reader, entryIndex, font, actionReader):
 	def _readLigatures(self, reader, font):
 		limit = len(reader.data)
 		numLigatureGlyphs = (limit - reader.pos) // 2
-		return [font.getGlyphName(g)
-		        for g in reader.readUShortArray(numLigatureGlyphs)]
+		return font.getGlyphNameMany (reader.readUShortArray(numLigatureGlyphs))
```
dscorbett, line 1212: There shouldn’t be a space before the parenthesis.

Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -927,9 +926,10 @@ def postRead(self, rawTable, font):
 				cls = rec.Class
 				startID = font.getGlyphID(start)
 				endID = font.getGlyphID(end) + 1
-				for glyphID in range(startID, endID):
-					if cls:
-						classDefs[font.getGlyphName(glyphID)] = cls
+				glyphNames = font.getGlyphNameMany(range(startID, endID))
+				if cls:
```
dscorbett, line 929: If `not cls`, then `glyphNames` is unused, so `glyphNames` should only be defined within this block.

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -570,8 +582,9 @@ def getReverseGlyphMap(self, rebuild=False):
 	def _buildReverseGlyphOrderDict(self):
 		self._reverseGlyphOrderDict = d = {}
 		glyphOrder = self.getGlyphOrder()
-		for glyphID in range(len(glyphOrder)):
-			d[glyphOrder[glyphID]] = glyphID
+		for glyphID,glyphName in enumerate(self.getGlyphOrder()):
```
dscorbett, line 585: You should use `glyphOrder` here instead of calling `self.getGlyphOrder` again, or remove the `glyphOrder` variable.

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -556,12 +558,22 @@ def getGlyphID(self, glyphName):
 					except (NameError, ValueError):
 						raise KeyError(glyphName)
 
-		glyphID = d[glyphName]
 		if glyphName != glyphOrder[glyphID]:
```
dscorbett, line 561: Under what circumstances would this condition be true? That is, how can the glyph order list and the reverse map get unsynchronized?

dscorbett (changes requested):
I installed this dev version of fontTools. My font’s build script produced an identical font file. There was no noticeable difference in speed.

Lib/fontTools/ttLib/ttFont.py
behdad: Only if client modifies the glyphOrder list manually after it has already been used to build reverse list.

behdad:
> There was no noticeable difference in speed.

Thanks for checking. That's expected, as I had put the fastpath in the main callsites. They just call new API now. Good to know that didn't regress.

behdad:
@anthrotype This is ready for your review & merge. Thanks.

Lib/fontTools/ttLib/ttFont.py
behdad: > Does that need to be supported? In [#1654 (comment)](https://github.com/fonttools/fonttools/pull/1654#issuecomment-505541892), @anthrotype wrote:
> 
> > Direct usages of glyphOrder attribute outside of get/setGlyphOrder were never supported. I don't think even fontTools library does that internally, even though technically it could. So I am not worried too much if user code is relying on that attribute.

I just wasn't convinced, and I given lack of documentation, I wasn't sold on @anthrotype's assertion and didn't want to possibly break that.

That said, I checked AFDKO and couldn't find any trace of that.

I'll add a commit to address that now. Yay.

In that case, I suggest we make `self.glyphOrder` be immutable.  I'll see if there's such a thing.  Also, we should make glyphOrder a property.

behdad: > I'll add a commit to address that now. Yay.
> 
> In that case, I suggest we make `self.glyphOrder` be immutable. I'll see if there's such a thing. Also, we should make glyphOrder a property.

Can we do these in a separate PR?

behdad: > I suggest we make `self.glyphOrder` be immutable. I'll see if there's such a thing

I meant other than `tuple`. But I suppose `tuple` works.

behdad: @dscorbett @anthrotype actually `GlyphOrder`'s own `fromXML` does this:
```
        def fromXML(self, name, attrs, content, ttFont):
                if not hasattr(self, "glyphOrder"):
                        self.glyphOrder = []
                        ttFont.setGlyphOrder(self.glyphOrder)
                if name == "GlyphID":
                        self.glyphOrder.append(attrs["name"])
```

Fixing up..

behdad: > Fixing up..

Not sure if possible without changing the XML machinery we have though.

behdad:
Okay, @dscorbett @anthrotype please review again. `setGlyphOrder()` must be called whenever glyphOrder is changed. I didn't change it to tuple because that would have made `fromXML` O(n^2).

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -544,67 +530,36 @@ def getGlyphNames2(self):
...
+			return "glyph%.5d" % glyphID
+
+	def getGlyphNameMany(self, lst):
+		glyphOrder = self.getGlyphOrder();
+		cnt = len(glyphOrder)
+		return [glyphOrder[gid] if gid < cnt else "glyph%.5d" % gid
+			for gid in lst]
+
+	def getGlyphID(self, glyphName):
 		glyphOrder = self.getGlyphOrder()
```
dscorbett, line 546: This variable is unused.

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -429,6 +413,8 @@ def get(self, tag, default=None):
 
 	def setGlyphOrder(self, glyphOrder):
 		self.glyphOrder = glyphOrder
+		if hasattr(self, '_reverseGlyphOrderDict'):
+			delattr(self, '_reverseGlyphOrderDict')
```
dscorbett, line 417: This can be simplified to `del self._reverseGlyphOrderDict`.

dscorbett (changes requested):

behdad:
Fixed. Thanks!

anthrotype:
Thanks @behdad for reducing our tech debt, and thanks @dscorbett and @khaledhosny for the reviews while I was away 👍

PR #2399, opened 2021-08-19
MNT: do not warn on import

tacaswell:
The py23 module is deprecated, but still heavily used within fontTools.  This
means that libraries which depend on fontTools will get a deprecation warning
that they can do nothing about on import of fontTools.

This does three things:

 - move the actual py23 implementation into a private module
 - changes all of the internal imports to use the private module
 - set the stack level in the py23 warning to warn where it is imported

In particular Matplotlib has recently picked up a dependency on fontTools and we have to special-case the warning from importing fontTools in our test suite (we otherwise fail on warnings).

I've added my name to the README as the contributing guides requests, but this is such a mechanical change, I'd be perfectly happy to not do so.

### DIFF
--- a/Lib/fontTools/agl.py
+++ b/Lib/fontTools/agl.py
@@ -26,7 +26,7 @@ This is used by fontTools when it has to construct glyph names for a font which
 doesn't include any (e.g. format 3.0 post tables).
 """
 
-from fontTools.misc.py23 import tostr
+from fontTools.misc._py23 import tostr
 import re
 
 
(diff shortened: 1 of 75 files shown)


khaledhosny:
As it happens, I’m working on this right now! But instead of moving the module to a private one, I’m moving the functions to textTools module and changing the imports.

tacaswell:
Fair enough, I went with the sledge hammer approach ⚒️ to solve by down-stream problem!

I'm going to close this but hold of deleting the branch for a bit in case you do want it.

PR #2400, opened 2021-08-19
Drop use of py23 module intenally

khaledhosny:
Move the rest of `py23` module to `textTools`, and change all imports to use `textTools` module, except the `test_py23.py` test which is kept until we decide to remove the module (if ever).

### DIFF
--- a/Doc/source/misc/index.rst
+++ b/Doc/source/misc/index.rst
@@ -25,7 +25,6 @@ utilities by fontTools, but some of which may be more generally useful.
    psCharStrings
    psLib
    psOperators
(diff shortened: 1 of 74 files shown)


anthrotype (approved):
Thanks Khaled, LGTM!
FYI, I'll be away until September, if you need to cut a release earlier you should be able to follow instructions I put in the README

khaledhosny:
>  FYI, I'll be away until September, if you need to cut a release earlier you should be able to follow instructions I put in the README

Thanks, enjoy your time.

PR #2411, opened 2021-09-20
Fix overflow check (again)

simoncozens:
My fix to @anthrotype's review comment in #2404 was premature. If a lookup could not be expressed as a particular format, the value in the `candidates` array would be `None`. The overflow check would try building it *anyway*, leading to ridiculous warnings:

```
WARNING: Contextual format 1 at foo.fea:5:5 overflowed ('NoneType' object is not iterable)
```

This skips over non-existent subtables when doing the overflow check.


anthrotype:
thanks for the quick fix!

Issue #2349, opened 2021-06-29
Add a facility to pass options to various fontTools modules


belluzj:
While working on https://github.com/fonttools/fonttools/pull/2326 where I need to pass a configuration option deep into the guts of fontTools, Behdad identified that fontTools has a general need for a facility to pass such configuration options from top-level API deep into the various modules that use the options.

Summary of desired features:
* Read conf from a file
* Get conf from tools that rely on fontTools (ufo2ft/fontmake/...)
* Get conf from scripts that call fontTools APIs
* Have global / default values
* Have per-TTFont values (TTFont.config attribute)
* Use conf in any fontTools module
* Potentially have a cascading fallback system (see below)

Opinions so far:

Behdad: I'm suggesting that we add a cfg parameter to TTFont such that various modules can be configured. Eg. by configuring "fontTools.otlLib.optimize.gpos.optimize = 9"; ideally that would be inherited / cascade, such that if not set, it will fall back to "fontTools.otlLib.optimize" to "fontTools.optimize" to "optimize". Something like that. Maybe "optimize" is not the best name and compressionLevel is (for 1-9). etc. But that's the idea.

Cosimo: 
* I think I'd prefer not to add a third-party dependency and try using built-in configparser only
* I think we can simply define one configparser section for each modue that has any configuration (e.g. "otLib.optimize.gpos", omitting the leading "fontTools"), containing the name=value for the module-level global variables, e.g. "optimize = 9"
* we define the configparser logic in a new fontTools.config module, initialized every time the fontTools top-level module is imported
* individual TTFonts can have a copy of the global configuration in that `cfg` attribute you suggested, so one could override individual settings at the font level
* we search for a "fonttools.cfg" first in the same fontTools package directory (we embed a pre-populated default config in the installed package), then in default locations (e.g. ~/.config/fonttools/ or ~/.local/share/fonttools/, /usr/local/share/fonttools/, etc.)
* the fontTools.config.Config class could be a dict-like object where the keys (sections) are the names of the modules (without "fontTools." prefix), the values other dicts keyed by the variable name to be configured
* and the string values parsed to int, float, bool, etc.
* About exposing it to higher-up modules, most of them have an entrypoint that builds a new TTFont from scratch, so doesn't have a TTFont instance where to put the options at first, so we need a nice way to pass the options from that entrypoint (e.g. ufo2ft's compile() or varLib's main) down to the TTFont when it starts to exist

[...]

PR #2416, opened 2021-09-24
Add config module as per #2349

belluzj:
See issue #2349

### DIFF
--- a/Lib/fontTools/config/__init__.py
+++ b/Lib/fontTools/config/__init__.py
@@ -1,0 +1,41 @@
+"""
+Define all configuration options that can affect the working of fontTools
+modules. E.g. optimization levels of varLib IUP, otlLib GPOS compression level,
+etc. If this file gets too big, split it into smaller files per-module.
+
+An instance of the Config class can be attached to a TTFont object, so that
+the various modules can access their configuration options from it.
+"""
+from textwrap import dedent
+
+from fontTools.misc.configTools import *
+
+
+class Config(AbstractConfig):
+    options = Options()
+
+OPTIONS = Config.options
+
+Config.register_option(
+    name="fontTools.otlLib.optimize.gpos:COMPRESSION_LEVEL",
+    help=dedent(
+        """\
+        GPOS Lookup type 2 (PairPos) compression level:
+            0 = do not attempt to compact PairPos lookups;
+            1 to 8 = create at most 1 to 8 new subtables for each existing
+                subtable, provided that it would yield a 50%% file size saving;
+            9 = create as many new subtables as needed to yield a file size saving.
+        Default: 0.
+
+        This compaction aims to save file size, by splitting large class
+        kerning subtables (Format 2) that contain many zero values into
+        smaller and denser subtables. It's a trade-off between the overhead
+        of several subtables versus the sparseness of one big subtable.
+
+        See the pull request: https://github.com/fonttools/fonttools/pull/2326
+        """
+    ),
+    default=0,
+    parse=int,
+    validate=lambda v: v in range(10),
+)
--- a/Lib/fontTools/misc/configTools.py
+++ b/Lib/fontTools/misc/configTools.py
@@ -1,0 +1,274 @@
+"""
+Code of the config system; not related to fontTools or fonts in particular.
+
+The options that are specific to fontTools are in :mod:`fontTools.config`.
+
+To create your own config system, you need to create an instance of
+:class:`Options`, and a subclass of :class:`AbstractConfig` with its
+``options`` class variable set to your instance of Options.
+
+"""
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from typing import (
+    Any,
+    Callable,
+    ClassVar,
+    Dict,
+    Iterator,
+    Mapping,
+    MutableMapping,
+    Union,
+)
+
+
+log = logging.getLogger(__name__)
+
+__all__ = [
+    "AbstractConfig",
+    "ConfigAlreadyRegisteredError",
+    "ConfigError",
+    "ConfigUnknownOptionError",
+    "ConfigValueParsingError",
+    "ConfigValueValidationError",
+    "Option",
+    "Options",
+]
+
+
+class ConfigError(Exception):
+    """Base exception for the config module."""
+
+
+class ConfigAlreadyRegisteredError(ConfigError):
+    """Raised when a module tries to register a configuration option that
+    already exists.
+
+    Should not be raised too much really, only when developing new fontTools
+    modules.
+    """
+
+    def __init__(self, name):
+        super().__init__(f"Config option {name} is already registered.")
+
+
+class ConfigValueParsingError(ConfigError):
+    """Raised when a configuration value cannot be parsed."""
+
+    def __init__(self, name, value):
+        super().__init__(
+            f"Config option {name}: value cannot be parsed (given {repr(value)})"
+        )
+
+
+class ConfigValueValidationError(ConfigError):
+    """Raised when a configuration value cannot be validated."""
+
+    def __init__(self, name, value):
+        super().__init__(
+            f"Config option {name}: value is invalid (given {repr(value)})"
+        )
+
+
+_NO_VALUE = object()
+
+
+class ConfigUnknownOptionError(ConfigError):
+    """Raised when a configuration option is unknown."""
+
+    def __init__(self, name, value=_NO_VALUE):
+        super().__init__(
+            f"Config option {name} is unknown"
+            + ("" if value is _NO_VALUE else f" (with given value {repr(value)})")
+        )
+
+
+@dataclass
+class Option:
+    help: str
+    """Help text for this option."""
+    default: Any
+    """Default value for this option."""
+    parse: Callable[[str], Any]
+    """Turn input (e.g. string) into proper type. Only when reading from file."""
+    validate: Callable[[Any], bool]
+    """Return true if the given value is an acceptable value."""
+
+
+class Options(Mapping):
+    """Registry of available options for a given config system.
+
+    Define new options using the :meth:`register()` method.
+
+    Access existing options using the Mapping interface.
+    """
+
+    __options: Dict[str, Option]
+
+    def __init__(self) -> None:
+        self.__options = {}
+
+    def register(
+        self,
+        name: str,
+        help: str,
+        default: Any,
+        parse: Callable[[str], Any],
+        validate: Callable[[Any], bool],
+    ) -> Option:
+        """Register a new option."""
+        return self.register_option(name, Option(help, default, parse, validate))
+
+    def register_option(self, name: str, option: Option) -> Option:
+        """Register a new option."""
+        if name in self.__options:
+            raise ConfigAlreadyRegisteredError(name)
+        self.__options[name] = option
+        return option
+
+    def __getitem__(self, __k: str) -> Option:
+        return self.__options.__getitem__(__k)
+
+    def __iter__(self) -> Iterator[str]:
+        return self.__options.__iter__()
+
+    def __len__(self) -> int:
+        return self.__options.__len__()
+
+
+_USE_GLOBAL_DEFAULT = object()
+
+
+class AbstractConfig(MutableMapping):
+    """
+    Create a set of config values, optionally pre-filled with values from
+    the given dictionary.
+
+    .. seealso:: :meth:`set()`
+
+    This config class is abstract because it needs its ``options`` class
+    var to be set to an instance of :class:`Options` before it can be
+    instanciated and used.
+
+    .. code:: python
+
+        class MyConfig(AbstractConfig):
+            options = Options()
+
+        MyConfig.register_option( "test:option_name", "This is an option", 0, int, lambda v: isinstance(v, int))
+
+        cfg = MyConfig({"test:option_name": 10})
+    """
+
+    options: ClassVar[Options]
+
+    @classmethod
+    def register_option(cls, *args, **kwargs) -> Option:
+        """Register an available option in this config system."""
+        return cls.options.register(*args, **kwargs)
+
+    _values: Dict[str, Any]
+
+    def __init__(
+        self,
+        values: Union[AbstractConfig, Dict] = {},
+        parse_values=False,
+        skip_unknown=False,
+    ):
+        self._values = {}
+        values_dict = values._values if isinstance(values, AbstractConfig) else values
+        for name, value in values_dict.items():
+            self.set(name, value, parse_values, skip_unknown)
+
+    def set(self, name: str, value: Any, parse_values=False, skip_unknown=False):
+        """Set the value of an option.
+
+        Args:
+            * `parse_values`: parse the configuration value from a string into
+                its proper type, as per its `Option` object. The default
+                behavior is to raise `ConfigValueValidationError` when the value
+                is not of the right type. Useful when reading options from a
+                file type that doesn't support as many types as Python.
+            * `skip_unknown`: skip unknown configuration options. The default
+                behaviour is to raise `ConfigUnknownOptionError`. Useful when
+                reading options from a configuration file that has extra entries
+                (e.g. for a later version of fontTools)
+        """
+        try:
+            option = self.options[name]
+        except KeyError:
+            if skip_unknown:
+                log.debug(
+                    "Config option %s is unknown (with given value %r)", name, value
+                )
+                return
+            else:
+                raise ConfigUnknownOptionError(name, value)
+
+        # Can be useful if the values come from a source that doesn't have
+        # strict typing (.ini file? Terminal input?)
+        if parse_values:
+            try:
+                value = option.parse(value)
+            except Exception as e:
+                raise ConfigValueParsingError(name, value) from e
+
+        if not option.validate(value):
+            raise ConfigValueValidationError(name, value)
+
+        self._values[name] = value
+
+    def get(self, name: str, default=_USE_GLOBAL_DEFAULT):
+        """
+        Get the value of an option. The value which is returned is the first
+        provided among:
+
+        1. a user-provided value in the options's ``self._values`` dict
+        2. a caller-provided default value to this method call
+        3. the global default for the option provided in ``fontTools.config``
+
+        This is to provide the ability to migrate progressively from config
+        options passed as arguments to fontTools APIs to config options read
+        from the current TTFont, e.g.
+
+        .. code:: python
+
+            def fontToolsAPI(font, some_option):
+                value = font.cfg.get("someLib.module:SOME_OPTION", some_option)
+                # use value
+
+        That way, the function will work the same for users of the API that
+        still pass the option to the function call, but will favour the new
+        config mechanism if the given font specifies a value for that option.
+        """
+        if name in self._values:
+            return self._values[name]
+        if default is not _USE_GLOBAL_DEFAULT:
+            return default
+        try:
+            return self.options[name].default
+        except KeyError as e:
+            raise ConfigUnknownOptionError(name) from e
+
+    def copy(self):
+        return self.__class__(self._values)
+
+    def __getitem__(self, name: str) -> Any:
+        return self.get(name)
+
+    def __setitem__(self, name: str, value: Any) -> None:
+        return self.set(name, value)
+
+    def __delitem__(self, name: str) -> None:
+        del self._values[name]
+
+    def __iter__(self) -> Iterator:
+        return self._values.__iter__()
+
+    def __len__(self) -> int:
+        return len(self._values)
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}({repr(self._values)})"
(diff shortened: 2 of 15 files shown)


belluzj:
I tried to move the `register_option(...)` to be located in the module that is concerned with the option, but that creates an issue when parsing the configuration: if the module in question is not imported yet, its configuration options are not registered yet and so the contents of the configuration cannot be validated.

This could be solved by:
* listing all options in the config module instead of in the various modules
* deferring validation of config contents to the point where the option values are used, hoping that the code that uses the option is in the same module that registered the option, so the option will have been registered by then (I don't like this because I think it would be better for the user if they get a crash asap if the config contains errors)
* importing all modules that declare options when the config module is loaded (not great because why import code that you won't use?)
* having in each module a `config` file that registers options for the module, and import all of these when the top-level `config` module is loaded. Eg. `otlLib.config`, `varLib.config` etc

@anthrotype or @behdad  would you be able to advise please?

image

anthrotype:
The first option -- "listing all options in the config module instead of in the various modules" --  is the easiest, though probably not the most elegant.
The fourth option (otlLib.config, varLib.config etc) may also work, however in fonttools the subpackages `__init__.py` are usually not empty, so importing `varLib.config` would also import varLib's `__init__`.

What about, instead of needing to have calls to `register_option()` from some python module, you define a text file or files (e.g. configparser's .ini files) that contain the names of the modules and the options and the respective default values, and upon importing the global `fontTools.config` we parse that file or files (fetching it via importlib.resources and embedded in the distribution as package_data) and populate the Config class (calling register_option in turn for each one defined in the default config)?

belluzj:
> a text file or files (e.g. configparser's .ini files) that contain the names of the modules and the options and the respective default values

That sounds OK except it's not Python code, so one would not be able to implement parsing/validation logic (or one would have to express that declaratively, like with JSON Schema for example). For that reason I would say it would be better as Python code. Also, I was quite happy with the idea that you could import the config options into your code, to refer to them without having hardcoded strings in the code; that wouldn't be simple with the text-file-approach.

Maybe the first option can be made to look organized enough by having `config.varLib`, `config.otlLib` etc modules to store the options of each module? It's still the same as option 1 but maybe it would make immediate sense to contributors. What do you think about that?

anthrotype:
> implement parsing/validation logic

anthrotype:
> implement parsing/validation logic

can you make an example of what you have in mind? I thought the options values would be simple str, float, int, or stuff that you can store as json.

> you could import the config options into your code, to refer to them without having hardcoded strings in the code

Not sure what you mean by hard coded string. You could still import the default config object and get the already-parsed values from it.

anthrotype:
> config.varLib, config.otlLib etc modules to store the options of each module?

yeah, if you actually need to have _code_ to register options, then that could be an option to keep it organized. I'm still not 100 % that a declarative approach wouldn't be sufficient for our use case.

belluzj:
> can you make an example of what you have in mind? I thought the options values would be simple str, float, int, or stuff that you can store as json.

Yes it's indeed very simple stuff, but still, assuming people can give options by writing them in a config file, we should expect to get the wrong type or values out of range. JSON schema would handle these cases for sure. See the tests below: parsing an input string to int, checking that the value is in the expected range.

https://github.com/fonttools/fonttools/blob/bd0e1563ead0f309237cff15800e8a169caf63fb/Tests/config_test.py#L44-L58

> Not sure what you mean by hard coded string. You could still import the default config object and get the already-parsed values from it.

See the two tests below: you have the `config` object, and you can index it either by string or by the `Option` object. I feel that indexing by option object should be the preferred way, it would help prevent typos; but I admit it's not a massive win at all, just a small thing.

https://github.com/fonttools/fonttools/blob/bd0e1563ead0f309237cff15800e8a169caf63fb/Tests/config_test.py#L20-L30

behdad:
Collecting options from source into a toplevel config.py module is another option. We can make that run in CI and fail CI if config.py changed.

behdad:
I'll comment on the rest of the approach soon. Thanks!

belluzj:
Cosimo: don't perform conversions from string to int in the default constructor; have a special one for parsing from file that does the conversions. In general you wouldn't expect an object that accepts an int value to also accept a string value and convert it.

Lib/fontTools/otlLib/optimize/gpos.py
```diff
@@ -3,22 +3,42 @@
...
-
-# NOTE: activating this optimization via the environment variable is
-# experimental and may not be supported once an alternative mechanism
-# is in place. See: https://github.com/fonttools/fonttools/issues/2349
-GPOS_COMPACT_MODE_ENV_KEY = "FONTTOOLS_GPOS_COMPACT_MODE"
-GPOS_COMPACT_MODE_DEFAULT = "0"
+from fontTools.config import register_option
+
+MODE_OPTION = register_option(
+    name="otlLib.optimize.gpos.mode",
```
belluzj, line 15: Cosimo: prepend `fontTools.`, actually use `f'{__name__}.mode'` (but that only works if the option is defined in the module)

belluzj:
Instead of reading configuration from files, we could read it from a lib key in the designspace / UFO or a custom parameter in Glyphs.app source. 

E.g. in designspace:
```xml
<lib>
  <key>org.fontTools.config</key>
  <dict>
    <key>fontTools.otlLib.optimize.gpos.mode</key>
    <int>8</int>
  </dict>
</lib>
```

in ufo2ft: 

```python
ttFont = TTFont(config=designspace.lib.get('org.fontTools.config', {}))
```

belluzj:
Cosimo: should this module be in misc? It's not font-related

Also maybe there's an off-the-shelf thing in general for all this?

behdad:
> Instead of reading configuration from files, we could read it from a lib key in the designspace / UFO or a custom parameter in Glyphs.app source.

It can come from there as well, yes.



> ```
>     <key>fontTools.otlLib.optimize.gpos.mode</key>
> ```

I find `mode` very confusing. Can we call this `compressionLevel`?

behdad:
I feel like the module is fundamental enough that should be at toplevel `fonttools.config`. But I'm fine if it goes in `misc.config`, and for everyone to use. That's a pattern we've been following. And I don't see anything simple off the shelf that we can use.

As for the TTFont property, at the risk of bikeshedding, I suggest we call it `font.cfg`. :D.

Lib/fontTools/varLib/plot.py
```diff
@@ -2,7 +2,7 @@
 
 from fontTools.varLib.models import VariationModel, supportScalar
 from fontTools.designspaceLib import DesignSpaceDocument
-from matplotlib import pyplot
+# from matplotlib import pyplot
```
belluzj, line 5: TODO: undo this change

Lib/fontTools/otlLib/optimize/gpos.py
belluzj: @anthrotype I did this but I went with the simple option 1, so I'm not using `__name__`

belluzj:
* about where to define options: I implemented the simplest solution (all options to be listed in the config module's `__init__.py`)
* the module stays top-level but it's split between `fontTools.config.generic` (not related to fonts) and `fontTools.config` (the one to use everywhere in fontTools and client libraries). I could move the `generic` part to `misc` if that makes sense? Or not.
* I renamed `TTFont.config` to `TTFont.cfg` and `mode` to `compressionLevel`
* I'm using a dataclass in the config module and rebased on top of `drop-3.6` so that tests pass (sorry for the noise in the diffs, I'll rebase on top of main when possible)

belluzj:
@anthrotype @behdad if you agree with how it is at the moment I'll start adding other options. I would also be happy to know which options you have in mind to add; I'm think to start with the parameters to TTFont().

Lib/fontTools/config/generic.py
```diff
@@ -0,0 +1,190 @@
...
+                value = option.parse(value)
+            except Exception as e:
+                raise ConfigValueParsingError(name, value) from e
+
+        if not option.validate(value):
+            raise ConfigValueValidationError(name, value)
+
+        self._values[name] = value
+
+    def get(self, option: Union[Option, str], default=_USE_GLOBAL_DEFAULT):
```
anthrotype, line 139: what's the point of allowing either an Option or a str for the key parameter of .get() and other dict-like methods? you're only using the Option's `name` attribute, a str, to lookup the global OPTIONS dict, so I think just accepting str would be enough, no?
What's the rationale?

behdad:
Thanks Jany. This is concise and good. I have some ideas forward. Let me sketch those.

Lib/fontTools/config/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+Define all configuration options that can affect the working of fontTools
+modules. E.g. optimization levels of varLib IUP, otlLib GPOS compaction, etc.
+If this file gets too big, split it into smaller files per-module.
+
+An instance of the Config class can be attached to a TTFont object, so that
+the various modules can access their configuration options from it.
+"""
+from textwrap import dedent
+
+from fontTools.config.generic import *
```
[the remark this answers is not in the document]
belluzj: Yes that makes sense, will do! I forgot about the __init__ behaviour you mention.

Lib/fontTools/otlLib/optimize/gpos.py
```diff
@@ -3,40 +3,16 @@
...
-    ),
-    default=0,
-    parse=int,
-    validate=lambda v: v in range(10),
-)
+from fontTools.config import OPTIONS
 
 log = logging.getLogger("fontTools.otlLib.optimize.gpos")
 
+LEVEL_OPTION = OPTIONS['fontTools.otlLib.optimize.gpos.compressionLevel']
```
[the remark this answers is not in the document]
belluzj: Will do both. The colon sounds a bit strange to me at first glance but maybe it will make lots of sense once I see it in the code.

Lib/fontTools/config/generic.py
belluzj: My idea here was that it is (marginally) more convenient and prevents typos. Imagine with just strings:

```python
def function():
    font.cfg['long option name without typo']
    ...lots of code...
    if something:
        font.cfg['long option name with a typo']
    ...lots of code...
    font.cfg['long option name without typo']
```
* long strings everywhere
* no editor autocomplete when using options/no squiggly underlines from the linter
* the typo would be silent until the code reaches the line of code with the typo (hopefully there's a unit test that covers it)

with the PR as it is now:

```python
THING = OPTIONS['long option name without typo']

def function():
    font.cfg[THING]
    ...lots of code...
    if something:
        font.cfg[THING]
    ...lots of code...
    font.cfg[THING]
```

* long string only once at the top
* autocomplete + linter to pick typos in the constant name
* if there's a typo in the long string it will explode on import

I do agree that these are subjective and marginal benefits, I'm happy to simplify the code and just use strings.

anthrotype: well, one could still do `font.cfg[THING.name]`.

behdad:
Copy/pasting from chat to Jany/Cosimo:

Jany, Cosimo, I have some Evil config ideass 😀
gimme a week to cook them.
Here's the sneak peak:
based on your design so far.
In your __getitem__
```
def __getitem__(s):

  if hasattr(s, '__module__'):
    s = s.__module__ + s.__name__
  elif hasattr(s, '__name__'):
    s = s.__name__
```
then accept a tuple: (path,node):
```
  s = path+node
```
in short, you can then do:
```
  cfg[obj,"COMPRESSION_LEVEL"]
```
and avoid repeating `"fontTools.ttLib....."`

any classtype or module can be passed there.

better yet, instead of a tuple, we can accept a colon as Cosimo liked 😛
```
  cfg[obj:"COMPRESSION_LEVEL"]
```

Next. If only a path is passed and no node, we return the subconfig. That is, `cfg[cls]` will return a new cfg that has subconfig for the class. `__getattr__` will also be overriden to do the right thing.  So we can do: `cfg[cls].COMPRESSION_LEVEL`

Finally, `__call__` will be used to set config on multiple objects as a context:

```
with cfg(obj1, obj2, obj3, mod1, cls1, ...):
  ...
```
and the main way to use the config is to use to populate key/value arguments of constructors, methods, and functions. The syntax for that I have not fully figured out yet.
```
  cfg.invoke(obj.method) (params)
```
That would pass all relevant cfg to method.

Anyway; this is what's been brewing in my head.

As for your Option type, I want use the typing module types directly.

No need for extra Option type.

Behdad Esfahbod,
Lemme copy/paste this on the issue.

Jany Belluz,
Ok, thanks

That sounds a bit complicated to be honest, but maybe it looks nice when using it in the code

Behdad Esfahbod,
Yeah I feel like ergonomically it works.

Jany Belluz,
Ok cool

(sounds like how a Ruby developer would think, not a Python one)
I was going for pedestrian and non-surprising

Behdad Esfahbod,
I understand 🙂

I'm not sure mine are good. Evil it might be.

But want to explore 😀

Jany Belluz,
Yes anyway on the issue would be good 🙂 I'm curious to see how the usage looks

belluzj:
@anthrotype I rebased on latest main

Lib/fontTools/misc/configTools.py
```diff
@@ -0,0 +1,186 @@
...
+        raise ConfigAlreadyRegisteredError(name)
+    option = Option(name, help, default, parse, validate)
+    OPTIONS[name] = option
+    return option
+
+
+_USE_GLOBAL_DEFAULT = object()
+
+
+class Config(MutableMapping):
```
anthrotype, line 95: I suggest we give Config a nice `__repr__` to aid debugging, I think MutableMapping doesn't make one for you.

Lib/fontTools/misc/configTools.py
```diff
@@ -0,0 +1,186 @@
...
+        )
+
+
+@dataclass
+class Option:
+    name: str
+    help: str
+    default: Any
+    # Turn input (e.g. string) into proper type. Only when reading from file.
+    parse: Callable
```
anthrotype, line 74: Callable[[str], Any]

Lib/fontTools/misc/configTools.py
```diff
@@ -0,0 +1,186 @@
...
+    # Return true if the given value is an acceptable value
+    validate: Callable
```
anthrotype, line 76: Callable[[Any], bool]

Lib/fontTools/misc/configTools.py
```diff
@@ -0,0 +1,186 @@
...
+            self.set(name, value, parse_values)
+
+    def set(self, name: str, value: Any, parse_values=False):
+        """
+        Set the value of an option.
+        """
+        try:
+            option = OPTIONS[name]
+        except KeyError:
+            # FIXME: (Jany)  What to do with options we don't recognize?
```
anthrotype, line 115: could be an init parameter to the Config class, like `skip_unknown=False`

Lib/fontTools/misc/configTools.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def __init__(self, name, value):
+        super().__init__(
+            f"Config option {name} is unknown (with given value {repr(value)})"
+        )
+
+
+@dataclass
+class Option:
+    name: str
```
belluzj, line 70: Cosimo: delete this attribute, not needed because it's already in the dictionary key of OPTIONS

Lib/fontTools/misc/configTools.py
belluzj: Print `f"Config({repr(self._values)})"`

Lib/fontTools/misc/configTools.py
```diff
@@ -0,0 +1,186 @@
...
+    help: str
+    default: Any
+    # Turn input (e.g. string) into proper type. Only when reading from file.
+    parse: Callable
+    # Return true if the given value is an acceptable value
+    validate: Callable
+
+
+OPTIONS: Dict[str, Option] = {}
```
belluzj, line 79: Make object with Mapping interface, have register_option as method, move global variable to fontTools.config.__init__ (and have the inner dict of options double underscore prefixed)

Lib/fontTools/misc/configTools.py
belluzj: Cosimo says: Or maybe a version the config file? Needs more thinking (maybe see what other applications are doing with their config). Maybe set() should always raise if you try to set an unknown option, but when you're parsing a file, maybe there it would make sense to skip unknown options?

Lib/fontTools/misc/configTools.py
```diff
@@ -0,0 +1,186 @@
...
+
+_USE_GLOBAL_DEFAULT = object()
+
+
+class Config(MutableMapping):
+    _values: Dict
+
+    @classmethod
+    def from_file(cls):
+        """TODO"""
```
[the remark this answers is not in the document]
belluzj: I removed the method for now

Lib/fontTools/varLib/merger.py
```diff
@@ -850,10 +846,10 @@ def merge(merger, self, lst):
 		# Compact the merged subtables
 		# This is a good moment to do it because the compaction should create
 		# smaller subtables, which may prevent overflows from happening.
-		mode = os.environ.get(GPOS_COMPACT_MODE_ENV_KEY, GPOS_COMPACT_MODE_DEFAULT)
-		if mode and mode != "0":
+		level = merger.font.cfg[COMPRESSION_LEVEL.name]
```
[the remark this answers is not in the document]
belluzj: I removed that and used the string directly everywhere

Lib/fontTools/misc/configTools.py
belluzj: I added skip_unknown=False

belluzj:
@anthrotype I fixed all your comments and added the new module to the Sphinx documentation.

I had an idea: how neat would it be if the list of registered options was parsed by Sphinx and added to the documentation page as well? I had a quick look and it sounds possible by adding some Sphinx code with new nodes and stuff, but I didn't go down the rabbit hole. Just a thought for later :)

anthrotype:
> would it be if the list of registered options was parsed by Sphinx 

that would be nice, yeah. File an issue so we don't forget

belluzj:
Done for the classvar and the issue

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -101,6 +102,7 @@ def __init__(self, file=None, res_name_or_index=None,
 		self.recalcTimestamp = recalcTimestamp
 		self.tables = {}
 		self.reader = None
+		self.cfg = Config(cfg)
```
anthrotype, line 105: how about here you check `if not isinstance(cfg, AbstractConfig): cfg = Config(cfg)` otherwise you self `self.cfg = cfg`. This would allow one to stick a custom super-set of fonttools own Config with additional flags that can be useful in ones workflow

belluzj: Ok will do now

anthrotype: We would still want each TTFont instance to have its distinct copy, since a config is mutable.
You current cast `Config(cfg)` achieves that already. To accomplish the same with my proposal above we need a `AbstractConfig.copy()` method that returns a new instance of the type with its own copy of the internal `self._values` dictionary. A shallow copy is fine, if any mutable values are used in individual configs, it's not our job to deepcopy those I think.

So in addition to [my earlier suggestion](https://github.com/fonttools/fonttools/pull/2416/files#r850336161), we add this to `AbstractConfig`:

```python
def copy(self):
    return self.__class__(self._values)
```

and then call it inside `TTFont.__init__`

```python
    self.cfg = Config(cfg) if not isinstance(cfg, AbstractConfig) else cfg.copy()
```

anthrotype: maybe also add a test that sets identical Config oject in two different fonts, and assert they have different `id()`;  or modify one font.cfg, and assert that it did not also change in the other font

belluzj: Done, I added a test for the use-case that you were describing, if you can check that it's what you meant?

anthrotype: > if you can check that it's what you meant?

yes, that's it

belluzj: Done for real now including the copy()

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -89,7 +91,7 @@ def __init__(self, file=None, res_name_or_index=None,
 			sfntVersion="\000\001\000\000", flavor=None, checkChecksums=0,
 			verbose=None, recalcBBoxes=True, allowVID=NotImplemented, ignoreDecompileErrors=False,
 			recalcTimestamp=True, fontNumber=-1, lazy=None, quiet=None,
-			_tableCache=None):
+			_tableCache=None, cfg={}):
```
anthrotype, line 94: in general mutable default arguments are bad, but in this case you always make a copy so it should be fine. Alternatively you default to `cfg=None` and then in method body do:

```
if cfg is None:
    cfg = Config()
self.cfg = cfg.copy() if isinstance(cfg, AbstractConfig) else Config(cfg)
```

anthrotype: actually it's fine like this

anthrotype (approved):

anthrotype (approved):

belluzj:
@anthrotype I wrote release notes in the description

PR #2422, opened 2021-10-07
Use a set for file names for clash checking

### LINKED ISSUE #2421: [ufoLib] userNameToFileName very slow for large `existing` lists
Profiling https://github.com/madig/noto-amalgamated/blob/main/amalgamate-noto.py shows that ~2 mins of ~10 mins are spent in userNameToFileName, specifically 

https://github.com/fonttools/fonttools/blob/df2916af615a3d5676d457d547496c3f43da3479/Lib/fontTools/ufoLib/filenames.py#L101

since `existing` is derived from `self._existingFileNames.values()` in https://github.com/fonttools/fonttools/blob/df2916af615a3d5676d457d547496c3f43da3479/Lib/fontTools/ufoLib/glifLib.py#L461.

`writeGlyph` is called per-glyph, I don't see an easy way to fix this. Changing the call to  `set(self._existingFileNames.values())` only makes it slower because now you make a set every time. And I'm not sure the list is guaranteed to be ordered?

madig:
Closes #2421.

Thanks @justvanrossum, right on the money :) In the mentioned script, runtime spent in `glyphNameToFileName` dropped from 20% to 0.15% :)

I smuggled a type annotation in and directed Python to postpone evaluation of annotations, to avoid the parsing overhead and also because the `set[str] | None` syntax is Python 3.9+.

### DIFF
--- a/Lib/fontTools/ufoLib/filenames.py
+++ b/Lib/fontTools/ufoLib/filenames.py
@@ -15,7 +15,7 @@ class NameTranslationError(Exception):
 	pass
 
 
-def userNameToFileName(userName: str, existing=[], prefix="", suffix=""):
+def userNameToFileName(userName: str, existing=set(), prefix="", suffix=""):
 	"""
 	existing should be a case-insensitive list
 	of all existing file names.
--- a/Lib/fontTools/ufoLib/glifLib.py
+++ b/Lib/fontTools/ufoLib/glifLib.py
@@ -10,6 +10,8 @@ in a folder. It offers two ways to read glyph data, and one way to write
 glyph data. See the class doc string for details.
 """
 
+from __future__ import annotations
+
 import logging
 import enum
 from warnings import warn
@@ -205,7 +207,7 @@ class GlyphSet(_UFOBaseIO):
 		self.glyphNameToFileName = glyphNameToFileNameFunc
 		self._validateRead = validateRead
 		self._validateWrite = validateWrite
-		self._existingFileNames = None
+		self._existingFileNames: set[str] | None = None
 		self._reverseContents = None
 
 		self.rebuildContents()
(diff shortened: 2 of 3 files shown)


Lib/fontTools/ufoLib/filenames.py
```diff
@@ -15,7 +15,7 @@ class NameTranslationError(Exception):
 	pass
 
 
-def userNameToFileName(userName: str, existing=[], prefix="", suffix=""):
+def userNameToFileName(userName: str, existing=set(), prefix="", suffix=""):
```
anthrotype, line 18: avoid the mutable default value here by setting `existing=None` and creting the empty set within the function body `if existing is None`

justvanrossum: Or default to an empty tuple. But it's a redundant change as `existing` will never be modified.

Lib/fontTools/ufoLib/glifLib.py
```diff
@@ -455,12 +457,10 @@ def writeGlyph(self, glyphName, glyphObject=None, drawPointsFunc=None, formatVer
 		fileName = self.contents.get(glyphName)
 		if fileName is None:
 			if self._existingFileNames is None:
-				self._existingFileNames = {}
-				for fileName in self.contents.values():
-					self._existingFileNames[fileName] = fileName.lower()
-			fileName = self.glyphNameToFileName(glyphName, self._existingFileNames.values())
+				self._existingFileNames = set(fileName.lower() for fileName in self.contents.values())
```
anthrotype, line 460: maybe use a set comprehension instead? `{filename.lower() for ...}`

anthrotype (approved):
Thanks Just and Nikolaus!

Lib/fontTools/ufoLib/glifLib.py
```diff
@@ -205,7 +207,7 @@ def __init__(
 		self.glyphNameToFileName = glyphNameToFileNameFunc
 		self._validateRead = validateRead
 		self._validateWrite = validateWrite
-		self._existingFileNames = None
+		self._existingFileNames: set[str] | None = None
```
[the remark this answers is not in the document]
madig: I don't think this is an issue as long as we postpone evaluation; mypy doesn't complain. I asked on the Python discord in the typing channel and someone told me it's fine :D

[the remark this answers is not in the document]
anthrotype: that means one should run static typechecker with 3.9+ -- I guess I'm fine with that

Lib/fontTools/ufoLib/filenames.py
anthrotype: > Hm, this function might be called from outside, so I'll have to guard against None inside.

it was not checking for None-ness before when it was defaulting to an empty (yet mutable) list. I don't think you need to guard for None.

Lib/fontTools/ufoLib/glifLib.py
anthrotype: granted, however making it an immutable set both in the default value (empty) and at the call site kills two or three birds: removes the need to use a mutable default or the check for None-ness

anthrotype: ok, don't want to be too strict with immutability etc. Just treat (or annotate even) the `existing` parameter as a a `Container` (we just use its `__contains__` method) so it can accept any collections that is a container; then as default value use an immutable type so you don't have to check for none when called without explicit value; and at the call site, use a set comprehension.

anthrotype (commented):
looks good, thanks!

Issue #2434, opened 2021-10-25
[designspace] Proposal to add `<fea>`


ctrlcctrlv:
コピペ of https://github.com/unified-font-object/ufo-spec/issues/194: (via https://twitter.com/fr_brennan/status/1452644352417800197)

I was talking to a bunch of designers and very surprised they don't even know what the binary format allows because Designspace spec has so thoroughly boxed them in.

Designspace allows only a fraction of what is possible in binary format, that is to say, while binary format allows either GPOS or GSUB to contain basically any number of FeatureVariations lookups, Designspace constrains itself to only GSUB LookupType 1, Single Substitution. Despite there being obvious use for all the other types, especially GPOS.

~~I propose to deprecate `<rule>` tag, replace with `<fea>` tag. `<fea>` tag will hold a subset of Adobe FEA code. It will not allow you to insert anything at the toplevel except `lookup`. Alternatively, if people prefer, `<fea>` tag can be provided in multiple and there will be an implied wrapping `lookup … { … }`.~~

_The new proposal is to have `<fea>` be a child of `<rule>`._

Problem solved, no need for @simoncozens idea of complex "variable FEA syntax" (_cf._ https://github.com/adobe-type-tools/afdko/pull/1350, https://github.com/fonttools/fonttools/pull/2432), no need to change FEA standard at all.

Besides, changing FEA standard is confusing because what, you're adding FeatureVariations to instance UFO FEA's? What if they differ in any way?

OK, I guess Simon would say "well I want a new font format which will allow a toplevel FEA", which, fine. Or someone else might say "I want a multiple master UFO with a toplevel FEA". Again, fine. (I disagree to both, I like UFO, but no need to slow down progress over something like that.)

~~Alternatively, `<rule>` can be kept, and `<sub>` deprecated instead, with `<fea>` living inside `<rule>` instead of becoming the new toplevel. I don't like this since it makes less sense, I would rather `<fea>` hold the `<conditionset>` and have a `filename` attribute, but whatever, it's OK.~~

# Detailed proposal

Add new section § **5.1.4 fea element** as follows:

-  Defines a group of OpenType features in **a subset of** Adobe [OpenType™ Feature File (FEA) format](http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html).
[...]

PR #2436, opened 2021-10-29
Designspace version 5 + method to convert from v5 to v4 + method to build STAT table and instance names

belluzj:
This is a continuation of the extension to the designspace format [initially proposed here](https://github.com/daltonmaag/file-format-discussion/tree/main/proposals/designspace_5) and discussed at the last UFO meeting in July 2020.

- [designspaceLib] Increment format version to 5.0
  - Add discrete axes, variable fonts, STAT information, either design- or user-space location on instances
    - Add `fontTools.designspaceLib.split` module to split a designspace into sub-spaces that interpolate and that represent the variable fonts listed in the document
  - Make instance names optional and allow computing them from STAT data instead
    - Add `fontTools.designspaceLib.statNames` module
  - Allow instances to have the same location as a previously defined STAT label
  - Deprecate some attributes
    - `SourceDescriptor`: `copyLib`, `copyInfo`, `copyGroups`, `copyFeatures`
    - `InstanceDescriptor`: `kerning`, `info`; `glyphs`: use rules or sparse sources
    - both: `location`: use the more explicit `designLocation` 
    - Note: all are soft deprecations and existing code should keep working
  - Update documentation for Python methods and the XML format
- [varLib] Add `build_many` to build several variable fonts from a single designspace document
  - Add `fontTools.varLib.stat` module to build STAT tables from a designspace document

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -18,6 +18,7 @@ Then you can make a variable-font this way:
 
 API *will* change in near future.
 """
+from typing import List
 from fontTools.misc.vector import Vector
 from fontTools.misc.roundTools import noRound, otRound
 from fontTools.misc.textTools import Tag, tostr
@@ -33,7 +34,9 @@ from fontTools.varLib.merger import VariationMerger
 from fontTools.varLib.mvar import MVAR_ENTRIES
 from fontTools.varLib.iup import iup_delta_optimize
 from fontTools.varLib.featureVars import addFeatureVariations
-from fontTools.designspaceLib import DesignSpaceDocument
+from fontTools.designspaceLib import DesignSpaceDocument, InstanceDescriptor
+from fontTools.designspaceLib.split import splitInterpolable, splitVariableFonts
+from fontTools.varLib.stat import buildVFStatTable
 from functools import partial
 from collections import OrderedDict, namedtuple
 import os.path
@@ -53,7 +56,7 @@ FEAVAR_FEATURETAG_LIB_KEY = "com.github.fonttools.varLib.featureVarsFeatureTag"
 # Creation routines
 #
 
-def _add_fvar(font, axes, instances):
+def _add_fvar(font, axes, instances: List[InstanceDescriptor]):
 	"""
 	Add 'fvar' table to font.
 
@@ -81,7 +84,8 @@ def _add_fvar(font, axes, instances):
 		fvar.axes.append(axis)
 
 	for instance in instances:
-		coordinates = instance.location
+		# Filter out discrete axis locations
+		coordinates = {name: value for name, value in instance.location.items() if name in axes}
 
 		if "en" not in instance.localisedStyleName:
 			if not instance.styleName:
@@ -198,11 +202,10 @@ def _add_avar(font, axes):
 
 	return avar
 
-def _add_stat(font, axes):
-	# for now we just get the axis tags and nameIDs from the fvar,
-	# so we can reuse the same nameIDs which were defined in there.
-	# TODO make use of 'axes' once it adds style attributes info:
-	# https://github.com/LettError/designSpaceDocument/issues/8
+def _add_stat(font):
+	# Note: this function only gets called by old code that calls `build()`
+	# directly. Newer code that wants to benefit from STAT data from the
+	# designspace should call `build_many()`
 
 	if "STAT" in font:
 		return
@@ -759,7 +762,8 @@ def load_designspace(designspace):
 	# Check all master and instance locations are valid and fill in defaults
 	for obj in masters+instances:
 		obj_name = obj.name or obj.styleName or ''
-		loc = obj.location
+		loc = obj.getFullDesignLocation(ds)
+		obj.designLocation = loc
 		if loc is None:
 			raise VarLibValidationError(
 				f"Source or instance '{obj_name}' has no location."
@@ -770,22 +774,18 @@ def load_designspace(designspace):
 					f"Location axis '{axis_name}' unknown for '{obj_name}'."
 				)
 		for axis_name,axis in axes.items():
-			if axis_name not in loc:
-				# NOTE: `axis.default` is always user-space, but `obj.location` always design-space.
-				loc[axis_name] = axis.map_forward(axis.default)
-			else:
-				v = axis.map_backward(loc[axis_name])
-				if not (axis.minimum <= v <= axis.maximum):
-					raise VarLibValidationError(
-						f"Source or instance '{obj_name}' has out-of-range location "
-						f"for axis '{axis_name}': is mapped to {v} but must be in "
-						f"mapped range [{axis.minimum}..{axis.maximum}] (NOTE: all "
-						"values are in user-space)."
-					)
+			v = axis.map_backward(loc[axis_name])
+			if not (axis.minimum <= v <= axis.maximum):
+				raise VarLibValidationError(
+					f"Source or instance '{obj_name}' has out-of-range location "
+					f"for axis '{axis_name}': is mapped to {v} but must be in "
+					f"mapped range [{axis.minimum}..{axis.maximum}] (NOTE: all "
+					"values are in user-space)."
+				)
 
 	# Normalize master locations
 
-	internal_master_locs = [o.location for o in masters]
+	internal_master_locs = [o.getFullDesignLocation(ds) for o in masters]
 	log.info("Internal master locations:\n%s", pformat(internal_master_locs))
 
 	# TODO This mapping should ideally be moved closer to logic in _add_fvar/avar
@@ -865,6 +865,38 @@ def set_default_weight_width_slant(font, location):
 			font["post"].italicAngle = italicAngle
 
 
+def build_many(designspace: DesignSpaceDocument, master_finder=lambda s:s, exclude=[], optimize=True, skip_vf=lambda vf_name: False):
+	"""
+	Build variable fonts from a designspace file, version 5 which can define
+	several VFs, or version 4 which has implicitly one VF covering the whole doc.
+
+	If master_finder is set, it should be a callable that takes master
+	filename as found in designspace file and map it to master font
+	binary as to be opened (eg. .ttf or .otf).
+
+	skip_vf can be used to skip building some of the variable fonts defined in
+	the input designspace. It's a predicate that takes as argument the name
+	of the variable font and returns `bool`.
+
+	Always returns a Dict[str, TTFont] keyed by VariableFontDescriptor.name
+	"""
+	res = {}
+	for _location, subDoc in splitInterpolable(designspace):
+		for name, vfDoc in splitVariableFonts(subDoc):
+			if skip_vf(name):
+				log.debug(f"Skipping variable TTF font: {name}")
+				continue
+			vf = build(
+				vfDoc,
+				master_finder,
+				exclude=list(exclude) + ["STAT"],
+				optimize=optimize
+			)[0]
+			if "STAT" not in exclude:
+				buildVFStatTable(vf, designspace, name)
+			res[name] = vf
+	return res
+
 def build(designspace, master_finder=lambda s:s, exclude=[], optimize=True):
 	"""
 	Build variation font from a designspace file.
@@ -898,7 +930,7 @@ def build(designspace, master_finder=lambda s:s, exclude=[], optimize=True):
 	# TODO append masters as named-instances as well; needs .designspace change.
 	fvar = _add_fvar(vf, ds.axes, ds.instances)
 	if 'STAT' not in exclude:
-		_add_stat(vf, ds.axes)
+		_add_stat(vf)
 	if 'avar' not in exclude:
 		_add_avar(vf, ds.axes)
 
--- a/Lib/fontTools/varLib/stat.py
+++ b/Lib/fontTools/varLib/stat.py
@@ -1,0 +1,142 @@
+"""Extra methods for DesignSpaceDocument to generate its STAT table data."""
+
+from __future__ import annotations
+
+from typing import Dict, List, Union
+
+import fontTools.otlLib.builder
+from fontTools.designspaceLib import (
+    AxisLabelDescriptor,
+    DesignSpaceDocument,
+    DesignSpaceDocumentError,
+    LocationLabelDescriptor,
+)
+from fontTools.designspaceLib.types import Region, getVFUserRegion, locationInRegion
+from fontTools.ttLib import TTFont
+
+
+def buildVFStatTable(ttFont: TTFont, doc: DesignSpaceDocument, vfName: str) -> None:
+    """Build the STAT table for the variable font identified by its name in
+    the given document.
+
+    Knowing which variable we're building STAT data for is needed to subset
+    the STAT locations to only include what the variable font actually ships.
+
+    .. versionadded:: 5.0
+
+    .. seealso::
+        - :func:`getStatAxes()`
+        - :func:`getStatLocations()`
+        - :func:`fontTools.otlLib.builder.buildStatTable()`
+    """
+    for vf in doc.getVariableFonts():
+        if vf.name == vfName:
+            break
+    else:
+        raise DesignSpaceDocumentError(
+            f"Cannot find the variable font by name {vfName}"
+        )
+
+    region = getVFUserRegion(doc, vf)
+
+    return fontTools.otlLib.builder.buildStatTable(
+        ttFont,
+        getStatAxes(doc, region),
+        getStatLocations(doc, region),
+        doc.elidedFallbackName if doc.elidedFallbackName is not None else 2,
+    )
+
+
+def getStatAxes(doc: DesignSpaceDocument, userRegion: Region) -> List[Dict]:
+    """Return a list of axis dicts suitable for use as the ``axes``
+    argument to :func:`fontTools.otlLib.builder.buildStatTable()`.
+
+    .. versionadded:: 5.0
+    """
+    # First, get the axis labels with explicit ordering
+    # then append the others in the order they appear.
+    maxOrdering = max(
+        (axis.axisOrdering for axis in doc.axes if axis.axisOrdering is not None),
+        default=-1,
+    )
+    axisOrderings = []
+    for axis in doc.axes:
+        if axis.axisOrdering is not None:
+            axisOrderings.append(axis.axisOrdering)
+        else:
+            maxOrdering += 1
+            axisOrderings.append(maxOrdering)
+    return [
+        dict(
+            tag=axis.tag,
+            name={"en": axis.name, **axis.labelNames},
+            ordering=ordering,
+            values=[
+                _axisLabelToStatLocation(label)
+                for label in axis.axisLabels
+                if locationInRegion({axis.name: label.userValue}, userRegion)
+            ],
+        )
+        for axis, ordering in zip(doc.axes, axisOrderings)
+    ]
+
+
+def getStatLocations(doc: DesignSpaceDocument, userRegion: Region) -> List[Dict]:
+    """Return a list of location dicts suitable for use as the ``locations``
+    argument to :func:`fontTools.otlLib.builder.buildStatTable()`.
+
+    .. versionadded:: 5.0
+    """
+    axesByName = {axis.name: axis for axis in doc.axes}
+    return [
+        dict(
+            name={"en": label.name, **label.labelNames},
+            # Location in the designspace is keyed by axis name
+            # Location in buildStatTable by axis tag
+            location={
+                axesByName[name].tag: value
+                for name, value in label.getFullUserLocation(doc).items()
+            },
+            flags=_labelToFlags(label),
+        )
+        for label in doc.locationLabels
+        if locationInRegion(label.getFullUserLocation(doc), userRegion)
+    ]
+
+
+def _labelToFlags(label: Union[AxisLabelDescriptor, LocationLabelDescriptor]) -> int:
+    flags = 0
+    if label.olderSibling:
+        flags |= 1
+    if label.elidable:
+        flags |= 2
+    return flags
+
+
+def _axisLabelToStatLocation(
+    label: AxisLabelDescriptor,
+) -> Dict:
+    label_format = label.getFormat()
+    name = {"en": label.name, **label.labelNames}
+    flags = _labelToFlags(label)
+    if label_format == 1:
+        return dict(name=name, value=label.userValue, flags=flags)
+    if label_format == 3:
+        return dict(
+            name=name,
+            value=label.userValue,
+            linkedValue=label.linkedUserValue,
+            flags=flags,
+        )
+    if label_format == 2:
+        res = dict(
+            name=name,
+            nominalValue=label.userValue,
+            flags=flags,
+        )
+        if label.userMinimum is not None:
+            res["rangeMinValue"] = label.userMinimum
+        if label.userMaximum is not None:
+            res["rangeMaxValue"] = label.userMaximum
+        return res
+    raise NotImplementedError("Unknown STAT label format")
--- a/Tests/varLib/stat_test.py
+++ b/Tests/varLib/stat_test.py
@@ -1,0 +1,167 @@
+from pathlib import Path
+
+import pytest
+from fontTools.designspaceLib import DesignSpaceDocument
+from fontTools.designspaceLib.split import Range
+from fontTools.varLib.stat import getStatAxes, getStatLocations
+
+
+@pytest.fixture
+def datadir():
+    return Path(__file__).parent / "../designspaceLib/data"
+
+
+def test_getStatAxes(datadir):
+    doc = DesignSpaceDocument.fromfile(datadir / "test_v5.designspace")
+
+    assert getStatAxes(
+        doc, {"Italic": 0, "width": Range(50, 150), "weight": Range(200, 900)}
+    ) == [
+        {
+            "values": [
+                {
+                    "flags": 0,
+                    "name": {
+                        "de": "Extraleicht",
+                        "en": "Extra Light",
+                        "fr": "Extra léger",
+                    },
+                    "nominalValue": 200.0,
+                    "rangeMaxValue": 250.0,
+                    "rangeMinValue": 200.0,
+                },
+                {
+                    "flags": 0,
+                    "name": {"en": "Light"},
+                    "nominalValue": 300.0,
+                    "rangeMaxValue": 350.0,
+                    "rangeMinValue": 250.0,
+                },
+                {
+                    "flags": 2,
+                    "name": {"en": "Regular"},
+                    "nominalValue": 400.0,
+                    "rangeMaxValue": 450.0,
+                    "rangeMinValue": 350.0,
+                },
+                {
+                    "flags": 0,
+                    "name": {"en": "Semi Bold"},
+                    "nominalValue": 600.0,
+                    "rangeMaxValue": 650.0,
+                    "rangeMinValue": 450.0,
+                },
+                {
+                    "flags": 0,
+                    "name": {"en": "Bold"},
+                    "nominalValue": 700.0,
+                    "rangeMaxValue": 850.0,
+                    "rangeMinValue": 650.0,
+                },
+                {
+                    "flags": 0,
+                    "name": {"en": "Black"},
+                    "nominalValue": 900.0,
+                    "rangeMaxValue": 900.0,
+                    "rangeMinValue": 850.0,
+                },
+            ],
+            "name": {"en": "Wéíght", "fa-IR": "قطر"},
+            "ordering": 2,
+            "tag": "wght",
+        },
+        {
+            "values": [
+                {"flags": 0, "name": {"en": "Condensed"}, "value": 50.0},
+                {"flags": 3, "name": {"en": "Normal"}, "value": 100.0},
+                {"flags": 0, "name": {"en": "Wide"}, "value": 125.0},
+                {
+                    "flags": 0,
+                    "name": {"en": "Extra Wide"},
+                    "nominalValue": 150.0,
+                    "rangeMinValue": 150.0,
+                },
+            ],
+            "name": {"en": "width", "fr": "Chasse"},
+            "ordering": 1,
+            "tag": "wdth",
+        },
+        {
+            "values": [
+                {"flags": 2, "linkedValue": 1.0, "name": {"en": "Roman"}, "value": 0.0},
+            ],
+            "name": {"en": "Italic"},
+            "ordering": 3,
+            "tag": "ital",
+        },
+    ]
+
+    assert getStatAxes(doc, {"Italic": 1, "width": 100, "weight": Range(400, 700)}) == [
+        {
+            "values": [
+                {
+                    "flags": 2,
+                    "name": {"en": "Regular"},
+                    "nominalValue": 400.0,
+                    "rangeMaxValue": 450.0,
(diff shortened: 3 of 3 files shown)


justvanrossum:
Hi Jany,
Thank you so much for this work! I've done a rough scan of the changes and in general I tend to agree with your changes.

With regard to your question:

> Related to the above: is it beneficial to have a default value on discrete axes?

I think so, yes: for example in an Upright/Italic setup, I consider "Upright" the default, and "Italic" the optional "special" one.

justvanrossum:
Btw. I'm not fully understanding the problem with using `fontTools.otlLib.builder.buildStatTable` instead of statmake. Yes, the statmake dependency should be removed, but why would a lib key be necessary? Coult varLib.build be changed to satisfy the needs? Does `fontTools.otlLib.builder.buildStatTable` need improvements?

belluzj:
We (Nikolaus and I) used statmake to get to a working POC fast because we're used to it and wanted to use it as a target format for the conversion script. But I don't think there's any problem with the fontTools STAT utilities, we just haven't looked into them properly yet. Statmake will certainly not be needed when building a VF directly from DSv5.

justvanrossum:
Slightly tangential: I'm often somewhat confused by the fact that instances are defined by designspace coordinates, and not userspace coordinates. The latter intuitively makes more sense to me. I wonder if other people feel the same way, and if yes, should we consider doing something about it? For instance by adding a "uservalue" attribute to the `<dimension>` element, to be mutually exclusive with "xvalue", or by adding some attribute to the `<location>` to specify in which space the dimensions are given.

belluzj:
I agree on instances, it's listed in the open questions :) happy to implement your suggestion next week

justvanrossum:
Urg! See how thoroughly I reviewed your work :) I mean, sorry :/ Yes, let's try to address this in DS5. I'm by no means 100% convinced of either of my suggestions, so if you have different ideas: let's hear it.

behdad:
LGTM in general as well.

And yes, I agree instances are better specified in user-space.

In fact, that has always been my preference. C.f also https://github.com/adobe-type-tools/afdko/pull/1350

In DS5 please let's make it possible to specify all coordinates in both userspace and designspace.

LettError:
Ooh that looks promising. I know my responsibility is fading here but if it is possible I'd like to make some notes over the weekend?

benkiel:
One thing in reading this that I really like is the `Variable font`; my understanding is that this would allow one to specify in the Designspace a larger variable font, and then subsets; for instance a variable font with an optical size and weight axis: this would allow generating a variable font with both of those axes, and then a variable font with just optical size at a particular weight value. Is that a correct reading, or am I off base?

belluzj:
@LettError that would be great, thanks in advance!

@benkiel yes absolutely, check the Aktiv Grotesk example file: https://github.com/fonttools/fonttools/blob/d67238f69cb749cea47c62cc24b78e11c468d3e2/Tests/designspaceLib/data/test_v5_aktiv.designspace#L211-L243

behdad:
>      <axis-subsets> 
>        <axis-subset name="Weight"/> 
>        <axis-subset name="Width"/> 
>        <axis-subset name="Italic" uservalue="1"/> 
>      </axis-subsets>

Can axes limit the range? I also don't think the name `axis-subset` is needed. Why not just `axis` with min/default/max, possibly omitted, etc?

LettError:
* userspace values in instances: I see it can be useful. But if possible it would be good to allow designspace coords as well, but never at the same time. 
* During the design of a system, sources come first, instances right after. Axis mapping is likely to come in quite late in the process. Which means instances with userspace locations might shift.
* DS5 would still allow anisotropic locations for instances, right? `xvalue="1" yvalue="2"` I understand such locations are not possible in Variable Fonts - they are useful in design.
* Discrete axes: good idea. I don't think there is a need for a default on a discrete axis
* `Deprecated: <copyLib, -Info, -Group, -Features> elements on <source> elements` - ok. But is there a way to record that a UFO source should only contribute glyph geometry? This is necessary for partial sources. It might also be useful to be able to indicate that a UFO source if for kerning only. I have not built such a thing, but maybe someone else has.
* `Deprecated: <glyphs>, <kerning>, <info> elements on <instance> elements` - ok.
* No comment on the STAT related things - if this is what you need. :)

madig:
If you'd like to look at a few things this should make possible, see https://github.com/daltonmaag/fonttools/tree/impl-designspace-v5/Tests/designspaceLib/data. I sketched out various use-cases:

- https://github.com/daltonmaag/fonttools/blob/impl-designspace-v5/Tests/designspaceLib/data/test_v5_aktiv.designspace: Producing several variable font subsets from the same big Designspace.
- https://github.com/daltonmaag/fonttools/blob/impl-designspace-v5/Tests/designspaceLib/data/test_v5_sourceserif.designspace: Producing two variable fonts that don't interpolate between themselves (upright vs. italic)
- https://github.com/daltonmaag/fonttools/blob/impl-designspace-v5/Tests/designspaceLib/data/test_v5_discrete.designspace: Nothing interpolates and the Designspace is just used to bind together sources in a project.
- https://github.com/daltonmaag/fonttools/blob/impl-designspace-v5/Tests/designspaceLib/data/test_v5_decovar.designspace: A font that exercises STAT format 4 records to mark free-standing locations in the Designspace, while remaining legible.

madig:
@LettError I was wondering the other day,

> DS5 would still allow anisotropic locations for instances, right? xvalue="1" yvalue="2" I know such locations are not possible in Variable Fonts - they are useful in design.

Are anisotropic locations allowed anywhere else? How should they to be evaulated in rules?

LettError:
> @LettError I was wondering the other day,
> 
> > DS5 would still allow anisotropic locations for instances, right? xvalue="1" yvalue="2" I know such locations are not possible in Variable Fonts - they are useful in design.
> 
> Are anisotropic locations allowed anywhere else? How should they to be evaulated in rules?

Type designers use designspaces to construct systems as part of the design process and anisotropic have been a very useful part of that. I don't expect any kind of support in a variable font pipeline. But is should be possible to store the data.

belluzj:
@behdad
* yes axes can limit range with  `userminimum="400" usermaximum="700" userdefault="400"` (since we're introducing this element we thought useful to prefix the attribute names with `user` to prevent confusion about user vs design). See example below: https://github.com/fonttools/fonttools/blob/d67238f69cb749cea47c62cc24b78e11c468d3e2/Tests/designspaceLib/data/test_v5.designspace#L172-L178

  The same file has several other examples with annotations. It works as you're suggesting: if you omit the min/max then the whole axis is included. Also, here is what happens to instances when you limit the range of an axis, e.g. range of "axis 1" in the example below:

  image

* About naming the element just `<axis>`, I think it might be confusing w.r.t. the top-level `<axis>` element, because the attributes are not the same. (e.g. `uservalue="1"` wouldn't make sense for an `<axis>` but it does for an `<axis-subset>`). If the concern about confusion is not warranted I'm happy to shorten the name to `<axis>`.
* About allowing user space locations wherever possible: yes, I'll look at all the places were it's possible and add that to the changes. Only one of user or design location will be allowed, as Erik suggests.

@LettError
* userspace values in instances: your suggestion sounds good, we'll add the option to use `uservalue="..."` to specify locations in user coordinates. About anisotropy, we are leaving both `xvalue` and `yvalue` untouched. We'll add a check that for each `<dimension>` element, only one coordinate system is used to specify the value.
  * Do you think that user coordinates values need to handle anisotropy? I would think not, because user coordinates is concept for VFs, which don't handle anisotropy anyway? Our current understanding is that design space coordinates on instances is the only place where anisotropy makes sense within the designspace file format. Is that right?
  * As for rules, should rules only apply to the xvalue?
* > But is there a way to record that a UFO source should only contribute glyph geometry? This is necessary for partial sources.
  
  We're planning to change/"generalize" the elements `<muteInfo>`, `<muteKerning>`, and `<mutedGlyphNames>` to also be applicable to varLib and implement support for them in the VF pipeline. Currently they're marked as used only by mutatorMath. This comes from feedback by @arrowtype that [sparse sources should be more explicit](https://github.com/daltonmaag/file-format-discussion/issues/11#issuecomment-942174891).
  * If we want to go further about sparse sources, we could also add "allowlist" versions of the attributes above, for example:
    * `includedGlyphNames` to list glyph names that the source contributes geometry for; provide an empty list for muting all glyphs, omit the element to keep all glyphs.
    * `includeFeatures` to list features that the source contributes. It could be more general than just `muteKerning`, e.g. you could contribute `kern` but not `mark` and `mkmk`, which is a common issue. Provide an empty list to not contribute any GPOS. Omit the element to contribute all features.
    
    What do you think?

Thanks both for the feedback!

LettError:
>   * Do you think that user coordinates values need to handle anisotropy? I would think not, because user coordinates is concept for VFs, which don't handle anisotropy anyway? Our current understanding is that design space coordinates on instances is the only place where anisotropy makes sense within the designspace file format. Is that right?

Yes. Anisotropy is useful during design. But userspace coordinates imply axis mapping and variable fonts where such coordinates have no meaning. So no anisotropy in userspace.

>   * As for rules, should rules only apply to the xvalue?

That seems reasonable.

>   We're planning to change/"generalize" the elements `<muteInfo>`, `<muteKerning>`, and `<mutedGlyphNames>` to also be applicable to varLib and implement support for them in the VF pipeline. Currently they're marked as used only by mutatorMath. This comes from feedback by @arrowtype that [sparse sources should be more explicit](https://github.com/daltonmaag/file-format-discussion/issues/11#issuecomment-942174891).

Ok, good!

>   * If we want to go further about sparse sources, we could also add "allowlist" versions of the attributes above, for example:
>     
>     * `includedGlyphNames` to list glyph names that the source contributes geometry for; provide an empty list for muting all glyphs, omit the element to keep all glyphs.
>     * `includeFeatures` to list features that the source contributes. It could be more general than just `muteKerning`, e.g. you could contribute `kern` but not `mark` and `mkmk`, which is a common issue. Provide an empty list to not contribute any GPOS. Omit the element to contribute all features.

Specifically for sparse kerning interventions there are some scenarios to think through, but I don't know if that needs to be solved in a format description.

Thanks!

ctrlcctrlv:
Thanks for working on this, many welcome changes making Designspace that much closer to general utility and descriptive of the entire OpenType standard.

I would ask that #2434 please be considered. This is the role of `<rule>` per your (¿)UML(?) chart:

image

As demonstrated by myself, this is not enough…both GPOS and GSUB can activate via FeatureVariations. Even if you were to keep the current `<sub>` it still will not work as a `Dict` because both key and value can be an array; value can be NULL (GSUB type 2, 0 replacement glyphs special case).

Coming down the pipeline is also COLRv1 which allows for varying the transparency of a layer of a glyph according to axis values. This is likely to be expressed in FEA syntax.

Please consider `<fea>`

### 〜:mage_man:**ﾋﾟｮﾝﾋﾟｮﾝ**

belluzj:
@ctrlcctrlv thanks for the comment. I feel your proposal is orthogonal to what we want to implement here (it's not tied with STAT data, and is very specific to the `<rule>` element that so far we were not planning to touch with this MR). May I suggest that your proposal be worked on as part of another merge request rather than be bundled with this one? I have a comment about your suggestion that I'll go make on the issue.

ctrlcctrlv:
OK :)

Tests/designspaceLib/data/out/.gitignore
```diff
@@ -0,0 +1,2 @@
+# This folder is meant to receive test output (DS5 -> DS4 conversions)
+*
```
anthrotype, line 2: why not use a temporary folder instead of a gitignored one?

Tests/designspaceLib/designspace_v5_test.py
```diff
@@ -0,0 +1,431 @@
...
+
+def test_convert_v5_document_aktiv_to_v4(datadir):
+    doc = DesignSpaceDocument.fromfile(datadir / "test_v5_aktiv.designspace")
+
+    import fontTools.designspaceLib.convert5to4
+
+    variable_fonts, stylespace = fontTools.designspaceLib.convert5to4.convert5to4(doc)
+    for vf_name, vf in variable_fonts.items():
+        vf.lib["org.statmake.stylespace"] = stylespace.to_dict()
+        vf.write(datadir / "out" / (vf_name + ".designspace"))
```
anthrotype, line 407: what is these convert_v5..to_v4 tests actually checking? I don't see any asserts

Lib/fontTools/designspaceLib/convert5to4.py
```diff
@@ -0,0 +1,538 @@
+# pyright: basic
+
+# FIXME: we're building a stylespace as in statmake but it's not part of fontTools
+# TODO: either move this converter out of fontTools so it can keep creating the stylespace
+#       or make this converter use the fontTools APIs for stat/generate data for these
+# TODO: The STAT table has a fallback name id for styles where everything is elided, but it
+#       is a name id, which exists in a VF, but not in a Designspace. Make accessible somehow?
```
anthrotype, line 7: maybe this elided fallback name could be yet another attribute in the DS5 _somewhere_

Lib/fontTools/designspaceLib/convert5to4.py
```diff
@@ -0,0 +1,538 @@
...
+#       It should be set to "2" most of the time.
+# FIXME: How to deal with Designspaces where e.g. the wght axis mapping differs slightly? See
```
anthrotype, line 9: "differs slightly"... from what?

Lib/fontTools/designspaceLib/convert5to4.py
```diff
@@ -0,0 +1,538 @@
...
+                    font=source.font,
+                    name=source.name,
+                    location=_filter_location(axes_with_range, source.location),
+                    layerName=source.layerName,
+                    familyName=source.familyName,
+                    styleName=source.styleName,
+                    # copyLib=False,
+                    # copyInfo=False,
+                    # copyGroups=False,
+                    # copyFeatures=False,
```
anthrotype, line 155: why are these here commented out?

Lib/fontTools/designspaceLib/convert5to4.py
```diff
@@ -0,0 +1,538 @@
...
+                    # localisedStyleName=instance.localisedStyleName
+                    # or vf_instance_names.localisedStyleName,
+                    # localisedStyleMapFamilyName=instance.localisedStyleMapFamilyName
+                    # or vf_instance_names.localisedStyleMapFamilyName,
+                    # localisedStyleMapStyleName=instance.localisedStyleMapStyleName
+                    # or vf_instance_names.localisedStyleMapStyleName,
+                    # Deprecated
+                    # glyphs=None,
+                    # kerning=True,
+                    # info=True,
```
anthrotype, line 205: same here, what's the purpose of commented out lines? if any, write accompanying comment

anthrotype (commented):
thanks Jany and Nikolaus! I left some comments below.
I see a lot of FIXMEs and TODOs. It'd be good to know what's the plan with each of those.
Regarding adding statmake as dependency, I think we should be able to avoid that. It seems like it's currently only being used as a transport mechanism to store the STAT-related data in the DSv4.lib. Maybe the whole convert5to4 should be moved elsewhere.

belluzj:
@chrissimpkins says: we should show this to font editors and plugin developers to make sure they're aware of the changes and would be happy to implement them or know how to stick to V4.

Maybe the library should keep saving as version 4 for a while or have an option to opt into saving as version 5? Or at least, opening a V4 and saving again it should save as V4.

chrissimpkins:
> @chrissimpkins says: we should show this to font editors and plugin developers to make sure they're aware of the changes and would be happy to implement them or know how to stick to V4.

Based on recent work with a designer who uses Erik's https://github.com/LettError/designSpaceRoboFontExtension to generate and edit DS files in Robofont.  This is simply a backwards compatibility consideration.  I recognize that Erik is involved in this thread and I don't think that it is anyone's responsibility to engage developers directly.  I don't have a good understanding of the breadth of built-in / 3rd party plugin DS editor tooling out there.  It would be useful to make sure that users have a way to get back to a state where their v4 based plugins work when they don't need expanded v5 functionality if any backwards incompatible changes are introduced.

benkiel:
100% this should have backwards compatibility built in; meaning v4 can still be opened, read/used, and saved back as v4. Breaking that will cause a lot of havoc when tools that are built on v4 assumptions cease to work when the library is updated.

I don't know what Glyphs and FontLab do with designspace files, but this should be put in front of them at the minimum and likely others. The very painful lesson from UFO3 is that if it's really hard for tool makers to implement, it quickly be; and then designers won't use v5.

anthrotype:
of couse, no reason adding DS v5 support should be a breaking change

> opening a V4 and saving again it should save as V4

yeah, makes sense

benkiel:
I should have been more specific about 'breaking change': I would expect that the API won't change, or at least be made to be backwards compatible so that when this is merged it doesn't then break tools expecting just v4. Seems like everyone agrees on this, just wanted to clearly state it.

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -8,6 +12,7 @@
 import posixpath
 from fontTools.misc import etree as ET
 from fontTools.misc import plistlib
+from fontTools.misc import etree as ElementTree
```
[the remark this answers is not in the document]
belluzj: Done

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -142,21 +168,95 @@ def __init__(
...
+        .. deprecated:: 5.0
+        """
         self.muteKerning = muteKerning
+        """bool. Indicates if the kerning data from this source
+        needs to be muted (i.e. not be part of the calculations).
+
+        MutatorMath + VarLib.
+
+        .. versionchanged:: 5.0
+            Now should also be used by varLib
```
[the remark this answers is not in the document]
belluzj: As discussed, I will extract that part of the proposal into its own PR and target v5.1 or something (the `<rule> <fea>` proposal)

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -142,21 +168,95 @@ def __init__(
...
         self.mutedGlyphNames = mutedGlyphNames or []
+        """list. Glyphnames that need to be muted in the
+        instances.
 
-    path = posixpath_property("_path")
-    filename = posixpath_property("_filename")
+        MutatorMath + VarLib.
+
+        .. versionchanged:: 5.0
+            Now should also be used by varLib
```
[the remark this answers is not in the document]
belluzj: Same, will go into its own proposal

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -266,34 +386,117 @@ def __init__(
...
+            Location can be either as in version 4 a dictionary that maps axis
+            names to (potentially anisotropic) locations, or, new in version 5,
+            a string that matches the name of a
+            :class:`LocationLabel <LocationLabelDescriptor>`.  In that second
+            case the instance should have the same location as the
+            LocationLabel.
+
+            The type of this property changes but as long as the library is used
+            to open a DSv4, the property will never be a string. Only when
+            opening a DSv5 will it start being a string.
```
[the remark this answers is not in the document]
belluzj: Removed the comment and also clarified the whole location situation (at least I think it's much clearer now)

Doc: https://fonttools--2436.org.readthedocs.build/en/2436/designspaceLib/python.html#fontTools.designspaceLib.InstanceDescriptor.locationLabel

Diagram extract: 
image

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -395,6 +662,183 @@ def map_backward(self, v):
...
+        self.location: Mapping[str, float] = location
+        self.elidable: bool = elidable
+        self.olderSibling: bool = olderSibling
+        self.labelNames: Dict[str, str] = labelNames or {}
+
+    @property
+    def defaultName(self) -> str:
+        return self.labelNames.get("en") or self.name
+
+    def __eq__(self, other):
```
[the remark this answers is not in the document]
belluzj: I have implement `__eq__` and `__repr__` on `SimpleDescriptor`, and used it in tests to write `assert doc.instances == [InstanceDescriptor(...)]`. This looks ripe for being replaced with dataclasses or attrs.

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -395,6 +662,183 @@ def map_backward(self, v):
...
 
+class DiscreteAxisDescriptor(SimpleDescriptor):
+    """Container for discrete axis data.
+
+    Use this for axes that do not interpolate.
+
+    .. versionadded:: 5.0
+    """
+
+    flavor = "axis"
```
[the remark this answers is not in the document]
belluzj: I don't know, I've added it to new classes with the same value as the XML tag for the class (however I'm not using that field while reading or writing XML). I don't know if we can remove all of them, since it's not underscored, it's supposed to be part of the public API. Also, I notice that RuleDescriptor doesn't have one.

@LettError would you know about these `flavor` attributes and what their use is?

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -395,6 +662,183 @@ def map_backward(self, v):
...
+
+class VariableFontDescriptor(SimpleDescriptor):
+    """Container for variations, sub-spaces of the Designspace."""
+
+    flavor = "variable-font"
+    _attrs = ('name', 'axisSubsets', 'lib')
+
+    def __init__(self, *, name, axisSubsets, lib=None):
+        self.name: str = name
+        self.axisSubsets: List[Union[AxisDescriptor, DiscreteAxisDescriptor]] = axisSubsets
```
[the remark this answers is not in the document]
belluzj: Thanks for spotting

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -828,9 +1285,173 @@ def readAxes(self):
...
+            userMinimum=minimum,
+            userMaximum=maximum,
+            elidable=elidable,
+            olderSibling=olderSibling,
+            linkedUserValue=linkedValue,
+            labelNames=labelNames,
+        )
+
+    def readLabels(self):
+        if not self.documentObject.formatVersion.startswith("5"):
```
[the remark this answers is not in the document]
belluzj: I introduced the property `formatTuple` and use it like so:

```python
if self.documentObject.formatTuple >= (5, 0) and "values" in axisElement.attrib:
```

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -971,8 +1592,12 @@ def _readSingleInstanceElement(self, instanceElement, makeGlyphs=True, makeKerni
                     styleMapFamilyName = styleMapFamilyNameElement.text
                     instanceObject.setStyleMapFamilyName(styleMapFamilyName, lang)
         instanceLocation = self.locationFromElement(instanceElement)
-        if instanceLocation is not None:
-            instanceObject.location = instanceLocation
+        instanceLocationString = instanceElement.attrib.get('location')
+        if instanceLocation is not None and instanceLocationString is not None:
+            raise DesignSpaceDocumentError('instance element must have at most one of the location="..." attribute or the nested location element')
+        # varLib expects the location to be an empty dictionary (not None)
+        # when only the <location> element in present with nothing inside
```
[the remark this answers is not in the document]
belluzj: Removed

Tests/designspaceLib/data/out/.gitignore
belluzj: Will do

Tests/designspaceLib/designspace_v5_test.py
belluzj: I'm going to move this converter code out of fontTools for now, and bring it back later using fontTools's STAT table tools instead of statmake. I'll take into account your feedback on this code then.

Lib/fontTools/designspaceLib/types.py
```diff
@@ -0,0 +1,81 @@
...
+from dataclasses import dataclass
+from typing import Dict, List, Mapping, Optional, Union
+
+
+@dataclass
+class Range:
+    __slots__ = "start", "end"
+
+    start: float
+    end: float
```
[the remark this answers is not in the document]
belluzj: Done

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -1180,49 +1874,97 @@ def updatePaths(self):
...
 
+    def addVariableFont(self, variableFontDescriptor: VariableFontDescriptor):
+        """Add the given ``variableFontDescriptor`` to :attr:`variableFonts`.
+
+        .. versionadded: 5.0
+        """
+        self.variableFonts.append(variableFontDescriptor)
+
+    def addVariableFontDescriptor(self, **kwargs):
+        """Instanciate a new :class:`VariableFontDescriptor` using the given
```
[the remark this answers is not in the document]
belluzj: Done

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -809,14 +1261,19 @@ def readAxes(self):
         axisElements = self.root.findall(".axes/axis")
         if not axisElements:
             return
+        is_v5: bool = self.documentObject.formatVersion.startswith("5")
         for axisElement in axisElements:
-            axisObject = self.axisDescriptorClass()
+            if is_v5 and "values" in axisElement.attrib:
```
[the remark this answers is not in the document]
belluzj: I think the formatTuple thing I added solves this issue

Lib/fontTools/designspaceLib/convert5to4.py
belluzj: I've added the name as a property on DesignSpaceDocument. I have a new question now (@madig):

https://github.com/fonttools/fonttools/blob/97a1bda5c8bcb9999587c24147aeb39b38e7adb5/Lib/fontTools/designspaceLib/convert5to4.py#L75-L80

Tests/designspaceLib/designspace_v5_test.py
belluzj: The tests are back and they're real now

Lib/fontTools/designspaceLib/convert5to4.py
belluzj: I removed that comment

belluzj:
This is ready for review; thanks everyone for the feedback so far! I removed the `statmake` dependency and I'm using `fontTools`' `buildStatTable` method, and I've updated `varLib` to use the designspace's STAT data for the STAT table.
* Please find the latest documentation here: [Python API + class diagram](https://fonttools--2436.org.readthedocs.build/en/2436/designspaceLib/python.html), [XML structure](https://fonttools--2436.org.readthedocs.build/en/2436/designspaceLib/xml.html)
* [Example input files with STAT data](https://github.com/daltonmaag/fonttools/tree/impl-designspace-v5/Tests/designspaceLib/data)
* [Example output files from the convert5To4 method](https://github.com/daltonmaag/fonttools/tree/impl-designspace-v5/Tests/designspaceLib/data/convert5to4_out) (check out the generated instance names, that's from the STAT data)
* [Tests to show that by default, reading v4 and writing out again stays v4, but if you add v5 data it upgrades the XML to v5](https://github.com/daltonmaag/fonttools/blob/impl-designspace-v5/Tests/designspaceLib/designspace_v5_test.py#L714-L724)

As per [Ben's advice](https://github.com/fonttools/fonttools/pull/2436#issuecomment-969774445), what do you think about implementing support for this in Glyphs.app, FontLab, Robofont, MFEK, Runebender, etc @schriftgestalt @twardoch @typemytype @ctrlcctrlv @cmyr?

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -1412,3 +2878,51 @@ def loadSourceFonts(self, opener, **kwargs):
...
+        .. versionadded:: 5.0
+        """
+        if self.formatVersion is None:
+            return (5, 0)
+        numbers = (int(i) for i in self.formatVersion.split("."))
+        major = next(numbers)
+        minor = next(numbers, 0)
+        return (major, minor)
+
+    def buildStatTable(self, ttFont: TTFont) -> None:
```
belluzj, line 2895: @anthrotype says: remove this method or move it to `varLib` + same for the `stat` and `statNames` modules

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -321,6 +665,117 @@ def setStyleMapFamilyName(self, styleMapFamilyName, languageCode="en"):
...
+    def getFullUserLocation(self, doc: 'DesignSpaceDocument') -> SimpleLocationDict:
+        """Get the complete user location for this instance.
+
+        .. seealso:: :meth:`getFullDesignLocation`
+
+        .. versionadded:: 5.0
+        """
+        return doc.map_backward(self.getFullDesignLocation(doc))
+
+    def getStatNames(self: InstanceDescriptor, doc: DesignSpaceDocument) -> StatNames:
```
belluzj, line 768: @anthrotype I'll also remove that method if I move the `statNames` module to `varLib`. However I'm using the STAT name generation stuff in the `convert5to4` method, so I'll have to depend on varLib there in order to keep calling the STAT name stuff. 

(that's because in DSv4 you have to write out the instance names yourself, but in v5 you can write nothing on the instances and it will all get generated from the STAT data)

Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -779,6 +779,8 @@ def preWrite(self, font):
 		format = 2
 		delta = None
 		for inID, outID in gidItems:
+			if inID is None:
+				print("Hello")
```
madig, line 783: Leftover?

belluzj: Yes thank you!

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -324,6 +325,32 @@ def getFullDesignLocation(self, doc: 'DesignSpaceDocument') -> AnisotropicLocati
                 result[axis.name] = axis.map_forward(axis.default)
         return result
 
+    def __deepcopy__(self, memo):
+        """Allow deep-copying a DesignSpace document without deep-copying
+        attached UFO fonts or TTFont objects. The :attr:`font` attribute
+        is shallow-copied instead.
```
madig, line 331: No copy involved, just passing on the reference.

Lib/fontTools/varLib/stat.py
```diff
@@ -0,0 +1,142 @@
...
+            name=name,
+            nominalValue=label.userValue,
+            flags=flags,
+        )
+        if label.userMinimum is not None:
+            res["rangeMinValue"] = label.userMinimum
+        if label.userMaximum is not None:
+            res["rangeMaxValue"] = label.userMaximum
+        return res
+    raise Exception("unreachable")
```
anthrotype, line 142: AssertionError or NotImplementedError

anthrotype (approved):
LGTM (with some minor comments below)

Would you mind also writing some brief high-level release notes? Maybe update the first message (i.e. PR decription).

Thanks again Jany for working on this!

Lib/fontTools/varLib/__init__.py
```diff
@@ -693,6 +696,7 @@ def normalize(name, value):
 		"rules",
 		"rulesProcessingLast",
 		"lib",
+		"doc",
```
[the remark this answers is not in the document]
belluzj: Nowhere, I removed it, thanks!

anthrotype (approved):

PR #2437, opened 2021-10-31
update to unicode 14.0

medicalwei:
In Debian we have unicode-data updated to 14.0 and encountered some test errors.

This PR updates unicodedata to 14.0 and fixes the related tests.  Please review.


khaledhosny (approved):

anthrotype:
thanks! Hopefull we will also have the unicodedata2 backport module (on which fontTools.unicodedata depends) also updated to Unicode 14.0 soon https://github.com/fonttools/unicodedata2/issues/44

PR #2440, opened 2021-11-06
[otlLib] Remove duplicates when building coverage

### LINKED ISSUE #2433: [otlLib] `buildCoverage` does not remove duplicate glyphs
If a class in a feature file includes duplicate glyphs, `buildCoverage` is called on a list that includes duplicate glyphs. `buildCoverage` sorts the input glyphs but does not remove duplicates.
```pycon
>>> import fontTools.otlLib.builder
>>> fontTools.otlLib.builder.buildCoverage(['A', 'B', 'A'], {'A': 1, 'B': 2}).glyphs
['A', 'A', 'B']
```
[The spec](https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#coverage-table) says “A Coverage table defines a unique index value, the _Coverage Index_, for each covered glyph”, but a glyph listed multiple times won’t have a unique index.

This can cause a problem when using the OpenType Sanitizer: if a CoverageFormat1 table lists more glyphs than are declared in 'maxp', OTS reports an error.

dscorbett:
This fixes #2433.

### DIFF
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -52,7 +52,7 @@ def buildCoverage(glyphs, glyphMap):
     if not glyphs:
         return None
     self = ot.Coverage()
-    self.glyphs = sorted(glyphs, key=glyphMap.__getitem__)
+    self.glyphs = sorted(set(glyphs), key=glyphMap.__getitem__)
     return self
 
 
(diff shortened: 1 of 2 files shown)


anthrotype (approved):
OK, I convinced myself it's the only right thing to do if we want unique coverage index per glyph.

PR #2444, opened 2021-11-10
[docs] Split table docs into individual pages

simoncozens:
This is the precursor to documenting each table convertor - which I started doing in the first commit, and then realised that the whole structure of the documentation would need to change if that was going to happen. The idea is to give the user an easier journey to get to the information they need about each table. Compare [this](https://fonttools.readthedocs.io/en/latest/ttLib/tables.html) with:

Screenshot 2021-11-10 at 10 20 41

### DIFF
--- a/Doc/source/ttLib/index.rst
+++ b/Doc/source/ttLib/index.rst
@@ -3,7 +3,7 @@ ttLib
 #####
 
 .. toctree::
-   :maxdepth: 1
+   :maxdepth: 2
 
    macUtils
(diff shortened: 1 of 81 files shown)


simoncozens:
As this is just a reorg and adds no content, and as it's needed for other table-level documentation, I'm going to merge this without review. I'll squash it so that it can be reverted if someone disagrees.

anthrotype:
Thanks Simon

PR #2446, opened 2021-11-10
Forbid empty classes (take 2)

simoncozens:
This follows on from #2445, and forbids the following FEA syntax:

```fea
markClass [] <anchor 150 -10> @TOPMARKS;

feature test {
    sub a by [];
    sub a by c [];
    sub a from [];
    sub f f [] by f;
    ignore sub a []';
    ignore sub [] a';
    sub a []' by b;
    sub [] a' by b;
    rsub [] by a;

    pos [] 120;
    pos a [] 120;
    enum pos a [] 120;
    pos cursive [] <anchor NULL> <anchor NULL>;
    pos base [] <anchor NULL> mark @TOPMARKS;
    pos mark [] <anchor NULL> mark @TOPMARKS;
    pos ligature [] <anchor NULL> mark @TOPMARKS;
    ignore pos a []';
    ignore pos [] a';

} test;
```

All of these rules were allowed before, and would fail with hard errors (i.e. non user-facing stuff, IndexErrors and the like) in the builder. Glyph deletion is still allowed with the *explicit* NULL glyph.

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -1331,9 +1331,15 @@ class PairPosStatement(Statement):
         """
         if self.enumerated:
             g = [self.glyphs1.glyphSet(), self.glyphs2.glyphSet()]
+            seen_pair = False
             for glyph1, glyph2 in itertools.product(*g):
+                seen_pair = True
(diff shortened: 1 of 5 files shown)


anthrotype (approved):
LGTM

PR #2447, opened 2021-11-10
Merge CFF rebased

khaledhosny:
Rebased #1319, but had to remove `.notdef` non-duplication since it conflicts with the change made in #1729 and needs bigger change to do properly.

TODO:
- [x] Add tests
- [x] Fix any blocking XXX


khaledhosny:
I’m not getting duplicate `.notdef` glyphs, so I’m not even sure what this code was about.

Lib/fontTools/merge.py
```diff
@@ -1048,10 +1155,29 @@ def merge(self, fontfiles):
...
+			for i, (glyphName, v) in enumerate(td.CharStrings.charStrings.items()):
+				if glyphName in mega:
+					n = mega[glyphName]
+					while (glyphName + "#" + repr(n)) in mega:
+						n += 1
+					mega[glyphName] = n
+					glyphName += "#" + repr(n)
+				mega[glyphName] = 1
+				d[glyphName] = v
+			cffTable.cff.topDictIndex[0].CharStrings.charStrings = d
```
khaledhosny, line 1177: I had also to change this code while rebasing to follow the same renaming logic below, but ideally we should de-duplicate this and do the renaming only once.

Lib/fontTools/merge.py
```diff
@@ -13,11 +13,33 @@
...
 import operator
 import logging
-
+import os
+from fontTools import subset
 
 log = logging.getLogger("fontTools.merge")
 timer = Timer(logger=logging.getLogger(__name__+".timer"), level=logging.INFO)
 
+def _desubtoutinize(otf):
```
anthrotype, line 22: by importing fontTools.subset.cff module, the CFF table gets a new `desubroutinize` method that you can call here (e.g. we do that in cffsubr: https://github.com/adobe-type-tools/cffsubr/blob/017660be89a7515abaccaa369130560049b232e1/src/cffsubr/__init__.py#L259-L268)
Maybe we should move [that code](https://github.com/fonttools/fonttools/blob/b29ae8003b5ff89b5df935f70ca584ac54112fea/Lib/fontTools/subset/cff.py#L481-L510) to fontTools.cffLib and have it available all the time, not just within the subsetter.

Lib/fontTools/merge.py
```diff
@@ -200,7 +222,7 @@ def merge(self, m, tables):
 	'macStyle': first,
 	'lowestRecPPEM': max,
 	'fontDirectionHint': lambda lst: 2,
-	'indexToLocFormat': recalculate,
+	'indexToLocFormat': first,
```
anthrotype, line 225: I forgot the reason of this change, and what has to do with CFF, i thought head.indexToLocFormat is for loca, hence TT only

Lib/fontTools/merge.py
```diff
@@ -993,20 +1078,39 @@ def merge(self, fontfiles):
 			A :class:`fontTools.ttLib.TTFont` object. Call the ``save`` method on
 			this to write it out to an OTF file.
 		"""
-		mega = ttLib.TTFont()
+		# Take first input file sfntVersion
+		sfntVersion = ttLib.TTFont(fontfiles[0]).sfntVersion
```
anthrotype, line 1082: how about you directly sniff the sfntVersion by reading the first 4 bytes, instead of creating a TTFont then throwing it away

Lib/fontTools/merge.py
```diff
@@ -993,20 +1078,39 @@ def merge(self, fontfiles):
...
+		cffTables = []
+		if sfntVersion == "OTTO":
+			for i, font in enumerate(fonts):
+				_desubtoutinize(font)
+				cffTables.append(font['CFF '])
+
 		glyphOrders = [font.getGlyphOrder() for font in fonts]
-		megaGlyphOrder = self._mergeGlyphOrders(glyphOrders)
+
+		# Handle glyphOrder merging and cff glyphs renaming together
```
anthrotype, line 1098: maybe add a little comment as to why one has to "handle glyphOrder merging and cff glyphs renaming together"

Lib/fontTools/merge.py
```diff
@@ -1018,13 +1122,16 @@ def merge(self, fontfiles):
 		allTags = reduce(set.union, (list(font.keys()) for font in fonts), set())
 		allTags.remove('GlyphOrder')
 
-		# Make sure we process cmap before GSUB as we have a dependency there.
+		# Make sure we process CFF before cmap, and cmap before GSUB as we have a dependency there.
```
anthrotype, line 1125: again unclear why cmap would depend on CFF

Lib/fontTools/merge.py
anthrotype: yes, this feels like two methods in one

Lib/fontTools/merge.py
```diff
@@ -371,6 +393,69 @@ def merge(self, m, tables):
...
+	for name in newfont.strings.strings:
+		if name not in glyphOrder:
+			storedNamesStrings.append(name)
+		else:
+			glyphOrderStrings.append(name)
+	chrset = list(newfont.charset)
+	newcs = newfont.CharStrings
+	# XXX CharStrings object does not have a 'charStringsIndex' attribute
+	# when read from a TTX file
+	newcsi = newcs.charStringsIndex
```
anthrotype, line 418: there's a `CharStrings.charStringsAreIndexed` attribute that you can test, like subset does here:
https://github.com/fonttools/fonttools/blob/b29ae8003b5ff89b5df935f70ca584ac54112fea/Lib/fontTools/subset/cff.py#L123

to support the non-indexed case, I believe you could convert the CharStrings object from indexed to non-indexed (removing the intermediation of the charStringsIndex), then treat it as non-indexed (i.e. charStrings being a Dict[str, CharString] instead of Dict[name, int])

anthrotype (commented):
Thanks Khaled for picking this up. It's been a while since I have read this code, I left some comment below

behdad:
LGTM. Thanks Khaled.

Lib/fontTools/merge.py
khaledhosny: https://github.com/fonttools/fonttools/pull/1289#discussion_r200960592

Lib/fontTools/merge.py
khaledhosny: Yes, I wasn’t feeling easy about how that code was so indirectly used.

Lib/fontTools/merge.py
anthrotype: ah ok thank you. I wonder if taking the first instead of recalculating could break things when attempting to merge TTFs that have different indexToLocFormat.

Lib/fontTools/merge.py
khaledhosny: Done. I notice that empty subroutines are removed separately using `remove_unused_subroutines()` I wonder if this should be folded into `desubroutinize()` or if there is a reason to keep it separate.

Lib/fontTools/merge.py
khaledhosny: Looks like `loca` table `compile()` always re-calculates it, so I don’t think it matters what value we set here.

Lib/fontTools/merge.py
khaledhosny: Done.

Lib/fontTools/merge.py
khaledhosny: Is reading from TTX files an issue for this module? It does not seem to allow taking TTFont’s as input, and input file names are always passed to TTFont constructor so using TTX files as input will not work.

Lib/fontTools/merge.py
khaledhosny: It does not even ha ve to do this as all fonts are opened a few lines below anyway.

Lib/fontTools/merge.py
khaledhosny: No idea either, I removed the lines that adds `CFF` before all other tables and my test fonts merged just fine.

Lib/fontTools/merge.py
anthrotype: > remove_unused_subroutines() I wonder if this should be folded into desubroutinize() or if there is a reason to keep it separate.

well, that's used when desubroutinize is False, so i guess they should stay separate
https://github.com/fonttools/fonttools/blob/9abfc00b71fc2d1aebe5f9f52ce737dac0fe7c05/Lib/fontTools/subset/cff.py#L452-L453

it's also used when --no-hinting

Lib/fontTools/subset/cff.py
```diff
@@ -478,36 +398,11 @@ def _delete_empty_subrs(private_dict):
 			del private_dict.rawDict['Subrs']
 		del private_dict.Subrs
 
+
+# TODO: This is now unused by this module, should we deprecate it?
```
anthrotype, line 402: yes, we should deprecate it. Please try use this decorator from 
https://github.com/fonttools/fonttools/blob/9abfc00b71fc2d1aebe5f9f52ce737dac0fe7c05/Lib/fontTools/misc/loggingTools.py#L521

Lib/fontTools/merge.py
```diff
@@ -1067,26 +1067,23 @@ def merge(self, fontfiles):
...
 		fonts = [ttLib.TTFont(fontfile) for fontfile in fontfiles]
-
-		if sfntVersion == "OTTO":
-			for font, glyphOrder, cffTable in zip(fonts, glyphOrders, cffTables):
-				font.setGlyphOrder(glyphOrder)
+		for font, glyphOrder, cffTable in zip(fonts, glyphOrders, cffTables):
+			font.setGlyphOrder(glyphOrder)
+			if cffTable:
 				# Rename CFF CharStrings to match the new glyphOrder.
+				# Using cffTable before reloading the fonts, because reasons.
```
anthrotype, line 1084: > Using cffTable before reloading the fonts, because reasons.

a bit cryptic, perhaps intentially so. But as far as I can tell, you are actually cffTable _after_ reloading, so I'm even more confused by the comment.

anthrotype: ah I now see. You are using the cffTable from the original fonts, before they have been reloaded. Ok, then why?

Lib/fontTools/merge.py
anthrotype: @JeremieHornus do you remember why you did that?

Lib/fontTools/merge.py
anthrotype: I see what you mean. Well, ideally we should support both states of the CharStrings object. You can add a TODO and file an issue to track that if you don't do this in the current PR.

Lib/fontTools/merge.py
JeremieHornus: I honnestly don't remember… that was a while ago…

anthrotype: no worries. Maybe it had something to do with glyph name collisions, when to-be-merged fonts contains same-named glyphs.. just a guess.

anthrotype:
@khaledhosny are you planning to add some minimal tests as well?

Lib/fontTools/merge.py
JeremieHornus: yes, likely to be something like that…

Lib/fontTools/merge.py
khaledhosny: Yes. I don’t fully understand what is going on, but if we don’t do this then duplicated glyphs gets messed up (they seem to be using the outlines from the first font) in the output font. This seems to be related to the font reloading step, if we stop doing this then all this becomes moot. My testing does not show any other difference with or without reloading.

Lib/fontTools/merge.py
khaledhosny: I’m testing with fonts that has duplicate glyph names and this change is not making any difference.

khaledhosny:
> @khaledhosny are you planning to add some minimal tests as well?

I want to add tests, but merge module does not have any integration tests and I’m not sure unit tests will be very helpful here. If I can figure an easy way to add some integration tests, I’ll do that.

Lib/fontTools/subset/cff.py
khaledhosny: Done.

Lib/fontTools/merge.py
khaledhosny: Should be supported now.

Lib/fontTools/merge.py
```diff
@@ -1252,7 +1253,14 @@ def main(args=None):
 		args = sys.argv[1:]
 
 	options = Options()
-	args = options.parse_opts(args)
+	args = options.parse_opts(args, ignore_unknown=['output-file'])
```
anthrotype, line 1256: why are you parsing it outside of the Options class, does it not support plan string args? I know at some point we should get rid of that and use stock argparse, bot for now.. I guess this is ok.

anthrotype (approved):

Lib/fontTools/merge.py
khaledhosny: This is essentially copied from `subset/__init__.py`.

anthrotype: ok

anthrotype:
Thanks Jeremie and Khaled!

PR #2448, opened 2021-11-11
Removed comment about missing “start” optional argument to `calcChecksum`

verbosus:
The optional “start” parameter isn’t there, so that part of the comment is not current.


anthrotype (approved):
thanks for spotting this! You just deleted a comment that has 22 years of history :)

PR #2452, opened 2021-11-15
Add support for subsetting OT-SVG table 

### LINKED ISSUE #534: Subset SVG table with pyftsubset
It's noted in [this issue](https://github.com/behdad/fonttools/issues/376) that the subetter doesn't know how to subset the SVG table. It would be nice if it did as SVG fonts can use all the optimization they can get :)

Currently, I have to explicitly tell pyftsusbet to leave the SVG table alone, as per instructions in the mentioned issue, so that should change too:

```
pyftsubset --verbose --no-subset-tables+='SVG ' --drop-tables-='SVG '  my-test-font.otf  --text='abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890ao123%‰€$¢£¥áâàäåãæçéêèëíîìïıñóôòöõøœšßúûùüýÿžÁÂÀÄÅÃÆÇÉÊÈËÍÎÌÏÑÓÔÒÖÕØŒŠÚÛÙÜÝŸ.,:;-–—•·...“‘’‘‚“”„‹›«»/\?!¿¡()[]{}†*&©®@#§+×=_°|'
```

anthrotype:
Fixes #534

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -10,6 +10,7 @@ from fontTools.pens.basePen import NullPen
 from fontTools.misc.loggingTools import Timer
 from fontTools.subset.util import _add_method, _uniq_sort
 from fontTools.subset.cff import *
+from fontTools.subset.svg import *
 import sys
 import struct
 import array
@@ -207,12 +208,12 @@ Font table options:
       Specify (=), add to (+=) or exclude from (-=) the comma-separated
       set of tables that will be be dropped.
       By default, the following tables are dropped:
-      'BASE', 'JSTF', 'DSIG', 'EBDT', 'EBLC', 'EBSC', 'SVG ', 'PCLT', 'LTSH'
+      'BASE', 'JSTF', 'DSIG', 'EBDT', 'EBLC', 'EBSC', 'PCLT', 'LTSH'
       and Graphite tables: 'Feat', 'Glat', 'Gloc', 'Silf', 'Sill'.
       The tool will attempt to subset the remaining tables.
       Examples:
-        --drop-tables-='SVG '
-            * Drop the default set of tables but keep 'SVG '.
+        --drop-tables-='BASE'
+            * Drop the default set of tables but keep 'BASE'.
         --drop-tables+=GSUB
             * Drop the default set of tables and 'GSUB'.
         --drop-tables=DSIG
@@ -2486,7 +2487,7 @@ class Options(object):
 
 	# spaces in tag names (e.g. "SVG ", "cvt ") are stripped by the argument parser
 	_drop_tables_default = ['BASE', 'JSTF', 'DSIG', 'EBDT', 'EBLC',
-				'EBSC', 'SVG', 'PCLT', 'LTSH']
+				'EBSC', 'PCLT', 'LTSH']
 	_drop_tables_default += ['Feat', 'Glat', 'Gloc', 'Silf', 'Sill']  # Graphite
 	_no_subset_tables_default = ['avar', 'fvar',
 				     'gasp', 'head', 'hhea', 'maxp',
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -8,6 +8,7 @@ from fontTools.ttLib.tables import otTables
 from fontTools.otlLib.maxContextCalc import maxCtxFont
 from fontTools.pens.basePen import NullPen
 from fontTools.misc.loggingTools import Timer
+from fontTools.subset.util import _add_method, _uniq_sort
 from fontTools.subset.cff import *
 import sys
 import struct
@@ -362,26 +363,6 @@ log.glyphs = MethodType(_log_glyphs, log)
 # main module's logger
 timer = Timer(logger=logging.getLogger("fontTools.subset.timer"))
 
-
-def _add_method(*clazzes):
-	"""Returns a decorator function that adds a new method to one or
-	more classes."""
-	def wrapper(method):
-		done = []
-		for clazz in clazzes:
-			if clazz in done: continue # Support multiple names of a clazz
-			done.append(clazz)
-			assert clazz.__name__ != 'DefaultTable', \
-					'Oops, table class not found.'
-			assert not hasattr(clazz, method.__name__), \
-					"Oops, class '%s' has method '%s'." % (clazz.__name__,
-									       method.__name__)
-			setattr(clazz, method.__name__, method)
-		return None
-	return wrapper
-
-def _uniq_sort(l):
-	return sorted(set(l))
 
 def _dict_subset(d, glyphs):
 	return {g:d[g] for g in glyphs}
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -329,6 +329,9 @@ Other font-specific options:
       Don't change the 'OS/2 usMaxContext' field. [default]
   --font-number=<number>
       Select font number for TrueType Collection (.ttc/.otc), starting from 0.
+  --pretty-svg
+      When subsetting SVG table, use lxml pretty_print=True option to indent
+      the XML output (only recommended for debugging purposes).
 
 Application options:
   --verbose
@@ -2554,6 +2557,7 @@ class Options(object):
 		self.timing = False
 		self.xml = False
 		self.font_number = -1
+		self.pretty_svg = False
 
 		self.set(**kwargs)
 
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -2692,7 +2692,7 @@ class Subsetter(object):
 	def _closure_glyphs(self, font):
 
 		realGlyphs = set(font.getGlyphOrder())
-		glyph_order = font.getGlyphOrder()
+		self.orig_glyph_order = glyph_order = font.getGlyphOrder()
 
 		self.glyphs_requested = set()
 		self.glyphs_requested.update(self.glyph_names_requested)
@@ -2811,6 +2811,24 @@ class Subsetter(object):
 
 		self.reverseEmptiedGlyphMap = {g:order[g] for g in self.glyphs_emptied}
 
+		if not self.options.retain_gids:
+			new_glyph_order = [
+				g for g in glyph_order if g in self.glyphs_retained
+			]
+		else:
+			new_glyph_order = [
+				g for g in glyph_order
+				if font.getGlyphID(g) <= self.last_retained_order
+			]
+		# We'll call font.setGlyphOrder() at the end of _subset_glyphs when all
+		# tables have been subsetted. Below, we use the new glyph order to get
+		# a map from old to new glyph indices, which can be useful when
+		# subsetting individual tables (e.g. SVG) that refer to GIDs.
+		self.new_glyph_order = new_glyph_order
+		self.glyph_index_map = {
+			order[new_glyph_order[i]]: i
+			for i in range(len(new_glyph_order))
+		}
 
 		log.info("Retaining %d glyphs", len(self.glyphs_retained))
 
@@ -2840,13 +2858,7 @@ class Subsetter(object):
 				del font[tag]
 
 		with timer("subset GlyphOrder"):
-			glyphOrder = font.getGlyphOrder()
-			if not self.options.retain_gids:
-				glyphOrder = [g for g in glyphOrder if g in self.glyphs_retained]
-			else:
-				glyphOrder = [g for g in glyphOrder if font.getGlyphID(g) <= self.last_retained_order]
-
-			font.setGlyphOrder(glyphOrder)
+			font.setGlyphOrder(self.new_glyph_order)
 
 
 	def _prune_post_subset(self, font):
--- a/Lib/fontTools/subset/svg.py
+++ b/Lib/fontTools/subset/svg.py
@@ -107,17 +107,17 @@ def closure_element_ids(
         unvisited = referenced
 
 
-def subset_elements(el: etree.Element, ids: Set[str]) -> bool:
+def subset_elements(el: etree.Element, retained_ids: Set[str]) -> bool:
     # Keep elements if their id is in the subset, or any of their children's id is.
     # Drop elements whose id is not in the subset, and either have no children,
     # or all their children are being dropped.
     el_id = el.attrib.get("id")
-    if el_id is not None and el_id in ids:
+    if el_id is not None and el_id in retained_ids:
         # if id is in the set, don't recurse; keep whole subtree
         return True
     keep = False
     for e in el:
-        keep |= subset_elements(e, ids)
+        keep |= subset_elements(e, retained_ids)
     if keep:
         return True
     assert len(el) == 0
--- a/Lib/fontTools/subset/svg.py
+++ b/Lib/fontTools/subset/svg.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import re
+from functools import lru_cache
 from itertools import chain, count, groupby
 from typing import Dict, Iterable, Iterator, List, Optional, Set, Tuple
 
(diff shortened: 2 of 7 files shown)


Lib/fontTools/subset/__init__.py
```diff
@@ -2505,7 +2490,7 @@ class UnknownOptionError(OptionError): pass
 
 	# spaces in tag names (e.g. "SVG ", "cvt ") are stripped by the argument parser
 	_drop_tables_default = ['BASE', 'JSTF', 'DSIG', 'EBDT', 'EBLC',
-				'EBSC', 'SVG', 'PCLT', 'LTSH']
```
rsheeter, line 2508: I don't think you necessarily have to change the drop default. Maybe that should be in it's own PR since it's not required to support subsetting?

anthrotype: normally when we add support for a new table, we also remove it from the drop-tables list, which is the list of tables for which the subsetter does not know yet how to subset so it simply drops them by default.

What do you mean "it's not required to support subsetting"?

Lib/fontTools/subset/__init__.py
```diff
@@ -2829,6 +2815,24 @@ def _closure_glyphs(self, font):
...
 
+		if not self.options.retain_gids:
+			new_glyph_order = [
+				g for g in glyph_order if g in self.glyphs_retained
+			]
+		else:
+			new_glyph_order = [
+				g for g in glyph_order
+				if font.getGlyphID(g) <= self.last_retained_order
+			]
```
rsheeter, line 2826: Optional, terser and arguably clearer if done with lambda and filter. Something like:

```
retention_fn = lambda g: g in self.glyphs_retained
if self.options.retain_gids:
  retention_fn = lambda g: font.getGlyphID(g) <= self.last_retained_order
new_glyph_order = list(filter(retention_fn, glyph_order))
```

Lib/fontTools/subset/svg.py
```diff
@@ -0,0 +1,235 @@
...
+NAMESPACES = {
+    "svg": "http://www.w3.org/2000/svg",
+    "xlink": "http://www.w3.org/1999/xlink",
+}
+XLINK_HREF = f'{{{NAMESPACES["xlink"]}}}href'
+
+if etree is not None:
+    # it's more efficient to compile XPath objects upfront than calling the same
+    # Element.xpath() several times
+    find_svg_elements_with_id = etree.XPath(".//svg:*[@id]", namespaces=NAMESPACES)
```
rsheeter, line 32: Optional, another approach would be to return these from a function with a functools annotation (lru_cache or cache depending on what the minimum python fonttools allows is).

Spares the etree None check because those functions are internal and only called from inside functions that have done the throw if no etree step.

Lib/fontTools/subset/svg.py
```diff
@@ -0,0 +1,235 @@
...
+        if declaration.count(":") == 1:
+            property_name, value = declaration.split(":")
+            property_name = property_name.strip()
+            result[property_name] = value.strip()
+        elif declaration.strip():
+            raise ValueError(f"Invalid CSS declaration syntax: {declaration}")
+    return result
+
+
+def iter_referenced_ids(tree: etree.Element) -> Iterator[str]:
```
rsheeter, line 70: Shouldn't the return type be Generator?

Lib/fontTools/subset/__init__.py
anthrotype: I simply moved that preexisting chunk of code

Lib/fontTools/subset/svg.py
```diff
@@ -0,0 +1,235 @@
...
+    find_svg_elements_with_glyph_href = etree.XPath(
+        ".//svg:*[starts-with(@xlink:href, '#glyph')]", namespaces=NAMESPACES
+    )
+
+
+def group_elements_by_id(tree: etree.Element) -> Dict[str, etree.Element]:
+    return {el.attrib["id"]: el for el in find_svg_elements_with_id(tree)}
+
+
+def parse_css_declarations(style_attr: str) -> Dict[str, str]:
```
rsheeter, line 56: Just to sanity check, do you specifically want to guarrantee Dict as opposed to any MutableMapping?

Lib/fontTools/subset/svg.py
anthrotype: the [python docs](https://docs.python.org/3.6/library/typing.html#typing.Generator) recommend to use the simpler Iterator[...] annotation when one doesn't use the Generator's SendType and ReturnType

Lib/fontTools/subset/svg.py
anthrotype: return types (unlike argument types) are specific

Lib/fontTools/subset/svg.py
```diff
@@ -0,0 +1,235 @@
...
+            if el_id not in elements:
+                # ignore dangling reference; not our job to validate svg
+                continue
+            referenced.update(iter_referenced_ids(elements[el_id]))
+        referenced -= element_ids
+        element_ids.update(referenced)
+        unvisited = referenced
+
+
+def subset_elements(el: etree.Element, ids: Set[str]) -> bool:
```
rsheeter, line 110: IIUC ids might be clearer as something like retained_ids

anthrotype: Done 92470ea

Lib/fontTools/subset/svg.py
```diff
@@ -0,0 +1,235 @@
...
+    # or all their children are being dropped.
+    el_id = el.attrib.get("id")
+    if el_id is not None and el_id in ids:
+        # if id is in the set, don't recurse; keep whole subtree
+        return True
+    keep = False
+    for e in el:
+        keep |= subset_elements(e, ids)
+    if keep:
+        return True
```
rsheeter, line 122: Do you specifically want to run subset_elements for every child? - if so, plz comment why. If not perhaps possible to make terser

```
if not any(subset_elements(e, retained_ids) for e in el):
  # remove parent
return el.getparent() is not None  # it's only none if we dropped
```

anthrotype: yes, I do want to run for every child. I don't want to do any(...) as that would stop at the first match. The subset_elements has the side effect of removing elements that are left empty

Lib/fontTools/subset/svg.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+def subset_elements(el: etree.Element, ids: Set[str]) -> bool:
+    # Keep elements if their id is in the subset, or any of their children's id is.
+    # Drop elements whose id is not in the subset, and either have no children,
+    # or all their children are being dropped.
+    el_id = el.attrib.get("id")
+    if el_id is not None and el_id in ids:
+        # if id is in the set, don't recurse; keep whole subtree
+        return True
```
rsheeter, line 117: I don't think you use el_id again so assuming None isn't valid in ids I think you could say

```
if el.attrib.get("id") in ids:
  return True
```

anthrotype: ok 12ed8b7

Lib/fontTools/subset/svg.py
anthrotype: thanks but I think using functools.lru_cache is a bit overkill for this

Lib/fontTools/subset/svg.py
```diff
@@ -0,0 +1,235 @@
...
+            # If the old index is missing, the element correspond to a glyph that was
+            # excluded from the font's subset.
+            # We rename it to avoid clashes with the new GIDs or other element ids.
+            new_id = f".{el_id}"
+            n = count(1)
+            while new_id in elements:
+                new_id = f"{new_id}.{next(n)}"
+
+        id_map[el_id] = new_id
+        el.attrib["id"] = new_id
```
rsheeter, line 157: This feels like it's doing two jobs and might be simpler if it was broken into handling each explicitly:

1. Assign new gids to things we're keeping
1. Rename things we are dropping to avoid clashes
   * Can we just destroy them? Why do we have to keep the things we aren't keeping and rename them to avoid id clashes?

Lib/fontTools/subset/svg.py
```diff
@@ -0,0 +1,235 @@
...
+
+    return id_map
+
+
+def update_glyph_href_links(svg: etree.Element, id_map: Dict[str, str]) -> None:
+    # update all xlink:href="#glyph..." attributes to point to the new glyph ids
+    for el in find_svg_elements_with_glyph_href(svg):
+        old_id = el.attrib[XLINK_HREF][1:]
```
rsheeter, line 165: suggest a function to resolve href targets; old_id = _href_target. Can comment there that 1: is to drop the leading #

Lib/fontTools/subset/svg.py
```diff
@@ -0,0 +1,235 @@
...
+        if old_id in id_map:
+            new_id = id_map[old_id]
+            el.attrib[XLINK_HREF] = f"#{new_id}"
+
+
+def ranges(ints: Iterable[int]) -> Iterator[Tuple[int, int]]:
+    # Yield (min, max) ranges of consecutive integers from the input set
```
rsheeter, line 172: Yield non-overlapping ranges, smaller #s first, I think? Any particular reason to not actually use Iterator[range]? - that would make inclusion of end value clear which is nice

anthrotype: > smaller #s first, I think?

what do you mean?

> Any particular reason to not actually use Iterator[range]? - that would make inclusion of end value clear which is nice

not sure about nicer, but in SVG doc records' ranges, the endGlyphID is inclusive, unlike built-in range()

Lib/fontTools/subset/svg.py
```diff
@@ -0,0 +1,235 @@
...
+    sorted_ints = sorted(set(ints))
+    # to group together consecutive ints, we use as 'key' the difference
+    # between their index in the (sorted) list and themselves, which stays
+    # the same for consecutive numbers
+    for _key, group in groupby(enumerate(sorted_ints), lambda i: i[0] - i[1]):
+        consecutive_ints = [v for _i, v in group]
+        yield (consecutive_ints[0], consecutive_ints[-1])
```
rsheeter, line 179: After sorting my instinct is to think you could use a reduce rather than a groupby. You add the new value to the last list in the accumulator if the last list is empty or it's last value is v-1 type of deal.

Lib/fontTools/subset/svg.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+@_add_method(ttLib.getTableClass("SVG "))
+def subset_glyphs(self, s) -> bool:
+    if etree is None:
+        raise ModuleNotFoundError("No module named 'lxml', required to subset SVG")
```
rsheeter, line 185: Depending how helpful we want to be "Consider pip install lxml" or some such?

anthrotype: too much hand-holding. and pip is not the only way to install (e.g. liunx distro)

Lib/fontTools/subset/svg.py
anthrotype: we can't destroy them, because they may contain other elements with id="glyph{digit}" to keep.
I think renaming all the id="glyph{digit}" attributes can logically be done in one go within the same method, whether we are keeping the glyph elements because their id is in our font subset (in which case we want to remap them from old to new glyph indices), or whether we are keeping them indirectly either because they are x-referenced from elements that we are keeping, or because they contain elements that we are keeping.
If we didn't do it in one go, there could be an intermediate state where clashes may still occur (we are updating the tree in-place).
We also return a id_map that has all the final old:new glyph id strings, with the values renamed to match the subset font's new glyph indices or renamed to avoid element id clashes. We can use this in the following `update_glyph_href_links` method where we search for all the `xlink:href="#glyph` references and update those as well.

Lib/fontTools/subset/svg.py
anthrotype: Done https://github.com/fonttools/fonttools/pull/2452/commits/40cf50f97d64a34d6620ff4b8f196a682e8a102f

Lib/fontTools/subset/__init__.py
```diff
@@ -2858,13 +2862,7 @@ def _subset_glyphs(self, font):
 				del font[tag]
 
 		with timer("subset GlyphOrder"):
```
rsheeter, line 2864: Maybe it doesn't matter but it seems like this is now timing something very different than before?

Lib/fontTools/subset/svg.py
rsheeter: Is it heavy? - I think I'd have thought of that as "lighter" than running code every time this file loads but maybe my perception of python "weight" is off

Lib/fontTools/subset/svg.py
rsheeter: Oh nice, I think we have the wrong type all over nanoemoji then :D

Lib/fontTools/subset/svg.py
rsheeter: You could use al then I think?

Lib/fontTools/subset/svg.py
rsheeter: Re smaller numbers, the comment could indicate that hte min,max ranges are both sorted and non-overlapping. As stated neither of those things is necessarily true.

Lib/fontTools/subset/svg.py
rsheeter: Nothing, my thought would just be to use the simplest tool that can solve the problem. Contrasting the "roughly equivalent to" for reduce vs group by would leave me thinking of reduce as "lighter" than groupby.

Tests/subset/svg_test.py
```diff
@@ -0,0 +1,455 @@
...
+
+    return fb.font
+
+
+def new_svg(**attrs):
+    return etree.Element("svg", {"xmlns": NAMESPACES["svg"], **attrs})
+
+
+def test_subset_svg_simple(empty_svg_font, tmp_path):
+    # 'simple' as in one glyph per svg doc
```
rsheeter, line 51: test_subset_otsvg_single_glyph_per_svg_doc (or some such) and drop the comment? - intent being clearer notion of what broke when a failure reports

Tests/subset/svg_test.py
```diff
@@ -0,0 +1,455 @@
...
+        "</svgDoc>",
+        '<svgDoc endGlyphID="3" startGlyphID="3">',
+        '  <![CDATA[<svg xmlns="http://www.w3.org/2000/svg"><path id="glyph3" d="M5,5"/></svg>]]>',
+        "</svgDoc>",
+        '<svgDoc endGlyphID="4" startGlyphID="4">',
+        '  <![CDATA[<svg xmlns="http://www.w3.org/2000/svg"><path id="glyph4" d="M6,6"/></svg>]]>',
+        "</svgDoc>",
+    ]
+
+    # same four glyphs, now retain gids
```
rsheeter, line 90: Would prefer to see do/don't retain gids report as separate test results. You can still share the code. One thought would be to make it a parameterized test on arguments to subset and expected xml.

rsheeter (approved):

Lib/fontTools/subset/__init__.py
rsheeter: If the convention is if it works don't drop it then it makes sense to remove it. I was interpreting it in a somewhat different light, more like tables we recommend be used. In retrospect several other things would drop by default if that was true so ... ignore me :D

Lib/fontTools/subset/svg.py
anthrotype: alright, I added lru_cache https://github.com/fonttools/fonttools/pull/2452/commits/f1e9241e9f7ed94247471efa1429d83ee31bd9cd

Lib/fontTools/subset/svg.py
anthrotype: In the end I changed it to use neither groupby nor reduce, just a simple generator that yields (start, end) when v-1 != end https://github.com/fonttools/fonttools/pull/2452/commits/55b2d4bc3b8228814533fc9a2cde7811353dc302

Lib/fontTools/subset/__init__.py
anthrotype: yeah, there wasn't much to time even before the change, but I'll keep it there.

Tests/subset/svg_test.py
anthrotype: Done 8bfeb3c

Lib/fontTools/subset/svg.py
anthrotype: `all` would also short-circuit (as soon as one is False). Actually I can use a list comprehension to run subset_elements on all children and avoid any() to short-circuit. I'll do that

PR #2453, opened 2021-11-18
[docs] Fix sphinx warnings

simoncozens:
It's currently hard to write new docs because there are so many warnings in the Sphinx output about bad markup that you can't see if any new documentation contains bad syntax. This PR fixes up all the markup errors and other Sphinx problems, such that the docs build without warnings. (Well, only one, about sympy, but now you can see any new warnings much more easily.)


belluzj:
@simoncozens Just so you know I'm updating the documentation in designspaceLib as part of #2436. I see that you've started touching a tiny bit the designspace files, in order to fix warnings there. That's fine and won't be an issue to merge but if you could not change too much that part of the documentation that would be helpful. Can I also do anything on my side to prevent conflicts later? (I'll make sure not to add more warnings)

simoncozens:
It's no problem. Designspace documentation is already pretty good :-) I'm focusing on the TrueType tables, which are the *really useful* but *really undocumented* bit of fontTools...

anthrotype (approved):

PR #2457, opened 2021-11-22
[docs] Fully document the glyf table

simoncozens:


dscorbett (changes requested):

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -215,16 +244,33 @@ def fromXML(self, name, attrs, content, ttFont):
...
+
+		Args:
+			glyphOrder ([str]): List of glyph names in order.
+		"""
 		self.glyphOrder = glyphOrder
 
 	def getGlyphName(self, glyphID):
+		"""Returns the name for the glyph with the given ID.
+
+		Raises a ``KeyError`` the glyph name is not found in the font.
```
dscorbett, line 257: “`KeyError` the glyph name” should be “`KeyError` if the glyph ID”.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -215,16 +244,33 @@ def fromXML(self, name, attrs, content, ttFont):
...
+		"""
 		return self.glyphOrder[glyphID]
 
 	def getGlyphID(self, glyphName):
+		"""Returns the ID of the glyph with the given name.
+
+		Raises a ``KeyError`` the glyph name is not found in the font.
```
dscorbett, line 264: “`KeyError` the” should be “`ValueError` if the”.

dscorbett (changes requested):

anthrotype (approved):
thanks!

PR #2462, opened 2021-12-02
subset/COLR: fix struct.error while subsetting Bungee COLR

### LINKED ISSUE #2461: error while subsetting Bungee COLR
Take this font from https://github.com/google/fonts/blob/17fd8411fad0c1fb14e4c8bd18c87f45f65454f8/ofl/bungeecolor/BungeeColor-Regular.ttf

and do

```
$ fonttools subset BungeeColor-Regular.ttf --text="À" --glyph-names --verbose
```

you get this struct.error:

```
[...]
  File "python3.9/site-packages/fontTools/ttLib/tables/otConverters.py", line 356, in write
    writer.writeValue(self.typecode, font.getGlyphID(value))
  File "python3.9/site-packages/fontTools/ttLib/tables/otBase.py", line 438, in writeValue
    self.items.append(struct.pack(f">{typecode}", value))
struct.error: ('required argument is not an integer', 'str', 2, 'LayerRecord[]', 'LayerRecordArray')
```

anthrotype:
Fixes #2461 

.. or will fix after I push the actual fix. For now it's just a test that fails with struct.error.

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -2064,6 +2064,20 @@ def subset_glyphs(self, s):
 	from fontTools.colorLib.unbuilder import unbuildColrV1
 	from fontTools.colorLib.builder import buildColrV1, populateCOLRv0
 
+	# only include glyphs after COLR closure, which in turn comes after cmap and GSUB
+	# closure, but importantly before glyf/CFF closures. COLR layers can refer to
+	# composite glyphs, and that's ok, since glyf/CFF closures happen after COLR closure
+	# and take care of those. If we also included glyphs resulting from glyf/CFF closures
+	# when deciding which COLR base glyphs to retain, then we may end up with a situation
+	# whereby a COLR base glyph is kept, not because directly requested (cmap)
+	# or substituted (GSUB) or referenced by another COLRv1 PaintColrGlyph, but because
+	# it corresponds to (has same GID as) a non-COLR glyph that happens to be used as a
+	# component in glyf or CFF table. Best case scenario we retain more glyphs than
+	# required; worst case we retain incomplete COLR records that try to reference
+	# glyphs that are no longer in the final subset font.
+	# https://github.com/fonttools/fonttools/issues/2461
+	s.glyphs = s.glyphs_colred
+
 	self.ColorLayers = {g: self.ColorLayers[g] for g in s.glyphs if g in self.ColorLayers}
 	if self.version == 0:
 		return bool(self.ColorLayers)
@@ -2776,6 +2790,7 @@ class Subsetter(object):
 					log.info("Closed glyph list over '%s': %d glyphs after",
 							 table, len(self.glyphs))
 					log.glyphs(self.glyphs, font=font)
+			setattr(self, f"glyphs_{table.lower()}ed", frozenset(self.glyphs))
 
 		if 'glyf' in font:
 			with timer("close glyph list over 'glyf'"):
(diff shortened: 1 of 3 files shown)


Lib/fontTools/subset/__init__.py
```diff
@@ -2776,6 +2790,7 @@ def _closure_glyphs(self, font):
 					log.info("Closed glyph list over '%s': %d glyphs after",
 							 table, len(self.glyphs))
 					log.glyphs(self.glyphs, font=font)
+			setattr(self, f"glyphs_{table.lower()}ed", frozenset(self.glyphs))
```
rsheeter, line 2793: Nit: it appears we unnecessarily write glyphs_bslned if bsln [what the heck is a bsln] exists?

Tests/subset/subset_test.py
```diff
@@ -1352,5 +1352,55 @@ def test_subset_svg_missing_lxml(ttf_path):
...
+        "Agrave.alt002",
+        "grave.alt002",
+    ]
+
+    color_layers = subset_font["COLR"].ColorLayers
+    assert ".notdef" in color_layers
+    assert "Agrave" in color_layers
+    # 'grave' not included in COLR because absent after COLR glyph closure;
+    # only added in subset font after subsequent glyf closure.
+    assert "grave" not in color_layers
```
rsheeter, line 1401: Suggested clarification

# Agrave 'glyf' uses grave. It should be retained in 'glyf' but NOT in COLR when we subset down to Agrave.

Tests/subset/subset_test.py
```diff
@@ -1352,5 +1352,55 @@ def test_subset_svg_missing_lxml(ttf_path):
...
+        "A",
+        "grave",  # included because glyf component of Agrave
+        "Agrave",
+        ".notdef.alt001",
+        ".notdef.alt002",
+        "A.alt002",
+        "Agrave.alt001",
+        "Agrave.alt002",
+        "grave.alt002",
+    ]
```
rsheeter, line 1394: Seems like more glyphs than you should need to minimally demonstrate the problem? - IIUC two glyf entries, one a component of the other, and two COLR glyphs, each using one of the glyf entries, would suffice? - subset down to only the glyf that has the other as a component and we should see the COLR glyph corresponding to the component get dropped.

IMO it might clearer if glyphs were renamed as well, e.g. glyf_with_colr1, glyph_with_colr2, glyf_with_colr1.component1, ... or w/e, not tied to those names, just hoping clearer names might be possible.

rsheeter (approved):
Release please :)

Lib/fontTools/subset/__init__.py
anthrotype: [`bsln`](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6bsln.html) is an Apple AAT table. It's true that attribute is unused, but same for `self.glyphs_cmaped` or `self.glyphs_glyfed`. I'd just keep it there, it's a pattern. Can be useful for debugging.

Tests/subset/subset_test.py
anthrotype: added your comment

Tests/subset/subset_test.py
anthrotype: yes, maybe it could be more minimal to aid clarity, but I prefer to use an existing font instead of making one from scratch or fiddle with it. It's good enough.

anthrotype:
> Release please :)

@rsheeter https://github.com/fonttools/fonttools/releases/tag/4.28.3

PR #2471, opened 2021-12-14
svgPathPen: add option to customize number formatting (e.g. rounding)

anthrotype:
If one wants to draw a TTFont's glyph outlines as SVG paths while at the same time apply some rounding to the float coordinates (e.g. introduced as result of some transformation), at the moment adding a RoundingPen filter to the pen chain is not sufficient when the glyph contains TrueType quadratic splines (qCurveTo). That's because the rounding would occur too early, before the qCurve get split into atomic quadratic Bezier segments (this is done in the superclass BasePen.qCurveTo method, that calls the SVGPathPen._qCurveToOne with each sub-curve), thus leaving unwanted floating-point coordinates in the SVG output.

An alternative approach is to make sure that the qCurves are decomposed before they reach the RoundingPen, by defining a new filter pen that does that, e.g.:

```python
from fontTools.pens.basePen import decomposeQuadraticSegment
from fontTools.pens.filterPen import FilterPen

class DecomposeTTQuadsPen(FilterPen):
    def qCurveTo(self, *points):
        for pt1, pt2 in decomposeQuadraticSegment(points):
            self._outPen.qCurveTo(pt1, pt2)
```

That works but for simplicity, I think it's good to add add to the SVGPathPen an optional ~~'round'~~ `ntos` parameter (that by default does nothing) that takes a `Callable[[float], str]` and can be used to customize the formatting of the SVG coordinates.

### DIFF
--- a/Lib/fontTools/pens/svgPathPen.py
+++ b/Lib/fontTools/pens/svgPathPen.py
@@ -1,18 +1,36 @@
+from typing import Callable
 from fontTools.pens.basePen import BasePen
 
 
-def pointToString(pt):
-    return " ".join([str(i) for i in pt])
+def pointToString(pt, ntos=str):
+    return " ".join(ntos(i) for i in pt)
 
 
 class SVGPathPen(BasePen):
+    """ Pen to draw SVG path d commands.
 
-    def __init__(self, glyphSet):
+    Example::
+        >>> pen = SVGPathPen(None)
+        >>> pen.moveTo((0, 0))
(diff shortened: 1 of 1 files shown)


anthrotype:
I renamed the option, instead of `round` callable takes float and returns float, I named the parameter `ntos` and it should now return a string directly (defaults to built-in `str`).

EsoFromTheEast (approved):

PR #2472, opened 2021-12-14
[feaLib] Fail with error code when build fails

### LINKED ISSUE #2459: [feaLib] "fonttools feaLib" should error out, not continue
If there's a parse/build error when using the feaLib command line tool, we currently do this:

https://github.com/fonttools/fonttools/blob/445108f735b22a5ca37f669808d47906d024fe24/Lib/fontTools/feaLib/__main__.py#L69-L73

i.e. we save the font anyway and exit with status code 0.

My Makefiles and I think this is a terrible idea, and I would like to change it. Any objections / thoughts?

simoncozens:
Fixes #2459 and my sanity.

### DIFF
--- a/Lib/fontTools/feaLib/__main__.py
+++ b/Lib/fontTools/feaLib/__main__.py
@@ -70,6 +70,7 @@ def main(args=None):
         if options.traceback:
             raise
         log.error(e)
+        sys.exit(1)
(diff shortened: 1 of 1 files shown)


anthrotype (approved):
LGTM

Issue #1260, opened 2018-05-11
fontTools merge.py: mark2mark OT feature broken


BlackFoundry:
When merging a freshly compiled font from "NotoSansThai-MM.glyphs" font with "NotoSans-Regular.ttf" (Latin) the "mkmk" OT feature seems to be broken when testing the resulting font (mark to mark not stacking as intended)

test case:
[gist](https://gist.github.com/BlackFoundry/7950dde3b67e2deded3729065fb894ec)
resulting NotoSansThaiOnly-Regular.ttf works as expected
while NotoSansMergedThai-Regular.ttf does not.

I am still investigating, but any insight would be much appreciated.

BlackFoundry:
Didn't found the reason of why, and not sure whether it comes from GPOS 'mkmk' or GSUB 'ccmp' features yet.
But swapping the order of the fonts to merge solved the issue: first Thai, second Latin.

behdad:
Is this still valid?

BlackFoundry:
yes: 
if Thai First, Latin second, mark to mark positioning works.
if Latin first, Thai second, mark to mark positioning does not work

behdad:
What's a good test sequence?  Marks in Thai don't work or Latin?

miloush:
Might be related (but also not): merging Noto Sans Grantha before Tamil results in broken vowel attachments and substitutions (red):

![image](https://user-images.githubusercontent.com/10546952/116980178-d5c22900-acbd-11eb-9076-3651d6e452da.png)

Merging Tamil first i.e. `pyftmerge NotoSansTamil-Regular.ttf NotoSansGrantha-Regular.ttf` works as expected (blue). The test string above is ப்பிபீபுபூவு.

BlackFoundry:
Yes I think the Tamil issue you show is related.
@behdad sorry I missed your message from 28 Nov 2018 😅!!

Marks in Thai do not work properly when merged with ordered list: [Latin, Thai]
Thai behaves normally when merged with ordered list: [Thai, Latin]

behdad:
I'm... hacking on the merger as we speak...

behdad:
Basically, we don't merge LookupFlags properly. I'll get on it.

behdad:
Fix in https://github.com/fonttools/fonttools/pull/2474
Please test.

dscorbett:
Are mark classes merged properly too?

satbyy:
[...]

PR #2473, opened 2021-12-14
Merger improvements

behdad:
With this branch I can try merging all of NotoSans Regular fonts:
```
behdad:ttf 1 (main)$ ~/fonttools/fonttools merge --verbose NotoSans*/NotoSans*-Regular.ttf --drop-tables+=vhea,vmtx
```
and hit the 64k glyph limit:
```
WARNING: Dropped mapping from codepoint 0X0374 to glyphId 'uni0374'
WARNING: Dropped mapping from codepoint 0X0374 to glyphId 'uni0374#1'
WARNING: Dropped mapping from codepoint 0X0374 to glyphId 'uni0374#2'
Merged 'cmap'.
Merging 'GSUB'.
WARNING: Have duplicates to resolve for font 37 but no GSUB: {'tildecomb': 'tildecomb#1', 'overlinecomb': 'uni0305#1', 'uni0484': 'uni0484#1', 'uni0487': 'uni0487#1', 'uniA66F': 'uniA66F#1'}
WARNING: Have duplicates to resolve for font 69 but no GSUB: {'hyphen': 'hyphen#17', 'uni02BC': 'uni02BC#5', 'uni02CD': 'uni02CD#1'}
WARNING: Have duplicates to resolve for font 90 but no GSUB: {'zeroguru': 'zeroguru#2', 'oneguru': 'oneguru#2', 'twoguru': 'twoguru#2', 'threeguru': 'threeguru#2', 'fourguru': 'fourguru#2', 'fiveguru': 'fiveguru#2', 'sixguru': 'sixguru#2', 'sevenguru': 'sevenguru#2', 'eightguru': 'eightguru#2', 'nineguru': 'nineguru#2'}
WARNING: Have duplicates to resolve for font 101 but no GSUB: {'uni20B9.beng': 'uni20B9#3'}
WARNING: Have duplicates to resolve for font 105 but no GSUB: {'gravecomb': 'uni0300', 'uni0306': 'uni0306#3', 'uni0307': 'uni0307#6', 'uni0308': 'uni0308#7', 'uni0313': 'uni0313#3', 'uni0483': 'uni0483#3', 'uni20DB': 'uni20DB#1'}
WARNING: Have duplicates to resolve for font 131 but no GSUB: {'hyphen': 'hyphen#24'}
WARNING: Have duplicates to resolve for font 139 but no GSUB: {'uni1736': 'uni1736#3'}
WARNING: Have duplicates to resolve for font 142 but no GSUB: {'uniA78B': 'uniA78B#3', 'uniA78C': 'uniA78C#3'}
WARNING: Have duplicates to resolve for font 149 but no GSUB: {'exclam': 'exclam#16', 'parenleft': 'parenleft#13', 'parenright': 'parenright#13', 'comma': 'comma#15', 'period': 'period#15', 'colon': 'colon#16', 'semicolon': 'semicolon#12', 'uni060C': 'uni060C#4', 'uni061B': 'uni061B#4', 'uni061F': 'uni061F#6', 'uni0660': 'uni0660#3', 'uni0661': 'uni0661#3', 'uni0662': 'uni0662#3', 'uni0663': 'uni0663#3', 'uni0664': 'uni0664#3', 'uni0665': 'uni0665#3', 'uni0666': 'uni0666#3', 'uni0667': 'uni0667#3', 'uni0668': 'uni0668#3', 'uni0669': 'uni0669#3', 'uni066A': 'uni066A#4', 'uni066B': 'uni066B#3', 'uni066C': 'uni066C#3', 'quoteleft': 'quoteleft#11', 'quoteright': 'quoteright#11', 'quotedblleft': 'quotedblleft#12', 'quotedblright': 'quotedblright#12', 'uniFDF2': 'uniFDF2#2', 'uniFDFD': 'uniFDFD#2'}
WARNING: Have duplicates to resolve for font 178 but no GSUB: {'exclam': 'exclam#17', 'quotedbl': 'quotedbl#12', 'quotesingle': 'quotesingle#11', 'parenleft': 'parenleft#14', 'parenright': 'parenright#14', 'asterisk': 'asterisk#10', 'plus': 'plus#9', 'comma': 'comma#16', 'hyphen': 'hyphen#29', 'period': 'period#16', 'slash': 'slash#12', 'nine': 'nine#10', 'colon': 'colon#17', 'semicolon': 'semicolon#13', 'less': 'less#8', 'equal': 'equal#9', 'greater': 'greater#8', 'question': 'question#15', 'asciicircum': 'asciicircum#8'}
WARNING: Have duplicates to resolve for font 181 but no GSUB: {'u3001': 'uni3001#2', 'u3002': 'uni3002#2', 'uni3008': 'uni3008#2', 'uni3009': 'uni3009#2', 'uni300A': 'uni300A#3', 'uni300B': 'uni300B#3', 'u300C': 'uni300C#1', 'u300D': 'uni300D#1', 'u300E': 'uni300E#1', 'u300F': 'uni300F#1', 'uni3010': 'uni3010#1', 'uni3011': 'uni3011#1', 'uni3014': 'uni3014#1', 'uni3015': 'uni3015#1', 'uni3016': 'uni3016#1', 'uni3017': 'uni3017#1', 'uni3018': 'uni3018#1', 'uni3019': 'uni3019#1', 'uni301A': 'uni301A#1', 'uni301B': 'uni301B#1'}
Merged 'GSUB'.

...
  File "/home/behdad/fonttools/Lib/fontTools/ttLib/tables/_g_l_y_f.py", line 898, in compileComponents
    data = data + compo.compile(more, haveInstructions, glyfTable)
  File "/home/behdad/fonttools/Lib/fontTools/ttLib/tables/_g_l_y_f.py", line 1464, in compile
    return struct.pack(">HH", flags, glyphID) + data
struct.error: 'H' format requires 0 <= number <= 65535
```


behdad:
Of course there's also https://github.com/fonttools/fonttools/issues/1260 that should be fixed.

Lib/fontTools/merge.py
```diff
@@ -424,9 +424,15 @@ def _glyphsAreSame(glyphSet1, glyphSet2, glyph1, glyph2):
 	g2 = glyphSet2[glyph2]
 	g1.draw(pen1)
 	g2.draw(pen2)
-	return (pen1.value == pen2.value and
-		g1.width == g2.width and
-		(not hasattr(g1, 'height') or g1.height == g2.height))
+	if pen1.value != pen2.value:
+		return False
+	tolerance = .05 if pen1.value else .20
```
rsheeter, line 429: Not finding the tolerence change entirely intuitive, a comment would help. Your IM idea of ignoring things like CR that never render also makes sense.

Also per IM, warning here would be nice

Lib/fontTools/merge.py
```diff
@@ -424,9 +424,15 @@ def _glyphsAreSame(glyphSet1, glyphSet2, glyph1, glyph2):
...
+	if (g1.width - g2.width) > g1.width * tolerance:
```
rsheeter, line 430: Doesn't this - and height below - give different results for the same two widths depending on whether g1 or g2 has larger/smaller value? - e.g. g1.w=100 g2.w=106 vs g1.w=106 g2.w=100.

I think I'd have expected something along the lines of `abs(g1.width - g2.width) > min(g1.width, g2.width) * tolerance` ?

Lib/fontTools/merge.py
```diff
@@ -436,6 +442,67 @@ class CmapUnicodePlatEncodings:
 	BMP = {(4, 3, 1), (4, 0, 3), (4, 0, 4), (4, 0, 6)}
 	FullRepertoire = {(12, 3, 10), (12, 0, 4), (12, 0, 6)}
 
+def _is_Default_Ignorable(u):
```
rsheeter, line 445: for those as ignorant as me a sentence clarifying what "ignorable" means in this context would be most appreciated :D

rsheeter: Would be nice if this was able to drive itself off unicode data and thus updated by just updating the data files (or the version of something like unicodedata2). Maybe not worthwhile for this alone but it's fairly easy to imagine subsets of unicode datafiles ending up scattered all over the place.

Lib/fontTools/merge.py
```diff
@@ -687,7 +756,10 @@ def merge(self, m, tables):
 	assert len(tables) == len(m.duplicateGlyphsPerFont)
 	for i,(table,dups) in enumerate(zip(tables, m.duplicateGlyphsPerFont)):
 		if not dups: continue
-		assert (table is not None and table is not NotImplemented), "Have duplicates to resolve for font %d but no GSUB: %s" % (i + 1, dups)
+		if table is None or table is NotImplemented:
+			log.warning("Have duplicates to resolve for font %d but no GSUB: %s" % (i + 1, dups))
```
rsheeter, line 760: Can we offer any further advice? - esp something of the form "you should fix X or check on Y"

behdad: I'll add some.

Lib/fontTools/merge.py
behdad: > for those as ignorant as me a sentence clarifying what "ignorable" means in this context would be most appreciated :D

"Default_Ignorable" is a Unicode technical term. I'll add a URL.

behdad: > Would be nice if this was able to drive itself off unicode data and thus updated by just updating the data files (or the version of something like unicodedata2). Maybe not worthwhile for this alone but it's fairly easy to imagine subsets of unicode datafiles ending up scattered all over the place.

Yes. We have this general problem in many places. For much of our unicode data we derive it using scripts. But those of this scale end up being manual.

Lib/fontTools/merge.py
behdad: > 
Doesn't this - and height below - give different results for the same two widths depending on whether g1 or g2 has larger/smaller value? - e.g. g1.w=100 g2.w=106 vs g1.w=106 g2.w=100.

It does give some preference to the first font... which makes sense since the value for that one will be used...

Lib/fontTools/merge.py
behdad: Will add comments.

Lib/fontTools/merge.py
```diff
@@ -473,6 +540,8 @@ def merge(self, m, tables):
 			if oldgid is None:
 				cmap[uni] = gid
 				fontIndexForGlyph[gid] = fontIdx
+			elif _is_Default_Ignorable(uni) or uni in (0x25CC,):
```
anthrotype, line 543: add comment about 0x25CC being "DOTTED CIRCLE"? I had to look that one up

Lib/fontTools/merge.py
anthrotype: we could parse DerivedCoreProperties.txt in MetaTools/buildUCD.py, like we do for blocks or scripts and make them available in fontTools.unicodedata, maybe in a separate PR

Lib/fontTools/merge.py
anthrotype: could even make that tolerance an option, the numbers as such look quite arbitrary

Lib/fontTools/merge.py
anthrotype: I also found that a bit weird. this is basically saying the glyphs are different if the second has slightly wider/taller advance, but same if slightly narrower. I don't immediatey see the logic.

behdad: Right. Meant to have `abs` in there, just not `min`.

behdad:
See https://github.com/fonttools/fonttools/pull/2476
I need to do more work on the `glyphsAreSame` to re-enable it.

behdad:
I addressed comments. Also split the merger code into multiple smaller modules. I like to merge this now, then go onto unifying identical glyphs in a followup PR.

Lib/fontTools/merge.py
```diff
@@ -771,7 +771,10 @@ def merge(self, m, tables):
 	assert len(tables) == len(m.duplicateGlyphsPerFont)
 	for i,(table,dups) in enumerate(zip(tables, m.duplicateGlyphsPerFont)):
 		if not dups: continue
-		assert (table is not None and table is not NotImplemented), "Have duplicates to resolve for font %d but no GSUB: %s" % (i + 1, dups)
+		if table is None or table is NotImplemented:
+			log.warning("Have non-identical duplicates to resolve for font %d but no GSUB. Are duplicates intended?: %s" % (i + 1, dups))
```
anthrotype: Minor optimisation: when interpolating strings in logging messages one usually uses the star-args at the end instead of `% (...)`, this way the string is interpolated only if/when the log message is actually emitted.

```diff
- log.warning("Have non-identical duplicates to resolve for font %d but no GSUB. Are duplicates intended?: %s" % (i + 1, dups))
+ log.warning("Have non-identical duplicates to resolve for font %d but no GSUB. Are duplicates intended?: %s", i + 1, dups)
```

Lib/fontTools/merge/__init__.py
```diff
@@ -20,8 +20,8 @@
 
 
 def _glyphsAreSame(glyphSet1, glyphSet2, glyph1, glyph2,
-                   advanceTolerance=.05,
-                   advanceToleranceEmpty=.20):
+				   advanceTolerance=.05,
```
anthrotype: since you're moving things around and bothering with the whitespace, do you mind running `black` on the whole merge modules?

Lib/fontTools/merge/__init__.py
```diff
@@ -210,17 +99,11 @@ def merge(self, fontfiles):
 		self.fonts = fonts
 		self.duplicateGlyphsPerFont = [{} for _ in fonts]
 
+		compute_mega_cmap(self, [font['cmap'] for font in fonts])
+
 		allTags = reduce(set.union, (list(font.keys()) for font in fonts), set())
 		allTags.remove('GlyphOrder')
 
-		# Make sure we process cmap before GSUB as we have a dependency there.
```
anthrotype: you got rid of this block that places cmap before GSUB in allTags list? Is it safe and the code/comment above was stale?

Lib/fontTools/merge/__init__.py
```diff
@@ -53,6 +53,13 @@ def __init__(self, options=None):
 
 		self.options = options
 
+	def _openFonts(self, fontfiles):
+		fonts = [ttLib.TTFont(fontfile) for fontfile in fontfiles]
+		for font,fontfile in zip(fonts, fontfiles):
+			font.fontfile = fontfile
+			font.name = font['name'].getDebugName(3)
```
anthrotype: hm so you're using the TTFont inner `__dict__` to store additional attributes. It works but I'm not a huge fan, it kind of muddies the interface and risk inadvertently overwriting things (who knows maybe in the future one adds public attributes named like those and forgets about this piece of code?). If you still want to use the object's inner dict, can you prefix the attributes with some underscore or label them e.g. `_meger__fontfile` or something?

anthrotype (approved):
Thanks Behdad! 

I like that you split the module to make it easier to navigate the code. I am not super excited about the blanket star-imports at the top, as it obscures where the symbols come from (they will appear undefined in linters). Would it be too much trouble if you replace them with explicit named imports?

behdad:
Thanks for the review. Great ideas.

Lib/fontTools/merge/__init__.py
behdad: Yes. I restructured the code so it's not needed anymore.

behdad: Because I'm calling `computeMegaCmap` explicitly now.

Lib/fontTools/merge/__init__.py
behdad: Is it just running `python -m black` on it? It's replacing all tabs with spaces and making some dubious decisions. Eg:
```
   return (
        u == 0x00AD
        or u == 0x034F  # Cf       SOFT HYPHEN
        or u == 0x061C  # Mn       COMBINING GRAPHEME JOINER
        or 0x115F <= u <= 0x1160  # Cf       ARABIC LETTER MARK
        or 0x17B4  # Lo   [2] HANGUL CHOSEONG FILLER..HANGUL JUNGSEONG FILLER
        <= u
        <= 0x17B5
        or 0x180B  # Mn   [2] KHMER VOWEL INHERENT AQ..KHMER VOWEL INHERENT AA
        <= u
        <= 0x180D
        or u  # Mn   [3] MONGOLIAN FREE VARIATION SELECTOR ONE..MONGOLIAN FREE VARIATION SELECTOR THREE
        == 0x180E
        or u == 0x180F  # Cf       MONGOLIAN VOWEL SEPARATOR
        or 0x200B <= u <= 0x200F  # Mn       MONGOLIAN FREE VARIATION SELECTOR FOUR
        or 0x202A <= u <= 0x202E  # Cf   [5] ZERO WIDTH SPACE..RIGHT-TO-LEFT MARK
        or 0x2060  # Cf   [5] LEFT-TO-RIGHT EMBEDDING..RIGHT-TO-LEFT OVERRIDE
        <= u
        <= 0x2064
        or u == 0x2065  # Cf   [5] WORD JOINER..INVISIBLE PLUS
        or 0x2066 <= u <= 0x206F  # Cn       <reserved-2065>
        or u == 0x3164  # Cf  [10] LEFT-TO-RIGHT ISOLATE..NOMINAL DIGIT SHAPES
        or 0xFE00 <= u <= 0xFE0F  # Lo       HANGUL FILLER
        or u == 0xFEFF  # Mn  [16] VARIATION SELECTOR-1..VARIATION SELECTOR-16
        or u == 0xFFA0  # Cf       ZERO WIDTH NO-BREAK SPACE
        or 0xFFF0 <= u <= 0xFFF8  # Lo       HALFWIDTH HANGUL FILLER
        or 0x1BCA0 <= u <= 0x1BCA3  # Cn   [9] <reserved-FFF0>..<reserved-FFF8>
        or 0x1D173  # Cf   [4] SHORTHAND FORMAT LETTER OVERLAP..SHORTHAND FORMAT UP STEP
        <= u
        <= 0x1D17A
        or u == 0xE0000  # Cf   [8] MUSICAL SYMBOL BEGIN BEAM..MUSICAL SYMBOL END PHRASE
        or u == 0xE0001  # Cn       <reserved-E0000>
        or 0xE0002 <= u <= 0xE001F  # Cf       LANGUAGE TAG
        or 0xE0020 <= u <= 0xE007F  # Cn  [30] <reserved-E0002>..<reserved-E001F>
        or 0xE0080 <= u <= 0xE00FF  # Cf  [96] TAG SPACE..CANCEL TAG
        or 0xE0100 <= u <= 0xE01EF  # Cn [128] <reserved-E0080>..<reserved-E00FF>
        or 0xE01F0  # Mn [240] VARIATION SELECTOR-17..VARIATION SELECTOR-256
        <= u
        <= 0xE0FFF
        or False  # Cn [3600] <reserved-E01F0>..<reserved-E0FFF>
```

behdad:
PTAL. Addressed all except for running black.

behdad:
Okay. Merging this and continuing.

PR #2477, opened 2021-12-16
[TSI0, TSI5] Derive number of entries to decompile from data length

jenskutilek:
When you add glyphs to a font containing VTT source tables, the number of glyphs will not match the number of entries in the VTT source tables anymore. This causes the decompilation of VTT source tables to fail.

This change fixes this by deriving the number of entries to decompile from the length of the table data in TSI0 and TSI5.

VTT itself also accepts TSI0 and TSI5 tables that are "too short" for the current number of glyphs.


anthrotype:
> the number of glyphs will not match the number of entries

but do you not also update maxp.numGlyphs when you add a glyph to the font?

anthrotype:
oh I see, you do update maxp.numGlyphs, hence the TSI tables become too short.
Can you tests?

what about the other TSI tables?

Lib/fontTools/ttLib/tables/T_S_I__5.py
```diff
@@ -12,13 +12,11 @@
...
 	def decompile(self, data, ttFont):
-		numGlyphs = ttFont['maxp'].numGlyphs
-		assert len(data) == 2 * numGlyphs
 		a = array.array("H")
 		a.frombytes(data)
 		if sys.byteorder != "big": a.byteswap()
 		self.glyphGrouping = {}
-		for i in range(numGlyphs):
+		for i in range(len(data) // 2):
 			self.glyphGrouping[ttFont.getGlyphName(i)] = a[i]
```
anthrotype, line 20: maybe we need to add a check for the case when TSI table is too long, contains more record than are glyphs. Not sure what error you get in this case, but probably best to add a more descriptive one

anthrotype: Also a logging warning may be a good idea when there are more glyphs than TSI records

anthrotype:
to fix the `lint` job failing, you should rebase on origin/master (or merge origin/master into your PR branch) to include this commit https://github.com/fonttools/fonttools/commit/22266631d95422f503570a92c47b4431d0e233d9

jenskutilek:
> what about the other TSI tables?

The other tables don't use the maxp.numGlyphs value. TSI2 uses the code from the TSI0 class and inherits the fix.

khaledhosny:
Are there still interest in merging this PR?

jenskutilek:
> Are there still interest in merging this PR?

Sure! I'll look at the comments and add some tests.

jenskutilek:
I added another detail: When making a test font with VTT hinting, I noticed that VTT will add zero-length `fpgm`, `prep`, and `cvt` tables. Such a font couldn't be roundtripped with ttx. Fixed in 07830f3.

I've added tests and log warnings when the number of TSI0/5 entries and glyphs don't match. Should be good to go now.

PR #2478, opened 2021-12-16
Fixed it 'fontTools.module 'fontTools.ttLib' has no attribute 'woff2'

bsy0317:


anthrotype:
hello! what problem are you trying to solve? How are you importing the woff2 module?

both of these already work:

```python
from fontTools.ttLib import woff2
```

or

```python
from fontTools.ttLib.woff2 import ...
```

PR #2494, opened 2022-01-06
Add FreeType-based Pen for rasterisation

### LINKED ISSUE #2492: Cross-platform Rasterisation Pen?
Is there a handy way to rasterise the paths obtained via pen protocol in a cross-platform manner?

- `CocoaPen`/`QuartzPen` is pretty handy unless you are on a non-macOS environment (e.g. Google Colab)
- `QtPen`/`WxPen` is useful only if it's already used within your codebase as a widget toolkit
- `skia-pathops` is a library dedicated to boolean operations but not for drawing

I might be missing something, but I couldn't find a solution that fully suits my needs. After all, I wrote a pen that finds the existing FreeType installation on the system, constructs `FT_Outline` and let FreeType rasterise the paths. As mentioned in #1769, I agree this is out of scope, but on the other hand I think it makes more sense to wrap it as a pen. If you are interested, I'll share and compile a PR.

takaakifuji:
Add a new pen that rasterises the given outlines with FreeType installed on the system. Closes #2492.

Some implemation notes:

- `FTPen`/`ftPen`: the _ft_ prefix conflicts with fontTools. More verbose naming should be less confusing.
- ~Currently it doesn't use _freetype-py_. Part of the reason is we don't need a access to `FT_Face` or `FT_Glyph` and all I was interested in was included in `ftoutln.c`. Will fix the ctypes portion if _freetype-py_ is preferred.~
- ~`load_freetype_lib()` looks quirky. My homebrew installation is at `~/homebrew`, and the only way I came up to locate the shared library was to use `subprocess`.~
- Dependencies (numpy, matplotlib, pillow) are lazily loaded rather than at the time of import.
- The default values for the metrics (`width=1000, ascender=880, descender=-120`) are questionable. Possibly there is a better way to specify the size of the resulted bitmap.
- For compactness, I embedded zlib-compressed bitmap data in the source code for the test cases.


justvanrossum:
Suggestion to use freetype-py, which already offers a ctypes-based FT wrapper, and it's on PyPI. https://pypi.org/project/freetype-py/

takaakifuji:
Oh, I didn't realise the _freetype-py_ distribution comes with the pre-built binary, thanks for the advice!

m4rc1e:
It would be great if this didn't rely on numpy. I understand this may be a pita though.

Lib/fontTools/pens/ftPen.py
```diff
@@ -0,0 +1,159 @@
...
+        )
+        outline = self.outline(offset=(0, -descender), even_odd=even_odd, scale=scale)
+        err = FT_Outline_Get_Bitmap(freetype.get_handle(), ctypes.byref(outline), ctypes.byref(bitmap))
+        if err != 0:
+            raise FT_Exception(err)
+        return buf.raw, (width, height)
+
+    def array(self, width=1000, ascender=880, descender=-120, even_odd=False, scale=None):
+        # Return a numpy array. Each element takes values in the range of [0.0, 1.0].
+        if not self.np:
```
justvanrossum, line 90: I would not cache PIL, numpy or matplotlib: imports are cached by Python already (via `sys.modules`), so importing again doesn't add a huge cost.

takaakifuji: Absolutely right. Thank you for pointing this out!

takaakifuji:
@m4rc1e The only mandatory dependency is freetype-py. PIL, numpy and matplotlib are optional.

```python
from fontTools.pens.ftPen import FTPen

font = TTFont('SourceSans3-Regular.otf')
glyph_set = font.getGlyphSet()
pen = FTPen(glyph_set)
glyph = glyph_set['A']
glyph.draw(pen)
width, ascender, descender = glyph.width, font['OS/2'].usWinAscent, -font['OS/2'].usWinDescent

# bytes
buf, size = pen.buffer(width, ascender, descender)
print(tuple(buf))

# PIL
print(pen.image(width, ascender, descender))

# numpy.ndarray
a = pen.array(width, ascender, descender)
print(type(a), a.shape)

# matplotlib
pen.show(width, ascender, descender)
```

The code above might give some clues. Currently `FTPen.buffer()` returns `bytes` of an 8-bit grayscale bitmap. It looks to me pixel modes other than `FT_PIXEL_MODE_GRAY` are not that useful.

takaakifuji:
After playing around with [uharfbuzz](https://github.com/harfbuzz/uharfbuzz), I slightly modified the method arguments to make top-to-bottom layout less painful. Also wrote a docstring. See the uharfbuzz example in the docstring for more practical application.

Lib/fontTools/pens/ftPen.py
```diff
@@ -0,0 +1,363 @@
...
+            contain: If ``True``, the image size will be automatically expanded
+                so that it fits to the bounding box of the paths. Useful for
+                rendering glyphs with negative sidebearings without clipping.
+            format: Optional format override. If omitted, the format to use is
+                determined from the filename extension.
+        
+        :Example:
+            >>> pen = FTPen(None)
+            >>> glyph.draw(pen)
+            >>> pen.save('glyph.png' width=500, height=1000)
```
anthrotype, line 311:
```suggestion
            >>> pen.save('glyph.png', width=500, height=1000)
```

Lib/fontTools/pens/ftPen.py
```diff
@@ -0,0 +1,363 @@
+# -*- coding: utf-8 -*-
+
+"""Pen to rasterize paths with FreeType."""
+
+__all__ = ['FTPen']
```
khaledhosny, line 5: Let's call it `freetypePen`/`FreeTypePen` to avoid the unnecessary confusion.

Lib/fontTools/pens/ftPen.py
```diff
@@ -0,0 +1,363 @@
...
+            format: Optional format override. If omitted, the format to use is
+                determined from the filename extension.
+        
+        :Example:
+            >>> pen = FTPen(None)
+            >>> glyph.draw(pen)
+            >>> pen.save('glyph.png' width=500, height=1000)
+        """
+        img = self.image(offset=offset, width=width, height=height, even_odd=even_odd, scale=scale, contain=contain)
+        img.save(fp, format=format, **kwargs)
```
khaledhosny, line 314: I think users can call `image` then save it themselves, wrapping it does not seem to provide much value.

Tests/pens/ftPen_test.py
```diff
@@ -0,0 +1,115 @@
...
+        pen = FTPen(None)
+        draw_cubic(pen)
+        offset, width, height = (0, 0), 500, 500
+        buf1, _ = pen.buffer(offset=offset, width=width, height=height)
+        buf2 = zlib.decompress(base64.b64decode(ZLIB_B64_BIN))
+        self.assertEqual(buf1, buf2)
```
anthrotype, line 44: asserting that the rendered buffer is exactly equal to something may be too precise for our tests, as it could fail unless a specific freetype version is used. Maybe we could relax this a bit, just assert that it is not empty or use Pillow to diff images within a threshold

anthrotype (commented):
Thanks for this, I think it can be very useful!

I left some minor comments below. I agree with Khaled's comment about naming it freeTypePen to avoid confusion with FontTools "ft" prefix, and about the save() method being superfluous.

anthrotype:
I think you need to add `freetype` to the https://github.com/fonttools/fonttools/blob/main/requirements.txt file, otherwise your tests will always be skipped when run on the CI environment.

Tests/pens/freetypePen_test.py
```diff
@@ -31,84 +30,92 @@ def star(pen):
     pen.lineTo((800, -200))
     pen.closePath()
 
+# Assume the buffers are equal when PSNR > 38dB.
```
anthrotype, line 33: pardon my ignorance, I had to google what "PSNR" means. Maybe a little explanation in a comment or a link to the wikipedia page for https://en.wikipedia.org/wiki/Peak_signal-to-noise_ratio could help clarify

anthrotype: also i'm curious about the 38db magic number, does it have any special meaning or is just some abitrary cut-off value?

anthrotype:
looks like the CI tests are failing with NameErrors because pytest is attempting to run your example docstrings as doctests. If those are not meant to be executed, I'd suggest you replace the special string prefix `>>>` with `>>`, like we do elsewhere, e.g.:

https://github.com/fonttools/fonttools/blob/e778f2c3f485101ad11cb595c427f579955c0714/Lib/fontTools/ttLib/ttFont.py#L21-L24

Tests/pens/freetypePen_test.py
takaakifuji: Thanks for the comment. It is a arbitrary cut-off value – at first I randomly took the `draw()` function from another test, and then I realised that I needed a test for quadratic curves as well, so `draw_quadratic()` contains a quadratic contour converted from the cubic one. That's where I chose `38db` as a threshold just to make the test pass when comparing the rendering results. I think the value is neither strict nor loose, but I don't have a good reasoning for the threshold. PSNR was one of the the easiest metrics for me to implement for comparison, but maybe there is better metrics for such a redering test.

anthrotype: I see, thanks. Honestly I'm not so sure about embedding those base64-encoded zlib-compressed binary blobs for the expected renderings in the freetypePen_test.py.. There's no way a reader could make any sense of them without _looking at_ them as an image with pixels on screen. 
Maybe it's not fonttools's job to validate what the exact freetype rendering should be, I believe we could simply accept whatever comes out of freetype as good (provided it doesn't crash python).. What do you think?

takaakifuji:
Thank you so much for taking the time for the review and the feedback! They are so helpful. The examples are not meant to be run as part of doctest, thanks for the suggestion. Another thing I'm not still 100% confident is the API design for positioning, dimension and scaling. I'd appreciate any feedback on it.

anthrotype:
> the API design for positioning, dimension and scaling...

maybe instead of separate offset and scale parameters, you could have a single transform that takes a 6-tuple (2x3 affine transformation)? When you have separate offset and scale it's not obvious what comes first, e.g. if one should first translate and then scale or viceversa. I believe you currently have the offset unscaled (whereas I'd have expected the offset to already be scaled).

Also, for the width/height parameters, I'd have thought they would not be affected by the scale parameter but would instead determine the final exact size of the buffer.

takaakifuji:
Thank you for the feedback! Now changed to pass a single transformation matrix instead of the cluttered offset/scale parameters. I also removed the `width=1000, height=1000` assumption. I think it looks a bit cleaner now. For tests, I added some [PGM format](https://en.wikipedia.org/wiki/Netpbm) images to the `data` directory, which eliminates the awful-looking base64 strings from the source code.

Tests/pens/freetypePen_test.py
takaakifuji: I agree that the tests are not for tracking the changes of FreeType. Still, I think they are valuable to verify the pen is emitting something meaningful. The reason I embedded test data in the source was because I was just hesitant to add test random images to the repo...and I've just come up with a simpler way to store them without size bloats in the latest commits. Please let me know if you have any concerns.

anthrotype (approved):
thank you, LGTM!

anthrotype:
as first time contributor, you may want to add your name to the [Acknowledgements](https://github.com/fonttools/fonttools#acknowledgements) section of the README

anthrotype:
correction, github reminds me you are not a first time contributor, nevertheless feel free to add your name if it's not already there :)

takaakifuji:
Thanks for the review! Modified the requirements/acknowledgements sections in README.

PR #2496, opened 2022-01-08
Fix type of fontName in suckfont

derwind:
In Python 3, `bytes` != `str`, so we explicitly convert to str, since the `fontdir.keys()` are of type `str`.

To check the behavior, for example, do as follows:

```python
>>> from fontTools.t1Lib import T1Font
>>> t1Font = T1Font('Tests/t1Lib/data/TestT1-Regular.pfa')
>>> gs = t1Font.getGlyphSet()
```

### DIFF
--- a/Lib/fontTools/misc/psLib.py
+++ b/Lib/fontTools/misc/psLib.py
@@ -365,6 +365,7 @@ def suckfont(data, encoding="ascii"):
 	m = re.search(br"/FontName\s+/([^ \t\n\r]+)\s+def", data)
 	if m:
 		fontName = m.group(1)
+		fontName = fontName.decode()
 	else:
 		fontName = None
 	interpreter = PSInterpreter(encoding=encoding)


justvanrossum:
I can't reprodruce a problem with the current code. Can you share you test font? Or a minimal test for that shows the problem?

derwind:
Sorry, I didn't explain the problem well. In fact, it may not cause any problem, but the path of "fall back, in case fontName wasn't found" is executed, so I fixed it.

Since it is an internal operation, it is difficult to check, but I have used the two patches attached below:

[psLib.py.before_fix.txt](https://github.com/fonttools/fonttools/files/7833662/psLib.py.before_fix.txt)
[psLib.py.after_fix.txt](https://github.com/fonttools/fonttools/files/7833663/psLib.py.after_fix.txt)

## Behavior before fix

First, apply a patch by

```
patch Lib/fontTools/misc/psLib.py < psLib.py.before_fix.txt
python setup.py install
```

then this shows:

```
>>> from fontTools.t1Lib import T1Font
>>> t1Font = T1Font('Tests/t1Lib/data/TestT1-Regular.pfa')
>>> t1Font.parse()
Not found, so apply fall back process
```

## Behavior after fix

Next, revert the previous patch by

```
git checkout Lib/fontTools/misc/psLib.py
```

and apply another patch by

```
patch Lib/fontTools/misc/psLib.py < psLib.py.after_fix.txt
python setup.py install
```

then this shows:

```
>>> from fontTools.t1Lib import T1Font
>>> t1Font = T1Font('Tests/t1Lib/data/TestT1-Regular.pfa')
>>> t1Font.parse()
Found
```

Therefore, I decided that calling `fontName.decode()` would execute the originally expected path.

justvanrossum:
Ok, makes sense, thanks!

PR #2500, opened 2022-01-12
FreeTypePen: fix missing RTD entry

takaakifuji:
#2494 fails to generate the corresponding RTD entry due to the lack of the dependency:

https://readthedocs.org/api/v2/build/15748934.txt
```
WARNING: autodoc: failed to import module 'freetypePen' from module 'fontTools.pens'; the following exception was raised:
No module named 'freetype'
```

This PR solves the issue by raising ImportError in `__init__`, rather than by adding a line to `Doc/docs-requirements.txt`.

### DIFF
--- a/Doc/docs-requirements.txt
+++ b/Doc/docs-requirements.txt
@@ -1,3 +1,4 @@
 sphinx==4.3.2
 sphinx_rtd_theme==1.0.0
 reportlab==3.6.5
+freetype-py==2.2.0
--- a/Lib/fontTools/pens/freetypePen.py
+++ b/Lib/fontTools/pens/freetypePen.py
@@ -11,17 +11,12 @@ import subprocess
 import collections
 import math
 
-have_freetype = False
-try:
-    import freetype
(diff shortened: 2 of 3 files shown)


anthrotype:
Will the inline documentation get built despite the import error? If not, we should just add freetype-py to the docs requirements

takaakifuji:
- https://fonttools--2494.org.readthedocs.build/en/2494/pens/freetypePen.html
- https://fonttools--2500.org.readthedocs.build/en/2500/pens/freetypePen.html

#2494 emits an empty page while the check shows green. I thought keeping `docs-requirements.txt` minimal was important, so I took the same approach to the Brotli binding  in `fontTools.ttLib.woff2`. If it's okay to add dependencies to `docs-requirements.txt`, I'll do that.

anthrotype:
thanks. It's preferable if the docs requirements include all the optional deps so we all the inline docs gets built.

> I took the same approach to the Brotli binding in fontTools.ttLib.woff2

Hm, I guess that mean we get no RTD autodocs for woff2 module, if so we should fix that.

Issue #2503, opened 2022-01-16
[T1Font] Unable to save charstrings in T1Font after changing them


derwind:
Currently, T1Font does not allow `saveAs` after `parse`. This means that we can't save charstrings after changing them. I wanted to change the hints, so I implemented the ability to recreate the PS stream: https://github.com/derwind/fonttools/commit/49b673b3ba04f71cb3da61c0168f4e18374ddda7

Type 1 is already a legacy format, and Adobe will discontinue support for it next year. If you think the improvement described above is still worthwhile, I would like to make a pull request. What do you think?

In addition to T1_SPEC.PDF, 5015.Type1_Supp.pdf, and PLRM.pdf, I referred to AFDKO's [t1write.c](https://github.com/adobe-type-tools/afdko/blob/develop/c/shared/source/t1write/t1write.c) a lot, because I am only familiar with Type 1 files converted from other formats (e.g. UFO) by AFDKO. Also, I am sorry to say that I have not do tests exhaustively, and it may be an immature implementation.

justvanrossum:
I am not against this although I fear it won't get a lot of use.

I'd be happy to review. Glad to see you wrote tests.

@anthrotype, would you consider this addition to t1Lib?

anthrotype:
Of course, please send a PR! Though I don't know much about T1 fonts so I can only do a superficial review

derwind:
Thank you very much.
At first I hesitated to implement these immature codes by myself, but I was the one who wanted the feature, so rather than bothering you, I thought I would implement it and share it with you.
It may not be used often, but I will make a pull request to help anyone in any way I can.

PR #2504, opened 2022-01-17
Added the ability to recreate the PS stream

derwind:
This is a pull request for #2503.

# Changes

Recreate the  PS stream if necessary during `saveAs`.

# Description

- To avoid affecting the existing behavior, I make sure that the additionally implemented codes will not be executed unless `parse` is executed and `saveAs` is called.
- Some implementations were not derived purely from the Type 1 standards, but were based on AFDKO, so there may be some bias.
- In addition, the Type 1 fonts available were limited for me, so the verifications by me are also limited.


Lib/fontTools/t1Lib/__init__.py
```diff
@@ -367,3 +513,72 @@ def stringToLong(s):
...
+font_dictionary_keys.remove("FontMatrix")
+
+FontInfo_dictionary_keys = copy.copy(_type1_fontinfo_order)
+# extend because AFDKO tx may use following keys
+FontInfo_dictionary_keys.extend([
+	"FSType",
+	"Copyright",
+])
+
+Praivate_dictionary_keys = [
```
justvanrossum, line 532: Typo in Private_dictionary_keys

Tests/t1Lib/t1Lib_test.py
```diff
@@ -52,23 +55,40 @@ def test_read_pfa_write_pfb(self):
...
+		saved_font = self.write(font, 'OTHER', dohex=True, doparase=True)
+		self.assertTrue(same_dicts(font.font, saved_font))
+
 	def test_read_with_path(self):
 		import pathlib
 		font = t1Lib.T1Font(pathlib.Path(PFB))
 
 	@staticmethod
-	def write(font, outtype, dohex=False):
+	def write(font, outtype, dohex=False, doparase=False):
```
justvanrossum, line 82: `doparase` -> `doparse` ?

Lib/fontTools/t1Lib/__init__.py
```diff
@@ -367,3 +513,72 @@ def stringToLong(s):
 	for i in range(4):
 		l += byteord(s[i]) << (i * 8)
 	return l
+
+
+# PS stream helpers
+
+font_dictionary_keys = copy.copy(_type1_pre_eexec_order)
```
justvanrossum, line 520: no need to use the copy module here, you can just do `list(_type1_pre_eexec_order)`

Lib/fontTools/t1Lib/__init__.py
```diff
@@ -367,3 +513,72 @@ def stringToLong(s):
...
+hintothers = \
+	"/OtherSubrs[{}{}{}{systemdict/internaldict known not{pop 3}{1183615869\n" + \
+	"systemdict/internaldict get exec dup/startlock known{/startlock get exec}{dup\n" + \
+	"/strtlck known{/strtlck get exec}{pop 3}ifelse}ifelse}ifelse}executeonly]def"
+# t1write.c:saveStdSubrs
+def bytechrjoin(iterable, joiner=b""):
+	return tobytes(joiner).join(bytechr(item) for item in iterable)
+std_subrs = [
+	# 3 0 callother pop pop setcurrentpoint return
+	bytechrjoin([0x8e, 0x8b, 0xc, 0x10, 0xc, 0x11, 0xc, 0x11, 0xc, 0x21, 0xb]),
```
justvanrossum, line 569: Why not use byte literals, like you do below?

Lib/fontTools/t1Lib/__init__.py
```diff
@@ -367,3 +513,72 @@ def stringToLong(s):
...
+	"ForceBold",
+	"LanguageGroup",
+	"password",
+	"lenIV",
+	"MinFeature",
+	"RndStemUp",
+]
+
+# t1write_hintothers.h
+hintothers = \
```
justvanrossum, line 560: This may look nicer as a triple-quoted string.

justvanrossum (commented):
First round of feedback.

Lib/fontTools/t1Lib/__init__.py
derwind: Thank you. I'll fix it.

Tests/t1Lib/t1Lib_test.py
derwind: Yes... I'll fix them.

Lib/fontTools/t1Lib/__init__.py
derwind: OK. I copied them from [t1write.c](https://github.com/adobe-type-tools/afdko/blob/develop/c/shared/source/t1write/t1write.c#L590-L605) and used them, which is why they look like this.

Lib/fontTools/t1Lib/__init__.py
derwind: OK, I will use `list()`. I checked the [Python 3.7 documents](https://docs.python.org/3.7/library/stdtypes.html#typesseq-list) just to be sure, and it says that list() will make a copy if _iterable_ is already a list.

Lib/fontTools/t1Lib/__init__.py
derwind: I'll use a triple-quoted string.

justvanrossum:
Looks ok to me. @anthrotype, what do you think? Given the small audience this will have, I'm tempted to just merge and not worry too much about further fine tuning the code.

anthrotype:
Sounds good to me, thanks!

PR #2506, opened 2022-01-18
[ufoLib] Expand reserved names and fix illegal characters

madig:
This PR fixes:

1. The "\0" was inside a raw string, meaning it wasn't getting substituted.
2. The "(" and ")" characters are listed in the specification as illegal, but weren't getting substituted.
3. The reserved names were missing LPT5-9 and COM5-9.
4. The previous code had "A:-Z:" literally in the reserved names, but did not expand them, so they were "dead weight". Colons were always replaced first so drive letters never survived to that stage anyway.

It does not fix:

1. Filenames must not end in a period or space.
2. Prefix and suffix aren't being treated.

### DIFF
--- a/Lib/fontTools/ufoLib/filenames.py
+++ b/Lib/fontTools/ufoLib/filenames.py
@@ -3,11 +3,88 @@ User name to file name conversion.
 This was taken from the UFO 3 spec.
 """
 
-illegalCharacters = r"\" * + / : < > ? [ \ ] | \0".split(" ")
-illegalCharacters += [chr(i) for i in range(1, 32)]
-illegalCharacters += [chr(0x7F)]
-reservedFileNames = "CON PRN AUX CLOCK$ NUL A:-Z: COM1".lower().split(" ")
-reservedFileNames += "LPT1 LPT2 LPT3 COM2 COM3 COM4".lower().split(" ")
+# Restrictions are taken mostly from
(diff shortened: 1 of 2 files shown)


belluzj:
What about the drive letters A: to Z: that Cosimo spotted?

belluzj:
(are they handled because the colon is removed before?)

madig:
Ah, right. Added to description.

Lib/fontTools/ufoLib/filenames.py
```diff
@@ -3,11 +3,88 @@
...
-reservedFileNames += "LPT1 LPT2 LPT3 COM2 COM3 COM4".lower().split(" ")
+# Restrictions are taken mostly from
+# https://docs.microsoft.com/en-gb/windows/win32/fileio/naming-a-file#naming-conventions.
+#
+# 1. Integer value zero, sometimes referred to as the ASCII NUL character.
+# 2. Characters whose integer representations are in the range 1 to 31,
+#    inclusive.
+# 3. Various characters that (mostly) Windows and POSIX-y filesystems don't
+#    allow, plus "(" and ")", as per the specification.
+illegalCharacters = {
```
benkiel, line 14: better as a tupple?

justvanrossum: Membership testing should be faster with a set.

benkiel: Got it.

anthrotype (approved):

PR #2512, opened 2022-01-21
Make comments somewhat more intelligent

ctrlcctrlv:
Allows one to be able to write comments more sanely from files, like:

https://github.com/ctrlcctrlv/FRBAmericanCursive/blob/702ab6b90401c9f28bb28d56f1549863ed434fd5/scripts/make_GDEF.py#L55

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -179,13 +179,20 @@ class Expression(Element):
 class Comment(Element):
     """A comment in a feature file."""
 
-    def __init__(self, text, location=None):
+    def __init__(self, text, location=None, raw=False):
         super(Comment, self).__init__(location)
+        self.raw = not text.strip().startswith("#")
         #: Text of the comment
         self.text = text
 
     def asFea(self, indent=""):
-        return self.text
+        if not self.raw:
+            return self.text
+        else:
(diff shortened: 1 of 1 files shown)


Lib/fontTools/feaLib/ast.py
```diff
@@ -179,13 +179,20 @@ class Expression(Element):
 class Comment(Element):
     """A comment in a feature file."""
 
-    def __init__(self, text, location=None):
+    def __init__(self, text, location=None, raw=False):
```
anthrotype, line 182: tha `raw` parameter is unused, remove it?

Lib/fontTools/feaLib/ast.py
```diff
@@ -179,13 +179,20 @@ class Expression(Element):
...
         super(Comment, self).__init__(location)
+        self.raw = not text.strip().startswith("#")
```
anthrotype, line 184: how about you check at contruction time whether the text starts with `#` and if it doesn't you add it at the beginning of each line. basically move the logic currently in asFea to the `__init__`. I would prefer that to having this stateful self.raw. A comment must always start with `# ` by definition; we allow to omit it to make it easier to code with but internally it has (must have) the leading `#` otherwise it isn't a comment.

PR #2513, opened 2022-01-21
[colrv1] ensure base glyphs with equal layers share the same PaintColrLayers

anthrotype:
When building a colrv1 table and the input dictionary contains multiple base glyphs that map to the same layers (equal in content, not identity), we need to make sure the same PaintColrLayers is used for all of them.
Currently what happens is, the subsequent glyphs with same layers content get a new unique PaintColrLayers with only one child, i.e. the shared PaintColrLayers (see failing test).
Fix coming in a follow-up commit.
(Note the first commit eded2086700f3f238aec730000e867e983f3a5b8 is minor, unrelated change)

### DIFF
--- a/Lib/fontTools/colorLib/builder.py
+++ b/Lib/fontTools/colorLib/builder.py
@@ -444,14 +444,12 @@ def _reuse_ranges(num_layers: int) -> Generator[Tuple[int, int], None, None]:
 
 
 class LayerListBuilder:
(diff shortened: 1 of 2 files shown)


anthrotype:
the failing test log for reference:
https://github.com/fonttools/fonttools/runs/4900896406?check_suite_focus=true#step:5:387

PR #2517, opened 2022-01-31
freetypePen: fit to content when rotated/skewed

takaakifuji:
I realised that I didn't give enough thought to the effect of rotate/skew transforms in the former PR (#2494). This PR will fix and clarify the behaviour when `width` or `height` is omitted, which is the use case I failed to cover. Also contains some minor fixes.


Lib/fontTools/pens/freetypePen.py
```diff
@@ -169,33 +169,49 @@ def buffer(self, width=None, height=None, transform=None, contain=False, evenOdd
...
         contain_x, contain_y = contain or width is None, contain or height is None
-        width, height = width or 0, height or 0
         if contain_x or contain_y:
-            bbox = self.bbox
-            bbox = transform.transformPoints((bbox[0:2], bbox[2:4]))
-            bbox = (*bbox[0], *bbox[1])
-            bbox_size = bbox[2] - bbox[0], bbox[3] - bbox[1]
             dx, dy = transform.dx, transform.dy
+            bbox = self.bbox
+            p1, p2, p3, p4 = transform.transformPoint((bbox[0], bbox[1])), transform.transformPoint((bbox[2], bbox[1])), transform.transformPoint((bbox[0], bbox[3])), transform.transformPoint((bbox[2], bbox[3]))
```
anthrotype, line 199: the line is too long, would you mind wrapping around 88 chars?
Or actually you could run `black` to reformat everything to consistent style. We plan to enforce that at some point throughout the codebase, for now we try to use it for newly added modules.

Lib/fontTools/pens/freetypePen.py
```diff
@@ -18,13 +18,13 @@
...
 from freetype.ft_errors import FT_Exception
 
-from fontTools.pens.basePen import BasePen
+from fontTools.pens.basePen import BasePen, PenError
 from fontTools.misc.roundTools import otRound
 from fontTools.misc.transform import Transform
 
 Contour   = collections.namedtuple('Contour', ('points', 'tags'))
 LINE      = 0b00000001
-CURVE     = 0b00000011
```
anthrotype, line 27: ah so all `LINE = CURVE = QCURVE = 0b00000001` basically, might be worth making it explicit in the code

Lib/fontTools/pens/freetypePen.py
takaakifuji: Right, then I'll use Black. Thanks for the suggestion!

anthrotype (approved):

PR #2526, opened 2022-02-09
Extend name table with more general functions

moontypespace:
extend name table with getFamilyName, getSubFamilyName and getNiceFullName + unittests


Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -120,6 +120,37 @@ def getDebugName(self, nameID):
...
+			if name is not None:
+				return name
+		return None
+
+	def getSubFamilyName(self):
+		for nameID in (22, 17, 2):
+			name = self.getDebugName(nameID)
+			if name is not None:
+				return name
+		return None
```
anthrotype, line 135: you do basically the same thing in getFamilyName and getSubFamilyName, with different list of ids, maybe you can factor out this logic into a method named `getBestDebugName` which takes a sequence of name ids integers, and returns one if it finds it or None, so you can use in both of these.

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -120,6 +120,37 @@ def getDebugName(self, nameID):
 		else:
 			return None
 
+	def getFamilyName(self):
+		for nameID in (21, 16, 1):
```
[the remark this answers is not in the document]
moontypespace: @anthrotype Thanks a lot for your valid feedback. I much appreciate it. I added the more general function like getFamilyName, so you don't need to care about the name ID :) I don't think the code much benefit from such long names like TYPOGRAPHIC_SUBFAMILY_NAME

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -120,6 +120,40 @@ def getDebugName(self, nameID):
 		else:
 			return None
 
+	def getBestDebugName(self, nameID):
```
anthrotype, line 123: that's not what I had in mind though. I was thinking of a more generic method like this one:

```python
def getBestDebugName(self, nameIDs):
    for nameID in nameIDs:
        name = self.getDebugName(nameID)
        if name is not None:
            return name
    return None
```

anthrotype: though 'best' is not the _best_ way to name this; it's basically just doing `getFirstDebugName`

Lib/fontTools/ttLib/tables/_n_a_m_e.py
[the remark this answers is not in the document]
anthrotype: at least add a comment next to these constants

Lib/fontTools/ttLib/tables/_n_a_m_e.py
moontypespace: @anthrotype Oh, sorry. My bad. I changed it. Thanks a lot.

moontypespace: @anthrotype  So, should I rename it from getBestDebugName to getFirstDebugName ?

anthrotype: yes, I think it describes better the intent

moontypespace:
> you may want to add your name to the list of contributors in the README, as first-time contributor (at least according to Github)

Looks like it needs another approval because I added my name.

anthrotype (approved):

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -120,6 +120,44 @@ def getDebugName(self, nameID):
...
+		# 22 = WWS SubFamily Name
+		# 17 = Typographic SubFamily Name
+		# 2 = SubFamily Name
+		return self.getFirstDebugName((22, 17, 2))
+
+	def getBestFullName(self):
+		# 4 = Full Name
+		# 6 = PostScript Name
+		for nameIDs in ((21, 22), (16, 17), (1, 2), (4, ), (6, )):
+			if len(nameIDs) == 2:
```
moi15moi, line 146: Why did you check if the lenght of nameIDs is 2?

moontypespace: Hi @moi15moi , thanks for asking. The reason behind is, because the loop starts with the preferred name table ID pairs (21, 22), (16, 17), (1, 2). And if none of them are present (which might be the case in obfuscated web font) then use only the (4, ), (6, ).

moi15moi: Ok, thanks.

I am trying to understand why getBestFullName, getBestFamilyName and getBestSubFamilyName does not always return me the good name.

I did this in my program and it seems to always work: https://github.com/moi15moi/FontCollector/blob/main/fontCollector.py#L304-L321

If you wanna try with an font that the name is not correctly decoded, here is an issue where I give one: https://github.com/fonttools/fonttools/issues/2579

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -120,6 +120,44 @@ def getDebugName(self, nameID):
...
+		# 21 = WWS Family Name
+		# 16 = Typographic Family Name
+		# 1 = Family Name
+		return self.getFirstDebugName((21, 16, 1))
+
+	def getBestSubFamilyName(self):
+		# 22 = WWS SubFamily Name
+		# 17 = Typographic SubFamily Name
+		# 2 = SubFamily Name
+		return self.getFirstDebugName((22, 17, 2))
```
moi15moi, line 140: I don't think we should always priorise the record 22 and 17.

There should be a parameter. For example, in my case, I would like to get the "real" SubFamily Name.

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -120,6 +120,44 @@ def getDebugName(self, nameID):
...
+			name = self.getDebugName(nameID)
+			if name is not None:
+				return name
+		return None
+
+	def getBestFamilyName(self):
+		# 21 = WWS Family Name
+		# 16 = Typographic Family Name
+		# 1 = Family Name
+		return self.getFirstDebugName((21, 16, 1))
```
moi15moi, line 134: I don't think we should always priorise the record 21 and 16.

There should be a parameter to specify what we want prioritized. For example, in my case, I would like to get the "real" Family Name.

PR #2528, opened 2022-02-10
Add windowsNames and macNames parameters to buildStatTable

moontypespace:
The function to create a STAT table created duplicate name table entries, because it did not take care about existing name table entries with the name name. This PullRequest takes care about it.

EDIT(anthrotype): `buildStatTable` calls `nameTable.addMultilingualName` which by default creates both mac and windows names; if the name table only contains one of the two (usually only windows, no mac ones), then `addMultilingualName` can't reuse the existing name IDs (or else it would produce inconsistent nameIDs between mac and windows platforms) and thus it creates duplicate records. This PR adds `windowsNames` and `macNames` parameters to `buildStatTable` so that one can select whether to only add one or both of the two sets. This way, when calling buildStatTable with a name able that only contains windows (or mac) names and there are some that are already present, they won't be unnecessarily duplicated provided that `macNames` (or `windowsNames`) is set to `False`.

### DIFF
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -2826,7 +2826,7 @@ def _buildAxisValuesFormat4(locations, axes, nameTable, windowsNames, macNames):
     return axisValues
 
 
-def _addName(nameTable, value, minNameID=0, windows=True, mac=True):
+def _addName(nameTable, value, minNameID=0, windows, mac):
     if isinstance(value, int):
         # Already a nameID
         return value
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -2802,7 +2802,7 @@ def _buildAxisRecords(axes, nameTable, windowsNames, macNames):
     return axisRecords, axisValues
 
 
-def _buildAxisValuesFormat4(locations, axes, nameTable, windowsNames=True, macNames=True):
+def _buildAxisValuesFormat4(locations, axes, nameTable, windowsNames, macNames):
     axisTagToIndex = {}
     for axisRecordIndex, axisDict in enumerate(axes):
         axisTagToIndex[axisDict["tag"]] = axisRecordIndex
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -2763,7 +2763,7 @@ def buildStatTable(ttFont, axes, locations=None, elidedFallbackName=2, windowsNa
         statTable.AxisValueCount = len(axisValues)
 
 
-def _buildAxisRecords(axes, nameTable, windowsNames=True, macNames=True):
+def _buildAxisRecords(axes, nameTable, windowsNames, macNames):
     axisRecords = []
     axisValues = []
     for axisRecordIndex, axisDict in enumerate(axes):
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -2649,7 +2649,7 @@ AXIS_VALUE_NEGATIVE_INFINITY = fixedToFloat(-0x80000000, 16)
 AXIS_VALUE_POSITIVE_INFINITY = fixedToFloat(0x7FFFFFFF, 16)
 
 
-def buildStatTable(ttFont, axes, locations=None, elidedFallbackName=2, windows=True, mac=True):
+def buildStatTable(ttFont, axes, locations=None, elidedFallbackName=2, windowsNames=True, macNames=True):
     """Add a 'STAT' table to 'ttFont'.
 
     'axes' is a list of dictionaries describing axes and their
@@ -2734,17 +2734,17 @@ def buildStatTable(ttFont, axes, locations=None, elidedFallbackName=2, windows=T
     ttFont["STAT"] = ttLib.newTable("STAT")
     statTable = ttFont["STAT"].table = ot.STAT()
     nameTable = ttFont["name"]
-    statTable.ElidedFallbackNameID = _addName(nameTable, elidedFallbackName, windows=windows, mac=mac)
+    statTable.ElidedFallbackNameID = _addName(nameTable, elidedFallbackName, windows=windowsNames, mac=macNames)
 
     # 'locations' contains data for AxisValue Format 4
-    axisRecords, axisValues = _buildAxisRecords(axes, nameTable, windows=windows, mac=mac)
+    axisRecords, axisValues = _buildAxisRecords(axes, nameTable, windowsNames=windowsNames, macNames=macNames)
     if not locations:
         statTable.Version = 0x00010001
     else:
         # We'll be adding Format 4 AxisValue records, which
         # requires a higher table version
         statTable.Version = 0x00010002
-        multiAxisValues = _buildAxisValuesFormat4(locations, axes, nameTable, windows=windows, mac=mac)
+        multiAxisValues = _buildAxisValuesFormat4(locations, axes, nameTable, windowsNames=windowsNames, macNames=macNames)
         axisValues = multiAxisValues + axisValues
 
     # Store AxisRecords
@@ -2763,13 +2763,13 @@ def buildStatTable(ttFont, axes, locations=None, elidedFallbackName=2, windows=T
         statTable.AxisValueCount = len(axisValues)
 
 
-def _buildAxisRecords(axes, nameTable, windows=True, mac=True):
+def _buildAxisRecords(axes, nameTable, windowsNames=True, macNames=True):
     axisRecords = []
     axisValues = []
     for axisRecordIndex, axisDict in enumerate(axes):
         axis = ot.AxisRecord()
         axis.AxisTag = axisDict["tag"]
-        axis.AxisNameID = _addName(nameTable, axisDict["name"], 256, windows=windows, mac=mac)
+        axis.AxisNameID = _addName(nameTable, axisDict["name"], 256, windows=windowsNames, mac=macNames)
         axis.AxisOrdering = axisDict.get("ordering", axisRecordIndex)
         axisRecords.append(axis)
 
@@ -2777,7 +2777,7 @@ def _buildAxisRecords(axes, nameTable, windows=True, mac=True):
             axisValRec = ot.AxisValue()
             axisValRec.AxisIndex = axisRecordIndex
             axisValRec.Flags = axisVal.get("flags", 0)
-            axisValRec.ValueNameID = _addName(nameTable, axisVal["name"], windows=windows, mac=mac)
+            axisValRec.ValueNameID = _addName(nameTable, axisVal["name"], windows=windowsNames, mac=macNames)
 
             if "value" in axisVal:
                 axisValRec.Value = axisVal["value"]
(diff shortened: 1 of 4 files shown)


anthrotype:
thanks for the PR. I don't think we should modify the behavior of nameTable.addName method. You could simply first try to getName you want and check if the string is the same, then skip adding the name and return the existing nameID.

anthrotype:
ideally this should work for both default (english) strings and for multilingual ones

justvanrossum:
> The function to create a STAT table created duplicate name table entries, because it did not take care about existing name table entries with the name name. This PullRequest takes care about it.

Can you be more specific of where/how this happens?

justvanrossum:
In case my suspion was right: there was some discussion about how #1985 should be a non-issue, and that the limitation in the spec was nonsensical (which is is). If the OT 1.9 spec no longer contains this limitation, I'd be happy to see https://github.com/fonttools/fonttools/commit/90c7c7fae1e886f54a04306611f882b3db4c7509 reverted somehow.

moontypespace:
> 



> 

@justvanrossum Thank you very much for your great input. That was the key :) It looks like it's related to the fact, that I stopped producing Mac name table entries in my fonts and because of that, it's creating duplicate name table entries. To fix that, I added the variable 'platforms', so people like me can specify it :)

Lib/fontTools/otlLib/builder.py
```diff
@@ -2649,7 +2649,7 @@ def build(self):
 AXIS_VALUE_POSITIVE_INFINITY = fixedToFloat(0x7FFFFFFF, 16)
 
 
-def buildStatTable(ttFont, axes, locations=None, elidedFallbackName=2):
+def buildStatTable(ttFont, axes, locations=None, elidedFallbackName=2, platforms=((1, 0, 0), (3, 1, 0x409))):
```
justvanrossum, line 2652: Since you're only using the platform tuples to check "mac" and/or "win", it would be better to use arguments named `windows` and `mac`, defaulting to `True`, so it's the same as in `addMultilingualName`.

Lib/fontTools/otlLib/builder.py
```diff
@@ -2763,21 +2763,21 @@ def buildStatTable(ttFont, axes, locations=None, elidedFallbackName=2):
         statTable.AxisValueCount = len(axisValues)
 
 
-def _buildAxisRecords(axes, nameTable):
+def _buildAxisRecords(axes, nameTable, platforms=((1, 0, 0), (3, 1, 0x409))):
```
justvanrossum, line 2766: So, the same here, but I think for all the private function (starting with `_`) you should need to bother with default values.

Tests/otlLib/builder_test.py
```diff
@@ -1402,6 +1402,85 @@ def test_buildStatTable(axes, axisValues, elidedFallbackName, expected_ttx):
     assert expected_ttx == ttx
 
 
+def test_buildStatTable_name_duplicates():
```
justvanrossum, line 1405: Does the main premise of this test need updating after your latest round of changes?

Perhaps you can minimize it somewhat? Or try to focus it more on the problem that you're fixing.

justvanrossum: Also: I would replace the doc string with a comment. Perhaps slightly reworded to be relevant to the "fixed state", instead of saying "currently" while meaning the broken past :)

Tests/otlLib/builder_test.py
```diff
@@ -1402,6 +1402,55 @@ def test_buildStatTable(axes, axisValues, elidedFallbackName, expected_ttx):
...
+    font_obj["name"].setName('Black', 266, 3, 1, 0x409)
+
+    font_obj["name"].setName('Weight', 270, 3, 1, 0x409)
+
+    expected_names = [x.string for x in font_obj["name"].names]
+
+    builder.buildStatTable(font_obj, AXES, windowsNames=True, macNames=False)
+    actual_names = [x.string for x in font_obj["name"].names]
+
+    assert expected_names == actual_names
```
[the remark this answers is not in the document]
moontypespace: @anthrotype  and @justvanrossum I extended some unittests and they fail. Please see: https://github.com/fonttools/fonttools/pull/2528/commits/51eda21b0a41d325d00bea74d94e6ae57c436a1b

addMultilingualName behaves a bit different then I would expect. This is the reason why I added the test: test_addMultilingualName_no_mac_but_win

moontypespace:
@justvanrossum and @anthrotype  What do you think about that: https://github.com/fonttools/fonttools/pull/2528/commits/ad1d82f810f3f9d7f168e70e1bc02b935cd2bf76

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -311,6 +311,16 @@ def addMultilingualName(self, names, ttFont=None, nameID=None,
 			if nameID is not None:
 				return nameID
 			nameID = self._findUnusedNameID()
+
+			nameID_win = self.findMultilingualName(names, windows=True, mac=False)
```
justvanrossum, line 315: I see what you're trying to do. I'm doubtful about the need: do we really need to handle this case? It seems that if you're building a font from scratch, you're best served by using the same settings for `mac` and `windows` for the entire build. In other words, if you're running into the situation where both mac and windows are requested but only one is present, it may be a symptom of an inconsistency elsewhere. I would be fine with `addMultilingualName` creating more name table entries than strictly necessary in this case.

That said, if we really do want to handle this in an optimal way, it should be structured slightly differently:

1. you're calling `findMultilingualName` for both mac and windows, regardless of whether they were requested.
2. you should check `nameID_win is not None and nameID_mac is None`, as `nameID_win` could legally be 0, which is also false. Same for the next if.
3. The second if should be an elif.
4. After the elif there should be an else in which the `nameID = self._findUnusedNameID()` call is done instead of before everything else, so it isn't called redunantly.

But I'd like some discussion whether we even want to handle this case.

anthrotype: > the situation where both mac and windows are requested but only one is present

I don't think it's worth supporting that use case. Producing an extra nameID is not the end of the world, and as Just said it's probably a hint of some other inconsistency in the way the name table is generated.

moontypespace: @justvanrossum @anthrotype Thank you very, very much for your feedback. I much appreciate the discussion and your valuable comments and thoughts. As discussed I went back one step, added a few more unittests and added some comments for explanation. Please see https://github.com/fonttools/fonttools/pull/2528/commits/e968e8fd7f39198e606e5605148599e5a50f2ffa

justvanrossum (commented):
Sorry, some last nitpicking.

Lib/fontTools/otlLib/builder.py
```diff
@@ -2850,4 +2850,5 @@ def _addName(nameTable, value, minNameID=0):
         return nameID
     else:
         raise TypeError("value must be int, str, dict or list")
-    return nameTable.addMultilingualName(names, minNameID=minNameID)
+
```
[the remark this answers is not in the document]
moontypespace: @justvanrossum Thanks for being picky with the code. I much appreciate high quality code and feedback from experienced people like you. Please see my last changes.

Tests/otlLib/builder_test.py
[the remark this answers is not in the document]
justvanrossum: Is action still needed here?

Tests/ttLib/tables/_n_a_m_e_test.py
```diff
@@ -305,6 +306,36 @@ def test_addMultilingualName_minNameID(self):
 		self.assertGreaterEqual(nameID, 256)
 		self.assertEqual(nameID, table.findMultilingualName(names, minNameID=256))
 
+	def test_addMultilingualName_no_mac_but_win(self):
+		'''
```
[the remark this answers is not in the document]
justvanrossum: This one is still open.

Tests/otlLib/builder_test.py
anthrotype: I think you could rename `test_buildStatTable_name_duplicates` to `test_buildStatTable_platform_specific_names`; the main change in this PR is the addition of `windowsNames` and `macNames` parameter to `buildStatTable`, not the behavior of `addMultilingualNames` as regards name duplicates, which hasn't changed and was already used by `buildStatTable`.

Lib/fontTools/otlLib/builder.py
```diff
@@ -2734,17 +2734,17 @@ def buildStatTable(ttFont, axes, locations=None, elidedFallbackName=2):
...
-    axisRecords, axisValues = _buildAxisRecords(axes, nameTable)
+    axisRecords, axisValues = _buildAxisRecords(axes, nameTable, windowsNames=windowsNames, macNames=macNames)
     if not locations:
         statTable.Version = 0x00010001
     else:
         # We'll be adding Format 4 AxisValue records, which
         # requires a higher table version
         statTable.Version = 0x00010002
-        multiAxisValues = _buildAxisValuesFormat4(locations, axes, nameTable)
+        multiAxisValues = _buildAxisValuesFormat4(locations, axes, nameTable, windowsNames=windowsNames, macNames=macNames)
```
anthrotype, line 2747: ok since we are nitpicking, this line is way too long, we should stay within 88 max characters. Because this `otlLib.builder` module was already passed through the `black` autoformatter once (according to git log), could you please run again the `black` command, on this file only?
```
$ pip install black
$ black Lib/fontTools/otlLib/builder.py
```

anthrotype: Also blacken `Tests/otlLib/builder_test.py` please, thanks

Tests/otlLib/builder_test.py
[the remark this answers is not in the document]
anthrotype: I would have split these into three separate sub-tests, parametrizing input/expected output, but if that's too much hassle, I'm good like this.

Lib/fontTools/otlLib/builder.py
moontypespace: Uh, these commands changed a lot. If this is what you want, please see the latest code commits @anthrotype @justvanrossum

Tests/ttLib/tables/_n_a_m_e_test.py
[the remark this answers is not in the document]
anthrotype: the name of the test `no_mac_but_win` as well as the comment at the top don't quite describe its content. This one is about testing that reuse of existing names does not occur when the existing names/platforms match only partially but not exactly the ones that were requested.

there's already a test named `test_addMultilingualNameReuse` which tests that names fully are reused, or that new names are added; so of the three assertions you only want to keep the second one (about inconsistency); the first and the third assertions are redundant and tested elsewhere.

The test name and description should be amended accordingly.

anthrotype:
@moontypespace you don't need to say "as requested by X or Y" all the time in the commit titles, it's just noise in the git log (especially for trivial things like running black). The reviewers' names are recorded in the PR page.
Besides, a commit message like "changes as requested by X" doesn't communicate anything, one cares more about what the changes are than who requested them. I think we could squash the commits before merging.

anthrotype:
I took the liberty of renaming the PR title and updating the description. "Improve build stat table" was a bit too generic.

anthrotype:
@moontypespace you probably missed this https://github.com/fonttools/fonttools/pull/2528/files/68fa3bbd478e07af9f37d3f85b49050205d5ed13#r804735438 as it was on a line that was later updated so github isn't showing it any more

moontypespace:
> 

@anthrotype You're right. I overlooked it. sorry. Thanks for reminding me. I updated the code.

anthrotype (approved):
we could go on neatpicking, but after a while it becomes subjective. Thanks for the patience. 

LGTM

justvanrossum (approved):
Thanks Olli!

PR #2531, opened 2022-02-20
freetypePen: allow drawing more than 32k points

takaakifuji:
When I try to typeset a longer text with uharfbuzz, `FreeTypePen` fails with the following error:
```
freetype.ft_errors.FT_Exception: FT_Exception:  (raster overflow)
```
This is because the number of the points must be kept < 32K per a `FT_Outline` struct. This PR adds `__enter__`/`__exit__` to the pen, which can be used as follows:
```python
pen = FreeTypePen(None)
for glyph in glyphs:
    with pen as sub_pen:
        glyph.draw(sub_pen)
```
Such 'subpen' allows us to group up contours on a per-glyph basis to avoid possible integer overflow.

### DIFF
--- a/Lib/fontTools/pens/freetypePen.py
+++ b/Lib/fontTools/pens/freetypePen.py
@@ -72,16 +72,17 @@ class FreeTypePen(BasePen):
             from fontTools.misc.transform import Offset
 
             en1, en2, ar, ja = 'Typesetting', 'Jeff', 'صف الحروف', 'たいぷせっと'
-            for text, font_path, direction, typo_ascender, typo_descender, vhea_ascender, vhea_descender, contain, features in (
-                (en1, 'NotoSans-Regular.ttf',       'ltr', 2189, -600, None, None, False, {"kern": True, "liga": True}),
-                (en2, 'NotoSans-Regular.ttf',       'ltr', 2189, -600, None, None, True,  {"kern": True, "liga": True}),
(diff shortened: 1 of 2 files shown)


justvanrossum:
Just checking: are you rendering an entire string of text as a single path? That is almost certainly not a good idea.

justvanrossum:
A "pen" is generally meant to draw the outlines of a single glyph. The "sub pen" concept seems to indicate the FreeTypePen is doing too many things, and should perhaps be factored into two or more classes, where the "pen" part is only responsible for collecting outline data and conerting to a form usable by FreeType.

takaakifuji:
Thank you for pointing this out!

I agree that the changes are getting off the original pen concept, even if it's purely for circumventing the limitation which I didn't noticed at first – I was expecting `FT_Outline` could hold 32-bit number of nodes without thinking, then I could easily hack to use the pen as a generic vector-to-raster surface (like DrawBot). That assumption was too optimistic. So, I realised that I needed 'hints' for where we can find the start of a glyph to break up the collected contours safely, and that's the sub-pen idea came from.

My another idea was to keep the original pen and add a new method/class for bit-blitting bitmaps, but I found it would become so out-of-place and hard to use. That might be another trail that the pen is
solving the problem in a wrong place. `SVGPathPen` is my favourite pen, and I understand the design decision why it doesn't pretend to be `SVGPen` but instead focuses on emitting the SVG path exclusively. In that sense `FreeTypePen` might have been stretching the goal to the weird direction right at the start.

I still think it has some benefits, but I'm not sure if I should give up trying to draw multiple glyphs with this pen. I'll give it more thought to possible different designs in a few days.

PR #2542, opened 2022-03-09
Add debug logger showing the glyph name for which gvar is built

twardoch:
When building gvar, some situations cause fontmake/varLib to fail, for example if a component has a coordinate that's >32k or <-32k. This adds a debug line that prints each glyph name for which gvar is built, so it’s easier to pinpoint where the faulty glyph is.


anthrotype:
i'm surpised to read Github says this is your "first contribution" to fonttools, is that so? :)

twardoch:
When Just and then Behdad were maintainers, I explained my suggestions with words, used issues etc. :) I haven’t contributed code via a PR yet :)

Issue #2060, opened 2020-09-11
issues while attempting to reorder glyphs in a TTFont


anthrotype:
TTFont.setGlyphOrder() simply does what it says on the tin, it sets the TTFont.glyphOrder list.
This works ok when one wants to _rename_ existing glyphs while keeping the same order (e.g. like ufo2ft.postProcessor does for uniXXXX names).
However, if one wants to _reorder_ glyphs (i.e. keeping the same names but changing the glyph indexes) in a font, simply doing TTFont.setGlyphOrder() and TTFont.save() may or may not achieve the result depending on the current state of the TTFont: i.e. depending on whether a given table (that uses this glyphOrder upon save/compile) had in turn been fully decompiled beforehand or not. That's because by design fonttools won't attempt to re-compile (and thus potentially update glyph indexes of) a table that was not previously loaded.

This is further complicated by the fact that some tables (e.g. glyf, OT layout ones, etc.) use the TTFont.lazy attribute to decide whether to fully decompile themselves on read, whereas others like cmap seem to always be decompiled lazily, no matter the lazy attribute.

Moreover, if a TTFont was not opened with lazy=False, there's no straightforward way to make that TTFont instance "eager" or non-lazy (it works if I save to temporary stream and re-open with lazy=False, but seems unelegant).

My immediate problem is I need to re-order glyphs of an existing TTFont in order to make use of OT-SVG glyph sharing feature, whereby glyphs defined within the same OT-SVG document must have contiguous glyph indexes (being referenced as ranges with min/max glyph IDs).

I wish fonttools provided an easier way to change glyph indexes without me having to remember to un-lazify the various tables that use such glyph indexes.

Granted, I could -- should? -- circumvent the issue by computing the final glyphOrder before my TTFont is generated to avoid having to modify its glyphOrder later in the build. But before I end up doing that, I wonder if anybody else faced a similar problem.

justvanrossum:
How about:
- new method to reorder, say `ttf.reorderGlyphs(...)`
- this method requires `lazy` to be False (will throw an error when lazy=True)
- loads all tables (perhaps warns when there are tables that fonttools doesn't know about)
- has some workaround for tables that load lazily even when lazy=False, such as cmap

How difficult this is to implement probably depends on how many tables always load lazily, regardless of the lazy flag.

anthrotype:
> this method requires lazy to be False (will throw an error when lazy=True)

[...]

PR #2544, opened 2022-03-10
Also update glyf's glyphOrder when calling TTFont.setGlyphOrder()

anthrotype:
see https://github.com/fonttools/fonttools/issues/2060#issuecomment-1063932428


rsheeter:
This has tripped me repeatedly and would be most appreciated :)

PR #2551, opened 2022-03-17
add ensureDecompiled method to decompile all the tables irrespective of lazy attribute

anthrotype:
While attempting to tackle glyph reordering in https://github.com/fonttools/fonttools/issues/2060, I needed a way to say "decompile everything in the font". The lazy=False unfortunately does not exactly do what it says to be doing, and that's confusing for people who are not familiar with the fonttools internals (some colleagues did stumble on this). Not only one has to set lazy=False, but one also has to iterate over all the tables and actually decompile them using silly loop like:

```python
for tag in font.keys():
    _ = font[tag]
```

A quick recap about `TTFont.lazy` attribute. It has three states: it defaults to None, and can be optionally be set to True or False.
When it is set to True, the font keeps a reference to the original input file (inside its SFTNReader); when it's None (or False), it copies the raw bytes of the input files to in-memory stream (allows overwriting input, or reading from unseekable streams). A TTFont decompiles a given table only when requested (in `__getitem__`) -- confusingly even when lazy=False.
Then, some tables take into account the lazy attribute to decide whether to load or not all their data upfront. In particular:
1) the glyf table decompile all the glyphs if lazy is False; it keeps them in compact (raw data) form if lazy is True or None (default), so they get expanded upon getting them.
2) All tables that are generated through the otData.py machinery are loaded lazily when lazy=True (`BaseTable.__getattr__` triggeres decompiling upon getting any attribute); they are fully decompiled if lazy is None (default) or False.
3) cmap completely ignores lazy attribute, and always loads its subtables lazily.

***

What this PR proposes to change is:

1) ~~have `TTFont(path, lazy=False)` immediately decompile all the tables' data, everything, like it says on the tin, so one doesn't need to then also decompile individual tables;~~ (decided to reverted this change)
2) add an `ensureDecompiled` method to TTFont, cmap, glyf and the otTables that allows to un-lazify them even when a font had been loaded with lazy=True or None; they get the same as opening it with lazy=False.
3) fix the fact that cmap doesn't respect lazy; make it load everything when lazy=False.

### DIFF
--- a/Lib/fontTools/ttLib/tables/otBase.py
+++ b/Lib/fontTools/ttLib/tables/otBase.py
@@ -4,6 +4,7 @@ import sys
 import array
 import struct
 import logging
+from typing import Iterator, NamedTuple, Optional
 
 log = logging.getLogger(__name__)
 
@@ -608,7 +609,7 @@ class BaseTable(object):
 			self.decompile(reader, font)
 		if recurse:
 			for subtable in self.iterSubTables():
-				subtable.ensureDecompiled(recurse)
+				subtable.value.ensureDecompiled(recurse)
 
 	@classmethod
 	def getRecordSize(cls, reader):
@@ -857,15 +858,36 @@ class BaseTable(object):
 
 		return self.__dict__ == other.__dict__
 
-	def iterSubTables(self):
+	class SubTableEntry(NamedTuple):
+		"""See BaseTable.iterSubTables()"""
+		name: str
+		value: "BaseTable"
+		index: Optional[int] = None  # index into given array, None for single values
+
+	def iterSubTables(self) -> Iterator[SubTableEntry]:
+		"""Yield (name, value, index) namedtuples for all subtables of current table.
+
+		A sub-table is an instance of BaseTable (or subclass thereof) that is a child
+		of self, the current parent table.
+		The tuples also contain the attribute name (str) of the of parent table to get
+		a subtable, and optionally, for lists of subtables (i.e. attributes associated
+		with a converter that has a 'repeat'), an index into the list containing the
+		given subtable value.
+		This method can be useful to traverse trees of otTables.
+		"""
 		for conv in self.getConverters():
-			value = getattr(self, conv.name, None)
+			name = conv.name
+			value = getattr(self, name, None)
 			if value is None:
 				continue
 			if isinstance(value, BaseTable):
-				yield value
+				yield self.SubTableEntry(name, value)
 			elif isinstance(value, list):
-				yield from (v for v in value if isinstance(v, BaseTable))
+				yield from (
+					self.SubTableEntry(name, v, index=i)
+					for i, v in enumerate(value)
+					if isinstance(v, BaseTable)
+				)
 
 
 class FormatSwitchingBaseTable(BaseTable):
--- a/Lib/fontTools/ttLib/tables/otBase.py
+++ b/Lib/fontTools/ttLib/tables/otBase.py
@@ -110,7 +110,6 @@ class BaseTTXConverter(DefaultTable):
 
 	def ensureDecompiled(self):
 		self.table.ensureDecompiled(recurse=True)
-		return self
 
 
 # https://github.com/fonttools/fonttools/pull/2285#issuecomment-834652928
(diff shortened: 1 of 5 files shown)


anthrotype:
i'm gonna add tests but first want to gather feedback

justvanrossum:
There's a lot of code out there that doesn't do `lazy=True`, and I'm afraid this is a huge potential performance regression for those cases.

What about the oposite: have a method that explicitly deompiles all?

TTFont was always about lazy loading, the `lazy` flag is somewhat conflated with "don't buffer the entire stream" and "decompile OTL tables lazily".

anthrotype:
> There's a lot of code out there that doesn't do lazy=True

wait, but lazy=False is opt-in, i'm not changing the default behavior of lazy=None..

anthrotype:
If you grep the code, right now the only difference between lazy=None (default) and lazy=False is in glyf table, which explicitly checks `if ttFont.lazy is False`. The otTables only do `if lazy` or `if not lazy` so there the distinction between None and False doesn't hold. It's either True or it isn't. So currently setting lazy=False has only the effect of loading all the glyphs at once when decompiling the glyf table. That's it. Other than that, it behaves exactly the same as not setting lazy attribute at all.

I would argue that we can give lazy=False a new meaning of load everything eagerly open opening the font can be useful and won't disrupt much existing workflows. If one bothered to set lazy=False to begin with, they definitely are willing to pay the performance hit for whatever reason.

Anyway, if lazy=False as "load everything upfront" turns out to be controvertial, I'm willing to only add the new ensureDecompiled method and keep the current behavior unchanged.

anthrotype:
I removed the auto-load-all-tables-if-lazy-is-false. One has to call `TTFont.ensureDecompiled` to achieve that.

justvanrossum:
(Not a big fan of the "return self for expression convenience" pattern. JS does this all over the place, but eg. Python's `list.append(item)` does not, as doing so might suggest to the naive reader that a new object is returned.)

anthrotype:
> "return self for expression convenience" pattern

hm ok, I reverted

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -862,6 +900,13 @@ def getRecordSize(cls, reader):
 		return NotImplemented
 
 	def getConverters(self):
+		try:
+			fmt = self.Format
+		except AttributeError:
+			# some FormatSwitchingBaseTables (e.g. Coverage) no longer have 'Format'
+			# attribute after fully decompiled, only gain one in preWrite before being
+			# recompiled.
+			return []
```
rsheeter, line 909: I don't find it completely obvious that [] is the right answer in that situation, if it is perhaps explain why in the comment?

rsheeter:
I'm enthused about lazy=False actually decompiling (and not being default). I loaded with lazy=False, checked isLoaded on the TTFont's keys, and had a wtf moment when it reported nothing was loaded.

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -596,13 +600,16 @@ def __getattr__(self, attr):
 
 		raise AttributeError(attr)
 
-	def ensureDecompiled(self):
+	def ensureDecompiled(self, recurse=False):
```
rsheeter, line 603: Doesn't recurse False default mean that by default ensureDecompiled does not actually ensure decompilation?

Lib/fontTools/ttLib/tables/otBase.py
anthrotype: they have no Format, so they don't/can't have format-specific converters, the only possible thing to return is nothing. These hand-coded otTables.py classes provide a nicer API that breaks away with the converters-driven class model, they are supposed to be at the end of a class tree (as leaves), they are not supposed to contain in turn any low-level converters-driven objects.

anthrotype: added comment

Lib/fontTools/ttLib/tables/otBase.py
anthrotype: `BaseTable.ensureDecompiled` pre-existed this PR, it's used in a few places. I don't want to change its default behavior. When calling `font["GSUB"].ensureDecompiled` method you are not calling this method directly, but through the `BaseTTXConverter.ensureDecompiled` (which I added) and the latter calls `BaseTable.ensureDecompiled(recurse=True)`

PR #2552, opened 2022-03-17
Use Harfbuzz Repacker for packing GSUB/GPOS tables

qxliu76:


Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -70,6 +79,12 @@ def compile(self, font):
 			try:
 				writer = OTTableWriter(tableTag=self.tableTag)
 				self.table.compile(writer, font)
+				if have_uHarfbuzz and (self.tableTag == "GSUB" or self.tableTag == "GPOS"):
+					table_data = writer.getAllDataFromHarfbuzz()
+					if len(table_data) == 0:
```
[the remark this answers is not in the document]
anthrotype: ideally I'd like to see something like

```python
try:
    return writer.getAllDataFromHarfbuzz()
except hb.RepackError as e:
    log.error("hb.repack failed, reverting to pure-python serializer; the error message was: %s", e)
    return writer.getAllData()
```

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -70,6 +79,12 @@ def compile(self, font):
 			try:
 				writer = OTTableWriter(tableTag=self.tableTag)
 				self.table.compile(writer, font)
+				if have_uHarfbuzz and (self.tableTag == "GSUB" or self.tableTag == "GPOS"):
```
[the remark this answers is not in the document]
qxliu76: Other tables don't have overflow issue, so we do not need use harfbuzz repacker

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -70,6 +79,12 @@ def compile(self, font):
 			try:
 				writer = OTTableWriter(tableTag=self.tableTag)
 				self.table.compile(writer, font)
+				if have_uharfbuzz and (self.tableTag == "GSUB" or self.tableTag == "GPOS"):
```
khaledhosny, line 82:
```suggestion
				if have_uharfbuzz and self.tableTag in ("GSUB", "GPOS"):
```

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -294,6 +309,23 @@ def getData(self):
 
 		return bytesjoin(items)
 
+	def getDataForHarfbuzz(self):
+		"""Assemble the data for this writer/table with all offset field set to 0"""
+		items = list(self.items)
+		numItems = len(items)
+		for i in range(numItems):
```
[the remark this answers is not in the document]
khaledhosny: Unresolving. You changed the old code and left the new code unchanged.

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -294,6 +306,23 @@ def getData(self):
...
+			item = items[i]
+
+			if hasattr(item, "getData"):
+				#offset value is not needed in harfbuzz repacker, so setting offset to 0 to avoid overflow here
+				if item.offsetSize == 4 or item.offsetSize == 2 or item.offsetSize == 3:
+					items[i] = {2:packUShort, 3:packUInt24, 4:packULong}[item.offsetSize](0)
```
khaledhosny, line 319:
```suggestion
					items[i] = {2: packUShort, 3: packUInt24, 4: packULong}[item.offsetSize](0)
```

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -294,6 +306,23 @@ def getData(self):
...
+				else:
+					raise ValueError(item.offsetSize)
+
+		return bytesjoin(items)
+
+
```
khaledhosny, line 325: One empty line too many.

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -7,6 +7,15 @@
...
 
+have_uharfbuzz = False
+try:
+	import uharfbuzz as hb
+	have_uharfbuzz = True
+except ImportError:
+	pass
+
+
+
```
khaledhosny, line 18: One empty line too many.

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -398,11 +427,92 @@ def _gatherTables(self, tables, extTables, done):
...
+		#merge virtual_links from parent
+		for idx in virtual_edges:
+			virtual_links.append((0, 0, idx))
+
+		sortCoverageLast = False
+		coverage_idx = 0
+		if hasattr(self, "sortCoverageLast"):
+			# Find coverage table
+			for i in range(numItems):
```
khaledhosny, line 444:
```suggestion
			for i, item in enumerate(self.items):
```

khaledhosny:
We probably need to install uharfbuzz on some of the CI jobs to get this new code tested. But we need a new uharfbuzz release first.

setup.py
```diff
@@ -124,6 +124,10 @@ def doraise_py_compile(file, cfile=None, dfile=None, doraise=False):
 	"pathops": [
 		"skia-pathops >= 0.5.0",
 	],
+	# for packing GSUB/GPOS tables with Harfbuzz repacker
+	"repacker": [
+		"uharfbuzz",
```
khaledhosny, line 129: This should require the first uharfbuzz version with the repack method.

qxliu76:
> We probably need to install uharfbuzz on some of the CI jobs to get this new code tested. But we need a new uharfbuzz release first.

could you help cut a uharfbuzz release?

khaledhosny:
> > We probably need to install uharfbuzz on some of the CI jobs to get this new code tested. But we need a new uharfbuzz release first.
> 
> could you help cut a uharfbuzz release?

I’ll do shortly, I need to cut a HarfBuzz release as well, so I might do it first.

qxliu76:
> > > We probably need to install uharfbuzz on some of the CI jobs to get this new code tested. But we need a new uharfbuzz release first.
> > 
> > 
> > could you help cut a uharfbuzz release?
> 
> I’ll do shortly, I need to cut a HarfBuzz release as well, so I might do it first.

thank you!

khaledhosny:
https://github.com/harfbuzz/uharfbuzz/releases/tag/v0.23.0

qxliu76:
> https://github.com/harfbuzz/uharfbuzz/releases/tag/v0.23.0

Thanks, I just updated uharfbuzz version

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -294,6 +307,19 @@ def getData(self):
...
 		return bytesjoin(items)
 
+	def getDataForHarfbuzz(self):
+		"""Assemble the data for this writer/table with all offset field set to 0"""
+		items = list(self.items)
+		for i, item in enumerate(items):
+			if hasattr(item, "getData"):
+				#offset value is not needed in harfbuzz repacker, so setting offset to 0 to avoid overflow here
+				if item.offsetSize == 4 or item.offsetSize == 2 or item.offsetSize == 3:
+					items[i] = {2: packUShort, 3: packUInt24, 4: packULong}[item.offsetSize](0)
```
khaledhosny, line 317: What about you define offsetSizes dict outside of the loop, something like:
```
offsetSizes = {2: packUShort, 3: packUInt24, 4: packULong}
for i, item in enumerate(items):
   ...
   if item.offsetSize in offsetSizes:
   ...
   else:
   ...
```

This is probably more efficient.

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -398,11 +424,89 @@ def _gatherTables(self, tables, extTables, done):
...
 		selfTables.append(self)
 
-	def getAllData(self):
-		"""Assemble all data, including all subtables."""
+	def _gatherGraphForHarfbuzz(self, tables, obj_list, done, objidx, virtual_edges):
+		real_links = []
+		virtual_links = []
+		item_idx = objidx
+
+		#merge virtual_links from parent
```
khaledhosny, line 432:
```suggestion
		# Merge virtual links from parent
```

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -294,6 +307,19 @@ def getData(self):
 
 		return bytesjoin(items)
 
+	def getDataForHarfbuzz(self):
+		"""Assemble the data for this writer/table with all offset field set to 0"""
+		items = list(self.items)
+		for i, item in enumerate(items):
+			if hasattr(item, "getData"):
+				#offset value is not needed in harfbuzz repacker, so setting offset to 0 to avoid overflow here
```
khaledhosny, line 315:
```suggestion
				# Offset value is not needed in harfbuzz repacker, so setting offset to 0 to avoid overflow here
```

anthrotype:
I can review this sometime this week

anthrotype (commented):
the tests you added is not actually exercising the new code because the test environment on the CI is initialized with the list from requirements.txt file, and that does not include uharfbuzz.

You should add this line at the end of requirements.txt file to make it run:

```
uharfbuzz==0.24.1
```

anthrotype:
As soon as I added `uharfbuzz==0.24.1` to the requirements.txt file and run the test suite with the command `tox -e py`, a bunch of tests failed because the checksumAdjustment value in the head table changed, which means harfbuzz repacker is doing its job of repacking while maintaining the rest of the data the same (the binary layout differences are abstracted away in the ttx dump).

We probably need to skip over the checksums differences. I'll try to fix that on main branch, then you can merge main into your PR branch to update.

I'm thinking we may probably want to run the test suite both with _and_ without uharfbuzz installed to confirm it works in either cases. We already do it for another optional dependency (lxml) that is used opportunistically in fontTools.ufoLib only when available.
Maybe for the sake of this PR, simply add uharfbuzz to requirements.txt, which means tests will always use it unconditionally, then I will follow up with another PR to make it run without as well.

anthrotype:
@qxliu76 ok, if you pull and merge origin/main into your branch, and update it like so, then the tests run using uharfbuzz and complete successfully. I fixed on main the few tests that were unnecessarily checking for checksums and timestamps.
 
```diff
diff --git a/requirements.txt b/requirements.txt
index d061b779..008bc8c5 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,3 +12,4 @@ skia-pathops==0.7.2; platform_python_implementation != "PyPy"
 ufoLib2==0.13.0
 pyobjc==8.1; sys_platform == "darwin"
 freetype-py==2.2.0
+uharfbuzz==0.24.1
```

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -70,6 +77,12 @@ def compile(self, font):
 			try:
 				writer = OTTableWriter(tableTag=self.tableTag)
 				self.table.compile(writer, font)
+				if have_uharfbuzz and self.tableTag in ("GSUB", "GPOS"):
+					try:
+						return writer.getAllDataUsingHarfbuzz()
+					except (ValueError, MemoryError, hb.RepackerError) as e:
+						log.error("hb.repack failed, reverting to pure-python serializer; the error message was: %s", e)
+						return writer.getAllData(remove_duplicate=False)
```
anthrotype, line 85: this except branch is currently never hit by the test suite. I know it's difficult to make the harfbuzz repacker fail, but I wonder if you can come up with a test that produces such a situation in order to trigger this and confirm that calling `writer.getAllData(remove_duplicate=False)` actually works.

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -398,11 +425,89 @@ def _gatherTables(self, tables, extTables, done):
...
+		obj_list.append((real_links,virtual_links))
+		item_idx += 1
+		done[id(self)] = item_idx
+		if sortCoverageLast:
+			virtual_edges.pop()
+
+		return item_idx
+
+	def getAllDataUsingHarfbuzz(self):
+		"""Assemble graph data and call Harfbuzz repacker to pack the table."""
```
anthrotype, line 482: for the casual onlooker, maybe it's good if you add a link to the uharfbuzz code where you describe the format of the inputs to hb.repack() method, e.g. https://github.com/harfbuzz/uharfbuzz/blob/c11c7ba/src/uharfbuzz/_harfbuzz.pyx#L1149

Even maybe to that public doc that describes how the hb repacker algorithm works.

anthrotype (commented):
this is looking good. One last thing I am wondering about is wether we need to provide an excape hatch to allow one to _not_ use the harfbuzz repacker even if uharfbuzz is importable. E.g. imagine we find out there's a bug somewhere in the way we hooked things up from python to c++ or anything really, it would be good to be able to disable this feature without requiring one to completely uninstall uharfbuzz in the same environment where fonttools is installed. I'm not sure where/how to add an option for that though. Maybe we can think about that for a follow up PR.

qxliu76:
> @qxliu76 ok, if you pull and merge origin/main into your branch, and update it like so, then the tests run using uharfbuzz and complete successfully. I fixed on main the few tests that were unnecessarily checking for checksums and timestamps.
> 
> ```diff
> diff --git a/requirements.txt b/requirements.txt
> index d061b779..008bc8c5 100644
> --- a/requirements.txt
> +++ b/requirements.txt
> @@ -12,3 +12,4 @@ skia-pathops==0.7.2; platform_python_implementation != "PyPy"
>  ufoLib2==0.13.0
>  pyobjc==8.1; sys_platform == "darwin"
>  freetype-py==2.2.0
> +uharfbuzz==0.24.1
> ```

Thanks! I have updated it accordingly

Lib/fontTools/ttLib/tables/otBase.py
qxliu76: yeah, I actually tested this except branch locally by changing uharfbuzz code, but I don't have a real testcase for this. Let me try and see if I can find or make a testcase

khaledhosny: I tried building GSans with this branch and hb.repack() is always failing.

qxliu76: > I tried building GSans with this branch and hb.repack() is always failing.

could you provide a testcase? I can send it to Garret for debugging. This is also a good testcase for except branch.

khaledhosny: I don’t know how to extract a test case from this, it is being built from UFO source. I get messages like this:
```
ERROR:fontTools.ttLib.tables.otBase:hb.repack failed, reverting to pure-python serializer; the error message was:
INFO:fontTools.ttLib.tables.otBase:Attempting to fix OTLOffsetOverflowError ('GPOS', 'LookupIndex:', 15, 'SubTableIndex:', None, 'ItemName:', None, 'ItemIndex:', None)
```

qxliu76: It looks like harfbuzz repacker failed to find a solution so fonttools falls back to the except branch as expected.
This "OTLOffsetOverflowError" is raised during writer.getAllData().

anthrotype: Yeah, the "attempting to fix offset overflow" is not an "ERROR", in fact it's marked as "INFO" in the logger output. This has been a source of confusions for ages..

anthrotype: @qxliu76 are you going to add a test case that covers the except branch?

anthrotype:
> it would be good to be able to disable this feature without requiring one to completely uninstall uharfbuzz

we can add a configuration option when we add that `fontTools.config` feature that @belluzj is working on in #2416 
For now I think we can ship this as is, with the mere presence of uharfbuzz signalling one wants to use hb.repack

Lib/fontTools/ttLib/tables/otBase.py
qxliu76: yes, Garret said he should be able to generate a graph that fails to repack. I will likely get that next Monday and then add a testcase

khaledhosny: My impression is that the HarfBuzz repacker would be useful specifically in offset overflow cases where FontTools is currently too slow, so I’m just curious why it is failing here.

qxliu76: This is just my guess: If subtables larger than 65kb exist, it's not possible for any offsets to point past them. In these cases the subtable can usually be split into two smaller subtables to allow for more flexibility in the ordering. Currently the harfbuzz repacker does not support table splitting. In the future we would like to explore either adding table splitting into harfbuzz repacker or finding a way to utilize fonttools table splitting before sending the graph to harfbuzz.

qxliu76: @anthrotype I just added a testcase that fails to repack because harfbuzz does not support table splitting. Since fonttoolsI is slow, I noticed that runtime for subset_test.py has bumped from 0.95s to 35s.

qxliu76: @anthrotype Besides runtime bump, the test file is pretty big as well, 14 MB. I wonder if we might need to find a different way to test this. Another way to setup the test would be to use something similar to the mock test, substitute the uharfbuzz object with a mock that always returns failure when the repack method is called. I'm not familiar with mocking stuff in python, not sure if you would have any suggestions.

anthrotype: yeah, it's too big. It's better to mock it.
Take a look at https://docs.python.org/3/library/unittest.mock.html

or if you want, I can add the mocked test after merging this PR

qxliu76: Thank you! I just removed this testcase.

anthrotype (approved):

anthrotype:
I will cut a release next week, before then I want to add the config option to disable hb-repack even when uharfbuzz is importable, using the new config feature just merged. I'll be back to work on Tuesday.

qxliu76:
> I will cut a release next week, before then I want to add the config option to disable hb-repack even when uharfbuzz is importable, using the new config feature just merged. I'll be back to work on Tuesday.

thank you!

Issue #2548, opened 2022-03-16
pyftsubset: problem with SVG table subsetting


frankrolf:
I tested `pyftsubset`’s SVG-table-subsetting capabilities with [Noto Emoji SVG](https://github.com/adobe-fonts/noto-emoji-svg/releases), and found an issue:
```
pyftsubset NotoColorEmoji-SVG.otf --unicodes=U+1F600,U+1F601,U+1F602,U+1F603,U+1F604,U+1F605,U+1F606,U+1F607,U+1F608,U+1F609,U+1F60A,U+1F60B,U+1F60C,U+1F60D
```
(subset to contain only a few emoji)

The resulting SVG table looks like this:

```
  <SVG>
    <svgDoc endGlyphID="1" startGlyphID="1">
[...]

PR #2553, opened 2022-03-18
[subset] fix subsetting OT-SVG when glyph id attribute is on the root <svg> element

anthrotype:
/cc @frankrolf

Fixes https://github.com/fonttools/fonttools/issues/2548

I tested locally with [Noto Emoji OT-SVG](https://github.com/adobe-fonts/noto-emoji-svg/releases) and it works now


frankrolf:
Thank you!

PR #2577, opened 2022-04-09
[CFF/T2] Ensure that pen.closePath() gets called for CFF2 charstrings

justvanrossum:
[CFF/T2] Ensure that pen.closePath() gets called for CFF2 charstrings. Fixes #2455

Issue #2455, opened 2021-11-20
T2CharString.draw not drawing properly


jansindl3r:
Hi, I want to take draw contour from cff font into `defcon's Glyph` object. It's it now drawn correctly as you can see in the attached pictures. Am I doing something wrong or is it something with fontTools? The fact that the same happens in RoboFont makes me believe that I am either one of more people doing something wrong or that something is wrong in FonTools. What do you think? Thanks, Jan

```from fontTools.ttLib import TTFont
from fontTools.cffLib import PrivateDict
from fontTools.pens.ttGlyphPen import TTGlyphPen
from fontTools.pens.pointPen import SegmentToPointPen
from fontTools.pens.t2CharStringPen import T2CharStringPen
from datetime import datetime
from defcon.objects.glyph import Glyph


font = TTFont("sourceSerif.otf")
cff2 = font["CFF2"]

if hasattr(cff2, "desubroutinize"):
    print(cff2.desubroutinize())


content = cff2.cff[cff2.cff.keys()[0]]
letter = content.CharStrings["A"]

print(letter)
glyph = Glyph()
pen = glyph.getPen()
letter.draw(pen)
go = cff2.getGlyphOrder()
cff_pen = T2CharStringPen(None, go, CFF2=True)
glyph.draw(cff_pen)

private = content.CharStrings["A"].private
content.CharStrings["O"] = cff_pen.getCharString(private=private)

font.save("output.otf")
```
drawing shape of A into glyph O
<img width="296" alt="Screenshot 2021-11-20 at 19 29 57" src="https://user-images.githubusercontent.com/44162266/142737294-654c5157-fa98-4d53-a814-9b31524e3d25.png">

the same issue happens when opening sourceSerif.otf in RoboFont
<img width="83" alt="Screenshot 2021-11-20 at 19 30 28" src="https://user-images.githubusercontent.com/44162266/142737309-65175eb7-0e75-4b18-b179-19cdee053b1a.png">
[bug.zip](https://github.com/fonttools/fonttools/files/7575039/bug.zip)

justvanrossum:
The problem seems to be that for CFF2, charstring.draw() doesn't output the final `pen.closePath()`. Looks like a bug to me.

justvanrossum:
Adding `extractor.endPath()` after this line seems to fix it:
[...]

Issue #1899, opened 2020-04-27
T2CharString.draw() does not emit final closePath() for CFF2 charstrings


justvanrossum:
This is likely caused by `endchar` being implicit in CFF2. Investigating.

justvanrossum:
Adding this at the end of `SimpleT2Decompiler.execute()` fixes it:

```python
		if getattr(self.private, "_isCFF2", False):
			self.op_endchar(None)
```

But I'm not sure that's the cleanest solution.

anthrotype:
maybe you could just call `pen.closePath()` from `_TTGlyphCFF.draw()` (or even better define a new `_TTGlyphCFF2` class in `fontTools.ttLib.ttFont` module and return that from `TTFont.getGlyphSet`).

anthrotype:
@justvanrossum  how about a patch like this?

```diff
diff --git a/Lib/fontTools/misc/psCharStrings.py b/Lib/fontTools/misc/psCharStrings.py
index b894653d..4b6c83d8 100644
--- a/Lib/fontTools/misc/psCharStrings.py
+++ b/Lib/fontTools/misc/psCharStrings.py
@@ -508,6 +508,18 @@ class T2OutlineExtractor(T2WidthExtractor):
                self.currentPoint = (0, 0)
                self.sawMoveTo = 0

+       def execute(self, charString):
+               super().execute(charString)
+
+               # endchar was removed in CFF2 (it's now implied); ensure the last path is closed
+               try:
+                       isCFF2 = self.private._isCFF2
+               except AttributeError:
+                       pass
+               else:
+                       if isCFF2:
+                               self.endPath()
+
        def _nextPoint(self, point):
                x, y = self.currentPoint
                point = x + point[0], y + point[1]
```

I think it's the job of T2OutlineExtractor to call the pen's closePath, not SimpleT2Decompiler to call endchar operator which is not there.

I'm doing `try.. except AttributeError` instead of `getattr(self.private, "_isCFF2", False)` because even `self.private` may not be there.. For some arcane reasons the `T1OutlineExtractor` is a sub-class of `T2OutlineExtractor` but it has a different constructor and doesn't set self.private attribute...

Do you have a test case?

anthrotype:
the `T2OutlineExtractor.closePath` method can be safely called more than once, as it's no-op when it has not seen a previous moveTo

justvanrossum:
> how about a patch like this?

I like that, that seems the cleanest, most correct way.

[...]

PR #2580, opened 2022-04-13
Handle implicit CFF2 closepath within T2OutlineExtractor

skef:
The very recently merged PR #2577 fixes the direct issue of #2455 and the surface issue in close-to-duplicate #1899 but the problem remains for any code inheriting from or using `T2OutlineExtractor` outside of the `T2CharString` context. The way things are left anyone making other use of `T2OutlineExtractor` also needs to add an `endPath()` call after calling `execute()` -- an undocumented and unusual requirement -- to work correctly with CFF2.

Other questions aside, from a basic object encapsulation standpoint it seems better to handle the issue from within T2OutlineExtractor (which is the level at which `closePath()` starts to be supported) rather than outside of it. 

This PR represents one approach, already discussed a bit in #1899 . I won't pretend this fills me with engineering joy -- it feels kind of grubby. But it works, both in terms of the test suite and my relevant code.

About that test suite -- it also passes without the `isSubr` addition, implying that there are no current tests catching spurious `closePath()`s. I'm willing to try to construct such a test as part of this PR assuming a bit of guidance.

### DIFF
--- a/Lib/fontTools/misc/psCharStrings.py
+++ b/Lib/fontTools/misc/psCharStrings.py
@@ -502,11 +502,20 @@ class T2OutlineExtractor(T2WidthExtractor):
 		T2WidthExtractor.__init__(
 			self, localSubrs, globalSubrs, nominalWidthX, defaultWidthX, private)
 		self.pen = pen
+		self.subrLevel = 0
 
 	def reset(self):
 		T2WidthExtractor.reset(self)
 		self.currentPoint = (0, 0)
 		self.sawMoveTo = 0
+		self.subrLevel = 0
+
+	def execute(self, charString):
+		self.subrLevel += 1
+		super().execute(charString)
+		self.subrLevel -= 1
+		if self.subrLevel == 0:
+			self.endPath()
 
 	def _nextPoint(self, point):
 		x, y = self.currentPoint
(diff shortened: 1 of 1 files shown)


skef:
This approach works as well. Would it make anyone happier?

```
diff --git a/Lib/fontTools/misc/psCharStrings.py b/Lib/fontTools/misc/psCharStrings.py
index 6014abc2..4e89b690 100644
--- a/Lib/fontTools/misc/psCharStrings.py
+++ b/Lib/fontTools/misc/psCharStrings.py
@@ -502,11 +502,20 @@ class T2OutlineExtractor(T2WidthExtractor):
                T2WidthExtractor.__init__(
                        self, localSubrs, globalSubrs, nominalWidthX, defaultWidthX, private)
                self.pen = pen
+               self.subrLevel = 0
 
        def reset(self):
                T2WidthExtractor.reset(self)
                self.currentPoint = (0, 0)
                self.sawMoveTo = 0
+               self.subrLevel = 0
+
+       def execute(self, charString):
+               self.subrLevel += 1
+               super().execute(charString)
+               self.subrLevel -= 1
+               if self.subrLevel == 0:
+                       self.endPath()
 
        def _nextPoint(self, point):
                x, y = self.currentPoint
```

justvanrossum:
Yeah, that feels a bit cleaner indeed.

A concern I had earlier while discussing #2455 that still stands: I wonder if we can avoid calling closePath twice for CFF by changing this:

https://github.com/fonttools/fonttools/blob/169731c7f5d96e96d531d35db477af5e727fe945/Lib/fontTools/misc/psCharStrings.py#L537-L540

into this:

```python
	def endPath(self):
		# In T2 there are no open paths, so always do a closePath when
		# finishing a sub path.
		if self.sawMoveTo:
			self.closePath()
```

justvanrossum:
(My last suggestion may be better in a new PR, as it's more of a followup to #2577 really, so feel free to ignore.)

skef:
@justvanrossum I'll redo this PR and take a look at that `closePath()` question as part of it.

skef:
OK, the PR is updated

justvanrossum (approved):
LGTM

PR #2586, opened 2022-04-18
chore(ci): add informational Codecov status checks

thomasrockhu-codecov:
Hi, Tom from Codecov here. I noticed that you were using Codecov but weren't actually getting any notifications on pull requests. I figured it would be useful to get some idea if code being changed is being tested, but also not blocking CI/merging. Let me know if this makes sense or if we can do something that would be helpful.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,5 +1,9 @@
 comment: false
 coverage:
     status:
-        project: off
-        patch: off
+        project:
+            default:
+                informational: true
+        patch:
+            default:
+                informational: true


anthrotype:
thanks, but we deliberately turned those off, as we find them a bit too noisy. We prefer to manually check the codecov website from time to time.

anthrotype:
ok I said _we_, but should have probably said _I_. If other fonttools devs think otherwise, feel free to reopen this.

anthrotype:
@thomasrockhu-codecov what I find noisy is the email notification which distracts the conversation. If there's a way to see some status/check in the PR page without also getting the email notification I'd be in favour of this.

justvanrossum:
I find codecov an utter nuisance, and I agree with closing.

anthrotype:
@thomasrockhu-codecov so do you confirm this PR will not trigger any additional comments from the codecov bot with accompanying email notifications, but only an informational (non-blocking) status check?

anthrotype:
@thomasrockhu-codecov I just did that, care to try? thanks

PR #2589, opened 2022-04-21
define USE_HARFBUZZ_REPACKER config option

anthrotype:
As discussed in #2552, this adds a config option to disable serializing GSUB/GPOS with uharfbuzz even when the latter can be imported.

I also took this opportunity to make some changes to the configTools. The Config mapping methods accept either a str (name of an option) or the Option object itself, which I think can be handy and avoids repetition when the latter is already in scope.

### DIFF
--- a/Lib/fontTools/misc/configTools.py
+++ b/Lib/fontTools/misc/configTools.py
@@ -17,9 +17,11 @@ from typing import (
     Callable,
     ClassVar,
     Dict,
-    Iterator,
+    Iterable,
     Mapping,
     MutableMapping,
+    Optional,
+    Set,
     Union,
 )
 
@@ -72,28 +74,29 @@ class ConfigValueValidationError(ConfigError):
         )
 
 
-_NO_VALUE = object()
-
-
 class ConfigUnknownOptionError(ConfigError):
     """Raised when a configuration option is unknown."""
 
-    def __init__(self, name, value=_NO_VALUE):
-        super().__init__(
-            f"Config option {name} is unknown"
-            + ("" if value is _NO_VALUE else f" (with given value {repr(value)})")
+    def __init__(self, option_or_name):
+        name = (
+            f"'{option_or_name.name}' (id={id(option_or_name)})>"
+            if isinstance(option_or_name, Option)
+            else f"'{option_or_name}'"
         )
+        super().__init__(f"Config option {name} is unknown")
 
 
-@dataclass
+@dataclass(frozen=True)
 class Option:
+    name: str
+    """Unique name identifying the option (e.g. package.module:MY_OPTION)."""
     help: str
     """Help text for this option."""
     default: Any
     """Default value for this option."""
     parse: Callable[[str], Any]
     """Turn input (e.g. string) into proper type. Only when reading from file."""
-    validate: Callable[[Any], bool]
+    validate: Optional[Callable[[Any], bool]] = None
     """Return true if the given value is an acceptable value."""
 
 
@@ -106,12 +109,14 @@ class Options(Mapping):
     """
 
     __options: Dict[str, Option]
+    __cache: Set[Option]
 
     def __init__(self, other: "Options" = None) -> None:
         self.__options = {}
+        self.__cache = set()
         if other is not None:
-            for name, option in other.items():
-                self.register_option(name, option)
+            for option in other.values():
+                self.register_option(option)
 
     def register(
         self,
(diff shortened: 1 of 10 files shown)


Lib/fontTools/misc/configTools.py
```diff
@@ -72,28 +74,29 @@ def __init__(self, name, value):
...
+    name: str
+    """Unique name identifying the option (e.g. package.module:MY_OPTION)."""
     help: str
     """Help text for this option."""
     default: Any
     """Default value for this option."""
     parse: Callable[[str], Any]
     """Turn input (e.g. string) into proper type. Only when reading from file."""
-    validate: Callable[[Any], bool]
+    validate: Optional[Callable[[Any], bool]] = None
```
belluzj, line 99: You could also make the default value be lambda _: True?

Lib/fontTools/misc/configTools.py
```diff
@@ -106,45 +109,71 @@ class Options(Mapping):
...
-    def register_option(self, name: str, option: Option) -> Option:
+    def register_option(self, option: Option) -> Option:
         """Register a new option."""
+        name = option.name
         if name in self.__options:
             raise ConfigAlreadyRegisteredError(name)
+        # sanity check option values are unique
+        assert option not in self.__cache
         self.__options[name] = option
+        self.__cache.add(option)
```
belluzj, line 140: Why do you need a cache, since you already have a dict() keyed by option name? Do you also want to ensure that the same option name is not represented by 2 different Option objects?

Lib/fontTools/misc/configTools.py
```diff
@@ -106,45 +109,71 @@ class Options(Mapping):
...
         return option
 
-    def __getitem__(self, __k: str) -> Option:
-        return self.__options.__getitem__(__k)
+    def is_registered(self, option: Option) -> bool:
+        """Return True if the option object is already registered."""
+        return option in self.__cache
```
belluzj, line 145: you could look for the name in the __options

belluzj (commented):
some comments, to be completed later

Lib/fontTools/misc/configTools.py
anthrotype: maybe, but I find a None more explicit for something that is mostly optional

Lib/fontTools/misc/configTools.py
anthrotype: exactly, right

anthrotype: I want Option objects to be unique. There should only be one source of truth, the global options mapping. I made them frozen so one does't try to re register the same object twice under different names (can't change the name). I could even make the Option dataclass `eq=False` so it compares and hashes by id only (like `object`). I actually did that in an earlier iteration of this code, then undid that to not overcomplicate things.

anthrotype: the cache is not actually necessary for I could check `option in self.values()`, but I think a set should speed it up

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -73,16 +76,34 @@ def compile(self, font):
...
+					"uharfbuzz not found, compiling '%s' with pure-python serializer",
+					self.tableTag,
+				)
 
 		while True:
 			try:
 				writer = OTTableWriter(tableTag=self.tableTag)
 				self.table.compile(writer, font)
-				if have_uharfbuzz and self.tableTag in ("GSUB", "GPOS"):
+				if use_hb_repack and have_uharfbuzz and self.tableTag in ("GSUB", "GPOS"):
```
khaledhosny, line 96: Should it raise an error if `use_hb_repack` is True and `have_uharfbuzz` is False?

anthrotype: No, I think it should not. USE_HARFBUZZ_REPACKER is either True or False, and it's True by default. What it does is it always tries to use uharfbuzz when importable, and if the latter errors out, falls back to pure-python serializer. Setting the option to False simply means ignore whether uharfbuzz is importable or not, just use the pure-python serializer all the time.
If we where to raise when USE_HARFBUZZ_REPACKER is True and have_uharfbuzz is False, then we'd need a tri-state boolean, defaulting to None, to distinguish the case when we try to use hb if available (default), or we request to use it and error out if not available (True).

anthrotype: Or we would have to change the default to False, so it'd only work if one explicitly sets it to True

Lib/fontTools/misc/configTools.py
anthrotype: same name is no guarantee that two option objects are the same thing. One could make a new Option object with the same name as one already registered (but e.g. different default) and it would pass as if it were registered if we only checked its name.

Lib/fontTools/ttLib/tables/otBase.py
anthrotype: alright, i'll make it a 3-state variable

Lib/fontTools/misc/configTools.py
belluzj: That feels like adding weird semantics to what the "identity" of an option is. I would rather implement it as just an extra check like this:

```python
class Config:
    __options = {}

    def check_option_is_same(option: Option):
         assert self.__options[option.name] is option

     # and then call the check above anytime you're given an Option instance
     # the check is O(1)
     # no need for an extra set
```

Lib/fontTools/misc/configTools.py
belluzj: The check `self.__options[option.name] is option` is both fast and more accurately locking down what should happen with options. What do you think?

anthrotype: `self.__options.get(option.name) is option` -- i'll do that

belluzj (approved):

Lib/fontTools/ttLib/tables/otBase.py
anthrotype: Done

PR #2590, opened 2022-04-21
Adding in code to recalculate xAvgCharWidth at the end of the merge p…

### LINKED ISSUE #2538: Pyftmerge doesn't recalculate xAvgCharWidth
I'm running a script to merge a small set of glyphs (xAvgCharWidth 1024) into a Japanese font that contains a large number of 2048 width glyphs (with xAvgCharWidth of 1991).

The xAvgCharWidth output of pyftmerge is 1507, which is the average of the xAvgCharWidth for each of the fonts. However, the correct xAvgCharWidth should be 1965, which reflects the actual composition of the merged font. 

Fontbakery flags this as an issue. 

I'd suggest re-calculating the xAvgCharWidth value post merge to ensure correctness of the output. Potentially the code could be pulled directly from the [fontbakery check](https://github.com/googlefonts/fontbakery/blob/main/Lib/fontbakery/profiles/os2.py)?

aaronbell:
As merge and fontmake don't recalculate xAvgCharWidth after merging fonts, this code addition fixes that.
Code pulled from fontbakery.

## PR Checklist
* [x] Closes #2538

### DIFF
--- a/Lib/fontTools/merge/__init__.py
+++ b/Lib/fontTools/merge/__init__.py
@@ -125,6 +125,21 @@ class Merger(object):
 
 		self._postMerge(mega)
 
+		# merge only averages the xAvgCharWidth of the fonts being merged rather than recalculating. This code (from FontBakery) recalculates.
+
+		width_sum = 0
+		count = 0
+		for glyph_id in mega['glyf'].glyphs:  # At least .notdef must be present.
+			width = mega['hmtx'].metrics[glyph_id][0]
+			# The OpenType spec doesn't exclude negative widths, but only positive
+			# widths seems to be the assumption in the wild?
+			if width > 0:
+				count += 1
+				width_sum += width
+
+		avgCharWidth = int(round(width_sum / count))
+		mega["OS/2"].xAvgCharWidth = avgCharWidth
+
 		return mega
 
 	def mergeObjects(self, returnTable, logic, tables):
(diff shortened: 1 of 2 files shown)


Lib/fontTools/merge/__init__.py
```diff
@@ -125,6 +125,21 @@ def merge(self, fontfiles):
...
+		for glyph_id in mega['glyf'].glyphs:  # At least .notdef must be present.
+			width = mega['hmtx'].metrics[glyph_id][0]
+			# The OpenType spec doesn't exclude negative widths, but only positive
+			# widths seems to be the assumption in the wild?
+			if width > 0:
+				count += 1
+				width_sum += width
+
+		avgCharWidth = int(round(width_sum / count))
+		mega["OS/2"].xAvgCharWidth = avgCharWidth
```
anthrotype, line 141: i suggest to use the exact same computation as ufo2ft
https://github.com/googlefonts/ufo2ft/blob/ad28eea062e0dd48678309bd9ef86dfcc85fa85a/Lib/ufo2ft/outlineCompiler.py#L556-L561

anthrotype:
FWIW that value is mostly informational, arguably unused/useless for layout, and isn't even clear how to calculate it (meaning changed between spec revisions).. But I don't mind, if you do care about updating it after merge, I'll take the PR.

aaronbell:
> if you do care about updating it after merge
There's a fontbakery check that uses the above code to evaluate the correctness of the xAvgCharWidth data, so it would probably be best to have merge align with FB. 

As for the method of calculation (and your various helpful comments above), feel free to make any modifications to the PR as you'd like! I just submitted this as Dave indicated that he wanted it fixed and I had a workable solution to get the ball rolling.

anthrotype:
Thanks @aaronbell and @khaledhosny 
fixed with https://github.com/fonttools/fonttools/pull/2591

PR #2591, opened 2022-04-21
[OS/2 / merge] unify recalcXAvgCharWidth method and use it for merger, subsetter, etc.

### LINKED ISSUE #2538: Pyftmerge doesn't recalculate xAvgCharWidth

anthrotype:
Fixes #2538, supersedes #2590 

/cc @aaronbell

### DIFF
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -491,7 +491,7 @@ class FontBuilder(object):
             assert (
                 "hmtx" in self.font
             ), "the 'hmtx' table must be setup before the 'OS/2' table"
-            self.font["OS/2"].recalcXAvgCharWidth(self.font)
+            self.font["OS/2"].recalcAvgCharWidth(self.font)
         if not (
             "ulUnicodeRange1" in values
             or "ulUnicodeRange2" in values
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -486,15 +486,12 @@ class FontBuilder(object):
         """Create a new `OS/2` table and initialize it with default values,
         which can be overridden by keyword arguments.
         """
-        if "xAvgCharWidth" not in values:
-            gs = self.font.getGlyphSet()
-            widths = [
-                gs[glyphName].width
-                for glyphName in gs.keys()
-                if gs[glyphName].width > 0
-            ]
-            values["xAvgCharWidth"] = int(round(sum(widths) / float(len(widths))))
         self._initTableWithValues("OS/2", _OS2Defaults, values)
+        if "xAvgCharWidth" not in values:
+            assert (
+                "hmtx" in self.font
+            ), "the 'hmtx' table must be setup before the 'OS/2' table"
+            self.font["OS/2"].recalcXAvgCharWidth(self.font)
         if not (
             "ulUnicodeRange1" in values
             or "ulUnicodeRange2" in values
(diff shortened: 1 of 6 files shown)


Lib/fontTools/ttLib/tables/O_S_2f_2.py
```diff
@@ -299,6 +300,19 @@ def recalcUnicodeRanges(self, ttFont, pruneOnly=False):
 		self.setUnicodeRanges(bits)
 		return bits
 
+	def recalcXAvgCharWidth(self, ttFont):
```
khaledhosny, line 303: I’d drop the `X` from the function name.

Lib/fontTools/fontBuilder.py
```diff
@@ -486,15 +486,12 @@ def setupOS2(self, **values):
...
-                for glyphName in gs.keys()
-                if gs[glyphName].width > 0
-            ]
-            values["xAvgCharWidth"] = int(round(sum(widths) / float(len(widths))))
         self._initTableWithValues("OS/2", _OS2Defaults, values)
+        if "xAvgCharWidth" not in values:
+            assert (
+                "hmtx" in self.font
+            ), "the 'hmtx' table must be setup before the 'OS/2' table"
+            self.font["OS/2"].recalcXAvgCharWidth(self.font)
```
khaledhosny, line 494: This adds a dependency on `hmtx` table that this function didn’t have before, it might break some existing code but since the fix would probably only require calling `setupHorizontalMetrics()` first, I think it is OK.

Lib/fontTools/merge/tables.py
```diff
@@ -132,7 +132,7 @@ def mergeOs2FsType(lst):
 	'*': first,
 	'tableTag': equal,
 	'version': max,
-	'xAvgCharWidth': avg_int, # Apparently fontTools doesn't recalc this
+	'xAvgCharWidth': first, # Will be recalculated at the end on the merged font
```
khaledhosny, line 135: There is also `recalculate` but I’m not sure what it does, in case you want to check it out.

khaledhosny (approved):

Lib/fontTools/fontBuilder.py
anthrotype: it already depeneded implicitly on hmtx being present, because it was getting a _TTGlyphSet which requires it

Lib/fontTools/merge/tables.py
anthrotype: the thing is OS/2 doesn't automatically recalculate xAvgCharWidth, client (in this case merger) need to call method explicitly, so recalculate doesn't fit here I think.

Lib/fontTools/fontBuilder.py
khaledhosny: I see, thanks.

Lib/fontTools/merge/tables.py
khaledhosny: OK.

Lib/fontTools/ttLib/tables/O_S_2f_2.py
anthrotype: dropped

Issue #2594, opened 2022-04-22
lots of logging ERROR when hb.repack fails


anthrotype:
As of v4.33 we try the use harfbuzz repacker when uharfbuzz can be imported, and fall back to our pure-python serializer when that fails. This can happen with relatively large fonts containing very big tables that need to be split up and hb.repack currently doesn't support that.
FontTools will emit a ([not very informative](https://github.com/harfbuzz/harfbuzz/issues/3546)) logging ERROR message in this case.

Now because of the way this was integrated into the existing overflow resolution loop, we keep retrying calling the HB repacker multiple times even as the first iteration failed.

The code is wrapped in a `while True:` loop where `OTLOffsetOverflowError` exception is caught and resolution happens; but this line here, even if has a `return` which normally would break from the loop, can still raise `OTLOffsetOverflowError` so the loop continues and hb repacker is tried again after the offset-resolution fixing attempt:

https://github.com/fonttools/fonttools/blob/f68ce14d7af48bc940b4ce220401bda6021ecd55/Lib/fontTools/ttLib/tables/otBase.py#L122

I think all these logging ERRORs are too noisy and might induce the user to think something is wrong whereas this is just the way things are currently.

So I propose we either demote the message to a DEBUG logging level, and/or we just try hb repacker one time only, and if that fails we won't try again at the next iteration.

This is what one would see (the font below is a proprietary one I can't share but you get the idea):

```
ERROR:fontTools.ttLib.tables.otBase:hb.repack failed to serialize 'GPOS', reverting to pure-python serializer; the error message was: RepackerError
INFO:fontTools.ttLib.tables.otBase:Attempting to fix OTLOffsetOverflowError ('GPOS', 'LookupIndex:', 2, 'SubTableIndex:', None, 'ItemName:', None, 'ItemIndex:', None)
ERROR:fontTools.ttLib.tables.otBase:hb.repack failed to serialize 'GPOS', reverting to pure-python serializer; the error message was: RepackerError
INFO:fontTools.ttLib.tables.otBase:Attempting to fix OTLOffsetOverflowError ('GPOS', 'LookupIndex:', 1, 'SubTableIndex:', 1, 'ItemName:', 'Coverage', 'ItemIndex:', None)
ERROR:fontTools.ttLib.tables.otBase:hb.repack failed to serialize 'GPOS', reverting to pure-python serializer; the error message was: RepackerError
[...]

PR #2595, opened 2022-04-22
[otBase] demote repacker ERROR to WARNING, only 1 per loop; don't exit at first failure

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/2594

* when hb repacker fails, we now emit a WARNING logging message, no longer an ERROR
* we only emit the warning message once, not multiple times at each iteration of the offset overflow resolution loop; it's always going to be the same anyway and it doesn't currently add any additional information besides the fact that it failed
* we no longer let the uharfbuzz error propagate (and thus stop execution if unhandled by client) even when the USE_HARFBUZZ_REPACKER option was explicitly set to True by the user, but we continue with the hybrid approach of trying hb first and if that fails attempt to fix offset overflow with pure-python serializer then re-trying until it completes. This is the normal state of things until hb repacker implements support for table splitting.

### DIFF
--- a/Lib/fontTools/ttLib/tables/otBase.py
+++ b/Lib/fontTools/ttLib/tables/otBase.py
@@ -96,6 +96,7 @@ class BaseTTXConverter(DefaultTable):
 						self.tableTag,
 					)
 
+		hb_first_error_logged = False
 		while True:
 			try:
 				writer = OTTableWriter(tableTag=self.tableTag)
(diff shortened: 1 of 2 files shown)


anthrotype:
basically, now the only difference between not setting the `USE_HARFBUZZ_REPACKER` option (default) and explicitly setting it to `True` is the fact that in the first case it will silently fall back to the default pure-python serializer in absence of uharfbuzz; whereas in the latter case, it will blow up with `ImportError` if uharfbuzz is not available in the python environment.

PR #2597, opened 2022-04-23
correct naming of dataclasses '__post_init__' method

m10d:
just happened to see this in a lint check for our own code


anthrotype:
thanks for spotting this

Issue #2602, opened 2022-04-27
'prune_post_subset' does not work for when 'self.Value' is an list of None's


Ristellise:
Hi! As mentioned in the title, when a `self.Value` is a list of `None`'s, the check for `if self.Value is None` returns `false`.

```py
@_add_method(otTables.SinglePos)
def prune_post_subset(self, font, options):
	if self.Value is None: # <-------------- Here.
		assert self.ValueFormat == 0
		return True

	# Shrink ValueFormat
	if self.Format == 1:
		if not options.hinting:
			self.Value.prune_hints()
		self.ValueFormat = self.Value.getEffectiveFormat()
	elif self.Format == 2:
		if not options.hinting:
			for v in self.Value:
				v.prune_hints()
		print(self.Value)
		self.ValueFormat = reduce(int.__or__, [v.getEffectiveFormat() for v in self.Value], 0)

	# Downgrade to Format 1 if all ValueRecords are the same
	if self.Format == 2 and all(v == self.Value[0] for v in self.Value):
		self.Format = 1
		self.Value = self.Value[0] if self.ValueFormat != 0 else None
		del self.ValueCount

	return True
```

The solution would be to replace the line with:
```py
	if self.Value is None or all(v is None for v in self.Value):
```
however, I'm not too sure what other parts will be affected.

The following catbox link is a font that exhibits this issue. I haven't extracted the ttx out of it yet.  
https://files.catbox.moe/audwuu.otf

to generate the ttx file, I used the following command:
```
pyftsubset audwuu.otf --output-file=encode --unicodes=0000-00FF,0370-04FF,2000-206F,20A0-20CF,2122-2122,3000-303F,203D-2044,2018-201F
```

anthrotype:
SinglePos `Value` attribute can be either a single `ValueRecord` object (format 1) or a list of `ValueRecord`s (format 2), or `None` if ValueFormat == 0. I don't think it can ever be a list of `None` values..

anthrotype:
hm, I think I managed to reproduce with the font above and command you provided. I'll take a look, thanks

```
$ fonttools subset --output-file=/tmp/audwuu.subset.otf --unicodes="0000-00FF,0370-04FF,2000-206F,20A0-20CF,2122-2122,3000-303F,203D-2044,2018-201F" /Users/clupo/Downloads/audwuu.otf
Traceback (most recent call last):
  File "/Users/clupo/Github/fonttools/.venv/bin/fonttools", line 33, in <module>
[...]

PR #2603, opened 2022-04-27
[subset] Fix subsetting SinglePosFormat2 with ValueFormat=0

anthrotype:
I first push a failing test to reproduce #2602, then follow up with a fix


anthrotype:
```
>   self.ValueFormat = reduce(int.__or__, [v.getEffectiveFormat() for v in self.Value], 0)
E   AttributeError: 'NoneType' object has no attribute 'getEffectiveFormat'
```

https://github.com/fonttools/fonttools/runs/6195927241?check_suite_focus=true#step:5:421

PR #2604, opened 2022-04-28
Add --fvar option to instancer

simoncozens:
I was surprised that I couldn't find an easy way to create static instances from fvar named instances, so I added one.

### DIFF
--- a/Lib/fontTools/varLib/instancer/__init__.py
+++ b/Lib/fontTools/varLib/instancer/__init__.py
@@ -1380,6 +1380,11 @@ def parseArgs(args):
         help="Update the instantiated font's `name` table. Input font must have "
         "a STAT table with Axis Value Tables",
     )
+    parser.add_argument(
+        "--fvar",
+        action="store_true",
+        help="Create static instances for each instance in the variable font's fvar table",
+    )
     loggingGroup = parser.add_mutually_exclusive_group(required=False)
     loggingGroup.add_argument(
         "-v", "--verbose", action="store_true", help="Run more verbosely."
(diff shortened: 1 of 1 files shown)


anthrotype:
I like Khaled's idea of an --named-instaces option that takes an optional regular expression that matches fvar named instances style name, if option argument is omitted defaults to `.*` to build the all

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -1413,16 +1421,41 @@ def parseArgs(args):
     if len(axisLimits) != len(options.locargs):
         parser.error("Specified multiple limits for the same axis")
 
+    if axisLimits and options.named_instances:
+        parser.error("Specified limits and --fvar")
```
anthrotype, line 1425:
```suggestion
        parser.error("Specified limits and --named-instances")
```

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -1413,16 +1421,41 @@ def parseArgs(args):
...
     varfont = TTFont(infile)
+    if options.named_instances:
+        for instance in varfont["fvar"].instances:
+            name = varfont["name"].getDebugName(instance.subfamilyNameID)
+            if not re.search(options.named_instances, name):
+                continue
+            axisLimits = instance.coordinates
+            log.info("Restricting axes: %s (%s)", axisLimits, name)
+            name = name.replace(" ", "")
+            outfile = os.path.splitext(infile)[0] + "-" + name + ".ttf"
```
anthrotype, line 1444: the -o/--output option is also incompatible with this new --named-instances, because the latter usually outputs multiple files, not a single one

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -1413,16 +1421,41 @@ def parseArgs(args):
...
 
 def main(args=None):
     """Partially instantiate a variable font."""
     infile, axisLimits, options = parseArgs(args)
-    log.info("Restricting axes: %s", axisLimits)
 
     log.info("Loading variable font")
     varfont = TTFont(infile)
+    if options.named_instances:
+        for instance in varfont["fvar"].instances:
```
anthrotype, line 1437: maybe move this whole for loop to its own function to `instantiateVariableFontNamedInstances`

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -1413,16 +1421,41 @@ def parseArgs(args):
...
+            name = varfont["name"].getDebugName(instance.subfamilyNameID)
+            if not re.search(options.named_instances, name):
+                continue
+            axisLimits = instance.coordinates
+            log.info("Restricting axes: %s (%s)", axisLimits, name)
```
anthrotype, line 1442: in this context, we use "restricting an axis" to mean level-3 instancing (i.e. modify min/max), as opposed to `pin an axis` for level-1 (pin at default) and 2 (pin at arbitrary location). If you still wanna log, keep it generic, like "Instancing axes: %s"

anthrotype:
a little test would also be nice

Lib/fontTools/varLib/instancer/__init__.py
khaledhosny: If `--output` is used this code should error, or, if it wants to be smart, use it for the first matching instance and error for the rest, so one can use it to output a single named instance (which happens to be my use case, since I use Makefile and rules that output multiple files is PITA).

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -1443,7 +1448,7 @@ def main(args=None):
             axisLimits = instance.coordinates
             log.info("Restricting axes: %s (%s)", axisLimits, name)
             name = name.replace(" ", "")
-            outfile = os.path.splitext(infile)[0] + "-" + name + ".ttf"
+            outfile = outprefix + "-" + name + ".ttf"
```
anthrotype, line 1451: this is not exactly what Khaled and I had in mind though, namely

> If --output is used this code should error, or, if it wants to be smart, use it for the first matching instance and error for the rest, so one can use it to output a single named instance

anthrotype: if you want to control where multiple outputs go, i think it'd be better to add a new explicit --output-dir option

anthrotype:
@simoncozens are you going to address the review comment above? if you don't have time to finish this it's ok, just let us know here so someone else can pick this up, thanks

simoncozens:
Sorry, yeah, I don't think I'm going to be picking this up any time soon. The project I needed it for is done now.

PR #2607, opened 2022-04-28
OS/2 ranges update olli

moontypespace:
I added:
getCodePageRange (equivalent to getUnicodeRange)
setCodePageRange (equivalent to setUnicodeRange)
recalcCodePageRange (equivalent to recalcUnicodeRanges)


moontypespace:
@justvanrossum What do you think about it?

Lib/fontTools/ttLib/tables/O_S_2f_2.py
```diff
@@ -300,6 +323,37 @@ def recalcUnicodeRanges(self, ttFont, pruneOnly=False):
...
+		bits = set()
+		for bit in OS2_CODEPAGE_RANGES:
+			codepage_name = OS2_CODEPAGE_RANGES[bit]
+			if codepage_name is None:
+				continue
+			codepage_encoding = CODEPAGERANGES_ENCODINGS[codepage_name]
+			if drop:
+				for enc in drop:
+					codepage_encoding.discard(enc)
+			if set() == codepage_encoding.difference(unicodes):
```
anthrotype, line 352: `if unicodes.issuperset(codepage_encoding)`

anthrotype:
hey! thanks for the contribution. The reason we never added something like this in the past -- besides codepages being windows legacy old cruft that should have no place in a unicode world -- is that there wasn't agreement on how to decide when a font is deemed to cover a given codepage. E.g. should it be only one, the majority, 95%, _all_ of the codepoints in the codepage, etc.. I think the spec leaves that to the font developer.
In your PR I see you opted for all codepoints (with the somewhat arbitrary exception of 0x000).

Lib/fontTools/unicodedata/codepages.py
```diff
@@ -0,0 +1,146 @@
+# -*- coding: utf-8 -*-
+# codepage ranges mainly from:
+# https://unicode.org/Public/MAPPINGS/VENDORS/MICSFT/ 
+# https://unicode.org/Public/MAPPINGS/VENDORS/MICSFT/WINDOWS/ 
+# https://unicode.org/Public/MAPPINGS/VENDORS/MICSFT/PC/
```
anthrotype, line 5: this codepages.py module could potentially be autogenerated, see similar scripts in MetaTools

moontypespace: @anthrotype Agree, but sadly not all. There is one codepage 708 I was not able to find any data I would be able to parse via script: https://github.com/moontypespace/fonttools/blob/caff9562e8bdb84e02b00c0a9d42c46c555ad292/Lib/fontTools/unicodedata/codepages.py#L39
The other character sets I have parsed via some line of code.

Lib/fontTools/ttLib/tables/O_S_2f_2.py
moontypespace: I have changed it, thanks :)

moontypespace:
> hey! thanks for the contribution. The reason we never added something like this in the past -- besides codepages being windows legacy old cruft that should have no place in a unicode world -- is that there wasn't agreement on how to decide when a font is deemed to cover a given codepage. E.g. should it be only one, the majority, 95%, _all_ of the codepoints in the codepage, etc.. I think the spec leaves that to the font developer. In your PR I see you opted for all codepoints (with the somewhat arbitrary exception of 0x000).

Jepp, you're right. The OpenType Spec is a little bit fuzzy. But these are specified character sets, not unicode ranges, why I thought it makes most sense to set the bit when the coverage is 100% (except 0x0000). But it's possible to add more unicodes to 'drop', if someone has a different opinion.

moontypespace:
@anthrotype Is there anything else you want me to change?

moontypespace:
@justvanrossum @anthrotype Excuse me, I really don't want o bother you with this unsexy topic, but is there anything else I can do to merge it into the master branch?

Lib/fontTools/unicodedata/codepages.py
anthrotype: how about we hard-code the odd codepage 708 only, and have a MetaTools script that generate the others that can be generted?

moontypespace:
Hi @anthrotype and @justvanrossum , I figured out that the fonttools subsetter takes care of the unicoderanges but not of the codepageranges. Wouldn't it be great if it would take care of it as well automatically?

anthrotype:
FWIW, my earlier comments still stand.

khaledhosny:
If no one is working on this, I can try to take over and finish it.

khaledhosny:
I was pretty sure I implemented this previously, but I couldn’t find it. Turned out it was in ufo2ft, so I’m going to move that to FontTools instead as it is much simpler.

https://github.com/googlefonts/ufo2ft/pull/254/commits/beca9c35e1dfb441458567fa0e0d9ac430b83398

khaledhosny:
https://github.com/fonttools/fonttools/pull/3328

PR #2609, opened 2022-04-29
[designspaceLib] Add some polish to the v5 paths

madig:
Found while working on https://github.com/googlefonts/ufo2ft/pull/598.

### DIFF
--- a/Lib/fontTools/designspaceLib/statNames.py
+++ b/Lib/fontTools/designspaceLib/statNames.py
@@ -113,12 +113,8 @@ def getStatNames(
 
     styleMapFamilyNames = {}
     for language in set(familyNames).union(styleNames.keys()):
-        try:
-            familyName = familyNames.get(language, familyNames["en"])
-            styleName = styleNamesForStyleMap.get(language, styleNamesForStyleMap["en"])
-        except KeyError:
-            LOGGER.warning("Cannot complete localized style map name because there are no English language fallbacks.")
-            continue 
+        familyName = familyNames.get(language, familyNames["en"])
+        styleName = styleNamesForStyleMap.get(language, styleNamesForStyleMap["en"])
         styleMapFamilyNames[language] = (familyName + " " + styleName).strip()
 
     return StatNames(
--- a/Lib/fontTools/designspaceLib/statNames.py
+++ b/Lib/fontTools/designspaceLib/statNames.py
@@ -113,8 +113,12 @@ def getStatNames(
 
     styleMapFamilyNames = {}
     for language in set(familyNames).union(styleNames.keys()):
-        familyName = familyNames.get(language, familyNames["en"])
-        styleName = styleNamesForStyleMap.get(language, styleNamesForStyleMap["en"])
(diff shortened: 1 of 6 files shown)


Lib/fontTools/designspaceLib/types.py
```diff
@@ -118,5 +133,9 @@ def getVFUserRegion(doc: DesignSpaceDocument, vf: VariableFontDescriptor) -> Reg
     # Any axis not mentioned explicitly has a single location = default value
     for axis in doc.axes:
         if axis.name not in vfUserRegion:
+            if not isinstance(axis.default, (int, float)):
```
anthrotype, line 136: how can that be? default should not be optional for an axis

Tests/designspaceLib/designspace_v5_test.py
```diff
@@ -185,6 +185,22 @@ def test_read_v5_document_simple(datadir):
                 styleName="Supports",
                 localisedFamilyName={},
             ),
+            SourceDescriptor(
+                filename="masters/masterTest2.ufo",
+                path=posix(str((datadir / "masters/masterTest2.ufo").resolve())),
+                name="master.ufo3",
+                layerName=None,
+                location={"Italic": 1.0, "Weight": 0.0, "Width": 100.0},
+                copyLib=False,
```
anthrotype, line 194: why do we keep initializing these deprecated attributes?

Lib/fontTools/designspaceLib/statNames.py
```diff
@@ -113,8 +113,12 @@ def getStatNames(
 
     styleMapFamilyNames = {}
     for language in set(familyNames).union(styleNames.keys()):
-        familyName = familyNames.get(language, familyNames["en"])
-        styleName = styleNamesForStyleMap.get(language, styleNamesForStyleMap["en"])
+        try:
+            familyName = familyNames.get(language, familyNames["en"])
+            styleName = styleNamesForStyleMap.get(language, styleNamesForStyleMap["en"])
+        except KeyError:
+            LOGGER.warning("Cannot complete localized style map name because there are no English language fallbacks.")
```
anthrotype, line 120: how likely is the situation that triggers this logging warning? if yes, it may deserve a test

Lib/fontTools/designspaceLib/types.py
madig: The type checker says it's possible.

anthrotype: well, then just assert it isn't the case

Tests/designspaceLib/designspace_v5_test.py
madig: Because the `True` ones are needed for equality testing.

Lib/fontTools/designspaceLib/statNames.py
madig: I think it happens if there's no default source _and_ you didn't set `en` names for something. I'll just remove it for now.

PR #2611, opened 2022-05-03
Handle one more valid real number format in svgLib parse_path()

tshinnic:
Using svg.draw(pen) and parse_path() from svgLib threw exception

      ValueError: could not convert string to float: 'a'

on a SVG <path> string generated by Inkscape. Altering the path
string might object to other string bits like 'Z' or 'v', or
give even stranger exceptions.

Finally tracked it down to the path containing particular valid
numbers like "-4e-5" or "1e-4". Changing these to "-4.0e-5" or
"1.0e-4" would stop the exceptions.  parse_path() was not
accepting valid SVG real numbers.

The specification for real number formats is a bit of a mess in
CSS land right now, but the reassuringly concrete spec is:
    https://www.w3.org/TR/css-syntax-3/#number-token-diagram
which allows for a real number having an exponent but without having
a fractional part, such as the number "1e3".

This change updates an RE to make fractional parts optional,
and adds a test for this valid SVG number format.

### DIFF
--- a/Lib/fontTools/svgLib/path/parser.py
+++ b/Lib/fontTools/svgLib/path/parser.py
@@ -16,10 +16,13 @@ ARC_COMMANDS = set("Aa")
 UPPERCASE = set('MZLHVCSQTA')
 
 COMMAND_RE = re.compile("([MmZzLlHhVvCcSsQqTtAa])")
+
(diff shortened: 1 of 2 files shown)


anthrotype (approved):
Thanks!

Issue #2579, opened 2022-04-12
How to get the proper Font Family name ?


moi15moi:
I have this font: https://mega.nz/file/S9ERDRpQ#bcPhS06kv-D5jt64aTNDbZVd6gZr6ZfJDYT91yYsoWk

From what I can see here, the family name is the nameID=1: https://docs.microsoft.com/en-us/typography/opentype/spec/name
This is why I always do `font['name'].names[1].XXX`

If I try to get the encoding, I get "mac_roman" when I should get "shift-jis"
Here is what I did `font['name'].names[1].getEncoding()`

If I print the names of the font, I have this (i think the problem is from this):
[<NameRecord NameID=0; PlatformID=1; LanguageID=11>, <NameRecord NameID=1; PlatformID=1; LanguageID=11>, <NameRecord NameID=2; PlatformID=1; LanguageID=11>, <NameRecord NameID=3; PlatformID=1; LanguageID=11>, <NameRecord NameID=4; PlatformID=1; LanguageID=11>, <NameRecord NameID=5; PlatformID=1; LanguageID=11>, <NameRecord NameID=6; PlatformID=1; LanguageID=11>, <NameRecord NameID=0; PlatformID=3; LanguageID=1041>, <NameRecord NameID=1; PlatformID=3; LanguageID=1041>, <NameRecord NameID=2; PlatformID=3; LanguageID=1041>, <NameRecord NameID=3; PlatformID=3; LanguageID=1041>, <NameRecord NameID=4; PlatformID=3; LanguageID=1041>, <NameRecord NameID=5; PlatformID=3; LanguageID=1041>, <NameRecord NameID=6; PlatformID=3; LanguageID=1041>]

I think the PlatformID=1 return "mac_roman" but the PlatformID=3 would return "shift-jis".
https://docs.microsoft.com/en-us/typography/opentype/spec/name#platform-encoding-and-language-ids

If I do, font['name'].names[1].toBytes().decode('shift-jis'), it return me the good font name.

In brief, what do I do wrong or is there a bug with the method getEncoding (https://fonttools.readthedocs.io/en/latest/_modules/fontTools/ttLib/tables/_n_a_m_e.html#NameRecord.getEncoding) ?

moi15moi:
Update: I edited my previous message and give better explanation.

moi15moi:
I just realize that one of the record fail to decompile (it is the second one)

```xml
[...]

Issue #236, opened 2015-04-16
[SPEC] Legacy Mac name table entries with PID=1


twardoch:
For the lack for a better place at this point, I'm putting this useful piece of information here as it relates potentially to https://github.com/behdad/fonttools/pull/235 . This info should actually be part of the OpenType or OFF spec, but isn’t. 

# Legacy Mac `name` table entries with `PID=1` #

As a rule, the `name` table strings are encoded using Unicode (UTF-16BE). However, the legacy Mac `name` table strings (with `PID=1`) use legacy “Script Manager” Mac encodings, not Unicode. 

To read them in correctly and write them out correctly into real Unicode, one would either need external libraries such as ICU encodings or Python codecs. ICU brings a set of converters, and the ICU “windows-100x” converters are in fact converters for legacy Mac encodings. Python also brings a set of codecs. But this is not *entirely helpful* because neither ICU nor Python actually bring codecs for *all* of the old Mac encodings that can potentially occur in fonts. Of course it would be stupid to add ICU as a dependency to fontTools, so, while I’m including these tables for people’s reference, I’m not actually advocating that we should convert to and from Unicode for the `PID=1` entries. 

**Instead**, the non-Unicode portions should be encoded using some *escapement*, and ideally somehow denoted as such. This might be *custom escapement* such as `\xNN` rather than `&#NN`; entities, because we claim that contents of the XML is UTF-8. This is crucial: we should *really* make it clear in tools such as fontTools and in files such as .ttx that a certain string *is Unicode* or that it is *not Unicode*. That aspect of `name` table is not really well solved at all in fontTools/TTX, and I cannot see how any simple patches that we do will make it really good. 

*Note: In the following tables, × means that an encoding is not supported by Python or ICU. and * denotes “emulated” solutions that don’t guarantee full compatibility (see last section).*

## Latin-script `EID=0` entries ##

For `EID=0`, the conversions are also dependent on the `LID`. The following is the mapping from `EID` and `LID` combinations to Python codecs (from the `codecs` module), ICUEncodingIDs, CoreFoundation LIDNames, MIMEnames and WindowsCodepageIDs. The hyperlinks connect to the mapping tables available at ICU and, in the MIMEName column, at unicode.org. 

EID | LID | Python | ICUEncID | LIDName | MIMEName | WinCPID | 
--- | --- | ------ | -------- | ------- | -------- | ------- | 
[...]

PR #2618, opened 2022-05-12
Decode correctly PlatformID 3 name

moi15moi:
### What does this PR does:
This PR use the same logic has DirectWrite to obtain the character encoding for an NameRecord.
It only concern the NameRecord from the Microsoft Platform.

The logic is from this comment: https://github.com/MicrosoftDocs/typography-issues/issues/956.

### Related issue:
- https://github.com/fonttools/fonttools/issues/2579
- https://github.com/fonttools/fonttools/issues/236

With this PR, here is the xml I get with the font in the issue #2579
```xml
  <name>
    <namerecord nameID="0" platformID="1" platEncID="0" langID="0xb" unicode="True">
      Copyright Fujitsu 1998
    </namerecord>
    <namerecord nameID="1" platformID="1" platEncID="0" langID="0xb" unicode="True">
      FjÉCÅ[É}310
    </namerecord>
    <namerecord nameID="2" platformID="1" platEncID="0" langID="0xb" unicode="True">
      Regular
    </namerecord>
    <namerecord nameID="3" platformID="1" platEncID="0" langID="0xb" unicode="True">
      Fujitsu:FjÉCÅ[É}310:1998
    </namerecord>
    <namerecord nameID="4" platformID="1" platEncID="0" langID="0xb" unicode="True">
      FjÉCÅ[É}310
    </namerecord>
    <namerecord nameID="5" platformID="1" platEncID="0" langID="0xb" unicode="True">
      Version 001.20
    </namerecord>
    <namerecord nameID="6" platformID="1" platEncID="0" langID="0xb" unicode="True">
      Fj-Ima310
    </namerecord>
    <namerecord nameID="0" platformID="3" platEncID="2" langID="0x411">
      Copyright Fujitsu 1998
    </namerecord>
    <namerecord nameID="1" platformID="3" platEncID="2" langID="0x411">
      Fjイーマ310
    </namerecord>
    <namerecord nameID="2" platformID="3" platEncID="2" langID="0x411">
      Regular
    </namerecord>
    <namerecord nameID="3" platformID="3" platEncID="2" langID="0x411">
      Fujitsu:Fjイーマ310:1998
    </namerecord>
    <namerecord nameID="4" platformID="3" platEncID="2" langID="0x411">
      Fjイーマ310
    </namerecord>
    <namerecord nameID="5" platformID="3" platEncID="2" langID="0x411">
      Version 001.20
    </namerecord>
    <namerecord nameID="6" platformID="3" platEncID="2" langID="0x411">
      Fj-Ima310
    </namerecord>
  </name>
```

### DIFF
--- a/Lib/fontTools/cu2qu/cu2qu.py
+++ b/Lib/fontTools/cu2qu/cu2qu.py
@@ -240,9 +240,9 @@ def split_cubic_into_three(p0, p1, p2, p3):
     mid2 = (p0 + 6 * p1 + 12 * p2 + 8 * p3) * (1 / 27)
     deriv2 = (4 * p3 - 3 * p1 - p0) * (1 / 27)
     return (
-        (p0, (2 * p0 + p1) / 3.0, mid1 - deriv1, mid1),
+        (p0, _complex_div_by_real(2 * p0 + p1, 3.0), mid1 - deriv1, mid1),
         (mid1, mid1 + deriv1, mid2 - deriv2, mid2),
-        (mid2, mid2 + deriv2, (p2 + 2 * p3) / 3.0, p3),
+        (mid2, mid2 + deriv2, _complex_div_by_real(p2 + 2 * p3, 3.0), p3),
     )
 
 
--- a/Tests/cu2qu/cu2qu_test.py
+++ b/Tests/cu2qu/cu2qu_test.py
@@ -254,6 +254,28 @@ def test_cu2qu_complex_division_floating_point_precision():
     assert result[1][5][1] == 326.49999999999994
 
 
+def test_cu2qu_split_cubic_into_three_complex_division():
+    # Same class of bug as test_cu2qu_complex_division_floating_point_precision above,
+    # but in split_cubic_into_three which has its own / 3.0 expressions not covered by
+    # the calc_cubic_points fix in PR #3930. These curves from JosefinSans-Italic
+    # (ordmasculine glyph, three masters) trigger n=3 splitting where the mathematical
+    # result is exactly 504.5. Cython's C complex division produces 504.49999999999994
+    # (just below .5, rounds down to 504) while pure Python gives 504.5000000000001
+    # (just above .5, rounds up to 505) — a ±1 coordinate difference in the final font.
+    # Pure Python already produces the correct result; this only fails under Cython.
+    # See: https://github.com/fonttools/fonttools/issues/3928
+
+    curves = [
+        [(217.833, 408.833), (299.667, 403.833), (372.333, 463.833), (382.5, 545.167)],
+        [(217, 413), (303, 408), (374, 468), (385, 551)],
+        [(213, 426), (313, 420), (381, 482), (392, 571)],
+    ]
+
+    result = curves_to_quadratic(curves, max_errors=[1.0] * len(curves))
+
+    assert result[0][3][1] == 504.5000000000001
+
+
 def test_curves_to_quadratic_rejects_mismatched_max_errors():
     curves = [[(0, 0), (0, 1), (2, 1), (2, 0)]]
     with pytest.raises(ValueError, match="max_errors must match"):


Lib/fontTools/misc/encodingTools.py
```diff
@@ -18,6 +18,7 @@
 		# See
 		# https://github.com/fonttools/fonttools/issues/236
 		0: { # Macintosh, platEncID==0, keyed by langID
```
moi15moi, line 20: Since it is write that, I am wondering if we should add all the langID from: https://docs.microsoft.com/en-us/typography/opentype/spec/name#macintosh-language-ids

The problem is that I don't know the encoding for each of them

moi15moi:
@behdad What do you think about that?

moi15moi:
_If I look at the font you posted in https://github.com/fonttools/fonttools/issues/2579, there are name table strings that are correctly identified as "shift_jis", but the actual data fails to be decompiled as such (this may be a problem with the font?)._

I have still no idea why it fails to decompiled the name.

I just saw that [fontconfig](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/main/src/fcfreetype.c#L733) can consider the language_id (11) to get the good encoding, so I reproduce something similar.

moi15moi:
I updated the code.
I was wrong on the method. Now, I relied on the microsoft documentation. See my comment about that.

I still don't have any proof that we should use _x_mac_japanese_ttx_ when the plaformID = 1, platEncID = 0 and langID = 11, but since it work, I highly doubt that this is not the right method.

moi15moi:
I installed a mac VM on my windows computer and I remove all the platformID=3 name from the font to be 100% sure that MACOS will show me the name this namerecord:
```xml
<namerecord nameID="1" platformID="1" platEncID="0" langID="0xb" unicode="True">
  FjÉCÅ[É}310
</namerecord>
```

I opened the font with [Font Book](https://support.apple.com/en-ca/guide/font-book/welcome/mac) (the default mac os font manager)
image

We can clearly see that it won't decode the font with x_mac_japanese_ttx.

Also, it is not wrote that it use kTextEncodingMacJapanese when the platEncID=0: https://developer.apple.com/documentation/coreservices/1399980-afp_text_encodings

I will remove the code I added

moi15moi:
@justvanrossum What do you think about the new change in the PR?

justvanrossum:
If "[All string data for platform 3 must be encoded in UTF-16BE](https://docs.microsoft.com/en-us/typography/opentype/spec/name#platform-specific-encoding-and-language-ids-windows-platform-platform-id-3)", then perhaps this section in `encodingTools.py` is wrong?

https://github.com/fonttools/fonttools/blob/21d22ae30f88b76fab5575f5f2e882b142ee1c08/Lib/fontTools/misc/encodingTools.py#L51-L60

But then again, the spec seems to contradict itself: for encodingID 4 it says "Big5", yet it should be encoded as UTF-16BE? I can't make sense of that. @PeterConstable, can you help out here what might be meant?

moi15moi:
I am not 100% sure. I also think that the documentation lacks a bit of precision.

Here are the possibilities:
1. Everything from the platform 3 should always be decoded with utf_16_be (which won't make any sense because why would there be encoding_id except for the cmap).
2. If the decode fail and the name is from the platform 3, try to decode it with utf_16_be. (In my opinion, it would make more sense. This spec is probably only for pretty old font.)

PeterConstable:
> If "[All string data for platform 3 must be encoded in UTF-16BE](https://docs.microsoft.com/en-us/typography/opentype/spec/name#platform-specific-encoding-and-language-ids-windows-platform-platform-id-3)"... 

That's a clear statement in the spec and means what it says.

> But then again, the spec seems to contradict itself: for encodingID 4 it says "Big5", yet it should be encoded as UTF-16BE? I can't make sense of that. @PeterConstable, can you help out here what might be meant?

The spec isn't contradicting itself, though perhaps there's a detail that isn't clear. At least, this is how I understand what historically has been intended:

The 'cmap' table can have subtables for different encodings for the same platform ID. For example,  the 'cmap' could have both 3/1 and 3/4 subtables. Whatever 'cmap' subtables for platform 3 are present, there should be name records in the 'name' table for the same set of platform 3 encodings. So, if the 'cmap' has 3/1 and 3/4 subtables, it should also have name records for both 3/1 and 3/4.

But whatever platform 3 encoding ID is specified in name records, the string data it references must be encoded in UTF-16BE.

The spec does not that some legacy CJK fonts were non-conformant in that they had 3/4 name records referencing string data encoded in code page 950. But it's explicitly stating that that is non-conformant.

If any aspect of that isn't clear enough in the spec, please open an issue against the relevant page that needs clarification.

justvanrossum:
Thank you Peter, that clears things up.

I do find it hard to understand *why* this is the case:

> So, if the 'cmap' has 3/1 and 3/4 subtables, it should also have name records for both 3/1 and 3/4.

Why would you need 3/1 and 3/4 subtables, if they're both going to to be UTF-16BE?

To move forward, it seems indeed all values in this dict should be `'utf_16_be'`:

https://github.com/fonttools/fonttools/blob/21d22ae30f88b76fab5575f5f2e882b142ee1c08/Lib/fontTools/misc/encodingTools.py#L51-L60

So that's what this PR should do.

moi15moi:
> To move forward, it seems indeed all values in this dict should be 'utf_16_be'

It should have an ellepsis for utf_16_be. And if the decode fail, it would use cp950.

justvanrossum:
The Ellipsis thing is for language code.

I'm not sure if falling back to anything is a wise thing here. We currently don't use cp950, and I don't see why we should do so now.

PeterConstable:
>Why would you need 3/1 and 3/4 subtables, if they're both going to be UTF-16BE?

Well, in 1990, that's what people thought should be done. In 2022, why does any font need a 3/4 'cmap' subtable?

Of course, some old fonts can have a 3/4 'cmap' subtable. If the 'name' table doesn't have corresponding 3/4 name records, I think the font should still be usable, though there may be some implementations that reject the font. 

If there are 3/4 name records (or 3/2, etc.), I'd try interpreting strings using UTF-16BE, but if any of the strings come out as gibberish or invalid (unpaired surrogate code points) then I might try interpreting that string using Big5 (or Shift-JIS for 3/2, etc.).

moi15moi:
To be sure what we exactly need, I want to have the "famous" _MingLi font included in the Traditional Chinese edition of Windows 95_ to test correctly.

I can't find it online, so I ask if it was possible to have it: https://github.com/MicrosoftDocs/typography-issues/issues/956

PeterCon:
Addendum... It turns out that what I stated earlier wasn't quite correct. What I described was what was relayed to me from a source that worked on the original GDI implementation back in the 90s. But since @moi15moi asked (in the typography-issues repo) for more info, I investigated and discovered that what GDI does is slightly different from what I described. I'll provide details in that other issue (https://github.com/MicrosoftDocs/typography-issues/issues/956). One part that still isn't clear to me is whether GDI (or DWrite) expect or give priority to name records with platform/encoding that correspond to 'cmap' subtables.

But I'm curious: have you actually encountered fonts that have 3/4 (or 3/3 or 3/5) name records?

moi15moi:
>But I'm curious: have you actually encountered fonts that have 3/4 (or 3/3 or 3/5) name records?

The only font I have who match these characteristic is ``MINGLI.TTF``

If you want to see if you have installed any font with those characteristic, you can try to run this:
```py
import shutil
import sys
from fontTools import ttLib
from matplotlib import font_manager

def main():
  fontsPath = font_manager.findSystemFonts()
  for fontPath in fontsPath:
    with open(fontPath, 'rb') as fontFile:
      fontType = fontFile.read(4)

    # We don't want collection font.
    if fontType != b'ttcf':
      font = ttLib.TTFont(fontPath)

      for name in font["name"].names:
        if name.platformID == 3 and name.platEncID in [3, 4, 5]:
          shutil.copy(fontPath, r"C:\Users\solmatech\Desktop\font_encoding_id_3_or_4_or_5")
          break

sys.exit(main())
```

moi15moi:
@justvanrossum I have pushed the change 1 week ago.

Is there still something you don't like or are not sure?

justvanrossum:
I was thinking perhaps the special cases should be handled by the caller of `getEncoding()`, instead of passing the nameID into it.

But I would really like to hear opinions from other maintainers.

moi15moi:
Since no maintainers share their opinion, do you still want everything done in ``getEncoding`` from ``_n_a_m_e`` or not?

moi15moi:
@justvanrossum Have you any news about from other maintainers?

behdad:
LGTM.

behdad:
Although, maybe we should implement the DWrite logic instead, which wouldn't need to pass down the nameID. I'm not sure.

moi15moi:
@behdad I don't really know either. I choose GDI, but we can also use DWrite. The important is to pick one of them ^^'

moi15moi:
> But I'm curious: have you actually encountered fonts that have 3/4 (or 3/3 or 3/5) name records?

@PeterCon Yes, here is one: [Font with PlatformID 3 and PlatEncID 3.zip](https://github.com/fonttools/fonttools/files/11012828/Font.with.PlatformID.3.and.PlatEncID.3.zip)

moi15moi:
@justvanrossum I just rebased with main and use DWrite logic.

moi15moi:
@justvanrossum Could I know what is blocking this PR from being merged?

justvanrossum:
I'm not comfortable with this change, but mostly because I don't fully understand the problem nor the solution, and because I think encodingTools.getEncoding() should be patched instead. I defer to other maintainers to make this call.

PR #2622, opened 2022-05-17
Add missing location argument to raise

madig:
Not sure if `location=None` is appropriate here.

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -1560,7 +1560,8 @@ class Builder(object):
     def add_conditionset(self, key, value):
         if not "fvar" in self.font:
             raise FeatureLibError(
-                "Cannot add feature variations to a font without an 'fvar' table"
+                "Cannot add feature variations to a font without an 'fvar' table",
+                location=None,
             )
 
         # Normalize


khaledhosny:
May be pass the location from calling sites?

madig:
Oh no, https://github.com/fonttools/fonttools/blob/2a6459b00404d46b628e2b6900d0e3345f51158a/Lib/fontTools/feaLib/builder.py#L234 has the same bug.

I'll see if I can figure out the calling site somehow.

anthrotype:
@madig good catch, I wonder did you actually encounter this error in real world or a linter pointed you to the missing parameter? Just assessing how likely/urgent this is

madig:
Yes, I manged to trigger this somehow, though I forgot how.

anthrotype:
IIUC, this issue triggers if one attempts to build features.fea containing `conditionset` statements into a non-variable font.
Instead of ignoring the location parameter of FeatureLibError, the caller should pass on the location to the builder `add_conditionset` method. The caller is a `ConditionsetStatement` which knows its `self.location`.
The `add_conditionset` method needs to be changed to accept a `location` parameter. I think this is a pattern seen elsewhere in feaLib.

https://github.com/fonttools/fonttools/blob/2a6459b00404d46b628e2b6900d0e3345f51158a/Lib/fontTools/feaLib/ast.py#L2055-L2071

PR #2626, opened 2022-05-25
Enable sharing across subtables of extension for harfbuzz packing.

garretrieger:
Harfbuzz repacker can handle subtables shared across extension lookups so there's no need to disallow such sharing. Add a parameter to _doneWriting to allow sharing to be enabled when preparing the object graph for passing to harfbuzz.


behdad:
LGTM.

anthrotype:
LGTM, thanks!

PR #2627, opened 2022-05-26
Clean up svgDoc compression

bungeman:
Previously, an entire `SVG ` table would be marked as compressed if any
of the decoded SVG documents in it were compressed. Then on encoding all
SVG documents would be considered for compression. The XML format had no
means to indicate if compression was desired.

Instead, mark each svgDoc with its compression status. When decoding
mark the svgDoc as compressed if the data was compressed. When encoding
try to compress the svgDoc if it is marked as compressed. In the XML
format the data itself is always uncompressed, but allow an optional
`compressed` boolean attribute (defaults to false) to indicate the
svgDoc should be compressed when encoded.

### DIFF
--- a/Lib/fontTools/subset/svg.py
+++ b/Lib/fontTools/subset/svg.py
@@ -202,7 +202,7 @@ def subset_glyphs(self, s) -> bool:
     glyph_index_map: Dict[int, int] = s.glyph_index_map
 
     new_docs: List[Tuple[bytes, int, int]] = []
-    for doc, start, end in self.docList:
+    for doc, start, end, compressed in self.docList:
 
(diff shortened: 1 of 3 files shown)


anthrotype:
thanks @bungeman for this. Makes sense, however I'm only concerned that a lot of client code (e.g. nanoemoji) may rely on the fact that docList is a 3-tuple and making it a 4-tuple would break code that unpacks the tuples like e.g. `for doc, start, end in self.docList: ...`.
This should have been a named tuple to begin with, instead of a bare, positional tuple, so that fields would be accessed via attribute lookup instead of by index.. Probably now it's too late for that.
maybe we could find a way to store this compressed flag somewhere else, e.g. a separate list of bool values of the same length of docList.

PR #2632, opened 2022-05-31
[docs] [designspace spec] Clarify expected types

ctrlcctrlv:


Doc/source/designspaceLib/xml.rst
```diff
@@ -1012,3 +1040,16 @@ The ``<lib>`` element contains arbitrary data.
     </lib>
 
 
+Type equivalence (`plist`)
+==========================
```
belluzj, line 1044: Why did you include this section? I don't think it's necessary, because most of the Designspace XML doesn't use plist (instead defines its own tags and attributes). At most it embeds plist for the lib sections, but the contents of these lib keys is not really defined in this document.

Doc/source/designspaceLib/xml.rst
```diff
@@ -53,6 +53,29 @@ Overview
...
+=====
+
+- **number**: A fixed point value `equivalent to Fixed in OpenType <https://docs.microsoft.com/en-us/typography/opentype/spec/otff#data-types>`_ (a 32-bit signed fixed-point number (16.16))
+- **font design (unsigned) float**: A floating point value `equivalent to (U)FWORD in OpenType <https://docs.microsoft.com/en-us/typography/opentype/spec/otff#data-types>`_
+- **float**: A floating point value `equivalent to F2DOT14 in OpenType <https://docs.microsoft.com/en-us/typography/opentype/spec/otff#data-types>`_
+
+     The F2DOT14 format consists of a signed, 2's complement integer and an unsigned fraction.
+
+- **(unsigned) integer**
+- **boolean**: A true/false value, may be expressed as ``0`` or ``1``
```
belluzj, line 68: I think you're uncovering a real consistency issue here with booleans because some boolean fields are written as 0/1, e.g. axis hidden=1

https://github.com/fonttools/fonttools/blob/42e4d66184be1acda05bed1896da6d61c084835e/Lib/fontTools/designspaceLib/__init__.py#L1454-L1455

and then hidden is read back as "any provided value means True" apparently:

https://github.com/fonttools/fonttools/blob/42e4d66184be1acda05bed1896da6d61c084835e/Lib/fontTools/designspaceLib/__init__.py#L1861-L1862

and some others are read/written as true/false, e.g. elidable and oldersibling

https://github.com/fonttools/fonttools/blob/42e4d66184be1acda05bed1896da6d61c084835e/Lib/fontTools/designspaceLib/__init__.py#L1485-L1488

https://github.com/fonttools/fonttools/blob/42e4d66184be1acda05bed1896da6d61c084835e/Lib/fontTools/designspaceLib/__init__.py#L1902-L1903

So in the end I would say that in addition to making the documentation about booleans more uniform, the PR should also edit the code to make the real situation also uniform. Maybe something like:
* when writing always output 0/1
  * because that's how it's been done so far (the 2 true/false examples above are recent additions from Nikolaus and me, I think they're an oversight because we haven't paid enough attention to the existing convention; I would say it's OK since these two new fields are still young to change the default writer to write 1 instead of true)
* when  reading accept any of "0", "1", "true", "false" and interpret it correctly
  * in the spirit of TCP wisdom "be conservative in what you do, be liberal in what you accept from others", and because I think someone reading the amended documentation quickly, seeing "boolean" as a type and not reading the introduction about types where it's mentioned that the format uses 0/1, would naturally try to write true/false.
  
What do you think?

Doc/source/designspaceLib/xml.rst
```diff
@@ -53,6 +53,29 @@ Overview
...
+    
+=====
+Types
+=====
+
+- **number**: A fixed point value `equivalent to Fixed in OpenType <https://docs.microsoft.com/en-us/typography/opentype/spec/otff#data-types>`_ (a 32-bit signed fixed-point number (16.16))
+- **font design (unsigned) float**: A floating point value `equivalent to (U)FWORD in OpenType <https://docs.microsoft.com/en-us/typography/opentype/spec/otff#data-types>`_
+- **float**: A floating point value `equivalent to F2DOT14 in OpenType <https://docs.microsoft.com/en-us/typography/opentype/spec/otff#data-types>`_
+
+     The F2DOT14 format consists of a signed, 2's complement integer and an unsigned fraction.
```
belluzj, line 65: This is an interesting precision but I wonder in practice how much success people have had with having their floating point values as written in e.g. designspace mappings be carried over to the font file with exactly the same value. Example practical use-case that comes up regularly: making a "staircase" mapping where a continuous change in user location maps to a brutal change in design location. How you would implement that is by making a mapping that has (for example):

| user | design |
| ---- | ---- |
| 100 | 1 |
| 199.9999 | 1 |
| 200 | 2 |
| 300 | 2 |

And ideally, the absolute perfect jump would be achieved by  having the "199.9999" value be exactly the F2DOT14 number that is before 200 as expressed in F2DOT14 (so that there's no opportunity for a ramp between the two numbers). Your addition of the precisions above to the spec makes me think that you're hinting at such a level of control; however I wonder how in practice that works out with compilers and such manipulating the floating point numbers, and whether the above is actually possible.

What use-case did you have in mind when mentioning such technical details as part of this format's documentation?

Doc/source/designspaceLib/xml.rst
ctrlcctrlv: Yep, that's a real problem all right. I'll add code changes as well.

Doc/source/designspaceLib/xml.rst
ctrlcctrlv: Well, this PR stems from me trying to figure out minor issues in work that the talented @ewonrael did for the MFEK Foundation in [Grant №1 of 2022](https://mfek.org/foundation/). "MFEK Serif" is Besley.

He specified one of the axes as ending at 87.5, which I didn't know was even possible. I.e.:

https://github.com/indestructible-type/Besley/blob/c87ddc8b1d852033865cb52680b6bddb3da1b943/sources/designspace/Besley.designspace#L102
```xml
<dimension name="Width" xvalue="87.5"/>
```

That turned out not to be what's wrong, but I documented it for others.

Doc/source/designspaceLib/xml.rst
ctrlcctrlv: I included this section due to `org.statmake.stylespace`. As in https://github.com/fonttools/fonttools/pull/2632#discussion_r886616940, the section comes to us due to a line in MFEK Serif's designspace:

https://github.com/indestructible-type/Besley/blob/c87ddc8b1d852033865cb52680b6bddb3da1b943/sources/designspace/Besley.designspace#L263

I strongly suggest nipping this in the bud now as `stylespace` isn't going anywhere.

belluzj: I would consider that the typing of the contents of a lib key (such as the stylespace lib key) is out of scope of the designspace documentation, because these lib keys could contain anything at all. In this particular case stylespace might sound related to the DS5 but as I say it's "just one of the lib keys" and as you say it's meant to be deprecated in favour of the DS5 additions around STAT anyway, so I think the section can go.

Doc/source/designspaceLib/xml.rst
belluzj: Ah OK, maybe is would be enough to say that floating point number are allowed, instead of referring to the particulars of the OT spec? I wonder how much one could argue that designspace is a "user facing format" that can/should be independent of the particular implementation of the OpenType format (which is also the only format really) but what if someone wants to make a designspace that is super-more-precise that F2DOT14 in its values, and uses a compiler or a font editor that doesn't care about OpenType? Maybe not a realistic scenario, but I'm not sure we should prevent it by tying the floating point values to the (limited) F2DOT14 format.

It's not even just about precision, but I realize now while writing that the F2DOT14 doesn't even go beyond the range of [-2,2], which is always exceeded by a Weight axis for example with values in the [0, 1000] range.

Doc/source/designspaceLib/xml.rst
ctrlcctrlv: Oh, OK. If indeed `org.statmake.stylespace` _is_ going somewhere than you're right, the section probably serves no purpose. :-)

PR #2634, opened 2022-06-01
[designspaceLib] Check for descriptor types with hasattr()

### LINKED ISSUE #2631: designspaceLib isinstance fails for custom descriptors
in Designspace editor we have custom descriptor objects that works nicely with NSObjects 

https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/designspaceLib/__init__.py#L1448-L1452

patch:

hasattr(axisObject, "values") --> DiscreteAxisDescriptor
hasattr(axisObject, "minimum") and hasattr(axisObject, "maximum") --> AxisDescriptor

belluzj:
...to allow custom classes that don't inherit the default descriptors.

Fixes #2631

### DIFF
--- a/Lib/fontTools/designspaceLib/__init__.py
+++ b/Lib/fontTools/designspaceLib/__init__.py
@@ -8,7 +8,7 @@ import os
 import posixpath
 from io import BytesIO, StringIO
 from textwrap import indent
-from typing import Any, Dict, List, MutableMapping, Optional, Tuple, Union
+from typing import Any, Dict, List, MutableMapping, Optional, Tuple, Union, cast
 
 from fontTools.misc import etree as ET
 from fontTools.misc import plistlib
@@ -1352,7 +1352,7 @@ class BaseDocWriter(object):
         minVersion = self.documentObject.formatTuple
         if (
             any(
-                isinstance(axis, DiscreteAxisDescriptor) or
+                hasattr(axis, 'values') or
                 axis.axisOrdering is not None or
                 axis.axisLabels
                 for axis in self.documentObject.axes
@@ -1445,10 +1445,10 @@ class BaseDocWriter(object):
             for label in axisObject.axisLabels:
                 self._addAxisLabel(labelsElement, label)
             axisElement.append(labelsElement)
-        if isinstance(axisObject, AxisDescriptor):
+        if hasattr(axisObject, "minimum"):
             axisElement.attrib['minimum'] = self.intOrFloat(axisObject.minimum)
             axisElement.attrib['maximum'] = self.intOrFloat(axisObject.maximum)
-        elif isinstance(axisObject, DiscreteAxisDescriptor):
+        elif hasattr(axisObject, "values"):
             axisElement.attrib['values'] = " ".join(self.intOrFloat(v) for v in axisObject.values)
         axisElement.attrib['default'] = self.intOrFloat(axisObject.default)
         if axisObject.hidden:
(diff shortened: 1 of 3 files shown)


belluzj:
@typemytype or @justvanrossum, would you be available to review this PR? I hope it works for you with these changes

justvanrossum:
It looks good to me, but I am not in a position to test unfortunately.

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -1445,10 +1445,10 @@ def _addAxis(self, axisObject):
...
                 self._addAxisLabel(labelsElement, label)
             axisElement.append(labelsElement)
-        if isinstance(axisObject, AxisDescriptor):
+        if hasattr(axisObject, "minimum"):
             axisElement.attrib['minimum'] = self.intOrFloat(axisObject.minimum)
             axisElement.attrib['maximum'] = self.intOrFloat(axisObject.maximum)
-        elif isinstance(axisObject, DiscreteAxisDescriptor):
+        elif hasattr(axisObject, "values"):
             axisElement.attrib['values'] = " ".join(self.intOrFloat(v) for v in axisObject.values)
         axisElement.attrib['default'] = self.intOrFloat(axisObject.default)
```
LettError, line 1453: This fails for `DiscreteAxisDescriptor` which does not have a default value. (Also not sure why discrete axes would not want a default value, but I did not follow any discussions about that)

belluzj: `DiscreteAxisDescriptor` has a default value: https://fonttools.readthedocs.io/en/latest/designspaceLib/python.html#fontTools.designspaceLib.DiscreteAxisDescriptor.default

LettError:
Re: my comment on line 1453: DiscreteAxisDescriptor causes a traceback on write.
(Why does a discrete axis not have a default value?)

```#python
from fontTools.designspaceLib import DesignSpaceDocument, SourceDescriptor, InstanceDescriptor, AxisDescriptor, RuleDescriptor, processRules, DiscreteAxisDescriptor

import fontTools
print(fontTools.version)

doc = DesignSpaceDocument()

#https://fonttools.readthedocs.io/en/latest/designspaceLib/python.html#axisdescriptor
a1 = AxisDescriptor()
a1.minimum = 1
a1.maximum = 1000
a1.default = 400
a1.name = "weight"
a1.tag = "wght"
a1.labelNames['fa-IR'] = "قطر"
a1.labelNames['en'] = "Wéíght"
a1.map = [(1.0, 10.0), (400.0, 66.0), (1000.0, 990.0)]
a1.axisOrdering = 1
doc.addAxis(a1)

a2 = DiscreteAxisDescriptor()
a2.values = [0, 1]
a2.name = "Italic"
a2.tag = "ITAL"
a2.labelNames['fr'] = "Italique"
#a2.map = [(0, 0), (1, -11)]
a2.axisOrdering = 2
#a2.axisLabels = [
#    AxisLabelDescriptor(name="Roman", userValue=0, elidable=True)
#]
doc.addAxis(a2)

path = "test.ds5.designspace"
doc.write(path)
```

Traceback:
```
  File "/Users/erik/code/ufoProcessor/Lib/ufoProcessor/ds5_makeTestDoc.py", line 34, in <module>
    doc.write(path)
  File "/Users/erik/code/ltr_py_env/lib/python3.7/site-packages/fontTools/designspaceLib/__init__.py", line 2482, in write
    writer.write()
  File "/Users/erik/code/ltr_py_env/lib/python3.7/site-packages/fontTools/designspaceLib/__init__.py", line 1303, in write
    self._addAxis(axisObject)
  File "/Users/erik/code/ltr_py_env/lib/python3.7/site-packages/fontTools/designspaceLib/__init__.py", line 1453, in _addAxis
    axisElement.attrib['default'] = self.intOrFloat(axisObject.default)
  File "/Users/erik/code/ltr_py_env/lib/python3.7/site-packages/fontTools/designspaceLib/__init__.py", line 1398, in intOrFloat
    if int(num) == num:
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

belluzj:
I would say it doesn't have a default value because you're not giving it one (I see you're doing `a1.default = 400` but not `a2.default = 0` for example). Is the sample code above from the documentation? If so we should update it to also set a default value on a2.

The class in general has a default property which default to None.

Lib/fontTools/designspaceLib/__init__.py
LettError: (traceback posted)

LettError:
Yes, this is from the documentation:
https://fonttools.readthedocs.io/en/latest/designspaceLib/python.html#discreteaxisdescriptor

belluzj:
Thanks, I updated the docstring

LettError:
Thank you!

anthrotype (approved):

PR #2635, opened 2022-06-01
[varLib] use nameIDs 2,17 and 6 for default instance

m4rc1e:
According to the spec:

> The default instance of a font is that instance for which the coordinate value of each axis is the defaultValue specified in the corresponding variation axis record. An instance record is not required for the default instance, though an instance record can be provided. When enumerating named instances, the default instance should be enumerated even if there is no corresponding instance record. If an instance record is included for the default instance (that is, an instance record has coordinates set to default values), then the nameID value should be set to either 2 or 17, and the postScriptNameID value should be set to 6.

https://docs.microsoft.com/en-us/typography/opentype/spec/fvar#instancerecord

Tests currently fail but if you're happy to include this, I'll fix them.

PR has been made due to https://github.com/googlefonts/fontbakery/issues/3703.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -74,6 +74,7 @@ def _add_fvar(font, axes, instances: List[InstanceDescriptor]):
 	fvar = newTable('fvar')
 	nameTable = font['name']
 
+	dflt_axis_coordinates = {v.tag: v.default for k,v in axes.items()}
 	for a in axes.values():
 		axis = Axis()
 		axis.axisTag = Tag(a.tag)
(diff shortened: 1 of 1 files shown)


behdad:
Let me ask this way: what font issue is this fixing?

behdad:
Okay, I see this is one of those times that I and the spec wording disagree. :))

khaledhosny:
`addMultilingualName` is the one that does reusing existing names.

m4rc1e:
Going to close this due to the discussion over at https://github.com/MicrosoftDocs/typography-issues/issues/946. Thanks all!

If we did want to implement this, we could use `addMultilingualName` as Khaled suggested.

Issue #2623, opened 2022-05-18
debugname return me arial when it should return me linkin


moi15moi:
I have this code and I am using this [font](https://github.com/fonttools/fonttools/files/8720284/linkin.zip)
```py
from fontTools import ttLib
font = ttLib.TTFont("C:\\Users\\example\\Desktop\\linkin.ttf")
print(font['name'].getDebugName(1).strip().lower())
```

It print me "arial" when it should print me "linkin" or "linkin park".

Here is the xml record:
```xml
  <name>
    <namerecord nameID="0" platformID="0" platEncID="3" langID="0x0">
      Typeface © The Monotype Corporation plc. Data © The Monotype Corporation plc/Type Solutions Inc. 1990-1992. All Rights Reserved
    </namerecord>
    <namerecord nameID="1" platformID="0" platEncID="3" langID="0x0">
      Arial
    </namerecord>
    <namerecord nameID="2" platformID="0" platEncID="3" langID="0x0">
      Regular
    </namerecord>
    <namerecord nameID="3" platformID="0" platEncID="3" langID="0x0">
      Monotype:Arial Regular:Version 2.76 (Microsoft)
    </namerecord>
    <namerecord nameID="4" platformID="0" platEncID="3" langID="0x0">
      Arial
    </namerecord>
    <namerecord nameID="5" platformID="0" platEncID="3" langID="0x0">
      Version 2.76
    </namerecord>
    <namerecord nameID="6" platformID="0" platEncID="3" langID="0x0">
      ArialMT
    </namerecord>
    <namerecord nameID="7" platformID="0" platEncID="3" langID="0x0">
      Arial® Trademark of The Monotype Corporation plc registered in the US Pat &amp; TM Off. and elsewhere.
    </namerecord>
    <namerecord nameID="8" platformID="0" platEncID="3" langID="0x0">
      Monotype Typography
    </namerecord>
    <namerecord nameID="9" platformID="0" platEncID="3" langID="0x0">
      Monotype Type Drawing Office - Robin Nicholas, Patricia Saunders 1982
    </namerecord>
    <namerecord nameID="10" platformID="0" platEncID="3" langID="0x0">
[...]

PR #2636, opened 2022-06-02
Sort naming table to get the same name has FontConfig

moi15moi:
This should close: https://github.com/fonttools/fonttools/issues/2623

I inspired myself from FontConfig.

This is still not perfect, because FontConfig use freetype for the postscript name. I am not able to fully understand what does freetype, but I know that freetype choose the right postscript name here: https://github.com/freetype/freetype/blob/c26f0d0d7e1b24863193ab2808f67798456dfc9c/src/sfnt/sfdriver.c#L1045-L1087

I added a new dependency, sortedcontainers. It is very usefull to always have a sortedlist.

### DIFF
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -469,7 +469,7 @@ class FontBuilder(object):
             variationsPostScriptNamePrefix (nameID 25)
         """
         nameTable = self.font["name"] = newTable("name")
-        nameTable.names = []
+        nameTable.initializeNames()
 
         for nameName, nameValue in nameStrings.items():
             if isinstance(nameName, int):
--- a/Lib/fontTools/ttLib/tables/_n_a_m_e.py
+++ b/Lib/fontTools/ttLib/tables/_n_a_m_e.py
@@ -1,14 +1,12 @@
 # -*- coding: utf-8 -*-
-from math import floor
 from fontTools.misc import sstruct
 from fontTools.misc.textTools import bytechr, byteord, bytesjoin, strjoin, tobytes, tostr, safeEval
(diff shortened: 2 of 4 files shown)


khaledhosny:
I don't think such a big breaking change is going to be accepted. This is going to break so much user code.

I think you are better off implementing your own name getting function on top of FontTools code.

moi15moi:
Then, I could implement a Strategy pattern.

The method getDebugGame would have 2 argument.
1: NameID (int)
2: Sorting strategy (by default, it would be what we currently have)

We could have these strategy: default (what we currently have), fontconfig, coretext (apple), DirectWrite, Windows GDI.

Like that, it would break nothing and would give a better way to get the nameid depending on the sort we want.

Want do you think about that? @khaledhosny

khaledhosny:
That is too specific for an API. You can really implement this is a few lines and save yourself all this truble:

```
names = sorted(font['name'].names, key=...)
for name in names:
  if name.nameID == 1:
     ...
```

moi15moi:
Why would it be too specific?

In my opinion, getDebugName is pretty useless right now because it does not respect any standard. Because of that, the result we can get is pretty useless.

khaledhosny:
If you say so.

moi15moi:
Ok, I have created a new branch where I implement the default and FontConfig sort: https://github.com/moi15moi/fonttools/tree/add-strategy-naming-table

I will try to contact apple (CoreText) and microsoft (DirectWrite, GDI) to know if I can have some information on how they pick up the name in the naming table since I don't have, obviously, access to their code.

I will close this pullrequest in 48 hours if no one say anything about it

PR #2643, opened 2022-06-06
Hatch: How does it work???

madig:
Test driving hatch.


anthrotype:
I don't know this hatch tool, and am not convinced we need to ditch setuptools or tox 🤔

madig:
Aaargh. I clicked the wrong button, this was supposed to stay in the dama fork because I'm just playing around. Sorry for the noise.

behdad:
What does it do though? :)

madig:
It's an exciting new tool that combines aspects of Peotry and Tox and a couple other things to make Python project management easier and it's already adopted by the PyPA. It fixes a couple things that annoy me about Poetry so I'm keen to try it out :)

PR #2645, opened 2022-06-09
Clean up svgDoc compression #2

anthrotype:
This is based on @bungeman's https://github.com/fonttools/fonttools/pull/2627 and it's supposed to supersede that one.

The difference is this tries to ensure that existing code that depends on `docList` containing sequences of three items `(doc, startGID, endGID)` continues to work without modification.

I define a new SVGDocument dataclass that exports a Sequence-like interface, only the first three elements can be read by their position in the fake tuple.

### DIFF
--- a/Lib/fontTools/subset/svg.py
+++ b/Lib/fontTools/subset/svg.py
@@ -14,6 +14,7 @@ except ModuleNotFoundError:
 
 from fontTools import ttLib
 from fontTools.subset.util import _add_method
+from fontTools.ttLib.tables.S_V_G_ import SVGDocument
 
 
 __all__ = ["subset_glyphs"]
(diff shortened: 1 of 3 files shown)


anthrotype:
PTAL @bungeman

bungeman:
Thanks, this seems really nice.

At the moment using `compressed` as a boolean like thing is probably fine. As a bikeshed I had been thinking that maybe the attribute should be called `encoding` with possible values of `auto` (the default, assume plain text unless starts with gzip magic), `plain-text` or `gzip` (or `gzip-deflate`). This is a little more in line with the wording in the specification and leaves a little more wiggle room in the future should anything unexpected happen. I used `compressed` simply because that was the name used with `getattr` previously for the "top level applies to all svgDocs" property.

anthrotype:
I guess we could add that later (if ever). Let's go with this for now

PR #2649, opened 2022-06-13
Add flag to mutator to allow recalc timestamp to be disabled.

garretrieger:


anthrotype:
thanks @garretrieger, I took the liberty of renaming it `--no-recalc-timestamp` (as hyphen is more canonical word separator with argparse generated options, and it also matches the name of the option in pyftsubset).
I also added the option to `fonttools varLib.instancer` CLI tool. By the way, the latter encompasses all the features of `varLib.mutator` and adds more, was actually supposed to supersede it (though we never officially deprecated mutator, perhaps we should).

PR #2650, opened 2022-06-14
Improve error msg for building substitutions with missing glyphs

jenskutilek:
The error message was not helpful for debugging. Give a hint that missing glyphs are referenced in a substitution.

Before:
```
TypeError: ("unsupported operand type(s) for -: 'NoneType' and 'int'", 0, 'SubTable[]', 134, 'Lookup[]', 'LookupList')
```

After:
```
ERROR:fontTools.ttLib.tables.otTables:Substitution references one or more missing glyphs. Check any items that contain a glyph ID of None:
ERROR:fontTools.ttLib.tables.otTables:[({'arrowleft.alt1': 1427}, {'arrowleft.alt1.3': None}), ({'uni23AF.alt1': 1436}, {'uni23AF.alt1.3': None}), ({'arrowright.alt1': 1445}, {'arrowright.alt1.3': None}), ({'arrowup.alt1': 1455}, {'arrowup.alt1.3': None}), ({'uni23D0.alt1': 1464}, {'uni23D0.alt1.3': None}), ({'arrowdown.alt1': 1473}, {'arrowdown.alt1.3': None})]
```

### DIFF
--- a/Lib/fontTools/ttLib/tables/otTables.py
+++ b/Lib/fontTools/ttLib/tables/otTables.py
@@ -779,6 +779,17 @@ class SingleSubst(FormatSwitchingBaseTable):
 		format = 2
 		delta = None
 		for inID, outID in gidItems:
+			if inID is None or outID is None:
+				log.error(
+					"Substitution references one or more missing glyphs. "
+					"Check any items that contain a glyph ID of None:"
+				)
+				log.error([
+						({v[0]: k[0]}, {v[1]: k[1]})
+						for k, v in sortableItems
+				])
+				raise KeyError
+
 			if delta is None:
 				delta = (outID - inID) % 65536
 


Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -779,6 +779,17 @@ def preWrite(self, font):
 		format = 2
 		delta = None
 		for inID, outID in gidItems:
+			if inID is None or outID is None:
+				log.error(
+					"Substitution references one or more missing glyphs. "
+					"Check any items that contain a glyph ID of None:"
```
anthrotype, line 785: why should the user check for Nones, can't the KeyError contain the name of the missing glyph?

jenskutilek: I thought it would be easier to find the feature that contains the error when a bit of context is given. It might be a mistake in the feature code, or a glyph that actually should be in the font, etc.

anthrotype: "a mistake in the feature code" should be caught earlier than that

"a glyph that actually should be in the font" is a faulty font or a bug somewhere

jenskutilek:
It happens in a feature that is added from a designspace rule via fontmake. Not sure what the exact sequence of steps is there, where the earliest opportunity of catching the error is.

Lib/fontTools/ttLib/tables/otTables.py
anthrotype: in general I don't like the combination of a logging ERROR message with a bare KeyError exception. There's no link between the two, and logging message can be ignored, the exception itself carries no message of its own

PR #2651, opened 2022-06-14
[varLib.iup] Rewrite force-set conditions & limit DP lookback length

behdad:
This does two things:

Fixes forced-set computation, which was wrong in multiple ways.
Debugged it. Is solid now... Famous last words.

Speeds up DP time by limiting DP lookback length. For Noto Sans,
IUP time drops from 23s down to 9s, with only a slight size increase
in the final font. This basically turns the algorithm from O(n^3) into
O(n).


anthrotype:
> For Noto Sans, IUP time drops from 23s down to 9s

that's great!

> Fixes forced-set computation, which was wrong in multiple ways

do you have a test case that shows how was it "wrong" and how this fixes it?

I'm actually not sure what's the current IUP test suite's coverage, so the mere fact that tests "pass" may not mean much.
The code is quite dense and I might not be able to do a full review pass right now.
But if you say it is "solid" now, I take you on your words.

Maybe @simoncozens who has worked on the Rust port recently can take a closer look.

behdad:
I added some comments at least.

behdad:
I'll add some tests.

behdad:
I'm seeing something really weird. On my Linux machine the speedup is much less than on my Mac :(. No idea why.

behdad:
> I'm seeing something really weird. On my Linux machine the speedup is much less than on my Mac :(. No idea why.

For NotoSans, calling varLib module without layout interpolation, and with this branch, takes 13s on my Mac *laptop*, and 30s on my AMD desktop. That's quite unexpected.

behdad:
> > I'm seeing something really weird. On my Linux machine the speedup is much less than on my Mac :(. No idea why.
> 
> For NotoSans, calling varLib module without layout interpolation, and with this branch, takes 13s on my Mac _laptop_, and 30s on my AMD desktop. That's quite unexpected.

My bad. I had debugging code committed accidentally.

behdad:
Still the Mac laptop is ~20% faster. That's amazing.

behdad:
Added a few tests. Granted, it's not much, but a start.

Tests/varLib/iup_test.py
```diff
@@ -0,0 +1,64 @@
...
+        assert forced == f
+
+        chain1, costs1 = iup._iup_contour_optimize_dp(delta, coords, f)
+        chain2, costs2 = iup._iup_contour_optimize_dp(delta, coords, set())
+
+        assert chain1 == chain2, f
+        assert costs1 == costs2, f
+
+if __name__ == "__main__":
+    sys.exit(unittest.main())
```
anthrotype, line 64: BTW, you don't need to import `unittest` just to call its `main()`; since you're using the pytest API, you can either remove the `if __name__ == "__main__":` block altogether and just calls a test module with `pytest name_of_test_module.py` (most of the test suite does that); or if you really want to run you test module like a script, you can add 

```python
if __name__ == "__main__":
    sys.exit(pytest.main(sys.argv))
```

which achieves the same thing as `unittest.main()`.

anthrotype:
> Added a few tests. 

better than nothing, thanks! Out of curiosity, do these pass or fail _without_ your patch?

simoncozens:
FWIW, there are Rust IUP unit tests at https://github.com/simoncozens/rust-font-tools/blob/07e0c6ec007004090da1b0dd783f6e61b3dd74e2/fonttools-rs/src/otvar/iup.rs#L461-L514

behdad:
@simoncozens Please port these changes to your Rust implementation.

Tests/varLib/iup_test.py
behdad: Thanks. Done.

behdad:
@anthrotype How does this look now?

Tests/varLib/iup_test.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+]
+
+class IupTest:
+
+    def setUp(self):
+        pass
+
+    def tearDown(self):
```
anthrotype, line 15: You don't need these two no-op setUp/tearDown methods, pytest discover the test class by its name ending in "Test"

anthrotype (approved):
LGTM

Tests/varLib/iup_test.py
```diff
@@ -0,0 +1,63 @@
+import fontTools.varLib.iup as iup
+import sys
+import pytest
+
+test_cases = [
```
anthrotype, line 5: Empty list is unused

behdad:
@anthrotype Thanks. Fixed.

behdad:
@anthrotype Is this good now?

behdad:
Ah sorry I missed the LGTM before.

PR #2655, opened 2022-06-15
Fix tests for Python 3.11

pnemade:
Submitting this Fedora 37 downstream patch by @hroncok (Miro Hrončok) 
This will fix running some tests against Python 3.11

### DIFF
--- a/Tests/ufoLib/ufoLib_test.py
+++ b/Tests/ufoLib/ufoLib_test.py
@@ -35,7 +35,7 @@ def test_formatVersionTuple(ufo_path):
     assert reader.formatVersionTuple == (3, 0)
     assert reader.formatVersionTuple.major == 3
     assert reader.formatVersionTuple.minor == 0
-    assert str(reader.formatVersionTuple) == "3.0"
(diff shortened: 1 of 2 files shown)


Tests/ufoLib/ufoLib_test.py
```diff
@@ -35,7 +35,7 @@ def test_formatVersionTuple(ufo_path):
     assert reader.formatVersionTuple == (3, 0)
     assert reader.formatVersionTuple.major == 3
     assert reader.formatVersionTuple.minor == 0
-    assert str(reader.formatVersionTuple) == "3.0"
+    assert str(reader.formatVersionTuple) in ("3.0", "UFOFormatVersion.FORMAT_3_0")
```
anthrotype, line 38: this one is problematic, why should it be different for 3.11...

anthrotype:
Fixed with https://github.com/fonttools/fonttools/commit/14f88fb781e2618a1e7ceaf2a5caa8809ac3d181

justvanrossum:
Upstream 3.11 issue has apparently been resolved: https://github.com/python/cpython/issues/94601

Issue #2328, opened 2021-06-01
[varLib] support building variable COLRv1 table from set of masters


anthrotype:
Now that we have support for 32-bit VarStore, required to variate some of the COLRv1 fields, fontTools.varLib should be updated to create a COLRv1.VarStore from a set of master fonts containing compatible COLRv1 tables.

behdad:
First step would be to produce a set of masters to use for testing / merging.

Do you suggest adapting `varLib` to do merging? Or update your toolchain to generate variations directly? We can do the former, but would require idential-structure limitations.

anthrotype:
yes, I was thinking varLib would do the merging, similarly to how it handles the other tables.
Indeed, COLR master tables would have to have identical structure.
nanoemoji currently can build a variable color font when given (via a config file) groups of SVG glyphs associated with a given location. It creates UFO masters for each and hands them off to ufo2ft to compile the variable font (which internally uses fontTools.varLib). But the COLR table is not varLib-merged yet, so only the outlines, metrics, etc. vary.

Issue #2211, opened 2021-03-03
[varLib.varStore] Return 0xFFFFFFFF if there's no variation


behdad:
Forking from https://github.com/simoncozens/fontFeatures/issues/39#issuecomment-790001038

Back when I was writing this code, there was no varIdx value that meant "no variation". So the callers of the OnlineVarStoreBuilder had to do the check themselves. Now'ish we added 0xFFFFFFFF to mean "no variations" to OpenType, so we can use that to make the client's job simpler.

I don't think it's an issue with backward compatibility. We'll update all our internal callsites and be done.

The optimizer can also choose to through empty ones out and return 0xFFFFFFFF. Currently it ends up spending a few bytes registering a region that has no variation and one data item with no deltas for it...

Note that, in constrained uses of the ItemVariationStore, like in HVAR/VVAR tables, it might be still beneficial to encode no-variations explicitly. So the builder should have a mode to do that.

behdad:
Also possibly make `OnlineVarStoreBuilder.finish()` return `None` if the resulting varStore holds no variations?  That would be a bigger API break... Currently in some places we do things like:
```
        store = merger.store_builder.finish()
        if not store.VarData:
                return
```

behdad:
My bad typo. I meant 0xFFFFFFFF.

behdad:
> Also possibly make `OnlineVarStoreBuilder.finish()` return `None` if the resulting varStore holds no variations? That would be a bigger API break...

@anthrotype WDYT?

anthrotype:
> return None if the resulting varStore holds no variations

I guess it'd be ok, I don't believe the OnlineVarStoreBuilder has extensively been used by used code outside of fontTools.varLib

behdad:
Sounds good. Thanks.

PR #2660, opened 2022-06-21
[varLib] add support for building variable COLR from set of master COLRv1 tables

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/2328

I'd like to add more unit tests before merge, but since this is already been used for generating [variable COLRv1 test fonts](https://github.com/googlefonts/color-fonts/pull/102) to aid Skia/FreeType implementation, as well as for stuff like https://github.com/BlackFoundryCom/black-renderer/pull/88, I think it's better to put this out for review so I stop tinkering with it.

### DIFF
--- a/Tests/varLib/data/TestVariableCOLR.designspace
+++ b/Tests/varLib/data/TestVariableCOLR.designspace
@@ -1,0 +1,18 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<designspace format="5.0">
+  <axes>
+    <axis tag="wght" name="Weight" minimum="400" maximum="700" default="400"/>
+  </axes>
+  <sources>
+    <source filename="master_ttx_varcolr_ttf/TestVariableCOLR-Regular.ttx" name="Regular">
+      <location>
+        <dimension name="Weight" xvalue="400"/>
+      </location>
+    </source>
+    <source filename="master_ttx_varcolr_ttf/TestVariableCOLR-Bold.ttx" name="Bold">
+      <location>
+        <dimension name="Weight" xvalue="700"/>
+      </location>
+    </source>
+  </sources>
+</designspace>
--- a/Tests/varLib/data/master_ttx_varcolr_ttf/TestVariableCOLR-Bold.ttx
+++ b/Tests/varLib/data/master_ttx_varcolr_ttf/TestVariableCOLR-Bold.ttx
@@ -1,0 +1,357 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ttFont sfntVersion="\x00\x01\x00\x00" ttLibVersion="4.33">
+
+  <GlyphOrder>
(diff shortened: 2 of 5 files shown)


Lib/fontTools/varLib/__init__.py
```diff
@@ -711,6 +713,28 @@ def _add_CFF2(varFont, model, master_fonts):
...
+	merger = COLRVariationMerger(model, axisTags, font)
+	merger.mergeTables(font, master_fonts)
+	store = merger.store_builder.finish()
+
+	colr = font["COLR"].table
+	if store.VarData:
+		mapping = store.optimize()
+		colr.VarStore = store
+		# the special 0xFFFF/0xFFFF (for no variations) always maps to itself
+		mapping[ot.NO_VARIATION_INDEX] = ot.NO_VARIATION_INDEX
```
behdad, line 726: This needs to go into the VarStore logic itself.
Generally this issue: https://github.com/fonttools/fonttools/issues/2211

behdad: Ie. Other places are affected as well.

behdad:
I reviewed this. Great work @anthrotype. Some of the changes to otBase machinery are a bit crude but I don't see any cleaner way. Thanks.

anthrotype:
reminder to self: given how with variability, pre-computing tight clip boxes is non-trivial (think of rotate, skew or other complex transforms), we may have to add an option to drop the clip boxes in the variable COLR (they are optional, renderer can compute on the fly from the paint graph), or better use some euristics to decide if it's safe or not to keep them on a glyph-by-glyph basis. Right now they are simply linearly interpolated like everything else.

Also see https://github.com/googlefonts/colr-gradients-spec/issues/289

anthrotype:
one notable change since my last push. Previously, the merger required that all masters had the same number of BaseGlyphList records, as well as the same LayerList count, and would fail as soon as it found a PaintColrLayers with a different FirstLayerIndex. But that immediately ruled out masters with "sparse" color glyph sets, which was too restrictive, given how varLib.build does support that when merging, say, gvar variations, for example.
Now, with the latest changes (esp. https://github.com/fonttools/fonttools/pull/2660/commits/e5029801d2c1739b8a51de926b3cb636edf2926e), it is possible to merge multiple master COLR tables containing different base color glyphs, or having different LayerList count or layer-reuse layout. Instead of treating PaintColrLayers as constant tables, I now traverse their references into the LayerList while merging the paint graph rooted at each base color glyph, and only require that each color glyph masters resolve to the same number of (flattened) layers. For the same reason, it's no longer necessary (even though it may speed things up a bit) to ensure that layer reuse is disabled when building each master COLR table, the algorithm works either way (duplicating references as needed to make sure the merger's in-place mutation doesn't go bad). And it's no longer necessary to unbuild the variable COLR table at the end just to be able to rebuild it with layer reuse turned on, since now the merger applies layer reuse as it goes along.

PR #2662, opened 2022-06-21
[varLib.varStore] Support NO_VARIATION_INDEX

behdad:
Feels a bit fragile but I think I got it right.

### DIFF
--- a/Lib/fontTools/varLib/instancer/__init__.py
+++ b/Lib/fontTools/varLib/instancer/__init__.py
@@ -633,6 +633,7 @@ def instantiateItemVariationStore(itemVarStore, fvarAxes, axisLimits):
         for major, deltas in enumerate(defaultDeltaArray)
         for minor, delta in enumerate(deltas)
     }
+    defaultDeltas[itemVarStore.NO_VARIATION_INDEX] = 0
     return defaultDeltas
 
 
--- a/Lib/fontTools/varLib/varStore.py
+++ b/Lib/fontTools/varLib/varStore.py
@@ -7,6 +7,10 @@ from functools import partial
 from collections import defaultdict
 
 
+NO_VARIATION_INDEX = 0xFFFFFFFF
+ot.VarStore.NO_VARIATION_INDEX = NO_VARIATION_INDEX
+
+
 def _getLocationKey(loc):
 	return tuple(sorted(loc.items(), key=lambda kv: kv[0]))
 
--- a/Lib/fontTools/varLib/varStore.py
+++ b/Lib/fontTools/varLib/varStore.py
@@ -178,6 +178,7 @@ class VarStoreInstancer(object):
 
 	def __getitem__(self, varidx):
 		major, minor = varidx >> 16, varidx & 0xFFFF
+		if varidx == NO_VARIATION_INDEX: return 0.
 		varData = self._varData
 		scalars = [self._getScalar(ri) for ri in varData[major].VarRegionIndex]
 		deltas = varData[major].Item[minor]
@@ -440,7 +441,7 @@ class _EncodingDict(dict):
 		return chars
 
 
-def VarStore_optimize(self):
+def VarStore_optimize(self, use_NO_VARIATION_INDEX=True):
 	"""Optimize storage. Returns mapping from old VarIdxes to new ones."""
 
 	# TODO
@@ -463,6 +464,10 @@ def VarStore_optimize(self):
 			for regionIdx,v in zip(regionIndices, item):
 				row[regionIdx] += v
 			row = tuple(row)
+
+			if use_NO_VARIATION_INDEX and not any(row):
+				front_mapping[(major<<16)+minor] = None
+				continue
 
 			encodings.add_row(row)
 			front_mapping[(major<<16)+minor] = row
@@ -546,9 +551,9 @@ def VarStore_optimize(self):
 			back_mapping[item] = (major<<16)+minor
 
 	# Compile final mapping.
-	varidx_map = {}
+	varidx_map = {NO_VARIATION_INDEX:NO_VARIATION_INDEX}
 	for k,v in front_mapping.items():
-		varidx_map[k] = back_mapping[v]
+		varidx_map[k] = back_mapping[v] if v is not None else NO_VARIATION_INDEX
 
 	# Remove unused regions.
 	self.prune_regions()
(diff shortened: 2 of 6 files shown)


Lib/fontTools/varLib/varStore.py
```diff
@@ -37,12 +41,14 @@ def setSupports(self, supports):
...
 
-	def finish(self, optimize=True):
+	def finish(self, optimize=True, empty=True):
 		self._regionList.RegionCount = len(self._regionList.Region)
 		self._store.VarDataCount = len(self._store.VarData)
 		for data in self._store.VarData:
 			data.ItemCount = len(data.Item)
 			data.calculateNumShorts(optimize=optimize)
+		if empty is None and not self._store.VarDataCount:
+			return None
```
anthrotype, line 51: it doesn't seem like you're using this empty=None parameter you just added anywhere. Where did you think it could be handy?

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -429,7 +429,7 @@ def setMvarDeltas(varfont, deltas):
         if mvarTag not in MVAR_ENTRIES:
             continue
         tableTag, itemName = MVAR_ENTRIES[mvarTag]
-        delta = deltas[rec.VarIdx]
+        delta = deltas.get(rec.VarIdx, 0)
```
anthrotype, line 432: hm, wouldn't it be better to add a `{NO_VARIATION_INDEX: 0}` before returning `defaultDeltas` from `instantiateItemVariationStore`? That way if `rec.VarIdx` really is no-op it works, but for whatever reason it points to some non existent VarIdx, it would fail with KeyError, this not hiding a bug somewhere in the font or code.
Like this:

```diff
diff --git a/Lib/fontTools/varLib/instancer/__init__.py b/Lib/fontTools/varLib/instancer/__init__.py
index 986ad5a4..91dd586f 100644
--- a/Lib/fontTools/varLib/instancer/__init__.py
+++ b/Lib/fontTools/varLib/instancer/__init__.py
@@ -429,7 +429,7 @@ def setMvarDeltas(varfont, deltas):
         if mvarTag not in MVAR_ENTRIES:
             continue
         tableTag, itemName = MVAR_ENTRIES[mvarTag]
-        delta = deltas.get(rec.VarIdx, 0)
+        delta = deltas[rec.VarIdx]
         if delta != 0:
             setattr(
                 varfont[tableTag],
@@ -633,6 +633,7 @@ def instantiateItemVariationStore(itemVarStore, fvarAxes, axisLimits):
         for major, deltas in enumerate(defaultDeltaArray)
         for minor, delta in enumerate(deltas)
     }
+    defaultDeltas[itemVarStore.NO_VARIATION_INDEX] = 0
     return defaultDeltas


diff --git a/Tests/varLib/instancer/instancer_test.py b/Tests/varLib/instancer/instancer_test.py
index 2757c3da..c3b6f71d 100644
--- a/Tests/varLib/instancer/instancer_test.py
+++ b/Tests/varLib/instancer/instancer_test.py
@@ -456,6 +456,8 @@ class InstantiateItemVariationStoreTest(object):

         defaultDeltaArray = []
         for varidx, delta in sorted(defaultDeltas.items()):
+            if varidx == varStore.NO_VARIATION_INDEX:
+                continue
             major, minor = varidx >> 16, varidx & 0xFFFF
             if major == len(defaultDeltaArray):
                 defaultDeltaArray.append([])

```

anthrotype: potentially (even though unlikely) a NO_VARIATION_INDEX could be found inside a GPOS ValueRecord or GDEF Anchor device table. You already modified `VarStoreInstancer.__getitem__` to return 0 when key is NO_VARIATION_INDEX, and this pseudo-dict `instancer` object in used in `MutatorMerger` for instancing GPOS/GDEF, so if a NO_VARIATION_INDEX occurs, it's handled there.

In `varLib.instancer:instantiateOTL` I call `MutatorMerger` with the `defaultDeltas` dict that are returned from `instantiateItemVariationStore` (not with an actual `VarStoreInstancer` object); if this dict doesn't contain a map for {NO_VARIATION_INDEX: 0} it could KeyError.

So my patch above would handle that case as well.

Lib/fontTools/varLib/varStore.py
```diff
@@ -7,6 +7,10 @@
 from collections import defaultdict
 
 
+NO_VARIATION_INDEX = 0xFFFFFFFF
```
anthrotype, line 10: after we merge #2660, otTables module would contain a NO_VARIATION_INDEX constant so here you could alias to that to avoid duplication (i know it's fairly trivial)

anthrotype (approved):
The changes to varLib.varStore.py LGTM, but see other comments

Tests/varLib/varStore_test.py
```diff
@@ -13,7 +13,7 @@
         (
             [{}, {"a": 1}],
             [
-                [10, 20],
```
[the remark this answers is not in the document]
behdad: It does. [10, 10] produces an empty delta which exercises `NO_VARIATION_INDEX` codepath.

Lib/fontTools/varLib/varStore.py
behdad: I don't use it currently. I thought in some situations we can do like:
```
  table.VarStore = varStore.optimize(empty=None)
```
But am fine dropping that commit.

Lib/fontTools/varLib/instancer/__init__.py
behdad: Thanks. WIll do.

Tests/varLib/varStore_test.py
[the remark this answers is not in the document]
anthrotype: ha! you're right. You may as well write a little comment to that effect

behdad:
Thanks @anthrotype. I fixed based on your feedback.

Lib/fontTools/varLib/varStore.py
anthrotype: i see. Well, until we find a use for it, i'd remove the extra code paths

Tests/varLib/varStore_test.py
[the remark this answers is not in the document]
behdad: Done.

Lib/fontTools/varLib/varStore.py
behdad: Okay. Let me do that.

PR #2665, opened 2022-06-23
[feaLib] show all missing glyphs at once

simoncozens:
I just had a frustrating situation where the FEA parser reported a missing glyph in my source, and then I fixed it and re-ran it, and then it reported another missing glyph, and I fixed it and re-ran it, and...

With this PR, "missing glyph" errors are gathered and reported at the end of parsing:

```
ERROR: The following glyph names are referenced but are missing from the glyph set:
 ampersand (first found at source/build/features.fea:43570:16)
 asterisk (first found at source/build/features.fea:43605:14)
```

### DIFF
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -73,6 +73,7 @@ class Parser(object):
         self.next_token_location_ = None
         lexerClass = IncludingLexer if followIncludes else NonIncludingLexer
         self.lexer_ = lexerClass(featurefile, includeDir=includeDir)
+        self.missing = {}
         self.advance_lexer_(comments=True)
 
     def parse(self):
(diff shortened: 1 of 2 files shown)


Lib/fontTools/feaLib/parser.py
```diff
@@ -2079,13 +2089,11 @@ def check_glyph_name_in_glyph_set(self, *names):
...
-                raise FeatureLibError(
-                    "The following glyph names are referenced but are missing from the "
-                    f"glyph set: {', '.join(missing)}",
-                    self.cur_token_location_,
-                )
+            for name in names:
+                if name in self.glyphNames_:
+                    continue
+                if name not in self.missing:
+                    self.missing[name] = self.cur_token_location_
```
anthrotype, line 2096: the docstring of `check_glyph_name_in_glyph_set` says "Raises if..." but it no longer raises, does it?

anthrotype (commented):
I can see how that can be handy, however I wonder if it's a problem delaying the exception at the very end of the parsing and whether there's some code in between that may assume that given glyph names exist after running `check_glyph_name_in_glyph_set`.
But maybe it's fine.

simoncozens:
You can create an ast.Glyph or ast.GlyphClass object with non-existent glyphs; it'll only fail when you try and build it. It's a semantic problem rather than a syntactic one. So reporting at the end of parsing is fine.

anthrotype:
so technically the parser _could_ finish its job without raising then..

anthrotype:
maybe Parser.parse() method could have an ignoreMissingGlyphs=False parameter.. well, nevermind. LGTM

anthrotype (approved):

Issue #2613, opened 2022-05-05
[varLib instancer] should set up italic style linking as an option


RosaWagner:
When slicing a VF with ital or slant axis, varLib instancer could set up proper italic style bits + italic angle + caret slope if `italic != 0` and if `slant != 0`. For example with and additional argument like `--is-italic` followed by an integer for italic angle.

For example when running:
`fonttools varLib.instancer "../fonts/variable/Nunito[ital,wght].ttf" ital=1 wght=200:1000 -o "../fonts/variable/Nunito-Italic[wght].ttf"`

Then we have to patch this kind of metadata:

```
for font_path in glob.glob("../fonts/variable/*Italic*.ttf"):
        with open(font_path, "rb") as f:
            print("Fix italic style linking in {}".format(font_path))
            ttFont = TTFont(f)
            ttFont["head"].macStyle = MacStyle.ITALIC
            ttFont["post"].italicAngle = -9
            ttFont["hhea"].caretSlopeRise = 1000
            ttFont["hhea"].caretSlopeRun = 158
            ttFont.save(font_path)
```

madig:
Here's some code to set up some kind of linking:

```py
    # style mapping
    styleMapStyleName = font["name"].getName(2, 3, 1, 0x409).toStr().lower()
    macStyle = []
    if styleMapStyleName == "bold":
        macStyle = [0]
    elif styleMapStyleName == "bold italic":
        macStyle = [0, 1]
    elif styleMapStyleName == "italic":
        macStyle = [1]
    font["head"].macStyle = intListToNum(macStyle, 0, 16)

    selection = font["OS/2"].fsSelection
    selection &= ~(1 << 0)
    selection &= ~(1 << 5)
    selection &= ~(1 << 6)
    if styleMapStyleName == "regular":
        selection |= 1 << 6
    elif styleMapStyleName == "bold":
        selection |= 1 << 5
    elif styleMapStyleName == "italic":
        selection |= 1 << 0
    elif styleMapStyleName == "bold italic":
        selection |= 1 << 0
        selection |= 1 << 5
    font["OS/2"].fsSelection = selection
```

anthrotype:
yeah, I like that.
basically what I suggested to do https://github.com/fonttools/fonttools/issues/1264#issuecomment-422425249 but never got around doing it

PRs welcome!

madig:
@RosaWagner Can you please test with the latest fontTools release and report what else instancer should be doing?

[...]

PR #2673, opened 2022-06-30
Set RIBBI bits when cutting instances

madig:
Tackles https://github.com/fonttools/fonttools/issues/2613.

Needs tests.

### DIFF
--- a/Lib/fontTools/varLib/instancer/__init__.py
+++ b/Lib/fontTools/varLib/instancer/__init__.py
@@ -90,12 +90,10 @@ from fontTools.varLib import builder
 from fontTools.varLib.mvar import MVAR_ENTRIES
 from fontTools.varLib.merger import MutatorMerger
 from fontTools.varLib.instancer import names
-from contextlib import contextmanager
 import collections
 from copy import deepcopy
 from enum import IntEnum
 import logging
-from itertools import islice
 import os
 import re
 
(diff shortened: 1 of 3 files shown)


Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -1273,9 +1271,57 @@ def instantiateVariableFont(
...
     return varfont
 
 
+def setRibbiBits(font):
+    """Set the `head.macStyle` and `OS/2.fsSelection` style bits
+    appropriately."""
+
+    english_ribbi_style = font["name"].getName(2, 3, 1, 0x409)
+    if english_ribbi_style is None:
+        log.warning(
```
anthrotype, line 1287: i think all these logging warnings are too noisy, make them debug or remove them. If one doesn't have those names it may just mean they are not interested in legacy style linking behavior.

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -1273,9 +1271,57 @@ def instantiateVariableFont(
         },
     )
 
+    # Set Regular/Italic/Bold/Bold Italic bits as appropriate, after the name
+    # table has been updated.
+    setRibbiBits(varfont)
```
anthrotype, line 1276: I wonder if you should do it only when updateFontNames is True, that's the thing that may or may not trigger changes in style name. If the original varfont that is being instanced didn't have those bit sets correctly it's not our job to fix them here. Only if the style name changed there's a chance these bits will need to be updated.

m4rc1e (commented):
Imo it would be great if this also worked for non-ribbi fonts. We have this functionality in gftools, https://github.com/googlefonts/gftools/blob/main/Lib/gftools/fix.py#L221-L264

I'm guessing if we instantiate a VF whose origin is Bold to a Medium static font, this isn't going to work correctly.

m4rc1e:
> if we instantiate a VF whose origin is Bold to a Medium static font

>> bits should be cleared, I think this does that already.

Yeah, I just took a look and it works as expected. What about going from a Bold Italic VF to a Medium Italic static?

madig:
So this PR does what you need?

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -1273,9 +1271,51 @@ def instantiateVariableFont(
...
+        setRibbiBits(varfont)
+
     return varfont
 
 
+def setRibbiBits(font):
+    """Set the `head.macStyle` and `OS/2.fsSelection` style bits
+    appropriately."""
+
+    english_ribbi_style = font["name"].getName(2, 3, 1, 0x409)
```
m4rc1e, line 1286: Nit. We have a NameID enum in https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/varLib/instancer/names.py so we could do: 

```PYTHON
font["name"].getName(NameID.SUBFAMILY_NAME, 3, 1, 0x409)
```

anthrotype (approved):

Issue #2672, opened 2022-06-30
[momentsPen] Definitions of (2nd) moments


n8willis:
Finally following up on the issue that I originally ran into (which in turn prompted #2401).

As I understand it, planar second moment of area w.r.t. X is defined as:

Ix = Int Int y^2 dA

and planar second moment of area w.r.t. Y is defined as:

Iy = Int Int x^2 dA

(Those are both as listed in https://en.wikipedia.org/wiki/Second_moment_of_area plus in similar online sources, like https://mathworld.wolfram.com/AreaMomentofInertia.html ) 

But MomentsPen defines them differently in the GreenPens it generates:
```
printGreenPen('MomentsPen', [
		      ('area', 1),
		      ('momentX', x),
		      ('momentY', y),
		      ('momentXX', x**2),
		      ('momentXY', x*y),
		      ('momentYY', y**2),
		     ])
```

Looks to me like the right-side `x`s and `y`s are all swapped.

That ended up causing me to run into problems when trying to generate pens for higher moments (and polars).

I think I might need someone to convince me that the FontTools code is correct here (it doesn't look right to me, and the external sources I've looked at seem to be in agreement with each other), but in any case, using notation that's different from the literature certainly causes confusion.

I think the _formulae_ used in statisticsPen to get meanX and meanY are correct, since it's dividing the correct integrals by the area, but the definitions (of what those integrals are) are swapped. That is, `xbar` should be `My / A`, and the code has My and Mx swapped in the GreenPens, but also swaps them in the "mean" computations.

I would also note that if the definitions are swapped, it's pretty likely to have escaped notice, since area is the only one most people are taking advantage of, and they probably just use areaPen, and the only direct usage of statisticsPen I've found on GitHub is Chris Simpkins's diff tools, which test on "plus" and "period" glyphs where you wouldn't notice the X/Y swap at all because they're symmetric.

Happy to be told that I've got it swapped around backwards in my head, rather than the code. But the code doesn't agree with the texts I see.

behdad:
I see what you mean. I named these myself, not after any literature source :(.

n8willis:
Been there! Took me ages to get the "about an axis" stuff straight in my head. Then the polar stuff ... yow.

Anyway, I am working on a PR; at least I think it won't affect anything beyond statisticsPen.

n8willis:
[...]

PR #2674, opened 2022-06-30
Moments fix

n8willis:
This targets #2672 .... Updating the First and Second Moment integral definitions that are passed to GreenPen, which then generate MomentsPen, which is used for some attributes in StatisticsPen.

It just swaps x and y for those integrals, but then it has to swap them back around where they get used in StatisticsPen.

I am fairly confident it's correct for the meanX and meanY cases; somewhat less certain for varianceX and varianceY; possibly there are ripple effects, and I have no idea where they might be felt.

People using mean/variance should not notice any difference, but anybody using the moment* attributes directly will see new output that hopefully makes more sense.

It also makes the notation match math texts elsewhere, which should make it easier to add new moments.


behdad:
I'm a bit undecided about this. How about at least we completely rename the variables, from `momentXX` to `XXmoment`, such that any code using one wouldn't work with the other?

behdad:
Also asking @anthrotype's opinion

n8willis:
> I'm a bit undecided about this. How about at least we completely rename the variables, from `momentXX` to `XXmoment`, such that any code using one wouldn't work with the other?

Brief thought, then I have to call it an evening, but it _might_ be interesting to name them `Planar1stMomentX` etc; which would leave the door open for the polar moments in a consistent fashion.

n8willis:
> > I'm a bit undecided about this. How about at least we completely rename the variables, from `momentXX` to `XXmoment`, such that any code using one wouldn't work with the other?

Sorry; one other clarification question: are you meaning that you want to preserve the older `momentXX` etc attributes? 

Because I think that would cause confusion. If anyone is actually using those directly (which is doubtful, I know), they could keep pinned the old version of the library, but they probably ought to be pushed to migrate their code. At least they ought to be flagged as deprecated to warn people, but I really don't think having both versions hang around has much benefit.

n8willis:
Here's where I'm at for a naming scheme. Either:

momentX -> Planar1stMomentAxisX
momentXX -> Planar2ndMomentAxisX
momentY -> Planar1stMomentAxisY
momentYY -> Planar2ndMomentAxisY

or:

momentX -> Planar1stMomentWrtX
momentXX -> Planar2ndMomentWrtX
momentY -> Planar1stMomentWrtY
momentYY -> Planar2ndMomentWrtY


I can rebase the `HigherMomentsPen` idea from #2340 on either of those schemes, and the higher moments and polar moments will be easy to understand based on their names. I feel like it'd be less confusing than putting XX and YY in the names, since the names are supposed to be a reference to the x and y axes.

The only tricky notation bit then would be what to call the various cross-moment parts (like XYY, XXY) that don't really have a clear meaning on their own, but would be convenient internally for algebraically getting the useful ones (as in [this comment](https://github.com/fonttools/fonttools/pull/2340#issuecomment-867737382)). But that could use some other, totally distinct naming scheme, like CrossMomentIntegralXXY or CrossMomentIntegralYYYX, and I add some comments to document why those are there.

But that's later. I'm happy to regenerate this bit with new naming scheme, but I'd rather here what the reviewers think about it and go with their preference once and for all. No real preference myself between "Wrt" and "Axis".

behdad:
> momentXX -> Planar2ndMomentAxisX

Shouldn't that be AxisY then?

n8willis:
Maybe I should take out the arrows from the previous comment; I guess they're confusing things.

n8willis:
Real question I wanted to ask is mainly whether `Planar2ndMomentAxisX` or `Planar2ndMomentWrtX` is better.

behdad:
> Real question I wanted to ask is mainly whether `Planar2ndMomentAxisX` or `Planar2ndMomentWrtX` is better.

I like the Wrt version better. What would the XY be called?

Re higher-level moments, If it doesn't slow things down, I'm fine with it. I don't remember if these are all computed together or separately... Looks like together, ie. adding higher-level will slow things down.

Polar moments are just linear combination of Planar moments so again, if can be added without cost, that's preferred.

behdad:
Adding Polar moment might be really low-cost. Give it a try and check the code.

behdad:
I just generated one with `momentRR` for Polar second moment, and the generator wasn't smart enough to reuse X^2 + Y^2 code for it. Which is weird...

n8willis:
> > Real question I wanted to ask is mainly whether `Planar2ndMomentAxisX` or `Planar2ndMomentWrtX` is better.
> 
> I like the Wrt version better. What would the XY be called?

Looking at Wikipedia again, I think `ProductMomentXY`, `ProductMomentXXY`, etc would probably do it without trying to be overly clever.

> 
> Re higher-level moments, If it doesn't slow things down, I'm fine with it. I don't remember if these are all computed together or separately... Looks like together, ie. adding higher-level will slow things down.
> 
> Polar moments are just linear combination of Planar moments so again, if can be added without cost, that's preferred.

Yeah, that's how it seemed to work out last time. The idea was to pull out the higher moments and put them in a separate pen, with the polar ones, because for most people StatisticsPen did everything they need and there was no reason to add the overhead, etc.

n8willis:
> Also, do you have a use-case for the planar moment, or the higher moments?

It's from research. The higher & polar moments were what Kindersley & Wiseman were using to find optical centers, particularly in the 80s/90s (long time after that book was published, which everyone's read).

Even though optical centers don't "solve" spacing, there's some useful things people could do with them. I dunno; mark-to-base centering, etc etc.

n8willis:
> I just generated one with `momentRR` for Polar second moment, and the generator wasn't smart enough to reuse X^2 + Y^2 code for it. Which is weird...

Interesting. Seems like it should be possible to force it to, but I haven't tried. Is the output deterministic?

Lib/fontTools/pens/momentsPen.py
```diff
@@ -274,19 +274,19 @@ def _curveToOne(self, p1, p2, p3):
...
+		self.Planar2ndMomentWrtY += r111*r131/9240 - r132*r72/9240 - r132*y3/4 + 3*r133*(r30 + r48)/3080 - r52*r70*(-r35 + y2)/9240 - r55*r77*(r11 - r48)/9240 - r6*(-84*r113 - r114 - r115 - r121*y3 + 54*r127 + r38*r52 + r55*r62 + 14*r56)/9240 + r60*(-r39*y1 + r64 + r67 + r78)/9240 + r61*(r118*y3 + r124*x2 - 14*r129 - 630*r130 + r30*x3 - r77*y0 + r85*x2 + r89 + 210*x1*y2)/9240 - r7*(27*r131 + 42*r132 + 70*r134 + r135)/9240 + x0**3*(r30 + r38 + 165*y0 + y3)/660 + x0*(42*r113 + 90*r116*y3 - r117*y0 - r119*y0 - r120*y0 - r121*y0 - 9*r122 - 27*r123 - 18*r126 - 30*r127 + r128 + r136*y2 + 162*r41*x1 + r42*x3 + r54 + 63*r59*y3 - r66*y0 - r68*x3 + r99*x1)/9240 - y0*(90*r110*x1 + 45*r131 + 14*r132 + 126*r133 + 42*r134 + r135 + r136*x2 + r25*r55 + 135*r52*x1 + r65*x3)/9240
 
 if __name__ == '__main__':
 	from fontTools.misc.symfont import x, y, printGreenPen
 	printGreenPen('MomentsPen', [
 		      ('area', 1),
```
behdad, line 286: Maybe capitalize `area` now as well?

Lib/fontTools/pens/momentsPen.py
```diff
@@ -274,19 +274,19 @@ def _curveToOne(self, p1, p2, p3):
...
-		      ('momentX', y),
-		      ('momentY', x),
-		      ('momentXX', y**2),
-		      ('momentXY', x*y),
-		      ('momentYY', x**2),
+		      ('Planar1stMomentWrtX', y),
+		      ('Planar1stMomentWrtY', x),
+		      ('Planar2ndMomentWrtX', y**2),
+		      ('ProductMomentXY', x*y),
```
behdad, line 290: Move this one after the next just for style?

Lib/fontTools/pens/momentsPen.py
n8willis: Well, that's breaking a test or two.

behdad:
Sorry for asking cosmetic changes, but please don't change the statisticsPen `area`; we don't want to change the API there.

Lib/fontTools/pens/momentsPen.py
n8willis: Looks like breaking people's code loudly on `area`->`Area` is going to be loud indeed.

behdad:
> Sorry for asking cosmetic changes, but please don't change the statisticsPen `area`; we don't want to change the API there.

Weird. You're not. I don't see why you pushed the last commit. Anyway, let me check the tests.

behdad:
Oh... Damn. We have a problem. I had forgotten that `StatisticsPen` subclasses `MomentsPen`. So we *are* breaking API. Shoot.

n8willis:
> Change of tactic. What do you think about reverting this back, and adding properties to StatisticsPen with the new names? Can also add the Polar 2nd moment there.

Wouldn't bother me at all. Although ... re-adding properties to restore the old names, while keeping the new names internally, might be the smoother way for doing the higher-moments junk.

behdad:
> re-adding properties to restore the old names, while keeping the new names internally, might be the smoother way for doing the higher-moments junk.

What's the plan for higher moments without slowing down? We can do a separate pen, since it's free basically.

n8willis:
> > re-adding properties to restore the old names, while keeping the new names internally, might be the smoother way for doing the higher-moments junk.
> 
> What's the plan for higher moments without slowing down? We can do a separate pen, since it's free basically.

Right; I was thinking separate pen for sure. But it just seems preferable to have the two generated pens be consistent with each other on how they're generated & notation, and then make the "compatibility properties" in statisticsPen be the part that 's the special case ... instead of having MomentsPen and HigherMomentsPen be different from each other internally.

behdad:
> Right; I was thinking separate pen for sure. But it just seems preferable to have the two generated pens be consistent with each other on how they're generated & notation, and then make the "compatibility properties" in statisticsPen be the part that 's the special case ... instead of having MomentsPen and HigherMomentsPen be different from each other internally.

Okay agreed.

n8willis:
OK; I think I've got that resolved. Need to figure out what Emacs is doing with the tabs here, I guess.

Didn't add the polar one yet, just to stay focused on fixing what's already there instead of new stuff. IMHO that could wait a little, since maybe there's something else interesting to add to StatisticsPen, but it's obviously pretty trivial to just do.

n8willis:
Other remaining question is do I need to do something to the docs pages....

Long-term, seems like that would be helpful, but then again very few of the pens have any significant documentation beyond the docstrings from the internal methods.

anthrotype:
> Passing to anthrotype 

honestly I don't know enough about this code or the underlying math to review this, I trust Behdad's LGTM so I'll just merge as is.

@n8willis could you please summarize the changes this PR introduces so we can update the next release notes? You could either write them in here as comment (or editing the first PR message at the top), or if you wish you could commit a change yourself to the `NEWS.rst` file adding a new entry at the very top following the style of the rest of the file (I can do that after merge if you prefer).

n8willis:
> @n8willis could you please summarize the changes this PR introduces so we can update the next release notes? You could either write them in here as comment (or editing the first PR message at the top), or if you wish you could commit a change yourself to the `NEWS.rst` file adding a new entry at the very top following the style of the rest of the file (I can do that after merge if you prefer).

Okay; how about:

[momentsPen/statisticsPen] Modified the formulae used internally to generate the various "moments" properties of StatisticsPen and added new properties that more directly reflect the mathematical meaning of the moments (e.g., `Planar1stMomentWrtX`). All of the old moment properties (e.g., `momentX`) remain, but should be considered deprecated. 

?

anthrotype:
sounds good, thanks

anthrotype:
@n8willis can you please resolve the merge conflict?

n8willis:
I'm not sure if I understand it, TBH.

Just because I haven't really interacted with the test framework, that is.

anthrotype:
around line 100 is the conflict, looks like this:

```
<<<<<<< moments-fix
		for item in ['area', 'Planar1stMomentWrtX', 'Planar1stMomentWrtY', 'Planar2ndMomentWrtX', 'Planar2ndMomentWrtY', 'ProductMomentXY', 'meanX', 'meanY', 'varianceX', 'varianceY', 'stddevX', 'stddevY', 'covariance', 'correlation', 'slant']:
			if item[0] == '_': continue
=======
		for item in ['area', 'momentX', 'momentY', 'momentXX', 'momentYY', 'momentXY', 'meanX', 'meanY', 'varianceX', 'varianceY', 'stddevX', 'stddevY', 'covariance', 'correlation', 'slant']:
>>>>>>> main
```

maybe @behdad pushed some changes to main branch that conflicts with yours, either you or him would have to decide what should be in that list of strings.

behdad:
> if item[0] == '_': continue

Yeah I removed this line because it wasn't necessary anymore.

n8willis:
OK; makes sense. Should I keep the old properties (`momentX`) in that list, though? Seems superfluous, as they're aliased, but maybe it matters?

behdad:
> OK; makes sense. Should I keep the old properties (`momentX`) in that list, though? Seems superfluous, as they're aliased, but maybe it matters?

No, print the new ones only. Thanks.

n8willis:
Done.

n8willis:
Sure is simpler just to not ever do any testing in your code. As I do with all my personal work.

n8willis:
I'm hung up on something when rebasing this, using `printGreenPen()` ... at that docstring-test addition, I guess. Raises this:
```
  File "/home/nate/code/fonttools/Lib/fontTools/misc/symfont.py", line 117, in printGreenPen
    '''from fontTools.pens.basePen import BasePen, OpenContourError
TypeError: not enough arguments for format string
```
That doesn't make sense to me; that's the first `print(` . I'm missing something here.... 

If there's a quick answer I've overlooked, I could use pointing in the right direction. Otherwise, I'm juggling a lot of unrelated stuff this week so I'll have to come back to it in a few days & see if it looks clearer with fresher eyes.

behdad:
My guess is that you merged conflicts incorrectly.

Lib/fontTools/misc/symfont.py
```diff
@@ -101,18 +101,21 @@ def _curveToOne(self, p1, p2, p3):
...
+Planar2ndMomentWrtYPen = partial(GreenPen, func=x*x)
+ProductMomentXYPen = partial(GreenPen, func=x*y)
 
 
 def printGreenPen(penName, funcs, file=sys.stdout):
 
 	print(
 '''from fontTools.pens.basePen import BasePen
 
+class %sOpenContourError(NotImplementedError):
```
anthrotype, line 116: this `%s` formatting characters should not be here

behdad: This whole class we decided should not be here since I moved it to BasePen.

PR #2675, opened 2022-07-02
[feaLib] Let the parser set nameIDs 1 through 6

dscorbett:
Since https://github.com/adobe-type-tools/afdko/commit/3fa06c51db37d3a2fb967552d61c47c9fbb8584c, only nameIDs 2 and 6 are reserved.


khaledhosny:
I’d go all the way and remove this check. This is an artificial limitation possibly influenced by how makeotf builds fonts and I don't see why we have to enforce it at all.

anthrotype:
I'd also vote for removing the check altogether

khaledhosny (approved):

PR #2678, opened 2022-07-04
Implement merging fonts to beyond 64k glyphs

behdad:


behdad:
I'm now hitting `LookupList` offset overflows now, which means I need to bump GSUB/GPOS structs to version 2.0. That's a bit hard to do in ttx currently.

anthrotype:
I'm away and won't be able to review this before I'll be back to work on 18 July, sorry!

khaledhosny:
> I don't know if I'm 100% comfortable with the latter idea

Why not?

khaledhosny:
> > I don't know if I'm 100% comfortable with the latter idea
> 
> Why not?

I mean, HarfBuzz supported is merged in main already and will be available in the next release.

behdad:
> before I begin to review, I was wondering what's the plan with this. Are you planning on keeping it as a separate branch (occasionally rebased on main) while the work progresses on the boring-expansion spec front, or do you intend to have it in main before any of it becomes official in the OT spec? I don't know if I'm 100% comfortable with the latter idea...

As Khaled said, this is already in HarfBuzz main branch. So in that sense it's ahead of where COLRv1 was when it got merged into fonttools. As for OT spec, it's likely that this work will never make it into OT spec, because it's quite likely that MS will not have resources to implement it. Implementing it in HarfBuzz was a week's work because of the HarfBuzz architecture that allowed templatizing types. MS implementation does not afford them the same from what I understand.

So, this is a HB-only font format for now. Next step is to generate a single-page spec.

anthrotype:
I'd suggest that you first `git rebase 78a939848` (78a939848 is the commit on main that immediately precedes the black PR #2925), fix the merge conflicts, and then rebase on, or merge in the `main` branch, resolving any conflicts by automatically running black

behdad:
Superseded by https://github.com/fonttools/fonttools/pull/4097

PR #2684, opened 2022-07-07
[designspaceLib] Don't make up bad PS names when no STAT data

belluzj:
See GitHub issue: https://github.com/googlefonts/ufo2ft/issues/630


justvanrossum:
Tested locally on our real font family: the problem seems to be gone indeed. Thank you!

belluzj:
I just checked that ufo2ft and fontmake's tox still works with this branch. Who could review the changes that's not Nikolaus or me? Thanks in advance!

justvanrossum:
Thank you!!

PR #2685, opened 2022-07-07
Fix typo in COLR merge

justvanrossum:


khaledhosny (approved):

anthrotype:
Oops, apologies. It's hard to hit that in a test font as only triggers when there are more than 255 layers in a color glyph and didn't find a way to test that without bloating the test suite.

anthrotype:
4th release in less than 24 hours! https://github.com/fonttools/fonttools/releases/tag/4.34.4

PR #2695, opened 2022-07-16
Allow variable scalar arithmetic

simoncozens:
Sometimes you have to sum, subtract, average, multiply, etc. variable scalars. (e.g. to answer questions like "What is the average location of this anchor across multiple glyphs?") This does that.


Lib/fontTools/feaLib/variableScalar.py
```diff
@@ -95,3 +95,66 @@ def add_to_variation_store(self, store_builder):
         store_builder.setSupports(supports)
         index = store_builder.storeDeltas(deltas)
         return int(self.default), index
+
+    def __sub__(self, other):
```
anthrotype, line 99: the canonical way for magic operator overrides is to first check if other is not an instance of the expected type, `return NotImplemented` so that the other operand has the chance to implement the operation:

```python
if not isinstance(other, VariableScalar):
    return NotImplemented
```

Tests/feaLib/varscalar_test.py
```diff
@@ -0,0 +1,103 @@
...
+    v.add_value({"wght": 400}, 10)
+    v.add_value({"wght": 800}, 20)
+
+    v = v - 30
+    assert v.value_at_location({"wght": 400}) == -20
+    assert v.value_at_location({"wght": 500}) == -17.5
+    assert v.value_at_location({"wght": 800}) == -10
+
+
+def test_add_varscalar(axes):
```
anthrotype, line 52: codecov reveals that `VariableScalar - VariableScalar` isn't tested. Also untested are subtraction/addition where second operand is a VariableScalar that doesn't vary.

anthrotype (approved):
LGTM, with comment

Lib/fontTools/feaLib/variableScalar.py
```diff
@@ -84,11 +84,17 @@ def default(self):
         return self.values[key]
 
     def value_at_location(self, location):
-        loc = location
+        if isinstance(location, dict):
+            loc = Location(location)
```
anthrotype, line 88: this `Location` might as well be an actual type, instead of a function disguised as one

Lib/fontTools/feaLib/variableScalar.py
```diff
@@ -109,3 +115,88 @@ def add_to_variation_store(self, store_builder):
         store_builder.setSupports(supports)
         index = store_builder.storeDeltas(deltas)
         return int(self.default), index
+
+    def __sub__(self, other):
+        if not isinstance(other, (VariableScalar, float, int)):
+            return NotImplemented
+        new = VariableScalar()
+        new.axes = self.axes
```
anthrotype, line 123: curious why VariableScalar's default constructor doesn't take an `axes` parameter but you set it after the fact?

Tests/feaLib/varscalar_test.py
```diff
@@ -0,0 +1,103 @@
...
+    assert v.value_at_location({"wght": 800}) == 20
+
+
+def test_add(axes):
+    v = VariableScalar()
+    v.axes = axes
+    v.add_value({"wght": 400}, 10)
+    v.add_value({"wght": 800}, 20)
+
+    v = v + 30
```
anthrotype, line 34: if you're going to do that `v = v + ...` often you could implement the in-place operators e.g `__iadd__/__isub__` etc.

Lib/fontTools/feaLib/variableScalar.py
```diff
@@ -109,3 +115,88 @@ def add_to_variation_store(self, store_builder):
...
+                    dict(loc),
+                    self.value_at_location(loc) - list(other.values.values())[0],
+                )
+        return new
+
+    def __add__(self, other):
+        if not isinstance(other, (VariableScalar, float, int)):
+            return NotImplemented
+        new = VariableScalar()
+        new.axes = self.axes
```
anthrotype, line 143: what if other.axes != self.axes? Is that valid? If so, please add a test for that.

PR #2697, opened 2022-07-18
Improvements to timing

simoncozens:
This PR does two things:

* adds a call stack to the `Timer` class, so that you can time individual operations and sub-operations
* adds a singleton `GlobalTimer` class, so that a Timer can be attached once and accessed from anywhere in the program (but is ignored if no Timer is attached).

This allows you to do things like:

(in fontmake)
```python
        if timing:
            configLogger(logger=timer.logger, level=logging.INFO)
            GlobalTimer.attach(timer)

...

    @GlobalTimer("load designspace sources")
    def _load_designspace_sources(self, designspace):
```

in ufo2ft:

```python
    GlobalTimer.start_step("merge")
    vfNameToTTFont = varLib.build_many(
        designSpaceDoc,
        exclude=excludeVariationTables,
        optimize=optimizeGvar,
        skip_vf=lambda vf_name: variableFontNames and vf_name not in variableFontNames,
    )
    GlobalTimer.end_step()
```

and get output like:

```
Starting to load designspace sources
Took 0.258s to load designspace sources
Starting to check UFO compatibility
Took 7.096s to check UFO compatibility
Starting to build variable fonts
  Starting to run 'call_preprocessor'
  Took 4.126s to run 'call_preprocessor'
...
  Starting to merge
  Took 9.087s to merge
  Starting to post-process final VFs
    Starting to run 'call_postprocessor'
    Took 2.076s to run 'call_postprocessor'
  Took 2.076s to post-process final VFs
  Starting to save VFs
  Took 0.031s to save VFs
Took 27.104s to build variable fonts
```

### DIFF
--- a/Lib/fontTools/misc/loggingTools.py
+++ b/Lib/fontTools/misc/loggingTools.py
@@ -404,6 +404,52 @@ class Timer(object):
 		return "%.3f" % self.elapsed
 
 
+class _Forwarder(type):
+	def __getattr__(self, name):
+		if self._instance:
+			return getattr(self._instance, name)
+		else:
+			return self.do_nothing
+
+	def do_nothing(self, *args, **kwargs):
+		pass
+
+class GlobalTimer(metaclass=_Forwarder):
+	_instance = None
+
+	@classmethod
+	def attach(self, timer):
+		self._instance = timer
+
+	def __init__(self, msg=None):
+		self.msg = msg
+
+	def __call__(self, func=None, **kwargs):
+		@wraps(func)
+		def wrapper(*args, **kwds):
+			if not self._instance:
+				return func(*args, **kwds)
+			if self.msg:
+				self._instance.msg = self.msg
+			else:
+				self._instance.msg = "run '%s'" % func.__name__
+			with self:
+				return func(*args, **kwds)
+		return wrapper
+
+	def __enter__(self):
+		if self._instance:
+			self._instance.start_step(self._instance.msg)
+		return self
+
+	def __exit__(self, exc_type, exc_value, traceback):
+		if exc_type:
+			return
+		if self._instance:
+			self._instance.end_step()
+		return self
+
+
 class ChannelsFilter(logging.Filter):
 	"""Provides a hierarchical filter for log entries based on channel names.
 
--- a/Lib/fontTools/misc/loggingTools.py
+++ b/Lib/fontTools/misc/loggingTools.py
@@ -282,7 +282,8 @@ class Timer(object):
 	# timeit.default_timer choses the most accurate clock for each platform
 	_time = timeit.default_timer
 	default_msg = "elapsed time: %(time).3fs"
-	default_format = "Took %(time).3fs to %(msg)s"
+	default_format = "%(indent)sTook %(time).3fs to %(msg)s"
+	start_step_format = "%(indent)sStarting to %(msg)s"
 
 	def __init__(self, logger=None, msg=None, level=None, start=None):
 		self.reset(start)
@@ -294,6 +295,7 @@ class Timer(object):
 		self.logger = logger
 		self.level = level if level is not None else TIME_LEVEL
 		self.msg = msg
+		self.stack = []
 
 	def reset(self, start=None):
 		""" Reset timer to 'start_time' or the current time. """
@@ -315,6 +317,20 @@ class Timer(object):
 		self.last = current
 		return self.elapsed
 
+	def start_step(self, msg):
+		indent = "  " * len(self.stack)
+		start_msg = self.start_step_format % {"msg": self.msg, "indent": indent}
+		self.logger.log(self.level, start_msg)
+		self.stack.append((self._time(), msg))
+		self.elapsed = 0.0
+		self.msg = msg
+
+	def end_step(self):
+		""" Finishes the current operation, logging the time elapsed. """
+		self.start, self.msg = self.stack.pop()
+		self.last = self.start
+		self.__exit__(None, None, None)  # Reuse finish/logging code
+
 	def formatTime(self, msg, time):
 		""" Format 'time' value in 'msg' and return formatted string.
 		If 'msg' contains a '%(time)' format string, try to use that.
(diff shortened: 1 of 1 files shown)


anthrotype:
wait, I don't understand why you need a GlobalTimer, aren't loggers as such already global?

anthrotype:
also the `start_step`/`end_step` methods, aren't those already covered by the context-manager behaviour of Timers (i.e. `with timer(...): ...`) ?

simoncozens:
> wait, I don't understand why you need a GlobalTimer, aren't loggers as such already global?

But timers aren't.

simoncozens:
> also the start_step/end_step methods, aren't those already covered by the context-manager behaviour of Timers (i.e. with timer(...): ...) ?

No, because the context-manager behaviour is currently designed to track multiple laps of the same step, not separate steps. I didn't want to change that behaviour, so I had to implement it in separate functions.

anthrotype:
I see. I haven't looked at this code in a long time, need to refresh my memory a bit. In any case, we should add tests to cover the new functionality.

anthrotype:
> so you can preserve the call stack between them so you get a report on the hierarchy

so this is basically to prettify the log messages with indentation, in a way that works across library boundaries

> the context-manager behaviour is currently designed to track multiple laps of the same step, not separate steps

I don't think so, where do you get that from? A new "step" is simply one where you change the message, e.g. this is from the `Timer` class docstring:

```python
        >>> timer = Timer(log)
	>>> with timer():
	...     time.sleep(0.01)
	elapsed time: ...s
	>>> with timer('redo it', level=logging.INFO):
	...     time.sleep(0.02)
	Took ... to redo it
```

simoncozens:
> so this is basically to prettify the log messages with indentation, in a way that works across library boundaries

Right.

> > the context-manager behaviour is currently designed to track multiple laps of the same step, not separate steps

> I don't think so, where do you get that from? 

From here:

```python
	def __exit__(self, exc_type, exc_value, traceback):
		""" End the current lap. If timer has a logger, log the time elapsed,
		using the format string in self.msg (or the default one).
		"""
		time = self.split()
```

anthrotype:
How would this work in practice? You'd have to change fontmake and all the dependant libraries that use timers to instead use this new GlobalTimer singleton (instead of instantiating their own). This would be no-op unless some caller higher up (in this particular case, a command line utility like fontmake) attaches a timer to this global, singleton timer.
I think this introduces a little too much complexity for relatively small benefit of having the timer messages indented. 
Also who owns the GlobalTimer? What if something in-between attaches a different timer to this global timer, is that even supported/desirable? I don't know.

simoncozens:
> You'd have to change fontmake and all the dependant libraries that use timers to instead use this new GlobalTimer singleton (instead of instantiating their own)

I think you're overestimating how much timing these libraries do already. :-)

You're right that I may be overcomplicated this, but maybe not. Here's how my thinking worked:

fontmake has a `--timing` flag which aims to turn on and off timing information on a global basis. But we don't actually have a global timing logger. You have to turn on DEBUG logging (which gives you a lot more information than you want just to find out timings.) Currently to add timing information to steps in ufo2ft (or any other dependent library), each library will have to instantiate its own Timer object and feed it some kind of logging object. So ufo2ft feeds it *its own* logger, and to find out the timing information you have to wade through all the mess of DEBUG messages and pick out:

```
DEBUG:ufo2ft.filters.base:Took 0.012s to run FlattenComponentsFilter on 624 glyphs
```

(Actually it logs this when logging=DEBUG, regardless of whether `--timing` is turned on or not, so the flag is a bit redundant.)

And so to implement the `--timing` flag - to turn on and off timing information on a global basis - you need (a) a global log facility, (b) which is configured independently of the verbosity of general library-specific log facilities, and (c) which all libraries log their timing information to. And hence a global timer.

anthrotype:
> I think you're overestimating how much timing these libraries do already. :-)

ah ok, so currently we are _not_ using timers across library boundaries, only inside fontmake itself, correct? You'd like to reuse the same timer in ufo2ft.

> You have to turn on DEBUG logging (which gives you a lot more information than you want just to find out timings.)

The `--timing` option attaches a handler to a logger named `"fontmake.timer"` so the latter's output should be independent from the output of the root logger (which is configured via basicConfig by the `--verbose` flag). What I mean is, `running fontmake --timer MyFont.glyphs` should give you timing messages coming from within fontmake itself at least, regardless of the --verbose flag.

You would like fontmake's --timing option to also control the output of ufo2ft's timers? Then within the current framework, you just have fontmake call configLogger on the logger named "ufo2ft.timer" which configures a handler for it in the same way fontmake does for its own timer-logger. No need to change anything (except yeah you don't get the nice indentation but I could live without that I think)

anthrotype:
basically I am suggesting that we just use continue using the existing features (globality, singleton-ness) that built-in logging machinery already provides

anthrotype:
To recap, we don't need any GlobalTimer, loggers are already global (logging.getLogger("name.of.logger") returns the same thing no matter where you call it).
If we want to add timing info in libraries like ufo2ft or fonttools similarly to how we already have in fontmake, we simply need to:

1) use the `fontTools.misc.loggingTools.Timer` object as usual and attach a logger with a name that fontmake can identiy as a logging the time, e.g. the fonttools subsetter does:
  
        from fontTools.misc.loggingTools import Timer

        timer = Timer(logger=logging.getLogger("fontTools.subset.timer"))
        ...
        with timer("load '%s'" % tag):
	        table = font[tag]
        # which emits a message like "Took 0.4s to load 'glyf'"

    in ufo2ft you could call the logger "ufo2ft.timer" for simplicity.

2) libraries (e.g. ufo2ft) can define loggers and log messages with them, but nothing happens if the same loggers are not configured by some applications (fontmake) using these libraries; configuring a logger means defining the verbosity level and attaching a handler (e.g. console or file handler) to a given logger (and automatically to all hiearachy of children with the namespace prefix); logging.basicConfig configures the root logger (which is the last resort where all unhandled logging messages go) and can only be used once per session (doesn't do anything if the root logger is already configured); fontTools.misc.loggingTools has a `configLogger` function that does something similar to basicConfig but for any logger, not just the root logger (and can be called multiple times on the same logger). You want to use that configLogger to configure the timing loggers of all the libraries used by fontmake that emit timing logs, in the exact same way fontmake configures its own timer logger. Here:

https://github.com/googlefonts/fontmake/blob/42feea769bbb8413dc8f81cbb4414207461e5534/Lib/fontmake/font_project.py#L121-L122

You'd basically add more `configLogger` calls for each of the timer-loggers you want to configure, using the same logging level as the fontmake timer. E.g.

    configLogger(logger="ufo2ft.timer", level=logging.DEBUG)
    configLogger(logger="fontTools.varLib.timer", level=logging.DEBUG)
    # etc.?

This is how python logging works, separation of concern between libraries defining loggers and emitting messages at given level of verbosity, and applications deciding how to configure each logger and what messages to filter and how to format the output and where to write the messages, etc.

PR #2704, opened 2022-07-19
[feaLib] Add vmtx and VORG support

punchcutter:
This adds vmtx support to feaLib. And to match makeotf this also fills out the VORG table for CFF or CFF2 fonts. Unlike everything else in feaLib this requires access to the outlines to get the bounds which can be expensive which is one reason to do it all in one go. Since we are only reading overrides here instead of building a vmtx from scratch it is assumed that the vmtx table already exists (probably from ufo2ft in most cases). This is one approach, but another possible approach would be to add builders to otlLib like `buildStatTable` where users could also build a vmtx and/or VORG from scratch if necessary. If that's more desireable I can change the PR, but this feaLib only setup is working well.

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -1923,6 +1923,25 @@ class VheaField(Statement):
         return "{} {};".format(keywords[self.key], self.value)
 
 
+class VmtxStatement(Statement):
+    """An entry in the ``vmtx`` table."""
+
+    def __init__(self, glyph, key, value, location=None):
+        Statement.__init__(self, location)
+        self.glyph = glyph
+        self.key = key
+        self.value = value
+
+    def build(self, builder):
+        """Calls the builder object's ``add_vmtx_field`` callback."""
+        builder.add_vmtx_pos(self.key, self.glyph, self.value)
+
+    def asFea(self, indent=""):
+        fields = ("VertOriginY", "VertAdvanceY")
+        keywords = dict([(x.lower(), x) for x in fields])
+        return "{} {} {};".format(keywords[self.key], self.glyph, self.value)
+
+
 class STATDesignAxisStatement(Statement):
     """A STAT table Design Axis
 
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -1,4 +1,5 @@
 from fontTools.misc import sstruct
+from fontTools.misc.roundTools import otRound
 from fontTools.misc.textTools import Tag, tostr, binary2num, safeEval
 from fontTools.feaLib.error import FeatureLibError
 from fontTools.feaLib.lookupDebugInfo import (
@@ -103,6 +104,7 @@ class Builder(object):
             "hhea",
             "name",
             "vhea",
+            "vmtx",
             "STAT",
         ]
     )
@@ -172,6 +174,8 @@ class Builder(object):
         self.hhea_ = {}
         # for table 'vhea'
         self.vhea_ = {}
+        # for table 'vmtx'
+        self.vmtx_ = {}
         # for table 'STAT'
         self.stat_ = {}
         # for conditionsets
@@ -205,6 +209,8 @@ class Builder(object):
             self.build_name()
         if "OS/2" in tables:
             self.build_OS_2()
+        if "vmtx" in tables:
+            self.build_vmtx()
         if "STAT" in tables:
             self.build_STAT()
         for tag in ("GPOS", "GSUB"):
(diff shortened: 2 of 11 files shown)


punchcutter:
One reason this came up is that while building a CFF2 and TTF variable font the CFF2 has VORG variation in the VVAR, but the TTF doesn't need that so copying the tables over copies irrelevant information. It makes more sense to build the vmtx here (and then VVAR gets built off that) when making a TTF variable.

anthrotype:
I haven't started reviewing this, but I have a perhaps naive question about the workflow 

> Since we are only reading overrides here instead of building a vmtx from scratch it is assumed that the vmtx table already exists

why does one need to "override" vmtx/VORG? Can't the compiler build those tables from source (UFO glifs) in the desired way without needing to "override" them in the features?
I understand that makeotf already allows that, and the vmtx table overrides are already part of the AFDKO Feature File spec and thus makes sense to implement here as well, and that you may have feature files with these that you also want to be able to build using fonttools instead of makeotf. But really what's their actual use? In an ideal world, one wouldn't need to override anything, just build. What's lacking/incomplete/incorrect in the current UFO=>{O,T}TF build pipeline that makes one want to define "overrides" for vmtx/VORG?

Also, the requirement that in order to build vmtx/VORG the outline tables be present is a but of a problem, if we would like in the future to build the OT features in parallel with building the outlines, since it creates a dependency between the two. If that could be avoided, the better.

Note, I'm not rejecting this outright, I'm simply trying to understand why it's needed. Sure, we can have it for interoperability with makeotf, but maybe there's a better way.

punchcutter:
Sure, maybe there's a better way in the grand scheme of things, but to match what's in the FEA spec this is how it works. Why override? That's like "why kern?" when you already have space between glyphs. Why spend all that time changing it? Because it looks better. From most fonts I've seen that do this there's not a lot of overriding going on. In Source Han Serif/Noto CJK there are around 250 overrides out of 65,535 glyphs. However, in another current project every glyph (thousands of them) needs an override because of the specifics of the design and vertical variation. 

Anyway, it might work to use `public.verticalOrigin` in the glif files (probably needs some updates in ufoLib2/ufo2ft, etc), but that's kind of irrelevant to this issue. I could also write a Python script to change it all in post. Or dump the ttx and edit that. Or any number of other options that have nothing to do with feaLib.

anthrotype:
Perhaps my question sounds a bit rethorical, what I meant was why does one need to "override" something in FEA instead of directly setting the advance height and public.verticalOrigin in the UFO glifs. I ask because I'm interested in knowing whether the regular UFO=>OTF build workflow is lacking something that one can only accomplish by using such FEA overrides.

punchcutter:
Well, the original sources for lots of things was never UFO so there was no way to set these things in the source file. Which is why the spec has vmtx. Since feaLib and this whole workflow _is_ based on UFO then it might be better to update sources rather than rely on feaLib. I believe the current build workflow is lacking something required to do this. ufoLib2 has some code to read public.verticalOrigin, but I don't see it being used anywhere. I can't say for sure if that's all that's required right now for this to work correctly.

anthrotype:
we are digressing, but ufo2ft can build vmtx and VORG, see

https://github.com/googlefonts/ufo2ft/blob/7d0a330626da6e4999349f3f416bfa508b2826ef/Lib/ufo2ft/outlineCompiler.py#L854-L90/4

If you could give it a try some time and report any issues, that would be lovely.

Regardless, we probably want to get support for these overrides in feaLib anyway given makeotf and FEA spec allow it.
I hope to be able to review this at some point this week, thank you Zachary.

punchcutter:
Yes, ufo2ft builds it, but it doesn't read public.verticalOrigin. And what's built anyway is incorrect in a couple places. This PR makes it so that the end result is the same as we get with makeotf.

anthrotype:
yes it does read verticalOrigin

https://github.com/googlefonts/ufo2ft/blob/7d0a330626da6e4999349f3f416bfa508b2826ef/Lib/ufo2ft/outlineCompiler.py#L64-L71

anthrotype:
> what's built anyway is incorrect in a couple places

we should file specific issues on the ufo2ft repo, I want to make its output correct and matching makeotf for vertical metrics tables.

punchcutter:
Yes, everything I've been doing is with `ufo2ft.compileVariableTTF`. I didn't see `public.verticalOrigin` there and it seemed like it wasn't using that, but I only just found that as an option. I can try to set that and see what happens, but all of this is really the lowest possible priority for me right now.

anthrotype:
> this is really the lowest possible priority for me right now.

no worries, thanks

punchcutter:
Ok, did a quick test and so far it seems to work if I set public.verticalOrigin in all the glifs. The defaultVerticalOrigin in VORG is different which I noted in this PR:

> For the VORG table ufo2ft finds the most frequent
> verticalOrigin and uses that to set defaultVertOriginY, but
> makeotf uses the OS/2.TypoAscender as stated in the
> OpenType Feature File Specification 9.h

anthrotype:
thanks for checking.
Are you sure regarding ufo2ft not defaulting to the OS/2 TypoAscender?

I found this: https://github.com/googlefonts/ufo2ft/pull/343

anthrotype:
though it is true that ufo2ft sets the VORG.defaultVertOriginY to the most frequent verticalOrigin. @moyogo contrinbuted that code in https://github.com/googlefonts/ufo2ft/pull/56
@punchcutter if you think that's incorrect and should be changed, please file an issue on ufo2ft and we can discuss about that

punchcutter:
That's the vertical origin, not the VORG defaultVertOriginY https://github.com/googlefonts/ufo2ft/blob/7d0a330626da6e4999349f3f416bfa508b2826ef/Lib/ufo2ft/outlineCompiler.py#L897

punchcutter:
Right, I can open an issue on ufo2ft for that.

anthrotype:
IIUC the vertical origins are actually the same, only the defaultVerticalOriginY is computed differently. I think using the most frequent one makes sense if you want to optimize the size of the VORG table

punchcutter:
I'm only talking about the default in VORG. Is the purpose of defaultVerticalOriginY only to optimize the VORG table? 
Using the most frequent doesn't necessarily seem correct to me. Just because a bunch of glyphs have a specific shared vertical origin value doesn't mean it's necessarily the right one for default. In my specific test case the default value gets set to something quite different from what's in the OS/2, but for this font every glyph is explicitly set anyway so the VORG defaultVerticalOriginY doesn't even get used. For Source Han/Noto CJK the most frequent and OS/2 would be the same so it wouldn't matter. In practice I don't know how much this value really gets used when OS/2 and most frequent value differ.

anthrotype:
> Is the purpose of defaultVerticalOriginY only to optimize the VORG table?

I think so, yes. The spec defines it as (emphasis mine)

> The y coordinate of a glyph’s vertical origin, in the font’s design coordinate system, to be _used if no entry is present for the glyph_ in the vertOriginYMetrics array.

and then it proceeds saying

> In a size-optimized implementation, glyphs whose vertical origin’s y coordinate equals defaultVertOriginY will not have an entry in this array. 
> If all glyphs in a font share the same defaultVertOriginY value, the length of the VORG table will be 8 bytes in a size-optimized implementation, since the vertOriginYMetrics array will be absent.

There's two considerations/questions here:

1) whether the individual glyph vertical origins in the VORG tables produced by ufo2ft (from UFO public.verticalOrigin lib key) and makeotf (by whatever means you do that in makeotf, presumably vmtx features.fea table override) match with one another. If so, the font are technically equivalent and produce same vertical glyph positioning. 
2) the implications of the difference b/w ufo2ft and makeotf as regards the choice of defaultVerticalOriginY (as defined above); if the VORG table produced by ufo2ft's choosing the most frequent verticalOrigin is smaller in size than one that simply uses the OS/2.sTypoAscender, then I would argue the different encoding of the VORG table that this entails is acceptable in light of what the spec calls "size-optimized implementation" and ufo2ft should not be changed to use TypoAscender as defaultVerticalOriginY simply to match maketof.

punchcutter:
Ok, so there's nothing to do here and this PR can be closed.

anthrotype:
Don't close it just yet. Supporting the full extent of the FEA spec, including vmtx table overrides, is desirable, there may be old sources that already contain this data and may take advantage of this feature. I will get to review this at some point.

NightFurySL2001:
Maybe instead of overriding the current values, the FEA vmtx syntax will be used when there is *no* vmtx table? This would make the feature syntax more useful by providing values in a centralised location instead of repeating multiple values over all glif files. (quite wasteful considering fonts with vmtx are CJK fonts, which generally contains 5k-10k+ glyphs, and all of the glifs require repeating the advance height and verticalOrigin in lib)

Do note that the original assumption and community standard expected by Adobe FEA syntax — where `sTypo` metric values will align with em-box — is broken by the new Google Fonts CJK vertical metrics requirements. See https://googlefonts.github.io/gf-guide/metrics.html#cjk-vertical-metrics and discussion raised by @aaronbell in https://github.com/google/fonts/issues/8911 . An Adobe FEA syntax update will be required for such cases.

behdad:
Would be nice to revive whatever bits of this that are still relevant.

PR #2711, opened 2022-07-28
Compile a couple of modules with Cython when available

behdad:
Even with no annotations, makes varLib.iup ~35% faster, and varLib.interpolatable 3x faster (by way of pens.momentsPen).


behdad:
I've pushed a couple small other commits here as well. Let me know if you prefer them as separate PRs.

anthrotype:
Thanks! I think the last two commits (where you add support for reading .designspace and .glyphs files in varLib.interpolatable tool) could be split to a separate PR ideally

anthrotype (approved):
LGTM, except the last two commits that could be split to a spearate PR (for visibility mostly)

behdad:
> Thanks! I think the last two commits (where you add support for reading .designspace and .glyphs files in varLib.interpolatable tool) could be split to a separate PR ideally

https://github.com/fonttools/fonttools/pull/2716

Issue #1959, opened 2020-05-15
design for a new processor or dispatch module implementing visitor pattern


anthrotype:
This has come up a few times, e.g. https://github.com/fonttools/fonttools/pull/1668#issuecomment-628970150 or https://github.com/fonttools/fonttools/issues/1951#issuecomment-628174958

quoting @behdad

> I'm also thinking about a new modules (possibly called processor or dispatch) that will improve on and centralize the programming scheme used in the merge, subset, and varLib modules. That will simplify a lot of things.

anthrotype:
In https://github.com/fonttools/fonttools/pull/1668#issuecomment-629173201, I mentioned https://github.com/mrocklin/multipledispatch/.

There is also another one called [methoddispatch](https://github.com/seequent/methoddispatch) that provides a `singledispatch` decorator (similar to the stdlib one) but designed to work with instance methods. This one also supports inheritance, by ensuring that each subclass has it's own independant copy of the dispatch registry

> Subclasses can extend the type registry of the function on the base class with their own overrides.

justvanrossum:
I don't like the limitation of the above-mentioned methoddispatch that specific implementations need to have distinct names.

Perhaps an external dependency isn't needed, as it's not super hard to just implement it ourselves. Here's a naive try that seems to work:

```python
def instancedispatch(generic_func):
    registry = {}
    def register(tp):
        def deco(func):
            registry[tp] = func
            return wrapper
        return deco
    def wrapper(self, arg0, *args, **kwargs):
        # TODO: traverse type(arg0).__mro__ instead of only looking at type(arg0)
        func = registry.get(type(arg0), generic_func)
        return func(self, arg0, *args, **kwargs)
    wrapper.register = register
    return wrapper


class A:

    @instancedispatch
    def myfunc(self, arg):
        print("A generic arg:", arg)

    @myfunc.register(int)
    def myfunc(self, arg):
        print("A int arg:", arg)


class B:

    @instancedispatch
    def myfunc(self, arg):
        print("B generic arg:", arg)

    @myfunc.register(int)
    def myfunc(self, arg):
        print("B int arg:", arg)

    
a = A()
a.myfunc("abc")
a.myfunc(123)

b = B()
b.myfunc("def")
b.myfunc(456)
```

anthrotype:
[...]

PR #2718, opened 2022-07-31
Add fontTools.misc.visitor, fontTools.ttLib.ttVisitor, Snippets/scale-upem.py, Snippets/print-json.py

behdad:
Fixes https://github.com/fonttools/fonttools/issues/1959


Lib/fontTools/misc/visitor.py
```diff
@@ -0,0 +1,93 @@
...
+			ret = visitorFunc(self, obj, *args, **kwargs)
+			if ret == False or (ret is None and self.defaultStop):
+			  return
+		if hasattr(obj, '__dict__'):
+			self.visitObject(obj, *args, **kwargs)
+		elif isinstance(obj, list):
+			self.visitList(obj, *args, **kwargs)
+
+
+class TTVisitor(Visitor):
```
anthrotype, line 78: maybe TTVisitor could be defined inside ttLib.ttFont module, next to TTFont class. Usually `misc` package contains things that are generic, not necessarily font-related

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -381,11 +381,13 @@ def keys(self):
...
-	def ensureDecompiled(self):
+	def ensureDecompiled(self, recurse=None):
 		"""Decompile all the tables, even if a TTFont was opened in 'lazy' mode."""
 		for tag in self.keys():
 			table = self[tag]
-			if self.lazy is not False and hasattr(table, "ensureDecompiled"):
+			if recurse is None:
+				recurse = self.lazy is not False
+			if recurse and hasattr(table, "ensureDecompiled"):
 				table.ensureDecompiled()
```
anthrotype, line 391: table.ensureDecompiled also has a `recurse=True` parameter, maybe you wanted to pass that on when you call it? i.e. `table.ensureDecompiled(recurse=recurse)`.

https://github.com/fonttools/fonttools/blob/9716b1c4556642f92d66d8ceee0225e6c5ade8bb/Lib/fontTools/ttLib/tables/otBase.py#L227-L228

It doesn't seem like you're calling `TTFont.ensureDecompiled` method with a `recurse` value other than `None`. I wonder it's actually needed.

Tests/misc/visitor_test.py
```diff
@@ -0,0 +1,37 @@
...
+        v = getattr(obj, attr)
+        v = visitor.scale(v)
+        setattr(obj, attr, v)
+    return False
+
+
+from fontTools.ttLib import TTFont
+import sys
+
+font = TTFont(sys.argv[1])
```
anthrotype, line 34: the file is called `visitor_test.py` (with a `_test` suffix) so pytest assumes it is a test module, but it's actually a script that takes argv, hence pytest run fails because as it tries to import the test module. Wrap that into a `if __name__ == "__main__"` or something if you'd like to keep that way

Lib/fontTools/misc/visitor.py
```diff
@@ -0,0 +1,93 @@
...
+	def visitList(self, obj, *args, **kwargs):
+		for value in obj:
+			self.visit(value, *args, **kwargs)
+
+	def visit(self, obj, *args, **kwargs):
+		visitorFunc = self._visitorsFor(obj).get(None, None)
+		if visitorFunc is not None:
+			ret = visitorFunc(self, obj, *args, **kwargs)
+			if ret == False or (ret is None and self.defaultStop):
+			  return
```
anthrotype, line 71: indentation looks off here, perhaps a mix of tabs and spaces. In general, I strongly recommend you just run `black` to autoformat this code (at least do for new modules). I know you like to hand-craft your code, but sometimes the odd formatting can get distracting )and we all have better things to do than argue about code style.. ;)

Lib/fontTools/misc/visitor.py
```diff
@@ -0,0 +1,93 @@
...
+		if hasattr(obj, '__dict__'):
```
anthrotype, line 72: in varLib.merger I had to special-case instances of enum.Enum (COLR tables use them e.g. CompositeMode) otherwise I would incur in a RecursionError; not sure if it's still relevant here, but I suggest you test if things still work as expected when an object tree contains some enum. Basically enums have a `__dict__` but we want to treat them as individual values, not collections of attributes

https://github.com/fonttools/fonttools/blob/9716b1c4556642f92d66d8ceee0225e6c5ade8bb/Lib/fontTools/varLib/merger.py#L133-L137

anthrotype (commented):
thanks for working on this! More tests needed but I like where this is going

behdad:
Thanks @anthrotype. This is currently a sketch... I expect to reiterate on it, and then put things in their proper places.

Lib/fontTools/ttLib/ttFont.py
behdad: cmap's ensureDecompiled doesn't take recurse, but yeah I think I'll fix that one too.

This change ended up being unused but I kept it. I'll probably put back recurse=False to the visitor.

Tests/misc/visitor_test.py
behdad: I've renamed that to `Snippets/scale-upem.py` now.

behdad:
Okay I think this is useful-enough to be merged now.

The `Snippets/scale-upem.py` added here already handles variable fonts as well.

behdad:
I've added a fairly functional `print-json.py` snippet as well. Python is able to parse back the json and I'm able to do things like:
```
j['glyf']['glyphs']['uni0040']['coordinates'][3][0]
```
which is the equivalent of:
```
font['glyf'].glyphs['uni0040'].coodinates[3][0]
```

behdad:
> I've added a fairly functional `print-json.py` snippet as well. Python is able to parse back the json and I'm able to do things like:

Doesn't work for CFF... We need to add `ensureDecompiled` there.

behdad:
Okay. Tests also added with 100% coverage.

m4rc1e:
This is great!

Could you add the scaler as a module to https://github.com/fonttools/fonttools/tree/main/Lib/fontTools like we do for pyftsubset and merge?

I plan on using this to diff fonts which have different upms.

Tests/ttLib/ttVisitor_test.py
```diff
@@ -0,0 +1,18 @@
...
+    @staticmethod
+    def getpath(testfile):
+        path = os.path.dirname(__file__)
+        return os.path.join(path, "data", testfile)
+
+    def test_ttvisitor(self):
+
+        font = TTFont(self.getpath("TestVGID-Regular.otf"))
+        visitor = TTVisitor()
+        visitor.visit(font, 1, 2, arg=3)
```
anthrotype, line 18: it's not immediately clear what this test is actually testing (what are those numbers), especially as I don't see any assertions.

Snippets/scale-upem.py
```diff
@@ -0,0 +1,156 @@
...
+
+    if len(sys.argv) != 3:
+        print("usage: scale-upem.py font new-upem")
+        sys.exit()
+
+    font = TTFont(sys.argv[1])
+    new_upem = int(sys.argv[2])
+
+    if 'CFF ' in font or 'CFF2' in font:
+        print("scale-upem: CFF/CFF2 fonts are not supported.", file=sys.stderr)
```
anthrotype, line 147: we should also add support for scaling COLR table fields that are font units (e.g. gradient coordinates. etc.), perhaps in a follow up PR 👍

Snippets/scale-upem.py
```diff
@@ -0,0 +1,156 @@
+from fontTools.ttLib.ttVisitor import TTVisitor
```
anthrotype, line 1: could you add a small module docstring at the top briefly summarizing what this scale-upem script does (and does not) currently support?

Snippets/print-json.py
```diff
@@ -0,0 +1,153 @@
+import fontTools.ttLib as ttLib
+from fontTools.ttLib.ttVisitor import TTVisitor
+from fontTools.misc.textTools import Tag
+from array import array
+
+
+class JsonVisitor(TTVisitor):
```
anthrotype, line 7: ah, this is fun. 
curious, why didn't you build a raw dict and pass it on to json library to serialize and opted for printing strings directly? For the sake of this snippet either is ok.

Lib/fontTools/ttLib/ttVisitor.py
```diff
@@ -0,0 +1,27 @@
+"""Specialization of fontTools.misc.visitor to work with TTFont."""
+
+from fontTools.misc.visitor import Visitor
+from fontTools.ttLib import TTFont
+
+
+class TTVisitor(Visitor):
+    def visitAttr(self, obj, attr, value, *args, **kwargs):
+        if isinstance(value, TTFont):
+            return False
```
anthrotype, line 10: why is that? A comment might help

Lib/fontTools/ttLib/ttVisitor.py
```diff
@@ -0,0 +1,27 @@
...
+    def visit(self, obj, *args, **kwargs):
+        if hasattr(obj, "ensureDecompiled"):
+            obj.ensureDecompiled(recurse=False)
+        super().visit(obj, *args, **kwargs)
+
+
+@TTVisitor.register(TTFont)
+def visit(visitor, font, *args, **kwargs):
+    if hasattr(visitor, "font"):
+        return False
```
anthrotype, line 22: this is to avoid recursion errors, right?

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -112,7 +112,7 @@ def decompile(self, data, ttFont):
 		if ttFont.lazy is False: # Be lazy for None and True
 			self.ensureDecompiled()
 
-	def ensureDecompiled(self):
+	def ensureDecompiled(self, recurse=None):
```
anthrotype, line 115: i think you want it to be False, not None (the latter value for recurse parameter is only relevant for TTFont which has the lazy attribute)

```suggestion
	def ensureDecompiled(self, recurse=False):
```

Also adding a comment as to why there's this parameter which is not even used inside this specific glyf table method would help our future selves

Lib/fontTools/ttLib/tables/_c_m_a_p.py
```diff
@@ -164,7 +164,7 @@ def decompile(self, data, ttFont):
 		if ttFont.lazy is False:  # Be lazy for None and True
 			self.ensureDecompiled()
 
-	def ensureDecompiled(self):
+	def ensureDecompiled(self, recurse=None):
```
anthrotype, line 167: like for glyf, I think you want `recurse=False`, not `None` (see my other comment related to glyf.ensureDecompiled)

Lib/fontTools/misc/visitor.py
```diff
@@ -0,0 +1,114 @@
...
+    def visitLeaf(self, obj, *args, **kwargs):
+        pass
+
+    def visit(self, obj, *args, **kwargs):
+        visitorFunc = self._visitorsFor(obj).get(None, None)
+        if visitorFunc is not None:
+            ret = visitorFunc(self, obj, *args, **kwargs)
+            if ret == False or (ret is None and self.defaultStop):
+                return
+        if hasattr(obj, "__dict__") and not isinstance(obj, enum.Enum):
```
anthrotype, line 107: thanks for special-handling enum.Enum. Mind adding a small test that visiting one such value works and doesn't end up tangled in recursion errors?

Lib/fontTools/misc/visitor.py
```diff
@@ -0,0 +1,114 @@
...
+            self.visit(value, *args, **kwargs)
+
+    def visitLeaf(self, obj, *args, **kwargs):
+        pass
+
+    def visit(self, obj, *args, **kwargs):
+        visitorFunc = self._visitorsFor(obj).get(None, None)
+        if visitorFunc is not None:
+            ret = visitorFunc(self, obj, *args, **kwargs)
+            if ret == False or (ret is None and self.defaultStop):
```
anthrotype, line 105: whe should document more prominently (in the Visitor class docstring maybe) that there special return values/options (False, None,  self.defaultStop) that let one interrupt the traversal when some object is visited, or by default stop as soon as a custom visitor has returned

anthrotype: I also wonder if, instead of generic False or None, we could use an enum (perhaps defined as a nested class inside Visitor itself so it's readily available to subclasses) which would give these a more explicit meaning

Lib/fontTools/misc/visitor.py
```diff
@@ -0,0 +1,114 @@
+"""Generic visitor pattern implementation for Python objects."""
+
+import enum
+
+
+class Visitor(object):
+
+    defaultStop = False
```
anthrotype, line 8: would be good to have an explicit test to cover the use case of defaultStop = True

Lib/fontTools/misc/visitor.py
```diff
@@ -0,0 +1,114 @@
...
+            if _visitors is None:
+                break
+
+            m = celf._visitors.get(typ, None)
+            if m is not None:
+                return m
+
+        return _default
+
+    def visitObject(self, obj, *args, **kwargs):
```
anthrotype, line 72: would be good to document as well that the `*args, **kwargs` are forwarded to the custom visitor implementations

anthrotype (commented):
this is very useful, thanks Behdad! I left some comments, PTAL

Tests/ttLib/ttVisitor_test.py
behdad: It just runs through it, that it doesn't crash. :))

behdad: The visitor.py is fully tested in the other module. The TTVisitor is just very short module. I'll add more test to it.

Those numbers just test that we can pass arbitrary arguments to the TTVisitor and it passes them through.  I'll add more tests here.

Snippets/scale-upem.py
behdad: And BASE table might need some work. Should do an inventory of all tables. Like subset module, we should warn on other tables we don't know how to scale that we know have coordinate data (like AAT, Graphite, ...)

Snippets/print-json.py
behdad: Right. That didn't occur to me. This was more of an inspector fun piece of code for debugging purposes kind of thing.

Lib/fontTools/ttLib/tables/_c_m_a_p.py
behdad: I used None because the argument is unused.

behdad:
Thanks for the review Cosimo. I'll address them soon.

Lib/fontTools/ttLib/ttVisitor.py
behdad: Yes.

Lib/fontTools/misc/visitor.py
behdad: I like None because that's what a function implicitly returns. And returning False from a visitor to stop iteration is common practice IMO. I kept it as is. I can change if you have strong opinion.

behdad:
> Could you add the scaler as a module

Done. Need better API, etc. But you can use it already.

behdad:
@anthrotype PTAL.

Lib/fontTools/misc/visitor.py
```diff
@@ -85,20 +94,40 @@ def visitObject(self, obj, *args, **kwargs):
             self.visitAttr(obj, key, value, *args, **kwargs)
 
     def visitAttr(self, obj, attr, value, *args, **kwargs):
+        """Called to visit an attribute of an object."""
         self.visit(value, *args, **kwargs)
 
     def visitList(self, obj, *args, **kwargs):
+        """Called to visit any value that is a list."""
```
anthrotype, line 101: i wonder if we ever need a visitTuple as well.. well, I guess one can always register a custom visit function for tuples if needed.

Lib/fontTools/scaleUpem.py
```diff
@@ -143,14 +143,14 @@ def visit(visitor, varData):
     import sys
 
     if len(sys.argv) != 3:
-        print("usage: scale-upem.py font new-upem")
+        print("usage: fonttools scaleUpem font new-upem")
```
anthrotype, line 156: maybe fontTools.ttLib.scaleUpem would be a better place than at the top package level, it feels like a special tool that works with the ttLib api more than its own separate "lib"

anthrotype (approved):
LGTM

Lib/fontTools/misc/visitor.py
behdad: I figured tuples are more often a leaf value than a container. But yeah, they can be registered as a type.

Lib/fontTools/scaleUpem.py
behdad: I'll move there if you like. I figured other tools/modules like like subset, merge, fontBuilder are also ttLib-specific but in toplevel.

Lib/fontTools/ttLib/scaleUpem.py
```diff
@@ -141,6 +142,51 @@ def visit(visitor, varData):
...
+def visit(visitor, record):
+    oldPaint = record.Paint
+
+    transform = otTables.Affine2x3()
+    transform.populateDefaults()
+    transform.xy = transform.yx = transform.dx = transform.dy = 0
+    transform.xx = transform.yy = visitor.scaleFactor
+
+    scale = otTables.Paint()
+    scale.Format = 12 # PaintTransform
```
anthrotype: you could use a PaintScaleUniform
https://github.com/fonttools/fonttools/blob/d275207dbc75d0dbe691f6f1c9fe3c9ffe01da5f/Lib/fontTools/ttLib/tables/otData.py#L1850-L1854

behdad: They use F2Dot14...

behdad: I can conditionally use it. But the transform is fine because all of them share it via offset, so it's only 1 byte worse than a PaintScaleUniform.

behdad: I first used it and I hit an out-of-range error at compile-time...

behdad: Working on it.

anthrotype: right, I had forgotten

behdad: https://github.com/fonttools/fonttools/commit/0c6b67e8eca0dbe4baa1038082ca473802e5123b

Issue #2339, opened 2021-06-16
[symfont] Autogenerating certain additional GreenPens dies


n8willis:
I've been using the MomentsPen `printGreenPen()` to generate some additional higher-order moments of interest. Although they're never "pretty" to look at, I found at least one func that defeats either symfont or sympy when I try to solve it. I'm not sure if I'm doing something wrong, or it's just undoable.

The tree with works-as-expected additions is here, if anybody cares to peruse: https://github.com/n8willis/fonttools/blob/moremoments/Lib/fontTools/pens/momentsPen.py , although it's not particularly different. Basically, calling printGreenPen for some additional moment integrals, like 
```
('momentXXXX', x**4),
('momentXXYY', x*x*y*y),
('momentPolar4', 2*x*x*y*y+x**4+y**4),
```
works fine, even if it takes a couple of minutes. But out of curiosity, I tried it on the 3rd-order-polar moment too; e.g.:
```
('momentPolar3', sp.sqrt(x**2+y**2)*(x**2+y**2)),
```

And it crashes the Jupyter kernel. Rewriting that func as `((x**2+y**2)**0.5)*(x**2+y**2))`, Jupyter stays with us, but it eventually kills sympy with this:
```
[...]

PR #2727, opened 2022-08-05
[WIP] Higher-moments pens

n8willis:
Work-in-progress on adding the higher moments to MomentsPen and StatisticsPen.

Right now it only has 2nd and 4th Polar moments.

It's looking to me like the odd-numbered Polar moments are not doable without changing `symfont`.

As in the [discussion on 2339](https://github.com/fonttools/fonttools/issues/2339#issuecomment-871592041), the original integral is all hyperbolic & trig & logs and doesn't really simplify. There might be numerical approximations for those functions that would work, but likely it'd either mean changing GreenPen or adding another one. (I haven't actually tried; that's just the current state and what I think from poking at it.)


behdad:
I've moved the definition of OpenContourError to basePen. Can you update that please?

behdad:
Humm. This needs rebasing on top of main since i changed symfont and momentsPen to use Cython.

behdad:
I can look into it in a few days if you don't feel like it.

n8willis:
> Humm. This needs rebasing on top of main since i changed symfont and momentsPen to use Cython.

> I've moved the definition of OpenContourError to basePen. Can you update that please?

Ahh; sure thing. That's just a side-effect from me not checking back in so frequently. Presumably that applies to the existing MomentsPen PR, too.

behdad:
> > Humm. This needs rebasing on top of main since i changed symfont and momentsPen to use Cython.
> 
> > I've moved the definition of OpenContourError to basePen. Can you update that please?
> 
> Ahh; sure thing. That's just a side-effect from me not checking back in so frequently. Presumably that applies to the existing MomentsPen PR, too.

Yes.

n8willis:
> > > Humm. This needs rebasing on top of main since i changed symfont and momentsPen to use Cython.
> > 
> > 
> > > I've moved the definition of OpenContourError to basePen. Can you update that please?
> > 
> > 
> > Ahh; sure thing. That's just a side-effect from me not checking back in so frequently. Presumably that applies to the existing MomentsPen PR, too.
> 
> Yes.

Yeah; I'll try and do this tonight. Been occupied doing other absurd things like DFTs of glyphs (which thankfully will never end up an Pens....).

behdad:
> Been occupied doing other absurd things like DFTs of glyphs

Interesting. I looked at that before as well. Didn't get anywhere useful. I know several people did in the context of kerning. Raph being one of those.

n8willis:
> Interesting. I looked at that before as well. Didn't get anywhere useful. I know several people did in the context of kerning. Raph being one of those.

Yeah, it doesn't really do anything practical (more importantly, nothing that can't also be done through simpler means). More of a way of thinking about space than finding it. But I do need to spend pages saying that and showing how it compares to other concepts that people talk about but are more practical.

PR #2728, opened 2022-08-07
[instancer] Towards L4 instancing

behdad:
Towards L4 instancing.

This does 2 things:

- Mess up `__init__` to pass triples instead of floats and min:max for every axis all the time...

- Add a `instancer.solver` module to do the heavy lifting.

The solver does _not_ currently implement L4 instancing. The algorithm is yet to be implemented. But the API is there so the `__init__` part can be fixed up to use it properly and undo my butchering and fix the tests.


behdad:
The tent-to-tent case handling is excruciating... but I think I will manage it.

behdad:
> The tent-to-tent case handling is excruciating... but I think I will manage it.

This is done and tested now.

anthrotype:
This is great! I'll take a look at this later this week. By the way, I love the ASCII art 👍

behdad:
The solver modules should be moved to varLib.models.

behdad:
> The solver modules should be moved to varLib.models.

Maybe not. Happy with where it is for now. We can always move it later.

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -785,7 +779,7 @@ def _instantiateFeatureVariations(table, fvarAxes, axisLimits):
             record, i, axisLimits, fvarAxes, axisIndexMap
         )
         if shouldKeep:
-            shouldKeep = _limitFeatureVariationRecord(record, axisLimits, axisOrder)
```
anthrotype: @behdad I don't understand this change, can you elaborate what is this "small fix" actually fixing? I think `_limitFeatureVariationRecord` should be given the opportunity to influence whether the record should be kept, that is if any condition is not met because outside the desired range. I think right now you're setting ConditionTable to None, but that's supposed to be a list..

behdad: Humm. You are right that this might be wrong still. My thinking was that an empty condition-set also means "always apply", and should not cause the variation record to be removed.

behdad: I think now `_instantiateFeatureVariationRecord` drops any record that is completely irrelevant. So I think the change is correct still. Remove this change and see the test failure.

anthrotype: in this case, what's the effect of keeping that empty condition-set that always applies if we have already applied the feature variation as the default feature record? We already swap {"uni0041": "uni0061"}, so what's the use of the empty alwats-apply condition that replaces the same {"uni0041": "uni0061"}?

anthrotype: In any case, as it currently stands in this PR, `_limitFeatureVariationRecord` has a couple issues:

- a couple comments are stale: "remapped axis index" (it no longer does that, because `_instantiateFeatureVariationRecord` already did before it); "remove entire record" (it doesn't do that either anymore, but that's under discussion in this thread)
- setting `record.ConditionSet.ConditionTable = newConditions` when `newConditions` might be None,  see that `else` branch which does `newConditions = None`; I think if you're going to keep the `ConditionTable`, the latter is supposed to be a list, empty if you like, but not a `None` value. Besides, that very line does not seem covered by the current test, but I also wonder if it can be _ever_ hit at all (I put an `assert 0` and tried a bunch of axisLimits but couldn't trigger it).

behdad: > * setting `record.ConditionSet.ConditionTable = newConditions` when `newConditions`

I thought that's fine. It is in HarfBuzz land. But might not be in OpenType land, you are right.





> in this case, what's the effect of keeping that empty condition-set that always applies if we have already applied the feature variation as the default feature record?

Humm. You are right, almost. Unless this rules had the effect of stopping further rules after it, in which case to preserve the logic we need to preserve the rule. Doe you understand what I mean?

behdad: Ah, I checked the code and you are right, the previous logic was right and I'm totally confused here.

behdad: Well not exactly. I just need to filter out None. Empty list should still be allowed.

behdad: I did that now.

behdad: https://github.com/fonttools/fonttools/commit/95a0fb4a0f933fa3bf0bfc8a60fa68fe12c9e6f7

anthrotype: > You are right, almost. Unless this rules had the effect of stopping further rules after it, in which case to preserve the logic we need to preserve the rule. Doe you understand what I mean?

no, but I will sleep over it and return with a fresh mind tomorrow

behdad: Okay I think I got it: If a rule becomes always-applicable (because the new designspace is a subset of its condition-set), then it should be lifted to default, *and all rules after it should be discarded*.

behdad: Let's make this code in a shape that it's not more broken that the existing code, then I'll rewrite it completely in a followup PR in a separate module for simplicity.

anthrotype: hm ok.

> Let's make this code in a shape that it's not more broken that the existing code,

are you going to do that?

behdad: I think after https://github.com/fonttools/fonttools/commit/95a0fb4a0f933fa3bf0bfc8a60fa68fe12c9e6f7 it should be roughly in that state.

anthrotype (approved):
LGTM! Amazing work, behdad!

Sorry if it took me so long to review this... 

I have a few minor changes I would like to make (e.g. expand single|ranges => triples only at the public interface and expect triples internally, fix tests accordingly) so if you don't mind I'm going to merge this and make a new PR on top.

behdad:
Thanks Cosimo!!

chrissimpkins:
This is fantastic.  Thanks so much Behdad!

PR #2729, opened 2022-08-07
Be able to getDebugName depending on the platformID

moi15moi:
### Why is this PR is needed?
Currently, if the for loop in the method **getDebugName** "meet" a name with a **platformID=1** and **langID=0**, it will directly return it. This mean that it can preferred a name from mac platform over a name from microsoft platform. 

However, here is what the Apple documentation mentions: [``Names with platformID 1 were required by earlier versions of macOS. Its use on modern platforms is discouraged``](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6name.html).

This is why something need to be changed in getDebugName.

### What logic is it based on?
- By default, use the same order has [FontConfig](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/d863f6778915f7dd224c98c814247ec292904e30/src/fcfreetype.c#L1078-1083)
- The method isEnglish is inspired from [FontConfig](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/d863f6778915f7dd224c98c814247ec292904e30/src/fcfreetype.c#L1111-1125)
- Finally, the sort logic is also from [FontConfig](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/d863f6778915f7dd224c98c814247ec292904e30/src/fcfreetype.c#L1128-1140)

### DIFF
--- a/Lib/fontTools/ttLib/tables/_n_a_m_e.py
+++ b/Lib/fontTools/ttLib/tables/_n_a_m_e.py
@@ -99,26 +99,26 @@ class table__n_a_m_e(DefaultTable.DefaultTable):
 					return namerecord
 		return None # not found
 
-	def getDebugName(self, nameID):
-		englishName = someName = None
-		for name in self.names:
-			if name.nameID != nameID:
-				continue
+	def getDebugName(self, nameID, platformIDs=(3, 0, 1, 2)):
+		"""
+		Parameters:
+			nameID: NameID from NamingTable.
+			platformIDs: Search the name depending on the platformID order.
+		Returns:
+			Decoded nameID.
+		"""
+		def isEnglish(name: NameRecord) -> bool:
+			return (name.platformID, name.langID) in ((1, 0), (3, 0x409))
+		filteredNamingTable = filter(lambda name: name.nameID == nameID and name.platformID in platformIDs, self.names)
+		sortedNamingTable = sorted(filteredNamingTable, key=lambda name: (platformIDs.index(name.platformID), name.platEncID, -isEnglish(name), name.langID))
+
+		for name in sortedNamingTable:
 			try:
 				unistr = name.toUnicode()
 			except UnicodeDecodeError:
 				continue
-
-			someName = unistr
-			if (name.platformID, name.langID) in ((1, 0), (3, 0x409)):
-				englishName = unistr
-				break
-		if englishName:
-			return englishName
-		elif someName:
-			return someName
-		else:
-			return None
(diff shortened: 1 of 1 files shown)


anthrotype:
The two sets of names should contain the same string so it shouldn't matter which one you pick.
But if you do already know that you want the platform 3 english nameID, you can just go on and use `getName` method directly. This `getDebugName` method is intended to give some good enough name string mostly for debugging purposes.

I'd be ok to simply switch the hard-coded priority and place windows platform 3 before the macintosh platform 0 in the list that is searched through, if we really want to give priority to unicode strings over legacy 8-bit encodings.

moi15moi:
> The two sets of names should contain the same string so it shouldn't matter which one you pick.

Like you said, it **_should_**. It is not always the case.

> But if you do already know that you want the platform 3 english nameID, you can just go on and use getName method directly.

Currently, `getName` does not prioritize English names. In my opinion, it should prioritize it like GDI or FontConfig. 
I would like to have your opinion about that.
https://github.com/fonttools/fonttools/blob/9716b1c4556642f92d66d8ceee0225e6c5ade8bb/Lib/fontTools/ttLib/tables/_n_a_m_e.py#L93-L100

anthrotype:
I'm not convinced that `getName` should prioritize English when `langID=None`. `None` here means "I don't care, get me the first one you can find". If one wants English Windows Unicode name one can explicitly do `getName(nameID, platformID=3, platEncID=1, langID=0x409)`.
`getDebugName` is for debugging or displaying some useful comments next to nameIDs in a ttx dump (e.g. for fvar axes or instances), in there it makes sense to prefer English.
I understand your point that macintosh legacy name strings don't always match the unicode windows ones, and I'm willing to change getDebugName to use the latter in preference of the former, basically 

```diff
- if (name.platformID, name.langID) in ((1, 0), (3, 0x409)):
+ if (name.platformID, name.platEncID, name.langID) in ((3, 1, 0x409), (1, 0, 0)):
```

(notice I included `platEncID` to make the match even more explicit)

moi15moi:
> (notice I included `platEncID` to make the match even more explicit)

I just tried to mimic [FontConfig](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/d863f6778915f7dd224c98c814247ec292904e30/src/fcfreetype.c#L1111-1125), but if you prefer it like that, it is still really close to it.

But finally, with hindsight, I think this PR should not be accepted.

PR #2717, opened 2022-07-29
[varLib.models] Generate narrower tents

behdad:
Based on https://github.com/fonttools/fonttools/pull/2218
but modify the general solver instead.

PR #2732, opened 2022-08-09
[instancer] Simplify a couple of things

behdad:
I think the proposed logic is cleaner.

### DIFF
--- a/Lib/fontTools/varLib/instancer/__init__.py
+++ b/Lib/fontTools/varLib/instancer/__init__.py
@@ -926,14 +926,9 @@ def instantiateAvar(varfont, axisLimits):
                     continue
                 fromCoord = normalizeValue(fromCoord, (axisRange.minimum, 0, axisRange.maximum))
 
-                if toCoord < 0:
-                    assert mappedMin != 0
-                    assert toCoord >= mappedMin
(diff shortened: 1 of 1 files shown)


anthrotype:
This all looks good to me, except for the first commit [[instancer] Simplify case 4 logic](https://github.com/fonttools/fonttools/pull/2732/commits/107b35b58c87dc8aec053928cac1f1ceb8d0a52c).

I think we discussed this together before I implemented the current code, we opted for extending the tent to its max -/+2.0 as a sort of optimization so that the magnitude of the deltas needed to obtain the desired effect would be smaller.
So I'd like to keep it, unless it has some other side effects which I am not seeing.

behdad:
> I think we discussed this together before I implemented the current code, we opted for extending the tent to its max -/+2.0 as a sort of optimization so that the magnitude of the deltas needed to obtain the desired effect would be smaller.
> So I'd like to keep it, unless it has some other side effects which I am not seeing.

Now I remember. You are right.

My only argument is that in https://github.com/fonttools/fonttools/pull/2717 we're going the opposite direction, generating non-overlapping tents to reduce rounding errors. But happy to remove this change. I'll merge the rest. Thanks.

Issue #2733, opened 2022-08-09
[instancer] Bug in _instantiateFeatureVariations()


behdad:
In `_instantiateFeatureVariations`:

- We first call `_instantiateFeatureVariationRecord`(); this renumbers the axisIdx of the record,

- We then call `_limitFeatureVariationRecord()`; this uses the axisIdx to lookup axisTag *using the old fvar axes*.

I *think* this should fix it:
```
diff --git a/Lib/fontTools/varLib/instancer/__init__.py b/Lib/fontTools/varLib/instancer/__init__.py
index b8a095dd..12f8edb0 100644
--- a/Lib/fontTools/varLib/instancer/__init__.py
+++ b/Lib/fontTools/varLib/instancer/__init__.py
@@ -787,12 +787,12 @@ def _instantiateFeatureVariationRecord(
     return applies, shouldKeep
 
 
-def _limitFeatureVariationRecord(record, axisRanges, fvarAxes):
+def _limitFeatureVariationRecord(record, axisRanges, axisOrder):
     newConditions = []
     for i, condition in enumerate(record.ConditionSet.ConditionTable):
         if condition.Format == 1:
             axisIdx = condition.AxisIndex
-            axisTag = fvarAxes[axisIdx].axisTag
+            axisTag = axisOrder[axisIdx]
             if axisTag in axisRanges:
                 axisRange = axisRanges[axisTag]
                 newRange = _limitFeatureVariationConditionRange(condition, axisRange)
@@ -836,7 +836,7 @@ def _instantiateFeatureVariations(table, fvarAxes, axisLimits):
             record, i, location, fvarAxes, axisIndexMap
         )
         if shouldKeep:
-            shouldKeep = _limitFeatureVariationRecord(record, axisRanges, fvarAxes)
+            shouldKeep = _limitFeatureVariationRecord(record, axisRanges, axisOrder)
 
         if shouldKeep and _featureVariationRecordIsUnique(record, uniqueRecords):
             newRecords.append(record)
```

anthrotype:
oh.. yes! you're right. The fix lgtm, but I think I prefer to have 8d69a52 splitted from the instancer-L4 PR of yours.

anthrotype:
ah sorry you did already split that off in a separate PR!

https://github.com/fonttools/fonttools/pull/2734

thanks

PR #2734, opened 2022-08-09
[instancer] Fix bug in _instantiateFeatureVariations()

behdad:
Fixes https://github.com/fonttools/fonttools/issues/2733


anthrotype (approved):
LGTM,

a test (e.g. two axes, one pinned, the second one restricted in range) would have been nice too

anthrotype:
we can add a test later, I want to include this in the upcoming release

PR #2738, opened 2022-08-12
TTVarGlyphset

behdad:
Just, this is pretty much your code integrated into TTFont. Needs some clean up still, but I think I like the general structure. I did minor testing. Can you do some more?


behdad:
I used this in the `varLib.interpolatable` tool already, seems to be working.

behdad:
I removed the compile though, used deepcopy instead.

behdad:
> I removed the compile though, used deepcopy instead.

And now a shallow copy.

justvanrossum:
Thanks for picking this up!

I'm not immediately able to contribute more testing. We probably have some test fonts in the the test suite that can be used for this?

behdad:
This is ready now. I visually tested it with the svgPathPen module as well.

justvanrossum:
Looks good to me, but I would prefer more test coverage before we merge.

behdad:
Sure. I'll add a subset RobotoFlex font with just letter 'I' and test it at various locations.

behdad:
Tests added.

justvanrossum (commented):
LGTM, would wait for review by Cosimo, though.

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -673,7 +678,7 @@ def getTableData(self, tag):
 		else:
 			raise KeyError(tag)
 
-	def getGlyphSet(self, preferCFF=True):
+	def getGlyphSet(self, preferCFF=True, location=None, normalized=False):
 		"""Return a generic GlyphSet, which is a dict-like object
```
anthrotype, line 682: docstring could mention the new location and normalized parameters, and the fact that currently this only works for TrueType-flavored variable fonts. Maybe you could add some TODO or file an issue for CFF2 support

Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -0,0 +1,214 @@
...
+class _TTVarGlyphSet(object):
+
+	def __init__(self, font, location, normalized=False):
+		from fontTools.varLib.models import normalizeLocation, piecewiseLinearMap
+		self._ttFont = font
+		if not normalized:
+			axes = {a.axisTag: (a.minValue, a.defaultValue, a.maxValue) for a in font['fvar'].axes}
+			location = normalizeLocation(location, axes)
+			if 'avar' in font:
+				avar = font['avar']
```
anthrotype, line 115: would be nice of `avar` had a `mapLocation` or something that did the same so we can share this logic in few other places (not necessarily in this same PR)

Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -0,0 +1,214 @@
...
+				continue
+			delta = var.coordinates
+			if None in delta:
+				if origCoords is None:
+					origCoords,control = glyf._getCoordinatesAndControls(self._glyphName, hMetrics, vMetrics)
+					endPts = control[1] if control[0] >= 1 else list(range(len(control[1])))
+				delta = iup_delta(delta, origCoords, endPts)
+			coordinates += GlyphCoordinates(delta) * scalar
+
+		from copy import copy
```
anthrotype, line 208: could be moved to the top-level scope

Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -0,0 +1,214 @@
...
+	def __getitem__(self, glyphName):
+		return _TTVarGlyphGlyf(self._ttFont, glyphName, self.location)
+
+	def get(self, glyphName, default=None):
+		try:
+			return self[glyphName]
+		except KeyError:
+			return default
+
+def setCoordinates(glyph, coord, glyfTable):
```
anthrotype, line 144: looks almost identical to `glyf._setCoordinates`, ideally we'd factor out the shared code. Maybe not worth in this case. I see you don't round here, whereas glyf._setCoordinates does, but I guess that's intentional.

Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -0,0 +1,214 @@
...
+	return horizontalAdvanceWidth, leftSideBearing, verticalAdvanceWidth
+
+
+class _TTVarGlyphGlyf(object):
+
+	def __init__(self, ttFont, glyphName, location):
+		self._ttFont = ttFont
+		self._glyphName = glyphName
+		self._location = location
+		#self.width, self.lsb = ttFont['hmtx'][glyphName]
```
anthrotype, line 182: maybe initialize these attributes to None or 0, and comment that the get updated once you draw the glyph.

Lib/fontTools/varLib/interpolatable.py
```diff
@@ -377,6 +377,37 @@ def main(args=None):
...
+                locs = set()
+                for variations in gvar.variations.values():
+                    for var in variations:
+                        loc = []
+                        for tag,val in sorted(var.axes.items()):
+                            loc.append((tag,val[1]))
+                        locs.add(tuple(loc))
+                # Rebuild locs as dictionaries
+                new_locs = [{}]
+                for loc in (sorted(locs, key=lambda v: (len(v), v))):
```
anthrotype, line 396: there's one parenthesis too much in **(** sorted(...) **)**

anthrotype (approved):
LGTM, with some minor comments

Lib/fontTools/ttLib/ttGlyphSet.py
behdad: I can add rounding just to be similar to the other case. It doesn't need rounding. I copied code from Just.

You are right that ideally this code should be refactored.

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -684,16 +689,28 @@ def getGlyphSet(self, preferCFF=True):
...
+			if location:
+				raise NotImplementedError # TODO
 			glyphs = _TTGlyphSet(self,
 			    list(self[table_tag].cff.values())[0].CharStrings, _TTGlyphCFF)
 
 		if glyphs is None and "glyf" in self:
-			glyphs = _TTGlyphSet(self, self["glyf"], _TTGlyphGlyf)
+			if location and 'gvar' in self:
+				glyphs = _TTVarGlyphSet(self, location=location, normalized=normalized)
+			else:
```
anthrotype, line 712: I think if `location is not None and "gvar" not in self` (i.e. non variable font) we should error out instead of silently return the default glyph set

behdad: > I think if `location is not None and "gvar" not in self` (i.e. non variable font) we should error out instead of silently return the default glyph set

That's debatable.  What if location points to an axis that doesn't exist in the varfont then? That's the same as location and no varfont... So I'm leaning towards the code as is.

PR #2739, opened 2022-08-13
[svgPen] Add main that prints SVG for input text

behdad:


behdad:
This helps me visually test glyphset draw functionality, which was hard to test previously.

khaledhosny (approved):

PR #2741, opened 2022-08-14
Lazy gvar

behdad:
Here's a possible sketch of lazy loading of gvar. Two tests fail, and has its limitations.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_v_a_r.py
+++ b/Lib/fontTools/ttLib/tables/_g_v_a_r.py
@@ -1,3 +1,4 @@
+from functools import partial
 from fontTools.misc import sstruct
 from fontTools.misc.textTools import safeEval
 from . import DefaultTable
@@ -36,7 +37,7 @@ GVAR_HEADER_FORMAT = """
 
 GVAR_HEADER_SIZE = sstruct.calcsize(GVAR_HEADER_FORMAT)
 
-class lazy_dict(dict):
+class _lazy_dict(dict):
 
     def get(self, k, *args):
         v = super().get(k, *args)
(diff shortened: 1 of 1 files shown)


behdad:
The main problem is when someone iterates over the .variations they don't get the actual value. Only if they use attr lookup syntax...

behdad:
I fixed `.items()` and `.values()`. The remaining failure is for when the `.variations` dictionary is tested for equality to another dictionary. That's harder to fix.

behdad:
Fixed remaining test.

Lib/fontTools/ttLib/tables/_g_v_a_r.py
```diff
@@ -97,23 +137,17 @@ def decompile(self, data, ttFont):
...
-			try:
-				self.variations[glyphName] = decompileGlyph_(
-					numPointsInGlyph, sharedCoords, axisTags, gvarData)
-			except Exception:
-				log.error(
-					"Failed to decompile deltas for glyph '%s' (%d points)",
-					glyphName, numPointsInGlyph,
-				)
-				raise
+			self.variations[glyphName] = get_lambda(gvarData, numPointsInGlyph)
```
justvanrossum, line 150: Wouldn't this be prettier with a `functools.partial`?

justvanrossum (approved):
Awesome! LGTM.

Lib/fontTools/ttLib/tables/_g_v_a_r.py
behdad: Thanks. Fixed.

justvanrossum: I was rather thinking that you could partial `decompileGlyph_` directly. I don't think `func` is really needed at all?

behdad: Oh, you're right! Let me do that. Thanks.

behdad: Done. So much nicer. Thanks.
https://github.com/fonttools/fonttools/pull/2738 is also ready :D.

PR #2743, opened 2022-08-15
support proposed PaintSweepGradient's angles with +1.0 (180°) bias

anthrotype:
In the upcoming draft of COLR spec, PaintSweepGradient's startAngle/endAngle are encoded with a +1.0 bias to allow for representation of a full +360° positive angle. Normal F2Dot14-fraction-of-half-circle angles can only represent angles between -360 <= angle < +360

This is a breaking change and will need to be coordinated with rendering implementations (at least FreeType/Skia).

/cc @drott


drott:
LGTM

drott (approved):

anthrotype:
For reference, see https://github.com/googlefonts/colr-gradients-spec/pull/362

PR #2744, opened 2022-08-15
[feaLib.builder] Fix for #2694: Remove GDEF warning

### LINKED ISSUE #2694: [feaLib.builder] Irrelevant "Must save GDEF when compiling a variable font" error when adding GSUB-only to VF
feaLib.builder insists on building GDEF, even if GPOS is not requested, for example if I only want to add GSUB to a VF.

I think the `elif self.varstorebuilder:` line below should read `elif "GPOS" in tables and self.varstorebuilder:`

https://github.com/fonttools/fonttools/blob/4139d8914abe788cc396dcb6ccc7c9cdb365ee89/Lib/fontTools/feaLib/builder.py#L227-L234

See also https://github.com/fonttools/fonttools/discussions/2690#discussioncomment-3131706

justvanrossum:
Only warn about not-requested-GDEF if building a VF *and* GPOS is requested but GDEF isn't.

Fixes #2694 more.

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -224,14 +224,12 @@ class Builder(object):
                 del self.font[tag]
         if any(tag in self.font for tag in ("GPOS", "GSUB")) and "OS/2" in self.font:
             self.font["OS/2"].usMaxContext = maxCtxFont(self.font)
-        gdef = self.buildGDEF() if "GDEF" in tables or "GPOS" in tables else None
         if "GDEF" in tables:
+            gdef = self.buildGDEF()
             if gdef:
                 self.font["GDEF"] = gdef
             elif "GDEF" in self.font:
                 del self.font["GDEF"]
-        elif self.varstorebuilder and gdef:
-            log.warning("GDEF is not requested, but is needed")
         if "BASE" in tables:
             base = self.buildBASE()
             if base:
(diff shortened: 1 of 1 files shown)


anthrotype (approved):
LGTM


(honestly I am tempted to remove the warning altogether, if GDEF is not requested, there has to be a reason why one does that, so no one is built, period.. but at least this PR reduces the chances that one sees that warning so I'm ok with that)

justvanrossum:
I would be fine with removing the warning, too. If @simoncozens agrees?

anthrotype:
somehow building GDEF unconditionally is causing other issues in ufo2ft, see https://github.com/fonttools/fonttools/issues/2747

simoncozens:
I'm OK with removing the warning; otherwise this patch looks right.

justvanrossum:
Ugh. Shall I adjust this PR to get rid of the warning altogether?

simoncozens:
Thanks. The original rationale for it was to help out downstream library users who may not have realised that they needed to build GDEF in the case of a variable font. But that's an absolutely tiny edge case.

justvanrossum:
Sorry for the mess in the previous attempt to fix. This is good now I think.

simoncozens:
As far as #2747 goes, it looks like building GDEF breaks things; not building GDEF obviously papers over the problem, but there's clearly a deeper issue there that I'll try and look at. (I know in my variable feature code PR to ufo2ft I had to clone the AST to avoid the problem.)

anthrotype (approved):

Issue #2748, opened 2022-08-16
ast.GlyphClass should contain list of ast.GlyphName objects or a list of str?


anthrotype:
this issue came up in https://github.com/fonttools/fonttools/issues/2747 and https://github.com/googlefonts/fontmake/issues/925 and in conversation with @simoncozens 

Basically, on the one hand, the `ast.GlyphClass` docstring as well as a unit test in feaLib/ast_test.py say that GlyphClass is a collection of `ast.GlyphName` objects (in turn a thin wrapper around a str for the actual glyph name, and its location in the file to aid debugging).

https://github.com/fonttools/fonttools/blob/853cadb0cdfba7b9891d8de6eccaac122633c70f/Lib/fontTools/feaLib/ast.py#L229-L235

https://github.com/fonttools/fonttools/blob/853cadb0cdfba7b9891d8de6eccaac122633c70f/Tests/feaLib/ast_test.py#L7-L9

However, when the feaLib.parser encounter one of these (e.g. `@SOMETHING = [A B C];`), it actually creates ast.GlyphClass objects that contain list of plan `str`, not `ast.GlyphName`.

```python
from io import StringIO
from fontTools.feaLib import ast
from fontTools.feaLib.parser import Parser

fea = Parser(StringIO("@SOMETHING = [A B C];")).parse()
assert isinstance(fea.statements[0], ast.GlyphClassDefinition)
assert isinstance(fea.statements[0].glyphs, ast.GlyphClass)
assert isinstance(fea.statements[0].glyphs.glyphs[0], str)  # not ast.GlyphName?!

print(fea.statements[0].glyphs.glyphs)  # ['A', 'B', 'C']
```

This is messed up. Simon is already looking into adding static typing info and cleaning this up.

anthrotype:
In this particular case, I believe ast.GlyphClass should contain ast.GlyphNames rather than str, this way the location of each glyph name in the class can be recorder and reported as needed.

It seems `Parser.expect_glyph_()` is currently returning a plain str, whereas potentially it could return an `ast.GlyphName(glyph, location=self.cur_token_location_)` since it already knows that glyph's location. Then subsequent code that assumes the return value was a str can be adapted to deal with an ast.GlyphName instead.

typoman:
FWIW for a feaLib subsetter and more advanced high level modifications to features I think all the glyphs better to be parsed as an `ast.GlyphName` not a string object. This would make it possible to add also more methods to this object.

Issue #2747, opened 2022-08-16
version 4.35.0 breaks our tests in gftools


m4rc1e:
v4.35.0 breaks two of our builder checks, https://github.com/googlefonts/gftools/runs/7839959593?check_suite_focus=true#step:5:186. Here's the traceback

```
ERROR:root:In 'Lora.glyphs' -> 'master_ufo/Lora.designspace': Generating fonts from Designspace failed: <fontTools.feaLib.ast.GlyphName object at 0x108267610>
Traceback (most recent call last):
  File "/Users/marcfoley/Type/fonttools/venv/lib/python3.10/site-packages/fontmake/font_project.py", line 1065, in run_from_designspace
    self._run_from_designspace_interpolatable(
  File "/Users/marcfoley/Type/fonttools/venv/lib/python3.10/site-packages/fontmake/font_project.py", line 1157, in _run_from_designspace_interpolatable
    self.build_variable_fonts(
  File "/Users/marcfoley/Type/fonttools/venv/lib/python3.10/site-packages/fontmake/font_project.py", line 384, in build_variable_fonts
    fonts = ufo2ft.compileVariableTTFs(
  File "/Users/marcfoley/Type/fonttools/venv/lib/python3.10/site-packages/ufo2ft/__init__.py", line 596, in compileVariableTTFs
    vfNameToBaseUfo = _compileNeededSources(
  File "/Users/marcfoley/Type/fonttools/venv/lib/python3.10/site-packages/ufo2ft/__init__.py", line 785, in _compileNeededSources
    ttfDesignSpace = compileInterpolatableFunc(
  File "/Users/marcfoley/Type/fonttools/venv/lib/python3.10/site-packages/ufo2ft/__init__.py", line 384, in compileInterpolatableTTFsFromDS
    for source, ttf in zip(result.sources, ttfs):
  File "/Users/marcfoley/Type/fonttools/venv/lib/python3.10/site-packages/ufo2ft/__init__.py", line 318, in compileInterpolatableTTFs
    compileFeatures(ufo, ttf, glyphSet=glyphSet, **kwargs)
  File "/Users/marcfoley/Type/fonttools/venv/lib/python3.10/site-packages/ufo2ft/__init__.py", line 507, in compileFeatures
    otFont = featureCompiler.compile()
  File "/Users/marcfoley/Type/fonttools/venv/lib/python3.10/site-packages/ufo2ft/featureCompiler.py", line 127, in compile
    self.setupFeatures()
[...]

PR #2753, opened 2022-08-18
Add type hints to feaLib.ast

simoncozens:
This is a big and noisy change; sorry. It also changes some public-facing aspects of the feaLib API (particularly the constructor parameters of AST object).

But in a good way! We previously had several situations where the documentation specified certain types, and either the builder expected or the parser generated something different. This lead to horrible bugs like #2747. I also found a couple of typos which were hidden by the lack of typing.

I suggest that because of the API change, we send a courtesy ping to anyone we know who's using it. (I know SIL are, at least.)

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -30,6 +30,7 @@ __all__ = [
     "MarkClassDefinition",
     "AlternateSubstStatement",
     "Anchor",
+    "NullAnchor",
     "AnchorDefinition",
     "AttachStatement",
     "AxisValueLocationStatement",
@@ -695,8 +696,8 @@ class Anchor(Expression):
 
     def __init__(
         self,
-        x: int,
-        y: int,
+        x: Optional[int],
+        y: Optional[int],
         name=None,
         contourpoint: int = None,
         xDeviceTable=None,
@@ -708,9 +709,14 @@ class Anchor(Expression):
         self.x, self.y, self.contourpoint = x, y, contourpoint
         self.xDeviceTable, self.yDeviceTable = xDeviceTable, yDeviceTable
 
+    def isNull(self) -> bool:
+        return self.x is None and self.y is None
+
     def asFea(self, indent="") -> str:
         if self.name is not None:
             return "<anchor {}>".format(self.name)
+        if self.isNull():
+            return "<anchor NULL>"
         res = "<anchor {} {}".format(self.x, self.y)
         if self.contourpoint:
             res += " contourpoint {}".format(self.contourpoint)
@@ -721,6 +727,9 @@ class Anchor(Expression):
             res += deviceToString(self.yDeviceTable)
         res += ">"
         return res
+
+
+NullAnchor = Anchor(None, None)
 
 
 class AnchorDefinition(Statement):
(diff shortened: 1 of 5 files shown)


anthrotype:
wow that's huge! I hope to be able to review next week (i'm travelling tomorrow).

Regarding the breaking changes, could you make a list of those? That'd help people adapting their code to the proposed API, for it's hard to disentagle those changes from the rest of the typing stuff.

simoncozens:
I think the only actual breaking change is that the `glyph` and `replacement` arguments to `MultipleSubstStatement` are now AST objects, not strings/lists of strings.

anthrotype:
oh then we can say it's a bug fix!

dscorbett:
Does there need to be a breaking change? You could use `Union` to support both the old and new types of those parameters.

simoncozens:
This isn't just a matter of type *hints*, we're changing the actual types - we're now calling `.glyph` on the `glyph` and `replacement` attributes. They need to be objects. Having these things sometimes be objects and sometimes be strings is what got us into a mess.

dscorbett:
Right. I’ll rephrase more explicitly. Does there need to be a breaking change? Can you automatically make the code accept either type but convert `str` to the new type in the constructor, and then use `Union` to support both the old and new types?

simoncozens:
Grr. This doesn't actually fix the "AST objects in buildCoverage" problem.

simoncozens:
This is because we haven't resolve #2748 (whether to put AST objects or strings into GlyphClass). Currently I've gone with the *parser*'s existing behaviour of putting strings in there, because that makes the type checking work for AST+parser. But I think we are moving in the direction of putting AST objects in there. I'll try that and see what shakes out.

anthrotype:
> we are moving in the direction of putting AST objects in there.

+1 to that

> Can you automatically make the code accept either type but convert str to the new type in the constructor

I agree we should try to do that, accept a union of the old/new type in the constructor parameter and coerce to the desired type for the instance field

anthrotype:
(i haven't had to time to review this yet, but I just wanted to point out I would like to avoid backward incompatible changes only for the sake of typechecking strictness; best if we could make explicit in those type hints all the implicit assumptions that the code currently makes; and if we do want to deprecate something, we at least issue deprecation warnings and wait a couple of releases)

simoncozens:
Urgh, there is another incompatibility here between feaLib.parser and ufo2ft.featureWriters.kernFeatureWriter.

kernFeatureWriter does this:

```python
            return {g.asFea() for g in classDef1.glyphSet()}
```

Previously `glyphSet` *usually* returned a list of AST objects but sometimes (in `GlyphName`) returned a list of strings. With this PR, it always returns strings.

behdad:
What's the status of this?

simoncozens:
Still waiting for @anthrotype to have time to review it. :-)

anthrotype:
did you fix the backward-incompatible changes as @dscorbett suggested?
also, there seems to be merge conflicts.

simoncozens:
> did you fix the backward-incompatible changes as @dscorbett suggested?

Yes, looks like I did:

```
        if isinstance(glyph, str):
            glyph = GlyphName(glyph, location)
        replacement = [
            GlyphName(glyph, location) if isinstance(glyph, str) else glyph
            for glyph in replacement
        ]
```

simoncozens:
I still think this is very much worth doing but it's going to need more work, so I'll pull it back to draft for now.

PR #2755, opened 2022-08-19
[scaleUpem] Support CFF/CFF2

behdad:


Lib/fontTools/ttLib/scaleUpem.py
```diff
@@ -144,6 +145,83 @@ def visit(visitor, obj, attr, kernTables):
...
+            value = getattr(topDict, attr, None)
+            if value is None:
+                continue
+            if isinstance(value, list):
+                _cff_scale(visitor, value)
+            else:
+                setattr(topDict, attr, visitor.scale(value))
+
+        topDict.FontMatrix[0] /= visitor.scaleFactor
+        topDict.FontMatrix[3] /= visitor.scaleFactor
```
anthrotype, line 198: what about FontMatrix[1] and FontMatrix[2]? (I honestly forgot and didn't check the CFF spec)

anthrotype (approved):

Lib/fontTools/ttLib/scaleUpem.py
```diff
@@ -1,13 +1,14 @@
 """Change the units-per-EM of a font.
 
-Currently does not support CFF fonts. AAT, Graphite tables are not supported
-either."""
+AAT and Graphite tables are not supported."""
```
anthrotype, line 3: should we mention that CFF/CFF2 get desubroutinized before scaling?

Issue #1156, opened 2018-01-23
makeOutputFile


behdad:
Right now we have makeOutputFileName. That's racy in choosing output name if overlap=False. We should replace with a function that 1. returns an open file, to fix race condition, and ideally 2. writes to a temp file and replaces to final file upon .close() or context __leave__. Then use it everywhere.

behdad:
While having race conditions is bad, I don't see this change happening any time soon.

anthrotype:
We could add a new one rather than replace the existing one as it may be used outside fonttools.
Something like this:

```python
import tempfile
import shutil
from contextlib import contextmanager
from pathlib import Path

@contextmanager
def move_tempfile_to_dest(dest_path):
    with tempfile.NamedTemporaryFile(delete=False) as tmp:
        try:
            yield tmp
            tmp.flush()
            shutil.move(tmp.name, dest_path)
        finally:
            Path(tmp.name).unlink(missing_ok=True)
```

though this one doesn't add file extension, nor appends the `#1`, `#2`, etc. suffix (which IIUC is where the raciness comes from)

PR #2756, opened 2022-08-19
[cliTools.makeOutputFileName] Add fragment argument

behdad:
And use it from subsetter and instancer.

Related to https://github.com/fonttools/fonttools/issues/1156


Lib/fontTools/misc/cliTools.py
```diff
@@ -6,21 +6,24 @@
 numberAddedRE = re.compile(r"#\d+$")
 
 
-def makeOutputFileName(input, outputDir=None, extension=None, overWrite=False):
+def makeOutputFileName(input, outputDir=None, extension=None, overWrite=False, fragment=""):
```
anthrotype, line 9: maybe "suffix" as opposed to "fragment" would be clearer as parameter name?

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -1468,9 +1469,8 @@ def main(args=None):
         updateFontNames=options.update_name_table,
     )
 
-    outfile = (
-        os.path.splitext(infile)[0]
-        + "-{}.ttf".format("instance" if isFullInstance else "partial")
+    fragment = "-instance" if isFullInstance else "-partial"
+    outfile = (makeOutputFileName(infile, overWrite=True, fragment=fragment)
```
anthrotype, line 1473: this should go to a new line in black style, re-run black on instancer modile

anthrotype (approved):

Issue #1252, opened 2018-04-28
[varLib] Extrapolate


behdad:
This is @LettError 's favorite feature. We should support it. I have ideas.

LettError:
Here "extrapolate" should mean "allowing the calculation of shapes outside the +/- 1.0 range", right?

behdad:
> Does this mean merely not clamping to normalized +/- 1.0 within interpolation in mutator/models?

No, requires more work than just that.

LettError:
Sorry to see.

behdad:
> Sorry to see.

I feel nerdsniped now. Lemme think about it. :)

PR #2757, opened 2022-08-19
[varLib] Support extrapolation

behdad:
Fixes https://github.com/fonttools/fonttools/issues/1252

@LettError can you test this?


behdad:
I realized this doesn't work. I'm working on it.

behdad:
Okay. It might work now. Locations must be in normalized space. Ie. -1,0,+1. It will interpolate for locations outside -1,+1 range.

LettError:
(I must admit struggled with testing this on Saturday and wasn't successful. I really want this to work, I will try again)

behdad:
Here's a test, and results for extrapolate=False, and extrapolate=True:
https://gist.github.com/behdad/755d6b91be26947612b39a6db8b0f352
false
true

behdad:
2D case:
https://gist.github.com/behdad/ce10b065da4338d4a01e3e796659061c
False
true

LettError:
These graphs look good!
* maybe check extrapolation on the negative, so < 0 and < -1
* what if there is only 1 master? or two identical masters -- the result would be flat?
* do avar values play any role in this?

behdad:
> * maybe check extrapolation on the negative, so < 0 and < -1

I'll try.

> * what if there is only 1 master? or two identical masters -- the result would be flat?

I believe so.

> * do avar values play any role in this?

VariationModel takes normalized values. So any avar processing comes before this.

behdad:
https://gist.github.com/behdad/aecf2726f911beeaeda7a15eff2af708
Figure_1

anthrotype (approved):
LGTM! (after resolving conflicts)

PR #2760, opened 2022-08-22
[varLib] Clean up __all__

behdad:
It's okay for our other modules to use these symbols, but they don't really belong in the API of varLib.


behdad:
We can move them to misc somewhere.

anthrotype (approved):

Issue #2759, opened 2022-08-22
TTFont.getGlyphSet() returning _TTVarGlyphSet where type hints/typechecker expect _TTGlyphSet


anthrotype:
the new `_TTVarGlyphSet` (#2738) that is returned from `TTFont.getGlyphSet()` is not a subclass of the old `_TTGlyphSet`, and neither is the `_TTVarGlyphGlyf` that it contains a subclass of the old `_TTGlyph`.
We have some internal code that uses type hints and is typechecked using pytype, which breaks with the latest fonttools 4.36.0 following the above change.

```
bad option 'fontTools.ttLib.ttGlyphSet._TTVarGlyphGlyf' in return type [bad-return-type]
           Expected: fontTools.ttLib.ttGlyphSet._TTGlyph
```

I think we should revise the class hierarchy of these glyphset/glyph objects and make sure that we continue to return a `_TTGlyphSet` that contains generic `_TTGlyph` objects and work around their respective differences inside subclasses.
I think it's doable.

PR #2761, opened 2022-08-22
[ttGlyphSet] Inherit ttVarGlyphSet from ttGlyphSet

behdad:
Fixes https://github.com/fonttools/fonttools/issues/2759

### DIFF
--- a/Lib/fontTools/ttLib/ttGlyphSet.py
+++ b/Lib/fontTools/ttLib/ttGlyphSet.py
@@ -107,7 +107,7 @@ class _TTGlyphGlyf(_TTGlyph):
 
 
 
-class _TTVarGlyphSet(object):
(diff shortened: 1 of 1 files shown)


anthrotype:
that's the very minimum change, but the constructors' signatures are different and the public fields/method also differ somewhat. I was working on a different patch that aligns the TTVarGlyphSet/TTVarGlyphGlyf to their superclasses. I will send you a PR for review in a bit

behdad:
These objects are created by our code, so it doesn't matter that the constructors are different. You're right about the rest.

anthrotype:
> These objects are created by our code, so it doesn't matter that the constructors are different. 

hm, you're right.. In that case, my patch reduces basically to yours plus a bunch of small things. I'll still file that PR

PR #2762, opened 2022-08-22
[ttGlyphSet] Inherit ttVarGlyphSet from ttGlyphSet

### LINKED ISSUE #2759: TTFont.getGlyphSet() returning _TTVarGlyphSet where type hints/typechecker expect _TTGlyphSet

anthrotype:
Fixes #2759

### DIFF
--- a/Lib/fontTools/ttLib/ttGlyphSet.py
+++ b/Lib/fontTools/ttLib/ttGlyphSet.py
@@ -107,19 +107,22 @@ class _TTGlyphGlyf(_TTGlyph):
 
 
 
-class _TTVarGlyphSet(object):
+class _TTVarGlyphSet(_TTGlyphSet):
 
 	def __init__(self, font, location, normalized=False):
-		from fontTools.varLib.models import normalizeLocation, piecewiseLinearMap
 		self._ttFont = font
+		self._glyphs = font['glyf']
+
 		if not normalized:
+			from fontTools.varLib.models import normalizeLocation, piecewiseLinearMap
+
 			axes = {a.axisTag: (a.minValue, a.defaultValue, a.maxValue) for a in font['fvar'].axes}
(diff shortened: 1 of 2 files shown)


Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -134,11 +136,15 @@ def keys(self):
...
 		return glyphName in self._ttFont['glyf']
+
 	__contains__ = has_key
 
 	def __getitem__(self, glyphName):
 		return _TTVarGlyphGlyf(self._ttFont, glyphName, self.location)
 
+	def __len__(self):
+		return len(self._ttFont['glyf'])
+
```
behdad, line 147: You can assign `self._ttFont['glyf']` as `self._glyphs` and remove a few of the methods now.

behdad (approved):

Lib/fontTools/ttLib/ttGlyphSet.py
anthrotype: done, thanks

Issue #2764, opened 2022-08-23
[varLib.models] Regression with "narrower tents"


justvanrossum:
I've found a serious regression with the "narrower tents" PR (#2717), specifically in this commit: 

- https://github.com/fonttools/fonttools/commit/688939394b8cae3b2a0d5747e435cbd828a14531

I will have to investigate further to see what's up. I'll try to write a reproducer.

justvanrossum:
The following script gives a different output before and after the above-mentioned commit:
```python
from fontTools.varLib.models import VariationModel


locations = [
    {},
    {"axis_A": 1.0},
    {"axis_B": 1.0},
    {"axis_A": 1.0, "axis_B": 1.0},
    {"axis_A": 0.5, "axis_B": 1.0},
    {"axis_A": 1.0, "axis_B": 0.5},
]


instanceLocation = {
    "axis_A": 0.5,
    "axis_B": 0.5,
}


masterValues = [
    0,
    10,
    20,
    70,
    50,
    60,
]

model = VariationModel(locations)
interpolatedValue = model.interpolateFromMasters(instanceLocation, masterValues)

# interpolatedValue differs before/after
# Before:
# - https://github.com/fonttools/fonttools/commit/c1dc2f3c81e89ba7224af65dcaa2fd46cae52a17
# After:
# - https://github.com/fonttools/fonttools/commit/688939394b8cae3b2a0d5747e435cbd828a14531
print(interpolatedValue)
```

Output before: 37.5
Output after: 47.5

justvanrossum:
@behdad

behdad:
I understand the problem. Let's revert that commit until I figure out a fix.

anthrotype:
in the meantime we reverted that commit, and pushed a new release without the "narrow tents": https://github.com/fonttools/fonttools/releases/tag/4.37.0

behdad:
Thanks @anthrotype 

I pushed PR https://github.com/fonttools/fonttools/pull/2766 but let's do more testing.

PR #2765, opened 2022-08-23
Revert "[varLib.models] Generate narrower tents"

anthrotype:
This reverts commit 688939394b8cae3b2a0d5747e435cbd828a14531.


anthrotype:
temporarily works around https://github.com/fonttools/fonttools/issues/2764

behdad:
Ah, thanks. I was on it. :)

anthrotype:
oh, ok ! it broke a couple tests, had to fix manually, let me know if they still make sense

behdad (approved):

behdad:
> oh, ok ! it broke a couple tests, had to fix manually, let me know if they still make sense

Yes. I was doing the same.

anthrotype:
i'll merge once CI passes and then cut bugfix release

behdad:
Now we should add test case from Just's bug report. I'll see if I can add that now.

behdad:
> Now we should add test case from Just's bug report. I'll see if I can add that now.

Don't wait for that though, obviously.

anthrotype:
i also need to fix the doctests..

behdad:
Test addded.

PR #2766, opened 2022-08-23
Narrow tents v2

behdad:
Fixes https://github.com/fonttools/fonttools/issues/2764


behdad:
I once tried proving the correctness of this change; ie. that only on-axis locations can be made narrow. But then the dog ate my homework.

behdad:
Thinking about it more, any full grid can be made narrow, but not anything else. That suggests that your approach of a simple vs general algorithm might be better, with the general algorithm being this version of mine. I'll think about it more.

justvanrossum:
> Thinking about it more, any full grid can be made narrow, but not anything else

It is a little broader than that, but not much. In https://github.com/fonttools/fonttools/pull/2218/commits/cb8c23afc94c48eb8721f7e28a9299ec4cfa34d6 I wrote:

	A model is "simple" when any two axes that actively cooperate
	(ie. have masters at non-zero positions for both axes) have a
	complete set of intermediates. For example, if axes "a" and "b"
	cooperate, and axis "a" has masters at 0, 0.5 and 1 from the
	neutral master, it must have those intermediates at all values
	for axis "b" as well.

It is worded clumsily, but I think you understand. A model can contain several "full grids".

behdad:
> A model can contain several "full grids".

Right. That's what I had in mind.

anthrotype:
what's the status of this one?

behdad:
Let's close for now. This doesn't quite do what Just's original simple-tents PR did. I'm happy to take that in if he still wants to. My brain can't come up with a generic way to do it.

justvanrossum:
Ok, I have reopened #2218, and I'll revisit as it's probably quite stale by now. I'm caught up with other things rn so it may be a while.

PR #2767, opened 2022-08-23
[varLib.plot] Add plotModelFromMasters

behdad:


Lib/fontTools/varLib/plot.py
```diff
@@ -131,6 +131,60 @@ def plotDocument(doc, fig, **kwargs):
...
+	Z = np.array(Z)
+
+	from matplotlib import pyplot
+	fix, ax = pyplot.subplots(subplot_kw={"projection": "3d"})
+
+	surf = ax.plot_surface(X, Y, Z)
+	pyplot.show()
+
+
+def plotModelFromMasters(model, masterValues):
```
anthrotype, line 179: the new function doesn't seem to be used anywhere, do you plan to hook it to main() via some flag I suppose?

behdad: > the new function doesn't seem to be used anywhere, do you plan to hook it to main() via some flag I suppose?

I can. For now it's a library function for visualization.

anthrotype: since there's no tests, would be nice to have an example of how to use

behdad: Added command-line for it.

anthrotype (approved):

PR #2768, opened 2022-08-23
[pens] Add cairoPen

behdad:


Lib/fontTools/pens/cairoPen.py
```diff
@@ -0,0 +1,26 @@
...
+        self.context = context
+
+    def _moveTo(self, p):
+        self.context.move_to(*p)
+
+    def _lineTo(self, p):
+        self.context.line_to(*p)
+
+    def _curveToOne(self, p1, p2, p3):
+        self.context.curve_to(*p1 + p2 + p3)
```
justvanrossum, line 23: This assumes p1-p3 are tuples or lists. The following is more general:
```suggestion
        self.context.curve_to(*p1, *p2, *p3)
```

anthrotype (approved):
LGTM, with Just's suggested fix

Issue #2772, opened 2022-08-24
Variable features and pairpos subtables


simoncozens:
This feature code doesn't do what it looks like it should:

```
lookup kern_Common {
    pos [p] g -5;
    pos [p] y (wght=900:-100 wght=400:-100);
} kern_Common;
```

This creates two class-based pairpos subtables (one for each value record format) and the second one never fires for the string "py".

Is this a bug? I think so. And I think a possible fix would be to promote the value record format of the first pair so that they can be placed in the same subtable. As I understand it, this is only a problem related to PairPos subtables, so I'd put the fix into `otlLib.builder.PairPosBuilder`.

simoncozens:
Weirder and weirder. In `PairPosBuilder`, we create a separate builder for each value format:

https://github.com/fonttools/fonttools/blob/d5a08b3b96f405b0fdca42e3f1f6baa32c59c687/Lib/fontTools/otlLib/builder.py#L1401-L1405

This is a `ClassPairPosSubtableBuilder`, which does its work by calling `buildPairPosClassesSubtable`. But `buildPairPosClassesSubtable` explicitly handles heterogenous value record formats! It first finds the "highest" value format for the list of value records:

https://github.com/fonttools/fonttools/blob/d5a08b3b96f405b0fdca42e3f1f6baa32c59c687/Lib/fontTools/otlLib/builder.py#L2128-L2129

and then promotes all VRs to that format:

https://github.com/fonttools/fonttools/blob/d5a08b3b96f405b0fdca42e3f1f6baa32c59c687/Lib/fontTools/otlLib/builder.py#L2143-L2152

So we don't actually need to create separate builders, and we can simplify the `PairPosBuilder` code to this:

```python
        builder = ClassPairPosSubtableBuilder(self)
        for glyphclass1, value1, glyphclass2, value2 in self.pairs:
            if glyphclass1 is self.SUBTABLE_BREAK_:
                builder.addSubtableBreak()
                continue
            builder.addPair(glyphclass1, value1, glyphclass2, value2)
        subtables = []
        if self.glyphPairs:
            subtables.extend(buildPairPosGlyphs(self.glyphPairs, self.glyphMap))
        subtables.extend(builder.subtables())
        lookup = self.buildLookup_(subtables)
```

All tests pass and my feature code now does what I expect it to.

anthrotype:
I'm trying to understand why it was done like that to begin with, and what the implications of your proposed change in #2776 are.
[...]

PR #2776, opened 2022-08-25
Promote pair-pos value records

simoncozens:
Use buildPairPosClassesSubtable's ability to promote value records; this allows variable and non-variable kerns to share a subtable. See discussion in #2772.


simoncozens:
This was exactly the same code as @behdad's, but with tests.

anthrotype (approved):

PR #2779, opened 2022-08-25
[gvar] Make iteration lazy as well

behdad:
This is continuation of #2741, making iteration lazy-load one item at a time as well.

Tests pass, but more testing would be good, in real applications.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_v_a_r.py
+++ b/Lib/fontTools/ttLib/tables/_g_v_a_r.py
@@ -54,14 +54,24 @@ class _lazy_dict(dict):
         return v
 
     def items(self):
-        if not hasattr(self, '_loaded'):
-            self._load()
-        return super().items()
+        gen = super().items()
+        def wrap(gen):
+            for v in gen:
+                if callable(v[1]):
+                    v = v[0],v[1]()
+                    self[v[0]] = v[1]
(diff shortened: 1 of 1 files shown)


justvanrossum:
By now I wonder if it isn't better to subclass `collections.abc.MutableMapping`. It only requires you to implement `__getitem__`, `__setitem__`, `__delitem__`, `__iter__`, and `__len__`, and everything else is done for you.

Lib/fontTools/ttLib/tables/_g_v_a_r.py
```diff
@@ -54,14 +54,24 @@ def __getitem__(self, k):
         return v
 
     def items(self):
-        if not hasattr(self, '_loaded'):
-            self._load()
-        return super().items()
+        gen = super().items()
+        def wrap(gen):
+            for v in gen:
```
justvanrossum, line 59: I would write `for k, v in gen` and aviod the noisy indexing below.

justvanrossum:
Or `collections.UserDict`: perhaps you can get away with only implementing `__getitem__`.

justvanrossum:
I made #2781 with UserDict so you can see.

anthrotype:
I think I also prefer #2781 because the less code the better

PR #2781, opened 2022-08-26
Use UserDict for lazy gar variations dict

justvanrossum:
This is an alternative to for #2779

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_v_a_r.py
+++ b/Lib/fontTools/ttLib/tables/_g_v_a_r.py
@@ -1,3 +1,4 @@
+from collections import UserDict
 from functools import partial
 from fontTools.misc import sstruct
 from fontTools.misc.textTools import safeEval
@@ -37,46 +38,14 @@ GVAR_HEADER_FORMAT = """
 
 GVAR_HEADER_SIZE = sstruct.calcsize(GVAR_HEADER_FORMAT)
 
-class _lazy_dict(dict):
+class _LazyDict(UserDict):
+	def __getitem__(self, k):
+		v = self.data[k]
+		if callable(v):
+			v = v()
+			self.data[k] = v
+		return v
 
-    def get(self, k, *args):
-        v = super().get(k, *args)
-        if callable(v):
-            v = v()
-            self[k] = v
-        return v
-
-    def __getitem__(self, k):
-        v = super().__getitem__(k)
-        if callable(v):
-            v = v()
-            self[k] = v
-        return v
-
-    def items(self):
-        if not hasattr(self, '_loaded'):
-            self._load()
-        return super().items()
-
-    def values(self):
-        if not hasattr(self, '_loaded'):
-            self._load()
-        return super().values()
-
-    def __eq__(self, other):
-        if not hasattr(self, '_loaded'):
-            self._load()
-        return super().__eq__(other)
-
-    def __neq__(self, other):
-        if not hasattr(self, '_loaded'):
-            self._load()
-        return super().__neq__(other)
-
-    def _load(self):
-        for k in self:
-            self[k]
-        self._loaded = True
 
 class table__g_v_a_r(DefaultTable.DefaultTable):
 	dependencies = ["fvar", "glyf"]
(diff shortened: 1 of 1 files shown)


Lib/fontTools/ttLib/tables/_g_v_a_r.py
```diff
@@ -39,6 +39,10 @@
 GVAR_HEADER_SIZE = sstruct.calcsize(GVAR_HEADER_FORMAT)
 
 class _LazyDict(UserDict):
+	def __init__(self, data):
+		super().__init__()
+		self.data = data
```
anthrotype, line 44: I don't think you need to override `__init__`, just pass the dict to the UserDict's default constructor and it will populate its internal self.data attribute with it:

https://github.com/python/cpython/blob/b69ba3d7233013ea8cb95a3e30e2338f7f76b359/Lib/collections/__init__.py#L1088-L1096

anthrotype: oh, maybe the `update` method that is called by the default constructor un-lazifies the data which defeats the purpose of the lazy dict?

Lib/fontTools/ttLib/tables/_g_v_a_r.py
```diff
@@ -37,46 +38,18 @@
 
 GVAR_HEADER_SIZE = sstruct.calcsize(GVAR_HEADER_FORMAT)
 
-class _lazy_dict(dict):
+class _LazyDict(UserDict):
```
anthrotype, line 41: maybe we could move this `_LazyDict` to `fontTools.misc.dictTools` or something, so can be used elsewhere.. Or maybe we do that if/when we find a use for it.

Lib/fontTools/ttLib/tables/_g_v_a_r.py
anthrotype: I just tried to delete the custom `_LazyDict.__init__`, and checked if the inner update() would evaluate the values, but it seems that it doesn't:

```python
In [1]: from fontTools.ttLib.tables._g_v_a_r import _LazyDict
In [2]: d = _LazyDict({"a": lambda: print("HEY!!!") or 1})  # initializing the lazy dict prints nothing
In [3]: d["a"]
HEY!!!
Out [3]: 1
```

justvanrossum: A custom __init__ indeed isn't needed, but I thought it's a bit wasteful to go through UserDict's __setitem__ if we know it just has to end up in data anyway.

Lib/fontTools/ttLib/tables/_g_v_a_r.py
justvanrossum: I think for now it's fine here.

Lib/fontTools/ttLib/tables/_g_v_a_r.py
anthrotype: hm ok. Maybe write a comment along those lines.
Usually doing `dict(d)` creates a shallow copy of the otherdict (which is what the default UserDict constructor also does), whereas in this case _LazyDict just sets its internal data to refer to that other dict. I suppose it's fine since this module is only consumer, but if this were a public LazyDict class making the copy would be more in line with the dict-like objects behavior.

anthrotype (approved):

Lib/fontTools/ttLib/tables/_g_v_a_r.py
justvanrossum: Look how I wrote it before: https://github.com/fonttools/fonttools/pull/2781/commits/cd68be6188b4c2711b1f0e47f7e4881d666a8356
Perhaps that makes it more explicit that it's a bit of a hack to avoid some unnecessary overhead.

anthrotype: you mean reverting to the first commit where you set data attribute after the fact? hm yeah if you like.
Or an alternative classmethod constructor like `self.variations = _LazyDict.with_data(variations)` so it's on one line.
In the end, it doesn't really matter because this it's only used once here.

behdad:
Looks much nicer indeed!

Issue #2777, opened 2022-08-25
missing assignment?


schriftgestalt:
https://github.com/fonttools/fonttools/blob/d5a08b3b96f405b0fdca42e3f1f6baa32c59c687/Lib/fontTools/ttLib/tables/otBase.py#L965

Am in missing something or is this `CountReference(table, conv.repeat, value=countValue)` never assigned to anything?

anthrotype:
it doesn't assign to a local variable, but the constructor has a side-effect of setting the count value (when not None) in the table (or if already set, asserts it's the same), see:

https://github.com/fonttools/fonttools/blob/d5a08b3b96f405b0fdca42e3f1f6baa32c59c687/Lib/fontTools/ttLib/tables/otBase.py#L766-L780

anthrotype:
yeah, it looks confusing. It could be refactored to make it look clearer

schriftgestalt:
Ok. But that is confusing. Relying on side effects sounds like a hack. Could we just set the table directly?

anthrotype:
yes as long as the logic in CountReference.setValue is maintained:

https://github.com/fonttools/fonttools/blob/d5a08b3b96f405b0fdca42e3f1f6baa32c59c687/Lib/fontTools/ttLib/tables/otBase.py#L777-L780

PR #2782, opened 2022-08-26
get the actual functionality out of the CountReference initializer

schriftgestalt:
to make it less confusing
see #2777


behdad:
But now we have duplicated code without any indication of that.
I think the old code with a comment would be better.

schriftgestalt:
I don’t think that duplicating there lines of code is a problem. But I’ll add a comment pointing to the CountReference code. 

I’ll check why the test fail next week.

PR #2784, opened 2022-08-27
[ttVarGlyphSet] Support CFF

behdad:
First stab. I'll do some cleanup. Will see what I can.


behdad:
I think I'm happy with this for now.

Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -178,12 +182,37 @@ def _setCoordinates(glyph, coord, glyfTable):
...
+
+class _TTVarGlyphCFF(_TTVarGlyph):
+
+	def draw(self, pen):
+		varStore = self._glyphs.varStore
+		if varStore is None:
+			blender = None
+		else:
+			from fontTools.varLib.varStore import VarStoreInstancer
+			vsInstancer = VarStoreInstancer(varStore.otVarStore, self._ttFont['fvar'].axes, self._location)
```
justvanrossum, line 203: Would it be more efficient to store the instancer in the glyph set object, since the instancer has an internal cache?

Lib/fontTools/misc/psCharStrings.py
```diff
@@ -410,17 +413,28 @@ def op_index(self, index):
...
+		if self.blender is None:
+			del self.operandStack[-(numOps-numBlends):] # Leave the default operands on the stack.
+		else:
+			argi = len(self.operandStack) - numOps
+			end_args = tuplei = argi + numBlends
+			while argi < end_args:
+				next_ti = tuplei + self.numRegions
+				deltas = self.operandStack[tuplei:next_ti]
+				delta = self.blender(self.vsIndex, deltas)
+				self.operandStack[argi] += otRound(delta)
```
justvanrossum, line 430: Is rounding here kosher? Don't we get an accumulation of rounding errors this way? Same at the HVAR site.

justvanrossum: To answer myself: no, it is not kosher. I have a test case that shows it:
image
This is the cap A from MutatorSans compiled as CFF2. Those points at the top align properly without the round (as they do in the TTF VF).

behdad:
Thanks Just. PTAL.

justvanrossum:
It works really well. Your last commit is hard to review as tabs got converted to spaces. Can you undo that?

behdad:
> It works really well. Your last commit is hard to review as tabs got converted to spaces. Can you undo that?

Ah, I ran black on it... I undid the tab to space. It still has some noise but hopefully you can review. Thanks.

justvanrossum (approved):
LGTM. There is plenty of opportunity for refactoring and cleaning up, but maybe later?

Curious for Cosimo's review.

justvanrossum:
Since this is a new module, maybe indeed go all the way with black + spaces. Although a separate commit for that would be neater.

justvanrossum:
I started sketching out a some cleanups for this module, but it's turning into real code.

Main ideas:
- we should only have to differentiate glyph set and glyph classes by `glyf` vs `CFF `/`CFF2`, not by variability
- use `collections.abc.Mapping` as the base class for glyph set

This approach so far seems to reduce code duplication, and increases clarity.

I'm still working on it, but once I'm done, it would make most sense to me to go on top of this PR. The current PR works fine and seems well tested already, so I'd be fine with merging it as it is now, and I'd then create a new PR branch. Does that sound reasonable?

behdad:
> * we should only have to differentiate glyph set and glyph classes by `glyf` vs `CFF `/`CFF2`, not by variability

I agree. I was reaching the same conclusion.

behdad:
> I'm still working on it, but once I'm done, it would make most sense to me to go on top of this PR. The current PR works fine and seems well tested already, so I'd be fine with merging it as it is now, and I'd then create a new PR branch. Does that sound reasonable?

Yes, I think this is well-contained enough to merge.

Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -178,16 +188,47 @@ def _setCoordinates(glyph, coord, glyfTable):
...
+		if "HVAR" in self._ttFont:
+			hvar = self._ttFont["HVAR"].table
+			varidx = self._ttFont.getGlyphID(self._glyphName)
+			if hvar.AdvWidthMap is not None:
+				varidx = hvar.AdvWidthMap.mapping[self._glyphName]
+			vsInstancer = VarStoreInstancer(
+				hvar.VarStore, self._ttFont["fvar"].axes, self._location
+			)
+			delta = vsInstancer[varidx]
+			self.width += delta
```
anthrotype, line 228: if you draw the same glyph multiple times, wouldn't this increment its advance width in-place all the time?

justvanrossum: Yes, but this should be resolved in #2789

behdad: I don't think so because `self.width` was reset a few lines earlier.

PR #2786, opened 2022-08-29
Don't crash when `post` format 1 is improperly used

### LINKED ISSUE #2736: Gracefully handle TrueType fonts that misuse `post` table format 1 ?
⚠️ **The attached font is broken** ⚠️ . It uses `post` table format 1 which [requires it to contain exactly the 258 glyphs in the standard Macintosh TrueType font file](https://docs.microsoft.com/en-us/typography/opentype/spec/post#version-10). This font has a lot more glyphs than that, and the first 258 are not ordered as expected by the spec.

Dumping the `hmtx` table with TTX fails with,
```
> ttx -t hmtx post_format_1.ttf
Dumping "post_format_1.ttf" to "post_format_1.ttx"...

ERROR: An exception occurred during the decompilation of the 'hmtx' table
  File "fontTools/ttLib/ttFont.py", line 420, in _readTable
    table.decompile(data, self)
  File "fontTools/ttLib/tables/_h_m_t_x.py", line 51, in decompile
    glyphName = glyphOrder[i]
```

Other font tools I tried handle this font just fine.
Can FontTools be made to gracefully handle this broken font?

[post_format_1.ttf.zip](https://github.com/fonttools/fonttools/files/9303651/post_format_1.ttf.zip)

miguelsousa:
Fixes #2736


justvanrossum:
Can you strip the test font from unneeded tables, and replace all glyphs with an empty glyph?

```
    tag     checksum    length    offset
    ----  ----------  --------  --------
    BASE  0x65235DBD        70     73608
    GDEF  0x0E600F0B       140     73468
    GPOS  0xC58F843E     21494     51972
    GSUB  0x676CF69A      4020     47952
    OS/2  0x5C09739A        96       360
    cmap  0x7D830F03      1820      2660
    glyf  0x1B788E11     40504      5584
    head  0x13A80257        54       236
    hhea  0x08BA0811        36       292
    hmtx  0xF1AD5BC9      2204       456
    loca  0x4BA3735A      1104      4480
    maxp  0x02370109        32       328
    name  0x0AD0EC64      1832     46088
    post  0x00010000        32     47920
```

miguelsousa:
Sure

miguelsousa:
New test font submitted
```
    tag     checksum    length    offset
    ----  ----------  --------  --------
    cmap  0x16D8EB36      1192      1828
    glyf  0x00000000         1      3804
    head  0x16A8DC50        54       140
    hhea  0x04E408B7        36       196
    hmtx  0xF0754561      1564       264
    loca  0x00000000       784      3020
    maxp  0x01880001        32       232
    post  0x00010000        32      3808
```

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -491,6 +491,13 @@ def getGlyphOrder(self):
 				# in combination with the Adobe Glyph List (AGL).
 				#
 				self._getGlyphNamesFromCmap()
+			elif len(glyphOrder) < self['maxp'].numGlyphs:
```
anthrotype, line 494: you could also merget this `elif` branch with the above `if` by doing `if glyphOrder is None or len(glyphOrder) < self.["maxp"].numGlyphs:` (while also moving the accompanying comments), since both are doing the same (i.e. calling `_getGlyphNamesFromCmap`)

anthrotype: or maybe actually, you should keep the `elif` but add a logging.warning to this one, since we are basically discarding the information in the post format1 table (first 258 glyphs named using standard mac order) and treating it as if it were format 3 (no names at all).

miguelsousa: Rebased on `main` and added warning log. PTAL

miguelsousa: Interestingly, the CI changes the logging level
https://github.com/fonttools/fonttools/runs/8096754180?check_suite_focus=true#step:5:387
What's the advice? Remove the `levelname` assert?

miguelsousa: Nevermind, figured out what was going on. There's another test that sets the logging level.

miguelsousa:
@anthrotype @justvanrossum OK to merge?

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -487,6 +487,14 @@ def getGlyphOrder(self):
...
+				#
+				# Not enough names found in the 'post' table.
+				# Can happen when 'post' format 1 is improperly used on a font that
+				# has more than 258 glyphs (the lenght of 'standardGlyphOrder').
+				#
+				log.warning("'post' table format 1 discarded; font has more than 258 glyphs")
```
anthrotype, line 496: the warning mentions format 1 or font having more than 258 glyphs, but I don't see that you actually check any of these two specific conditions. You simply test if the glyphOrder that returned from post table is smaller than the total number of glyphs in the font according to maxp. Could it also be an invalid truncated post format 2 that contains less glyphs than there are in maxp.numGlyphs? (i'm speculating, I actually don't know if that's even possible). Maybe better to simply warn "Not enough names found in the 'post' table, generating them from cmap instead".

PR #2788, opened 2022-08-29
[ttVarGlyphSet] Convert tabs to spaces

justvanrossum:
It's a new module, it's already been run through Black, but kept the tabs for easier review in #2784. Now convert tabs to spaces.


behdad:
LGTM.

PR #2789, opened 2022-08-29
[ttGlyphSet] Refactor/rebuild class hierarchy

justvanrossum:
- only differentiate between glyf and CFF/CFF2 implementations, not var vs non-var
- use collections.abc.Mapping to get a more consistent dict-like object with less code
- prefer HVAR metrics over gvar metrics
- move some responsibilities from the _TTGlyphSet objects to the _TTGlyph objects
- adjust some tests to the changes

### DIFF
--- a/Lib/fontTools/ttLib/ttFont.py
+++ b/Lib/fontTools/ttLib/ttFont.py
@@ -696,12 +696,11 @@ class TTFont(object):
 		axes space.
 		"""
 		if ("CFF " in self or "CFF2" in self) and (preferCFF or "glyf" not in self):
-			glyphSetClass = _TTGlyphSetCFF
+			return _TTGlyphSetCFF(self, location, normalized)
 		elif "glyf" in self:
-			glyphSetClass = _TTGlyphSetGlyf
+			return _TTGlyphSetGlyf(self, location, normalized)
 		else:
 			raise TTLibError("Font contains no outlines")
-		return glyphSetClass(self, location, normalized)
 
 	def getBestCmap(self, cmapPreferences=((3, 10), (0, 6), (0, 4), (3, 1), (0, 3), (0, 2), (0, 1), (0, 0))):
 		"""Returns the 'best' Unicode cmap dictionary available in the font
--- a/Lib/fontTools/ttLib/ttFont.py
+++ b/Lib/fontTools/ttLib/ttFont.py
@@ -4,7 +4,12 @@ from fontTools.misc.configTools import AbstractConfig
 from fontTools.misc.textTools import Tag, byteord, tostr
 from fontTools.misc.loggingTools import deprecateArgument
 from fontTools.ttLib import TTLibError
-from fontTools.ttLib.ttGlyphSet import _TTGlyph, _TTGlyphSetCFF, _TTGlyphSetGlyf
+from fontTools.ttLib.ttGlyphSet import (
+    _TTGlyph,
+    _TTGlyphSetCFF,
+    _TTGlyphSetGlyf,
+    _normalizeLocation,
+)
 from fontTools.ttLib.sfnt import SFNTReader, SFNTWriter
 from io import BytesIO, StringIO
 import os
@@ -695,10 +700,12 @@ class TTFont(object):
 		as in the normalized (-1..+1) space, otherwise it is in the font's defined
 		axes space.
 		"""
+		if location:
+			location = _normalizeLocation(location, self) if "fvar" in self else None
 		if ("CFF " in self or "CFF2" in self) and (preferCFF or "glyf" not in self):
-			return _TTGlyphSetCFF(self, location, normalized)
+			return _TTGlyphSetCFF(self, location)
 		elif "glyf" in self:
-			return _TTGlyphSetGlyf(self, location, normalized)
+			return _TTGlyphSetGlyf(self, location)
 		else:
 			raise TTLibError("Font contains no outlines")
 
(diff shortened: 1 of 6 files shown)


behdad (approved):
Looks great! Thanks.

justvanrossum:
I think I'm done fiddling now. I added `TTFont.normalizeLocation()` as a public method, as I expect it to be generally useful.

Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -1,158 +1,193 @@
...
+                    varStore.otVarStore, font["fvar"].axes, location
+                )
+                self.blender = instancer.interpolateFromDeltas
+
+    def __getitem__(self, glyphName):
+        return _TTGlyphCFF(self, glyphName)
 
 
-class _TTGlyph(object):
+class _TTGlyph:
```
anthrotype, line 78: maybe we could make _TTGlyph an abstract base class and define draw/drawPoints as abstract methods that must be overridden by subclasses

anthrotype: FWIW, also `_TTGlyphSet` in theory is abstract, with particular subclasses needing to override `__getitem__` at least

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -684,44 +679,68 @@ def getTableData(self, tag):
...
+		else:
+			raise TTLibError("Font contains no outlines")
 
-		if glyphs is None and "glyf" in self:
-			if location and 'gvar' in self:
-				glyphs = _TTVarGlyphSet(self, self["glyf"], _TTVarGlyphGlyf,
-										location, normalized)
-			else:
-				glyphs = _TTGlyphSet(self, self["glyf"], _TTGlyphGlyf)
+	def normalizeLocation(self, location):
```
anthrotype, line 714: nice, maybe we could reuse this in the instancer as well

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -4,12 +4,7 @@
 from fontTools.misc.textTools import Tag, byteord, tostr
 from fontTools.misc.loggingTools import deprecateArgument
 from fontTools.ttLib import TTLibError
-from fontTools.ttLib.ttGlyphSet import (
-    _TTGlyphSet, _TTGlyph,
-    _TTGlyphCFF, _TTGlyphGlyf,
```
anthrotype, line 9: I wonder if we should keep the `_TTGlyphSet`, `_TTGlyphCFF` and `_TTGlyphGlyf` symbols here, even if unused. There may be client code that may expect to find these in `ttFont` module, as they were there before things where moved to the new `ttGlyphSet`. I know they are underscore prefixed and thus private, but one may still be using them in typing annotations for function parameters/return values that include those.

anthrotype (approved):
LGTM, with some non-blocking comments

Lib/fontTools/ttLib/ttGlyphSet.py
justvanrossum: Ah yes, makes sense for `__TTGlyph`, will do.

`_TTGlyphSet` is already an actual abstract baseclass, via `collections.abc.Mapping`.

Lib/fontTools/ttLib/ttFont.py
justvanrossum: The only reason `_TTGlyph` is there is because it's used as a typing thing in the remove overlaps module. Even there I wonder: does it need to be this concrete (but abstract) class, or should we define our own abstract typing declarations? "What is a glyph set" / "What is a glyph".

I don't want to keep more "just in case", unless there's actual client code that uses these (but really shouldn't).

anthrotype: > should we define our own abstract typing declarations? "What is a glyph set" / "What is a glyph".

yeah maybe we could do that, perhaps using typing.Protocol.

anthrotype: e.g. I found code like this one in our internal codebase:

```python
  def to_ttglyph(self, character: Text) -> ttFont._TTGlyph:
    """Converts a character to its associated ttFont._TTGlyph.
    ...
```

what should client code do in this case?

Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -75,7 +76,7 @@ def __getitem__(self, glyphName):
         return _TTGlyphCFF(self, glyphName)
 
 
-class _TTGlyph:
+class _TTGlyph(metaclass=ABCMeta):
```
anthrotype, line 79: I tend to use the simpler `class _TTGlyph(ABC)` but this is basically the same, so it's ok

anthrotype (approved):

Lib/fontTools/ttLib/ttGlyphSet.py
justvanrossum: Ah yes, that's better, thanks.

Lib/fontTools/ttLib/ttFont.py
justvanrossum: I don't really know. We could rename it `TTGlyph`, and the client code should ideally import it from ttGlyphSet.py instead of from ttFont.py.

What aspects of the `_TTGlyph` instance does the caller of `to_ttglyph()` actually use? Is it expecting all kinds of internals of that object, or does it expect draw/drawPoints methods and maybe the metrics attrs?

simoncozens:
This broke a lot of software. Deprecation would have been helpful.

PR #2792, opened 2022-08-30
[otlLib] In PairPos, always build one class-based subtable

behdad:
Fixes https://github.com/fonttools/fonttools/issues/2772


simoncozens:
#2776 is this plus tests, right?

anthrotype:
closing as duplicate of #2776

PR #2794, opened 2022-08-31
[gvar] be non lazy for TTFont.lazy=False, add ensureDecompiled

anthrotype:
like cmap, or glyf or OTL tables, we need to respect lazy=False flag and decompile everything upfront, also we want to add an ensureDecompiled method (called by TTFont.ensureDecompiled) to allow loading everything in one go even when a font had been opened lazily.


behdad:
LGTM.

PR #2797, opened 2022-09-01
Avoid crash when handling unbounded conditions during designspace splitting

Hoolean:
## Context

Designspace documents can contain unbounded conditions, which [omit one of "minimum" or "maximum"](https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#condition-element). In fontTools, the absence of either is represented with None, both when designspace files [are parsed](https://github.com/fonttools/fonttools/blob/1306a71/Lib/fontTools/designspaceLib/__init__.py#L1838-L1850), and when designspace objects [are processed in varLib](https://github.com/fonttools/fonttools/blob/bb6fd8d22e7aa20d8ae6553c15f5686fbaf7d862/Lib/fontTools/varLib/__init__.py#L670-L677).

## Problem

For unbounded conditions, the splitting functions expect "minimum" or "maximum" [to be absent](https://github.com/fonttools/fonttools/blob/1306a71db323556232b1f5ece1c2ab77476d0884/Lib/fontTools/designspaceLib/split.py#L356-L357), instead of present with a value of None. This means that the None values are never substituted by an appropriate value, and so propagate further until the program crashes.

ufo2ft [uses splitting](https://github.com/googlefonts/ufo2ft/blob/6d618f6/Lib/ufo2ft/__init__.py#L750) as part of its build pipeline, and so unbounded conditions break the build process downstream.

## Fix

This PR changes the splitting code to expect "minimum" and "maximum" to always be present, with potential values of None. This makes the behaviour consistent with the rest of the library, and fixes the crash.

In addition, a test has been added to check for regressions.

### Caveat

At least [one area of the existing code](https://github.com/fonttools/fonttools/blob/1306a71/Lib/fontTools/designspaceLib/__init__.py#L1422-L1430) accounts for the values being None AND the keys being absent. This is probably overkill, as the codebase is already heavily dependent on both keys always being present. For this reason, checking for both cases was avoided here.

## Future work

We could type the conditions in DesignSpaceDocument to avoid issues like this in the future. In addition, we could add tests downstream that compile projects with unbounded conditions, as this would have highlighted the issue sooner.

### DIFF
--- a/Lib/fontTools/designspaceLib/split.py
+++ b/Lib/fontTools/designspaceLib/split.py
@@ -352,7 +352,7 @@ def _extractSubSpace(
 def _conditionSetFrom(conditionSet: List[Dict[str, Any]]) -> ConditionSet:
     c: Dict[str, Range] = {}
     for condition in conditionSet:
-        minimum, maximum = condition["minimum"], condition["maximum"]
+        minimum, maximum = condition.get("minimum"), condition.get("maximum")
         c[condition["name"]] = Range(
             minimum if minimum is not None else -math.inf,
             maximum if maximum is not None else math.inf,
--- a/Lib/fontTools/designspaceLib/split.py
+++ b/Lib/fontTools/designspaceLib/split.py
@@ -352,9 +352,10 @@ def _extractSubSpace(
 def _conditionSetFrom(conditionSet: List[Dict[str, Any]]) -> ConditionSet:
     c: Dict[str, Range] = {}
     for condition in conditionSet:
+        minimum, maximum = condition["minimum"], condition["maximum"]
         c[condition["name"]] = Range(
-            condition.get("minimum", -math.inf),
-            condition.get("maximum", math.inf),
+            minimum if minimum is not None else -math.inf,
+            maximum if maximum is not None else math.inf,
         )
     return c
 
--- a/Tests/designspaceLib/split_test.py
+++ b/Tests/designspaceLib/split_test.py
@@ -160,13 +160,15 @@ def test_convert5to4(datadir, tmpdir, test_ds, expected_vfs):
 @pytest.mark.parametrize(
     ["unbounded_condition"],
     [
+        ({"name": "Weight", "minimum": 500},),
+        ({"name": "Weight", "maximum": 500},),
         ({"name": "Weight", "minimum": 500, "maximum": None},),
         ({"name": "Weight", "minimum": None, "maximum": 500},),
     ],
 )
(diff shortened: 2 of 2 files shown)


Lib/fontTools/designspaceLib/split.py
```diff
@@ -352,9 +352,10 @@ def maybeExpandDesignLocation(object):
 def _conditionSetFrom(conditionSet: List[Dict[str, Any]]) -> ConditionSet:
     c: Dict[str, Range] = {}
     for condition in conditionSet:
+        minimum, maximum = condition["minimum"], condition["maximum"]
```
anthrotype, line 355: should this allow for the "maximum" or "minimum" keys to actually be missing from the `condition` dict? Or are they guaranteed to be there, sometimes with a `None` value? I'm actually not sure who creates these `condition` dicts. But it's probably safe to do it nonetheless, as that seemed to be the original intent of this code.

```suggestion
        minimum, maximum = condition.get("minimum"), condition.get("maximum")
```

Hoolean: Hi @anthrotype :)

These are the condition dicts from `DesignSpaceDocument.rules`. The keys are guaranteed to be there [when created through designspace parsing](https://github.com/fonttools/fonttools/blob/1306a71/Lib/fontTools/designspaceLib/__init__.py#L1838-L1850), and varLib [assumes that they are always there](https://github.com/fonttools/fonttools/blob/bb6fd8d22e7aa20d8ae6553c15f5686fbaf7d862/Lib/fontTools/varLib/__init__.py#L670-L677) too. This is what made me feel comfortable taking a similar approach.

With that said, other areas of the code, such as [rule evaluation](https://github.com/fonttools/fonttools/blob/1306a71/Lib/fontTools/designspaceLib/__init__.py#L395-L402) and [writing to XML](https://github.com/fonttools/fonttools/blob/1306a71/Lib/fontTools/designspaceLib/__init__.py#L1422-L1430), do perform this extra check.

I agree that there is no harm in including it, as it feels like a stronger guarantee of safety to me too, and bears no cost :+1: With that said, we should be mindful that other areas (e.g. varLib) might be more picky about the dicts they take.

Shall I add it in?

anthrotype: yeah, let's use `dict.get` thanks

Hoolean: Super! I have added the checks, and a few test cases to make sure this stays somewhat supported too.

anthrotype (approved):

PR #2798, opened 2022-09-05
Speed up varscalar with caching

simoncozens:
This speeds up variable scalar compilation by caching computed variation models and also sharing them between variable scalars with the same masters. Since in an ordinary font workflow there will be one fixed set of masters and hence one variation model, this saves a lot of varLib.models compilation time. When compiling Roboto Serif with a variable feature, this PR takes compilation time down from 170 seconds to 70 seconds.

### DIFF
--- a/Lib/fontTools/feaLib/variableScalar.py
+++ b/Lib/fontTools/feaLib/variableScalar.py
@@ -1,4 +1,11 @@
 from fontTools.varLib.models import VariationModel, normalizeValue
+from functools import cached_property, lru_cache
+
+# We will often use exactly the same locations (i.e. the font's
+# masters) for a large number of variable scalars. Instead of
+# creating a model for each, let's share the models.
+model_pool = {}
+
 
 
 def Location(loc):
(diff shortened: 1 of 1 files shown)


Lib/fontTools/feaLib/variableScalar.py
```diff
@@ -1,4 +1,11 @@
 from fontTools.varLib.models import VariationModel, normalizeValue
+from functools import cached_property, lru_cache
+
+# We will often use exactly the same locations (i.e. the font's
+# masters) for a large number of variable scalars. Instead of
+# creating a model for each, let's share the models.
+model_pool = {}
```
belluzj, line 7: Maybe find a right time to empty this dict, or cap its size, so it doesn't become a memory leak in a potential scenario of using fontTools in a long-running Python process?

Lib/fontTools/feaLib/variableScalar.py
```diff
@@ -81,11 +88,17 @@ def value_at_location(self, location):
...
-        return VariationModel(locations)
-
+        key = tuple([tuple(x.items()) for x in locations])
+        if key in model_pool:
+            return model_pool[key]
+        m = VariationModel(locations)
+        model_pool[key] = m
+        return m
+
+    @lru_cache
```
belluzj, line 101: Is `self.values` mutable somehow? If not, then this could also be just @cached_property

Lib/fontTools/feaLib/variableScalar.py
anthrotype: maybe make it a class property of VariableScalar and add a class method to clear_cache() so caller can dump it when finished. Otherwise these will be kept alive forever

Lib/fontTools/feaLib/variableScalar.py
```diff
@@ -1,4 +1,5 @@
 from fontTools.varLib.models import VariationModel, normalizeValue
+from functools import lru_cache
```
anthrotype, line 2: unused

anthrotype (approved):

Issue #2800, opened 2022-09-05
[COLRv1] Drop DeltaSetIndexMap when not needed


drott:
In an experimental font generated with:
https://github.com/drott/color-fonts/tree/noDeltaSetIndexMap
fontTools creates and empty DeltaSetIndexMap. 

```
    <VarIndexMap Format="0">
      <!-- Omitted values default to 0xFFFF/0xFFFF (no variations) -->
      <Map index="0" outer="0" inner="0"/>
    </VarIndexMap>
```

I suggest to drop this when not needed for saving space. 

I am in the process of fixing this part in [FreeType in this issue (#1178)](https://gitlab.freedesktop.org/freetype/freetype/-/issues/1178).

PS: FF nightly seems to support those already.

anthrotype:
unless, of course, the VarStore VarData items are interspersed with lots of zeros to replace those NO_VARIATION_DATA but I'm not actually sure that would produce a smaller COLR table in the end (as opposed to an optimized VarStore plus an indirect DeltaSetIndexMap)

PR #2803, opened 2022-09-05
don't add no-op DeltaSetIndexMap when not needed

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/2800

### DIFF
--- a/Tests/varLib/varLib_test.py
+++ b/Tests/varLib/varLib_test.py
@@ -1,7 +1,10 @@
+from fontTools.colorLib.builder import buildCOLR
 from fontTools.ttLib import TTFont, newTable
-from fontTools.varLib import build, load_designspace
+from fontTools.ttLib.tables import otTables as ot
+from fontTools.varLib import build, load_designspace, _add_COLR
 from fontTools.varLib.errors import VarLibValidationError
 import fontTools.varLib.errors as varLibErrors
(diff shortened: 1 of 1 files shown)


behdad:
Lgtm.

Lib/fontTools/varLib/__init__.py
```diff
@@ -722,8 +722,11 @@ def _add_COLR(font, model, master_fonts, axisTags, colr_layer_reuse=True):
 	if store:
 		mapping = store.optimize()
 		colr.VarStore = store
+		# don't add DeltaSetIndexMap for identity mapping
+		colr.VarIndexMap = None
 		varIdxes = [mapping[v] for v in merger.varIdxes]
-		colr.VarIndexMap = builder.buildDeltaSetIndexMap(varIdxes)
+		if any(i != varIdxes[i] for i in range(len(varIdxes))):
+			colr.VarIndexMap = builder.buildDeltaSetIndexMap(varIdxes)
```
behdad, line 729: Thinking more, maybe we move that functionality into `buildDeltaSetIndexMap`, activated with a new parameter for backward compatibility, or default-enabled after auditing all places it is used.

anthrotype: hm.. currently, all the varLib.builders functions return `self`, none of them ever returns `None`. I think it's fine like this

PR #2811, opened 2022-09-12
[instancer] downgrade GSUB/GPOS version if there are no FeatureVariations after instancing

qxliu76:


anthrotype:
thanks! I simplified it a bit and added a test in https://github.com/fonttools/fonttools/pull/2812

Issue #2763, opened 2022-08-22
codecov fails randomly with error about missing repository upload token


anthrotype:
E.g. see 

```
[2022-08-22T16:02:20.777Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
```

Re-running the failed jobs works most of the times, but it's annoying. Anybody knows how to properly fix this?

PR #2825, opened 2022-09-20
Fix for #2824: fix arguments in calls to (glyf) glyph.draw() and glyph.drawPoints()

### LINKED ISSUE #2824: [ttGlyphSet] glyphSet + offset not passed correctly to glyf glyph
Reported here: https://github.com/adobe-type-tools/afdko/issues/1560

justvanrossum:
This fix also exposed a second bug, where lsb and tsb were not set.

Fixes #2824


anthrotype:
I mean, I do see this is the right fix but I wonder how come it didn't make https://github.com/adobe-type-tools/afdko/issues/1560 crash

justvanrossum:
I think I can add a test by grabbing Adobe's test font.

justvanrossum:
I modified the font a bit so it was easier to integrate into the existing test parameters.

anthrotype (approved):
thanks

anthrotype:
but getCoordinates, called by glyf.Glyph.drawPoints uses it, to follow up component references, no?

https://github.com/fonttools/fonttools/blob/796c1f13303f0af515166e95bff1191f8d4e7f16/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L1078

justvanrossum:
The offset is needed when `lsb` doesn't match the glyph's `xMin`.

The "real" problems were:
- the offset wasn't calculated correctly for VS
- the offset wasn't passed to `glyph.draw()`

The `glyphSet` argument is effectively ignored, so there was no crash.

The Adobe test font contains a glyph that has an `lsb` of 30, but an `xMin` of 25, so when drawing, the outline needs to be offset to the right by 5 units.

We didn't test that case, but now we do.

justvanrossum:
Yes, but there is an earlier switch on components, so that code is never reached:

https://github.com/fonttools/fonttools/blob/796c1f13303f0af515166e95bff1191f8d4e7f16/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L1260-L1264

anthrotype:
ah, thanks for the clarifications! Then I think we should prehaps deprecate the unused parameter and remove it altogether from the glyf table draw/drawPoints interface (not in this PR, maybe we should file an issue about that too if there isn't one already).

xMin != lsb is relatively uncommon (I believe ufo2ft always makes sure they align) so I'm less worried about the issue now.
I think we can merge this. I can do a quick release now nonetheless.

anthrotype:
@justvanrossum maybe you can add a comment about the fact that "The Adobe test font contains a glyph that has an lsb of 30, but an xMin of 25"?

anthrotype:
you merged already, no problem then. Let's just update the mentioned issue https://github.com/fonttools/fonttools/issues/2824 with these additional details and keep the test code as is

justvanrossum:
I can still add a comment on main, good point actually.

anthrotype:
ok I'll make the release in a bit, after you updated the code, thanks

justvanrossum:
Done: https://github.com/fonttools/fonttools/commit/6e6b86d41a72ac398b78dd0aba4d3051b96741f8

justvanrossum:
Another Codecov glitch? https://github.com/fonttools/fonttools/actions/runs/3088881083/jobs/4995869043

anthrotype:
yeah, re-running the failed jobs usually works.. I have no idea #2763

josh-hadley:
@anthrotype @justvanrossum thanks for your efforts on this — I appreciate the speedy response!

justvanrossum:
@josh-hadley apologies for the breakage, and thanks for the report!

PR #2827, opened 2022-09-20
Avoid negative hmtx values when creating font from variable CFF2 font

### LINKED ISSUE #2821: Calling instantiateVariableFont can lead to negative advance widths in hmtx
Hi!

Calling `instantiateVariableFont` can introduce negative advance widths in the `hmtx` table:

https://github.com/fonttools/fonttools/blob/796c1f13303f0af515166e95bff1191f8d4e7f16/Lib/fontTools/varLib/mutator.py#L121-L149

We could probably replace line 146 by `entry[0] = max(0, entry[0] + width_delta)`. I can open a PR if you want.

(Carefully crafted fonts shouldn’t trigger this problem, but I’ve got it using a subset font where advance of unused glyphes has been set to 0.)

liZe:
Fix #2821.

### DIFF
--- a/Lib/fontTools/varLib/mutator.py
+++ b/Lib/fontTools/varLib/mutator.py
@@ -143,7 +143,7 @@ def interpolate_cff2_metrics(varfont, topDict, glyphOrder, loc):
 
 		if lsb_delta or width_delta:
 			if width_delta:
-				entry[0] += width_delta
+				entry[0] = max(0, entry[0] + width_delta)
 			if lsb_delta:
 				entry[1] = lsb
 			hmtx[gname] = tuple(entry)


behdad:
Fix looks arbitrary though. This is speced nowhere.

anthrotype:
well, hmtx advance can't be negative, they are encoded as unsigned 16-bit integer, no?

behdad:
> well, hmtx advance can't be negative, they are encoded as unsigned 16-bit integer, no?

Correct. Though applications might interpret them otherwise. I'd say this is a case of a bad font. The fact that the advance is encoded as unsigned in hmtx is orthogonal to whether the combined advance has to be nonnegative.

behdad:
We have other places in the format also that the variations can take a value outside the limits of its original value.

anthrotype:
FWIW we also do a similar thing for truetype-flavored VFs in the varLib.instancer, see

https://github.com/fonttools/fonttools/blob/a2eef358be401cfa1c783f24f48a000424457152/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L435-L438

behdad:
Okay I see; I forgot we are talking about instancing. Nevermind.

Issue #2831, opened 2022-09-21
[varLib] Optimize gvar isComposite workaround hack


behdad:
In https://github.com/fonttools/fonttools/commit/9c3dde3ce6be74b7d06127c2d9a0c17484d638a0 we made varLib write gvar variation entries for composite glyphs even if they are all-zero vectors.

This is overkill. The requirement from the bug it's trying to fix is that composite glyphs have *some* entry in gvar table. So we just need to write one variation entry per composite glyph, not for all masters. For fonts with many masters the empty entries add up.

(I found this while discussing https://github.com/fonttools/fonttools/discussions/2830)

PR #2832, opened 2022-09-21
[varLib] Optimize gvar isComposite workaround hack

behdad:
In 9c3dde3 we made varLib write gvar variation entries for composite glyphs even if they are all-zero vectors.

This is overkill. The requirement from the bug it's trying to fix is that composite glyphs have some entry in gvar table. So we just need to write one variation entry per composite glyph, not for all masters. For fonts with many masters the empty entries add up.

Fixes https://github.com/fonttools/fonttools/issues/2831

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -272,8 +272,10 @@ def _add_gvar(font, masterModel, master_ttfs, tolerance=0.5, optimize=True):
 		endPts = control.endPts
 
 		for i,(delta,support) in enumerate(zip(deltas[1:], supports[1:])):
+			# For Composite glyphs we want to write at least one entry.
 			if all(v == 0 for v in delta.array) and not isComposite:
 				continue
+			isComposite = False
 			var = TupleVariation(support, delta)
 			if optimize:
 				delta_opt = iup_delta_optimize(delta, origCoords, endPts, tolerance=tolerance)


anthrotype:
makes sense. Did you or someone else verify that this workaround still works in CoreText and doesn't break anything else?

anthrotype:
@Lorp commented in https://github.com/fonttools/fonttools/issues/1381#issuecomment-1217119263

> This bug in macOS 12.4 seems to apply to fonts that have the workaround:
> [Variable fonts rendering bug: composite glyphs disappear when all deltas are zero (https://feedbackassistant.apple.com/feedback/11009662)

unfortunately the link isn't publicly accessible. Laurence, could you provide more details here?

jenskutilek:
[This still applies](https://github.com/fonttools/fonttools/pull/1788#issuecomment-1217120235):

> Since it was a problem introduced with macOS 10.14, and was apparently fixed latest in 10.14.6, the time to remove the workaround should be now...

IMHO there is no reason somebody who can run macOS 10.14 on their machines can't update to 10.14.6. We're talking about Macs manufactured in 2012 here. The original workaround should be removed.

anthrotype:
oh I missed that, so you're saying this bug that this workaround was meant to address can no longer be reproduced?

anthrotype:
I'd be fine to completely remove the workaround, especially if this creates other problems on more recent macOSs as @Lorp suggested. I could only read the title of that "[Variable fonts rendering bug: composite glyphs disappear when all deltas are zero](https://feedbackassistant.apple.com/feedback/11009662)" 
@jenskutilek can you confirm that? I don't have a mac at hand right now.

jenskutilek:
I can't see the feedback at Apple either.

I have made two test fonts, with and without the workaround, but can't reproduce the disappearing composites (in TextEdit on macOS 12.6 anyway). Perhaps I'm not doing it right ...

Bildschirmfoto 2022-09-22 um 15 48 06

[SudoGvarTest.zip](https://github.com/fonttools/fonttools/files/9626346/SudoGvarTest.zip)

jenskutilek:
@Lorp can you reproduce it with the fonts I attached?

behdad:
> makes sense. Did you or someone else verify that this workaround still works in CoreText and doesn't break anything else?

No I haven't.

Lorp:
Here is my [bug report to Apple Feedback Assistant](https://feedbackassistant.apple.com/feedback/11009662) posted 2022-08-02 (link may not work):

> I have encountered a bug in the Apple renderer for variable fonts such that composite glyphs *disappear* if their gvar data moves the x and y of all “points” by a zero delta. In gvar data for composites, the initial point IDs refer to component IDs, and the last 4 refer to metrics. Of course such glyphs don’t actually need deltas at all. However the widely used tool fontmake produces such fonts, which AFAIK are spec compliant. One can fix the issue on a per font basis by deleting the gvar entry for that glyph. Adding a single non-zero delta to any of the points also stops the glyph from disappearing. I am using TTX to show me what is in the gvar, and have not examined the precise encoding of the zero values. I presume that the issue would persist if the deltas were all zero for one delta set, but non-zero for another delta set in the same glyph — though I have not tested that.
>
> An important extra detail is that the issue does NOT occur at the default instance (wght=100). This means the issue does not occur in apps without variation control such as TextEdit. [edit: in TextEdit, indeed non-default instances show the problem]
>
> An easy way to reproduce the issue is to drag drag the font [attached] onto onto a website that attaches it as a webfoot, such as Axis-Praxis or Font Gauntlet. Then paste the characters ⬤⾦㊎ and adjust the Weight slider. You will see that the 3rd character disappears. Note that the issue does NOT occur in FontGoggles (which bypasses Apple rendering).
>
> [WhiteOnBlack-VF-zeroed-composite-gvar#1.ttf.zip](https://github.com/fonttools/fonttools/files/9627954/WhiteOnBlack-VF-zeroed-composite-gvar.1.ttf.zip)

My system is macOS 12.4.

On 2022-09-09 Apple posted a reply stating “Please verify this issue with macOS 13 beta 7 and update your bug report with your results”. **I do not have the ability to test the fix.** The status of the bug has changed to “Potential fix identified - In macOS 13”.

Lorp:
@jenskutilek the ÜÖÄ in your fonts do not show any obvious issues when moving axes away from defaults locations (macOS 12.4)

anthrotype:
Thanks Laurence. Just to double check, you can still reproduce the bug you originally reported in your current macOS version?

@jenskutilek can you also reproduce it using Laurence's font in the above linked zip file?

I think the string to display the glyphs in the font should be the following: ⬤⾦㊎

```
 ‎2B24 BLACK LARGE CIRCLE
 ‎2FA6 KANGXI RADICAL GOLD
 ‎328E CIRCLED IDEOGRAPH METAL
```

Again, didn't bring a mac with me today so can't test on my own

jenskutilek:
Yes, I can repro on macOS 12.6 with @Lorp’s font, but not with mine.

anthrotype:
weird. what's different between the two? Could you maybe record a short video of your screen, if that's not too much trouble? I know QuickTime Player can do that

behdad:
What's the conclusion here?

jenskutilek:
Sorry, I missed that. Here's a screen recording on macOS 12.6.1:


https://user-images.githubusercontent.com/3185950/198260230-fc653f93-45a8-4595-ad20-52a7b157e642.mp4

gvar with "zeroed" composites makes the component disappear in @Lorp’s font, but not in mine (shown both with and without forced zero gvar entries, no difference in behaviour).

There are many differences between the repro and non-repro fonts, for a start there are three masters in mine.

In my opinion, as the bug that required the original workaround of adding zero entries to gvar, has been fixed in macOS 10.14.6 (released July 22, 2019), the workaround should be removed.

anthrotype:
since we decided to merge #1788, this is no longer needed, hence closing

PR #2837, opened 2022-09-30
[subset] Keep nameIDs used by CPAL palette entry labels

jackmcDaMa:
Hello, I'm Jack from the SD team at Dalton Maag. We found this while trying to subset a font with CPALv1. I assume that it was just missing an enumerate() ?

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -2553,6 +2553,10 @@ def prune_pre_subset(self, font, options):
 		if stat.table.AxisValueArray:
 			nameIDs.update([val_rec.ValueNameID for val_rec in stat.table.AxisValueArray.AxisValue])
 		nameIDs.update([axis_rec.AxisNameID for axis_rec in stat.table.DesignAxisRecord.Axis])
+	cpal = font.get('CPAL')
+	if cpal and cpal.version == 1:
+		nameIDs.update(cpal.paletteLabels)
+		nameIDs.update(cpal.paletteEntryLabels)
 	if '*' not in options.name_IDs:
 		self.names = [n for n in self.names if n.nameID in nameIDs]
 	if not options.name_legacy:
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -2271,9 +2271,10 @@ def prune_post_subset(self, font, options):
 		name_table.names = [
 			n for n in name_table.names
 			if (
-				n.nameID not in dropped_labels or
+				n.nameID not in dropped_labels
 				# Only remove nameIDs in the user range and if they're not explicitly kept
-				n.nameID < 256 or n.nameID in options.name_IDs
+				or n.nameID < 256
+				or n.nameID in options.name_IDs
 			)
 		]
 	return bool(self.numPaletteEntries)
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -2258,8 +2258,23 @@ def prune_post_subset(self, font, options):
 	self.numPaletteEntries = len(self.palettes[0])
 
 	if self.version == 1:
-		self.paletteEntryLabels = [
-			label for i, label in enumerate(self.paletteEntryLabels) if i in retained_palette_indices
+		kept_labels = []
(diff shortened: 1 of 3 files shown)


belluzj:
@anthrotype do you think this needs a test?

anthrotype:
ouch, thanks! Would you mind adding a little test as well?

anthrotype:
the `colrv1_path` fixture used in the Tests/subset/subset_test.py module uses fontBuilder API to setupCPAL, you can add some palette lables in there I think

https://github.com/fonttools/fonttools/blob/bf6078af3e836e99e741debf3aa8a805ff288b53/Tests/subset/subset_test.py#L1232

https://github.com/fonttools/fonttools/blob/bf6078af3e836e99e741debf3aa8a805ff288b53/Lib/fontTools/fontBuilder.py#L860-L866

jackmcDaMa:
I've added a test but it fails. It appears that maybe the nameIDs from the CPAL table are not taken into account and the actual name strings are dropped from the name table?
image

Should we do something about CPAL in the subset code? For example here:
image

anthrotype:
> Should we do something about CPAL in the subset code? 

you're right, yes we should!

anthrotype:
the prune_pre_subset should keep all nameIDs that are referenced by the CPAL table, but then probably we want also a prune_post_subset to remove the ones that are no longer used..

anthrotype:
1) in `prune_pre_subset` for the `name` table, like we do for `fvar` or `STAT` collect the set of all the nameIDs used by the CPAL table, we want to keep them all by default (unless explicitly excluded by --name-IDs CLI option). It looks like we currently drop them all (LOL!).
2) inside CPAL's `prune_post_subset`, immediately as we prune `paletteEntryLabels` we also collect the nameIDs for the entry labels that we are dropping, and remove those from the `name` table.

jackmcDaMa:
@anthrotype It is ready for review now

Lib/fontTools/subset/__init__.py
```diff
@@ -2258,8 +2258,23 @@ def collect_colors_by_index(paint):
...
+		dropped_labels = []
+		for i, label in enumerate(self.paletteEntryLabels):
+			if i in retained_palette_indices:
+				kept_labels.append(label)
+			else:
+				dropped_labels.append(label)
+		self.paletteEntryLabels = kept_labels
+		# Remove dropped labels from the name table.
+		name_table = font["name"]
+		name_table.names = [
```
anthrotype, line 2271: with my black hat on, I think the `or` should be placed on the next line, one condition per line

```python
name_table.names = [
    n
    for n in name_table.names
    if (
        n.nameID not in dropped_labels
        # Only remove nameIDs in the user range and if they're not explicitly kept
        or n.nameID < 256
        or n.nameID in options.name_IDs
    )
]
```

Tests/subset/subset_test.py
```diff
@@ -1364,6 +1390,77 @@ def test_subset_COLRv1_and_CPALv1(colrv1_cpalv1_path):
         "second color",
         "third color",
     ]
+    # check that the "first color" name is dropped from name table
+    font = TTFont(colrv1_cpalv1_path)
+    for n in font["name"].names:
+        if n.toUnicode() == "first color":
+            first_color_nameID = n.nameID
+            break
+    assert first_color_nameID is not None
```
anthrotype, line 1399: you should initialize `first_color_nameID` variable to None before the `for` loop

Tests/subset/subset_test.py
```diff
@@ -1364,6 +1390,77 @@ def test_subset_COLRv1_and_CPALv1(colrv1_cpalv1_path):
...
+    subset_path = colrv1_cpalv1_path.parent / (colrv1_cpalv1_path.name + ".subset")
+
+    # figure out the name ID of first color so we can keep it
+    font = TTFont(colrv1_cpalv1_path)
+
+    for n in font["name"].names:
+        if n.toUnicode() == "first color":
+            first_color_nameID = n.nameID
+            break
+    assert first_color_nameID is not None
```
anthrotype, line 1413: same here, set `first_color_nameID = None` before the loop, otherwise if it doesn't `break` it's a NameError

anthrotype (approved):

PR #2846, opened 2022-10-12
[varLib.models] Attempt to fix #2843 by computing the axis ranges

### LINKED ISSUE #2843: variationModel.interpolateFromMasters doesn't extrapolate as expected
With `extrapolate=True` there is an issue with`variationModel.interpolateFromMasters` as it clips results in 2 value models. Results are as expected in 3 value models. 

```python
from fontTools.varLib.models import VariationModel

# extrapolate under and over a 3 value system
locations = [
    dict(wght=-1),
    dict(wght=0),
    dict(wght=1),    
]

values = [30, 40, 50]

m = VariationModel(locations, extrapolate=True)
assert m.interpolateFromMasters(dict(wght=-1.5), values) == 25
assert m.interpolateFromMasters(dict(wght=-1), values) == 30
assert m.interpolateFromMasters(dict(wght=0), values) == 40
assert m.interpolateFromMasters(dict(wght=0.5), values) == 45
assert m.interpolateFromMasters(dict(wght=1), values) == 50
assert m.interpolateFromMasters(dict(wght=1.5), values) == 55


# extrapolate under and over a 2 value system
locations = [
    dict(wght=0),
    dict(wght=1),    
]

values = [40, 50]

m = VariationModel(locations, extrapolate=True)

# values under 0 are clipped.
print(m.interpolateFromMasters(dict(wght=-1), values))
print(m.interpolateFromMasters(dict(wght=-2), values))
print(m.interpolateFromMasters(dict(wght=-3), values))

#> 40.0
#> 40.0
#> 40.0

# this one fails
#assert m.interpolateFromMasters(dict(wght=-1), values) == 30

# the rest works as expected
assert m.interpolateFromMasters(dict(wght=0), values) == 40
assert m.interpolateFromMasters(dict(wght=0.5), values) == 45
assert m.interpolateFromMasters(dict(wght=1), values) == 50
assert m.interpolateFromMasters(dict(wght=1.5), values) == 55
```

justvanrossum:
The problem @LettError pointed out in #2843 could be resolved if `supportScalar()` uses the axis min/max values, instead of hard-coding them to -1 and +1.

With this change, the snippet below shows extrapolations as expected:
```python
from fontTools.varLib.models import VariationModel


def formatRow(row, cellWidth=8):
    fmt = f"{{v:>{cellWidth}}}"
    return "".join(fmt.format(v=v) for v in row)


locations = [{}, {"x": 1}, {"y": 1}, {"x": 1, "y": 1}]

model = VariationModel(locations, extrapolate=True)

values = [100, 200, 300, 400]

rng = [-1, 0, 1, 2]

print(formatRow([""] + rng))
for x in rng:
    row = [x]
    for y in rng:
        v = model.interpolateFromMasters({"x": x, "y": y}, values)
        row.append(v)
    print(formatRow(row))
```

Output:
```
              -1       0       1       2
      -1  -200.0     0.0   200.0   400.0
       0  -100.0   100.0   300.0   500.0
       1     0.0   200.0   400.0   600.0
       2   100.0   300.0   500.0   700.0
```

I think this keeps the original functionality as intended by @behdad, it just assumes less of the input design space.

I also believe this lifts the requirement for the locations to be normalized when extrapolating.

### DIFF
--- a/Tests/varLib/models_test.py
+++ b/Tests/varLib/models_test.py
@@ -36,10 +36,12 @@ def test_supportScalar():
     assert supportScalar({"wght": 0.2}, {}) == 1.0
     assert supportScalar({"wght": 0.2}, {"wght": (0, 2, 3)}) == 0.1
     assert supportScalar({"wght": 2.5}, {"wght": (0, 2, 4)}) == 0.75
-    assert supportScalar({"wght": 4}, {"wght": (0, 2, 2)}) == 0.0
-    assert supportScalar({"wght": 4}, {"wght": (0, 2, 2)}, extrapolate=True) == 2.0
-    assert supportScalar({"wght": 4}, {"wght": (0, 2, 3)}, extrapolate=True) == 2.0
-    assert supportScalar({"wght": 2}, {"wght": (0, 0.75, 1)}, extrapolate=True) == -4.0
+    assert supportScalar({"wght": 3}, {"wght": (0, 2, 2)}) == 0.0
+    assert supportScalar({"wght": 3}, {"wght": (0, 2, 2)}, extrapolate=True, axisRanges={"wght": (0, 2)}) == 1.5
+    assert supportScalar({"wght": -1}, {"wght": (0, 2, 2)}, extrapolate=True, axisRanges={"wght": (0, 2)}) == -0.5
+    assert supportScalar({"wght": 3}, {"wght": (0, 1, 2)}, extrapolate=True, axisRanges={"wght": (0, 2)}) == -1.0
+    assert supportScalar({"wght": -1}, {"wght": (0, 1, 2)}, extrapolate=True, axisRanges={"wght": (0, 2)}) == -1.0
+    assert supportScalar({"wght": 2}, {"wght": (0, 0.75, 1)}, extrapolate=True, axisRanges={"wght": (0, 1)}) == -4.0
 
 
 @pytest.mark.parametrize(
--- a/Tests/varLib/models_test.py
+++ b/Tests/varLib/models_test.py
@@ -46,6 +46,34 @@ def test_supportScalar():
         supportScalar({"wght": 2}, {"wght": (0, 0.75, 1)}, extrapolate=True, axisRanges=None)
 
 
+def test_model_extrapolate():
+    locations = [{}, {"a": 1}, {"b": 1}, {"a": 1, "b": 1}]
+    model = VariationModel(locations, extrapolate=True)
+    masterValues = [
+     100, 200,
+     300, 400]
+    testLocsAndValues = [
+        ({"a": -1, "b": -1}, -200),
+        ({"a": -1, "b": 0}, 0),
+        ({"a": -1, "b": 1}, 200),
+        ({"a": -1, "b": 2}, 400),
+        ({"a": 0, "b": -1}, -100),
+        ({"a": 0, "b": 0}, 100),
+        ({"a": 0, "b": 1}, 300),
+        ({"a": 0, "b": 2}, 500),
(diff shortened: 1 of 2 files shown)


behdad:
Thanks. This is the fix I had in mind as well.

> I also believe this lifts the requirement for the locations to be normalized when extrapolating.

Correct. Please update the comments and tests.

LGTM Otherwise.

justvanrossum:
Cool, thanks.

I'm considering dropping the fallback to -1/+1 here:
```python
axisMin, axisMax = axisRanges.get(axis, (-1, +1))
```
and simply require that axisRanges is filled in completely.

Tests/varLib/models_test.py
```diff
@@ -36,10 +36,12 @@ def test_supportScalar():
...
-    assert supportScalar({"wght": 4}, {"wght": (0, 2, 2)}) == 0.0
-    assert supportScalar({"wght": 4}, {"wght": (0, 2, 2)}, extrapolate=True) == 2.0
-    assert supportScalar({"wght": 4}, {"wght": (0, 2, 3)}, extrapolate=True) == 2.0
-    assert supportScalar({"wght": 2}, {"wght": (0, 0.75, 1)}, extrapolate=True) == -4.0
+    assert supportScalar({"wght": 3}, {"wght": (0, 2, 2)}) == 0.0
+    assert supportScalar({"wght": 3}, {"wght": (0, 2, 2)}, extrapolate=True, axisRanges={"wght": (0, 2)}) == 1.5
+    assert supportScalar({"wght": -1}, {"wght": (0, 2, 2)}, extrapolate=True, axisRanges={"wght": (0, 2)}) == -0.5
+    assert supportScalar({"wght": 3}, {"wght": (0, 1, 2)}, extrapolate=True, axisRanges={"wght": (0, 2)}) == -1.0
+    assert supportScalar({"wght": -1}, {"wght": (0, 1, 2)}, extrapolate=True, axisRanges={"wght": (0, 2)}) == -1.0
+    assert supportScalar({"wght": 2}, {"wght": (0, 0.75, 1)}, extrapolate=True, axisRanges={"wght": (0, 1)}) == -4.0
```
anthrotype, line 44: should you add a test for extrapolate=True and axisRanges=None? Or if that's never possible, maybe assert it isn't None?

justvanrossum: Good point, will fix.

anthrotype: also, it appears we don't have any tests that construct VariationModel with extrapolate=True, maybe you could add a few?

anthrotype: and also maybe some tests with multiple axes (which behdad said extrapolation algorithm may get trickier), I see we only tests with one wght axis

anthrotype (approved):
LGTM

behdad:
So, this changed things in the case of not extrapolating. If I have only a non-default master at wght=.5, I expect the tent to be 0,.5,1. But with this change we now get 0,.5,.5. We basically made it a requirement that axis extremes have a master.

behdad:
> So, this changed things in the case of not extrapolating. If I have only a non-default master at wght=.5, I expect the tent to be 0,.5,1. But with this change we now get 0,.5,.5. We basically made it a requirement that axis extremes have a master.

Actually it might not have been this change. Looks like we have some duplicate code as well:
https://github.com/fonttools/fonttools/blob/ba20aa2f56c90da06000aac7110a99ccfacb8239/Lib/fontTools/varLib/models.py#L427-L435

https://github.com/fonttools/fonttools/blob/ba20aa2f56c90da06000aac7110a99ccfacb8239/Lib/fontTools/varLib/models.py#L293-L302

PR #2847, opened 2022-10-13
Allow extrapolation in normalization

simoncozens:
We are allowing extrapolation in the variation model, but when we normalise values, these are still clamped to the axis min/max (-1.0<=>1.0). This allows normalized values to extend beyond that range.

This may not be the right approach, because it looks like the extrapolation option of the variation model allows for values which aren't normalised but which just have a min/max range. But I don't understand that well enough to know how it handles defaults.

### DIFF
--- a/Lib/fontTools/varLib/models.py
+++ b/Lib/fontTools/varLib/models.py
@@ -43,7 +43,7 @@ def subList(truth, lst):
     return [l for l, t in zip(lst, truth) if t]
 
 
-def normalizeValue(v, triple):
+def normalizeValue(v, triple, extrapolate=False):
     """Normalizes value based on a min/default/max triple.
 
       >>> normalizeValue(400, (100, 400, 900))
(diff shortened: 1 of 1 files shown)


justvanrossum:
This looks good to me. Some tests would be nice.

The fact that VariationModel can work with unnormalized values (extrapolating or not) is tangential I think.

anthrotype:
> Some tests would be nice.

+1, even if just a couple of doctests

> how it handles defaults.

default is always assumed to be at 0 for VariationModel, whether or not min/max are -1/+1 or beyond that

simoncozens:
> default is always assumed to be at 0 for VariationModel, whether or not min/max are -1/+1 or beyond that

Ah, OK, so this is indeed tangential, and will be needed for the glyphsLib thing.

I'll work on tests.

anthrotype:
behdad's instancer-L4 PR contain exactly the same change as yours.
https://github.com/fonttools/fonttools/pull/2728/commits/ccbbd9ba119d8ba9992267cf3459217719e7251e

i'm going to merge this and add a couple tests myself, thanks

PR #2849, opened 2022-10-13
Fix ZeroDivisionErrors with normalizeLocation(extrapolate=True)

anthrotype:
https://github.com/fonttools/fonttools/pull/2847 is incomplete, because it fails with ZeroDivisionError for things like:

```
normalizeLocation({"wght": 100}, {"wght": (400, 400, 700)}, extrapolate=True)
normalizeLocation({"wght": 1000}, {"wght": (400, 700, 700)}, extrapolate=True)
```

tests fails at the moment, fix in follow-up commit.

### DIFF
--- a/Lib/fontTools/varLib/models.py
+++ b/Lib/fontTools/varLib/models.py
@@ -61,13 +61,16 @@ def normalizeValue(v, triple, extrapolate=False):
         )
     if not extrapolate:
         v = max(min(v, upper), lower)
-    if v == default:
-        v = 0.0
-    elif v < default:
-        v = (v - default) / (default - lower)
-    else:
-        v = (v - default) / (upper - default)
-    return v
+
+    if v == default or lower == upper:
+        return 0.0
+
+    if (v < default and lower != default) or (v > default and upper == default):
+        return (v - default) / (default - lower)
+    elif (v > default and upper != default) or (v < default and lower == default):
+        return (v - default) / (upper - default)
+
+    raise AssertionError("that's not possible")
 
 
 def normalizeLocation(location, axes, extrapolate=False):
--- a/Lib/fontTools/varLib/models.py
+++ b/Lib/fontTools/varLib/models.py
@@ -67,10 +67,11 @@ def normalizeValue(v, triple, extrapolate=False):
 
     if (v < default and lower != default) or (v > default and upper == default):
         return (v - default) / (default - lower)
-    elif (v > default and upper != default) or (v < default and lower == default):
+    else:
+        assert (v > default and upper != default) or (
+            v < default and lower == default
+        ), f"Ooops... v={v}, triple=({lower}, {default}, {upper})"
         return (v - default) / (upper - default)
-
-    raise AssertionError("that's not possible")
 
 
 def normalizeLocation(location, axes, extrapolate=False):
--- a/Tests/varLib/models_test.py
+++ b/Tests/varLib/models_test.py
@@ -31,19 +31,104 @@ def test_normalizeLocation():
     assert normalizeLocation({"wght": 1001}, axes) == {"wght": 0.0}
 
 
+@pytest.mark.parametrize(
+    "axes, location, expected",
+    [
+        # lower != default != upper
+        ({"wght": (100, 400, 900)}, {"wght": 1000}, {"wght": 1.2}),
+        ({"wght": (100, 400, 900)}, {"wght": 900}, {"wght": 1.0}),
+        ({"wght": (100, 400, 900)}, {"wght": 650}, {"wght": 0.5}),
+        ({"wght": (100, 400, 900)}, {"wght": 400}, {"wght": 0.0}),
+        ({"wght": (100, 400, 900)}, {"wght": 250}, {"wght": -0.5}),
+        ({"wght": (100, 400, 900)}, {"wght": 100}, {"wght": -1.0}),
+        ({"wght": (100, 400, 900)}, {"wght": 25}, {"wght": -1.25}),
+        # lower == default != upper
+        (
+            {"wght": (400, 400, 900), "wdth": (100, 100, 150)},
+            {"wght": 1000, "wdth": 200},
+            {"wght": 1.2, "wdth": 2.0},
+        ),
+        (
+            {"wght": (400, 400, 900), "wdth": (100, 100, 150)},
+            {"wght": 25, "wdth": 25},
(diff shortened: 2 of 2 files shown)


Tests/varLib/models_test.py
```diff
@@ -31,19 +31,104 @@ def test_normalizeLocation():
...
     assert supportScalar({"wght": 0.2}, {}) == 1.0
     assert supportScalar({"wght": 0.2}, {"wght": (0, 2, 3)}) == 0.1
     assert supportScalar({"wght": 2.5}, {"wght": (0, 2, 4)}) == 0.75
     assert supportScalar({"wght": 3}, {"wght": (0, 2, 2)}) == 0.0
-    assert supportScalar({"wght": 3}, {"wght": (0, 2, 2)}, extrapolate=True, axisRanges={"wght": (0, 2)}) == 1.5
-    assert supportScalar({"wght": -1}, {"wght": (0, 2, 2)}, extrapolate=True, axisRanges={"wght": (0, 2)}) == -0.5
-    assert supportScalar({"wght": 3}, {"wght": (0, 1, 2)}, extrapolate=True, axisRanges={"wght": (0, 2)}) == -1.0
-    assert supportScalar({"wght": -1}, {"wght": (0, 1, 2)}, extrapolate=True, axisRanges={"wght": (0, 2)}) == -1.0
-    assert supportScalar({"wght": 2}, {"wght": (0, 0.75, 1)}, extrapolate=True, axisRanges={"wght": (0, 1)}) == -4.0
+    assert (
```
anthrotype, line 83: ignore changes to `test_supportScalar` below, it's just black reformatting the whole file..

anthrotype:
https://github.com/fonttools/fonttools/actions/runs/3244488440/jobs/5320732749 exposes the ZeroDivisionErrors

behdad:
In such cases we should return the value of the last point.

anthrotype:
> In such cases we should return the value of the last point.

hm, is not that what you mean? https://github.com/fonttools/fonttools/pull/2849/commits/3c935f95ca3d624b5b70d522f664059da6a92bd9

Lib/fontTools/varLib/models.py
```diff
@@ -61,13 +61,16 @@ def normalizeValue(v, triple, extrapolate=False):
...
+
+    if v == default or lower == upper:
+        return 0.0
+
+    if (v < default and lower != default) or (v > default and upper == default):
+        return (v - default) / (default - lower)
+    elif (v > default and upper != default) or (v < default and lower == default):
+        return (v - default) / (upper - default)
+
+    raise AssertionError("that's not possible")
```
behdad, line 73: I'd go with ValueError.

anthrotype: it's deliberately AssertionError because it can't _ever_ be hit if the logic above is correct

anthrotype: basically that line is never executed

Lib/fontTools/varLib/models.py
```diff
@@ -61,13 +61,16 @@ def normalizeValue(v, triple, extrapolate=False):
...
-    elif v < default:
-        v = (v - default) / (default - lower)
-    else:
-        v = (v - default) / (upper - default)
-    return v
+
+    if v == default or lower == upper:
+        return 0.0
+
+    if (v < default and lower != default) or (v > default and upper == default):
```
behdad, line 68: I'm not sure about the logic change. If  `default == upper`, I don't think we should extrapolate using `lower`.

anthrotype: hm.. then I misunderstood the intent of this extrapolate=True.. Isn't this what we want?

```python
In [1]: from fontTools.varLib.models import *

In [2]: normalizeLocation({"wght": 550}, {"wght": (100, 400, 400)}, extrapolate=False)  # clamps to 0.0
Out[2]: {'wght': 0.0}

In [3]: normalizeLocation({"wght": 550}, {"wght": (100, 400, 400)}, extrapolate=True)  # extends beyond
Out[3]: {'wght': 0.5}
```

Lib/fontTools/varLib/models.py
anthrotype: > Then the previous elif can be an else, no?

yes, but I wanted to make the conditions explicit.. I can change to else (and add the conditions in comments)

justvanrossum: Or assert the condition.

anthrotype: @justvanrossum before I make any changes, first we need to agree on the semantics of this code. What do you think of https://github.com/fonttools/fonttools/pull/2849#discussion_r994953694 ?

Lib/fontTools/varLib/models.py
anthrotype: note that normalizeLocation is currently never called with extrapolate=True, the VariationModel already takes locations that are assumed to be normalized (i.e. 0 as default).
Only your PR branch for instancing-L4 calls normalizeValue (which this normalizeLocation calls upon) with extrapolaate=True, in the rebaseTent method.. 

I think @simoncozens needs normalizeLocation(extrapolate=True) for his Smart Components glyphsLib PR, see https://github.com/googlefonts/glyphsLib/pull/822/files#diff-63af9ecacc6b3c6fa71c485ef068d441e09bbb7a7f936291eaa700e18d1611c9R119

justvanrossum: IMO what Cosimo shows is exactly what we want. You normalize using the axis min/default/max, but don't clamp.

anthrotype: so, have we reached a consensus? 
@simoncozens have you tried seeing if using this proposed `normalizeLocation(extrapolate=True)` does what you want in your glyphsLib PR? It's the main reason we are adding this

simoncozens: Yes, it works.

anthrotype: ok thanks. Then I think we can merge this since it's useful. If it's desirable to not do that, but instead raise ValueError as behdad suggested, we could add that as an option later. I personally can't think of a situation where that would apply considering that extrapolate is already opt-in and disabled by default.
By the way, the `instancer-l4` branch tests continue to pass after I locally merge in the current `fix-division-by-zero` PR branch, so the latter changes don't seem to affect the new instancer.solver.

PR #2857, opened 2022-10-19
[glyf] Return component transform as part of control data

behdad:
This makes sure that when we are building variable fonts, we check that all masters have the same component transform in composite glyphs.  We were not checking for this before.


anthrotype:
ideally ufo2ft should decompose these automatically whenever the component transforms don't match across masters.

See https://github.com/googlefonts/ufo2ft/issues/507#issuecomment-1110876311

anthrotype:
but I guess we still want to merge this regardless of what ufo2ft does/will do, so LGTM

PR #2860, opened 2022-10-20
[otlLib|featureVars] sort name records after building STAT, set redundant Counts

anthrotype:
ufo2ft.postProcessor does an unnecessary save/reload of the TTFont it produces in order to normalize the data before comparaing against expected TTX dump, as a few things like redundant Count values aren't built upfront but automatically updated upon compiling the font.

Since in this case it's trivial and cost us nothing to simply set those fields, we just do that and ufo2ft can avoid compiling before asserting XML dump is as expected.


anthrotype:
forgot to link the ufo2ft issue https://github.com/googlefonts/ufo2ft/issues/485

PR #2861, opened 2022-10-21
[instancer-l4] expand singles/tuples to triples upfront and use triples throughout

anthrotype:
Also renamed AxisTent => AxisTriple because I think "tent" is more appropriate to visualize master supports (i.e. a triangle with a peak in the middle), here it's simply the new desired min, default, max values of an axis, aka the axis limits.
I could have named the triple AxisLimit(s) but the latter name was already used to mean a map from axis tags to axis values so I decided to keep the term "AxisLimits" to mean exactly that; I even made a custom mapping that automatically expands singles/tuples so the rest of the code internally can expect triples.
The public facing `instantiateVariableFont` of course continues to accept a hybrid map of axis tags to single values or 2- (and now also 3-) tuples, which are normalized to this AxisLimits.
A lot of words for nothing.

### DIFF
--- a/Lib/fontTools/varLib/instancer/__init__.py
+++ b/Lib/fontTools/varLib/instancer/__init__.py
@@ -51,19 +51,30 @@ Note that, unlike varLib.mutator, when an axis is not mentioned in the input
 location, the varLib.instancer will keep the axis and the corresponding deltas,
 whereas mutator implicitly drops the axis at its default coordinate.
 
-The module currently supports only the first three "levels" of partial instancing,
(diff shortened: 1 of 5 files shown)


anthrotype:
the previous `AxisRange` and `NormalizedAxisRange` namedtuples are deprecated, but kept them as disguised AxisTriples for now, I don't think anybody was using these anyway.

I may add an `__all__` to varLib.instancer to make it clear what's public or not.

behdad:
Thanks for this. Looks great!

Issue #2360, opened 2021-07-10
Improve the performance of SVG table


ACCia:
In decompileentrylist function, can I use struct library instead of sstruct library to parse

`docIndexEntry, data2 = sstruct.unpack2(doc_index_entry_format_0, data2, DocumentIndexEntry())`

This behavior results in slow parsing when the SVG table is retained by the subset

behdad:
If you can speed up the code, please send a Pull Request.

Also, how slows is the code for you? How do you know it's the unpack2 that is slow/ I highly doubt it.

behdad:
> Also, how slows is the code for you? How do you know it's the unpack2 that is slow? Did you profile the code?

I can see how this is slow...

PR #2864, opened 2022-10-26
[subset/SVG] Speed up

behdad:
Subsetting NotoColorEmojiSvg for one glyph.

Before: 280ms
After: 80ms

Fixes https://github.com/fonttools/fonttools/issues/2360

### DIFF
--- a/Lib/fontTools/ttLib/tables/S_V_G_.py
+++ b/Lib/fontTools/ttLib/tables/S_V_G_.py
@@ -70,7 +70,8 @@ class table_S_V_G_(DefaultTable.DefaultTable):
 			data2 = data[pos:]
 			entries = []
 			for i in range(self.numEntries):
-				docIndexEntry, data2 = sstruct.unpack2(doc_index_entry_format_0, data2, DocumentIndexEntry())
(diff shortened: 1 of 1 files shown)


Lib/fontTools/ttLib/tables/S_V_G_.py
```diff
@@ -70,7 +70,9 @@ def decompile(self, data, ttFont):
 			data2 = data[pos:]
 			entries = []
 			for i in range(self.numEntries):
-				docIndexEntry, data2 = sstruct.unpack2(doc_index_entry_format_0, data2, DocumentIndexEntry())
+				# Extract record_data and pass to unpack2, to avoid copying data2 each time.
+				record_data = data2[i * doc_index_entry_format_0Size:(i + 1) * doc_index_entry_format_0Size]
+				docIndexEntry, _ = sstruct.unpack2(doc_index_entry_format_0, record_data, DocumentIndexEntry())
```
anthrotype, line 75: use `sstruct.unpack` here

behdad: Done. Thanks.

anthrotype (approved):

Issue #2140, opened 2021-01-02
[varLib.featureVars] lookups for 'rvrn' should probably be inserted at the beginning


justvanrossum:
It turns out Apple messed up the implementation of `rvrn`: despite the spec saying it should be processed early, it follows the lookup order.

https://twitter.com/justvanrossum/status/1345364798423363584

So fonts built with designspace rules saying `<rules processing="first">` (which use the `rvrn` feature), *behave* like `<rules processing="last">` on macOS (at least on 10.15).

We should probably work around that by inserting the `rvrn` lookups before all other lookups.

anthrotype:
interesting. Could you upload a test font that reproduces the issue?
Inserting the lookup would mess all other lookups referencing other lookups by their index..

justvanrossum:
The necessary remapping if the lookup indices was part of what prompted me to tack them at the and (as well as that I still think that's the only relevant behavior).

Here's a test font with a character 'a', a `smcp` feature, and a variable axis `TEST` that swaps in a swashy alt for both the regular and the smcp glyph. The '`rvrn` version does not work as intended in HB (because rvrn), but *does* work in CoreText:

[FeaVarTestFont.zip](https://github.com/fonttools/fonttools/files/5760771/FeaVarTestFont.zip)

khaledhosny:
Or rvrn could be dropped and ccmp be used instead (not sure if it works in MS Office though, which most likely the sole reason even exists).

moyogo:
The issue would remain with `ccmp`, the feature variation lookups would happen after other `ccmp` lookups when they maybe should happen before.

arrowtype:
I’ve just checked on macOS Big Sur (Version 11.2), and this isn’t fixed there yet.

arrowtype:
To explore this a bit further, I edited a GSUB table ttx to put my rvrn swaps first, then adjusted the indexes and references for following features back to keep things working. 

(Overall, my goal is to make opsz-activated alts, so I’m subbing `R→R.default` at > opsz 24, then have `ss20` swap `R.default→R.grot`. I’m doing something similar but for < opsz 24 for a few other characters such as `I l g`.)

That still works in HarfBuzz, and gets things working in InDesign. Strangely, though, even though a partial attempt at this worked in CoreText earlier (with just the R swap), my more-thorough rearrangement (adjusting two lookups) no longer seems to work. Maybe I’ve messed something up, though.

arrowtype:
[...]

PR #2867, opened 2022-10-28
[varLib/featureVars] Insert rvrn lookups at beginning

behdad:
Fixes https://github.com/fonttools/fonttools/issues/2140
Fixes https://github.com/googlefonts/fontmake/issues/950

### DIFF
--- a/Lib/fontTools/varLib/featureVars.py
+++ b/Lib/fontTools/varLib/featureVars.py
@@ -7,6 +7,7 @@ from fontTools.misc.dictTools import hashdict
 from fontTools.misc.intTools import popCount
 from fontTools.ttLib import newTable
 from fontTools.ttLib.tables import otTables as ot
+from fontTools.ttLib.ttVisitor import TTVisitor
 from fontTools.otlLib.builder import buildLookup, buildSingleSubstSubtable
 from collections import OrderedDict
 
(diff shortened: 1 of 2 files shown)


behdad:
The part for renumbering of (Chain)Context lookups is untested.

Lib/fontTools/varLib/featureVars.py
```diff
@@ -414,19 +415,42 @@ def makeSubstitutionsHashable(conditionalSubstitutions):
         condSubst.append((conditionSet, substitutions))
     return condSubst, sorted(allSubstitutions)
 
+class ShifterVisitor(TTVisitor):
```
anthrotype, line 418: Ha, nice use of a visitor!

Lib/fontTools/varLib/featureVars.py
```diff
@@ -414,19 +415,42 @@ def makeSubstitutionsHashable(conditionalSubstitutions):
...
+    def __init__(self, shift):
+        self.shift = shift
+
+@ShifterVisitor.register_attrs(
+    (
+        (ot.SubstLookupRecord, ("LookupListIndex")),  # GSUB
+        (ot.PosLookupRecord, ("LookupListIndex")),  # GPOS
```
anthrotype, line 425: why not add `FeatureRecord` as well here? it has the same "LookupListIndex" attribute

behdad: `FeatureRecord.LookupListIndex` is a list of numbers, not just a number.

behdad: I mean, sure, I can add another block. Let me do that.

behdad:
PTAL!

anthrotype (approved):
LGTM!

Issue #1362, opened 2018-11-06
io.UnsupportedOperation: seek


weiweihuanghuang:
```
  File "~/Library/Application Support/Glyphs/Scripts/fontTools/ttLib/ttFont.py", line 127, in __init__
    file.seek(0)
io.UnsupportedOperation: seek
```

Getting this error from https://github.com/googlefonts/gf-glyphs-scripts/issues/38

anthrotype:
@behdad Was that seek(0) introduced when we added support for ttCollection maybe?
I was under the impression that TTFont supported unseekable streams (e.g. stdin or zipfile), as that was the reason we read all incoming data into a BytesIO when lazy=False.

PR #2869, opened 2022-10-28
[ttLib] Add main to ttFont and ttCollection

behdad:
Just opens font files from commandline.

### DIFF
--- a/Lib/fontTools/ttLib/__init__.py
+++ b/Lib/fontTools/ttLib/__init__.py
@@ -19,21 +19,68 @@ from fontTools.ttLib.ttFont import *
 from fontTools.ttLib.ttCollection import TTCollection
 
 
-
 def main(args=None):
-	"""Open font with TTFont() or TTCollection()"""
+	"""Open font with TTFont() or TTCollection()
+
+	  ./fonttools ttLib [-oFILE] [-yNUMBER] files...
+
+	If multiple files are given on the command-line,
+	they are each opened (as a font or collection),
+	and added to the font list.
+
+	If -o (output-file) argument is given, the font
+	list is then saved to the output file, either as
+	a single font, if there is only one font, or as
+	a collection otherwise.
+
+	If -y (font-number) argument is given, only the
+	specified font from collections is opened.
+
+	The above allow extracting a single font from a
+	collection, or combining multiple fonts into a
+	collection.
+
+	If --lazy or --no-lazy are give, those are passed
+	to the TTFont() or TTCollection() constructors.
+	"""
 	from fontTools import configLogger
 
 	if args is None:
 		args = sys.argv[1:]
 
-	for f in args:
-		try:
-			font = TTFont(f)
-		except TTLibFileIsCollectionError:
-			font = TTCollection(f)
+	import getopt
+	options, files = getopt.getopt(args, "o:y:",
+		["lazy", "no-lazy"])
 
-		#font.save(outfile)
+	fontNumber = -1
+	outFile = None
+	lazy = None
+	for option, value in options:
+		if option == "-o":
+			outFile = value
+		elif option == "-y":
+			fontNumber = int(value)
+		elif option == "--lazy":
+			lazy = True
+		elif option == "--no-lazy":
+			lazy = False
+
+	fonts = []
+	for f in files:
+		try:
+			font = TTFont(f, fontNumber=fontNumber, lazy=lazy)
+			fonts.append(font)
+		except TTLibFileIsCollectionError:
+			collection = TTCollection(f, lazy=lazy)
+			fonts.extend(collection.fonts)
+
+		if outFile is not None:
+			if len(fonts) == 1:
+				fonts[0].save(outFile)
+			else:
+				collection = TTCollection()
+				collection.fonts = fonts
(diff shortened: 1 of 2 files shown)


behdad:
I can add the docstrings and help stuff to show up in "./fonttools --help" as well. Let me do that.

justvanrossum:
Why is this this useful and why should these modules become subpackages?

behdad:
> Why is this this useful

Helps debugging errors in opening fonts. A niche for sure but I find myself firing up Python console to just TTFont() files often. I'm happy to leave them unlisted.

> and why should these modules become subpackages?

If I add `__name__ == "__main__"` in `ttFont.py` it hits include-order issue because ttFont is first included by ttLib; I'll get a warning from Python.

behdad:
(I was debugging https://github.com/fonttools/fonttools/issues/1362 when I had this idea.)

behdad:
Another usecase I was dealing with today was to measure memory consumption of opening fonts.

The modules can also take `--[no-]lazy`, `--save`, etc. as well.

anthrotype:
hm I'm not convinced this should be added, these CLI entry points don't do anything useful for the end-user, and a developer can always write a three line script to test stuff

khaledhosny:
Opening and saving a font with FontTools is a good optimization trick, currently one has to do it the slow way using ttx twice or write a script. I think it is worthwhile to have these CLI entries.

anthrotype:
well.. if at least two of you find these useful, I'm OK to add them! Make it optionally save and then it becomes more useful like Khaled suggested

justvanrossum:
`ttLib` itself doesn't have a `__main__`, perhaps these ideas could fit there? I would be a less intrusive change.

behdad:
> `ttLib` itself doesn't have a `__main__`, perhaps these ideas could fit there? It would be a less intrusive change.

Currently it is non-trivial to distinguish TTC from non-TTC without peeking at the font bytes... I can make ttLib main just do TTFont for now...

behdad:
> Currently it is non-trivial to distinguish TTC from non-TTC without peeking at the font bytes... I can make ttLib main just do TTFont for now...

Or have a `--collection`

behdad:
Or I can return a specific TTLibError subclass and catch it. Let me do that.

behdad:
Okay, simplified it.

Will look into save, --font-number, etc. It can be used to extract a font from a collection, and possibly to combine fonts into a collection as well.

behdad:
PTAL! This is ready except for docstring.

behdad:
I wrote a rudimentary docstring, but `--help` is not implemented. I suggest merging for now?

Lib/fontTools/ttLib/__init__.py
```diff
@@ -15,3 +17,61 @@ def debugmsg(msg):
...
 from fontTools.ttLib.ttFont import *
 from fontTools.ttLib.ttCollection import TTCollection
+
+
+def main(args=None):
+	"""Open font with TTFont() or TTCollection()
+
+    ./fonttools ttLib [-oFILE] [-yNUMBER] files...
+
+	If multiple files are give on the command-line,
```
anthrotype, line 27:
```suggestion
	If multiple files are given on the command-line,
```

Lib/fontTools/ttLib/__init__.py
```diff
@@ -15,3 +17,70 @@ def debugmsg(msg):
 
 from fontTools.ttLib.ttFont import *
 from fontTools.ttLib.ttCollection import TTCollection
+
+
+def main(args=None):
+	"""Open font with TTFont() or TTCollection()
+
+    ./fonttools ttLib [-oFILE] [-yNUMBER] files...
```
anthrotype, line 25: the indentation seems off

Lib/fontTools/ttLib/__init__.py
```diff
@@ -15,3 +17,70 @@ def debugmsg(msg):
...
+    If --lazy or --no-lazy are give, those are passed
+    to the TTFont() or TTCollection() constructors.
+	"""
+	from fontTools import configLogger
+
+	if args is None:
+		args = sys.argv[1:]
+
+	import getopt
+	options, files = getopt.getopt(args, "o:y:",
```
anthrotype, line 52: i'd have expected to use argparse instead of getopt, but it's ok

justvanrossum: argparse is so much better. The Python docs for getopt start with this highlighted box:

> Note The [getopt](https://docs.python.org/3/library/getopt.html#module-getopt) module is a parser for command line options whose API is designed to be familiar to users of the C getopt() function. Users who are unfamiliar with the C getopt() function or who would like to write less code and get better help and error messages should consider using the [argparse](https://docs.python.org/3/library/argparse.html#module-argparse) module instead.

I think fonttools should use argparse for all new code.

justvanrossum:
> Currently it is non-trivial to distinguish TTC from non-TTC without peeking at the font bytes

Looking at the file name extension is better than nothing. I've only ever seen `.ttc` and `.otc`.

behdad:
Ported to argparse. PTAL!

anthrotype (approved):
LGTM

PR #2870, opened 2022-10-29
CI: Test on Python 3.11

madig:


madig:
I don't understand why pip tries to compile lxml instead of grabbing the wheel.

Issue #2163, opened 2021-01-30
[typing] Ok to add typing_extensions as dep for older Pythons?


madig:
I'm looking at adding typing to some parts of the code base (currently: pens) to make them nicer to work with in IDEs and mypy.

Example: looking at `AbstractPointPen.addPoint`, the `segmentType` can be one of a few strings or `None`. Python's `typing.Literal` covers that, but was added only in Python 3.8. For older Pythons, there's a backport package at https://pypi.org/project/typing-extensions/ that contains that type.

Another example is `Protocol`, which can be used to declare that a class implements some interface, but without necessarily deriving from some common base class. I suppose pens fit that description, but I'm not sure if there is some `isinstance` shenanigans going on somewhere outside fontTools?

Long story short, is it ok to add `typing-extensions` as a conditional dependency for Python < 3.8?

justvanrossum:
Shouldn't `AbstractPen` and `AbstractPointPen` become proper "abc"s first?

1. I'm personally not very keen on adding typing everywhere: it's a huge amount of additional noise for a remarkably small benefit. There's a lot of overengineering here.

2. To add another dependency (even if only for < 3.9) to get the ultimate precision for a call signature in a single method seems overkill.

madig:
I can do the ABC stuff, yeah!

1. Typing is indeed some work, but I found the niceness adding up. A library with typing is nicer to use than one without in my experience.
2. Oh well :o

madig:
https://www.youtube.com/watch?v=Lj_9TyT3V98

Here someone from Facebook talks about gradually adding types to production code.

behdad:
> Long story short, is it ok to add `typing-extensions` as a conditional dependency for Python < 3.8?

Or just requirement to 3.8? What is the timeline for that?

anthrotype:
> What is the timeline for that?

3.7 is EOL in end of June 2023, but I've already seen a few projects dropping support for it.
https://endoflife.date/python

behdad:
Close?

madig:
Yes.

PR #2879, opened 2022-11-03
Drop Python3.7, require 3.8+ and run tests on 3.11 too

### LINKED ISSUE #2878: drop support for 3.7?
Python 3.7 is EOL in about 7 months in June 2023 (see https://endoflife.date/python)

Other popular projects like numpy or scipy are already requiring Python 3.8 in their latest versions.

I think we can do the same without waiting for June 2023. Any objections?

anthrotype:
Fixes #2878 and #2163


anthrotype:
We currently test the minimum and maximum required pythons, previously it was 3.7 (on linux only) and 3.10 (on all of linux, mac and windows).
Initially, in this PR, I tried to bump those to 3.8 (linux) and 3.11 (linux, mac, win), but encountered issues because lxml (again!) doesn't provide wheels for 3.11 for all three. So I reverted to testing 3.10 on linux, mac and windows for the main test suite. 
I still left one CI build that's using Python 3.11, that's the one that tests building the cython modules, and only runs on linux and runs test that don't require lxml so it worked in that case.

So this is ready to be merged, provided everyone is happy about dropping support for Python 3.7.
I will interpret silence as consent if nobody raises concerns within a couple days time.

madig:
I'm wondering if we want to start using pip-tools for requirements management? The text files seem handwritten to me currently, and some deps are pinned in tox instead. We could instead pip-compile several requirements files ready to be used as is in testing? Then there's one way to do dep testing.

anthrotype:
sounds good, wanna send a PR?

anthrotype:
I just reverted this PR with https://github.com/fonttools/fonttools/commit/f21e6768a7e970e3ec0f3c4b818f47f8dfd1c92b. We shall re-revert once a new RoboFont is released, which I hear will use (at least) Python 3.9. I can't speak of Robofont schedule, but I would like to be able to drop Python 3.7 support by the end of this year.

Issue #2737, opened 2022-08-11
[instancer] _instantiateFeatureVariations() limitation when a range rule becomes applicable


behdad:
With current logic, it's possible that, as a result of pinning one axis, a rule becomes default-applicable, but we do not apply it by default. That is wrong.

Let me take an example from the test suite's `InstantiateFeatureVariationsTest::test_partial_instance`, second test.  Imagine the font has these rules:
```
        font = makeFeatureVarsFont( 
            [ 
                ([{"wght": (0.20886, 1.0)}], {"uni0024": "uni0024.nostroke"}), 
                ([{"cntr": (0.75, 1.0)}], {"uni0041": "uni0061"}), 
                ( 
                    [{"wght": (-1.0, -0.45654), "cntr": (0, 0.25)}], 
                    {"uni0061": "uni0041"}, 
                ), 
            ] 
        ) 
```

Now if we pin `{"wght": -1.0}`, currently the test suite says no `appliedSub`, and the following `expectedRecords`:
```
                [ 
                    ({"cntr": (0, 0.25)}, {"uni0061": "uni0041"}), 
                    ({"cntr": (0.75, 1.0)}, {"uni0041": "uni0061"}), 
                ], 
```

However, this is wrong. Because now the rule `({"cntr": (0, 0.25)}, {"uni0061": "uni0041"})` is default-applicable because it starts at 0, so we should also copy its substitution into `appliedSub`.

But if we do that, the `appliedSub` will be applied to the entire space. So we then need to *disable* that at the end with a catch-all rule. That is:

```
                [ 
                    ({"cntr": (0, 0.25)}, {"uni0061": "uni0041"}), 
                    ({"cntr": (0.75, 1.0)}, {"uni0041": "uni0061"}), 
                    ({}, {}),
                ], 
```

Fixing this is hard but doable. I can do it but I prefer to do after merging the L4 branch.

anthrotype:
I'm not sure I understand what you mean by "default-applicable because it starts at 0"..
We don't want to apply the substitution to the entire space, but only apply it when "cntr" is within 0 and 0.25.
It's not clear to me why we would need a catch-all rule at the end. E.g. if one selects an instance at cntr=0.5 (in the partial VF with only cntr, without wght), I would expect that since none of the ranges match, no substitution happens.

behdad:
> I'm not sure I understand what you mean by "default-applicable because it starts at 0"..

Because the rule includes 0, it should apply by the default instance.

[...]

PR #2880, opened 2022-11-03
Instancer feature vars

behdad:
Fix featureVars instantiation.

Fixes https://github.com/fonttools/fonttools/issues/2737

> most implementations do NOT process variation-specific tables at the default-instance (0) location. As such, if there's a condition that applies at 0, it won't be processed. So its substitutions must be promoted to be the default lookup. But since we don't want the default lookup to apply to the entire designspace, we need to undo its effect by adding a catch-all rule and no-substitutions.


Lib/fontTools/varLib/instancer/featureVars.py
```diff
@@ -0,0 +1,178 @@
+
+from fontTools.ttLib.tables import otTables as ot
+from fontTools.varLib.models import normalizeValue
+from copy import deepcopy
+import logging
+
+
+log = logging.getLogger("fontTools.varLib.instancer.featureVars")
```
anthrotype, line 8: we probably don't need this level of granularity and it's fine to use a logger named "fontTools.varLib.instancer" like the parent module

Tests/varLib/instancer/instancer_test.py
```diff
@@ -1975,7 +2000,7 @@ def test_negative_var(self, var, axisTag, newMin, expected):
 def test_limitFeatureVariationConditionRange(oldRange, newLimit, expected):
     condition = featureVars.buildConditionTable(0, *oldRange)
 
-    result = instancer._limitFeatureVariationConditionRange(
+    result = instancer_featureVars._limitFeatureVariationConditionRange(
```
anthrotype, line 2003: why not just call `instancer.featureVars._limitFeatureVariationConditionRange`? why use the alias `instancer_featureVars` with the underscore

anthrotype (approved):
LGTM, thanks!

PR #2883, opened 2022-11-09
Update references of branch name from master to main

pnemade:
This PR will update the existing references of branch name master to branch name main.

### DIFF
--- a/Tests/ttLib/data/TestTTF-Regular.ttx
+++ b/Tests/ttLib/data/TestTTF-Regular.ttx
@@ -468,7 +468,7 @@
       https://github.com/fonttools/fonttools
     </namerecord>
     <namerecord nameID="14" platformID="1" platEncID="0" langID="0x0" unicode="True">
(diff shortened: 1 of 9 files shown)


Tests/ttLib/data/TestTTF-Regular.ttx
```diff
@@ -468,7 +468,7 @@
       https://github.com/fonttools/fonttools
     </namerecord>
     <namerecord nameID="14" platformID="1" platEncID="0" langID="0x0" unicode="True">
-      https://github.com/fonttools/fonttools/blob/master/LICENSE
+      https://github.com/fonttools/fonttools/blob/main/LICENSE
```
anthrotype, line 471: hm that's weird.. changing "master" -> "main" in this name string makes the `Tests/ttLib/woff2_test.py::WOFF2WriterTTFTest::test_hmtx_trasform` fail.. I'll take a look

pnemade:
Thank you for looking into this PR

PR #2884, opened 2022-11-09
[woff2] Support encoding/decoding OVERLAP_SIMPLE glyf flags

### LINKED ISSUE #2576: [woff2] Add support for overlapSimpleBitmap
The [WOFF2 spec](https://www.w3.org/TR/WOFF2/#glyf_table_format) was updated to support retaining the OVERLAP_SIMPLE glyf flags, and the official encoder/decoder has recently been updated to support that: https://github.com/google/woff2/pull/149

Our own woff2 encoder/decoder also need to be updated.

anthrotype:
Fixes #2576

This updates our `fontTools.ttLib.woff2` encoder/decoder to support retaining the `OVERLAP_SIMPLE` glyf flag, following the updated [WOFF 2.0 specification](https://www.w3.org/TR/WOFF2/#glyf_table_format) and official google/woff2 implementation.

### DIFF
--- a/Lib/fontTools/ttLib/woff2.py
+++ b/Lib/fontTools/ttLib/woff2.py
@@ -523,7 +523,8 @@ woff2TransformedTableTags = ('glyf', 'loca')
 
 woff2GlyfTableFormat = """
 		> # big endian
-		version:                  L  # = 0x00000000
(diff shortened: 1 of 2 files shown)


m4rc1e (approved):
LGTM

Issue #2885, opened 2022-11-09
[varLib Instancer] Adjust usWeightClass and names


guidoferreyra:
When instantiating a variable font and the default axes values has has been moved, should the instancer adjust usWeightClass and nameID 1 and 6?

anthrotype:
I think we already do that, have you used --update-name-table option?

guidoferreyra:
You are right! I neglected about `updateFontNames` 🤦 

So only usWeightClass doesn’t get up updated

anthrotype:
> So only usWeightClass doesn’t get up updated

that should also be supported, see

https://github.com/fonttools/fonttools/blob/12bf60aee01e65a1fef2d74814abb7259cfc974d/Lib/fontTools/varLib/instancer/__init__.py#L1170

guidoferreyra:
So I will double check my code because it didn’t worked for me. Thanks!

anthrotype:
how are you calling the instancer? Are you creating a static instance (with all axes fully instantiated) or a new VF with axis defaults changed?

anthrotype:
ah I think I see the problem now.. We only use `axisLimits.pinnedLocation()` when setting default OS/2 weight/width (so we only update those if wght/wdth are pinned), whereas we should also consider the modified axis defaults when the axes are retained. I'll fix that, thanks!

PR #2886, opened 2022-11-09
[instancer] update OS/2 class and post.italicAngle when default moved (L4)

anthrotype:
Previously we were only updating these when pinning wght/wdth/slnt axes. Now we do whenever any of these axes defaults are changed, whether or not the axes are pinned or kept

Fixes https://github.com/fonttools/fonttools/issues/2885


guidoferreyra:
Thank you! I will be testing this on my project!

Issue #1737, opened 2019-10-01
unicodedata.ot_tag_to_script: returns None for "math"


madig:
The function `ufo2ft.featureWriters.ast.getScriptLanguageSystems` determines the OT tag for scripts, but `unicodedata.ot_tag_to_script` does not seem to have a script name for `math`. 

The font https://github.com/madig/DejaVu-Font-Sandbox/tree/master/sources/cubic/DejaVuSans.ufo happens to have some GSUB features dealing with math glyphs, with this declaration at the top: `languagesystem math dflt;` (see https://github.com/madig/DejaVu-Font-Sandbox/blob/master/sources/cubic/DejaVuSans.ufo/features.fea).

According to https://en.wikipedia.org/wiki/Script_(Unicode)#List_of_scripts_in_Unicode, `ot_tag_to_script` should return ISO 15924's `Zmth`, even though this isn't an official script in Unicode? Or should it return something else? Or should ufo2ft do something different?

wollmers:
It should use the script tags of the OpenType Specification wich include `math`. https://docs.microsoft.com/en-us/typography/opentype/spec/scripttags

behdad:
`Zmth` is correct.

wollmers:
@behdad 
> `Zmth` is correct.

"Correct" regarding to which reference? In this case we have 3 conflicting standards, e.g. for 

```
$ uni print U+1D400
     cpoint  dec    utf-8       html       name
'𝐀'  U+1D400 119808 f0 9d 90 80 &#x1d400;  MATHEMATICAL BOLD CAPITAL A (Uppercase_Letter)
````

- OpenType: `math`
- [ISO 15924](https://www.unicode.org/iso15924/codechanges.html): `Zmth`
- Unicode: `Zyyy` Code for undetermined script = Common

A usual solution would be to define both `math` (as alias, maybe with with a deprecation warning) and `Zmth`.

`Zyyy` makes no sense in font context.

IMHO OpenType should review their specification and deprecate `math` in favour of `Zmth`.

behdad:
> > `Zmth` is correct.
> 
> "Correct" regarding to which reference?

I mean, the return value of `unicodedata.ot_tag_to_script` for `math` should be `Zmth`. That's what HarfBuzz uses.

behdad:
@madig does this work for you:
```diff
diff --git a/Lib/fontTools/unicodedata/OTTags.py b/Lib/fontTools/unicodedata/OTTags.py
[...]

PR #2889, opened 2022-11-11
[unicodedata] map Zmth<->math in ot_tag_{to,from}_script

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/1737

### DIFF
--- a/Lib/fontTools/unicodedata/OTTags.py
+++ b/Lib/fontTools/unicodedata/OTTags.py
@@ -22,9 +22,14 @@ SCRIPT_EXCEPTIONS = {
     "Yiii": "yi  ",
     "Nkoo": "nko ",
     "Vaii": "vai ",
+    "Zmth": "math",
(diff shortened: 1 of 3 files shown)


behdad (approved):

madig:
Also needs an entry in `unicodedata.Scripts.NAMES` or `unicodedata.script_horizontal_direction("Zmth")` will raise. Or maybe it needs special treatment somewhere?

anthrotype:
> Also needs an entry in unicodedata.Scripts.NAMES or unicodedata.script_horizontal_direction("Zmth") will raise.

you know you can call `script_horizontal_direction` with a default parameter if you don't want it to raise an exception.
What's the script horizontal direction for Zmth? I suppose LTR?

madig:
Hm. K! The callsite in question is ufo2ft's KernFeatureWriter :)

PR #2891, opened 2022-11-14
[designspaceLib] Add DS.getAxisByTag and refactor getAxis

madig:
I find myself doing this often enough that I decided to upstream it :)


anthrotype (approved):
bonus point for a little test

PR #2892, opened 2022-11-16
Clarify user expectations for getStatNames

### LINKED ISSUE #2890: [Designspace v5] getStatNames should probably consider instance attributes at the given location for computing names
I think currently, only labels are considered. I.e. if I get the STAT names for a location with an instance, the PostScript name should probably reflect the instance's family and style name. Or maybe not?

madig:
Closes #2890


anthrotype (approved):
SGTM, though maybe we could also mention something along the same lines in the designspace xml docs, in addition to these inlined docstrings?

belluzj:
In the XML docs we should mention the name attributes on instances are no longer mandatory if you have STAT info in the file

Issue #2877, opened 2022-11-02
Empty name entry generated if elidable="true" in Designspace v5 `label` element


arrowtype:
I originally posted this issue to FontMake, but I learned over there that varLib is what interprets the designspace into the STAT table (looks like it happens in [varLib/stat.py](https://github.com/fonttools/fonttools/blob/afa0998be236aef3ebfa3445d5441195da2247de/Lib/fontTools/varLib/stat.py)?

In summary, if I add ` elidable="true"` to the "Regular" wght instance `<label>` element, but I don’t add `elidedfallbackname="Regular"` to the `<axis>` element, I get an empty name ID in the resulting variable font’s `name` table.

More details at https://github.com/googlefonts/fontmake/issues/952

This isn’t urgent for me because I have ended up _not_ eliding "Regular" in the case of the current font project. And, I now know to include both eliding tags when I do wish to elide the style name, in other projects. But, this seems like an unexpected and probably unintended result.

Please let me know if any more details would be helpful. Thank you!

justvanrossum:
Can confirm — I just ran into this yesterday. I also ended up not eliding Regular.

anthrotype:
@justvanrossum how do you suggest we fix? Assume "Regular" when all particles get elided or simply raise an error and let the user set the fallback explicitly?

justvanrossum:
Not sure. Is it possible (and desirable) to end up with _no_ style name? So the full name of the regular instance would simply be "FamilyName". 

If so, then the empty name table entry may be correct (and fontbakery should not complain about it). 

If not, then an error sounds like the best option here.

anthrotype:
empty style name sounds like a font bug to me, that's why they defined the elided fallback name, to prevent that from happening...

justvanrossum:
This kind of configuration used to be quite common:
- `Times`
- `Times-Bold`
- `Times-Italic`
- `Times-BoldItalic`

The elided fallback name was an afterthought (STAT 1.1) so that you *can* provide a fallback if all particles get elided, not that you *must*. (Or at least, that's how I'm reading it.) The [spec](https://learn.microsoft.com/en-us/typography/opentype/spec/stat#style-attributes-header) says:

> The elidedFallbackNameId field provides a name that _**can**_ be used when composing a name if all of the axis-value names are elidable. 

(Emphasis mine)

The only way I can imagine how the above configuration can be done with `fvar` named instance subfamilyNameIDs is to allow an empty string.

[...]

PR #2903, opened 2022-11-22
[designspaceLib] Provide fallback for `elidedFallbackName`

madig:
Provide a fallback for the DS-wide `elidedFallbackName`.

This is a bit unfortunate, because the `DesignSpaceDocument.elidedFallbackName` is a string, not a localization dict. However, if it is not set by the user, we just insert a generic English "Regular". Still better than having an empty style name.

This also fixes a bug where a RIBBI style would end up with a style mapping like `stylemapfamilyname="Source Serif 4 Bold Italic" stylemapstylename="regular"`. RIBBI styles should not be part of the style map family name.

Closes https://github.com/fonttools/fonttools/issues/2877.

### DIFF
--- a/Tests/designspaceLib/data/convert5to4_output/SourceSerif4Variable-Italic.designspace
+++ b/Tests/designspaceLib/data/convert5to4_output/SourceSerif4Variable-Italic.designspace
@@ -180,7 +180,7 @@
       <kerning/>
       <info/>
     </instance>
-    <instance name="Source Serif 4 Italic" familyname="Source Serif 4" stylename="Italic" filename="Source Serif 4-Italic.ttf" postscriptfontname="SourceSerif4Italic-Regular" stylemapfamilyname="Source Serif 4 Italic" stylemapstylename="regular">
+    <instance name="Source Serif 4 Italic" familyname="Source Serif 4" stylename="Italic" filename="Source Serif 4-Italic.ttf" postscriptfontname="SourceSerif4Italic-Regular" stylemapfamilyname="Source Serif 4" stylemapstylename="regular">
       <location>
         <dimension name="weight" xvalue="394"/>
         <dimension name="optical" xvalue="20"/>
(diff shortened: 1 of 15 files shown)


madig:
Hm, wait, this still leaves the `fvar` name for the fully elided location empty.

Tests/designspaceLib/data/convert5to4_output/SourceSerif4Variable-Italic.designspace
```diff
@@ -196,7 +196,7 @@
       <kerning/>
       <info/>
     </instance>
-    <instance name="Source Serif 4 Bold Italic" familyname="Source Serif 4" stylename="Bold Italic" filename="Source Serif 4-Bold Italic.ttf" postscriptfontname="SourceSerif4Italic-Bold" stylemapfamilyname="Source Serif 4 Bold Italic" stylemapstylename="regular">
+    <instance name="Source Serif 4 Bold Italic" familyname="Source Serif 4" stylename="Bold Italic" filename="Source Serif 4-Bold Italic.ttf" postscriptfontname="SourceSerif4Italic-Bold" stylemapfamilyname="Source Serif 4" stylemapstylename="regular">
```
belluzj, line 199: This change sounds wrong, either keep the stylemapfamilyname, or change the stylemapstylename to bold italic?

madig: The DS has no style linking though? How is getStatName supposed to do the right thing?

belluzj: No style linking = stylemapfamilyname is set to the familyname + stylename, and stylemapstylename is set to regular

belluzj: Otherwise, applications that don't understand the 16/17 named ids are going to see a bunch of "Source Serif 4" fonts and won't be able to distinguish them, that's why we put the stylename into the stylemapfamilyname.

The other option, if the DS had style linking, would be to set the stylemapstylename correctly, in this case bold italic

Issue #2896, opened 2022-11-18
Can it be possible to replace problematic word `master` from this project code?


pnemade:
When I searched for "master" word, I found many entries that matched `master` and `masters` word/variable references in source code. I also saw same reference in https://github.com/fonttools/fonttools/tree/main/Doc/source/designspaceLib files.

Can upstream think if it can be replaced by some other alternative word?

anthrotype:
thanks for the reminder. Can you make a PR?

for branch name, "master" -> "main"

for designspace "master" -> "source"

behdad:
Changing it in varLib would be harder / I don't think possible at this point.

PR #2904, opened 2022-11-23
Replace `master` word in designspace document files

pnemade:
This PR is an attempt to fix some `master` word entries in `fonttools/Doc/source/` directory files.
Please review this PR and do suggest if needed any changes.

This PR is created to fix issue #2896

### DIFF
--- a/Doc/source/designspaceLib/scripting.rst
+++ b/Doc/source/designspaceLib/scripting.rst
@@ -59,7 +59,7 @@ Make a descriptor object and add it to the document.
 -  The ``tag`` attribute is the one of the registered `OpenType
    Variation Axis
    Tags <https://www.microsoft.com/typography/otspec/fvar.htm#VAT>`__
--  The default master is expected at the intersection of all
+-  The default source is expected at the intersection of all
    default values of all axes.
 
 Option: add label names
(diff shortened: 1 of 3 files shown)


Doc/source/designspaceLib/scripting.rst
```diff
@@ -195,16 +195,16 @@ this into something clever.
...
-    # You can specify different masters
+    # You can specify different sources
     # for this specific glyph.
-    # You can also give those masters new
+    # You can also give those sources new
     # locations. It's a miniature designspace.
     # Remember the "name" attribute we assigned to the sources?
-    glyphData['masters'] = [
-        dict(font="master.thin",
+    glyphData['sources'] = [
```
belluzj, line 203: Here you may need to change the matching code that expects this dictionary key

pnemade: are you suggesting to change the occurrences of `glyphData['masters']` with `glyphData['sources']` in `Lib/fontTools/designspaceLib/*` source files?

belluzj: Yes basically, however I'm not familiar with that part of the designspace spec, maybe @LettError and others can advise. I think maybe font editors who rely on the format would need updating as well if the key changes from "masters" to "sources", that is, code outside this repository would need to adapt. But I'm not sure 100%.

LettError:
I don't think there are any tools left that need  these glyph-specific designspaces.

PR #2912, opened 2022-12-07
[cu2quPen] Add Cu2QuMultiPen

behdad:


behdad:
Docs to be added.

anthrotype:
> this changes most pen method signatures to an extent that I don't think it should inherit from AbstractPen anymore

yeah, I agree. There's no point inheriting if changing all methods' signatures anyway.
I wonder if we could instead define a generic filter pen called something like ZipPen or ZippingPen (for lack of better name) that would take a list of other segment pens (in your case Cu2QuPen) and would call each respective pen commands by zipping through the provided arguments.
I see Behdad already has a use case, would be nice to support this in fonttools.

https://github.com/behdad/smarties/blob/93d3f32c83ccb1cedb3ed10a7c5858a0790a8326/smarties.py#L569-L571

behdad:
> General remark: this changes most pen method signatures to an extent that I don't think it should inherit from AbstractPen anymore.

Right. That occurred to me later. Will remove AbstractPen.

behdad:
Added docs.

Lib/fontTools/pens/cu2quPen.py
```diff
@@ -258,3 +258,102 @@ def _drawPoints(self, segments):
...
+
+    This pen does not follow the normal SegmentPen protocol. Instead, its
+    moveTo/lineTo/qCurveTo/curveTo methods take a list of tuples that are
+    arguments that would normally be passed to a SegmentPen, one item for
+    each of the pens in other_pens.
+    """
+
+    def __init__(self, other_pens, max_err, reverse_direction=False):
+        if reverse_direction:
+            other_pens = [ReverseContourPen(pen) for pen in other_pens]
```
anthrotype, line 284: does this need to set (or forward on) the newly added `outputImplicitClosingLine` param from https://github.com/fonttools/fonttools/pull/2922 ?

anthrotype (commented):
generally LGTM but I'd like some tests in https://github.com/fonttools/fonttools/blob/main/Tests/pens/cu2quPen_test.py

Lib/fontTools/pens/cu2quPen.py
behdad: Yes, when that's merged.

behdad: Done.

anthrotype:
I'm going to merge this so it gets included in the big auto-formatting PR #2925 - I filed https://github.com/fonttools/fonttools/issues/2926 to track adding unit tests for the new pen

PR #2918, opened 2022-12-08
[gvar] Don't expand glyph for unused pointCount

behdad:
Keep it in the API though.

The compile code doesn't use the pointCount. This avoids expanding the glyph unnecessarily (and makes my smarties code work as well, since it builds glyphs that cannot be expanded by current fonttools).


Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -632,6 +632,7 @@ def compileSharedTuples(axisTags, variations,
 def compileTupleVariationStore(variations, pointCount,
                                axisTags, sharedTupleIndices,
                                useSharedPoints=True):
+	# pointCount is actually unused. Keeping for API compat.
```
anthrotype, line 635: you can do `del pointCount` to make that explicit

anthrotype (approved):

PR #2919, opened 2022-12-09
Unify pen argument names

jenskutilek:
Some pens used `baseGlyphName` and `transform` as argument names, others used `glyphName` and `transformation`. This PR unifies the argument names and adds some type annotations.

### DIFF
--- a/Lib/fontTools/cu2qu/ufo.py
+++ b/Lib/fontTools/cu2qu/ufo.py
@@ -24,6 +24,7 @@ the resulting splines are interpolation-compatible.
 """
 
 import logging
+from typing import Tuple
 from fontTools.pens.basePen import AbstractPen
 from fontTools.pens.pointPen import PointToSegmentPen
 from fontTools.pens.reverseContourPen import ReverseContourPen
@@ -91,7 +92,11 @@ class GetSegmentsPen(AbstractPen):
     def endPath(self):
         self._add_segment('end')
 
-    def addComponent(self, glyphName, transformation):
+    def addComponent(
+		self,
+		glyphName: str,
+		transformation: Tuple[float, float, float, float, float, float],
+	) -> None:
         pass
 
 
--- a/Lib/fontTools/cu2qu/ufo.py
+++ b/Lib/fontTools/cu2qu/ufo.py
@@ -65,38 +65,38 @@ class GetSegmentsPen(AbstractPen):
     duplicated between segments.
     """
 
-    def __init__(self):
+    def __init__(self) -> None:
         self._last_pt = None
         self.segments = []
 
-    def _add_segment(self, tag, *args):
+    def _add_segment(self, tag, *args) -> None:
         if tag in ['move', 'line', 'qcurve', 'curve']:
             self._last_pt = args[-1]
         self.segments.append((tag, args))
 
-    def moveTo(self, pt):
+    def moveTo(self, pt) -> None:
         self._add_segment('move', pt)
 
-    def lineTo(self, pt):
+    def lineTo(self, pt) -> None:
         self._add_segment('line', pt)
 
-    def qCurveTo(self, *points):
+    def qCurveTo(self, *points) -> None:
         self._add_segment('qcurve', self._last_pt, *points)
 
-    def curveTo(self, *points):
+    def curveTo(self, *points) -> None:
         self._add_segment('curve', self._last_pt, *points)
 
-    def closePath(self):
+    def closePath(self) -> None:
         self._add_segment('close')
 
-    def endPath(self):
+    def endPath(self) -> None:
         self._add_segment('end')
 
     def addComponent(
-		self,
-		glyphName: str,
-		transformation: Tuple[float, float, float, float, float, float],
-	) -> None:
+        self,
+        glyphName: str,
+        transformation: Tuple[float, float, float, float, float, float],
+    ) -> None:
         pass
 
 
--- a/Lib/fontTools/pens/basePen.py
+++ b/Lib/fontTools/pens/basePen.py
@@ -47,6 +47,7 @@ __all__ = [
 ]
 
 
+PenGlyphSet: TypeAlias = Optional[Dict[str, Any]]
 PenPoint: TypeAlias = Tuple[float, float]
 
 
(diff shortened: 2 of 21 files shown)


Lib/fontTools/cu2qu/ufo.py
```diff
@@ -91,7 +92,11 @@ def closePath(self):
     def endPath(self):
         self._add_segment('end')
 
-    def addComponent(self, glyphName, transformation):
+    def addComponent(
+		self,
```
anthrotype, line 96: mixing tabs and spaces?
feels weird we add typing only to this method and not the rest.  Maybe for now just rename the parameters if you don't have time to add type hints to all.

jenskutilek: C&P error ...

jenskutilek:
The tests fail ... didn’t see it locally as tox didn’t run because of lxml installation problems. I’m on it …

Lib/fontTools/cu2qu/ufo.py
jenskutilek: I've started to add typing annotations to the pens. But I had to stop now, because it gets complicated and even requires some code rewriting to pass the type checks. But at least there is more typing than before now :)

madig: Hm, where would it require code rewriting?

benkiel:
Putting in a warning here, lots of things that are not fontTools pull in pens/subclass pens; please be sure that you aren't introducing breaking changes

Lib/fontTools/pens/basePen.py
```diff
@@ -36,34 +36,42 @@
...
-__all__ =  ["AbstractPen", "NullPen", "BasePen", "PenError",
-			"decomposeSuperBezierSegment", "decomposeQuadraticSegment"]
+__all__ = [
+	"AbstractPen", "NullPen", "BasePen", "PenError", "decomposeSuperBezierSegment",
+	"decomposeQuadraticSegment", "PenPoint",
+]
+
+
+PenGlyphSet: TypeAlias = Optional[Dict[str, Any]]
+PenPoint: TypeAlias = Tuple[float, float]
```
justvanrossum, line 51: I don't agree that a "pen point" is a an x, y tuple. It is a sequence with length two, containing two numbers.

justvanrossum (commented):
I don't have time to do a full review, but I notice `typing_extensions` is sneaking in here as a new dependency. I don't have a strong opinion about that right now, but there was some discussion, also in the context of dropping support for Python 3.7.

madig:
> Putting in a warning here, lots of things that are not fontTools pull in pens/subclass pens; please be sure that you aren't introducing breaking changes

Hm, I'm not sure how one can meaningfully "be sure" about it, other than to drop the PR? Maybe a GitHub search can help find some API users.

Lib/fontTools/pens/basePen.py
```diff
@@ -350,20 +374,24 @@ def decomposeSuperBezierSegment(points):
...
+					0.5 * (pt2[1] + temp[1]),
+				)
 				bezierSegments.append((pt1, pt2, pt3))
 				pt1, pt2, pt3 = temp, None, None
 	bezierSegments.append((pt1, points[-2], points[-1]))
 	return bezierSegments
 
 
-def decomposeQuadraticSegment(points):
+def decomposeQuadraticSegment(points: List[Union[PenPoint, None]]) -> List[Tuple[PenPoint, PenPoint]]:
```
madig, line 394: You can `from __future__ import annotations` at the top and then rewrite this as `def decomposeQuadraticSegment(points: 
list[PenPoint | None]) -> list[tuple[PenPoint, PenPoint]]:` and save yourself the `List`, `Union` and `Tuple` imports :)

Lib/fontTools/pens/boundsPen.py
```diff
@@ -27,8 +27,8 @@ def __init__(self, glyphSet, ignoreSinglePoints=False):
 		self.init()
 
 	def init(self):
-	    self.bounds = None
-	    self._start = None
+		self.bounds = None
```
madig, line 30: What happened to the indentation here?

Lib/fontTools/pens/basePen.py
jenskutilek: Thanks! I wasn't sure if Python 3.7 can deal with that notation.

madig: The annotations import was added in 3.7. You still need to import the typings for type aliases though... until we have 3.10 minimum.

jenskutilek:
@madig I agree that it is probably better to tackle this in separate steps. Should I prepare a new branch that only fixes the formatting for all pen code?

Lib/fontTools/cu2qu/ufo.py
jenskutilek: I needed to add some assertions, otherwise the type checker would complain about variables possibly being None, e.g.:

https://github.com/fonttools/fonttools/pull/2919/files#diff-713c63d7dd345e61c8c06754b31811b60345f57b1b2c70fdf7bb8aaf7227c92aR408-R420

jenskutilek:
> > Putting in a warning here, lots of things that are not fontTools pull in pens/subclass pens; please be sure that you aren't introducing breaking changes
> 
> Hm, I'm not sure how one can meaningfully "be sure" about it, other than to drop the PR? Maybe a GitHub search can help find some API users.

Should we add deprecation messages when somebody uses the "old" argument names? The only things that actually changed are `baseGlyphName` → `glyphName` and `transform` → `transformation` in some pens `addComponent` arguments.

anthrotype:
> deprecation messages

anthrotype:
> deprecation messages

a client would get TypeError and it's going to be too late for issuing deprecation warning..

jenskutilek:
I've made a formatting-only PR in #2923.

Lib/fontTools/pens/basePen.py
madig: @anthrotype actually, I'm wondering about the `points` argument, Is `None` really possible? The docstring says the last point must be an on-curve point, implying that it cannot be `None`?

anthrotype: Yes the last point in the list (and only the last) can be None, which indicates a closed TrueType contour containing only off-curve points and no on-curves..

madig: Yeah, but in this instance? It's only called from https://github.com/fonttools/fonttools/blob/16884008daa45a04cfe9a526aff767053e24b09a/Lib/fontTools/pens/basePen.py#L348 where the last point is taken care of?

anthrotype: Good point you are right the caller handles that case already, if this function expects all points to be non-None that we can type it without the Optional, maybe add an assertion too

Issue #2913, opened 2022-12-08
[ttGlyphPen] Add a preserve_geometry option


behdad:
I'm using ttGlyphPen to build variation data; so it's important to preserve geometry. Currently I see the following code that might be an issue, that needs to be disabled:
```
# if first and last point on this path are the same, remove last
```

I'll submit a PR, unless @anthrotype wants to beat me to it.

jenskutilek:
I think `TTGlyphPointPen` does a better job preserving all points.

justvanrossum:
Also tend to use point pens if geometry needs to be preserved.

anthrotype:
I also agree that using the TTGlyphPointPen (or pointPen protocol in general) would be better at preserving geometry, in fact we use that in ufo2ft.
That piece of code looks suspicious though, I don't remember why it was added in the first place.
I think it might have something to do with the implied-lineTo-start-point-in-closed-contours rule in segment pen protocol.

If you really need to use segment-oriented pens you can try use the adapter pens in pointPen.py module.
Which reminds the PointToSegmentPen has a parameter `outputImpliedClosingLine` that forces the closing lineTo command to always be emitted (and [cu2qu itself uses that to make sure it can convert curves compatibly](https://github.com/anthrotype/cu2qu/commit/855166c518a08e26e1f7c17034b0b6ee403bb408)).

https://github.com/fonttools/fonttools/blob/96de954d295854f3294f576f78ab177c8d2b069e/Lib/fontTools/pens/pointPen.py#L213-L229

Maybe we could add a similar logic to TTGlyphPen itself.

behdad:
> Maybe we could add a similar logic to TTGlyphPen itself.

That sounds good to me. Thanks!

PR #2921, opened 2022-12-09
[ttGlyphPen] Add outputImpliedClosingLine argument

behdad:
Fixes https://github.com/fonttools/fonttools/issues/2913

### DIFF
--- a/Lib/fontTools/pens/ttGlyphPen.py
+++ b/Lib/fontTools/pens/ttGlyphPen.py
@@ -164,6 +164,15 @@ class TTGlyphPen(_TTGlyphBasePen, LoggingPen):
     drawMethod = "draw"
     transformPen = TransformPen
 
+    def __init__(
+        self,
+        glyphSet: Optional[Dict[str, Any]] = None,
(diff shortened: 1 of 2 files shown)


Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -164,6 +164,14 @@ class TTGlyphPen(_TTGlyphBasePen, LoggingPen):
     drawMethod = "draw"
     transformPen = TransformPen
 
+    def __init__(self,
+                 glyphSet: Optional[Dict[str, Any]] = None,
+                 handleOverflowingTransforms: bool = True,
+                 outputImpliedClosingLine: bool = False) -> None:
```
justvanrossum, line 170: I think this module is already using black, so perhaps reformat with black.

anthrotype (approved):
LGTM pending formatting changes, and ideally a test in https://github.com/fonttools/fonttools/blob/main/Tests/pens/ttGlyphPen_test.py

behdad:
Added tests.

Issue #2914, opened 2022-12-08
[reverseContourPen] Add a `outputImpliedClosingLine` option


behdad:
I have code for generating variation data, when passing that through reverseContourPen, one of my glyphs suddenly has one fewer point.

anthrotype:
can you provide a test case so we can track this down? 
also have you considered using a pointPen which by definition preserves geometry, instead?
There's a ReverseContourPointPen in fontTools.pens.pointPen
https://github.com/fonttools/fonttools/blob/96de954d295854f3294f576f78ab177c8d2b069e/Lib/fontTools/pens/pointPen.py#L407

i wonder if this should be moved (or aliased) to fontTools.pens.reverseContourPen

behdad:
> can you provide a test case so we can track this down?

I'll try to produce one.

> also have you considered using a pointPen which by definition preserves geometry, instead?

That did not occur to me no. At this point I'm too invested in the SegmentPen to switch :(. Which is to say, my code is complete! :)

behdad:
I debugged it and it's the same issue that I identified in ttGlyphPen:
```
            if firstOnCurve != lastOnCurve: 
                # emit an implied line between the last and first points 
                yield "lineTo", (lastOnCurve,) 
                contour[-1] = (lastType, 
                               tuple(lastPts[:-1]) + (firstOnCurve,)) 
```

If I make it always take that branch, then my code succeeds.

behdad:
Similar to https://github.com/fonttools/fonttools/issues/2913#issuecomment-1342524038, maybe add `outputImpliedClosingLine`.

PR #2922, opened 2022-12-09
[reverseContourPen] Add outputImpliedClosingLine

behdad:
Fixes https://github.com/fonttools/fonttools/issues/2914


behdad:
I was wrong. This doesn't completely fix my issue. Investigating.

behdad:
> I was wrong. This doesn't completely fix my issue. Investigating.

Works now.

anthrotype (approved):

anthrotype (commented):
there's https://github.com/fonttools/fonttools/blob/main/Tests/pens/reverseContourPen_test.py if you'd like to add a couple tests

behdad:
Tests added. Thanks!

Issue #2233, opened 2021-03-22
[Housekeeping] Formatting all code with black


madig:
Hi all,
Cosimo and me would like to format all code with https://github.com/psf/black to unify style and automate the boring tasks of fitting your code into the surrounding style. This would also include a CI lint check to make sure the black style is adhered to for any incoming code.

Two known difficulties:
1. Existing PRs may need to be rebased on top of the newly formatted code.
2. Making `git blame` work as before requires an extra step: https://github.com/psf/black#migrating-your-code-style-without-ruining-git-blame

Thoughts?

justvanrossum:
> Making git blame work as before requires an extra step: https://github.com/psf/black#migrating-your-code-style-without-ruining-git-blame

There I read: "The one caveat is that GitHub and GitLab do not yet support ignoring revisions using their native UI of blame"

Which is literally the only way I ever use blame, so I'm not super excited at this moment.

anthrotype:
even in github blame UI one can always click through past commits when exploring the git blame history, so I don't think this is such a problem. We have a lot of noisy commits where we do "minor" (whatever that means) or apply formatting changes

justvanrossum:
Ok, fair enough. And as a side effect we'll lose all tabs, yes? That'd be worth a party.

alerque:
My philosophy on this is that the sooner code styling gets normalized the sooner you get to the end of the long tail of disruption caused.

1. Existing PR's should be easy to fix one of two ways depending on complexity:
    a) If they can be automatically rebased on master *now*, they can be rebased automatically after the change by running the reformatter on each of the commits in the PR first, then rebasing on the new master.
    b) If the PR cannot currently be rebased anyway but it could be merged (for example `git rerere` artifacts have already been developed for it) then merge the old master pre-rewite into the PR branch to resolve conflicts, then run the formatter in the branch, merge again from the new master, then squash the two merges and format commit to get a clean merge from the new format.

2. Is web UI blame usage frequent enough that the extra click/level of blame history would really be a bother?

behdad:
Sgtm.

khaledhosny:
[...]

PR #2923, opened 2022-12-12
Format all pens with black

jenskutilek:
... and fix flake8 errors


tox.ini
```diff
@@ -41,6 +41,15 @@ commands =
...
     coverage html
 
+[testenv:lint]
+skip_install = true
+deps =
+    black
+    flake8
+commands =
+    black --check --diff Lib/fontTools/pens Lib/fontTools/cu2qu/ufo.py Lib/fontTools/ufoLib/glifLib.py Lib/fontTools/varLib/interpolatable.py
+    flake8 Lib/fontTools/pens Lib/fontTools/cu2qu/ufo.py Lib/fontTools/ufoLib/glifLib.py Lib/fontTools/varLib/interpolatable.py
```
madig, line 51: Maybe these should be put into the config files like with mypy so we can more easily extend them as time goes on.

- https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-format
- https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-filename

madig: Actually, I would pin black and flake8 in dev-requirements.txt, because new versions can and will break CI when you're working on something else.

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[tool.black]
+line-length = 120
+target-version = ["py38"]
```
madig, line 3:
```suggestion
target-version = ["py37"]
```

:)

pyproject.toml
```diff
@@ -0,0 +1,11 @@
+[tool.black]
+line-length = 120
```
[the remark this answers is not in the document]
madig: I think the default of 88 is fine in most circumstances.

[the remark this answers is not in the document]
jenskutilek: Should we use black's default of 88? In other projects I use 79 (flake8 default), but the lines are quite short then.

anthrotype:
I prefer we do one PR (one commit) for running black on the entire repository #2233 (as well as enforcing black --check on CI), without mixing it with other unrelated flake8 stuff, those can be done separately. @madig is working on the black part, so please hold on

PR #2927, opened 2022-12-13
[feaLib] Sort name table entries in builder

### LINKED ISSUE #2909: feaLib does not sort name table
Although the name table is sorted when it is compiled, if it is written out without being compiled (such as to TTX) then the entries will not be in the order expected by the spec.

In practice, I suspect this is only really an issue in the feaLib test suite, where there are [several](https://github.com/fonttools/fonttools/blob/5a0dc4bc8dfaa0c7da146cf902395f748b3cebe5/Tests/feaLib/data/spec8d.ttx#L5) [cases](https://github.com/fonttools/fonttools/blob/main/Tests/feaLib/data/spec8c.ttx#L11) where the expected TTX has name entires in an order that is not valid per the spec.

This has come up for me because I am reusing the (very nice!) feaLib test suite for the development of a new feature compiler, and am running into these differences.

This is not a huge problem for me; I have a mechanism for indicating that certain diffs are 'acceptable', so I can work around this reasonably easily. I do think that it would be nice if the TTX in these cases did reflect the expected order in the compiled table.

cmyr:
This patch ensures that feaLib always produces a name table with entries sorted in the order proscribed by the spec: platform id, encoding id, language id, name id.

This breaks some tests, and so I have manually updated the test data to match the new outputs.

---

I'm not very familiar with this project, so very possible i have overlooked or omitted something, please let me know! There was one question I had, will point it out in the diff.

closes #2909

### DIFF
--- a/Tests/feaLib/data/spec8c.ttx
+++ b/Tests/feaLib/data/spec8c.ttx
@@ -2,17 +2,17 @@
 <ttFont>
 
   <name>
-    <namerecord nameID="256" platformID="3" platEncID="1" langID="0x409">
-      Feature description for MS Platform, script Unicode, language English
-    </namerecord>
-    <namerecord nameID="256" platformID="3" platEncID="1" langID="0x411">
-      Feature description for MS Platform, script Unicode, language Japanese
-    </namerecord>
     <namerecord nameID="256" platformID="1" platEncID="0" langID="0x0" unicode="True">
       Feature description for Apple Platform, script Roman, language unspecified
     </namerecord>
     <namerecord nameID="256" platformID="1" platEncID="1" langID="0xc" unicode="True">
       Feature description for Apple Platform, script Japanese, language Japanese
+    </namerecord>
+    <namerecord nameID="256" platformID="3" platEncID="1" langID="0x409">
+      Feature description for MS Platform, script Unicode, language English
+    </namerecord>
+    <namerecord nameID="256" platformID="3" platEncID="1" langID="0x411">
+      Feature description for MS Platform, script Unicode, language Japanese
     </namerecord>
   </name>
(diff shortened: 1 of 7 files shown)


Tests/feaLib/data/spec8c.ttx
```diff
@@ -39,7 +39,7 @@
         <Feature>
           <FeatureParamsStylisticSet>
             <Version value="0"/>
-            <UINameID value="256"/>  <!-- Feature description for MS Platform, script Unicode, language English -->
+            <UINameID value="256"/>  <!-- Feature description for Apple Platform, script Roman, language unspecified -->
```
cmyr, line 42: so I'm curious about how the comment strings are determined. Is it merely the text of the first entry with this ID?

anthrotype: for the comments we use name table's `getDebugName` method:

https://github.com/fonttools/fonttools/blob/164d6ead033c506c19647778c3e739f029b0ad5f/Lib/fontTools/ttLib/tables/_n_a_m_e.py#L102-L121

anthrotype (approved):

Tests/feaLib/data/spec8c.ttx
anthrotype: so yeah, if Macintosh English (platformID=1 langID=0) names exist that match the requested nameID, we return those since they come before Windows English names (platformID=3 langID=0x409)

khaledhosny: We probably change that code to drop platform and script part if multiple entries of the same NamID exist as it is just misleading. Or may be drop it unconditionally, does not sound like much useful information.

PR #2928, opened 2022-12-13
set codecov to not fail_ci_if_error

anthrotype:
Codecov keeps failing to upload the coverage report seemingly at random, e.g.

https://github.com/fonttools/fonttools/actions/runs/3686952402/jobs/6240048994#step:8:40

I thought I had already fixed it by setting a CODECOV_TOKEN in the repo secrets with #2763 

but it still keeps pestering us so I think the only option left is to not have the PR checks fail with codecov errors.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -63,7 +63,7 @@ jobs:
         file: coverage.xml
         flags: unittests
         name: codecov-umbrella
-        fail_ci_if_error: true
+        fail_ci_if_error: false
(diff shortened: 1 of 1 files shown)


anthrotype:
hm I'm not 100% sure but it might be a github "feature" that the repository's secrets such as CODECOV_TOKEN are not available from pull requests such as one early on (#2927) from @cmyr who was contributing for the first time and is not currently an org member?

I can see that codecov managed to find its token from the environment variables in this PR of mine, unlike with his..

PR #2936, opened 2022-12-22
Add typing to pens

madig:
This an alternative to https://github.com/fonttools/fonttools/pull/2919 where I just apply the typing but keep the argument names the same. They could be changed in a separate PR.


benkiel:
@madig Can you comment on how merging this may impact things that aren't font tools that use the base pens?

madig:
It shouldn't impact any users. I kept the argument names unchanged, but added e.g. the `identifier` kwarg where it was missing; that shouldn't impact anyone either, because all of those instances were followed by `kwargs`  anyway.

knutnergaard:
All pens annotated in #4037

Issue #2949, opened 2023-01-09
[feaLib] ignore statements without marked glyphs


cmyr:
Another instance where it feels like there is a discrepancy between the spec and feaLib's behaviour.

Per the spec ([5.f.ii. Specifying exceptions to the Chain Sub rule](http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#5fii-specifying-exceptions-to-the-chain-sub-rule)):

>The backtrack and lookahead sequences may be omitted, but there must be at least one marked glyph or glyph class.

However the [GSUB_5_formats.fea](https://github.com/fonttools/fonttools/blob/main/Tests/feaLib/data/GSUB_5_formats.fea) test case consists of `ignore sub` statements that do not have any marked glyphs or glyphclasses.

My understanding is that in the absence of any marked glyphs, feaLib chooses to interpret all glyphs/classes as belonging to the input sequence.

So: Is this intentional? If so, is there any particular rationale?

cc @simoncozens, who added this particular test case.

cmyr:
Interesting; you're correct, given the input,

```fea
feature cool {
    ignore sub e x;
} cool;
```

makeotf produces a type-6 lookup where 'e' is the input sequence and 'x' is lookahead. feaLib, on the otherhand, produces a type-6 lookup where both 'e' and 'x' are part of the input sequence.

I'm not really sure what to think about this? I suppose for my purposes it makes more sense to copy the feaLib behaviour, *if it is intentional*, but I think I would also print a warning indicating that this syntax is ambiguous.


also for further fun: if I change the above to,

```fea
feature cool {
    ignore sub e x;
    sub y e' by f;
} cool;
```

then the ignore statement gets promoted to type 6, but then the second statement seems to get added to the same subtable, which is probably a bug? the ttx looks like,

```xml
<LookupList>
      <!-- LookupCount=2 -->
      <Lookup index="0">
        <LookupType value="6"/>
        <LookupFlag value="0"/>
        <!-- SubTableCount=1 -->
        <ChainContextSubst index="0" Format="1">
          <Coverage>
            <Glyph value="e"/>
          </Coverage>
          <!-- ChainSubRuleSetCount=1 -->
          <ChainSubRuleSet index="0">
            <!-- ChainSubRuleCount=2 -->
            <ChainSubRule index="0">
              <!-- BacktrackGlyphCount=0 -->
              <!-- InputGlyphCount=2 -->
              <Input index="0" value="x"/>
[...]

PR #2950, opened 2023-01-09
[feaLib] Fix handled of "ignore" with unmarked glyphs

khaledhosny:
Fixes https://github.com/fonttools/fonttools/issues/2949

- If there are no marked glyphs in an "ignore" statement, issue a warning. The spec disallows this but makeotf allows it. It is most likely a typo, so a warning is warranted.
- Mark the first glyph not all the glyphs, this matches makeotf.
- In asFea() always mark the input glyph.
- Update the tests.

### DIFF
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -524,27 +524,33 @@ class Parser(object):
                 )
             return (prefix, glyphs, lookups, values, suffix, hasMarks)
 
-    def parse_chain_context_(self):
+    def parse_ignore_glyph_pattern_(self, sub):
         location = self.cur_token_location_
         prefix, glyphs, lookups, values, suffix, hasMarks = self.parse_glyph_pattern_(
             vertical=False
         )
-        chainContext = [(prefix, glyphs, suffix)]
-        hasLookups = any(lookups)
+        if any(lookups):
+            raise FeatureLibError(
+                f'No lookups can be specified for "ignore {sub}"', location
+            )
+        if not hasMarks:
+            error = FeatureLibError(
+                f'Ambiguous "ignore {sub}", there should be least one marked glyph',
+                location,
+            )
+            log.warning(str(error))
+            suffix, glyphs = glyphs[1:], glyphs[0:1]
+        chainContext = (prefix, glyphs, suffix)
+        return chainContext
+
+    def parse_ignore_context_(self, sub):
+        location = self.cur_token_location_
+        chainContext = [self.parse_ignore_glyph_pattern_(sub)]
         while self.next_token_ == ",":
             self.expect_symbol_(",")
-            (
-                prefix,
-                glyphs,
-                lookups,
-                values,
-                suffix,
-                hasMarks,
-            ) = self.parse_glyph_pattern_(vertical=False)
-            chainContext.append((prefix, glyphs, suffix))
-            hasLookups = hasLookups or any(lookups)
+            chainContext.append(self.parse_ignore_glyph_pattern_(sub))
         self.expect_symbol_(";")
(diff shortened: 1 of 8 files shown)


cmyr (approved):
Thanks for looking into this! I think this approach sounds good, and I will adopt it. :)

Lib/fontTools/feaLib/parser.py
```diff
@@ -524,46 +524,44 @@ def parse_glyph_pattern_(self, vertical):
...
+        if any(lookups):
+            raise FeatureLibError(
+                f'No lookups can be specified for "ignore {sub}"', location
+            )
+        if not hasMarks:
+            error = FeatureLibError(
+                f'Ambiguous "ignore {sub}", there should be least one marked glyph',
+                location,
+            )
+            log.warning(str(error))
```
anthrotype, line 541: no need to make a FeatureLibError exception if the only thing you need is str(error), just pass the string as log.warning message

anthrotype:
> if there are no unmarked glyphs in an "ignore" statement

I suppose you meant "if there are no marked glyphs"

Tests/feaLib/data/GSUB_5_formats.fea
```diff
@@ -1,16 +1,16 @@
 lookup GSUB5f1 {
-    ignore sub three four;
-    ignore sub four five;
+    ignore sub three' four';
```
anthrotype, line 2: when we parse an ignore statement with no marked glyphs, we issue a warning and assume that only the first is marked; however when we convert it back to string with asFea() we put the apostrophe on all the glyphs, not just the first... I find this a bit odd. Shouldn't asFea also include a similar logic?

anthrotype (approved):
Thanks, I'm ok with issuing a warning and match makeotf on this when parsing. But see my other comments on asFea().

Tests/feaLib/data/GSUB_5_formats.fea
khaledhosny: Only the marked glyph (either explicitly or implicitly) is marked in `asFea()`. I marked all the glyphs here to keep the expected ttx output unchanged (previously we were marking all the glyphs). This reminds me, I need to add a new builder test for the implicit marking.

Lib/fontTools/feaLib/parser.py
khaledhosny: This is to show the location with the warning (this is used elsewhere in the file, but if there is a better way, I can use it).

anthrotype: oh right, that's ok then.

Tests/feaLib/data/GSUB_5_formats.fea
anthrotype: > I marked all the glyphs here to keep the expected ttx output unchanged (previously we were marking all the glyphs)

but we don't want to keep the output unchanged, since we are no longer implicitly marking all the glyphs but only the first, no?

khaledhosny: I’m not sure what this test is testing, its name suggests it is testing various subtable formats of GSUB 5 lookup type, so I tried to keep the output unchanged since I presumed this is what it tests not the `ignore` statements.

anthrotype: ok then I would rather fix the input test fea such that it no longer has ambiguously unmarked glyphs

khaledhosny: This is what I did already.

anthrotype: oh I see, I got confused, sorry for the noise and thanks

PR #2957, opened 2023-01-17
[glyf] endPointsOfContours is unsigned

behdad:


anthrotype:
good catch btw

behdad:
Needs a very large number to hit...

justvanrossum:
32000 points is an awful lot for a glyph :)

I don't recall. Perhaps it's a mistake, perhaps it's based on an obsolete version of the spec.

justvanrossum (approved):

justvanrossum:
It seems FreeType also uses a signed short for the endPtsOfContours array:

https://github.com/freetype/freetype/blob/65f85237060a9a21abe8521b2ed14634fec005f1/src/truetype/ttgload.c#L374

I really wonder if it might have been int16 an old version of the spec.

justvanrossum:
> I really wonder if it might have been int16 in an old version of the spec.

The evidence rather points to it having been uint16 all along, and this was a genuine 23-year-old bug by yours truly.

https://web.archive.org/web/19991008110742/http://voyager.cns.ohiou.edu/~sadkins/web_library/fonts/font_specs/4-4ch.html

Let's make some fonts with 65535 points in the outlines!

PR #2958, opened 2023-01-17
VarComposites

behdad:
I've started implementing these. decompile() seems to work so far.


justvanrossum:
You should probably run black.

behdad:
Thanks. Will do every few commits.

behdad:
@justvanrossum I have a question for you:

Currently I'm setting all transformation components on VarComponent, and flags controls which ones are compiled. This is a bit uneasy to work with since if client modifies the transformation components they also need to modify the flags. We cannot just compile transformation components smartly since they should all match across masters when compiling a variable font...

An alternative would be to NOT set attribute for those components that should not be compiled into the font. That would be inconvenient to use when working with the component though.

Can you advise which one you prefer?

behdad:
@justvanrossum This branch now has compile/decompile/toXML/fromXML.

It's a bit crude as axisIndices are retained as integers not tags. To convert to tags I need to pass fvar down to glyf compile/decompile, which is a bit inconvenient.

justvanrossum:
> the glyf compiler should calculate flags based on presence of attributes only, and cannot drop components that have default value

I don't follow. Ideally (I think), all fields would be present as attributes, but only those that don't have the default value get compiled into the binary.

justvanrossum:
Ah yes, now I get it. The variations in gvar have to match up.

behdad:
Currently `Glyph.decompile` only takes the `glyf` table. Not sure how to pass `fvar` to it without changing API drastically.

behdad:
> Ah yes, now I get it. The variations in gvar have to match up.

Oh. If .flags is missing we can calculate it!

justvanrossum:
> Currently Glyph.decompile only takes the glyf table. Not sure how to pass fvar to it without changing API drastically.

Maybe `table__g_l_y_f.compile()` and `table__g_l_y_f.decompile()` can assign a reference to the `ttFont` for later use. It's a bit ugly, but then again: circular refs stopped being a big deal a long time ago, and I'm sure the API would have looked different if it were designed today... Or use a weakref.

behdad:
> circular refs

I'll try that. The cff table does it just fine.

behdad:
> > Currently Glyph.decompile only takes the glyf table. Not sure how to pass fvar to it without changing API drastically.
> 
> Maybe `table__g_l_y_f.compile()` and `table__g_l_y_f.decompile()` can assign a reference to the `ttFont` for later use. It's a bit ugly, but then again: circular refs stopped being a big deal a long time ago, and I'm sure the API would have looked different if it were designed today... Or use a weakref.

Or I can just save the axisTags. That's all I need.

justvanrossum:
> Or I can just save the axisTags. That's all I need. 

Ah yes. As long as they're refreshed on each compile/decompile call, that should be fine.

behdad:
All done.

behdad:
Now onto actually drawing these. I'll see if I can get my head around it.

behdad:
I still have to handle the following flag:
```
If ScaleY is missing: take value from ScaleX (to be discussed here: BlackFoundryCom/variable-components-spec#2)
```
It gets messy.

justvanrossum:
> It gets messy.

I bet :( Feel free to simplify this in the spec — we can be pragmatic about this.

behdad:
@justvanrossum I did a first pass at drawing as well. Works for my Hangul test font but not Hanzi (deeper components with recursive transformations I think). Debugging.

behdad:
> @justvanrossum I did a first pass at drawing as well. Works for my Hangul test font but not Hanzi (deeper components with recursive transformations I think). Debugging.

Fixed. Was a typo.

behdad:
> I still have to handle the following flag:
> 
> ```
> If ScaleY is missing: take value from ScaleX (to be discussed here: BlackFoundryCom/variable-components-spec#2)
> ```
> 
> It gets messy.

I made the compiler handle this flag automatically.

behdad:
This is now works in:
- ttx
- scaleUpem
- subset
- merge
- varLib.instancer when not instancing VarComposite axes

behdad:
Would be nice to have: WOFF2.

behdad:
@justvanrossum  This is ready for review.

justvanrossum:
> This is ready for review

Thanks, I'll try to review it before the weekend.

behdad:
> Would be nice to have: WOFF2.

Done.

behdad:
I've added tests also, so should be good on that front.

behdad:
@justvanrossum @anthrotype Anything I can do to help review?

behdad:
Humm. I'm facing a weird issue with WOFF2. Debugging.

behdad:
I tried adding a test for woff2 and I'm seeing a failure because when glyf table is being expanded, there's no `fvar` table in the font. The `ttFont`  only has 4 tables in it. It seems like the woff2 compiler works inplace and remove all tables and add them back one-by-one transformed and compiled, and ignoring the ttLib's table dependencies? That's a bit tricky to fix. Let me see if I can load the `glyf` table first thing.

behdad:
> I tried adding a test for woff2 and I'm seeing a failure because when glyf table is being expanded, there's no `fvar` table in the font. The `ttFont` only has 4 tables in it. It seems like the woff2 compiler works inplace and remove all tables and add them back one-by-one transformed and compiled, and ignoring the ttLib's table dependencies? That's a bit tricky to fix. Let me see if I can load the `glyf` table first thing.

Solved.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -1080,20 +1147,25 @@ def recalcBounds(self, glyfTable):
...
+            return struct.unpack(">h", self.data[:2])[0] == -1 if self.data else False
         else:
             return self.numberOfContours == -1
 
-    def __getitem__(self, componentIndex):
-        if not self.isComposite():
-            raise ttLib.TTLibError("can't use glyph as sequence")
-        return self.components[componentIndex]
+    def isVarComposite(self):
+        """Test whether a glyph has components"""
```
justvanrossum, line 1164:
```suggestion
        """Test whether a glyph has variable components"""
```

Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -29,6 +35,17 @@ def __init__(self, font, location, glyphsMapping):
...
 
+    def pushLocation(self, location, reset: bool):
+        self.locationStack.append(self.location)
+        if reset:
+            self.location = self.originalLocation.copy()
+        else:
+            self.location = self.location.copy()
+        self.location.update(location)
+
+    def popLocation(self):
```
justvanrossum, line 46: Consider making pushLocation() a context manager with the `contextlib.contextmanager` decorator.

Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -126,15 +142,61 @@ def draw(self, pen):
...
+            t = t.scale(comp.scaleX, comp.scaleY)
+            t = t.skew(-comp.skewX / 180 * math.pi, comp.skewY / 180 * math.pi)
+            t = t.translate(-comp.tCenterX, -comp.tCenterY)
+
+            if isPointPen:
+                tPen = TransformPointPen(pen, t)
+            else:
+                tPen = TransformPen(pen, t)
+
+            self.glyphSet.pushLocation(
```
justvanrossum, line 191: If `pushLocation` were a context manager, you could use `with`, and get rid of the pop call.

justvanrossum (commented):
Thanks Behdad, this looks very good.

It would be nice if the test coverage were better, esp. in the `glyf` table implementation.

Lib/fontTools/ttLib/ttGlyphSet.py
behdad: Done. Thanks.

behdad:
> It would be nice if the test coverage were better, esp. in the glyf table implementation.

CodeCov is currently failing. I can add more tests to hit more codepaths when that comes back up.

behdad:
> It would be nice if the test coverage were better, esp. in the glyf table implementation.

`glyf` is almost all covered by tests now.

Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -29,6 +36,19 @@ def __init__(self, font, location, glyphsMapping):
...
+    @contextmanager
+    def pushLocation(self, location, reset: bool):
+        self.locationStack.append(self.location)
+        if reset:
+            self.location = self.originalLocation.copy()
+        else:
+            self.location = self.location.copy()
+        self.location.update(location)
+
+        yield None
```
justvanrossum, line 48: The "correct" pattern is to wrap the yield in a try/finally:
```
        try:
            yield None
        finally:
            self.location = self.locationStack.pop()
```

Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -126,15 +145,60 @@ def draw(self, pen):
         how that works.
         """
         glyph, offset = self._getGlyphAndOffset()
+
+        if self.glyphSet.locationStack:
+            offset = 0  # Offset should only apply at top-level
```
anthrotype, line 151: btw, I didn't see any mention of this detail about the offset only applying at top-level in the BE draft glyf1.md spec... perhaps it should.

behdad: It's the undocumented behavior of the glyf table that toplevel glyphs are shifted to align their lsb to the hmtx table. Components never are. We need this here because VarComposite components are handled by ttGlyphSet whereas regular Composite components are handled in the glyf table.  So this is not really a glyf1-specific behavior, but I'll add a note there.

behdad: Done:

Note: While it is the (undocumented?) behavior of the glyf table that glyphs loaded are shifted to align their LSB to that specified in the hmtx table, much like regular Composite glyphs, this does not apply to component glyphs being loaded as part of a VariableComposite glyph.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -938,6 +999,12 @@ def compileComponents(self, glyfTable):
             data = data + struct.pack(">h", len(instructions)) + instructions
         return data
 
+    def compileVarComponents(self, glyfTable):
+        data = b""
+        for component in self.components:
+            data = data + component.compile(glyfTable)
+        return data
```
anthrotype, line 1006: maybe do `return b"".join(c.compile(glyfTable) for c in self.components)`

behdad:
I added a `VarTransform` class and used it instead. Code reads nicer.

behdad:
> I added a `VarTransform` class and used it instead. Code reads nicer.

To be used in pen-protocol extension.

behdad:
@justvanrossum I added `addVarComponent` to pen protocols. Please check.

Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -126,15 +153,59 @@ def draw(self, pen):
...
+        )
+
+        for comp in glyph.components:
+
+            with self.glyphSet.pushLocation(
+                comp.location, comp.flags & VarComponentFlags.RESET_UNSPECIFIED_AXES
+            ):
+                try:
+                    pen.addVarComponent(
+                        comp.glyphName, comp.transform, self.glyphSet.rawLocation
```
behdad, line 195: Maybe I should directly pass `comp.location` here. That would be more truthful to the data. Yeah, I think that's more correct.

behdad: Oh now I remember why I can't do that. If I do that then I also need to pass the `RESET` flag. So I kept it higher level. I'm not quite happy with this though. Happy to discuss. @justvanrossum

justvanrossum:
> I added addVarComponent to pen protocols. Please check.

Ooohh lovely. Will check soon and report back. Thanks!

Lib/fontTools/misc/transform.py
```diff
@@ -400,16 +400,27 @@ def Scale(x, y=None):
...
-    translateX: float = 0
-    translateY: float = 0
-    rotation: float = 0  # in degrees, counter-clockwise
-    scaleX: float = 1
-    scaleY: float = 1
-    skewX: float = 0  # in degrees, counter-clockwise
-    skewY: float = 0  # in degrees, counter-clockwise
-    tCenterX: float = 0
-    tCenterY: float = 0
+    def __init__(
```
anthrotype, line 403: you could use a dataclass so you get the constructor, eq and repr for free

anthrotype:
I am not sure if that VarTransform belongs to the fontTools.misc.transform module. That is a generic module for Affine 2D transformations that is not font-specific, hence why it is in misc sub-package. This class is only used for glyf1 VarComponents, so maybe belongs there. Besides, the skewX negates the angle, which I think is to match opentype slant conventions, so again font-specific. Also the name "VarTransform" looks problematic if we were to keep it inside misc.transform module; what's "Var" about it, besides it being used by glyf1 VarComponents.

I think there is a use case for having an object like this in misc.transform, i.e. one that allows one to construct a 2x3 affine transform from a set of decomposed translation, rotation, scale, skew and center (and ideally also a way to deconstruct a 2x3 affine into such decomposed vectors). But then we need to find a different name (e.g. DecomposedTransform maybe conveys this?) that is not tied to OT or glyf.

Also, I think we want to use dataclasses in this case (not namedtuple because you want it to be mutable as far as I can see).

justvanrossum:
Agreed that VarTransform doesn't belong in transform.py. It's fine in _g_l_y_f.py I think. Also +1 on using a dataclass for it.

anthrotype:
Another thing I am not quite conviced about is the need to pass this VarTransform object as a parameter the addVarComponent method in the pen protocol. Usually we stick to raw python data structures for these parameters (floats, bool, tuples, etc.). Maybe we have addVarComponent method take keyword arguments like translateX, trasnlateY, etc. instead, and one can do `pen.addVarComponent(..., **dataclasses.asdict(comp.transform))` to pass it down the pen?

anthrotype:
I find myself needing something like that in glyphsLib actually (in .glyphs3 components no longer have a 2x3 affine transform, but a set of optional pos/angle/scale attributes ... yep, skew is dead, and need to be able to convert between the two forms in both directions, so having it in misc.transform would be handy)

justvanrossum:
> Usually we stick to raw python data structures for these parameters

Hm, this object needs names, and I think there are too many to turn them all into kwargs. I would prefer the definition to be abstract: "an object with these atrtributes + types", so a named tuple would work equally well as a dataclass instance.

I now see that VarTransform can't really go into _g_l_y_f.py, due to the type declaration in basePen.py. Do we need fontTools.types perhaps? I mean, I'm not super against putting it in misc.transform, but it feels a bit out of place, too, there.

> compose_transform.py

Ah yes, it would be very useful to have that included in fonttools.

anthrotype:
we can keep it in misc.transform, renamed DecomposedTransform (like in your compose_transform.py), make it a dataclass..

> "an object with these atrtributes + types"

that's also fine.

Lib/fontTools/pens/basePen.py
```diff
@@ -131,6 +132,19 @@ def addComponent(
...
+        glyphName: str,
+        transformation: VarTransform,
+        location: Dict[str, float],
+    ) -> None:
+        """Add a VarComponent sub glyph. The 'transformation' argument
+        must be a VarTransform from the fontTools.misc.transform module,
+        and the 'location' argument must be a dictionary mapping axis tags
+        to their locations.
+        """
+        raise NotImplementedError
```
anthrotype, line 146: IIUC, in the ttGlyphSet.draw/drawPoints you are interpreting NotImplementedError exception as a signal that the given pen cannot handle drawing varComponents and thus one wishes to "decompose" them. However, wouldn't it be better to use AttributeError (or hasattr(pen, "addVarComponent") instead of NotImplementedError? The latter usually is used for astract classes where the expectation is that all concrete subclasses _must_ implement the abstract method, otherwise the concrete class cannot be instantiated. Whereas in your case, this addVarComponet seems to be intentionally designed to be "optionally implementable".
An optional method with that interface looks more suitable for this maybe. So if a pen wants to handle them varComponents it can define such a method, otherwise they will be drawn decomposed.

Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -126,15 +153,59 @@ def draw(self, pen):
...
+                    )
+                except NotImplementedError:
```
anthrotype, line 197: this could fail with AttributeError if a pen only implements the current interface (qucks like a pen) without actually subclassing directly from the fonttools base pens... is it intentional? or should we not instead decompose in that case? (see my other comment)

Lib/fontTools/pens/basePen.py
anthrotype: or... could the whole decomposing logic be moved to the basePen.DecomposingPen (or another similar mixin), the one that takes a glyphSet to resolve references? Then you would no longer catch NotImplementedError from within TTGlyphSet.draw, but simply call the pen's addVarComponent method and leave it to the pen itself to do the decomposing (or not do it)... I don't know what's better, honestly.

Lib/fontTools/ttLib/ttGlyphSet.py
anthrotype: how about we decouple this discussion about enhancing the pen protocol to make it "aware" of varcomponents to a separate follow-up PR? I think before this changes, this PR was relatively self-contained (mostly about glyf and the rest simply accompanying changes to accommodate the latter), and almost at merging stage; whereas changing the pen protocol could have wider effects, maybe we need to think about it more.

Lib/fontTools/ttLib/ttGlyphSet.py
justvanrossum: Yeah, I just experienced this first-hand: my Fontra point pen does not inherit from AbstractPointPen (although it probably should).

That said: even flattening currently doesn't work for me like it did yesterday. Debugging.

Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -180,19 +187,24 @@ def _drawVarComposite(self, glyph, pen, isPointPen):
...
+                    t = comp.transform.toTransform()
+                    if isPointPen:
+                        tPen = TransformPointPen(pen, t)
+                    else:
+                        tPen = TransformPen(pen, t)
+
+                        if isPointPen:
```
justvanrossum, line 204: Found the problem: this section is indented one level too deep, so it only works for a segment pen.

Lib/fontTools/ttLib/ttGlyphSet.py
justvanrossum: I pushed the fix. Hope you don't mind.

Lib/fontTools/ttLib/ttGlyphSet.py
anthrotype: I don't immediately understand what a "raw" location means.. Passing through the comp.location _and_ the reset flag seems better to me, if the intent is to delegate to the pen implementation

justvanrossum: The reset flag is mostly an optimization, we were hoping to keep the pen protocol stuff higher level so it will not have to worry about it. Especially since the compiler can figure out the best setting for it.

That said, you've got a point about fleshing out the pen stuff in a separate PR. (Although it could also be that we're almost there.)

justvanrossum:
GuessSmooth[Point]Pen needs to pass var components through. Will gladly contribute that, but will await a decision whether the pen stuff should be separated from this PR or not.

behdad:
> I see @justvanrossum wrote something along this in https://github.com/googlefonts/variable-components-in-ufo/blob/main/compose_transform.py, maybe it could be integrated in misc.transform..

I have half of that now. Will add the other half.

Lib/fontTools/pens/basePen.py
behdad: Moving the decomposing logic to the pen is hard because we need to keep more state (the font's location at least) that the pen doesn't have. That's why currently the VarComposite pens are drawn in ttGlyphSet.

behdad: I mean, I can make basePen.DecomposingPen call a private method of ttGlyphPen...

Lib/fontTools/ttLib/ttGlyphSet.py
anthrotype: ok, I got it now. It's more like the "actual" or "effective" varcomponent location at that particular node in tree, not simply its nominal location (i.e. as encoded in the glyf table) that takes into account its reset flag as well as its parents, basically abstracting away the whole reset flag business

Lib/fontTools/pens/basePen.py
behdad: > An optional method with that interface looks more suitable for this maybe. So if a pen wants to handle them varComponents it can define such a method, otherwise they will be drawn decomposed.

So, you suggest basePen wouldn't have this method at all? I find that clumsy.

Lib/fontTools/ttLib/ttGlyphSet.py
behdad: Yes. That's the idea.

Lib/fontTools/pens/basePen.py
anthrotype: > you suggest basePen wouldn't have this method at all?

yeah that'd also be weird. What I was a bit concerned about was that defining an addVarComponent on AbstractBasePen in the same way as the other abstract methods (i.e. return NotImplementedError) conveys the idea that any concrete class that inherit from them ought to implement it, whereas in this case it looks like a pen may deliberately "decide" to leave it unimplemented in order to trigger the ttGlyphSet decomposing behavior..

anthrotype:
> we document addVarComponent's as any Python object having these nine attribute

i'm fine with that. Remember now we are 3.8+ so we can actually start using the typing.Protocol for things like that

https://docs.python.org/3/library/typing.html#typing.Protocol

behdad:
> i'm fine with that. Remember now we are 3.8+ so we can actually start using the typing.Protocol for things like that
> 
> https://docs.python.org/3/library/typing.html#typing.Protocol

I might need help with that.

Lib/fontTools/misc/transform.py
```diff
@@ -346,6 +349,44 @@ def toPS(self):
         """
         return "[%s %s %s %s %s %s]" % self
 
+    def toDecomposed(self) -> "DecomposedTransform":
```
justvanrossum, line 352: Not sure if it's better, but I was thinking rather of a class method on DecomposedTransform:
```python
decomp = DecomposedTransform.fromTransform(t)
```
That way we can easily convert 6-tuples, too.

behdad: I can add both?

anthrotype:
thanks for adding Transform.toDecompose() 👍 
Is that inconsistency between skewX direction (clockwise in Transform, counter-clockwise in DecomposedTransform) actually needed? Shouldn't they follow the same direction? I propose to match the way the current Transform treat skewX (angle of vertical lines relative to the Y axis, in clockwise direction)

Lib/fontTools/misc/transform.py
justvanrossum: Yeah, why not.

anthrotype: I also like the classmethod on DecomposedTransform

anthrotype: we could even have both

anthrotype:
since the two classes live in the same module i would expect them to match with one another.. the caller could take care of applying a minus sign to the skewX

behdad:
> since the two classes live in the same module i would expect them to match with one another.. the caller could take care of applying a minus sign to the skewX

Yeah unfortunate indeed. It would definitely cause headaches if we represent skewX one way in UFO and another way in the binary though, but probably cleanest from fonttools point of view.

anthrotype:
also would be nice to document not just the direction and the fact that the angles are in degrees (or radians, another minor inconsistency between the two classes), but also explicitly relative to what (x or y axes) the angle is measured, it might be obvious to you but I get always confused between skewX/Y

behdad:
> also would be nice to document not just the direction and the fact that the angles are in degrees (or radians, another minor inconsistency between the two classes), but also explicitly relative to what (x or y axes) the angle is measured, it might be obvious to you but I get always confused between skewX/Y

I would do if I knew for certain :)))

behdad:
> * Be inconsistent with the spec how we represent skew

I changed it to this.

justvanrossum:
> since the two classes live in the same module

Maybe they should not, then? 

DecomposedTransform is a different beast with different requirements, as also witnessed by the degree/radians difference.

behdad:
> I mean, I can make basePen.DecomposingPen call a private method of ttGlyphPen...

I tried this. It's too fragile. Because if BasePen doesn't decompose VarComposite glyphs, it breaks all code that previously was working because now they won't handle addVarComposite(). As you said addVarComposite is an opt-in, not mandatory method now.

anthrotype:
i think they do belong to the same module. The radians/degree difference can also be amended, we could make all radians and shift responsibility to the caller to convert to/from degrees. It's nice they are close to each other, they represent the same thing in the end.
This convention whereby positive skewX/Y angles make vertical/horizontal lines "converge" towards one another in the positive X/Y quadrant seems to be the norm, svg skewx and skewy attributes effectively do the same as we currently do in Transform.skew method (with the difference that in SVG the y axis is flipped and goes down, but if you ignore that, they work the same): i.e. skewX angle is the clockwise angle between vertical lines relative to the Y axis; skewY is the counter-clockwise angle between horizontal lines and the X axis. Makes sense that it's the way it is.

justvanrossum:
Ok, you mostly convinced me. 

(Although in UFO I would not want to see radians. Hm, neither in TTX, for that matter.)

anthrotype:
> the degree/radians difference

this difference could be justified with the fact that the decomposed form using degrees is supposed to be more human-friendly, higher-level than the low-level 2x3 affine using radians.

justvanrossum:
> this difference could be justified with the fact that the decomposed form using degrees is supposed to be more human-friendly, higher-level than the low-level 2x3 affine using radians.

I would be slightly happier indeed if we could keep the angles in degrees.

behdad:
Okay, then code is good as is then with my adjustment?

behdad:
> > this difference could be justified with the fact that the decomposed form using degrees is supposed to be more human-friendly, higher-level than the low-level 2x3 affine using radians.
> 
> I would be slightly happier indeed if we could keep the angles in degrees.

I think Cosimo agreed to that.

behdad:
So. There's no `DecomposingRecordingPointPen`. There's a discrepancy that `RecordingPen` records components whereas `RecordingPointPen` decomposes components.

What should `RecordingPointPen` do with VarComposites?

behdad:
> What should `RecordingPointPen` do with VarComposites?

Currently it decomposes them, like Composites are.

anthrotype:
RecordingPointPen should pass them through, same way as RecordingPen. We can add a corresponding DecomposingRecordingPointPen for consistency (I think I made one a couple times someplace else which now I can't remember)

behdad:
> RecordingPointPen should pass them through, same way as RecordingPen. We can add a corresponding DecomposingRecordingPointPen for consistency (I think I made one a couple times someplace else which now I can't remember)

Okay.

Lib/fontTools/pens/basePen.py
anthrotype: > As you said addVarComposite is an opt-in, not mandatory method now.

to me raising NotImplementedError communicates a sense of urgency, "implement me!". However, now this seems to be part of the interface itself, a way for a pen to deliberately shift the responsibility of decomposing varcomponents to the "context" (the DecomposingRecordingPen overrides the RecordingPen.addVarComponents by re-raising NotImplementedError exactly to communicate to TTGlyphSet that it indeed wants them to be decomposed).
Maybe this is ok and I am overthinking this. Maybe it's ok that for the time being only TTGlyphSet knows what varComponents are and can decompose them.

behdad: Would `raise AttributeError` would look nicer to you?

behdad: With a comment?

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -1080,20 +1145,25 @@ def recalcBounds(self, glyfTable):
         recomputed when the ``coordinates`` change. The ``table__g_l_y_f`` bounds
         must be provided to resolve component bounds.
         """
-        coords, endPts, flags = self.getCoordinates(glyfTable)
-        self.xMin, self.yMin, self.xMax, self.yMax = calcIntBounds(coords)
+        try:
+            coords, endPts, flags = self.getCoordinates(glyfTable)
+            self.xMin, self.yMin, self.xMax, self.yMax = calcIntBounds(coords)
+        except NotImplementedError:
```
justvanrossum, line 1151: The `pass` here causes the old xMin, yMin etc. to be left alone, which makes messes up the lsb calculation in ttGlyphSet. I *think* you work around this by setting offset to zero in there's a location stack, but I fear the "normal" drawing (not as part of a composite) of such a glyph is still messed up.

Lib/fontTools/pens/basePen.py
anthrotype: hm.. I didn't think of that, actually. Yeah, it kind of communicates better the idea (even if it's technically lying). It's like the equivalent of not defining the method at all (which is the norm for all existing pens), but communicates that it is a possible optional method with given signature that one may want to implement in order to handle varcomponents (unless they want them decomposed). Good idea!

Lib/fontTools/ttLib/tables/_g_l_y_f.py
behdad: My understanding was that offset doesn't apply to components. But now that I come to think about it, I'm probably wrong here. Ugh. This is really messy :(.

behdad: In HarfBuzz we only apply the shift to toplevel glyph. That's what tripped me. I'm utterly confused now. Let me study fonttools code a bit, and possibly construct a font.

justvanrossum: Sorry, I'm also confused: I'm observing things in my code that do not happen in your code, and it relates to the offset. My conclusion above may be wrong. I'll debug further.

behdad: So indeed HarfBuzz and FontTools have different behavior in this regard currently.  I subset MutatorSans to have the Composite glyph Á. Then I changed the component accent glyph's lsb in the hmtx to shift it horizontall. HarfBuzz still renders the Á unchanged, while FontTools render it with the accent shifted.

I need volunteers to test this in Windows / Mac. The font is attached.
[comp.ttf.gz](https://github.com/fonttools/fonttools/files/10581839/comp.ttf.gz)

behdad: > I need volunteers to test this in Windows / Mac. The font is attached.

Mac shows the accent shifted. So HarfBuzz is in the wrong *compared to Mac*. We need Windows testing as well.

behdad: > Mac shows the accent shifted. So HarfBuzz is in the wrong compared to Mac. We need Windows testing as well.

FreeType agrees with HarfBuzz though.

behdad: For reference:

HarfBuzz and FreeType rendering (applying offset only at top-level; not applying to components):
image

behdad: Mac rendering:
Screenshot 2023-02-03 at 12 00 30 PM

behdad: FontTools rendering:
image

behdad:
> [[basePen] Add addVarComposite() to DecomposingPen](https://github.com/fonttools/fonttools/pull/2958/commits/f4544da777a16a880a2efb3d54236186c3b76db9)

@anthrotype Or did you want the `AttributeError` in the AbstractPen (as well)?

Lib/fontTools/ttLib/tables/_g_l_y_f.py
behdad: Windows rendering (different font but same hack):
image

So, looks like Mac is in the minority here and FontTools follows it. @nedley

behdad: The implications for FontTools is that `DecomposingPen` cannot work as it does, and the Composite decomposing should also be moved to `ttGlyphSet`. @justvanrossum

behdad: @nedley The issue we are discussing is that if a component glyph in a glyf table has different xMin and hmtx lsb, Windows and FreeType do _not_ shift it by the difference when using the component, whereas CoreText seems to shift.

Ie. Windows and FreeType only shift top-level glyphs by the difference of xMin/lsb.

behdad: @justvanrossum In fact, fontTools itself has internal inconsistency in that `decomposingPen` applies offset to components, but `_g_l_y_f.Glyph.getCoordinates` doesn't!

behdad:
> > [[basePen] Add addVarComposite() to DecomposingPen](https://github.com/fonttools/fonttools/pull/2958/commits/f4544da777a16a880a2efb3d54236186c3b76db9)
> 
> @anthrotype Or did you want the `AttributeError` in the AbstractPen (as well)?

I did.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
justvanrossum: We should probably open a new issue for this.

The fact that drawing components should not apply the offset is new to me. How is this written in the spec?

It indeed makes the current way to decompose things impossible to handle correctly for non-zero offsets.

That said: I'm still confused. Do I understand correctly that the lsb in hmtx should make no difference for glyphs used as (var) components?

Yet, if I hack up Tests/ttLib/data/varc-ac00-ac01.ttf to have different lsb values for glyph00003, glyph00004 and glyph00005, I don't get a correct rendered uniAC00. I tested with
```
python -m fontTools.pens.svgPathPen Tests/ttLib/data/varc-ac00-ac01.ttf 가 > test.svg
```
And
```
python -m fontTools.pens.svgPathPen varc-ac00-ac01-lsb-hack.ttf 가 > test-lsb-hack.svg
```
with the attached font.
[varc-ac00-ac01-lsb-hack.ttf.zip](https://github.com/fonttools/fonttools/files/10582330/varc-ac00-ac01-lsb-hack.ttf.zip)

Perhaps I am misunderstanding things.

behdad: > We should probably open a new issue for this.

Agreed. Will do.

> The fact that drawing components should not apply the offset is new to me. How is this written in the spec?

The HarfBuzz code calls the whole offset application "undocumented behavior".   Indeed, I cannot find anywhere in the spec that says glyph shape should be shifted to align xMin to lsb.

> It indeed makes the current way to decompose things impossible to handle correctly for non-zero offsets.

Right.

> That said: I'm still confused. Do I understand correctly that the lsb in hmtx should make no difference for glyphs used as (var) components?

That's what Windows and FreeType seem to be doing.

> Yet, if I hack up Tests/ttLib/data/varc-ac00-ac01.ttf to have different lsb values for glyph00003, glyph00004 and glyph00005, I don't get a correct rendered uniAC00.

Humm. Let me try. That's exactly why I added the `offset = 0` change.

justvanrossum: I think in general, it is undesirable to have lsb to be not equal to xMin. fontmake generally doesn't (can't?) produce fonts like that, and I'm pretty sure neither does Glyphs.

I would be fine with fonttools saying: "lsb != xMin is not supported in the glyphset draw API for glyphs that are used as components".

The fact that you only now discovered a mismatch between Windows and  macOS says a lot about how uncommon this is in real-world fonts.

That said, I was hoping to fix the problem I am seeing on my end (with hacked up Fontra code to use addVarComponent to extract variable components from .ttf) by fixing the test font to have lsb == xMin in all glyphs. And that didn't work, so I must look further. Unless something is fishy in the ttGlyphSet implementation, which is entirely possible, given how incredibly hairy this offset stuff is.

behdad: > Yet, if I hack up Tests/ttLib/data/varc-ac00-ac01.ttf to have different lsb values for glyph00003, glyph00004 and glyph00005, I don't get a correct rendered uniAC00.

Something broken in my `fromXML` impl I think. If I diff the two fonts, all the `translateX` and `translateY` are gone. That's a relief. Debugging.

behdad: > Something broken in my fromXML impl I think. If I diff the two fonts, all the translateX and translateY are gone. That's a relief. Debugging.

The compile code was broken. Fixed. With that I recreated your work and it renders fine still; with FontTools and HarfBuzz.

justvanrossum: > Indeed, I cannot find anywhere in the spec that says glyph shape should be shifted to align xMin to lsb.

The only thing I find is this, which is rather indirect:

> Note: The scaler will perform better if the glyph coordinates have been created such that the xMin is equal to the lsb. For example, if the lsb is 123, then xMin for the glyph should be 123. If the lsb is -12 then the xMin should be -12. If the lsb is 0 then xMin is 0. If all glyphs are done like this, set bit 1 of flags field in the 'head' table.
 
https://learn.microsoft.com/en-us/typography/opentype/spec/glyf#glyph-headers

And nothing about components indeed.

justvanrossum:
I ran into an issue with using addVarComponent: a shallow copy is not enough when instantiating vaiable components. The following fixes it.

```diff
diff --git a/Lib/fontTools/ttLib/ttGlyphSet.py b/Lib/fontTools/ttLib/ttGlyphSet.py
index 12afd4cae..d6f49d1bb 100644
--- a/Lib/fontTools/ttLib/ttGlyphSet.py
+++ b/Lib/fontTools/ttLib/ttGlyphSet.py
@@ -3,7 +3,7 @@
 from abc import ABC, abstractmethod
 from collections.abc import Mapping
 from contextlib import contextmanager
-from copy import copy
+from copy import copy, deepcopy
 from types import SimpleNamespace
 from fontTools.misc.fixedTools import otRound
 from fontTools.misc.loggingTools import deprecateFunction
@@ -288,12 +288,14 @@ def _setCoordinates(glyph, coord, glyfTable):
 
     if glyph.isComposite():
         assert len(coord) == len(glyph.components)
-        glyph.components = [copy(comp) for comp in glyph.components]  # Shallow copy
+        # Shallow copy is enough: all attributes that we change are scalars
+        glyph.components = [copy(comp) for comp in glyph.components]
         for p, comp in zip(coord, glyph.components):
             if hasattr(comp, "x"):
                 comp.x, comp.y = p
     elif glyph.isVarComposite():
-        glyph.components = [copy(comp) for comp in glyph.components]  # Shallow copy
+        # Deep copy needed because we mutate comp.transformation and comp.location
+        glyph.components = [deepcopy(comp) for comp in glyph.components]
         for comp in glyph.components:
             coord = comp.setCoordinates(coord)
         assert not coord
```
If you want I can push this change (but I'm soon signing off for today).

behdad:
> If you want I can push this change (but I'm soon signing off for today).

Ah right. Yes please. Thanks.

behdad:
> ```diff
> +        # Deep copy needed because we mutate comp.transform and comp.location
> ```

The location one is recreated, but transform is modified in place. This broke after I added `DecomposedTransform`. Thanks for the fix.

justvanrossum:
> The location one is recreated, but transform is modified in place. This broke after I added DecomposedTransform. Thanks for the fix.

Pushed.

justvanrossum:
> The location one is recreated, but transform is modified in place

In retrospect: would it perhaps be better to copy the transform explicitly in setCoordinates?

behdad:
> In retrospect: would it perhaps be better to copy the transform explicitly in setCoordinates?

Yeah I can do that. I'll take care of it.

behdad:
> > In retrospect: would it perhaps be better to copy the transform explicitly in setCoordinates?

That's the better fix anyway, to reset fields not in the flag to their defaults.

justvanrossum:
Hi @behdad, after learning this from the [head table spec](https://learn.microsoft.com/en-us/typography/opentype/spec/head):

> Note that, in a variable font with TrueType outlines, the left side bearing for each glyph must equal xMin, and bit 1 in the flags field must be set. [ ... ]

...I would love to request for the variable component test fonts in Tests/ttLib/data to be fixed to ensure `lsb == xMin` for all glyphs. I can gladly patch and push to this PR branch. I don't know how these were originally made, perhaps you'd prefer to do this yourself? I would just manually tweak the ttx and recompile.

behdad:
> ...I would love to request for the variable component test fonts in Tests/ttLib/data to be fixed to ensure `lsb == xMin` for all glyphs. I can gladly patch and push to this PR branch. I don't know how these were originally made, perhaps you'd prefer to do this yourself? I would just manually tweak the ttx and recompile.

Feel free to patch and push. I'm working on my tools to fix the original fonts. I find it hard to calculate xMin (and hence lsb) for VarComposite glyphs themselves though, so right now I'm setting both to 0.

justvanrossum:
> Feel free to patch and push

Will do, thanks.

> I find it hard to calculate xMin (and hence lsb) for VarComposite glyphs themselves though

While there may be some chicken & egg involved: it should be doable with ControlBoundsPen via ttGlyphSet, after you've first set them to zero.

behdad:
> > Feel free to patch and push
> 
> Will do, thanks.

Looks like `varc-ac00-ac01.ttf` has correct lsb, no? It's probably just `varc-6868.ttf` that is broken.

behdad:
> > > Feel free to patch and push
> > 
> > 
> > Will do, thanks.
> 
> Looks like `varc-ac00-ac01.ttf` has correct lsb, no? It's probably just `varc-6868.ttf` that is broken.

Fixed.

behdad:
> While there may be some chicken & egg involved: it should be doable with ControlBoundsPen via ttGlyphSet, after you've first set them to zero.

Right. It's too much logic to duplicate otherwise in the build stage.

justvanrossum:
I don't see correct lsb values in varc-ac00-ac01.ttf:
```
  <hmtx>
    <mtx name=".notdef" width="1000" lsb="100"/>
    <mtx name="glyph00003" width="966" lsb="189"/>
    <mtx name="glyph00004" width="966" lsb="627"/>
    <mtx name="glyph00005" width="966" lsb="179"/>
    <mtx name="uniAC00" width="966" lsb="47"/>
    <mtx name="uniAC01" width="966" lsb="35"/>
  </hmtx>

    <TTGlyph name="glyph00003" xMin="-414" yMin="-153" xMax="67" yMax="14">
    <TTGlyph name="glyph00004" xMin="-311" yMin="-97" xMax="28" yMax="354">
    <TTGlyph name="glyph00005" xMin="-516" yMin="-244" xMax="63" yMax="14">
    <TTGlyph name="uniAC00" xMin="46" yMin="-107" xMax="959" yMax="851">
    <TTGlyph name="uniAC01" xMin="34" yMin="-116" xMax="959" yMax="851">
```

justvanrossum:
It's *especially* the component glyph lsb's that are messing things up on my side.

behdad:
> I don't see correct lsb values in varc-ac00-ac01.ttf:

Weird. I wonder where those lsb's coming from then. Let me take another look at my tools. Feel free to fix up.

behdad:
> Weird. I wonder where those lsb's coming from then

I see. I'm getting the lsb from the original glyph in the font, which happens to not match my new font. I'll try to fix my tool but feel free to patch the font yourself as well.

behdad:
I'm working on updating the font.

behdad:
Done.

behdad:
The decomposeTransform logic here has a bug. See:
https://math.stackexchange.com/questions/13150/extracting-rotation-scale-values-from-2d-transformation-matrix#comment8965285_2888105

behdad:
> The decomposeTransform logic here has a bug. See:
> https://math.stackexchange.com/questions/13150/extracting-rotation-scale-values-from-2d-transformation-matrix#comment8965285_2888105

I once tried to derive this but apparently I've missed a lot of cases... We need to find a better derivation.
https://lists.w3.org/Archives/Public/www-style/2010Sep/0697.html

behdad:
> Maybe I just add a special case at the top...

Or maybe the behavior is expected? @justvanrossum

behdad:
It can be fixed by propagating the sign of a/d to the corresponding scale components. But my confidence is low to make such changes without extensive tests.

justvanrossum:
TBH I just grabbed that code from SO without much thought, and definitely not much understanding of the maths. Not sure what to do, but special cases for [a 0 0 d e f] and maybe [0 b c 0 e f] may not be so bad.

behdad:
> TBH I just grabbed that code from SO without much thought, and definitely not much understanding of the maths. Not sure what to do, but special cases for [a 0 0 d e f] and maybe [0 b c 0 e f] may not be so bad.

A patch like this should work, I just haven't tested it beyond the trivial case:
```
@@ -434,15 +434,18 @@ class DecomposedTransform:
         scaleX = scaleY = 0
         skewX = skewY = 0
 
+        def sign(v):
+            return +1 if v >= 0 else -1
+
         # Apply the QR-like decomposition.
         if a != 0 or b != 0:
-            r = math.sqrt(a * a + b * b)
-            rotation = math.acos(a / r) if b > 0 else -math.acos(a / r)
+            r = sign(a) * math.sqrt(a * a + b * b)
+            rotation = math.acos(a / r) if b >= 0 else -math.acos(a / r)
             scaleX, scaleY = (r, delta / r)
             skewX, skewY = (math.atan((a * c + b * d) / (r * r)), 0)
         elif c != 0 or d != 0:
-            s = math.sqrt(c * c + d * d)
-            rotation = math.pi / 2 - (math.acos(-c / s) if d > 0 else -math.acos(c / s))
+            s = sign(d) * math.sqrt(c * c + d * d)
+            rotation = math.pi / 2 - (math.acos(-c / s) if d >= 0 else -math.acos(c / s))
             scaleX, scaleY = (delta / s, s)
             skewX, skewY = (0, math.atan((a * c + b * d) / (s * s)))
         else:
```

behdad:
Added tests and fixed.

behdad:
> Added tests and fixed.

Err. Actually fixing.

Lib/fontTools/pens/pointPen.py
```diff
@@ -60,6 +61,21 @@ def addComponent(
...
+        location: Dict[str, float],
+        identifier: Optional[str] = None,
+        **kwargs: Any,
+    ) -> None:
+        """Add a VarComponent sub glyph. The 'transformation' argument
+        must be a DecomposedTransform from the fontTools.misc.transform module,
+        and the 'location' argument must be a dictionary mapping axis tags
+        to their locations.
+        """
+        raise NotImplementedError
```
anthrotype, line 77: you made the AbstractPen (segment-oriented protocol) raise AttributeError, but here in the AbstractPointPen you are raising NotImplementedError.. you should do the same for both respective's addVarComponent methods

anthrotype: since the TTGlyphSet now only catches AttributeError, and not also NotImplementedError, it means that a pointpen that inherits from AbstractPointPen but doesn't implement addVarComponent will now crash when used with the new varComponents-aware TTGlyphSet

anthrotype:
I see you added some generic xml-roundtripping tests for varComponents in the unit tests for scaleUpem module.. that's not ideal, althoughj some tests is always better than no tests. The glyf table module has unit test module of its own, in [Tests/ttLib/tables/glyf_test.py](https://github.com/fonttools/fonttools/blob/main/Tests/ttLib/tables/_g_l_y_f_test.py), that's where I'd expect to find things like those.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -1299,6 +1377,8 @@ def trim(self, remove_hinting=False):
                 i += 2 + instructionLen
             # Remove padding
             data = data[:i]
+        elif self.isVarComposite():
+            raise NotImplementedError
```
anthrotype, line 1381: if varComposites cannot (unlike simple and regular composite glyphs) have hinting instructions, should this `pass` or return early instead of raising NotImplementedError? Just a few lines above within the same `trim` method you do:

```pyhon
                elif self.isVarComposite():
                    pass  # Doesn't have hinting
```

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -1165,6 +1235,8 @@ def getCoordinates(self, glyfTable):
                 allCoords.extend(coordinates)
                 allFlags.extend(flags)
             return allCoords, allEndPts, allFlags
+        elif self.isVarComposite():
+            raise NotImplementedError
```
anthrotype, line 1239: maybe add some error message to this, in case somebody attempts to use the `_g_l_y_f.Glyph.draw` and `drawPoints` method directly, suggesting to use the _TTGlyph.draw/drawPoints (which do handle VarComposites)

Lib/fontTools/ttLib/tables/_g_l_y_f.py
anthrotype: > The pass here causes the old xMin, yMin etc. to be left alone

so, what do we expect recalcBounds to do with bbox when a glyph contain varcomponents?

justvanrossum: I think we have to use ttGlyphSet and ControlBoundsPen and do it the slow but correct way. There may be some chicken/egg problem involved, as that drawing code also calls recalcBounds.

However: I don't think the drawing code *should* call recalcBounds, because the only reason it does so is to calculate the offset, in case xMin is different from lsb. But: we do not have to cater for that situation, as lsb *has* to be equal to xMin for variable fonts.

I'm afraid it may take some refactoring to get this right. Perhaps postpone until after this PR lands?

Lib/fontTools/ttLib/tables/_g_l_y_f.py
behdad: But this is about removing padding, not just hinting. It can and should eventually be implemented. I'll get on implementing it.

behdad:
> I see you added some generic xml-roundtripping tests for varComponents in the unit tests for scaleUpem module.. that's not ideal, althoughj some tests is always better than no tests. The glyf table module has unit test module of its own, in [Tests/ttLib/tables/glyf_test.py](https://github.com/fonttools/fonttools/blob/main/Tests/ttLib/tables/_g_l_y_f_test.py), that's where I'd expect to find things like those.

Ah thanks. I was wrongly looking for ttx tests and not finding them. Will move there.

behdad:
This seems to be in good shape again.

justvanrossum:
> This seems to be in good shape again.

Looks good to me i general, and the `pen.addVarComponent()` api is now usable for glyph extraction uses cases.

I do worry that we don't have a solution for `glyph.recalcBounds()` for variable components. The factoring between `_g_l_y_f.py` and `ttGlyphSet.py` makes it seemingly difficult to implement. At the same time, `ttGlyphSet._setCoordinates()` does too much work in some cases, where we don't need the `lsb - xMin` offset (because is is supposed to be zero anyway). I have vague idea of adding a new class "DrawingContext" to `_g_l_y_f.py`, that would contain the common code for `recalcBounds` and `ttGlyphSet`. Needs sketching out.

behdad:
> I do worry that we don't have a solution for `glyph.recalcBounds()` for variable components. The factoring between `_g_l_y_f.py` and `ttGlyphSet.py` makes it seemingly difficult to implement. At the same time, `ttGlyphSet._setCoordinates()` does too much work in some cases, where we don't need the `lsb - xMin` offset (because is is supposed to be zero anyway). I have a vague idea of adding a new class "DrawingContext" to `_g_l_y_f.py`, that would contain the common code for `recalcBounds` and `ttGlyphSet`. Needs sketching out.

We can do that in a followup PR. Sad thing is we don't have the font around; I suppose we can add a glyf.font link.

behdad:
@anthrotype @justvanrossum Can we move towards merging this?

anthrotype (approved):
LGTM

a follow up PR can deal with recalcBounds for varComposites

anthrotype:
> We can do that in a followup PR.

let's also file an issue tracking implementation of recalcBouds for varComposites

behdad:
> > We can do that in a followup PR.
> 
> let's also file an issue tracking implementation of recalcBouds for varComposites

https://github.com/fonttools/fonttools/issues/2986

Issue #2962, opened 2023-01-22
Cannot resave TrueType font with sbix table containing glyphs with ref


PoomSmart:
Specifically, this line:

https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/ttLib/tables/sbixGlyph.py#96

`self.rawdata = sstruct.pack(sbixGlyphHeaderFormat, self) + self.imageData`

Glyphs with `graphicType` as `dupe` do not have `imageData` so what happens here is a `bytes` object is being concatenated with `NoneType`.
Error: `TypeError: can't concat NoneType to bytes`

You can find this kind of font from for example, iOS 14/15/16 AppleColorEmoji font.

PR #2963, opened 2023-01-22
Handle glyphs with dupe graphic type on compile correctly

PoomSmart:
Instead of appending glyph raw data with `imageData` for `graphicType=dupe` which will always be `NoneType`, append with the reference glyph ID.

Closes https://github.com/fonttools/fonttools/issues/2962


behdad:
Thanks!

anthrotype (approved):

lanun135 (commented):
J

PR #2967, opened 2023-01-26
Odd check pattern

rsheeter:
Perhaps I simply misread but it looks like we might be saying continue if `P ⊂ C` and `C ⊂ P` which could be directly written as P == C?

Spotted while reading models.py in hopes of learning something.

### DIFF
--- a/Lib/fontTools/varLib/models.py
+++ b/Lib/fontTools/varLib/models.py
@@ -365,12 +365,12 @@ class VariationModel(object):
             # Walk over previous masters now
             for prev_region in regions[:i]:
                 # Master with extra axes do not participte
-                if not set(prev_region.keys()).issubset(locAxes):
+                if set(prev_region.keys()) != locAxes:
                     continue
                 # If it's NOT in the current box, it does not participate
                 relevant = True
                 for axis, (lower, peak, upper) in region.items():
-                    if axis not in prev_region or not (
+                    if not (
                         prev_region[axis][1] == peak
                         or lower < prev_region[axis][1] < upper
                     ):


Lib/fontTools/varLib/models.py
```diff
@@ -365,12 +365,12 @@ def _computeMasterSupports(self):
...
-                if not set(prev_region.keys()).issubset(locAxes):
+                if set(prev_region.keys()) != locAxes:
                     continue
                 # If it's NOT in the current box, it does not participate
                 relevant = True
                 for axis, (lower, peak, upper) in region.items():
-                    if axis not in prev_region or not (
+                    if not (
                         prev_region[axis][1] == peak
                         or lower < prev_region[axis][1] < upper
```
rsheeter, line 375: Noob question, I thought lower, peak, upper were inclusive in which case ... shouldn't this be <= ?

behdad:
> if P ⊂ C and C ⊂ P

I'm confused. Where's this in the code?  I see one subset-check only.

behdad:
> > if P ⊂ C and C ⊂ P
> 
> I'm confused. Where's this in the code? I see one subset-check only.

Oh now I understand. Let me think.

Lib/fontTools/varLib/models.py
behdad: If peak of one master is on the lower of another master, the two masters don't really interact and there's nothing to do.

behdad:
> > > if P ⊂ C and C ⊂ P
> > 
> > 
> > I'm confused. Where's this in the code? I see one subset-check only.
> 
> Oh now I understand. Let me think.

Yes I believe you are right. I just wrote the code they way it came to me. Update the comments maybe?

Issue #2965, opened 2023-01-25
pyftsubset on COLRv1 font gives unexpected results


RoelN:
I'm trying to isolate the "foreground_color" glyphs shown on https://roettsch.es/var_colrv1.html

<img width="1024" alt="image" src="https://user-images.githubusercontent.com/4570664/214676694-3fc4ec3b-9c61-4f42-8995-d7cbeb48124f.png">

Downloading the font used on that page, and using pyftsubset:

```
pyftsubset --verbose --unicodes="U+F0B00U+F0B01U+F0B02U+F0B03U+F0B04U+F0B05U+F0B06U+F0B07" test_glyphs-glyf_colr_1_variable.ttf
```

Then using the font in my own test page:

<img width="991" alt="image" src="https://user-images.githubusercontent.com/4570664/214677544-03e29bcb-d163-4fbe-a202-3683e503282a.png">

It looks like a `ClipList` has wrong values, except for the first glyph (0xf0b00). The first glyph has the same ClipBox as in the original font, the other 7 get a different ClipBox applied.

anthrotype:
Hey Roel, thanks for the report!
I cannot seem to reproduce the issue on my end. I downloaded the font and run the same pyftsubset command as yours, then made two test page for the original and subsetted font (you'll find them attached below).

[test_glyphs-glyf_colr_1_variable.zip](https://github.com/fonttools/fonttools/files/10508840/test_glyphs-glyf_colr_1_variable.zip)

When I load the pages (e.g. running a local http server with `python -m http.server 8081`) in either Chrome (both latest stable and canary) or Firefox both fonts display correctly.

Can you check the font you made is the same as the one I made?
What fonttools and browser version are you testing with?

RoelN:
Hey Cosimo,

I'm on TTX 4.38.0, currently the newest when installed via Homebrew. I haven't tried a local build of FontTools yet. But I guess there's a difference in those versions, as my subsetted version from the original ttf in your zip differs from your subsetted version:

```diff
diff --git a/differ.ttx b/differ.ttx
index d8f6188..c1b1c58 100644
--- a/differ.ttx
+++ b/differ.ttx
@@ -18,7 +18,7 @@
     <!-- Most of this table will be recalculated by the compiler -->
     <tableVersion value="1.0"/>
     <fontRevision value="1.0"/>
-    <checkSumAdjustment value="0x6c7b2c40"/>
+    <checkSumAdjustment value="0xfa84d3c0"/>
[...]

PR #2968, opened 2023-01-26
copy OTTableReader before copying/pickling lazily loaded table

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/2965

In otTables.py we have a custom implementation of ClipList.postRead where we build a `clips` dictionary of ClipBox records keyed by glyph name, ensuring each value is a distinct object (they may be shared for ranges of glyph IDs in the binary table):

https://github.com/fonttools/fonttools/blob/f73220816264fc383b8a75f2146e8d69e455d398/Lib/fontTools/ttLib/tables/otTables.py#L1325-L1332

We call `copy.copy(rec.ClipBox)`. If the TTFont was loaded lazily like the subsetter does by default (unless one passes --no-lazy), then the ClipBox record is only decompiled when any of its attribute is accessed (thanks to our magic `BaseTable.__getattr__`).

In python <= 3.10 we didn't need to do anything special and calling copy.copy on a lazy object would automatically decompile it (I don't know exactly how, something again about our magic `__getattr__`).

Whereas since python3.11, calling `copy.copy(rec.ClipBox)` does _not_ automatically decompile our lazy objects.

So we should define a custom `__getstate__` as explained in the [python docs](https://docs.python.org/3/library/copy.html) to control how an object state gets copied/pickled.

In the case of lazy objects we could either call `ensureDecompiled` before returing `self.__dict__`, or better keep the object in lazy mode make a shallow copy of its OTTableReader. This way each distinct copy owns a different reader and can be un-lazified independently instead of stepping on each other's toes.

I hope that makes sense. It does fix the bug @RoelN found out in #2956, now I need to figure out how to test this..

### DIFF
--- a/Lib/fontTools/ttLib/tables/otBase.py
+++ b/Lib/fontTools/ttLib/tables/otBase.py
@@ -896,6 +896,15 @@ class BaseTable(object):
             for subtable in self.iterSubTables():
                 subtable.value.ensureDecompiled(recurse)
 
+    def __getstate__(self):
+        # before copying/pickling 'lazy' objects, make a shallow copy of OTTableReader
+        # https://github.com/fonttools/fonttools/issues/2965
+        if "reader" in self.__dict__:
+            state = self.__dict__.copy()
+            state["reader"] = self.__dict__["reader"].copy()
+            return state
+        return self.__dict__
+
     @classmethod
     def getRecordSize(cls, reader):
         totalSize = 0
--- a/Tests/subset/subset_test.py
+++ b/Tests/subset/subset_test.py
@@ -1388,8 +1388,8 @@ def colrv1_path(tmp_path):
         clipBoxes={
             "uniE000": (0, 0, 200, 300),
             "uniE001": (0, 0, 500, 500),
-            "uniE002": (100, 100, 400, 400),
-            "uniE003": (-50, -50, 350, 350),
+            "uniE002": (-50, -50, 400, 400),
+            "uniE003": (-50, -50, 400, 400),
         },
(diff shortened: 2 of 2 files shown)


behdad:
Makes sense. Thanks for figuring this out!

behdad (approved):

anthrotype:
> ensureDecompiled

actually calling ensureDecompiled in `__getstate__` (or in `__copy__`) would be weird in that it would modify the state of self in a method that is supposed to "get" something (as opposed to change the object), so I think this makes more sense

anthrotype:
Maybe it's because of this change in python 3.11:

https://docs.python.org/3/library/pickle.html#object.__getstate__

> Changed in version 3.11: Added the default implementation of the __getstate__() method in the [object](https://docs.python.org/3/library/functions.html#object) class.

Since BaseTable didn't define its own `__getstate__`, and since in python < 3.11 neither did the super class `object`, then the custom `BaseTable.__getattr__` method would be triggered whenever copy or pickle modules attempt to get `__getstate__` through `getattr()` built-in, and the custom `__getattr__` would automatically decompile our lazy objects.

Whereas with 3.11, since `object` parent class _does_ implement a default `__getstate__`, the `__getattr__` is not triggered and the lazy object remained lazy. But then each shallow copy were still sharing the same reader, which is single-use only, so decompiling one instance would destroy the other one.

anthrotype:
after merging #2970 into this PR branch that includes the fix, the tests that were passing on 3.10 but failing on 3.11 now pass on both

PR #2970, opened 2023-01-26
reproducer for bug subsetting COLR.ClipList with shared clips #2965

anthrotype:
modify test so that a couple color glyphs share the same clipbox to trigger issue copying lazy objects 

https://github.com/fonttools/fonttools/pull/2968


anthrotype:
passes on Python 3.10 but fails on Python 3.11 as expected:
https://github.com/fonttools/fonttools/actions/runs/4017863990/jobs/6902768414#step:5:3632

Issue #2959, opened 2023-01-18
[instancer] Should limit axes to actual fvar values


behdad:
It looks to me like if I tell instancer to change axes limits to what's beyond those of fvar, it actually succeeds and sets those in fvar.

Say, my font only has wght axis 300:700. If I ask instancer to "limit" wght axis to 100:1000, it writes 100,1000 as new min/max into fvar table. That sounds wrong to me.

PR #2971, opened 2023-01-26
[instancer] Limit axis ranges to that of fvar

behdad:
Fixes https://github.com/fonttools/fonttools/issues/2959

A bit crude but I think I got it right.

Should add a test.


anthrotype (approved):
LGTM (pending a small test .. and fixing the failing tests)

Issue #2150, opened 2021-01-17
Duplicate substitution rules are not optimized away


dscorbett:
I have an automatically generated feature file with two duplicate substitution rules. Here it is, with other rules removed for clarity.
```AFDKO
feature blws {
    script dupl;
    language dflt;
    sub u1BC08' @CONTEXT by u1BC08 u1BCA0;
    sub u1BC08' @CONTEXT by u1BC08 u1BCA0;
} blws;
```
When I add the file and build the font, fontTools generates a chaining contextual substitution lookup with one subtable for each rule. It does not deduplicate the rules. Each duplicate rule’s subtable refers to a different multiple substitution lookup, instead of reusing the same multiple substitution lookup. Is this something that fontTools could optimize?

khaledhosny:
Reusing lookups is problematic, since the order of the lookups can affect the final output.

dscorbett:
I don’t understand. Do you mean that
```AFDKO
feature blws {
    script dupl;
    language dflt;
    sub u1BC08' @CONTEXT by u1BC08 u1BCA0;
    sub u1BC08' @CONTEXT by u1BC08 u1BCA0;
} blws;
```
and
```AFDKO
feature blws {
    script dupl;
    language dflt;
    sub u1BC08' @CONTEXT by u1BC08 u1BCA0;
} blws;
```
can behave differently?

khaledhosny:
No, but if there are other lookups in the middle between the two, it could.

khaledhosny:
Anyway, I *think* the table compilers in the FontTools are smart and will deduplicate the data, even if the lookups are not deduplicated, so the file size difference is probably not that significant.

dscorbett:
Can you give an example where an extra substitution would make the presence or absence of the second duplicate line matter?

khaledhosny:
```fea
feature liga {
 script latn;
  sub a' b' by c;
  sub d  by b;
  sub a' b' by c;
} liga;
```
```
$ hb-shape test.ttf abad
[c=0+931|c=2+931]
```

```fea
lookup ab {
  sub a' b' by c;
} ab;

feature liga {
 script latn;
  lookup ab;
  sub d  by b;
  lookup ab;
} liga;
```

```
$ hb-shape test2.ttf abad
[c=0+931|a=2+1122|b=3+1069]
```

khaledhosny:
This behavior was intentional as #512 said this is makeotf behavior, but I think we don’t have to follow makeotf here.

[...]

PR #2974, opened 2023-01-27
[feaLib] Further merge inline single substitutions

khaledhosny:
Fixes https://github.com/fonttools/fonttools/issues/2150 (see the discussion there).


behdad:
That was quick! Thanks. LGTM.

PR #2983, opened 2023-02-06
Revert #2963

PoomSmart:
I didn't realize this line of code https://github.com/PoomSmart/fonttools/blob/main/Lib/fontTools/ttLib/tables/sbixGlyph.py#L131 already handles ref glyph type. The reference glyph ID is already assigned to `self.imageData`. Hence, no need for the workaround.

I still do no understand why it did not work and I had to use the workaround before...

### DIFF
--- a/Lib/fontTools/ttLib/tables/sbixGlyph.py
+++ b/Lib/fontTools/ttLib/tables/sbixGlyph.py
@@ -93,12 +93,7 @@ class Glyph(object):
         if self.graphicType is None:
             self.rawdata = b""
         else:
-            rawdata = sstruct.pack(sbixGlyphHeaderFormat, self)
-        if self.graphicType == "dupe":
-            rawdata += struct.pack(">H", ttFont.getGlyphID(self.referenceGlyphName))
-        else:
-            rawdata += self.imageData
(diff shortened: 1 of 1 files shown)


anthrotype:
that line you quoted handles "dupe" glyphs in fromXML (importing a ttx file into the table object), whereas the code you modified in #2963 handles them in the compile method (when serializing a TTFont to bytes), I think we probably need both.

PoomSmart:
@anthrotype Ah. It's true. I guess I can close this PR.

anthrotype:
actually, your earlier fix is doubly correct, because a glyph index may well change and when compiling we want to use the updated glyph ID. Thanks again

PoomSmart:
@anthrotype Oh. I think there's a problem with this code in `sbixStrike.py` (around line 80) that made my script failed when calling `ttFont.save()` in Python but not failed when I run `ttx -o TTF -x TTX`

image

It calls `compile()` only.

Whereas in the same file but `fromXML()` function:

image

It calls `fromXML()` before calling `compile().

`fromXML()` happens to handle `ref` to `self.imageData` as I said in OP.

PR #2988, opened 2023-02-07
[glyf] Support cubic curves

behdad:
https://github.com/harfbuzz/boring-expansion-spec/issues/41


behdad:
Can be used with this patch to `ufo2ft`:
```
diff --git a/Lib/ufo2ft/preProcessor.py b/Lib/ufo2ft/preProcessor.py
index 1543916a..64c8f0d1 100644
--- a/Lib/ufo2ft/preProcessor.py
+++ b/Lib/ufo2ft/preProcessor.py
@@ -330,13 +330,13 @@ class TTFInterpolatablePreProcessor:
             for func in funcs:
                 func(ufo, glyphSet)
 
-        fonts_to_quadratic(
-            self.glyphSets,
-            max_err=self._conversionErrors,
-            reverse_direction=self._reverseDirection,
-            dump_stats=True,
-            remember_curve_type=self._rememberCurveType and self.inplace,
-        )
+        #fonts_to_quadratic(
+        #    self.glyphSets,
+        #    max_err=self._conversionErrors,
+        #    reverse_direction=self._reverseDirection,
+        #    dump_stats=True,
+        #    remember_curve_type=self._rememberCurveType and self.inplace,
+        #)
 
         # TrueType fonts cannot mix contours and components, so pick out all glyphs
         # that have contours (`bool(len(g)) == True`) and decompose their
```
tested with the `-o=variable` output.

With NotoSansArabic-VF, I see 10% size improvement.

behdad:
See https://github.com/fonttools/fonttools/issues/2989
Currently I'm passing all offcurve point sequences to `curveTo`, which will interpret them as a "super-Bezier". Not what we want.

behdad:
HarfBuzz counterpart: https://github.com/harfbuzz/harfbuzz/pull/4107

behdad:
I've started a spec at:
https://github.com/harfbuzz/boring-expansion-spec/blob/main/glyf1-cubicOutlines.md

The full implementation handling all corner-cases needs more work (and is quite cumbersome). But I like feedback and whether this can move forward.

I like to land the HarfBuzz patch and ship it in HB 7 soon, even if it only handles simple cases.

Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -195,7 +195,13 @@ def moveTo(self, pt: Tuple[float, float]) -> None:
         self._addPoint(pt, 1)
 
     def curveTo(self, *points) -> None:
-        raise NotImplementedError
+        assert len(points) % 2 == 1
+        for pt in points[:-1]:
+            self._addPoint(pt, 0x80)
```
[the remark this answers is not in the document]
behdad: Yes. Will do.

Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -266,9 +272,23 @@ def endPath(self) -> None:
...
             raise PenError("Tried to end an empty contour.")
+
+        contourStart = self.endPts[-1] + 1 if self.endPts else 0
         self.endPts.append(len(self.points) - 1)
         self._currentContourStartIndex = None
 
+        # Resolve types for any cubic segments
+        flags = self.types
+        for i in range(contourStart, len(flags)):
+            if flags[i] == 3:
```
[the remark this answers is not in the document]
behdad: Will name it. It's a special number to mean we need to post-process to change the previous off-curves into a cubic off-curve.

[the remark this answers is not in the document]
behdad: Done. PTAL.

behdad:
This now implements all-offcurve outlines and cubic splines. I'm working on woff2 untransformed outlines, and I'll see if I can get cu2qu all_quadratic=False mode.

behdad:
> I'll see if I can get cu2qu all_quadratic=False mode.

Perhaps that should be done as a followup PR.

behdad:
> I'll see if I can get cu2qu all_quadratic=False mode.

This is done in `cu2qu` now. Looking into the pens.

behdad:
> > I'll see if I can get cu2qu all_quadratic=False mode.
> 
> This is done in `cu2qu` now. Looking into the pens.

All done and tested.

behdad:
I think this is ready now. I now have to go finish my HarfBuzz implementation.

anthrotype:
@behdad shouldn't the `head.glyphDataFormat` automatically be set to 1 when cubic beziers (or variable components) are present?

anthrotype:
we also need tests for ttGlyphPen's preserveTopology=False (dropping of implied oncurves), I can do that after lunch.

behdad:
> @behdad shouldn't the `head.glyphDataFormat` automatically be set to 1 upon compiling head (or glyf?) table whenever cubic beziers (or variable components) are present?

I don't think we should automatically do that. It would require expanding all glyphs that is very expensive. The subsetter for example goes to great lengths to avoid that.

anthrotype:
ok so it would be the responsibility of a higher-level compiler (ufo2ft, or fontTools.fontBuilder) to set that head.glyphDataFormat accordingly when building the font. At least our qu2cu command line tool could set that if/when it introduces any cubics in glyf?

Tests/pens/ttGlyphPen_test.py
```diff
@@ -611,7 +613,7 @@ def test_cubic(self):
 
         for pen in (spen, ppen):
 
-            glyph = pen.glyph()
+            glyph = pen.glyph(preserveTopology=False)
```
anthrotype, line 616: this seems no-op in this particular case. I would rather have distinct tests for this parameter with both True/False and expected outputs of each.

behdad: I just added a new test. Will modify it to test True/False.

Tests/pens/ttGlyphPen_test.py
```diff
@@ -638,6 +638,96 @@ def test_cubic(self):
                     ("closePath", ()),
                 ]
 
+    def test_cubic_topology(self):
+        for preserveTopology in (False, True):
```
anthrotype, line 642: this looks like could be parametrized with `pytest.mark.parametrize` decorator

anthrotype: parametrizing both the input and the expected output

behdad: While it's possible, it would have 8 lengthy parameters. I'm not sure it's a good idea. WDUT?

behdad: I'm trying it.

anthrotype: only 3 parameters, preserve_topology, expected_coordinates, expected_flags

behdad: Done.

anthrotype: thanks! that's much better

anthrotype:
@behdad can you add to TTGlyphPen docstring to explain what `preserveTopology` parameter does? Maybe we should remame it to `dropImpliedOnCurves` or something that is clearer on its own?

anthrotype:
btw, re: head.glyphDataFormat, does harfbuzz even look at that when parsing glyf?

behdad:
> At least our qu2cu command line tool could set that if/when it introduces any cubics in glyf?

Done.

behdad:
> btw, re: head.glyphDataFormat, does harfbuzz even look at that when parsing glyf?

No. I take it to be intentional.

Lib/fontTools/qu2cu/cli.py
```diff
@@ -31,6 +32,12 @@ def _font_to_cubic(input_path, output_path=None, **kwargs):
         glyph.draw(pen)
         glyf[glyphName] = ttpen.glyph(preserveTopology=False)
 
+    if stats["4"]:
```
anthrotype, line 35: that looks weird.. what's '4'?

behdad: 4 means "cubic", 3 means "quadratic".

behdad: I'll add a comment.

behdad:
> > @behdad can you add to TTGlyphPen docstring to explain what `preserveTopology` parameter does? Maybe we should remame it to `dropImpliedOnCurves` or something that is clearer on its own?
> 
> Okay.

Done.

Lib/fontTools/qu2cu/cli.py
anthrotype: because in cu2qu.ufo and also in cu2quPen the stats dict keys are number of off-curves, I think, not number of total control points including start and end points... See how they do len(curve) - 2

anthrotype: whereas your Qu2CuPen seems inconsistent in that regard

behdad: I thought it's just at printing time that we do the -2. Let me find that code.

behdad: You are correct. I'll fix.

anthrotype:
how much work would adding support for all_quadratic=False be for cu2qu.ufo.fonts_to_quadratic as well? we can also file an issue and do later if/when needed

Lib/fontTools/qu2cu/cli.py
behdad: Ah. The pen is correct. I should use 2 instead of 4 here.

anthrotype: > Ah. The pen is correct. I should use 2 instead of 4 here.

are you sure the pen is correct? It's not doing - 2

https://github.com/fonttools/fonttools/blob/085b489012b45e55ff5eac25bf80c7781311c8ba/Lib/fontTools/pens/qu2cuPen.py#L52-L56

behdad: No. Let me clean this up.

behdad:
> how much work would adding support for all_quadratic=False be for cu2qu.ufo.fonts_to_quadratic as well? we can also file an issue and do later if/when needed

Let me take a look. Isn't too much work.

behdad:
> > how much work would adding support for all_quadratic=False be for cu2qu.ufo.fonts_to_quadratic as well? we can also file an issue and do later if/when needed
> 
> Let me take a look. Isn't too much work.

Now I remember. I took a look but didn't know UFO enough to know what to do. What should I put in `CURVE_TYPE_LIB_KEY`?

behdad:
Ah I see it's a cu2qu private key. We can still set it to quadratic I guess.

Lib/fontTools/qu2cu/cli.py
anthrotype: so to decide whether you end up with any cubics, basically, you're checking that the stats contain an entry for the count of curves/splines with 2 offcurves. However, is it conceivable that the Qu2CuPen keeps all the quad splines that have 2 off-curves as quadratic and doesn't actually convert them to cubics? In which case, you are setting the glyphDataFormat to 1 unnecessarily?

behdad:
@anthrotype you are right about me setting the head table version wrong. I was confused with the cu2qu all_quadratic=False mode. So, I suggest upgrading the table version unconditionally. WDYT?

anthrotype:
> What should I put in CURVE_TYPE_LIB_KEY?

well, technically, in the case of all_quadratic=False the value is neither 'cubic' (defaul when omitted) nor 'quadratic'...
How about "mixed"?

anthrotype:
> upgrading the table version unconditionally

for the qu2cu command line tool, maybe that's enough. But surely we could come up with a better way to detect this than checking the stats dict..

anthrotype:
the CURVE_TYPE_LIB_KEY is mostly about caching the state inside the ufo itself and avoiding re-running cu2qu on an already converted ufo.

anthrotype:
I also think the qu2cu.cli should have a flag to set Qu2CuPen all_cubic=True if one wants to force the output to be only cubic (could be useful for testing purposes)

behdad:
> I also think the qu2cu.cli should have a flag to set Qu2CuPen all_cubic=True if one wants to force the output to be only cubic (could be useful for testing purposes)

I'll add.

cu2qu.cli can also gain an inverse of all_quadratic when I add the support to ufo.py...

behdad:
> how much work would adding support for all_quadratic=False be for cu2qu.ufo.fonts_to_quadratic as well? we can also file an issue and do later if/when needed

Done. Can you look into testing it?

anthrotype:
both curves_to_quadratic (with all_quadratic=False) and quadratic_to_curves (with all_cubic=False) return list of tuples of either length three (for quads) or lenght 4 (for cubics).. so the caller (be it a pen or the cu2qu.ufo module which also would like to know if it did actually modify a given curve segment) should be able use the lengths of the returned tuples to determine whether or not the function has actually done anything to the input

anthrotype:
>  Can you look into testing it?

yes

Lib/fontTools/qu2cu/cli.py
```diff
@@ -59,6 +60,14 @@ def main(args=None):
         default=0.001,
         help="maxiumum approximation error measured in EM (default: 0.001)",
     )
+    parser.add_argument(
+        "-c",
+        "--all-cubic",
+        type=bool,
+        default=False,
+        action=argparse.BooleanOptionalAction,
```
anthrotype, line 68: i think that's 3.9+ only, you should probably use `action="store_true"`

behdad:
> cu2qu.cli can also gain an inverse of all_quadratic when I add the support to ufo.py...

What should I call that argument? `--mixed`? Umm.

anthrotype:
--mixed sounds good

behdad:
Nevermind. They are somehow stuck on old code. No idea why. I already fixed the code.

Lib/fontTools/cu2qu/ufo.py
```diff
@@ -143,22 +143,25 @@ def _set_segments(glyph, segments, reverse_direction):
...
     assert all(len(s) == n for s in new_points[1:]), "Converted incompatibly"
 
     spline_length = str(n - 2)
     stats[spline_length] = stats.get(spline_length, 0) + 1
 
-    return [("qcurve", p) for p in new_points]
+    if all_quadratic:
+        return [("qcurve", p) for p in new_points]
+    else:
+        return [("qcurve" if len(p) == 3 else "curve", p) for p in new_points]
```
anthrotype, line 161: I'm not 100% sure this line is correct. The `_segments_to_quadratic` is called when zipping through on each curves at ith position along the interpolatable contours. I expect its return value to be all qcurve or all curve, not a mix of the two, otherwise it'd break interpolability.
Maybe that's impossible because `curves_to_quadratic` already does the right thing, in which case this line could be further simplified because we already know `n = len(new_points[0])` a few lines above, if that's == 3 and all_quadratic=False, then we know (or assert) that the segment type is qcurve, otherwise curve... Does that make sense?

behdad: Yes, I believe it can be simplified as you said.

behdad:
I think this is good to merge now.

behdad:
Next step would be ufo2ft / fontmake support...

anthrotype:
> ufo2ft / fontmake support...

basically we just need a way to pass down all_quadratic=False to cu2qu fonts_to_quadratic (to get the mixed quad+cubic curves)

behdad:
> > ufo2ft / fontmake support...
> 
> basically we just need a way to pass down all_quadratic=False to cu2qu fonts_to_quadratic (to get the mixed quad+cubic curves)

Correct..

PR #2993, opened 2023-02-13
qu2cu

behdad:


behdad:
Based on https://github.com/googlefonts/cu2qu/pull/29/files#diff-0ad9d3d850d1321d2e2abe961cabbef118134bb5c155ee65103968c8c2fe16acR274-R286

Lib/fontTools/qu2cu/qu2cu.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+@cython.locals(_1_3=cython.double, _2_3=cython.double)
+@cython.locals(
+    p0=cython.complex,
+    p1=cython.complex,
+    p2=cython.complex,
+    p1_2_3=cython.complex,
+)
+def elevate_quadratic(p0, p1, p2, _1_3=1 / 3, _2_3=2 / 3):
```
rsheeter, line 47: Suggest adding a citation so those new to curves can learn something, e.g. https://pomax.github.io/bezierinfo/#reordering

rsheeter:
I was unable to locate the tests. Now we have cu2qu and qu2cu you could round trip things, perhaps cherry-picking cases where that works out with nice #s.

behdad:
> I was unable to locate the tests. Now we have cu2qu and qu2cu you could round trip things, perhaps cherry-picking cases where that works out with nice #s.

Yeah right now there's just a main that roundtrips. Will need tests indeed.

behdad:
The algorithm is slow (`O(n^3)`). But for short segments that `cu2qu` generates it should not matter.

behdad:
Later we can add `quadratics_to_curve` that takes multiple (joining) quadratic splines and converts them.

behdad:
I think this is ready now. With tests.

behdad:
I've rebased it on cubic-glyf branch, but that's only required for the cli, which is not tested. So can be rebased on main and merged.

behdad:
> can be rebased on main

Done.

behdad:
Okay I'm finally happy with this.

Lib/fontTools/qu2cu/qu2cu.py
```diff
@@ -0,0 +1,327 @@
...
+
+
+def quadratics_to_curves(pp, tolerance=0.5, all_cubic=False):
+    is_complex = type(pp[0][0]) is complex
+    if not is_complex:
+        pp = [[complex(x, y) for (x, y) in p] for p in pp]
+
+    q = [pp[0][0]]
+    cost = 0
+    costs = [0]
```
anthrotype, line 173: it's unclear what these `costs` are referring to or how they are used, can you please add a comment at least to help the reader?

anthrotype (approved):
LGTM, thank you very much for pushing this to the finishing line! 👏

PR #2995, opened 2023-02-17
reverseContourPen_test: check outputImpliedClosingLine works as expected

anthrotype:
In https://github.com/fonttools/fonttools/pull/2922, @behdad added an `outputImpliedClosingLine` parameter to the `ReverseContourPen`, plus a test that only expected a certain number of segments. 

I was looking at this again today, and got a little suspicious as I noticed the number of pen commands increases from _4_ in the default `outputImpliedClosingLine=False` case, to _6_ in the `outputImpliedClosingLine=True` case, whereas I would only expect it to increase only by +1, i.e. to _5_, in the latter case, given it no longer omits the closePath's closing line.

It turns out, with `outputImpliedClosingLine=True`, not only the implied closing line is emitted, but also an additional, duplicate lineTo right after moveTo is emitted, which I belive is incorrect.

Here I first push the failing test that exemplifies the issue, without changing the code. I'll follow up with the fix.

### DIFF
--- a/Lib/fontTools/pens/reverseContourPen.py
+++ b/Lib/fontTools/pens/reverseContourPen.py
@@ -62,7 +62,7 @@ def reversedContour(contour, outputImpliedClosingLine=False):
         if closed:
             # for closed paths, we keep the starting point
             yield firstType, firstPts
-            if outputImpliedClosingLine or firstOnCurve != lastOnCurve:
+            if firstOnCurve != lastOnCurve:
                 # emit an implied line between the last and first points
                 yield "lineTo", (lastOnCurve,)
                 contour[-1] = (lastType, tuple(lastPts[:-1]) + (firstOnCurve,))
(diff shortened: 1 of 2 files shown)


anthrotype:
as expected, CI test fails https://github.com/fonttools/fonttools/actions/runs/4204996225/jobs/7296452293

```
E       AssertionError: assert [('moveTo', (...osePath', ())] == [('moveTo', (...osePath', ())]
E         At index 1 diff: ('lineTo', ((0, 0),)) != ('lineTo', ((0, 10),))
E         Left contains one more item: ('closePath', ())
E         Use -v to get more diff
```

If I pass -v verbose to pytest, we get the full diff. Notice the additional `lineTo((0, 0))` after the `moveTo((0, 0))`:

```
E         Full diff:
E           [
E            ('moveTo',
E         +   ((0,
E         +     0),)),
E         +  ('lineTo',
E             ((0,
E               0),)),
E            ('lineTo',
E             ((0,
E               10),)),
E            ('lineTo',
E             ((10,
E               0),)),
E            ('lineTo',
E             ((0,
E               0),)),
E            ('closePath',
E             ()),
E           ]
```

I am _not_ referring to the final `lineTo((0, 0))` before `closePath`, that is correct and as expected because `outputImpliedClosingLine=True`.

anthrotype:
https://github.com/fonttools/fonttools/pull/2995/commits/847ce2866e2f0f5366c8404f3514403de246127e fixes it, this is now ready for review

behdad (approved):

Issue #902, opened 2017-03-29
maxp.maxSizeOfInstructions is not calculated?


khaledhosny:
I was trying to make a font with defective `glyf` table glyphs for testing OTS, when I found that `Tests/ttx/data/TestTTF.ttf` is already rejected by OTS. This seems to be because `maxSizeOfInstructions` is set to 0, but some glyphs have `instructionLength` of 2.

It seems `maxSizeOfInstructions` is not calculated when generating `maxp` table, I’m not sure if this is intentional or not.

anthrotype:
Yes, we currently don't recalculate that. We probably should.

The problem is that there is some ambiguity as to how properly calculate that value.
I myself raised the issue with the OpenType mailing list almost two years ago, and Greg Hitchcock from Microsoft replied with a long and detailed history of this maxp field.

TLDR...

This should be based on the largest set of instructions (in the glyf table) for a single simple or composite glyph.

Tools like VTT or ttfautohint also consider `fpgm` and `prep` when computing this value, but apparently are wrong.

FontValidator also warns when the value does not match the "Maximum byte count for glyph instructions" (as the `maxp` spec puts it).

lemzwerg:
OK, just rereading Greg's e-mail...

The sentence 'Tools like VTT or ttfautohint [...] are apparently wrong' is highly debatable :-) Given that no implementation is actually using the max value, and that the (former) de-facto tool for creating bytecode (VTT)  disagrees or disagreed for a long time with Greg's interpretation, and that the wording in the specification is still ambiguous in the newest OpenType specification, we have to live with the dichotomy of two (or even more) valid values.  Using the maximum value of the bytecode sizes in the `fpgm`, `prep`, and `glyf` does certainly work.  Admittedly, it might be too large, but it doesn't do any harm.

behdad:
> I was trying to make a font with defective glyf table glyphs for testing OTS, when I found that Tests/ttx/data/TestTTF.ttf is already rejected by OTS. This seems to be because maxSizeOfInstructions is set to 0, but some glyphs have instructionLength of 2.

Relax ots I'd say.

> It seems maxSizeOfInstructions is not calculated when generating maxp table, I’m not sure if this is intentional or not.

Maybe sometimes it should.  I don't want it to happen in subetter though.  Maybe same as other recalc stuff?  Also, it will force expansion of all glyf entries which is costly.  We can implement a fastpath that doesn't expand.  I did that for some other stuff in the subsetter.

[...]

PR #3002, opened 2023-02-21
Calculate maxSizeOfInstructions

madig:
Closes https://github.com/fonttools/fonttools/issues/902

I chose to implement the simplest strategy according to the spec, because why not.

Waiting for feedback before adjusting the tests.


madig:
Cosimo says: 

> it's ambiguous how to compute as there are several competing interpretatinos and it's useless, so fonttools should not bother

anthrotype:
well, I take back the "useless".. I'm open to reconsider if having this field incorrectly set (whatever the defintion of 'correct' is) produces actual undersirable effects. E.g. does OTS still fails? 
If it were me, I would actually go for the more relaxed calculation that also includes the global fpgm and prep, as a larger maxSizeOfInstructions should in theory cause less harm than smaller. But first I would still like to have evidences of any such harm caused.

Issue #3003, opened 2023-02-23
varLib is creating multiple, disconnected feature variations groups from a single rule with a single condition


arrowtype:
I have a variable font in which I am trying to do the typical GSUB of a simplified Q and Ø at heavy weights, when the inner counter can no longer support a bar.

I am compiling this from UFOs and a (v5) designspace, using FontMake version `3.5.1` and fonttools version `4.38.0`.

I have a `rule` that looks like this:

```xml
    <rule name="Open Rounds">
      <conditionset>
        <condition name="Weight" minimum="728" maximum="1000"/>
        <condition name="Optical Size" minimum="8" maximum="72"/>
      </conditionset>
      <sub name="Q" with="Q.open"/>
      <sub name="Oslash" with="Oslash.open"/>
      <sub name="oslash" with="oslash.open"/>
    </rule>
```

(I have also tried the above without any `Optical Size` condition, but the docs seem ambiguous about whether it is suggested to include axes if a rule should happen throughout the whole thing)

I don’t have any other rules that reference the `Q` or `Q.open`. However, this is the result I’m getting:

https://user-images.githubusercontent.com/45946693/221049580-a5bfd7c8-1b10-446e-bd16-0fa527676b7e.mov

Interestingly, this "flickering" doesn’t happen on the max end of `opsz`.

If I use TTX to see the GSUB table, I can see that the substitution lookup is referenced in multiple different `FeatureVariationRecord`s, which is somewhat surprising to me, but I don’t know enough about those to know whether that is significant.

My axes look like this, if it is useful:

```xml
  <axes elidedfallbackname="Regular">
    <axis tag="opsz" name="Optical Size" minimum="8" maximum="72" default="8">
      <labelname xml:lang="en">Optical Size</labelname>
      <map input="72" output="72"/>
      <map input="20" output="40"/>
      <map input="8" output="8"/>
    </axis>
    <axis tag="wght" name="Weight" minimum="1" maximum="1000" default="700">
      <labelname xml:lang="en">Weight</labelname>
      <map input="1000" output="1000"/>
      <map input="900" output="850"/>
      <map input="800" output="775"/>
      <map input="700" output="700"/>
      <map input="600" output="600"/>
      <map input="500" output="500"/>
      <map input="400" output="400"/>
      <map input="300" output="300"/>
      <map input="200" output="200"/>
      <map input="100" output="100"/>
      <map input="1" output="1"/>
      <labels>
        <label uservalue="1" name="Hairline"/>
[...]

PR #3005, opened 2023-02-24
[varLib.featureVars] Fix bug in overlayBox

behdad:
See added test. Previously we were cutting remainder box even if the intersection had dimensions not present in the remainder box. Ouch!

Fixes https://github.com/fonttools/fonttools/issues/3003

Tests need updating.


behdad:
Tests happy now.

Issue #3006, opened 2023-02-24
Error on sorting NameRecords, comparison not implemented?


kontur:
Compiling a font with the Fontmake toolchain and with Fonttools 4.38.0 I get:

```
  File "/Users/johannes/.pyenv/versions/3.10.4/envs/mrhandy/lib/python3.10/site-packages/fontTools/ttLib/tables/_n_a_m_e.py", line 62, in compile
    names.sort() # sort according to the spec; see NameRecord.__lt__()
TypeError: '<' not supported between instances of 'NameRecord' and 'NameRecord'
```
in https://github.com/fonttools/fonttools/blob/111d2c5c0f3de8804843ad5dfdf4b18c8b107966/Lib/fontTools/ttLib/tables/_n_a_m_e.py#L62

Checking the [NameRecord.__lt__()](https://github.com/fonttools/fonttools/blob/111d2c5c0f3de8804843ad5dfdf4b18c8b107966/Lib/fontTools/ttLib/tables/_n_a_m_e.py#L581) method it seems something is obscuring the actual error in this blanket try block, and as per the comments, it explains:

```
			# This can only happen for
			# 1) an object that is not a NameRecord, or
			# 2) an unlikely incomplete NameRecord object which has not been
			#    fully populated, or
			# 3) when all IDs are identical but the strings can't be encoded
			#    for their platform encoding.
			# In all cases it is best to return NotImplemented.
			return NotImplemented
```

The original source of the error for my case was `'charmap' codec can't encode character '\u0132' in position 35: character maps to <undefined> SELF Alternative form of ‘J’ and Dutch ‘Ĳ’` for a stylistic set name that uses the encoded `Ĳ` (not `I` `J`). This probably falls under 3) but not obscuring this error with a generic `NotImplement` would be heeeeeeaps more helpful. In fact, emitting a warning and passing might actually be a viable option, too. I reckon accidental encoding mishaps or impossibilities might happen rather frequently.

This should be reproducible with a font with this in it's features:

```
feature ss01 {
	featureNames {
		name "Alternative form of ‘J’ and Dutch ‘Ĳ’";
                name 1 "Alternative form of ‘J’ and Dutch ‘Ĳ’";
	};
        #...
} ss01;
```

It's the second line with the platform ID (and omitted script ID which falls back to Latin) that is causing the issue.

anthrotype:
[...]

PR #3007, opened 2023-02-24
Make NameRecord comparison not fail on encoding errors #3006

kontur:
- Distinguish raised errors in comparing / sorting of `NameRecord`s to not fail sorting on encoding errors of the NameRecord, but retain behavior of failing on attribute errors or incomplete NameRecords
- Emit warning when encountering NameRecord encoding issues
- Added test for sorting incomplete NameRecords (the code comments in the comparison function mention such cases, but I don't think there was an actual test for it)
- Modified test for sorting NameRecords with encoding errors (previously failing test should now a) not fail, b) perform sorting based on available IDs)

I tested with `pytest` only; got `tox` to run some tests (lints okay), but coverage tests I could not manage to get running:

```
ERROR: Could not find a version that satisfies the requirement lxml==4.9.0 (from versions: none)
ERROR: No matching distribution found for lxml==4.9.0

py310-cov: exit 1 (0.59 seconds) /Users/johannes/Projects/fonttools> python -m pip install --only-binary=lxml 'coverage>=4.3' lxml==4.9.0 pytest pytest-randomly -r requirements.txt pid=73604
```

### DIFF
--- a/Lib/fontTools/ttLib/tables/_n_a_m_e.py
+++ b/Lib/fontTools/ttLib/tables/_n_a_m_e.py
@@ -607,34 +607,40 @@ class NameRecord(object):
 
     def __lt__(self, other):
         if type(self) != type(other):
-            return NotImplemented
+            return NotImplemented()
 
         try:
-            # implemented so that list.sort() sorts according to the spec.
             selfTuple = (
                 self.platformID,
                 self.platEncID,
                 self.langID,
                 self.nameID,
-                self.toBytes(),
             )
             otherTuple = (
                 other.platformID,
                 other.platEncID,
                 other.langID,
                 other.nameID,
-                other.toBytes(),
             )
-            return selfTuple < otherTuple
-        except (UnicodeEncodeError, AttributeError):
+        except (AttributeError):
             # This can only happen for
             # 1) an object that is not a NameRecord, or
             # 2) an unlikely incomplete NameRecord object which has not been
-            #    fully populated, or
-            # 3) when all IDs are identical but the strings can't be encoded
-            #    for their platform encoding.
-            # In all cases it is best to return NotImplemented.
-            return NotImplemented
+            #    fully populated
+            return NotImplemented()
+
+        try:
+            # Include the actual NameRecord string in the comparison tuples
+            selfTuple = selfTuple + (self.toBytes(),)
+            otherTuple = otherTuple + (other.toBytes(),)
+        except (UnicodeEncodeError) as e:
+            # toBytes caused an encoding error in either of the two, so content
+            # to sorting based on IDs only
+            log.warning("NameRecord sorting failed to decode: %s" % e)
+
+        # Implemented so that list.sort() sorts according to the spec by using
+        # the order of the tuple items and their comparison
+        return selfTuple < otherTuple
 
     def __repr__(self):
         return "<NameRecord NameID=%d; PlatformID=%d; LanguageID=%d>" % (
--- a/Tests/ttLib/tables/_n_a_m_e_test.py
+++ b/Tests/ttLib/tables/_n_a_m_e_test.py
@@ -72,14 +72,33 @@ class NameTableTest(unittest.TestCase):
         ]
         table.compile(None)
 
-    def test_names_sort_bytes_str_encoding_error(self):
+    def test_names_sort_attributes(self):
         table = table__n_a_m_e()
+        # Create an actual invalid NameRecord object
+        broken = makeName("Test", 25, 3, 1, 0x409)
+        delattr(broken, "platformID")
         table.names = [
-            makeName("Test寬", 25, 1, 0, 0),
-            makeName("Test鬆鬆", 25, 1, 0, 0),
+            makeName("Test", 25, 3, 1, 0x409),
+            broken,
         ]
+        # Sorting these two is impossible, expect an error to be raised
         with self.assertRaises(TypeError):
             table.names.sort()
+
+    def test_names_sort_encoding(self):
+        """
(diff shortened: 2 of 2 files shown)


Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -607,34 +607,40 @@ def fromXML(self, name, attrs, content, ttFont):
 
     def __lt__(self, other):
         if type(self) != type(other):
-            return NotImplemented
+            return NotImplemented()
```
anthrotype, line 610: NotImplemented is a singleton value, not an exception class to be instantiated (not to be confused with NotImplementedError) so remove the `()`

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -607,34 +607,40 @@ def fromXML(self, name, attrs, content, ttFont):
...
             otherTuple = (
                 other.platformID,
                 other.platEncID,
                 other.langID,
                 other.nameID,
-                other.toBytes(),
             )
-            return selfTuple < otherTuple
-        except (UnicodeEncodeError, AttributeError):
+        except (AttributeError):
```
anthrotype, line 625: remove the no-op parentheses here

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -607,34 +607,40 @@ def fromXML(self, name, attrs, content, ttFont):
...
             # This can only happen for
             # 1) an object that is not a NameRecord, or
             # 2) an unlikely incomplete NameRecord object which has not been
-            #    fully populated, or
-            # 3) when all IDs are identical but the strings can't be encoded
-            #    for their platform encoding.
-            # In all cases it is best to return NotImplemented.
-            return NotImplemented
+            #    fully populated
+            return NotImplemented()
```
anthrotype, line 630: remove parentheses

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -607,34 +607,40 @@ def fromXML(self, name, attrs, content, ttFont):
...
+
+        try:
+            # Include the actual NameRecord string in the comparison tuples
+            selfTuple = selfTuple + (self.toBytes(),)
+            otherTuple = otherTuple + (other.toBytes(),)
+        except (UnicodeEncodeError) as e:
```
anthrotype, line 636: parentheses...

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -607,34 +607,40 @@ def fromXML(self, name, attrs, content, ttFont):
...
             # 2) an unlikely incomplete NameRecord object which has not been
-            #    fully populated, or
-            # 3) when all IDs are identical but the strings can't be encoded
-            #    for their platform encoding.
-            # In all cases it is best to return NotImplemented.
-            return NotImplemented
+            #    fully populated
+            return NotImplemented()
+
+        try:
```
anthrotype, line 632: instead of a new try: except block you can simply add another `except UnicodeEncodeError:` branch

anthrotype: one can have as many `except SomeException:` branches as needed all in the same `try:`

anthrotype: sorry I see you did intentionally as you no longer try to encode in the first try block.

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -607,34 +607,40 @@ def fromXML(self, name, attrs, content, ttFont):
...
+            # Include the actual NameRecord string in the comparison tuples
+            selfTuple = selfTuple + (self.toBytes(),)
+            otherTuple = otherTuple + (other.toBytes(),)
+        except (UnicodeEncodeError) as e:
+            # toBytes caused an encoding error in either of the two, so content
+            # to sorting based on IDs only
+            log.warning("NameRecord sorting failed to decode: %s" % e)
```
anthrotype, line 639:
```suggestion
            log.warning("NameRecord sorting failed to encode: %s" % e)
```

anthrotype: I would make that a `log.error`

Lib/fontTools/ttLib/tables/_n_a_m_e.py
kontur: Yikes, think autocomplete done a goofy one on me here.

Lib/fontTools/ttLib/tables/_n_a_m_e.py
kontur: yep, also tried some `try…else…finally` approach first, but it just ended up more convoluted and hard to read.

Lib/fontTools/ttLib/tables/_n_a_m_e.py
kontur: Okay by me, after all it _is_ an actual error that causes the message.

Lib/fontTools/ttLib/tables/_n_a_m_e.py
kontur: Hm, yes. I remember having to read up on returning this instead of raising a NotImplementedError, confused the two syntaxes/types.

kontur:
All updated as per your comments, thanks for reviewing :)

PR #3012, opened 2023-03-01
[subset] Compute xAvgCharWidth even when --no-prune-unicode-ranges is given

belluzj:
The conditions in the subsetter were so that re-computing the xAvgCharWidth would only happen when also pruning Unicode ranges, while the two are not connected. This makes the two options independent from one another.


anthrotype:
thank you

PR #3017, opened 2023-03-02
prevent cython.compiled raise AttributeError if cython not properly installed

anthrotype:
It's possible sometimes that `import cython` does _not_ fail but then attempting to get `cython.compiled` raises `AttributeError` exception. It actually happened in our internal production environment... (still unclear as to the root cause).
So here in this PR, to check that the cython library is correctly installed I also catch AttributeError raised when the `compiled` variable is not there (even though it _must_ be there for the 'real' cython, but sometimes a 'pretend' one sneaks in, probably because sys.path has been messed with).

somewhat similar issue to https://github.com/pydantic/pydantic/pull/573 and https://github.com/ipython/ipython/issues/13294

### DIFF
--- a/Lib/fontTools/cu2qu/cu2qu.py
+++ b/Lib/fontTools/cu2qu/cu2qu.py
@@ -17,9 +17,13 @@
 
 try:
     import cython
-except ImportError:
+
+    COMPILED = cython.compiled
+except (AttributeError, ImportError):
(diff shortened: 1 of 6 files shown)


anthrotype:
hold on, test-cython CI failed..

anthrotype:
CI failures unrelated (codecov token gain aargh...)

Issue #2659, opened 2022-06-21
varLib CLI can't build multiple variable fonts from Designspace v5 


anthrotype:
The https://github.com/fonttools/fonttools/pull/2436 PR that defined DesignSpace v5, also added support to varLib to build several variable fonts from a single designspace v5 document (via `varLib.build_many` method).
However the `main()` script used when calling `fonttools varLib` from the command line is still using the old, single-VF `varLib.build()` method that only builds one VF at a time.
It should use the new `build_many` instead, and add a CLI flag to filter/select which VF to build. And also maybe an `--output-dir` to choose where to save multiple outputs.

anthrotype:
(@belluzj feel free to un-assign yourself in case you're not available at this time, thanks)

PR #3024, opened 2023-03-07
[varLib] Don't overwrite pre-existing STAT

anthrotype:
when master UFOs contain STAT table definition inside features.fea, varLib should not overwrite that with its own (potentially incomplete) STAT but keep whatever is there (587b0fbf115dea2302e86c81aacbf245621ea32d).

This fixes https://github.com/googlefonts/fontmake/issues/985 and https://github.com/googlefonts/ufo2ft/issues/723

/cc @kontur 

Also while I am at, I changed fonttools varLib CLI to call the new `build_many` API which supports DesignSpace v5 with multiple variable fonts within the same doc (#2659).

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -995,11 +995,11 @@ def build_many(
             vf = build(
                 vfDoc,
                 master_finder,
-                exclude=list(exclude) + ["STAT"],
+                exclude=exclude,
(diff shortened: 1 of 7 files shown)


kontur:
Thanks for the quick action on this one 👍

anthrotype:
@kontur I hope you don't mind me reusing your test files from https://github.com/googlefonts/fontmake/issues/985

PR #3027, opened 2023-03-08
[COLRv1] Add method to automatically compute ClipBoxes, w/ optional quantization

anthrotype:
ClipBoxes are optional in COLRv1 but can make rendering more efficient. Currently colorLib.builder.buildCOLR takes an optional clipBoxes parameter with user-defined clipBoxes (nanoemoji computes those upfront and provides them to colorLib). Other clients like glyphsLib don't compute the clipBoxes for us, so COLRv1 fonts built from fontmake via glyphsLib have no clip boxes.
This adds a method so one can ask fonttools to compute them, taking into account various paint transforms.
For example one can do:

```python

font = TTFont("path/to/input/font.ttf")
font["COLR"].table.computeClipBoxes(font.getGlyphSet(), quantization=100)
font.save("path/to/output/font.ttf")
```

This currently only works for static, non variable COLRv1 tables, which is OK, since for variable COLR is probably preferable _not_ to provide explicit clipboxes and let renderer derive them at runtime, since computing a bounding box that works throughout the entire variation space is non trivial in presence of variable rotation/skew.

This can also be handy for https://github.com/googlefonts/glyphsLib/issues/847

### DIFF
--- a/Lib/fontTools/misc/arrayTools.py
+++ b/Lib/fontTools/misc/arrayTools.py
@@ -282,6 +282,27 @@ def intRect(rect):
     return (xMin, yMin, xMax, yMax)
 
 
+def quantizeRect(rect, factor=1):
+    """
+    >>> bounds = (72.3, -218.4, 1201.3, 919.1)
+    >>> quantizeRect(bounds)
+    (72, -219, 1202, 920)
+    >>> quantizeRect(bounds, factor=10)
+    (70, -220, 1210, 920)
+    >>> quantizeRect(bounds, factor=100)
+    (0, -300, 1300, 1000)
+    """
+    if factor < 1:
+        raise ValueError(f"Expected quantization factor >= 1, found: {factor!r}")
+    xMin, yMin, xMax, yMax = normRect(rect)
+    return (
+        int(math.floor(xMin / factor) * factor),
+        int(math.floor(yMin / factor) * factor),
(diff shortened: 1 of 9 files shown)


anthrotype:
(I plan to add tests tomorrow, but thanks behdad for taking a look)

behdad:
This looks okay except that it doesn't handle PaintColrGlyph?

anthrotype:
it does, because iterPaintSubTables traverses those as well, and if any PaintGlyphs are found within their sub-trees, their bounds will be computer as well

behdad:
FWIW we have a more complicated logic of this implemented in HarfBuzz, which is probably unnecessarily complex but allows for multiple clips inside each other... It also has the job of determining boundedness of the paint:

https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-paint-extents.hh
https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-paint-extents.cc

behdad:
> it does, because iterPaintSubTables traverses those as well, and if any PaintGlyphs are found within their sub-trees, their bounds will be computer as well

Ah okay.

anthrotype:
PaintComposite might need special treatment as well; right now we set clipbox to the union of the source/backdrop clipboxes; the actual one might actually be smaller depending on the composite mode, but at least it's never going to be larger than that, so I think what we have is sufficient.

PR #3029, opened 2023-03-09
[glifLib] Wrap XML library exceptions with glifLib types when parsing glifs

Hoolean:
Hello!

This PR abstracts across XML library exceptions with glifLib errors when parsing glyphs, which allows dependent projects to catch syntax errors without requiring logic to account for which XML library fonttools is using internally (e.g. for implementing fonttools/ufoLib2#264).

The PR also also adds tests to ensure that the exception we expose remains stable across future releases.

### Affected functions

* `glifLib.readGlyphFromString()`
* `glifLib.GlyphSet.readGlyph()`

### Alternative

If we wanted to abstract over lxml/xml exceptions more broadly, we could catch them in `etree.py` and wrap them with a fonttools-specific XmlParsingError; this would also have the cost of maintaining wrapper functions (e.g. for `etree.fromstring()`) where there currently are none.

### DIFF
--- a/Lib/fontTools/ufoLib/glifLib.py
+++ b/Lib/fontTools/ufoLib/glifLib.py
@@ -416,17 +416,33 @@ class GlyphSet(_UFOBaseIO):
         if validate is None:
             validate = self._validateRead
         text = self.getGLIF(glyphName)
-        tree = _glifTreeFromString(text)
-        formatVersions = GLIFFormatVersion.supported_versions(
-            self.ufoFormatVersionTuple
-        )
-        _readGlyphFromTree(
-            tree,
-            glyphObject,
-            pointPen,
-            formatVersions=formatVersions,
-            validate=validate,
-        )
+        try:
+            tree = _glifTreeFromString(text)
+            formatVersions = GLIFFormatVersion.supported_versions(
+                self.ufoFormatVersionTuple
+            )
+            _readGlyphFromTree(
+                tree,
+                glyphObject,
+                pointPen,
+                formatVersions=formatVersions,
+                validate=validate,
+            )
+        except GlifLibError as glifLibError:
+            # Re-raise with a note that gives extra context, describing where
+            # the error occurred.
+            fileName = self.contents[glyphName]
+            try:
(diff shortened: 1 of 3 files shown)


Lib/fontTools/ufoLib/glifLib.py
```diff
@@ -1082,10 +1082,14 @@ def _glifTreeFromFile(aFile):
...
-        root = etree.fromstring(data, parser=etree.XMLParser(remove_comments=True))
-    else:
-        root = etree.fromstring(data)
+    try:
+        if etree._have_lxml:
+            root = etree.fromstring(data, parser=etree.XMLParser(remove_comments=True))
+        else:
+            root = etree.fromstring(data)
+    except Exception as etree_exception:
+        raise GlifLibError("GLIF contains invalid XML.") from etree_exception
```
anthrotype, line 1091: we can't also print the name of the glyph, can we?

anthrotype: we probably need to catch the exception higher in the call stack to be able to print the name -- but I think it's worth it

Hoolean: Yes agreed :+1: I think the ideal exception raising would:
- print the .glif path and name in fonttools
- print further the UFO and layer name in ufoLib2

This PR is intended as a first step towards that in giving us something non-lxml/xml-specific to hang on to when we see the error at a higher level; what do you think about something like this:
```python
except GlifLibError as glif_lib_error:
    raise GlifLoadError(name, glif_path) from glif_lib_error
```
in `readGlyphFromString()`/`glifLib.GlyphSet.readGlyph()`?

Hoolean: yes I think so, making a new GlifLoadError type, although currently not sure whether any of the other exceptions that are thrown should be wrapped too; e.g. `readGlyph()` currently raises KeyError, GLIFFormatError, UnsupportedGLIFFormat explicitly as well

anthrotype (approved):
LGTM

PR #3031, opened 2023-03-10
[sbix] Fix UnboundLocalError: cannot access local variable 'rawdata' where

liZe:


anthrotype:
oops. thanks. Looks like that line wasn't tested..

anthrotype:
looks like none of the sbix module is in fact. I'll file a separate issue about that, and merge this for now

liZe:
> looks like none of the sbix module is tested in fact. I'll file a separate issue about that, and merge this for now

Thanks a lot!

PR #3034, opened 2023-03-13
[otlib] Add a config option to write GPOS 7 lookups

khaledhosny:
See the inline comment and option documentation

### DIFF
--- a/Lib/fontTools/config/__init__.py
+++ b/Lib/fontTools/config/__init__.py
@@ -62,8 +62,8 @@ Config.register_option(
     name="fontTools.otlLib.builder:WRITE_GPOS7",
     help=dedent(
         """\
-        macOS before 13.2 didn’t support GPOS LookupType 7 (none-chaining
-        ContextPos lookups), so FontTools.otLib.builder disables a file size
+        macOS before 13.2 didn’t support GPOS LookupType 7 (non-chaining
+        ContextPos lookups), so FontTools.otlLib.builder disables a file size
         optimisation that would use LookupType 7 instead of 8 when there is no
         chaining (no prefix or suffix). Set to True to enable the optimization.
         """
--- a/Lib/fontTools/config/__init__.py
+++ b/Lib/fontTools/config/__init__.py
@@ -57,3 +57,18 @@ Config.register_option(
     parse=Option.parse_optional_bool,
     validate=Option.validate_optional_bool,
 )
+
+Config.register_option(
+    name="fontTools.otlLib.builder:WRITE_GPOS7",
+    help=dedent(
+        """\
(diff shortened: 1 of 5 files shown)


Lib/fontTools/otlLib/builder.py
```diff
@@ -369,10 +369,19 @@ def build(self):
...
         # correctly process GPOS7 lookups, so for now we force contextual
         # positioning lookups to be chaining (GPOS8).
-        if self.subtable_type == "Pos":  # horrible separation of concerns breach
+        #
+        # This seems to be fixed as of macOS 13.2, but we keep disabling this
+        # for now until we are no longer concerned about old macOS versions.
+        # But we allow people to opt-out of this with the config key below.
+        write_gpos7 = self.font.cfg.get("fontTools.otlLib.builder:WRITE_GPOS7")
+        # horrible separation of concerns breach
+        if not write_gpos7 and self.subtable_type == "Pos":
```
anthrotype, line 384: What exactly is "horrible"?

anthrotype (commented):
LGTM

Lib/fontTools/config/__init__.py
```diff
@@ -57,3 +57,18 @@ class Config(AbstractConfig):
     parse=Option.parse_optional_bool,
     validate=Option.validate_optional_bool,
 )
+
+Config.register_option(
+    name="fontTools.otlLib.builder:WRITE_GPOS7",
+    help=dedent(
+        """\
+        macOS before 13.2 didn’t support GPOS LookupType 7 (none-chaining
```
dscorbett, line 65: “none-chaining” should be “non-chaining”.

Lib/fontTools/config/__init__.py
```diff
@@ -57,3 +57,18 @@ class Config(AbstractConfig):
...
+        ContextPos lookups), so FontTools.otLib.builder disables a file size
```
dscorbett, line 66: “FontTools.otLib.builder” should be “fontTools.otlLib.builder”.

Lib/fontTools/otlLib/builder.py
khaledhosny: That is an existing comment, I moved it up to improve code formatting.

anthrotype: I know it preexisted, but I don't understand why it is horrible breach of separation of concerns

khaledhosny: That is a question to @simoncozens I guess.

anthrotype:
can we merge this? I'd like to cut a fonttools release soon as I'd like to use some newly added functionality in ufo2ft, and this PR seems ready to me

PR #3038, opened 2023-03-14
Handle string input to Program.fromAssembly()

### LINKED ISSUE #3036: Weird XML output for ttProgram.Program.fromAssembly()
When a TrueType program is built from assembly, and then immediately dumped to XML, the output format is very weird, one character per line.

```python
from fontTools.misc.xmlWriter import XMLWriter
from fontTools.ttLib.tables.ttProgram import Program
from io import BytesIO


p = Program()
p.fromAssembly("PUSHB[ ]\n0 0\nPOP[ ]\nPOP[ ]")

buf = BytesIO()

p.toXML(XMLWriter(buf), None)
print(buf.getvalue())


p.fromBytecode(p.getBytecode())

buf = BytesIO()

p.toXML(XMLWriter(buf), None)
print(buf.getvalue())
```

```xml
b'<?xml version="1.0" encoding="UTF-8"?>\n<assembly>\n  P\n  U\n  S\n  H\n  B\n  [\n   \n  ]\n  \n\n  0\n   \n  0\n  \n\n  P\n  O\n  P\n  [\n   \n  ]\n  \n\n  P\n  O\n  P\n  [\n   \n  ]\n</assembly>\n'
```

When the program is assembled to bytecode, and then read from bytecode with Program.fromBytecode, and then dumped to XML, it looks as seen in ttx dumps.

```python
from fontTools.misc.xmlWriter import XMLWriter
from fontTools.ttLib.tables.ttProgram import Program
from io import BytesIO


p = Program()
p.fromAssembly("PUSHB[ ]\n0 0\nPOP[ ]\nPOP[ ]")
p.fromBytecode(p.getBytecode())

buf = BytesIO()

p.toXML(XMLWriter(buf), None)
print(buf.getvalue())
```

```xml
b'<?xml version="1.0" encoding="UTF-8"?>\n<assembly>\n  PUSHB[ ]\t/* 2 values pushed */\n  0 0\n  POP[ ]\t/* PopTopStack */\n  POP[ ]\t/* PopTopStack */\n</assembly>\n'
```

Functionally, there seems to be no difference in the two notations. But is this expected?

jenskutilek:
Fixes #3036

### DIFF
--- a/Lib/fontTools/ttLib/tables/ttProgram.py
+++ b/Lib/fontTools/ttLib/tables/ttProgram.py
@@ -1,8 +1,10 @@
 """ttLib.tables.ttProgram.py -- Assembler/disassembler for TrueType bytecode programs."""
+from __future__ import annotations
 
 from fontTools.misc.textTools import num2binary, binary2num, readHex, strjoin
 import array
 from io import StringIO
+from typing import List
 import re
 import logging
 
@@ -1067,10 +1069,25 @@ class Program(object):
     def __init__(self):
         pass
 
+    @property
+    def assembly(self):
+        if hasattr(self, "_assembly"):
+            return self._assembly
+        raise AttributeError
+
+    @assembly.setter
+    def assembly(self, value: List[str] | str):
+        if isinstance(value, list):
+            self._assembly = value
+        elif isinstance(value, str):
+            self._assembly = value.splitlines()
+        else:
+            raise TypeError
+
     def fromBytecode(self, bytecode):
         self.bytecode = array.array("B", bytecode)
-        if hasattr(self, "assembly"):
-            del self.assembly
+        if hasattr(self, "_assembly"):
+            del self._assembly
 
     def fromAssembly(self, assembly):
         self.assembly = assembly
@@ -1083,7 +1100,7 @@ class Program(object):
         return self.bytecode.tobytes()
 
     def getAssembly(self, preserve=True):
-        if not hasattr(self, "assembly"):
+        if not hasattr(self, "_assembly"):
             self._disassemble(preserve=preserve)
         return self.assembly
 
@@ -1159,7 +1176,7 @@ class Program(object):
         if name == "assembly":
             self.fromAssembly(strjoin(content))
             self._assemble()
-            del self.assembly
+            del self._assembly
         else:
             assert name == "bytecode"
             self.fromBytecode(readHex(content))
--- a/Tests/ttLib/tables/ttProgram_test.py
+++ b/Tests/ttLib/tables/ttProgram_test.py
@@ -12,6 +12,13 @@ DATA_DIR = os.path.join(CURR_DIR, "data")
 TTPROGRAM_TTX = os.path.join(DATA_DIR, "ttProgram.ttx")
 # TTPROGRAM_BIN = os.path.join(DATA_DIR, "ttProgram.bin")
 
+ASSEMBLY = [
+    "PUSH[ ]",
+    "0 4 3",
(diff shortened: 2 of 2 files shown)


Lib/fontTools/ttLib/tables/ttProgram.py
```diff
@@ -1067,10 +1069,25 @@ class Program(object):
     def __init__(self):
         pass
 
+    @property
+    def assembly(self):
+        if hasattr(self, "_assembly"):
+            return self._assembly
```
anthrotype, line 1075: just `return self._assembly` is sufficient if you want to raise AttributeError

Lib/fontTools/ttLib/tables/ttProgram.py
```diff
@@ -1067,10 +1069,25 @@ class Program(object):
...
+        raise AttributeError
+
+    @assembly.setter
+    def assembly(self, value: List[str] | str):
+        if isinstance(value, list):
+            self._assembly = value
+        elif isinstance(value, str):
+            self._assembly = value.splitlines()
+        else:
+            raise TypeError
```
anthrotype, line 1085: `raise TypeError(f"expected str or List[str], got {type(value).__name__}")`

Tests/ttLib/tables/ttProgram_test.py
```diff
@@ -90,6 +97,18 @@ def test__bool__(self):
         assert p.assembly.pop() == "SVTCA[0]"
         assert not bool(p)
 
+    def test_from_assembly_list(self):
+        p = Program()
+        p.fromAssembly(ASSEMBLY)
+        asm = p.getAssembly(preserve=True)
```
anthrotype, line 103: preserve=True is already the default

Lib/fontTools/ttLib/tables/ttProgram.py
```diff
@@ -1067,10 +1069,25 @@ class Program(object):
     def __init__(self):
         pass
 
+    @property
+    def assembly(self):
```
anthrotype, line 1073: since you typed the other one, how about `def assembly(self) --> List[str]:`

Lib/fontTools/ttLib/tables/ttProgram.py
jenskutilek: Ah, sure.

Lib/fontTools/ttLib/tables/ttProgram.py
```diff
@@ -1067,10 +1069,23 @@ class Program(object):
...
+
     def fromBytecode(self, bytecode):
         self.bytecode = array.array("B", bytecode)
-        if hasattr(self, "assembly"):
-            del self.assembly
+        if hasattr(self, "_assembly"):
+            del self._assembly
 
     def fromAssembly(self, assembly):
         self.assembly = assembly
```
anthrotype, line 1091: I don't think you even need to add the getter/setter for `self.assebly`, you can simply move that code you put in the `assembly` setter (that splits a str into a List[str]) directly inside the `fromAssembly` method, which will ensure that `self.assembly` is a list. There is already a `getAssembly` so I don't see the need for yet another property getter with the additional self._assembly private attribute...

Lib/fontTools/ttLib/tables/ttProgram.py
jenskutilek: I added some more annotations.

anthrotype (approved):
LGTM

Do you mind swashing the commits before merge? Thanks

anthrotype:
squashing, not swashing 😄

Issue #3040, opened 2023-03-14
ConditionSets and VariableScalars should have `avar` applied to coordinates


madig:
I found a case where a mapping in a Designspace file like so:

```xml
  <axes>
    <axis tag="wght" name="Weight" minimum="100" maximum="900" default="100">
      <map input="100" output="22"/>
      <map input="600" output="115"/>
      <map input="900" output="185"/>
    </axis>
  </axes>
  <rules processing="last">
    <rule name="conditional_sub">
      <conditionset>
        <condition name="Weight" minimum="116" maximum="185"/>
      </conditionset>
      <sub name="e" with="a"/>
    </rule>
  </rules>
```

means that "e" is substituted by "a" at wght=606 and upwards (116 in design coordinates is a tad over 600 in user coordinates), whereas with a feature file snippet like this:

```
lookup conditional_sub {
    sub e by a;
} conditional_sub;

conditionset test {
    wght 606 900;
} test;

variation rlig test {
    lookup conditional_sub;
} rlig;
```

it is substituted at wght=644 and upwards.

anthrotype:
From @simoncozens' https://github.com/adobe-type-tools/afdko/pull/1350#issuecomment-955038971

> userspace values as taken from the fvar table (wght=400 in the Nunito example), which is how we’ve interpreted things so far

so, yes, whenever fvar userspace coordinates are given, avar needs to be taken into account when normalizing. 

E.g. by doing https://github.com/fonttools/fonttools/blob/032616b10685bfca139ff702d1ffc6458891d980/Lib/fontTools/varLib/instancer/__init__.py#L1069-L1074

simoncozens:
This is similarly true for locations used in variable scalars.

madig:
Hm, so, at what level/where should I fix this? There are a couple of normalization functions floating around, i.e. `feaLib.builder.Builder.add_conditionset` is using `varLib.models.normalizeValue`.

madig:
So, at least when changing add_conditionset like so:

```
        if "avar" in self.font:
            mapping = self.font["avar"].segments
            value = {
                axis: tuple(
                    floatToFixedToFloat(piecewiseLinearMap(v, mapping[axis]), 14)
                    for v in condition_range
                )
                for axis, condition_range in value.items()
            }
```

I get almost the same cross-over point (606 instead of 605 with the Designspace). So I guess there's a rounding issue in here somewhere.

anthrotype:
[...]

PR #3042, opened 2023-03-14
Conditionset: consult avar to normalize values

madig:
Closes https://github.com/fonttools/fonttools/issues/3040.

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -2065,7 +2065,7 @@ class ConditionsetStatement(Statement):
         self.conditions = conditions
 
     def build(self, builder):
-        builder.add_conditionset(self.name, self.conditions)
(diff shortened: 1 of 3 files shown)


anthrotype (approved):
LGTM

Lib/fontTools/feaLib/ast.py
```diff
@@ -2063,9 +2063,10 @@ def __init__(self, name, conditions, location=None):
         Statement.__init__(self, location)
```
anthrotype, line 2063: doesn't `Statement.__init__(self, location)` already store `self.location`?

anthrotype: yes, I think that's done in `Element.__init__`, and `Statement` in turn should inherit from Element:

https://github.com/fonttools/fonttools/blob/7b7882355783bc9dd2dd94d6cfcde07c95965ba4/Lib/fontTools/feaLib/ast.py#L163

PR #3043, opened 2023-03-15
Apply `avar` also to variable locations

### LINKED ISSUE #3040: ConditionSets and VariableScalars should have `avar` applied to coordinates

madig:
Follow-up to #3042.

1. Guard against missing `avar` entries when normalizing condition set locations
2. Apply `avar` also to variable locations

Closes #3040.

### DIFF
--- a/Lib/fontTools/feaLib/variableScalar.py
+++ b/Lib/fontTools/feaLib/variableScalar.py
@@ -8,15 +8,6 @@ def Location(loc):
 class VariableScalar:
     """A scalar with different values at different points in the designspace."""
 
-    # We will often use exactly the same locations (i.e. the font's
-    # masters) for a large number of variable scalars. Instead of
-    # creating a model for each, let's share the models.
-    model_pool = {}
-
-    @classmethod
-    def clear_cache(cls):
-        cls.model_pool = {}
-
     def __init__(self, location_value={}):
         self.values = {}
         self.axes = {}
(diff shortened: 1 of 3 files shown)


Lib/fontTools/feaLib/variableScalar.py
```diff
@@ -83,29 +83,37 @@ def default(self):
...
         values = list(self.values.values())
-        return self.model.interpolateFromMasters(loc, values)
+        return self.model(avar).interpolateFromMasters(loc, values)
 
-    @property
-    def model(self):
+    def model(self, avar=None):
         key = tuple(self.values.keys())
         if key in self.model_pool:
             return self.model_pool[key]
```
anthrotype, line 96: if the VariableScalar.values that are used as the key in the model_pool chace are user-space, pre-normalization.. but then you pass in avar as a parameter, which in turn might change the normalized location stored inside the VariationModel...
the cache's key should either include avar somehow, or it should have normalized values to be avar-independent, otherwise there's a risk that you get a wrong cached model, one that doesn't take the passed-in avar into account.

madig: Ugh. Maybe (store into/load from) cache first, then apply avar? What's a good benchmark for ensuring this doesn't undo the optimization?

anthrotype: how about we do not store the model cache inside the VariableScalar class scope _at all_? Can you try moving it to the feaLib builder's state?

Issue #3045, opened 2023-03-16
[varLib] regression in v4.39.1, builds incomplete STAT 


anthrotype:
[ufo2ft intergration tests](https://github.com/googlefonts/ufo2ft/actions/runs/4438446665/jobs/7789512575?pr=718) alerted me of a regression introduced with https://github.com/fonttools/fonttools/pull/3024 just released today in 4.39.1, whereby the STAT info present in DesignSpace v5 are being ignored and only an incomplete STAT table is built.

I identified the culrpit and and working on a quick fix

anthrotype:
I had to temporarily "yank" the 4.39.1 release from PyPI because the fix is a bit more involved than I first thought. Again apologies

PR #3046, opened 2023-03-16
[varLib] don't skip buildVFStatTable from DSv5 with 'STAT' info

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/3045


madig (commented):
Seems sensible to me.

Issue #3069, opened 2023-03-29
pyftmerge OTF changes width of glyphs in output file


KatjaSchimmel:
I'm trying to merge two OTF (one with only Latin Glyphs, one with only Arabic glyphs). In the output file, the right sidebearing of the glyphs change depending on which OTF is named first in the terminal command.
For example:
- Latin OTF first (LA-Regular.otf)
`pyftmerge L-Regular.otf A-Regular.otf --output-file=LA-Regular.otf
`
The right sidebearing of all Arabic glyphs change.
- Arabic OTF first (AL-Regular.otf)
`pyftmerge A-Regular.otf L-Regular.otf --output-file=AL-Regular.otf
`
The right sidebearing of all Latin glyphs change.

I did several tests to find out the value added or subtracted from the right sidebearing. 
In my latest test I merged a Latin OTF containing only **H** with an Arabic OTF containing only **alef** and **space**. The output file showed following:
- Latin OTF first (LA-Regular.otf)
LSB of H = 100 units
RSB of H = 100 units
LSB of alef = 95
RSB of alef = 335 (should be 95, 240 units are added)
width of space = 500 (should be 260, 240 units are added)
- Arabic OTF first (AL-Regular.otf)
LSB of H = 100 units
RSB of H = -148 units (should be 100, 248 units are subtracted)
LSB of alef = 95
RSB of alef = 95
width of space = 260

The value of 240 is being added to the Arabic Glyphs, in the merged "LA" file. The value of 248 subtracted from the Latin glyphs, in the merged "AL" file. I'm adding a screenshot of the CFF table. 

I also found out that this value changes, when merging bigger charactersets. For example I merged one Latin OTF with Latin Plus characterset and one Arabic OTF with Arabic Language coverage. The value added/subtracted from the right sidebearing was 176 units for both merged output files.

I have no idea why the width of the glyphs change and where the values come from. Do you know what is happening here?

Merging two TTF is no problem. All widths are maintained.

<img width="1033" alt="Screenshot 2023-03-29 at 15 01 16" src="https://user-images.githubusercontent.com/32390109/228543617-90cdbd77-3900-455c-9c0e-a2f650b98d21.png">

anthrotype:
[...]

PR #3070, opened 2023-03-29
[merge] Handle differing default/nominalWidthX in CFF

behdad:
Fixes https://github.com/fonttools/fonttools/issues/3069

### DIFF
--- a/Lib/fontTools/merge/tables.py
+++ b/Lib/fontTools/merge/tables.py
@@ -3,6 +3,7 @@
 # Google Author(s): Behdad Esfahbod, Roozbeh Pournader
 
 from fontTools import ttLib, cffLib
+from fontTools.misc.psCharStrings import T2WidthExtractor
 from fontTools.ttLib.tables.DefaultTable import DefaultTable
 from fontTools.merge.base import add_method, mergeObjects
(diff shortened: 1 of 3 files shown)


behdad:
The test expectation rightly needs to be updated IMO.

KatjaSchimmel:
Hi! I was just wondering if this will be merged soon. I am looking forward to testing it

Lib/fontTools/merge/tables.py
```diff
@@ -278,6 +286,18 @@ def merge(self, m, tables):
                 newcs.charStrings[name] = i
                 newcs.charStringsIndex.items.append(None)
         for name in cs.charStrings:
+            if widthsDiffer:
+                c = cs[name]
+                defaultWidthXToken = object()
+                extractor = T2WidthExtractor([], [], nominalWidthX, defaultWidthXToken)
+                extractor.execute(c)
+                width = extractor.width
+                if width != defaultWidthXToken:
```
anthrotype, line 295: since you want to test for identity with the unique `object()` you may as well us `is` operator: 
```suggestion
                if width is not defaultWidthXToken:
```

anthrotype (approved):
belated LGTM!

PR #3073, opened 2023-04-05
Optionally keep empty lookups for interpolatability

madig:
WIP.

For dealing with https://github.com/googlefonts/fontmake/issues/894#issuecomment-1492672869, we may want to inject empty lookups into fonts, to give varLib something to grab onto. We could do this at the feature level in ufo2ft, before passing the font to varLib to merge. This is some hacking to make feaLib not drop empty lookups.

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -18,6 +18,7 @@ from fontTools.otlLib.builder import (
     ChainContextPosBuilder,
     ChainContextSubstBuilder,
     LigatureSubstBuilder,
+    LookupBuilder,
     MultipleSubstBuilder,
     CursivePosBuilder,
     MarkBasePosBuilder,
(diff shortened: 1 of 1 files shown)


Lib/fontTools/feaLib/builder.py
```diff
@@ -89,6 +90,25 @@ def addOpenTypeFeaturesFromString(
...
+    def __init__(self, font, location):
+        super().__init__(font, location, "GPOS", 2)
+
+    def build(self):
+        from fontTools.ttLib.tables import otTables as ot
+
+        lookup = ot.Lookup()
+        lookup.LookupType = 2
+        lookup.LookupFlag = 0
+        lookup.SubTable = []
```
anthrotype, line 106: Are we sure that we don't actually want tp have `lookup.SubTable = None` here instead of the empty list?
We need to check what varLib makes of these, whether the it disinguishes one with SubTable=[] vs one with SubTable=None. In general what we want in these cases is that this particular lookup "does not participate" in variations basically.

Lib/fontTools/feaLib/builder.py
```diff
@@ -1069,6 +1089,8 @@ def add_lookup_call(self, lookup_name):
         assert lookup_name in self.named_lookups_, lookup_name
         self.cur_lookup_ = None
         lookup = self.named_lookups_[lookup_name]
+        if lookup is None and self.cur_feature_name_ in ("kern", "dist"):
```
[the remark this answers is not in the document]
madig: I think you need to specify the subtype of the lookup and I'm not sure how to make them universal.

Lib/fontTools/feaLib/builder.py
behdad: This would be really interesting indeed, to allow sparse models at the lookup level.

anthrotype: but it's impossible with FEA because truly empty lookups have no type, it's their rules that give them their type. Unless you want to hack this by adding dummy `pos .notdef .notdef 0` ...

behdad: It's okay to give the lookup type 0. Except that both fonttools compiler and varLib would choke on it then. :(

behdad: Can we extend the format?

anthrotype: we could...

PR #3075, opened 2023-04-05
[varLib.interpolatable] Allow for sparse masters

### LINKED ISSUE #3074: [varLib.interpolatable] Allow for sparse masters
Hi!

I love using `varLib.interpolatable` but I noticed it doesn't work well with sparse masters. I think the default behavior should still be to report missing glyphs, but perhaps there should be a `--ignore-missing` flag to allow the tool to run without reporting missing glyphs as errors. It should still check for compatibility between glyphs in masters in which they exist, though.

For example:
- Font1 has `a` and `e` glyphs
- Font2 just has `e`
- Font3 has `a` and `e`

`varLib.interpolatable --ignore-missing` should check Font1, Font2, and Font3 for compatibility in `e`, but only Font1 and Font3 for `a`. Crucially, if the sparse master happens to be the first master, it _should not_ skip `a` altogether.

colinmford:
This PR fixes #3074.

It adds a `--ignore-missing` flag to allow `varLib.interpolatable` to check sparse masters without raising missing glyphs as errors.

Functionally, I edited the `test` function to not assume all the masters had the same glyphset, and more specifically not to rely on the first master in the `glyphsets` list (`m0`) as much.

I certainly welcome edits/suggestions if you have any!


colinmford:
Oh, this PR also adds a `--quiet` flag for user convenience — the script will not print anything, it'll just return exit code `1` or `0`

Lib/fontTools/varLib/interpolatable.py
```diff
@@ -284,15 +294,21 @@ def add_problem(glyphname, problem):
                                     "path": pathIx,
                                     "node": nodeIx,
                                     "master_1": names[0],
```
behdad, line 296: 0 should be m0idx here?

colinmford: Yes - thank you! Fixed!

behdad:
Thanks Colin! Looks good to me.

Did you test it with Glyphs, Designspace, and variable font by any chance?

colinmford:
@behdad only ttfs and designspace — I can write those tests for Glyphs, Designspace, and VF, though!

behdad:
> I can write those tests for Glyphs, Designspace, and VF, though!

That would be great. Thank you!

colinmford:
@behdad Added more tests ! VF didn't really make sense for `--ignore-missing`, but I added the rest — for sparse masters and there were only tests for otf and ttf for "non-sparse" spaces, so I added VF, ufo, designspace and Glyphs tests for those too.

behdad:
> @behdad Added more tests! VF didn't really make sense for `--ignore-missing`, but I added the rest — for sparse masters and there were only tests for otf and ttf for "non-sparse" spaces, so I added VF, ufo, designspace and Glyphs tests for those too.

Thank you very much! Appreciated.

behdad:
Can you please run `black` on the modified files, such that the lint test passes?

colinmford:
@behdad of course — done!

colinmford:
Sorry — new to `black`... hopefully this time is the charm

Lib/fontTools/varLib/interpolatable.py
```diff
@@ -365,7 +393,21 @@ def main(args=None):
...
     )
     parser.add_argument(
-        "inputs", metavar="FILE", type=str, nargs="+", help="Input TTF/UFO files"
+        "--quiet",
+        action="store_true",
+        help="Only exit with code 1 or 0, no output",
+    )
+    parser.add_argument(
+        "--ignore-missing",
+        action="store_true",
```
anthrotype, line 402: I wonder if maybe this should be True by default, i.e. expect inputs to be sparse by default, and maybe we add an option to `--report-missing-glyphs`.

anthrotype (approved):

anthrotype:
Thanks Colin!
Looks like the tests are failing because glyphsLib is not installed and is required for some of the tests you added. Adding glyphsLib to the list of fonttools' own dependencies (even if only the optional "extras") is a bit of a chicken and egg because glyphsLib in turn depends on fonttools.
could you try adding it to the requirements.txt file (which is used by tox to initialise the test virtual environment), similarly to how we did for other test-only deps like ufoLib2. Add a comment saying that it's only required for the interpolatable_test.py.
https://github.com/fonttools/fonttools/blob/main/requirements.txt

anthrotype:
if you could also maybe remove some of the private lib cruft that is in the test UFO/glyphs file which is not strictly needed for the tests, it would be better, so we reduce the number/size of test files added with this PR, thanks

Lib/fontTools/varLib/interpolatable.py
anthrotype: I would argue it is an improvement of the tool, more than a breaking change ;)
@behdad WDYT?

anthrotype:
can you please use `pytest.importorskip("glyphsLib")` for the test methods where glyphsLib is going to be imported? We want the fonttools core tests to be able to run without third-party deps. When importorskip fails to import the named module, the given test is skipped.

like we do e.g. here for pathops in the instancer_test.py:
https://github.com/fonttools/fonttools/blob/58bc16e58f48367d902cbaeaa9b6a6e5971a9b72/Tests/varLib/instancer/instancer_test.py#L1537

https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-importorskip

anthrotype:
>  I added glyphsLib to the requirements

that's fine, the requirements.txt is only used for setting up the test venv. The canonical set of dependencies (used by pip when installing fonttools) is listed inside setup.py. I think it makes sense not to add glyphsLib to the fonttools[interpolatable] set of extra deps because glyphsLib already depends on fonttools and that would be a circular dependency; if one has already glyphsLib, one has also fonttools and can use varLib.interpolatable if they need so.

Lib/fontTools/varLib/interpolatable.py
behdad: I'm fine either way.

colinmford: I'm fine either way as well, but speaking as someone who is using this script as a sort of pre-build check, I would personally rather opt-in to a behavior than to opt-out of it, especially since in this case it's changing the default behavior. 

I was also worried about causing trouble for other scripts that used this script in some way — for example fontbakery uses it for a check.

anthrotype: I'm fine to keep it opt-in

colinmford: Ok, seems like we're agreed — will resolve this conversation

colinmford:
Looks like we're good to go!

colinmford:
Thanks, @anthrotype and @behdad!

PR #3092, opened 2023-04-25
Fix missing composite instructions in ttx

### LINKED ISSUE #3091: [4.39+] Composite TrueType instructions are discarded when dumping to XML
I found a serious bug that apparently was introduced in release 4.39.0.

In the attached demo font, the B glyph is a composite which has TrueType instructions. When dumped with ttx in FontTools 4.39.0+, the instructions of any composites are missing from the ttx file.

[2023-composite-hints.zip](https://github.com/fonttools/fonttools/files/11322045/2023-composite-hints.zip)

jenskutilek:
Fixes #3091 and adds a test.


anthrotype (approved):
good catch! makes sense

anthrotype:
feel free to merge this when the test pass, I will cut a bugfix release later today

PR #3094, opened 2023-04-28
Expose and fix case where interpolability was destroyed by reverse

### LINKED ISSUE #3093: Reversing interpolation compatible outlines should produce compatible outlines when outputImpliedClosingLine=True
A subpath (sequence of drawing commands starting with a move and containing no other move) is interpolation compatible with another subpath iff both use the same drawing commands in the same order. 

For example, move, curve, close (MCZ) is compatible with all other MCZ sequences.

It should not matter:

* What the values of the associated points, if any, are
* Whether we interact with the curve as a point stream, drawing command stream (context-aware drawing commands, e.g. a quad has two points, implicitly starting at the current position), or segments (self-contained drawing commands, e.g. a quad has 3 points rather than implicitly using current as the first)

Reversing paths shouldn't cause them to become interpolation incompatible. https://gist.github.com/rsheeter/cef6cf71fa9e519c94f9f0d23906e5de demonstrates a case where this appears not to be the case.

rsheeter:
Fixes #3093 by applying @behdad proposed change and then making the tests happy again.

### DIFF
--- a/Lib/fontTools/pens/pointPen.py
+++ b/Lib/fontTools/pens/pointPen.py
@@ -237,6 +237,7 @@ class PointToSegmentPen(BasePointToSegmentPen):
                 # 'outputImpliedClosingLine' option) in order to disambiguate this case from
                 # the implied closing 'lineTo', otherwise the duplicate point would be lost.
                 # See https://github.com/googlefonts/fontmake/issues/572.
+
                 if (
                     i + 1 != nSegments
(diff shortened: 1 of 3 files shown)


rsheeter:
After extensive discussion, it appears we agree the the svg/pdf/postscript guidance calls to always emit the line but we don't want that in fonttools so withdrawing. We will try it out in fontmake-rs and see where that takes us :)

Issue #3097, opened 2023-04-29
feature variations lookup order only rvrn should be first


punchcutter:
https://github.com/fonttools/fonttools/pull/2867/files

This change seems to have broken processing=last features because they are also getting forced to reorder unless I'm looking at something wrong. With the update from 4.38.0 in afdko we have a broken build now.

behdad:
> processing=last features

Excuse my fading memory. What are those?

punchcutter:
rclt by default usually, but could be subbed for rlig. Could of course be anything in the world and should still work.

behdad:
My question was where does the terminology come from. Now I see it's from designspace.

punchcutter:
Ah, yes. Of course I'd expect it always in lookup order except for the ridiculous case of rvrn.

PR #3099, opened 2023-05-01
[featureVars] Process lookups for features other than rvrn last

behdad:
Fixes https://github.com/fonttools/fonttools/issues/3097

One test needs updated expectations.

### DIFF
--- a/Lib/fontTools/varLib/featureVars.py
+++ b/Lib/fontTools/varLib/featureVars.py
@@ -14,7 +14,9 @@ from collections import OrderedDict
 from .errors import VarLibError, VarLibValidationError
 
 
-def addFeatureVariations(font, conditionalSubstitutions, featureTag="rvrn"):
+def addFeatureVariations(
+    font, conditionalSubstitutions, featureTag="rvrn", processLast=None
+):
     """Add conditional substitutions to a Variable Font.
 
     The `conditionalSubstitutions` argument is a list of (Region, Substitutions)
@@ -45,6 +47,9 @@ def addFeatureVariations(font, conditionalSubstitutions, featureTag="rvrn"):
     # >>> f.save(dstPath)
     """
 
+    if processLast is None:
+        processLast = featureTag != "rvrn"
+
     _checkSubstitutionGlyphsExist(
         glyphNames=set(font.getGlyphOrder()),
         substitutions=conditionalSubstitutions,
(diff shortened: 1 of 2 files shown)


behdad:
This basically undoes https://github.com/fonttools/fonttools/pull/2867/files and adds an option for it.

behdad:
@punchcutter Can you please test this?

Lib/fontTools/varLib/featureVars.py
```diff
@@ -14,7 +14,9 @@
 from .errors import VarLibError, VarLibValidationError
 
 
-def addFeatureVariations(font, conditionalSubstitutions, featureTag="rvrn"):
+def addFeatureVariations(
+    font, conditionalSubstitutions, featureTag="rvrn", processLast=None
```
anthrotype, line 18: this new `processLast` parameter does not seem to be used anywhere, I think it's ok if remove it altogether and automatically append when featureTag != "rvrn", otherwise insert at index 0 for "rvrn". I don't see its utility, why would one ever want to _not_ insert at 0 for "rvrn" or, conversely, to insert at 0 (instead of append) for any other feature tag that is != "rvrn".
Right now we ignore the `processing` attribute of DS rules if there is a `featureVarsFeatureTag` key in the designspace lib, since the two methods were intended to be alternative to one another:

https://github.com/fonttools/fonttools/blob/f026853cb28df60006347cecf69f57e76c56d91e/Lib/fontTools/varLib/__init__.py#L55-L56

I mean, we could chose to apply the rules `processing` attribute whenever present and pass it down to featureVars.addFeatureVariations. Only in this case it would make sense to keep the `processLast` parameter for addFeatureVariations function (though I am not 100% sure how useful that would be)

anthrotype (approved):
LGTM, but see earlier comment about unused processLast parameter

Lib/fontTools/varLib/featureVars.py
behdad: > I mean, we could chose to apply the rules processing attribute whenever present and pass it down to featureVars.addFeatureVariations. Only in this case it would make sense to keep the processLast parameter for addFeatureVariations function (though I am not 100% sure how useful that would be)

That was my intention, to wire it up.

behdad: Wired up now.

behdad: I think I better understand your point about not exposing this parameter now. Let me think.

behdad: Removed.

anthrotype (approved):

PR #3103, opened 2023-05-08
feaLib: support multiple substitution with classes

khaledhosny:
This is a GlyphsApp extension, (partially) documented here, but it is a useful one as it allows concise glyph insertion using decomposition without having to maintain the list of glyphs in two places. E.g.

```fea
@upper = [A-Z];

lookup insert {
  sub @upper by @upper connector;
} insert;

feature ccmp {
  sub @upper' lookup insert x;
} ccmp;
```

As apposed to:

```fea
@upper = [A-Z];

lookup insert {
  sub A by A connector;
  sub B by B connector;
  sub C by C connector;
  sub D by D connector;
  sub E by E connector;
  sub F by F connector;
  sub G by G connector;
  sub H by H connector;
  sub I by I connector;
  sub J by J connector;
  sub K by K connector;
  sub L by L connector;
  sub M by M connector;
  sub N by N connector;
  sub O by O connector;
  sub P by P connector;
  sub R by R connector;
  sub S by S connector;
  sub T by T connector;
  sub U by U connector;
  sub V by V connector;
  sub W by W connector;
  sub X by X connector;
  sub Y by Y connector;
  sub Z by Z connector;
} insert;

feature ccmp {
  sub @upper' lookup insert x;
} ccmp;
```


Lib/fontTools/feaLib/parser.py
```diff
@@ -924,23 +924,25 @@ def parse_substitute_(self):
...
-            not reverse
-            and len(old) == 1
-            and len(old[0].glyphSet()) == 1
-            and len(new) > 1
-            and max([len(n.glyphSet()) for n in new]) == 1
-            and num_lookups == 0
-        ):
+        # Format A: "substitute f_f_i by f f i;"
+        # Format B: "substitute [f_i f_l] by [f f] [i l];"
+        # Format C: "substitute [f_i f_l] by f [i l];"
```
anthrotype, line 929: where do these A, B and C come from? I don't see them listed in http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#5.b

Lib/fontTools/feaLib/ast.py
```diff
@@ -1256,23 +1256,29 @@ def build(self, builder):
...
-                    prefix,
-                    glyph,
-                    suffix,
-                    self.replacement,
-                    self.forceChain,
-                )
+        if hasattr(self.glyph, "glyphSet"):
+            originals = self.glyph.glyphSet()
         else:
+            originals = [self.glyph]
```
anthrotype, line 1262: codecov says this line isn't covered, should it be?

anthrotype (approved):
I'm not against adding support for this as it's convenient but I worry that implementations and upstream spec may diverge too much. Ideally we should make a PR against spec and update it accordingly.

Can you also please add a link to the Glyphs docs for this somewhere in the code?

Otherwise the code looks good to me.

anthrotype (approved):
I'm not against adding support for this as it's convenient but I worry that implementations and upstream spec may diverge too much. Ideally we should make a PR against spec and update it accordingly.

Can you also please add a link to the Glyphs docs for this somewhere in the code?

Otherwise the code looks good to me.

anthrotype:
maybe you could also add a test in which you call a named lookup (as in your example from the PR description)

Lib/fontTools/feaLib/parser.py
khaledhosny: I pulled the numbers out of my hat, as I was planning to propose the additional syntax to AFDKO. I’m goingto add a link to Glyphs documentation there, and name them GlyphsApp format 1 and 2.

Lib/fontTools/feaLib/ast.py
khaledhosny: I’m not sure. Going from the parser, this should never happen, but the old code has a similar check so I guessed it would be for constructing the ast classes directly so one can use strings for glyph names instead of `ast.GlyphName()`.

khaledhosny:
> maybe you could also add a test in which you call a named lookup (as in your example from the PR description)

Done.

khaledhosny:
https://github.com/adobe-type-tools/afdko/pull/1651

PR #3105, opened 2023-05-13
feaLib: Dedupe multiple substitutions with classes

dscorbett:
This is a follow-up to #3103. If the same glyph appears multiple times in the input class, only the first copy’s rule can ever be applied, so the rest can be skipped. Without this change, if the rule is in a chaining substitution lookup, `Builder.add_multiple_subst` creates duplicate multiple substitution lookups.


khaledhosny (approved):
LGTM. Would be nice to have a test.

Issue #3111, opened 2023-05-18
Merging kerning lookup statements fails when default source has value of 0


madig:
A font with contextual kerning had the following line in a lookup for the default source:

```fea
pos @group1 @group2 @group3' 0;
```

while the other masters had different values. This fails when merging into a VF, because the default source's zero gets turned into a `None`, which then looks to varLib merging as if the default source was missing, raising a `VariationModelError`.

anthrotype:
I believe the culprit is this use of ValueRecord.getEffectiveFormat() (instead of getFormat) in the varLib.merger that turns 0s into Nones:

https://github.com/fonttools/fonttools/blob/65bc6105f7aec3478427525d23ddf2e3c8c4b21e/Lib/fontTools/varLib/merger.py#L359-L361

behdad:
Can I have a test please?

madig:
```py
from fontTools.designspaceLib import DesignSpaceDocument
from ufo2ft import compileVariableTTF
from ufoLib2 import Font

ds = DesignSpaceDocument()
ds.addAxisDescriptor(name="wght", tag="wght", minimum=100, maximum=900, default=400)
ds.addSourceDescriptor(font=Font(), location=dict(wght=100))
ds.addSourceDescriptor(font=Font(), location=dict(wght=400))
ds.addSourceDescriptor(font=Font(), location=dict(wght=900))

ds.sources[0].font.newGlyph("a").unicode = ord("a")
ds.sources[0].font.newGlyph("b").unicode = ord("b")
ds.sources[0].font.features.text = "feature kern { pos a b b' 100; } kern;"

ds.sources[1].font.newGlyph("a").unicode = ord("a")
ds.sources[1].font.newGlyph("b").unicode = ord("b")
ds.sources[1].font.features.text = "feature kern { pos a b b' 0; } kern;"

ds.sources[2].font.newGlyph("a").unicode = ord("a")
ds.sources[2].font.newGlyph("b").unicode = ord("b")
ds.sources[2].font.features.text = "feature kern { pos a b b' -100; } kern;"

font = compileVariableTTF(ds, inplace=True)
font.save("test.ttf")
```

Change 0 to 1 and it werks.

behdad:
Alternate fix. I prefer this one:
```diff
diff --git a/Lib/fontTools/otlLib/builder.py b/Lib/fontTools/otlLib/builder.py
index 7a5848ad8..af166401a 100644
[...]

PR #3112, opened 2023-05-19
[varLib/merger] Fix merging of SinglePos with pos=0

behdad:
Fixes https://github.com/fonttools/fonttools/issues/3111


anthrotype (approved):
LGTM, ideally the test didn't simply check we don't crash but asserted the output is as expected (e.g. a ttx dump of GPOS table should be enough)

PR #3114, opened 2023-05-22
[subset] Prune GDEF.MarkGlyphSetsDef

khaledhosny:
Remove empty sets and fixup Lookup.MarkFilteringSet to use the new indices.

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -1513,6 +1513,12 @@ def closure_glyphs(self, s, cur_glyphs=None):
 def subset_glyphs(self, s):
     self.SubTable = [st for st in self.SubTable if st and st.subset_glyphs(s)]
     self.SubTableCount = len(self.SubTable)
+    if hasattr(self, "MarkFilteringSet") and self.MarkFilteringSet is not None:
+        if self.MarkFilteringSet not in s.used_mark_sets:
+            self.MarkFilteringSet = None
+            self.LookupFlag &= ~0x10
+        else:
+            self.MarkFilteringSet = s.used_mark_sets.index(self.MarkFilteringSet)
     return bool(self.SubTableCount)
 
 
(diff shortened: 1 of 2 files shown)


khaledhosny:
Needs a test.

anthrotype (approved):
LGTM!

anthrotype:
does hb-subset do the same? if it, I suppose it should also be updated /cc @qxliu76 @garretrieger @behdad

khaledhosny:
> does hb-subset do the same? if not, I suppose it should also be updated

Looks like hb-subset does not prune them.

anthrotype:
> Looks like hb-subset does not prune them.

maybe file an issue upstream in harfbuzz/harfbuzz

khaledhosny:
> > Looks like hb-subset does not prune them.
> 
> thanks for checking, maybe file an issue upstream in harfbuzz/harfbuzz

https://github.com/harfbuzz/harfbuzz/issues/4239

Lib/fontTools/subset/__init__.py
```diff
@@ -2091,17 +2097,14 @@ def subset_glyphs(self, s):
...
                 coverage.subset(glyphs)
 
-        # TODO: The following is disabled. If enabling, we need to go fixup all
-        # lookups that use MarkFilteringSet and map their set.
-        # indices = table.MarkGlyphSetsDef.Coverage = \
-        #   [c for c in table.MarkGlyphSetsDef.Coverage if c.glyphs]
-        # TODO: The following is disabled, as ots doesn't like it. Phew...
-        # https://github.com/khaledhosny/ots/issues/172
-        # table.MarkGlyphSetsDef.Coverage = [c if c.glyphs else None for c in table.MarkGlyphSetsDef.Coverage]
+        s.used_mark_sets = [i for i, c in enumerate(markGlyphSets.Coverage) if c.glyphs]
```
behdad, line 2105: This introduces a dependency on the order the tables are subset. We don't enforce that right now.

khaledhosny: I thought `_sort_tables` would sort the tables alphabatically, so `GDEF` always comes first, but reading again it seem to leave the other tables in their original order. May be I should add `GDEF`, `GSUB` and `GPOS` their?
https://github.com/fonttools/fonttools/blob/a537f97b44118e7a14cc6e507701e9fca95b81e5/Lib/fontTools/subset/__init__.py#L3498C4-L3502

behdad: Sounds right.

khaledhosny: https://github.com/fonttools/fonttools/pull/3118

Issue #2614, opened 2022-05-06
[ttLib.DSIG] Error when DSIG XML dump is split at XML parse buffer size


justvanrossum:
When compiling the attached TTX file to TTF, it crashes like this:

```
$ ttx -f roundtrip.ttx
Compiling "roundtrip.ttx" to "roundtrip.ttf"...
Parsing 'GlyphOrder' table...
Parsing 'head' table...
Parsing 'hhea' table...
Parsing 'maxp' table...
Parsing 'OS/2' table...
Parsing 'hmtx' table...
Parsing 'cmap' table...
Parsing 'loca' table...
Parsing 'glyf' table...
Parsing 'name' table...
Parsing 'post' table...
Parsing 'DMMY' table...
Parsing 'DSIG' table...
ERROR: Unhandled exception has occurred
Traceback (most recent call last):
  File "/Users/just/code/git/fonttools_just/Lib/fontTools/ttx.py", line 405, in main
    process(jobs, options)
  File "/Users/just/code/git/fonttools_just/Lib/fontTools/ttx.py", line 387, in process
    action(input, output, options)
  File "/Users/just/code/git/fonttools_just/Lib/fontTools/misc/loggingTools.py", line 372, in wrapper
    return func(*args, **kwds)
  File "/Users/just/code/git/fonttools_just/Lib/fontTools/ttx.py", line 298, in ttCompile
    ttf.importXML(input)
  File "/Users/just/code/git/fonttools_just/Lib/fontTools/ttLib/ttFont.py", line 348, in importXML
    reader.read()
  File "/Users/just/code/git/fonttools_just/Lib/fontTools/misc/xmlReader.py", line 47, in read
    self._parseFile(self.file)
  File "/Users/just/code/git/fonttools_just/Lib/fontTools/misc/xmlReader.py", line 72, in _parseFile
    parser.Parse(chunk, 0)
  File "/Users/sysadmin/build/v3.7.6/Modules/pyexpat.c", line 471, in EndElement
  File "/Users/just/code/git/fonttools_just/Lib/fontTools/misc/xmlReader.py", line 155, in _endElementHandler
    self.currentTable.fromXML(name, attrs, content, self.ttFont)
  File "/Users/just/code/git/fonttools_just/Lib/fontTools/ttLib/tables/D_S_I_G_.py", line 96, in fromXML
    sigrec.fromXML(name, attrs, content, ttFont)
[...]

PR #3115, opened 2023-05-22
xmlReader: join consecutive text data that had been cut by buffered xml parser

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/2614

First I add a test as reproducer and then follow up with the fix

### DIFF
--- a/Lib/fontTools/misc/xmlReader.py
+++ b/Lib/fontTools/misc/xmlReader.py
@@ -148,7 +148,19 @@ class XMLReader(object):
 
     def _characterDataHandler(self, data):
         if self.stackSize > 1:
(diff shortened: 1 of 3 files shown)


anthrotype:
reproduced [here](https://github.com/fonttools/fonttools/actions/runs/5045127834/jobs/9049008644?pr=3115#step:5:3503)

```
  File "/opt/hostedtoolcache/Python/3.8.16/x64/lib/python3.8/base64.py", line 87, in b64decode
    return binascii.a2b_base64(s)
binascii.Error: Incorrect padding
```

justvanrossum (approved):

PR #3118, opened 2023-05-22
[subset] Make sure GDEF is processed before GSUB and GPOS

khaledhosny:
See https://github.com/fonttools/fonttools/pull/3114#pullrequestreview-1437266522


behdad:
LGTM. Thanks.

PR #3119, opened 2023-05-23
fontBuilder: add glyphDataFormat=0; error with accidentally cubic outlines

### LINKED ISSUE #3113: Make the upcoming glyf table with cubics opt-in, behind a flag
Now that a glyf table with support for cubic curves is on the roadmap for OpenType 2.0 and that fontTools already supports that output, it can happen that such a glyf table gets produced inadvertently and results in a font that doesn't work except in bleeding-edge environments that run the latest Harfbuzz.

Previously, trying to introduce cubic béziers into the glyf table would crash, and I think it should keep crashing by default for the time being, until at least the spec of the new table format is officially released and implementers can catch up. The new format could be behind a flag, either on fontTools using the config system, or specifically on ttGlyphPens as suggested by @anthrotype . 

We'll look into reproducing this issue and possibly propose a PR.

anthrotype:
Fixes #3113

### DIFF
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -648,7 +648,7 @@ class FontBuilder(object):
         for fontDict in topDict.FDArray:
             fontDict.Private.vstore = vstore
 
-    def setupGlyf(self, glyphs, calcGlyphBounds=True, checkFormat=True):
+    def setupGlyf(self, glyphs, calcGlyphBounds=True, validateGlyphFormat=True):
         """Create the `glyf` table from a dict, that maps glyph names
         to `fontTools.ttLib.tables._g_l_y_f.Glyph` objects, for example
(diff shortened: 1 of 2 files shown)


anthrotype:
> I don't quite understand the use for the checkFormat argument.

Initially I didn't have that, but then I thought that checking _all_ glyf glyph flags may be relatively expensive operation to be doing all the time and if a client knows that all the glyphs that are passed to the fontBuilder have already been cu2qu'ed beforehand.. it's just a waste of time.

justvanrossum:
> Initially I didn't have that, but then I thought that checking all glyf glyph flags may be relatively expensive operation to be doing all the time and if a client knows that all the glyphs that are passed to the fontBuilder have already been cu2qu'ed beforehand.. it's just a waste of time.

Ahh, that makes a lot of sense. Maybe reword the doc string to make it clearer that that is what it's for? And/or consider calling it `validateGlyphFormat`? (Or anything else with the word `validate` in it.)

Issue #3116, opened 2023-05-22
building native binary wheels for fonttools and publishing them to PyPI


anthrotype:
FontTools contains a bunch of cython-accelerated modules (notably fontTools.cu2qu, plus others) that are only available when compiling the package from source (git clone or sdist) within an environment where cython and a compiler toolchain are available.

Ideally we'd build and publish these pre-compiled native binary wheel packages to PyPI, alongside the existing pure python wheel, so that when one `pip install fonttools` and a wheel is available for the target architecture, pip will download and install that automatically without the need to compile from source.

Right now fontmake/ufo2ft still depend on the standalone cu2qu for this exact reason, because for the old project we have set up and published binary wheels, whereas we have not yet for the new "upstream" copy that's in fontTools.cu2qu. The latter has received some additions lately (e.g a new `all_quadratic=True` parameter that can be set to False to let cubics in when advantageous) which have not been backported to the standalone cu2qu (which we did deprecate actually, though the repo isn't archived yet).

Anyway, I propose we set up a new repository at https://github.com/fonttools/fonttools-wheels where we pull in fonttools as a submodule and configure the Github Actions CI to build and publish the native wheels immediately after we cut new releases.
This way we do not make the current fonttools CI slower by adding a lot of builds for each combination of python minor version/ platform / architectures we'd like to support.

Ideally Cython would support Py_LIMITED_API so we can build a single wheel per platform, not one for each python minor version we support, but support for that is still not stable in cython (they are working on it but is buggy/experimenal and can't rely on that yet).

I'll do the work, just filing this in case others have different/better opinions.

anthrotype:
Looks like it worked! 
https://pypi.org/project/fonttools/4.39.4/#files
https://github.com/fonttools/fonttools-wheels/actions/runs/5060028329

Let me know if you find any issues installing these binary wheels.

Note that if one really prefers not to install the precompiled binary wheel, one can always request the sdist with --no-binary option: e.g.

`pip install --no-binary fonttools fonttools`

The pure wheel is still on PyPI of course (e.g. pypy will download that one for I made no specific wheels for it, not worth it).

anthrotype:
I had to delete the wheel files I uploaded yesterday for the current PyPI release, since a user pointed out that the hashes for this release changed and this doesn't look good: https://github.com/fonttools/fonttools/issues/3120
[...]

PR #3121, opened 2023-05-24
Add workflow config file for building native binary wheels

anthrotype:
copied from https://github.com/fonttools/fonttools-wheels/blob/main/.github/workflows/ci.yml

Part of https://github.com/fonttools/fonttools/issues/3116

Thanks @RickyDaMa for the tip about `workflow_dispatch` (to enable it to be run manually), let's see if it works

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -12,7 +12,6 @@ on:
 
 env:
   PROJECT_NAME: fonttools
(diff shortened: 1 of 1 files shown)


anthrotype:
I think the other `publish.yml` worflow should be folded into this one. I'll do that after lunch.
I might have to just merge to main before Github Actions realizes there are new workflows to be run.

Issue #3049, opened 2023-03-18
[designspace] Representing avar2 mappings in designspace documents


Lorp:
The [avar2 spec](https://github.com/harfbuzz/boring-expansion-spec/blob/main/avar2.md) is defined, the need for the new format has been demonstrated with various use cases [[video](https://www.youtube.com/watch?v=j7unMVZOcaw) / [slides](https://docs.google.com/presentation/d/1jLiUvh4PoYz2sw0T0Fygj2HoMrBhmJQuGgAmoMc_Sns/edit#slide=id.p
)] including much smaller binary size for usable parametric fonts, and a few [fonts are being tested](https://lorp.github.io/samsa/src/samsa-avar2.html). However it remains difficult to make avar2 fonts since there is no high-level method to generate an avar2 table.

**The following draft spec represents avar2 mappings in .designspace documents — when implemented it will enable avar2 fonts to be built much more easily.**

* [**Proposal for representing avar2 mappings in designspace documents**](https://github.com/harfbuzz/boring-expansion-spec/blob/main/avar2-in-designspace.md)

Reviews appreciated!

anthrotype:
Thank you Laurence! I finally read that document, and here's my initial comments.

> The avar2 use-case classes already identified (distortion, parametric, HOI, fences) are not representable in the “sources/masters” style. We need to allow the min, peak, max style of variation region specification.

could you elaborate on this a bit, or make an example? it's not immediately clear why, at a first read. I mean, surely having access to "full-power OT variation math" might be desirable per se, but because it diverges so much from the DesignSpace's current approach for the other variable data, I think we should clarify why the latter doesn't work for avar2 and the latter requires more low level control on the region definitions. I'm not arguing that that's not true, just would like to understand better why this is the case.

> Represent avar2 in absolute user coordinates

"user" coordinates mean those from fvar. But isn't avar2 supposed to happen _after_ avar1, by which (normalized) user coordinates are turned into (normalized) internal coordinates? So, if avar1 mappings are present, the input to avar2 cannot be considered to be in user coordinates any more, I believe. I think both the input and output of avar2 should be considered to be in internal, design-space coordinates, as mapped by avar1 mappings if present; if there aren't any, then user == design coordinates so that distintion doesn't matter.

> The <input> element contains one or more <dimension> elements and thereby defines a variation region

[...]

Issue #3086, opened 2023-04-21
[instancer] Update for avar2


behdad:

behdad:
This might be hard.

anthrotype:
can you elaborate why it might be hard?

justvanrossum:
Here is an idea (that may not be a _general_ solution): I have found that is possible to recreate the .designspace-style mappings from avar2 (at least as long avar2 was built from such mappings). From there it should be possible to slice/instance the mappings, similarly how outline data is sliced/instantiated, and then rebuild avar2.

behdad:
> Here is an idea (that may not be a _general_ solution): I have found (*) that it is possible to recreate the .designspace-style mappings from avar2 (at least as long as avar2 was built from such mappings).

@Lorp and I were just talking about this. Here's how the reverse conversion can fail even if it was generated using varLib.models:

Here's the mappings:
- wght=0 -> wgt2=0
- wght=.5 -> wgt2=0
- wght=1. -> wgt2=1.

Here there are two tents, at wght=.5 and wght=1.; however, since the delta at wght=.5 is zero, it won't be encoded in the ItemVariationStore. So your code won't distinguish it from:
- wght=0 -> wgt2=0
- wght=1. -> wgt2=1.

We need to consider the min/max of the tents as well somehow. And then filter out the unnecessary masters? I'll think more about it.

justvanrossum:
Ah, good catch. Looking at my code, it will be relatively easy to take the tent min/max into account. It indeed currently only looks at [PeakCoord](https://github.com/googlefonts/fontra/blob/574354d09feaf1b3d6c503c4ec79c5436a8f0429/src/fontra/backends/opentype.py#L182).

behdad:
> it will be relatively easy to take the tent min/max into account.

Right. The question then is do we get too many extra mappings? But it should at least work.

Lorp:
Will it work to infer 2^n+1 mappings from every region encoded in the IVS? (where n=number of dimensions with delta!=0, thus encoding all of the region’s corners, and the +1 is for the peak). Then we remove duplicates.

Lorp:
Apologies for multiple edits of above. I think I’m done :)

PR #3123, opened 2023-05-24
[designspace] Add avar2 mapping support

behdad:
This just uses list of list of dictionaries. Should use proper objects.

https://github.com/fonttools/fonttools/issues/3049


behdad:
cc @justvanrossum Opening this such that we can work on it together.

behdad:
I updated this to actually build an avar2 table, and successfully built one.

Lorp:
Is this based on my proposed <map> syntax?

behdad:
> Is this based on my proposed `<map>` syntax?

Yes. Just using `<mapping>` instead of `<map>`.

Lorp:
Are you dealing with supporting masters, thus only calculating duovar deltas after calculating supporting monovar deltas etc?

behdad:
> Are you dealing with supporting masters, thus only calculating duovar deltas after calculating supporting monovar deltas etc?

It's a full VariationModel solver.

behdad:
Based on this experiment, I suggest this way forward:
* If an output axis is specified with `xvalue`, then it *must* be also specified as an input axis,
* Otherwise, and output axis can have a `delta` attribute, that has a value in *normalized* (-1..+1) range to be added to the axis value.

I think this makes most sense. I've willing to hear alternatives.

Lorp:
That sounds like an excellent refinement.

Can you confirm your avar2 builder is happy crossing O? (If specified absolutely, the delta must be worked out in two steps.)

behdad:
> That sounds like an excellent refinement.
> 
> Can you confirm your avar2 builder is happy crossing 0? (If specified absolutely, such a delta must be worked out in two steps.)

It looks like it's happy to. In fact, the varLib modeler requires a region at 0.

davelab6:
>> If an output axis is specified with `xvalue`, then it must be also specified as an input axis

I am not so familiar with the implemention, and was curious about the name of this variable; I pinged Behdad to ask, what is the x in xvalue, and he kindly explained, 

> it comes from mutatorMath, which supports anisotropic interpolation, so there's `xvalue` and `yvalue`; and if yvalue is omitted, it's copied from xvalue. But in varfonts we only support isotropic interpolation, so you always only see xvalue

behdad:
cc @LettError @justvanrossum

Lorp:
Recall that there are valid concerns from @anthrotype @justvanrossum @belluzj about the terminology in the spec, so we should not regard it as firm. In particular I’m happy to avoid use of "xvalue". See #3049.

LettError:
"xvalue" is there to record anisotropic locations. This is perhaps not very useful for variable font but it is where this specification started. As long as anisotropy remains possible I'm fine with any other name.

Lib/fontTools/designspaceLib/__init__.py
```diff
@@ -2547,6 +2565,9 @@ def __init__(self, readerClass=None, writerClass=None):
 
         self.axes: List[Union[AxisDescriptor, DiscreteAxisDescriptor]] = []
         """List of this document's axes."""
+
+        self.mapping = []
```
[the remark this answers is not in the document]
behdad: Right.

behdad:
> Based on this experiment, I suggest this way forward:
> 
> * If an output axis is specified with `xvalue`, then it _must_ be also specified as an input axis,
> * Otherwise, an output axis can have a `delta` attribute, that has a value in _normalized_ (-1..+1) range to be added to the axis value.

This makes it ugly to represent in the code. Currently how it works, is that if an axis is only specified in output, then it will work as a delta from the axis's default value. Maybe we just document it that way and go home? @Lorp

behdad:
@Lorp Can you work on a patch to the designspace spec?

behdad:
@Lorp Instead of `<map><region>` what do you think of `<mappings><mapping>`?

behdad:
> @Lorp Instead of `<map><region>` what do you think of `<mappings><mapping>`?

What's what I'm using in the `designspaceLib` code.

Lorp:
Sounds good. We don’t talk about regions anywhere else in designspace. I’m mostly away for the weekend, but will do it soon.

behdad:
Should the dimension element have axis name instead of tag?

behdad:
> Should the dimension element have axis name instead of tag?

Done.

behdad:
This is ready except for documentation.

justvanrossum:
> Currently how it works, is that if an axis is only specified in output, then it will work as a delta from the axis's default value.

Does this mean this is mostly (only?) useful for hidden axes? The situation sounds like that to me: public axes A and B completely determine the value of hidden axis C.

justvanrossum:
The following could perhaps be enforced for non-hidden axes only?

> If an output axis is specified with xvalue, then it _must_ be also specified as an input axis.

(I'm trying to imagine a use case for not specifying the input value that does not involve a hidden axis and I am failing so far.)

Lib/fontTools/varLib/__init__.py
```diff
@@ -208,6 +212,51 @@ def _add_avar(font, axes):
...
+        interesting = True
+
+        derived = [
+            {
+                axes[name].tag: models.normalizeValue(v, vals_triples[axes[name].tag])
+                for name, v in mapping.inputLocation.items()
+            }
+            for mapping in mappings
+        ]
+        source = [
```
anthrotype, line 225: i'm a bit confused by the names of the local variables `derived` and `source`, respectively associated with `inputLocation` and `outputLocation`.. (i'd have expected input being a "source" and output something "derived"). Could you clarify?

behdad: I'll rename to `derived axes` and `source axes`, or just inputaxes and outputaxes.

behdad:
> > Currently how it works, is that if an axis is only specified in output, then it will work as a delta from the axis's default value.

> Does this mean this is mostly (only?) useful for hidden axes? The situation sounds like that to me: public axes A and B completely determine the value of hidden axis C.

Correct.

Lib/fontTools/varLib/__init__.py
anthrotype: Ok but I still don't get our input is "derived" and output is "source"

behdad: Think of a 'wght' axis that is derived from the parametric axes. The 'wght' is input and it drives the parametric axes. But it's really the derived axis.

behdad:
> The following could perhaps be enforced for non-hidden axes only?
> 
> > If an output axis is specified with xvalue, then it _must_ be also specified as an input axis.

Possibly. I just don't like adding unnecessary constraints. Maybe a warning.

behdad:
> > If an output axis is specified with xvalue, then it _must_ be also specified as an input axis.
> 
> (I'm trying to imagine a use case for not specifying the input value that does not involve a hidden axis and I am failing so far.)

Well. A very crude idea is: "if opsz is reduced, reduce the weight slightly"...

khaledhosny:
> > > If an output axis is specified with xvalue, then it _must_ be also specified as an input axis.
> > 
> > 
> > (I'm trying to imagine a use case for not specifying the input value that does not involve a hidden axis and I am failing so far.)
> 
> Well. A very crude idea is: "if opsz is reduced, increase the weight slightly"...

Or rather increase the weight and width slightly.

behdad:
How does the designspace spec versioning work? What would be the new version with this addition?

justvanrossum:
> Well. A very crude idea is: "if opsz is reduced, increase the weight slightly"...

I'm not very excited about this implicit delta-behavior. I may be in favor of requiring an explicit input value for wght then.

Alternatively: if the input is not given, then the output determines the absolute value for that axis _regardless_ of input.

Put another way, I am not convinced that "adding 50 to wght regardless of what wght currently is" is going to be so common that it warrants this rather confusing behavior.

behdad:
> I'm happy to warn if the axis is not hidden.

Done.

Lorp:
> enforced for non-hidden axes only?

Behdad’s use case for a simple opsz axis makes sense. The case is made in [Roboto Flex documentation](https://material.io/blog/roboto-flex) for "fine tuning" using parametric axes. I don’t see that axes intended for fine tuning should be forced to be hidden. IMO the hidden axis flag should have zero consequences beyond being hidden in end user UI.

behdad:
> > enforced for non-hidden axes only?
> 
> Behdad’s use case for a simple opsz axis makes sense. The case is made in [Roboto Flex documentation](https://material.io/blog/roboto-flex) for "fine tuning" using parametric axes. I don’t see that axes intended for fine tuning should be forced to be hidden. IMO the hidden axis flag should have zero consequences beyond being hidden in end user UI.

If the user specifies input axis at default, the resulting font is exactly the same. That's because of how OT variations work...

justvanrossum:
> > Alternatively: if the input is not given, then the output determines the absolute value for that axis regardless of input.

> We discussed this. It is not obvious how to normalize such number though, and it would look out of place. 

I may misunderstand or have misread, but I don't see this exact proposal discussed in this issue. I'm also not sure what you mean by "it would look out of place".

In designspace axis mappings, I would like _all output values to be absolute values_, as absolute master values in a variation model. 

I would like this to resemble outline interpolation (in designspace/UFO) as closely as possible, so we end up with a model that people already understand. In my mind, it can work like this:

- The designer defines a set of locations, akin to drawing a set of master outlines
- Like an outline master, a "location" master has a location
- Now we have two locations: the input location and the output (master) location
- The variation model solver turns this into deltas suitable for the avar2 format

While this wouldn't _directly_ allow for "add N to wght and M to wdth at opsz X", it is similar to how you would implement a "ascender height" axis with designspace/UFO: you provide a new drawing with only some points moved up or down, while other points stay put. An editor could probably provide improved ergonomics, but the format doesn't need unintuitive exceptions for this to work.

Lorp:
Designspace spec patch here (rename `<map>` to `<mappings>`, `<region>` to `<mapping>`, remove references to minimum and maximum values (we leave this to the VariationModel solver), rewrite introduction)

https://github.com/harfbuzz/boring-expansion-spec/pull/92

behdad:
> Designspace spec patch here (rename `<map>` to `<mappings>`, `<region>` to `<mapping>`, remove references to minimum and maximum values (we leave this to the VariationModel solver), rewrite introduction)
> 
> [harfbuzz/boring-expansion-spec#92](https://github.com/harfbuzz/boring-expansion-spec/pull/92)

Thanks. We still need to update the upstream spec:
https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html

That's the only part of this PR left to do. Can you help with that?

behdad:
I assume updating the doc is a prerequisite to merging this PR. I'll see if I can work on it.

In the mean time, yes, would be great to build a few fonts...

behdad:
> Thanks. We still need to update the upstream spec:
> https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html

I did a first pass at this.

behdad:
> In designspace axis mappings, I would like _all output values to be absolute values_, as absolute master values in a variation model.

They are currently all absolute values, and I like it that way as well.

behdad:
This is ready for a full review now I think. @justvanrossum @anthrotype

behdad:
I chose the version 5.1 in the documentation for the new elements. Is there anywhere else the version needs to be bumped or handled?

behdad:
> I chose the version 5.1 in the documentation for the new elements. Is there anywhere else the version needs to be bumped or handled?

Adjusted when writing out document.

behdad:
Splitting axisMappings is left as TODO. I'm a bit unsure how to do that currently.

behdad:
> Splitting axisMappings is left as TODO. I'm a bit unsure how to do that currently.

I took a stab at it.

justvanrossum:
> > In designspace axis mappings, I would like all output values to be absolute values, as absolute master values in a variation model.

> They are currently all absolute values, and I like it that way as well.

But not if there's no input value for a given axis, right? Here you fall back to zero to compute the delta:

https://github.com/fonttools/fonttools/blob/77c00719ba5ddc96f9fb4e1dba905b059183c935/Lib/fontTools/varLib/__init__.py#L256

It is this behavior that I am not comfortable with. Either we should ensure the axis does have an input value, or the output should not be an implicit delta but an absolute value, regardless of the input. If the latter is hard to do, then let's require the input value.

behdad:
> But not if there's no input value for a given axis, right? Here you fall back to zero to compute the delta:
> 
> https://github.com/fonttools/fonttools/blob/77c00719ba5ddc96f9fb4e1dba905b059183c935/Lib/fontTools/varLib/__init__.py#L256

I like to point out that the 0 is in the normalized space, which means the axis default value. So, this is the same as specifying the input axis as its default. So, it's the default value for an axis being omitted, which is normal behavior for axes. If you add a declaration for the input axis at its default location, the behavior would not change as far as I can think of.

justvanrossum:
> I like to point out that the 0 is in the normalized space, which means the axis default value. So, this is the same as specifying the input axis as its default. So, it's the default value for an axis being omitted, which is normal behavior for axes. If you add a declaration for the input axis at its default location, the behavior would not change as far as I can think of.

Ok thanks, I think I get what you mean now. I'll try to update my mental model :)

Lib/fontTools/designspaceLib/split.py
```diff
@@ -225,6 +226,20 @@ def maybeExpandDesignLocation(object):
                 )
             )
 
+    # TODO Trim out-of-range values? :-(
+    subDoc.axisMappings = mappings = []
+    subDocAxes = [axis.name for axis in subDoc.axes]
```
anthrotype, line 231: maybe use a set instead of list for subDocAxes

Lib/fontTools/designspaceLib/split.py
```diff
@@ -225,6 +226,20 @@ def maybeExpandDesignLocation(object):
...
+    for mapping in doc.axisMappings:
+        if not all(axis in subDocAxes for axis in mapping.inputLocation.keys()):
+            continue
+        assert all(axis in subDocAxes for axis in mapping.outputLocation.keys())
```
anthrotype, line 235: assert here means this should never trigger. IIUC the assumption is that if all the axis in the inputLocation are in this subDoc, then also all those in the outputLocation must be. If it's conceivable that this assertion may trigger because of result of user error, then we should provide a better error message than plain assertion. If this is simply checking inner code logic and can never trigger, then i'm fine with leaving as such.

Lib/fontTools/varLib/__init__.py
```diff
@@ -208,6 +212,62 @@ def _add_avar(font, axes):
...
+                axes[name].tag: models.normalizeValue(v, vals_triples[axes[name].tag])
+                for name, v in mapping.outputLocation.items()
+            }
+            for mapping in mappings
+        ]
+
+        # If base-master is missing, insert it at zero location.
+        if not any(all(v == 0 for k, v in loc.items()) for loc in inputLocations):
+            inputLocations.insert(0, {})
+            outputLocations.insert(0, {})
```
anthrotype, line 238: you're only checking if base master is missing from inputLocations but then add to outputLocations unconditionally, should you check each one independently?

Lib/fontTools/varLib/__init__.py
```diff
@@ -208,6 +212,62 @@ def _add_avar(font, axes):
...
+
+        model = models.VariationModel(inputLocations, axisTags)
+        storeBuilder = varStore.OnlineVarStoreBuilder(axisTags)
+        storeBuilder.setModel(model)
+        varIdxes = {}
+        for t in axisTags:
```
anthrotype, line 244: let's call this `t` just tag for easier reading

anthrotype: i'm ok to use single-letter variables on a one-liner generator expression or list comprehension, but here it's used on multiple lines and need to remind myself what it means

Lib/fontTools/varLib/__init__.py
```diff
@@ -208,6 +212,62 @@ def _add_avar(font, axes):
...
+            masterValues = []
+            for vo, vi in zip(outputLocations, inputLocations):
```
anthrotype, line 246: should we assert that outputLocations and inputLocations have the same length?

Lib/fontTools/varLib/__init__.py
```diff
@@ -208,6 +212,62 @@ def _add_avar(font, axes):
...
+                if t not in vo:
+                    masterValues.append(0)
+                    continue
+                if t not in vi and t not in hiddenAxes:
+                    log.warning(
+                        "No input location specified for non-hidden axis '%s' in axis mapping %s",
```
anthrotype, line 252: maybe the warning could mention what's the consequence of not specifying an input location for a non-hidden axis

Lib/fontTools/varLib/__init__.py
```diff
@@ -856,6 +907,9 @@ def load_designspace(designspace):
         axes[axis_name] = axis
     log.info("Axes:\n%s", pformat([axis.asdict() for axis in axes.values()]))
 
+    axisMappings = ds.axisMappings
+    log.info("Mapping:\n%s", pformat(axisMappings))
```
anthrotype, line 911: maybe only log that if axisMappings is non-empty

Tests/designspaceLib/data/test_avar2.designspace
```diff
@@ -0,0 +1,1481 @@
...
+    <axis tag="wdth" name="Width" minimum="62.5" maximum="100" default="100">
+      <map input="62.5" output="70"/>
+      <map input="75" output="79"/>
+      <map input="87.5" output="89"/>
+      <map input="100" output="100"/>
+    </axis>
+    <mappings>
+      <mapping>
+        <input>
+	  <dimension name="Justify" xvalue="0"/>
```
anthrotype, line 25: even in this test_avar2.designspace file I see that non-hidden axes are omitted from the input location, so it would trigger the warning above, see https://github.com/fonttools/fonttools/pull/3123/files#r1212966176

if this is normal thing to do, I would just remove the warning, turn it into info or debug

Tests/designspaceLib/data/test_avar2.designspace
```diff
@@ -0,0 +1,1481 @@
...
+        <input>
+	  <dimension name="Justify" xvalue="-100"/>
+        </input>
+        <output>
+	  <dimension name="Weight" xvalue="26"/>
+        </output>
+      </mapping>
+      <mapping>
+        <input>
+	  <dimension name="Justify" xvalue="+100"/>
```
anthrotype, line 41: the "+" symbol is more like an embellishment here, I happens to work because we eval this as python float but I would rather stick to plain digits

Tests/designspaceLib/data/test_avar2.designspace
```diff
@@ -0,0 +1,1481 @@
...
+    </source>
+  </sources>
+  <instances>
+    <instance name="Noto Sans Arabic Thin" familyname="Noto Sans Arabic" stylename="Thin" filename="../instance_ufo/NotoSansArabic-Thin.ufo" stylemapfamilyname="Noto Sans Arabic Thin" stylemapstylename="regular">
+      <location>
+        <dimension name="Weight" xvalue="26"/>
+        <dimension name="Width" xvalue="100"/>
+      </location>
+      <lib>
+        <dict>
```
anthrotype, line 110: can we get rid of this unused lib stuff?

Tests/designspaceLib/data/test_avar2.designspace
```diff
@@ -0,0 +1,1481 @@
...
+            <array>
+              <string>styleMapFamilyName</string>
+              <string>Noto Sans Arabic ExtCond Blk</string>
+            </array>
+          </array>
+        </dict>
+      </lib>
+    </instance>
+  </instances>
+  <lib>
```
anthrotype, line 1393: same here, we can remove this whole lib

Tests/designspaceLib/data/test_avar2.designspace
```diff
@@ -0,0 +1,1481 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<designspace format="5.0">
+  <axes>
+    <axis tag="jstf" name="Justify" minimum="-100" maximum="100" default="0"/>
```
anthrotype, line 4: where is this lowercase "jstf" axis (impliing an official registered axis) defined? I don't see it in 
https://learn.microsoft.com/en-us/typography/opentype/spec/dvaraxisreg#registered-axis-tags

suggest making this uppercase JSTF until it's officially defined.

By the way, why a justification axis to influence the weight? Is that a realistic scenario -- just curious

Tests/designspaceLib/designspace_test.py
```diff
@@ -688,6 +689,37 @@ def test_axisMapping():
     assert r == [("ddd", [(0, 0.0), (300, 0.5), (600, 0.5), (1000, 1.0)])]
 
 
+def test_axisMappingsRoundtrip(tmpdir):
+    # tests of axisMappings in a document, roundtripping.
+    import pathlib
```
anthrotype, line 694: feel free to `from pathlib import Path` at the top of the module

Tests/designspaceLib/split_test.py
```diff
@@ -211,3 +211,10 @@ def test_optional_min_max_internal(condition, expected_set: ConditionSet):
     """Check that split's internal helper functions produce the correct output
     for conditions that are partially unbounded."""
     assert _conditionSetFrom([condition]) == expected_set
+
+
+def test_avar2(datadir):
+    ds = DesignSpaceDocument()
+    ds.read(datadir / "test_avar2.designspace")
+    out = list(splitInterpolable(ds))[0][1]
```
anthrotype, line 219: this test is a bit cryptic, what's that `[0][1]` indices, what is `out`?  maybe use self-explanatory variables or add comment to avoid one having to go read the source code to unsterstand what's going on

anthrotype: it's also not exercising the splitInterpolable method much, as the test designspace contains a single space sub-space/VF setup so you're just asserting that it passes through. Better than nothing of course

Lib/fontTools/designspaceLib/split.py
behdad: I think an error is more appropriate indeed.  I'm not very familiar with how axes are chosen for the split to know if it can happen or not. Help needed.

Lib/fontTools/varLib/__init__.py
behdad: The two are parallel lists at this point. Ie. the outputLocations is the value corresponding to the input location. So when we insert in one we have to insert into other. Since there's no data for the default master available, the value would be empty dictionary.

Lib/fontTools/varLib/__init__.py
behdad: They have to. They're both coming out of `mappings` list earlier. I can add an assert though, yes.

Lib/fontTools/varLib/__init__.py
behdad: Done. The 0 means axis default value.

Tests/designspaceLib/data/test_avar2.designspace
behdad: We're still a bit unsure how this would pan out. I'll update the test.

Tests/designspaceLib/data/test_avar2.designspace
anthrotype: +1 width

Lib/fontTools/designspaceLib/split.py
behdad: Reading the calls to the split function, it looks like all non-discrete axes are always included in the split. So indeed an assert should be fine. Am I right @belluzj?

The only issue outstanding is to limit the range if the input range is limited. That might be tricky though as in https://github.com/fonttools/fonttools/issues/3086

behdad: > Reading the calls to the split function, it looks like all non-discrete axes are always included in the split.

I'm wrong I think. Adding a test and making it an error.

behdad:
@anthrotype Thanks for the review. I think I addressed them all.

behdad:
@khaledhosny Is this something you can use in your font to build the avar2 table?

behdad:
> @khaledhosny Is this something you can use in your font to build the avar2 table?

I guess not if you're going from Glyphs to varLib directly.

khaledhosny:
> @khaledhosny Is this something you can use in your font to build the avar2 table?

Probably. I create a DS in code and pass it to varLib, so I can add the mapping there.

khaledhosny:
This fixes the warnings:
```diff
diff --git a/scripts/build.py b/scripts/build.py
index 8d19a4a..63d4531 100644
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -726,11 +726,11 @@ def build(font, instance, args):
         ds.addSource(source)

     inputs = [
-        {"Justification": 0},
-        {"Justification": 90},
-        {"Justification": 100},
-        {"Justification": -50},
-        {"Justification": -100},
+        {"Justification": 0, "Spacing": 0, "Mashq": 10},
+        {"Justification": 90, "Spacing": 0, "Mashq": 10},
+        {"Justification": 100, "Spacing": 0, "Mashq": 10},
+        {"Justification": -50, "Spacing": 0, "Mashq": 10},
+        {"Justification": -100, "Spacing": 0, "Mashq": 10},
     ]

     outputs = [
```

But it feels redundant. Now I need to remember to update these if I ever change the default location of any of the axes.

khaledhosny:
Also I’m seeing this ttx diff compared to my code:
```diff
   <avar>
     <version major="2" minor="0"/>
     <segment axis="SPAC">
+      <mapping from="-1.0" to="-1.0"/>
+      <mapping from="0.0" to="0.0"/>
+      <mapping from="1.0" to="1.0"/>
     </segment>
     <segment axis="MSHQ">
+      <mapping from="-1.0" to="-1.0"/>
+      <mapping from="0.0" to="0.0"/>
+      <mapping from="1.0" to="1.0"/>
     </segment>
     <segment axis="jstf">
+      <mapping from="-1.0" to="-1.0"/>
+      <mapping from="0.0" to="0.0"/>
+      <mapping from="1.0" to="1.0"/>
     </segment>
```

Probably harmless but also redundant.

behdad:
> But I don’t understand these warnings, what am I doing wrong?

Make the `SPAC` axis hidden and they should go away.

behdad:
> This fixes the warnings:
> 
> ```diff
> diff --git a/scripts/build.py b/scripts/build.py
> index 8d19a4a..63d4531 100644
> --- a/scripts/build.py
> +++ b/scripts/build.py
> @@ -726,11 +726,11 @@ def build(font, instance, args):
>          ds.addSource(source)
> 
>      inputs = [
> -        {"Justification": 0},
> -        {"Justification": 90},
> -        {"Justification": 100},
> -        {"Justification": -50},
> -        {"Justification": -100},
> +        {"Justification": 0, "Spacing": 0, "Mashq": 10},
> +        {"Justification": 90, "Spacing": 0, "Mashq": 10},
> +        {"Justification": 100, "Spacing": 0, "Mashq": 10},
> +        {"Justification": -50, "Spacing": 0, "Mashq": 10},
> +        {"Justification": -100, "Spacing": 0, "Mashq": 10},
>      ]
> 
>      outputs = [
> ```
> 
> But it feels redundant. Now I need to remember to update these if I ever change the default location of any of the axes.

Yeah it's one of the issues we've been discussing. Currently we warn if a non-hidden output axis is not specified.

khaledhosny:
> > But I don’t understand these warnings, what am I doing wrong?
> 
> Make the `SPAC` axis hidden and they should go away.

But I want it the other way around, `SPAC` and `MSHQ` visible and `jstf` hidden since it should be use by justification algorithms and not directly by the user, unlike the other two.

behdad:
> > > But I don’t understand these warnings, what am I doing wrong?
> > 
> > 
> > Make the `SPAC` axis hidden and they should go away.
> 
> But I want it the other way around, `SPAC` and `MSHQ` visible and `jstf` hidden since it should be used by justification algorithms and not directly by the user, unlike the other two.

Interesting. Yeah I think we might want to nuke that warning then.

justvanrossum:
Yes, let's get rid of the warning. But since leaving out the input value and putting it at the default value should behave the same way, I am also surprised at the ttx diff.

khaledhosny:
> But since leaving out the input value and putting it at the default value should behave the same way, I am also surprised at the ttx diff.

Sorry of I didn't make this clear, the ttx diff is not related to this issue but when generating avar with varLib in general (I’m comparing with the code I was using earlier which left the segments empty).

behdad:
Removed warning.

Is this good to merge then?

anthrotype:
@behdad what about that diff that Khaled noticed?

behdad:
> @behdad what about that diff that Khaled noticed?

That's because of this pre-existing code block:
https://github.com/fonttools/fonttools/blob/84cebca6a1709085b920783400ceb1a147d51842/Lib/fontTools/varLib/__init__.py#L149-L156

anthrotype:
Are we all happy to merge this? We can start making avar2 fonts and come back fix things as needed

anthrotype:
@behdad hm that commit you quoted is quite old, from 2017
https://github.com/fonttools/fonttools/commit/04eacf13cde64cfc277b4223f592299be0edb6c9

> I’m comparing with the code I was using earlier which left the segments empty

@khaledhosny does that mean you had not rebuilt your fonts since 2017?

khaledhosny:
> @behdad hm that commit you quoted is quite old, from 2017 [04eacf1](https://github.com/fonttools/fonttools/commit/04eacf13cde64cfc277b4223f592299be0edb6c9)
> 
> > I’m comparing with the code I was using earlier which left the segments empty
> 
> @khaledhosny does that mean you had not rebuilt your fonts since 2017?

I wasn’t using DS to generate avar table.

khaledhosny (approved):

PR #3124, opened 2023-05-24
[varStore] Improve optimize algorithm

behdad:
This was always supposed to be how it works, but was missed in the initial implementation apparently.

I should probably add extensive tests for the optimization algorithm...


behdad:
Saves 408 bytes building NotoSansArabic.

anthrotype:
hm in the absence of a test that touches those lines, it's a bit too hard for me to reivew so I'll pass. feel free to merge, this is your code

behdad:
Adding Jany and Nikolaus in case one of them wants to read and understand the algorithm.  I'm willing to add more comments. And yes I'll think about adding tests.

rsheeter:
The doc comment is most appreciated. +1 to tests.

> Saves 408 bytes building NotoSansArabic.

Are there cases where it saves more?

behdad:
> The doc comment is most appreciated. +1 to tests.
> 
> > Saves 408 bytes building NotoSansArabic.
> 
> Are there cases where it saves more?

In NotoSans it saved 2kb, but that's 0.1%.  For NotoSansArabic it was 0.5%. I don't expect it to be much, but hey, it's free savings.

behdad:
> In NotoSans it saved 2kb

Another bugfix (thanks to new tests), and it saves 4.5kb now.

behdad:
> Saves 408 bytes building NotoSansArabic.

Now at 936 bytes after a few other fixes.

PR #3126, opened 2023-05-24
Var store quantize

behdad:


behdad:
Not sure if we want to set quantization higher than one by default in any fontmake-related codepath.

Lib/fontTools/varLib/varStore.py
```diff
@@ -551,8 +551,16 @@ def VarStore_optimize(self, use_NO_VARIATION_INDEX=True):
...
-                row[regionIdx] += v
+
+            if quantization == 1:
+                for regionIdx, v in zip(regionIndices, item):
+                    row[regionIdx] += v
+            else:
+                for regionIdx, v in zip(regionIndices, item):
+                    row[regionIdx] += (
+                        round(v / quantization) * quantization
+                    )  # TODO: round towards 0
```
anthrotype, line 562: why "round towards 0"?

you can use `decimal.Decimal`'s `quantize` method with a `ROUND_DOWN` rounding method:

https://docs.python.org/3/library/decimal.html#decimal.ROUND_DOWN

though it might be quicker to simply do `floor(v) if v > 0 else ceil(v)`

anthrotype:
> Not sure if we want to set quantization higher than one by default in any fontmake-related codepath.

I don't know, maybe an option in the public varLib.build/build_many methods that ufo2ft uses. But I'm not sure if one wants to use the same quantization for all the VarStores.
What did you have in mind for this? Make GPOS variable kerning tables smaller? Did somebody ask for this?

anthrotype:
I think it's fine to have this option available in the `fonttools varLib.varStore` command-line tool only for now, one can already start experiment with it and if needed we can surface it to varLib's public api

anthrotype (approved):

Lib/fontTools/varLib/varStore.py
behdad: I'll remove the comment. round is fine.

behdad: What I meant was a round that rounds .5 to 0, such that eg. with quantization 2, value 1 will become 0, not 2.

anthrotype: ah that's https://docs.python.org/3/library/decimal.html#decimal.ROUND_HALF_DOWN

anthrotype: yeah i think as a size saving optimization this quantization should make the deltas smaller, would be strange if it made them bigger

behdad:
> I think it's fine to have this option available in the `fonttools varLib.varStore` command-line tool only for now (that one only touches GDEF VarStore as used in GPOS), one can already start experiment with it and if needed we can surface it to varLib's public api

Sounds good. Thanks.

Lib/fontTools/varLib/varStore.py
anthrotype: (though the sharing comes from reuse, not the delta magnitude itself)

PR #3127, opened 2023-05-25
Var store optimize more

behdad:


behdad:
I'm happy with this code now. It was due for a refresh!

Lib/fontTools/varLib/varStore.py
```diff
@@ -570,89 +568,63 @@ def VarStore_optimize(self, use_NO_VARIATION_INDEX=True, quantization=1):
...
-                )
-            )
-            separate_gain = this_gain + other_gain
+    # Repeatedly pick two best encodings to combine, and combine them.
 
-            if combined_gain - separate_gain > best_gain:
-                best_idx = i
-                best_gain = combined_gain - separate_gain
+    heap = []
+    for i, other_encoding in enumerate(todo):
```
cmyr, line 578: i'm confused here, are we ever using `other_encoding`? it looks like we rebind it in the next loop without having touched it. Maybe it was supposed to be `encoding`, and then you wouldn't need to manually bind that, below?

cmyr (approved):
Okay I mostly justed used this as an excuse to familiarize myself with ItemVariationStore.

Out of curiosity, do we have a sense of the savings provided this optimization?

Lib/fontTools/varLib/varStore.py
behdad: You are correct.

behdad:
> Okay I mostly justed used this as an excuse to familiarize myself with ItemVariationStore.
> 
> Out of curiosity, do we have a sense of the savings provided this optimization?

Honestly not much. In the few kilobytes range. For NotoSans-VF it's 10kb.

PR #3129, opened 2023-05-25
privData needs to padded with bytes not str

benkiel:
One letter, but an important one.


behdad:
How did this ever work?!

anthrotype:
ouch. thanks

looks like this wasn't covered by tests. Mind adding a simple one somewhere?

anthrotype:
> How did this ever work?!

it didn't 😅

benkiel:
@anthrotype in the middle of trying to get out a font release, tbh. Will try to come back and make a test later this week

benkiel:
@anthrotype where would be best to put a test? 

I can see either putting it in `main_test.py` for `ttLib` or in `sfnt_test.py` — the first is easier as it already has methods for getting a font object to use, and it tests WOFF a bit already, but the direct place is `sfnt_test.py`. (yes, there could also be `woff_test.py` but that's the most work.

anthrotype:
Where it's more convenient to you, any test is better than nothing thanks

benkiel:
@anthrotype ok, test added

benkiel:
Ok to merge?

anthrotype (approved):
LGTM thanks!

PR #3130, opened 2023-05-26
[feaLib] Support variable ligature caret position

khaledhosny:
Allow variable scaler in ligature caret position and build CaretValueFormat3 with DeviceTable. Does not support non-variable device table, but can be added if someone really really wants it.


Lib/fontTools/otlLib/builder.py
```diff
@@ -2509,9 +2509,14 @@ def buildAttachPoint(points):
 
 def buildCaretValueForCoord(coord):
     # 500 --> otTables.CaretValue, format 1
+    # (500, DeviceTable) --> otTables.CaretValue, format 3
     self = ot.CaretValue()
-    self.Format = 1
-    self.Coordinate = coord
+    if isinstance(coord, tuple):
```
anthrotype, line 2514: instead of optionally taking a 2-tuple, how about adding an additional deviceTable=None parameter, when not None you'd make Format 3 table?

Lib/fontTools/otlLib/builder.py
```diff
@@ -2573,7 +2578,8 @@ def buildLigGlyph(coords, points):
     # ([500], [4]) --> otTables.LigGlyph; None for empty coords/points
     carets = []
     if coords:
-        carets.extend([buildCaretValueForCoord(c) for c in sorted(coords)])
+        coords = sorted(coords, key=lambda c: isinstance(c, tuple) and c[0] or c)
```
anthrotype, line 2581: if c[0] == 0 the `and c[0]` test would evaluate to False and the `or c` would kick in, so the tuple as a whole instead of its first item would involutarily be passed on. I think you meant something like this:

```suggestion
        coords = sorted(coords, key=lambda c: c[0] if isinstance(c, tuple) else c)
```

Lib/fontTools/otlLib/builder.py
khaledhosny: This would only move the optional tuple down to `buildLigGlyph` and make the code their slightly uglier without changing much. `buildLigGlyph` takes a list of coordinates, so if were were to split off the devices we would need to lists and keep them in sync (e.g. when we sort the coordinates).

anthrotype (approved):

Issue #3139, opened 2023-05-29
[instancer, L4] unexpected weight jumps if VF default is repositioned, when VF contains partial/support/sparse sources


arrowtype:
Level 4 of the Instancer is described as the following in the docs:

> L4
> moving the default location of an axis, by specifying (min,defalt,max) values:
> `font = instancer.instantiateVariableFont(varfont, {“wght”: (100, 300, 700)})`

It seems to work very well for most fonts, but I’ve noticed what seems to a case in which it doesn’t quite work as expected.

In my project Name Sans, there are several sparse masters / support sources, for glyphs which need a little extra contrast control in lighter weights. Name Sans has a default source of 700 (Bold), plus full sources at wght=1 and wght=1000. It has sparse sources at weights 100, 200, 500, and 800. I would like to add a default of 400 to the desktop fonts, where it would help the fonts work more intuitively for users (and probably just generally better in some apps). 

Setting this new default seems to work at first... but it has an unintended consequence: there is a weight "jump" at wght=200 in all glyphs that have a support at that location (`G a e n b` and quite a few more). There is also a very subtle jump at wght=460, in glyphs with a support at 500 (there are fewer, but it includes `a e s w`). I can’t see other jumps, so far.

Here’s a video showing before/after:

https://github.com/fonttools/fonttools/assets/45946693/4945a08f-22be-422c-a553-7fe3fa3f3de6

I am using the instancer with this script: https://gist.github.com/arrowtype/9fefe9633cae500bbaf0000230f6a3ed

I’m using FontTools 4.39.4 to instance the font (and to build it, with FontMake).

Here are two variable fonts, one before the instancer, and one after. They have a limited/trial glyphset, so I can share them here, and folks are welcome to use them for testing purposes / inspecting internal font data:
[namesans-variable-trial-fonttools_issue_3139.zip](https://github.com/fonttools/fonttools/files/11594331/namesans-variable-trial-fonttools_issue_3139.zip)

I’m happy to send more sources, etc, if that might be helpful!

behdad:
Oops! I'll take a look. Thanks.

arrowtype:
Well, as usual, I’m not sure whether it’s my error or something in FontTools. But, it seems to only happen after running the instancer, so I think it might be a bug? In any case, thank you so much for taking a look!

behdad:
Fix in https://github.com/fonttools/fonttools/pull/3140
@arrowtype please test.

arrowtype:
[...]

PR #3140, opened 2023-05-29
[instancer-solver] Fix a bug

behdad:
Fixes https://github.com/fonttools/fonttools/issues/3139

Tests to be added.


behdad:
@qxliu76 Please port this fix and test to HarfBuzz. Thanks.

PR #3146, opened 2023-06-01
implied oncurve points for interpolatable glyphs

anthrotype:
we have a drop_implied_oncurves function inside ttGlyphPen, but that being a pen only works for a single glyph.

Thus I modified this to work on one or more (interpolatable) glyphs and to only prune oncurve points that are impliable for _all_ of the input glyphs, such that the modified glyphs maintain interpolability.

I also moved it to the glyf table module. I'd like to use this in ufo2ft to massage the glyphs of master TTFs before it calls varLib.build.

fontc does a similarly eager pruning of oncurve points and I'd like to have this done (optionally) in fontmake-py as well.

### DIFF
--- a/Lib/fontTools/pens/ttGlyphPen.py
+++ b/Lib/fontTools/pens/ttGlyphPen.py
@@ -17,48 +17,57 @@ import math
 __all__ = ["TTGlyphPen", "TTGlyphPointPen"]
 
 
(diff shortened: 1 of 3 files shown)


behdad:
LGTM. Thanks!

behdad (approved):

PR #3147, opened 2023-06-02
[varlib] add --drop-implied-oncurves option

anthrotype:
Follow up from #3146 

I decided to add this to varLib directly instead of ufo2ft, since the method for dropping the oncurve points was already here.

The option is False by default.
Don't expect much savings, mostly in the hundred of bytes for normal fonts so usually not worth rebuilding yours.

I was thinking of adding a tolerance to increase the chances of pruning oncurves (right now it has to be exactly in the middle, we use math.isclose). But maybe another time, I just need this for testing purposes (diffing against fontc)

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -28,6 +28,7 @@ from fontTools.misc import xmlWriter
 from fontTools.misc.filenames import userNameToFileName
 from fontTools.misc.loggingTools import deprecateFunction
 from enum import IntFlag
+from types import SimpleNamespace
 from typing import Set
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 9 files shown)


Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -1549,16 +1551,40 @@ def dropImpliedOnCurvePoints(*interpolatable_glyphs: Glyph) -> Set[int]:
...
+            numContours = glyph.numberOfContours
+        elif glyph.numberOfContours != numContours:
+            raise ValueError(
+                f"Incompatible number of contours for glyph at master index {i}: "
+                f"expected {numContours}, found {glyph.numberOfContours}"
+            )
+
+        if flags is None:
+            flags = glyph.flags
+        elif glyph.flags != flags:
```
behdad, line 1580: Should you also check for same `endPtsOfContours`?

behdad (approved):
Just a minor comment. LGTM otherwise. Thanks.

Lib/fontTools/ttLib/tables/_g_l_y_f.py
anthrotype: Right, it I'd conceivable that numContours and flags be equal but endPtsOfContours are different, will update, thanks

anthrotype: done https://github.com/fonttools/fonttools/pull/3147/commits/5b93100616199945e23cdcd664242b68ad9bf0f8

PR #3156, opened 2023-06-07
Drop impliable oncurves either before or after rounding

anthrotype:
Currently (#3146) we only compute implied oncurves after coordinates have already been rounded. We should compute if an oncurve point is in between previous/next off-curve points both before they get rounded and afterwards, and accept it either way.
The TTGlyphPens need to allow leaving the coordinates un-rounded, so that ufo2ft can set `round=noRound` when building master interpolatable TTFs; then it can send these to varLib.build and have the impied points dropped (when impliable for all masters).

### DIFF
--- a/Lib/fontTools/pens/ttGlyphPen.py
+++ b/Lib/fontTools/pens/ttGlyphPen.py
@@ -1,5 +1,5 @@
 from array import array
-from typing import Any, Dict, Optional, Tuple
+from typing import Any, Callable, Dict, Optional, Tuple
 from fontTools.misc.fixedTools import MAX_F2DOT14, floatToFixedToFloat
 from fontTools.misc.loggingTools import LogMixin
 from fontTools.pens.pointPen import AbstractPointPen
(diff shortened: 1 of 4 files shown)


Lib/fontTools/pens/ttGlyphPen.py
```diff
@@ -127,7 +127,12 @@ def _buildComponents(self, componentFlags):
             components.append(component)
         return components
 
-    def glyph(self, componentFlags: int = 0x04, dropImpliedOnCurves=False) -> Glyph:
+    def glyph(
+        self,
+        componentFlags: int = 0x04,
+        dropImpliedOnCurves: bool = False,
+        roundCoordinates: bool = True,
```
behdad, line 134: We typically just pass a `round=round` where one can pass `round=noRound` from `roundTools`. The code can optimize for `noRound`. I'm fine with what you have though.

anthrotype: ah makes sense, thanks for the suggestion, I'll do that

PR #3162, opened 2023-06-12
Bug fix for `TypeError` in `_reverseBytes()` in module `E_B_D_T_`, currently breaks on `bytes` arguments longer than 1

robhagemans:
In the `E_B_D_T_` module, `_reverseBytes()` with `bytes` arguments longer than 1 lead to a recursive call from a loop. However, a loop over `bytes` produces `int`s, which have no `len()`, leading to a `TypeError`.

Fixed with `isinstance` as it appears the function is intended to be applicable to both `int` and `bytes` arguments.

Contributing guidelines suggest pull request authors should be listed in the README - let me know if I should add my name there, happy to do so but this is a kinda trivial edit.

### DIFF
--- a/Lib/fontTools/ttLib/tables/E_B_D_T_.py
+++ b/Lib/fontTools/ttLib/tables/E_B_D_T_.py
@@ -258,7 +258,7 @@ def _memoize(f):
     class memodict(dict):
         def __missing__(self, key):
             ret = f(key)
-            if len(key) == 1:
+            if isinstance(key, int) or len(key) == 1:
                 self[key] = ret
             return ret
 
--- a/Lib/fontTools/ttLib/tables/E_B_D_T_.py
+++ b/Lib/fontTools/ttLib/tables/E_B_D_T_.py
@@ -271,7 +271,7 @@ def _memoize(f):
 # opposite of what makes sense algorithmically and hence this function.
 @_memoize
 def _reverseBytes(data):
-    if len(data) != 1:
+    if isinstance(data, bytes) and len(data) != 1:
         return bytesjoin(map(_reverseBytes, data))
     byte = byteord(data)
     result = 0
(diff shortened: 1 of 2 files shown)


anthrotype:
thanks! It looks this code isn't much tested otherwise we'd realised by now.
I think this part of the code is triggered when one dumps a font with EBDT table with ttx using -z bitwise option, right?
I just tried with a random bitmap font using this PR branch and I still got a TypeError, this time coming from the `_memoize` decorator. I had to patch it like so:

```diff
diff --git a/Lib/fontTools/ttLib/tables/E_B_D_T_.py b/Lib/fontTools/ttLib/tables/E_B_D_T_.py
index 4527d0769..1ba6f4e77 100644
--- a/Lib/fontTools/ttLib/tables/E_B_D_T_.py
+++ b/Lib/fontTools/ttLib/tables/E_B_D_T_.py
@@ -258,7 +258,7 @@ def _memoize(f):
     class memodict(dict):
         def __missing__(self, key):
             ret = f(key)
-            if len(key) == 1:
+            if isinstance(key, int) or len(key) == 1:
                 self[key] = ret
             return ret

```

> let me know if I should add my name there, happy to do so but this is a kinda trivial edit.

robhagemans:
Ah, I didn't come across that other error. I think I might have disabled the `_memoize` before testing... 

What I was trying to do is construct a bitmap font from scratch (which works, code is here https://github.com/robhagemans/monobit/tree/master/monobit/formats/sfnt if interested. Some of this might be worthwhile as part of the font builder but it would need more work and testing)

anthrotype:
that's a cool project! 
Do you mind incorporating the proposed changes above?
I'm thinking where to best add a test for this since there is currently no unit test module specific for EBDT table. Maybe a doctest in the _reverseBytes method is sufficient for now

robhagemans:
Sure, I'll add the changes to the PR

robhagemans:
I added the change to `_memoize`. Also added a doctest but it seems the `_memoize` decorator prevents it from being detected and run. Not sure how to fix that, I'm not very familiar with doctest.

Issue #3085, opened 2023-04-20
varLib does not check if glyf component flags are the same, should it?


anthrotype:
in the `_GlyphControls` namedtuple returned from `glyf._getCoordinatesAndControls` method, as used by varLib._add_gvar when calculating deltas from master glyf tables, the field for composite glyphs' flags is `None`, so it is not considered when checking that components are interpolation compatible across masters.

https://github.com/fonttools/fonttools/blob/1c283756a5e39d69459eea80ed12792adc4922dd/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L380-L381

given that glyf component flags cannot "vary" (unlike, say, dx,dy translation), I believe the expectation should be that they are identical across all masters; so if any of the master TTFs are such that, e.g., the USE_MY_METRICS flag would _not_ be set, the default instance of the VF should not force the flag on the entire variation space.
I haven't thought about other flags so far.

thougths?

anthrotype:
when ufo2ft sets the USE_MY_METRICS, it does it per single master UFO independently from one another; so it is conceivable that in some master TTFs some components will get the flag while in other masters a corresponding component may not. But fonttools doesn't check if these are the same.
My question is, should it? does it matter?

behdad:
You're right. The components are in a separate entry I think? The dx,dy get encoded in the points I believe.

anthrotype:
> The dx,dy get encoded in the points I believe

yes, among the "coordinates" part returned from _getCoordinatesAndControls

behdad:
It's not just component flags. We don't check that the component transformation, if any, other than translate, is the same either.

behdad:
> It's not just component flags. We don't check that the component transformation, if any, other than translate, is the same either.

Nevermind. I fixed that before:
https://github.com/fonttools/fonttools/blob/1c283756a5e39d69459eea80ed12792adc4922dd/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L402

PR #3163, opened 2023-06-13
varLib: ensure USE_MY_METRICS component flag is set consistently or not at all

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/3085


anthrotype:
maybe this would better done in ufo2ft which is the one that automatically sets these flags to begin with, so closing here. 
I'm also not entirely sure what the exact consequences are of these flags not being set consistently across masters..
Maybe whatever the default master has in terms of USE_MY_METRICS is ok for the entire VF and doesn't really matter what the rest of the non-default masters do...

PR #3164, opened 2023-06-13
Add voltLib.voltToFea module and command line tool

khaledhosny:
A tool for converting VOLT VTP projects to feature files.

Essentially https://github.com/TiroTypeworks/TiroTools/tree/master/Volto with minor modifications to fit into FontTools.


behdad:
Maybe a more descriptive name?

khaledhosny:
I have no particular attachment to the name, but it is the name used over at [https://github.com/TiroTypeworks/TiroTools](https://github.com/TiroTypeworks/TiroTools)

anthrotype:
how about moving it inside voltLib and renaming the module voltToFea, like the class is also called?

khaledhosny:
At one point it was supposed to grow a fea to volt convertor as well, and may be even a volt compiler. This didn’t happen though, and no idea if it ever will.

anthrotype:
We could add more submodules if/when needed. I think having a both a fontTools.voltLib and fontTools.volto is confusing and one "volt" too much for the top level modules.

khaledhosny:
> We could add more submodules if/when needed. I think having a both a fontTools.voltLib and fontTools.volto is confusing and one "volt" too much for the top level modules.

Done.

anthrotype (approved):
LGTM! Many thanks to Khaled and TiroTypeworks for contributing this to fonttools!

Issue #3170, opened 2023-06-19
[macCreatorType] setMacCreatorAndType() clears additional metadata


NSGod:
The `com.apple.FinderInfo` extended attribute contains a full 32 bytes of information that makes up the Mac metadata. If I take an existing `'LWFN'` PostScript Type 1 outline font, and take a look at it with `xattr`, it shows the following:

```
mdouma46@MacBookPro15 Adobe Garamond % xattr -p com.apple.FinderInfo AGarBolCopy
0000   4C 57 46 4E 41 53 50 46 21 0C 00 00 00 00 00 00    LWFNASPF!.......
0010   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00    ................
```
This particular file had a red label color and custom icon set. You'll notice that the first 8 bytes are the file type and creator code, but then there are 2 non-zero bytes afterwards that hold that other information.

If I use the existing `setMacCreatorAndType()` command to set the creator code to `'FNGL'`, for example, it results in the following being shown by `xattr`:

```
mdouma46@MacBookPro15 Adobe Garamond % xattr -p com.apple.FinderInfo AGarBolCopy
0000   4C 57 46 4E 46 4E 47 4C 00 00 00 00 00 00 00 00    LWFNFNGL........
0010   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00    ................
```

As you can see, the extra info after the file type and creator code has been zeroed out.

I'll be submitting a pull request which should resolve this issue.

PR #3171, opened 2023-06-19
[macCreatorType] preserve existing other metadata when setting filetype & creator

NSGod:
This addresses #3170.

I'll just note that I'm not fluent in Python, so I don't know if there'd be a better way to fix this than what I came up with.

### DIFF
--- a/Lib/fontTools/misc/macCreatorType.py
+++ b/Lib/fontTools/misc/macCreatorType.py
@@ -52,5 +52,11 @@ def setMacCreatorAndType(path, fileCreator, fileType):
 
         if not all(len(s) == 4 for s in (fileCreator, fileType)):
             raise TypeError("arg must be string of 4 chars")
-        finderInfo = pad(bytesjoin([fileType, fileCreator]), 32)
+        try:
+            existingFinderInfo = xattr.getxattr(path, "com.apple.FinderInfo")
+        except (KeyError, IOError):
+            pass
+        else:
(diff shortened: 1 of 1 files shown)


anthrotype:
thanks, this looks good, though I haven't tested it. I wonder how difficult it'd be to add a test.

Also it doesn't look like this setMacCreatorAndType function is used anywhere in fontTools... Anybody remembers why it's there? /cc @justvanrossum 

@NSGod out of curiousity, what do you use this method for?

justvanrossum:
> Anybody remembers why it's there?

It quite a legacy thing for MacOS Classic. I found two spots in old fonttools where it was used, and rightfully removed later. I can't find other references to it.

NSGod:
> @NSGod out of curiousity, what do you use this method for?

Well, I haven't actually used fontTools all that much yet, or that code at all, for that matter (still learning Python). I've written software that converts between the older formats and newer ones, so I'm keenly aware of the importance of the `com.apple.FinderInfo`. It was kind of an effort to force myself to learn Python.

While the type and creator code are no longer important, the rest of the info includes things like file label and custom icon bit might be worthwhile to preserve if overwriting files.

anthrotype:
so it looks like neither us nor yourself are using that piece of code, I am tempted to just get rid of it or deprecate it/leave as is and if one cares about retaining those attributes they can implement this in their own code..

Issue #3177, opened 2023-06-20
[varLib instancer, L4] Applying a new default location shifts up weight of Bold, etc


arrowtype:
I have a variable font with a wght range of 1–1000, with a default location of 700. More details about the project and designspace in [this related, previous issue](https://github.com/fonttools/fonttools/issues/3139).

I’m attempting to use the varLib Instancer to create a new default location at wght=400, for the desktop fonts only (leaving the webfonts without the added filesize). This is mostly working well, but I’ve noticed a small but meaningful issue: the upper end of the weight range gets shifted to become bolder than before. This deviates from my design intent, and also creates a difference between variable desktop and web fonts, plus between variable and static desktop fonts.

My best guess at a cause is that perhaps the AVAR table is interacting with the font outlines in an unexpected way, after the additional location is added. My weight axis is mapped in the 700–1000 range, and this issue seems to occur between 400 and 1000, and is most visible around 700. 

Here’s a before/after of setting the new default:

> <img width="1101" alt="image" src="https://github.com/fonttools/fonttools/assets/45946693/e7c15c29-5b15-46f3-a84b-6bce81020716">

Here’s a closer view of the two, overlapped:

> <img width="498" alt="image" src="https://github.com/fonttools/fonttools/assets/45946693/707d366a-a768-4f9b-8c85-cee81ae2e5ce">

I’ve come across this issue whether running the Instancer via Python or via a terminal command. For example, I can set the new default like this, and it produces the error:

```bash
fonttools varLib.instancer "AT-NameSansVariable.Trial.ttf" wght=1:400:1000
```

Version: FontTools 4.40.0

Thank you so much for taking a look!

### Test fonts

The following fonts show the before/after example of this. They are trial fonts, so feel free to use them as needed to reproduce and test this issue (or other varLib/fontTools stuff, if relevant).

[AT-NameSansVariable.Trial.ttf.zip](https://github.com/fonttools/fonttools/files/11805876/AT-NameSansVariable.Trial.ttf.zip)
[AT-NameSansVariable.Trial.varLibTest-dflt400.ttf.zip](https://github.com/fonttools/fonttools/files/11805877/AT-NameSansVariable.Trial.varLibTest-dflt400.ttf.zip)

behdad:
Thanks Stephen! This is bad. And indeed quite probably avar related. I'll take a look.

arrowtype:
Glad I could perhaps help uncover something here. Thanks so much for taking a look!

[...]

PR #3179, opened 2023-06-21
L4 fixes

behdad:
Fixes https://github.com/fonttools/fonttools/issues/3177

See https://github.com/fonttools/fonttools/issues/3177#issuecomment-1601335484


behdad:
Err. Needs more work.

behdad:
This still seems to produce wrong results with avar present. Sigh.

behdad:
> This still seems to produce wrong results with avar present. Sigh.

I think I fixed it.

behdad:
Tests fixed.

behdad:
I think this is good now.

behdad:
With the latest changes the test coverage is complete for the change.

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -250,6 +252,50 @@ def __post_init__(self):
...
+
+        totalDistance = distanceNegative * -lower + distancePositive * default
+
+        if v >= 0:
+            vDistance = (default - v) * distancePositive
+        else:
+            vDistance = -v * distanceNegative + distancePositive * default
+
+        if totalDistance == 0:
+            # This happens
```
anthrotype, line 292: when does this situation happen specifically?
IIUC, totalDistance == 0 only when both distanceNegative and distancePositive are 0, which seems unlikely to me in real-world because it would mean fvar axis min==default==max hence the two distances from default zero out.

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -250,6 +252,50 @@ def __post_init__(self):
...
+
+        if v > default:
+            return (v - default) / (upper - default)
+
+        # v < default
+
+        if lower >= 0:
+            return (v - default) / (default - lower)
+
+        # lower < 0 and v < default
```
anthrotype, line 282: I see that your example make sense but I'm still not 100% clear why this situation in which default >=0 and v < default and lower < 0 needs to take into account the positive/negative distances, whereas the other cases do not. Is it because in this particular case the value `v` falls in a segment that may span both the negative and positive sides and thus the ratio between these two matters?

behdad: > Is it because in this particular case the value v falls in a segment that may span both the negative and positive sides and thus the ratio between these two matters?

Correct.

behdad:
> I find it strange that we didn't catch this before, when L4 was not implemented yet. Can you confirm that nothing basically has/will change after this PR for the pinning-axis and restrict-axis-but-keep-default cases of partial instancing, and that these changes only affect when move-axis-default is involved?

Correct. Because as you noted, this only affects cases where the new default moves, and the new range for min-side or max-side spans over the previous default, which only happens with L4. And yes I'm also surprised we didn't catch this before.

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -250,6 +252,50 @@ def __post_init__(self):
...
+
+        # default >= 0 and v != default
+
+        if v > default:
+            return (v - default) / (upper - default)
+
+        # v < default
+
+        if lower >= 0:
+            return (v - default) / (default - lower)
```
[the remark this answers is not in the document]
behdad: Let's leave as is. If someone tries to extrapolate when lower=default, that's an error or at least a case we want to study.

Lib/fontTools/varLib/instancer/__init__.py
behdad: Ah. Now I see it was happening when I was intentionally passing 0,0 as distances in featureVars.py. I removed it, but am not confident it's correct. Lemme think.

behdad: I believe it's correct.

behdad:
Thanks @anthrotype. I think I addressed all your feedback.

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -234,9 +234,29 @@ def limitRangeAndPopulateDefault(self, fvarTriple) -> "AxisTriple":
 
 
 @dataclasses.dataclass(frozen=True, order=True, repr=False)
-class NormalizedAxisTripleAndDistances(AxisTriple):
+class NormalizedAxisTrip(AxisTriple):
```
anthrotype, line 237: typo? NormalizedAxisTrip**le**

anthrotype (approved):
thanks LGTM! Remember to fix the typo

PR #3183, opened 2023-06-25
[FontBuilder] Update setupAvar() to also support avar-2, fixing _add_avar() call site

### LINKED ISSUE #3182: [FontBuilder] _add_avar() is called with the wrong amounts of arguments
`_add_avar()` was changed for `avar-2`, but the call site in fontBuilder.py wasn't updated, so calling `builder.setupAvar()` fails.

I'll post a PR.

justvanrossum:
Fixes #3182.

The avar-2 bit is untested, but I did add a minimal FontBuilder avar test that would have caught #3182.

### DIFF
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -707,7 +707,7 @@ class FontBuilder(object):
 
         addFvar(self.font, axes, instances)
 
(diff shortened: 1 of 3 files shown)


behdad:
LGTM. Sorry about the breakage.

PR #3185, opened 2023-06-26
Remove redundant name table records

m4rc1e:
I've needed a way to remove redundant name table records for many years so I thought I'd just write it. I'm using Behdad's visitor pattern to collect all the name records which are being used by the fvar, STAT, CPAL etc.

You can get redundant name records when you replace a font's STAT table using the builder's buildSTAT function or when hot fixing with other scripts etc.

My plan is to make `removeUnusedNameRecords` a method of `table__n_a_m_e`. I just wanted to open this pr to get feedback before I progress too far. No worries if you don't want this pr either.

So far I've covered every table in [ttLib/tables/otData](https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/ttLib/tables/), STAT, fvar and CPAL. Please let me know if I'm missing any.


Snippets/clean-nametable.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+logger = logging.getLogger()
+
+
+class NameRecordVisitor(TTVisitor):
+    def __init__(self):
+        self.seen = set()
+
+    def removeUnusedNameRecords(self, font):
```
behdad, line 19: Maybe remove this from the visitor object, move it somewhere else.

behdad (commented):
Looks good!

Snippets/clean-nametable.py
m4rc1e: Thanks for looking. I've moved and converted the function so it's a static method of the name table. Imo it feels like belongs here. Happy to drop the static method and just have it as a plain func instead. I've moved it away from the Snippets dir so it can be imported into other libs and tools.

Happy to add tests if you're ok with the above. Still not too sure if I've found all the places in a font which can have a nameID though.

behdad:
Thanks. We should see if we can use this, or at least compare notes with, the subsetter.

behdad:
> [ @m4rc1e ](/m4rc1e)
> ` [Use NameRecordVisitor in subsetter](/fonttools/fonttools/pull/3185/commits/edf8891fba54015c5d37a6397b7b95f14dfe5662)`

Slick! I just wonder if we need to add a table order dependency to process `name` last.

m4rc1e:
>  I just wonder if we need to add a table order dependency to process name last.

I'm not too familiar with all of the subsetting code. However, if I skim, I see `self._prune_pre_subset` and `_subset_glyphs` methods call `_sort_tables`. Is it sufficient to just move the "name" table to last pos in `_sort_tables`? if not, I'll keep digging.

behdad:
Thanks. Checking the subset code, I think we need to move the name-table subsetting from `prune_pre_subset` to `prune_post_subset`. That might be enough.

There's also some "name" table code in the "CPAL" subsetting method, which should be removed.

I don't remember why "name" table is in `_sort_tables`. May be possible to remove it now, maybe not.

behdad:
> There's also some "name" table code in the "CPAL" subsetting method, which should be removed.

Can you please do this as well?

Thanks. I let @anthrotype review the rest.

Lib/fontTools/subset/__init__.py
```diff
@@ -3463,7 +3427,11 @@ def _subset_glyphs(self, font):
             font.setGlyphOrder(self.new_glyph_order)
 
     def _prune_post_subset(self, font):
-        for tag in font.keys():
+        tableTags = font.keys()
+        if "name" in tableTags:
```
anthrotype, line 3431: perhaps add little comment why we need to post-subset the name table last

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -223,6 +227,28 @@ def removeNames(self, nameID=None, platformID=None, platEncID=None, langID=None)
...
+        for record in ttFont["name"].names:
+            # Name IDs 26 to 255, inclusive, are reserved for future standard names.
+            # https://learn.microsoft.com/en-us/typography/opentype/spec/name#name-ids
+            if record.nameID < 256:
+                continue
+            if record.nameID not in visitor.seen:
+                toDelete.add(record.nameID)
+
+        if not toDelete:
+            log.info("Name table has no redundant records, skipping")
```
anthrotype, line 246: demote to DEBUG or remove entirely? Maybe the method could return the set of nameIDs that it has deleted so that caller can decide whether they wish to log this or not

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -1138,3 +1164,50 @@ def __repr__(self):
...
+        visitor.seen.add(inst.subfamilyNameID)
+
+    for axis in obj.axes:
+        visitor.seen.add(axis.axisNameID)
+
+
+@NameRecordVisitor.register(ttLib.getTableClass("CPAL"))
+def visit(visitor, obj):
+    for nameID in obj.paletteLabels + obj.paletteEntryLabels:
+        if obj.version == 1:
```
anthrotype, line 1212: move this `if` out of the loop, it's constant. Maybe do the same as the code in the subsetter that you deleted:

```python
    if obj.version == 1:
        visitor.seen.update(obj.paletteLabels)
        visitor.seen.update(obj.paletteEntryLabels)
```

anthrotype (approved):
LGTM, with comments

anthrotype:
the snippet doesn't do much besides loading a TTFont, calling remoteUnusedNames method, and saving back; do we actually need it?

Lib/fontTools/ttLib/tables/_n_a_m_e.py
m4rc1e: I've removed the log. Method now returns the deleted records nameIDs.

m4rc1e:
Thanks Cosimo. I've implemented your feedback and removed the snippet. The comment may be too long for your liking. Lmk and I'll rephrase.

anthrotype (approved):

Issue #3189, opened 2023-06-27
[glyf-1 cubic curves] Last cubic segment becomes quadratic


justvanrossum:
This is happening when drawing a variable composite glyph from a TTGlyphSet onto a TTGlyphPointPen. All outlines are cubic in the source. The bug is either in the TTGlyphSet code or in TTGlyphPointPen. Or somewhere else. (I will dig deeper, and fix if I can.)

<img width="589" alt="image" src="https://github.com/fonttools/fonttools/assets/4246121/48055075-6b35-41c1-9226-7e0e3abc7896">

behdad:
Thanks. Pretty sure it's in `Glyph.draw()`.

behdad:
Is this an on-curve-less outline?

behdad:
> > Is this an on-curve-less outline?
> 
> No, it's an ordinary mix of on-curve and off-curve points.

I can take a look if you send me a font.

justvanrossum:
The bug is in `TTGlyphPen` and `TTGlyphPointPen`:

```python
from fontTools.pens.ttGlyphPen import TTGlyphPen, TTGlyphPointPen
from fontTools.pens.recordingPen import RecordingPointPen

points = [
    ((0, 0), "curve"),
    ((0, 0), None),
    ((0, 0), None),
    ((0, 0), "curve"),
    ((0, 0), None),
    ((0, 0), None),
]

if True:
    pen = TTGlyphPointPen(None)
    pen.beginPath()
    for (x, y), segmentType in points:
        pen.addPoint((x, y), segmentType=segmentType)
    pen.endPath()
else:
    pen = TTGlyphPen(None)
    pen.moveTo((0, 0))
    pen.curveTo((0, 0), (0, 0), (0, 0))
    pen.curveTo((0, 0), (0, 0), (0, 0))
    pen.closePath()

glyph = pen.glyph()

rpen = RecordingPointPen()
glyph.drawPoints(rpen, None)

for m, args, kwargs in rpen.value:
    if args:
        pt, tp, *_ = args
        print(m, pt, tp)
    else:
        print(m)
```

In both cases, the output is:
```
beginPath
addPoint (0, 0) qcurve
addPoint (0, 0) None
addPoint (0, 0) None
addPoint (0, 0) curve
addPoint (0, 0) None
addPoint (0, 0) None
endPath
```
Where the first segmentType should be "curve" instead of "qcurve".

justvanrossum:
[...]

PR #3190, opened 2023-06-27
[cubic-glyf] Fix drawPoints() contour ending with cubic offCurve

behdad:
Fixes https://github.com/fonttools/fonttools/issues/3189


behdad:
> How should an all-offcurve cubic contour be represented in pointPen? It looks the same as a quadratic one, no?

Oh, the lone cubic on-curve with None coordinates I think. Fixing.

justvanrossum (approved):
LGTM!

justvanrossum:
> How should an all-offcurve cubic contour be represented in pointPen? It looks the same as a quadratic one, no?

That is a design flaw in the point pen protocol, that is reflected in UFO outlines, too.

behdad:
Filed https://github.com/fonttools/fonttools/issues/3191.

PR #3194, opened 2023-06-29
Designspace optional name

behdad:


behdad:
The first commit is already in main. My bad for getting included here.

rsheeter:
Can we also update https://fonttools.readthedocs.io/en/latest/designspaceLib/python.html#fontTools.designspaceLib.SourceDescriptor.name to mention that we generate a name if none is present? - https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/designspaceLib/__init__.py#L2311. Currently it just says optional.

rsheeter (approved):
LGTM

justvanrossum (approved):

PR #3200, opened 2023-07-07
Return three-byte operands from `encodeIntCFF`

dscorbett:
According to [the CFF spec](https://adobe-type-tools.github.io/font-tech-notes/pdfs/5176.CFF.pdf), there is a three-byte integer format identical to the one in Type 2 CharStrings, so `encodeIntCFF` should support it.

### DIFF
--- a/Lib/fontTools/misc/psCharStrings.py
+++ b/Lib/fontTools/misc/psCharStrings.py
@@ -197,11 +197,14 @@ t2Operators = [
 
 def getIntEncoder(format):
     if format == "cff":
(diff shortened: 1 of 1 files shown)


behdad:
Shouldn't the name be two-byte op, given that in the code op 29 is called four-byte op?

Also, how did this go unnoticed?!

dscorbett:
I renamed the variable to `twoByteOp` for consistency with `fourByteOp`, but I left the commit message the same because the CFF spec refers to it as a 3-byte format.

PR #3201, opened 2023-07-07
Convert panoseDefault to namespace object

### LINKED ISSUE #3197: [fontBuilder] Default panose value is set to a dict
When using fontBuilder to build a font, the default panose is set as a dictionary and not a object namespace causing issues downstream.

### Discussed in https://github.com/fonttools/fonttools/discussions/3196

<div type='discussions-op-text'>

<sup>Originally posted by **NightFurySL2001** July  2, 2023</sup>
3. `extractor` raises two error: `AttributeError: 'dict' object has no attribute 'bFamilyType'` ...
I can't catch the bug for the former. Maybe it's due to `fontbuilder` doesn't build panose?
</div>

<div type='discussions-op-text'>
<sup>Originally posted by @anthrotype in https://github.com/robotools/extractor/issues/58#issuecomment-1617773977</sup>

looks to me like a bug in the fontBuilder, because it leaves the OS/2 table's `panose` attribute as a dictionary (as in `_panoseDefaults`), whereas it should set it to a `Panose` object like the OS/2 table's decompile method does. The reason it works when saving from the fontBuilder is that the fonttools' `sstruct.pack` method accepts either an object or a dictionary and so compile works regardless.
We should fix it in fonttools. @NightFurySL2001 do you mind filing an issue or working on a PR even?
As a workaround, you could temporarily save the fontBuilder's font (to a tempfile or BytesIO stream), reload it and then try to extractor it. I believe it will work then, for the panose attribute will be an object and not a dict upon decompiling.
</div>

NightFurySL2001:
Fix #3197

### DIFF
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -132,10 +132,11 @@ fb.save("test.otf")
 from .ttLib import TTFont, newTable
 from .ttLib.tables._c_m_a_p import cmap_classes
 from .ttLib.tables._g_l_y_f import flagCubic
+from .ttLib.tables.O_S_2f_2 import Panose
 from .misc.timeTools import timestampNow
 import struct
 from collections import OrderedDict
-from types import SimpleNamespace
+
 
 _headDefaults = dict(
     tableVersion=1.0,
@@ -263,18 +264,7 @@ _nameIDs = dict(
 # to insert in setupNameTable doc string:
 # print("\n".join(("%s (nameID %s)" % (k, v)) for k, v in sorted(_nameIDs.items(), key=lambda x: x[1])))
 
-_panoseDefaults = SimpleNamespace(
-    bFamilyType=0,
-    bSerifStyle=0,
-    bWeight=0,
-    bProportion=0,
-    bContrast=0,
-    bStrokeVariation=0,
-    bArmStyle=0,
-    bLetterForm=0,
-    bMidline=0,
-    bXHeight=0,
-)
+_panoseDefaults = Panose()
 
 _OS2Defaults = dict(
     version=3,
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -135,7 +135,7 @@ from .ttLib.tables._g_l_y_f import flagCubic
 from .misc.timeTools import timestampNow
 import struct
 from collections import OrderedDict
-
+from types import SimpleNamespace
 
 _headDefaults = dict(
     tableVersion=1.0,
@@ -263,7 +263,7 @@ _nameIDs = dict(
 # to insert in setupNameTable doc string:
 # print("\n".join(("%s (nameID %s)" % (k, v)) for k, v in sorted(_nameIDs.items(), key=lambda x: x[1])))
 
-_panoseDefaults = dict(
+_panoseDefaults = SimpleNamespace(
     bFamilyType=0,
     bSerifStyle=0,
     bWeight=0,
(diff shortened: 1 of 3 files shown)


Lib/fontTools/fontBuilder.py
```diff
@@ -263,17 +263,19 @@ def drawTestGlyph(pen):
...
-    bSerifStyle=0,
-    bWeight=0,
-    bProportion=0,
-    bContrast=0,
-    bStrokeVariation=0,
-    bArmStyle=0,
-    bLetterForm=0,
-    bMidline=0,
-    bXHeight=0,
+_panoseDefaults = SimpleNamespace(
```
justvanrossum, line 266: You could use `SimpleNamespace` instead of `dict` directly, there's no needed for the nested `**dict(...)`

NightFurySL2001: Thanks for the pointer, I updated the commit

anthrotype:
I'd rather the OS2 table module's `Panose` object which is used when decompiling or loading from XML. Just add an `__init__` to that takes keyword-only arguments and initializes all self attributes to zeros, then fontBuilder can just use`_panoseDefaults=Panose()`.

https://github.com/fonttools/fonttools/blob/3235454fcd720f270e18ddf00b058df83da63547/Lib/fontTools/ttLib/tables/O_S_2f_2.py#L27

If one wishes to override any panose attributes when building OS/2 with fontBuilder they can pass the latter a Panose instance, and can easily construct one with the new constructor.

Lib/fontTools/ttLib/tables/O_S_2f_2.py
```diff
@@ -25,6 +25,14 @@
 
 
 class Panose(object):
+    def __init__(self, panoseDict=None):
+        formatstring, names, fixes = sstruct.getformat(panoseFormat)
+        for name in names:
+            if panoseDict is not None and name in panoseDict:
+                setattr(self, name, safeEval(panoseDict[name]))
+            else:
+                setattr(self, name, 0)
```
anthrotype, line 34: Was thinking more something like this:

```suggestion
    def __init__(self, **kwargs):
        _, names, _ = sstruct.getformat(panoseFormat)
        for name in names:
            setattr(self, name, kwargs.pop(name, 0))
        for k in kwargs:
            raise TypeError(f"Panose() got an unexpected keyword argument {k!r}")
```

anthrotype: don't need to eval, we're not accepting strings in the init

anthrotype:
a little test in fontBuilder_test.py would be nice too, e.g. constructing OS/2 once without overriding panose defaults, and another overriding with custom Panose instance, asserting the panose attribute is an object with expected attrs

Tests/fontBuilder/fontBuilder_test.py
```diff
@@ -420,3 +420,49 @@ def test_unicodeVariationSequences(tmpdir):
...
+        "bArmStyle": 5,
+        "bLetterForm": 2,
+        "bMidline": 3,
+        "bXHeight": 4,
+    }
+
+    for name in panoseValues:
+        assert getattr(fb.font["OS/2"].panose, name) == 0
+
+    fb.setupOS2(panoseValues)
```
anthrotype, line 462: no, I don't think setupOS2 works like that, you want to construct a panose = Panose(**panoseValues) instance beforehand, and pass that one to fb.setupOS2(panose=panose)

Tests/fontBuilder/fontBuilder_test.py
```diff
@@ -420,3 +420,49 @@ def test_unicodeVariationSequences(tmpdir):
     fb.setupCharacterMap(cmap, uvs)
     fb.save(outPath)
     _verifyOutput(outPath, tables=["cmap"])
+
+
+def test_setupPanose(is_ttf, keep_glyph_names, make_cff2, post_format):
```
anthrotype, line 425: you are not using these pytest fixture arguments, just define it as  `def test_setupPanose():`

PR #3204, opened 2023-07-13
Fix typing of script_horizontal_direction()

Hoolean:
Without explicit annotation, some type checkers inferred that the type of the 'default' argument could only be `type[KeyError]`. This was the case in unicodedata_test.py, where pyright disallowed `"LTR"`. This PR adds annotations to avoid this, fixing the issue in the test, and in external code depending on the API.

### Extras

Some explicit conversions of literals to `str` were removed, as Python 3 no longer requires them and in one case they would have required a less specific return type to be used.

### Caveats

* Some of the other functions in this file have the same semantics and suffer from the same type error, and so this fix could also be extended to them as usage requires.
* `HorizDirection` and `T` may become importable; do we want to avoid this for either?

### DIFF
--- a/Lib/fontTools/unicodedata/__init__.py
+++ b/Lib/fontTools/unicodedata/__init__.py
@@ -1,7 +1,11 @@
+from __future__ import annotations
+
 from fontTools.misc.textTools import byteord, tostr
 
 import re
 from bisect import bisect_right
+from typing import Literal, TypeVar, overload
+
 
 try:
     # use unicodedata backport compatible with python2:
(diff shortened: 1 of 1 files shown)


anthrotype:
LGTM thanks

> Some of the other functions in this file have the same semantics and suffer from the same type error, and so this fix could also be extended to them as usage requires.

feel free to fix those too or just leave it for when the need arises

> HorizDirection and T may become importable; do we want to avoid this for either?

I don't see much of a problem with that

Hoolean:
> feel free to fix those too or just leave it for when the need arises

Cheers Cosimo, will probably leave out of this PR as I have not been prodding them with the same testing

> I don't see much of a problem with that

Nice nice! Happy for the branch to be merged if you're happy with it too 🚀

Hoolean:
Thanks again :)

Issue #3168, opened 2023-06-18
[varLib] Support for sparse kerning / anchors


behdad:
I want to make sparse kerning / anchor masters possible.

behdad:
Here's my current idea for sparse kerning. We'll add an extension keyword to the .fea format `override`. Then the following block:
```
override feature kern {

} kern;
```
will signify an sparse kerning block. The implementation simply create a class-kerning table that has 0x7FFF instead of 0 for missing entries. In `varLib.merger` then we just replace the `0x7FFF` values with `None`, which will automatically remove them from the interpolation model.

behdad:
Same idea `override` can be extended to other lookup types.

behdad:
@anthrotype wdyt?

rsheeter:
@cmyr might also be interested. 

Current behavior, iiuc https://github.com/googlefonts/fontmake/issues/998 correctly is, not entirely intuitive: if a static being merged:

* has _any_ GPOS fontmake fills holes with UFO kerning algo, which notably falls back to 0 (https://unifiedfontobject.org/versions/ufo3/kerning.plist/#kerning-value-lookup-algorithm )
* has no GPOS fontmake produces interpolated kerning

The ability to specify _some_ and let the rest interpolate is notably missing, I quite like the idea. I'm curious if type designers would find this useful, @davelab6 wdyt?

davelab6:
VERY USEFUL! :)

behdad:
> VERY USEFUL! :)

Cool. I'll prioritize.

I need input on the .fea file extension (`override` keyword). cc @simoncozens @khaledhosny

rsheeter:
@skef penny for Adobe's thoughts?

skef:
There are a lot of potential questions in this space.

I'll start by noting a deep rabbit hole we went down considering whether to support a distributed data model in a VF-first workflow. That is, a model where a single hierarchy of fea files could still support supplying kerning data for individual masters separately. At this point we're leaning against that model because although it could work fine in common cases, in the general case it will sometimes be hard to match up lookups between the per-master lists, and accordingly hard to describe the heuristics for doing so and to implement those heuristics in the varying implementations out in the field, which don't currently do anything like that. Given that except for things like chaining rules most of these files are derived from other sources, and can now be given in a centralized format with the appropriate "sparseness caveats" it seems like more trouble than its worth. 

[...]

PR #3209, opened 2023-07-18
[varLib.merger] Support sparse CursivePos

behdad:
Implement merging of sparse CursivePos.

Part of https://github.com/fonttools/fonttools/issues/3168


anthrotype (approved):
LGTM

PR #3210, opened 2023-07-18
Fix wrong created date in fontBuilder

martinleopold:
Bug introduced in https://github.com/fonttools/fonttools/commit/a421f9045a4d1e130daece08350ed7c1ed300ee7#diff-c01f84c2ac99ae2b60f071d4a38ff88c2653d27f19e62e11bef3dc7fe7f68925R352

Causes created date to be set to 0 instead of the current timestamp.


justvanrossum (approved):
Good catch, thanks!

PR #3214, opened 2023-07-20
[subset] Revert use of NameVisitor to fix speed regression

### LINKED ISSUE #3213: Performance regression in 4.41.0?
See: https://github.com/satbyy/go-noto-universal/issues/59

Using 4.41.0 the builds take longer than 6 hours, but using 4.40.0 it takes 20ish minutes. Not too sure what is going on since I am not familiar with either project but just thought I would point it out

anthrotype:
Fixes #3213

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -15,7 +15,6 @@ from fontTools.subset.util import _add_method, _uniq_sort
 from fontTools.subset.cff import *
 from fontTools.subset.svg import *
 from fontTools.varLib import varStore  # for subset_varidxes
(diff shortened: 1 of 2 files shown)


anthrotype:
@imran-iq @satbyy could you please try to install fonttools from this PR branch and see if that fixes the speed regression you experienced?

`pip install -e git+https://github.com/fonttools/fonttools.git@subset-speed-regression#egg=FontTools`

behdad:
Lgtm. Unfortunate. cc @m4rc1e.

m4rc1e:
Thanks for just removing it from the subsetter only.

PR #3215, opened 2023-07-20
Cache repeated subtables while building COLR

simoncozens:
When building a complex COLR table, particularly for a font with a large number of glyphs, there can be *many* repeated subtables. Instead of rebuilding each subtable from scratch, a simple cache can be used to speed up the build.

For example, building [Bitcount](https://bitcount.typenetwork.com) without the cache takes 274 seconds on my computer; with the cache,  it takes 94.

### DIFF
--- a/Lib/fontTools/colorLib/table_builder.py
+++ b/Lib/fontTools/colorLib/table_builder.py
@@ -5,6 +5,8 @@ colorLib.table_builder: Generic helper for filling in BaseTable derivatives from
 
 import collections
 import enum
+import json
+import copy
 from fontTools.ttLib.tables.otBase import (
     BaseTable,
     FormatSwitchingBaseTable,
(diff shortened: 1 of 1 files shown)


simoncozens:
...but this only works when you're sending a plain JSON-style object to `buildPaint`, not a mix of dictionaries and pre-compiled subtables...

anthrotype:
hm yeah, the API supports a mix of raw dicts and prebuilt subtables and I'd like to keep that way. Also I'm a bit meh on having KBs of JSON for the cache dict keys..

anthrotype:
I'm worried about the potential risk of collisions when using hash instead of the strings themselves as the key. This SO suggests using more cryptographically robust hashlib (also in stdlib).
Also I think pickle is faster than json in general.

anthrotype:
also how about we make this caching optional by adding a parameter to the buildCOLR table that trickles down to the TableBuilder constructor? Maybe for your particular use case this may make things faster, but for the general case I am not sure whether the cost of pickling and caching outweighs the benefits

simoncozens:
Closing in favour of #3221

anthrotype:
> Closing in favour of https://github.com/fonttools/fonttools/pull/3221

my PR was based off yours, the idea was if you liked that suggestion you could merge that first into your PR branch and proceed from there. But since you closed this one I have rebased my PR onto main. Feel free to pick that one up

PR #3217, opened 2023-07-20
Faster name visitor subset

behdad:
https://github.com/fonttools/fonttools/issues/3213#issuecomment-1644158826

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -15,6 +15,7 @@ from fontTools.subset.util import _add_method, _uniq_sort
 from fontTools.subset.cff import *
 from fontTools.subset.svg import *
 from fontTools.varLib import varStore  # for subset_varidxes
+from fontTools.ttLib.tables._n_a_m_e import NameRecordVisitor
 import sys
 import struct
 import array
@@ -2585,25 +2586,10 @@ def prune_post_subset(self, font, options):
 
     if self.version == 1:
         kept_labels = []
-        dropped_labels = []
         for i, label in enumerate(self.paletteEntryLabels):
             if i in retained_palette_indices:
                 kept_labels.append(label)
-            else:
-                dropped_labels.append(label)
         self.paletteEntryLabels = kept_labels
-        # Remove dropped labels from the name table.
-        name_table = font["name"]
-        name_table.names = [
-            n
-            for n in name_table.names
-            if (
-                n.nameID not in dropped_labels
-                # Only remove nameIDs in the user range and if they're not explicitly kept
-                or n.nameID < 256
-                or n.nameID in options.name_IDs
-            )
-        ]
     return bool(self.numPaletteEntries)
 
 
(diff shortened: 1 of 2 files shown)


Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -242,6 +242,9 @@ def removeUnusedNames(ttFont):
             if record.nameID not in visitor.seen:
                 toDelete.add(record.nameID)
 
+        if not toDelete:
+            return
+        log.info(f"Deleting name records with NameIDs {toDelete}")
```
anthrotype, line 247: these three lines above we don't want regardless

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -1208,3 +1211,15 @@ def visit(visitor, obj):
...
         visitor.seen.update(obj.paletteLabels)
         visitor.seen.update(obj.paletteEntryLabels)
+
+
+@NameRecordVisitor.register(
+    (
+        ttLib.getTableClass("cmap"),
+        ttLib.getTableClass("glyf"),
+        ttLib.getTableClass("CFF "),
+        ttLib.getTableClass("CFF2"),
```
anthrotype, line 1221: hm I don't know.. what if a font has a huge/deep table that you don't know about? It feels to me the selection of tables to visit should be opt-in, not opt-out

behdad: Sounds good. Let me see if I can implement that easily.

Lib/fontTools/subset/__init__.py
```diff
@@ -2915,32 +2901,10 @@ def prune_pre_subset(self, font, options):
...
-        nameIDs.update(
-            [axis_rec.AxisNameID for axis_rec in stat.table.DesignAxisRecord.Axis]
-        )
-    cpal = font.get("CPAL")
-    if cpal and cpal.version == 1:
-        nameIDs.update(cpal.paletteLabels)
-        nameIDs.update(cpal.paletteEntryLabels)
+def prune_post_subset(self, font, options):
+    visitor = NameRecordVisitor()
+    visitor.visit(font)
```
anthrotype, line 2906: why do we need to "visit" the entire font if we know already where we expect to find nameIDs?

behdad: PTAL

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -1208,3 +1208,16 @@ def visit(visitor, obj):
...
         visitor.seen.update(obj.paletteEntryLabels)
+
+
+@NameRecordVisitor.register(ttLib.TTFont)
+def visit(visitor, font, *args, **kwargs):
+    if hasattr(visitor, "font"):
+        return False
+
+    visitor.font = font
+    for tag in ("GSUB", "GPOS", "fvar", "CPAL", "STAT"):
```
anthrotype, line 1219: how about storing this list of table tags that can contain nameID references as a class attribute of NameRecordVisitor?

behdad: Okay.

behdad: Done.

anthrotype:
it would be nice if @imran-iq could give this PR branch a try with their go-noto font build to confirm it is fast enough?

anthrotype (approved):

behdad:
Here's a timing I observed:
```
$ time ./fonttools subset --gids=0-10000 --font-number=0 SourceHanNotoCJK.ttc --timing
Took 0.000s to load font
Took 0.000s to compile glyph list
Took 0.000s to load 'maxp'
Took 0.000s to prune 'maxp'
Took 0.041s to load 'cmap'
Took 0.000s to prune 'cmap'
Took 0.000s to load 'post'
Took 0.000s to prune 'post'
Took 0.000s to load 'CFF '
Took 0.017s to prune 'CFF '
Took 0.000s to load 'GPOS'
Took 0.009s to prune 'GPOS'
Took 0.000s to load 'GSUB'
Took 0.015s to prune 'GSUB'
Took 0.012s to close glyph list over 'cmap'
Took 0.068s to close glyph list over 'GSUB'
Took 1.175s to close glyph list over 'CFF '
Took 0.011s to subset 'cmap'
Took 0.000s to subset 'post'
Took 0.013s to subset 'CFF '
Took 0.000s to subset 'VORG'
Took 0.056s to subset 'hmtx'
Took 0.016s to subset 'vmtx'
Took 0.000s to subset 'GDEF'
Took 0.090s to subset 'GPOS'
Took 0.010s to subset 'GSUB'
Took 0.000s to subset GlyphOrder
Took 0.000s to prune 'head'
Took 2.202s to prune 'CFF '
Took 0.000s to prune 'GDEF'
Took 0.002s to prune 'GPOS'
Took 0.005s to prune 'GSUB'
Took 0.186s to prune 'name'
Took 1.160s to compile and save font
Took 5.102s to make one with everything (TOTAL TIME)

real	0m5.222s
user	0m5.147s
sys	0m0.059s
```
Without name closure it takes `4.925` (as in without this PR). The name pruning took `0.186s`, which I think is acceptable.

With the previous approach this was taking minutes before I interrupted it...

PR #3223, opened 2023-07-23
[varLib] Add avarPlanner

behdad:
Currently deduces "correct" weight axis mapping based on glyph average weights of axis min/default/max.

EDIT: Plans width and slant axes now too.


behdad:
I have to generate waterfalls and observe whether the generated output makes visual sense.

behdad:
I made some waterfalls, and the results look good! In each picture, the left column is the results produced by this tool, and the right column is the original font, with any avar mapping it might have.

NotoSansArabic-VF:
image

RobotoFlex: (has no avar table)
image

NotoSansCJKtc-VF:
image

behdad:
cc @twardoch

behdad:
For reference: https://www.diacritics.club/family-steps

Lorp:
Can you clarify what this is doing? Have you implemented Lucas and Pablo curves as HOI in avar2? (which would be a fabulous new use case)

behdad:
> Can you clarify what this is doing? Have you implemented Lucas and Pablo curves as HOI in avar2? (which would be a fabulous new use case)

No. It's just avar1. It's an alternative to Lucas & Pablo curves. I think it produces better results than those. It works by calculating the font's gray "color" using fontTools areaPen, and then treating the weight as a log-linear variable of the gray color.

behdad:
> It's an alternative to Lucas & Pablo curves. I think it produces better results than those.

What's curious is that @lucasfonts has the right idea that the stem-widths (in his case) / blackness (in my case) should be proportionally the same across adjacent weights:

  https://www.lucasfonts.com/learn/interpolation-theory

That cleanly maps to a exponential function though. I'm not sure why he didn't arrive at that and arrived at a more complicated formula there.

behdad:
Sample output:
```
$ ./fonttools varLib.avarPlanner RobotoFlex.ttf -q
    <axis tag="wght" name="Weight" minimum="100" maximum="1000" default="400">
      <map input="100" output="100"/>
      <map input="150" output="137.044"/>
      <map input="200" output="178.423"/>
      <map input="250" output="224.677"/>
      <map input="300" output="276.498"/>
      <map input="350" output="334.62"/>
      <map input="400" output="400"/>
      <map input="450" output="435.201"/>
      <map input="500" output="472.464"/>
      <map input="550" output="512.077"/>
      <map input="600" output="554.097"/>
      <map input="650" output="598.675"/>
      <map input="700" output="646.102"/>
      <map input="750" output="696.45"/>
      <map input="800" output="749.934"/>
      <map input="850" output="806.78"/>
      <map input="900" output="867.231"/>
      <map input="950" output="931.557"/>
      <map input="1000" output="1000"/>
    </axis>
```

behdad:
cc @LettError

behdad:
Waterfalls HTML files for testing:
[waterfalls.zip](https://github.com/fonttools/fonttools/files/12139237/waterfalls.zip)

tiroj:
The Arabic waterfall comparison is a good example of both what a nice progression the new avarPlanner produces and why in a practical typographic context it might not be ideal. Functionally, the relationship between Regular and Bold weights is the most important part of the progression, because this combination is most common in typical text—for historical reasons—so the Bold weight needs a certain ‘punch’ relative to the Regular. That’s why I sometimes end up tweaking my progression to make the Bold slightly heavier than the calculated weight. In the Noto Sans Arabic waterfall, the progression on the left is certainly nice and smooth, but the nominal Bold weight (presuming we’re looking at 100–900 wight class range) is a little timid.

behdad:
> The Arabic waterfall comparison is a good example of both what a nice progression the new avarPlanner produces and why in a practical typographic context it might not be ideal. Functionally, the relationship between Regular and Bold weights is the most important part of the progression, because this combination is most common in typical text—for historical reasons—so the Bold weight needs a certain ‘punch’ relative to the Regular. That’s why I sometimes end up tweaking my progression to make the Bold slightly heavier than the calculated weight. In the Noto Sans Arabic waterfall, the progression on the left is certainly nice and smooth, but the nominal Bold weight (presuming we’re looking at 100–900 wight class range) is a little timid.

Thanks John. For that reason I added the `--weight-pins` argument, such that user can pin certain weights. Another example of its need would be to adjust the 400 of the CJK sample.

behdad:
The slant code basically implements the same idea as:
https://twitter.com/jenskutilek/status/1083243100036837384
but using real calculated slant values.

behdad:
I also implemented `opsz` axis mapping. Picture (left: after; right: original font):

image

The rows here have sizes that proportionally (NOT linearly) get larger. So a smooth straight line is desirable in my opinion.

tiroj:
Can we see that waterfall at target sizes and, more importantly, the individual text size targeting from 8–16pt?

behdad:
> Can we see that waterfall at target sizes

image

Left the tool's, right the font's existing avar'.

behdad:
> Can we see that waterfall at target sizes and, more importantly, the individual text size targeting from 8–16pt?

image

tiroj:
This looks promising, but I will need to find some time to play with it with a better opsz test font. This sans serif is pretty tightly spaced even at its smallest size. I’ll look around for a more traditional text face in which the spacing between 6pt and 10pt changes more significantly.

LettError:
I would like to test it on an opsz thing I have on my desk but can not share.

behdad:
> I would like to test it on an opsz thing I have on my desk but can not share.

Sure. With fonttools main branch, do:
```
fonttools varLib.avarPlanner varfont.ttf
```

Let us know how it goes!

Lorp:
Shouldn’t this be calibrated against a corpus of Regular and Bold fonts, Regular and Condensed fonts, etc?

behdad:
> Shouldn’t this be calibrated against a corpus of Regular and Bold fonts, Regular and Condensed fonts, etc?

From my cursory look around, the actual weights of Regular/Bold and their ratio varies widely across families (specially when you look across different scripts). So the tool cannot suggest any absolute numbers for those. Instead it logarithmically interpolates between Regular/Black to suggest where Bold should be *perceptually*. Given Blacks I've seen, this undershoots the designer-set Bold. That is, there's just not enough variance between most Bold and Blacks. The difference is not as pronounced as Regular to Bold.

behdad:
> > No. It's just avar1. It's an alternative to Lucas & Pablo curves. I think it produces better results than those. It works by calculating the font's gray "color" using fontTools areaPen, and then treating the weight as log-linear in the gray color.
> 
> Could you please explain the curve and

From what I wrote on Twitter:
https://twitter.com/behdadesfahbod/status/1729557135862780000

I think it's first good to review this: [https://diacritics.club/family-steps](https://t.co/PnXEp7KvdV)

My method is closer to Luc(as)'s. Except that it uses actual ink area instead of adjusting the stem widths only.

I think it's first good to review this: [https://diacritics.club/family-steps](https://t.co/PnXEp7KvdV)

In my understanding, you are suggesting the linear progression.

My method is closer to Luc(as)'s. Except that it uses actual ink area instead of adjusting the stem widths only.

The way it works is that it first samples the weight axis at a few points (8; doesn't matter) to build a function of from the axis value to the actual ink percentage on the page. This is done using the `fontTools.pens.areaPen`.

After that, it inserts the requested number of in-betweens, such that the *ratio* of ink between successive ones are the same.

That is, it produces a *geometric* progression of the ink percentage.

> are you calculating the gray 'color' using a key glyph?

I'm calculating the color from the entire glyphset.

PR #3234, opened 2023-08-01
Sparse cff2

behdad:


behdad:
Ignore the first commit. I pushed it to main now.

behdad:
Okay now only `test_varlib_build_sparse_CFF2` test failes, which might need new expectations.

behdad:
Tests pass now.

anthrotype (approved):
LGTM

I'll push a test after lunch

PR #3235, opened 2023-08-02
varLib: use sentinel to mark glyph metrics as sparse

anthrotype:
as discussed in https://github.com/googlefonts/ufo2ft/issues/501, sometimes one wants to have a gyph in a non-default sparse master without it participating in HVAR/VVAR. For outlines (both glyf and CFF) we allow empty glyphs in non-default masters to mean the same as missing, thus non participating. This PR enables a similar thing for hmtx and vmtx advances and sidebearings (we don't actually build deltas for lsb/tsb the sentinel value includes both to make it less likely that it occurs in real world fonts).

With this PR, if a glyph has both an advance of 65535 (0xFFFF) and a left/top sidebearing of -32768 (-0x8000), it is ignored when building HVAR or VVAR, as if is were missing from a master.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -563,9 +563,17 @@ def _get_advance_metrics(
 ):
     vhAdvanceDeltasAndSupports = {}
     vOrigDeltasAndSupports = {}
+    # HACK: we treat width 65535 and lsb -32768 as sentinel values to signal that a
+    # glyph from a non-default master should not participate in computing {H,V}VAR,
+    # as if it were missing. Allows to variate other glyph-related data independently
+    # from glyph metrics
+    sparse_metrics = (0xFFFF, -0x8000)
     for glyph in glyphOrder:
         vhAdvances = [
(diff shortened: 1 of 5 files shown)


anthrotype:
actually, since UFO doesn't let one specify the sidebearings directly (they are computed from the outlines), it's best to use (0xFFFF, 0) as sentinel, with the sidebearings being 0, which is what an empty glyph would normally get.

justvanrossum:
> actually, since UFO doesn't let one specify the sidebearings directly (they are computed from the outlines), it's best to use (0xFFFF, 0) as sentinel, with the sidebearings being 0, which is what an empty glyph would normally get.

Hm, I was assuming this would be a hack between ufo2ft and fonttools, for the intermediate otf/ttf master files only, and not be visible at the UFO level: the glyph would simply not exist there.

anthrotype:
what if you want to have a composite glyph where you variate the component offsets without it participating in the HVAR/VVAR? you could add the base component glyphs as empty glyphs (availing yourself of the rule that non-default empty glyphs are assumed to be missing/non-participating), but you would be forced to have some non-empty hmtx

Issue #3236, opened 2023-08-03
Roundtripping NotoColorEmoji-Regular.ttf makes COLR table grow


behdad:
COLR table of original font is 1058175 bytes. Just roundtripping it through TTX makes it grow by 20kb to 1079817.

I have no explanation for this. COLRv1 uses all 24 or larger offsets types, so all object-sharing should happen all the same. This is very unsettling to me currently. :-)

Font file downloaded from https://fonts.google.com/noto/specimen/Noto+Color+Emoji (Click on "Download family").

behdad:
I checked how many objects each packer is producing, and hb-subset's was about 3000 fewer.

I think I see what's going on. In FontTools packer, each `OTTableWriter` has a member `offsetSize`, which is the width of the offset *to this table*. So FontTools won't share an object that is reached to by different offset-sizes.  The COLR table uses a mix of 32bit and 24bit offsets. I believe that's what's going on here.

PR #3241, opened 2023-08-03
[otBase/packer] Allow sharing tables reached by different offset sizes

behdad:
This makes us match hb-subset now.

Fixes https://github.com/fonttools/fonttools/issues/3236

The patch isn't the cleanest. But I suggest we commit as is. The part that is not clean is that the offsetSize is passed to OTTableWriter first, then removed from it and put in the OffsetToWriter instead...

I might try cleaning it up.


behdad:
> The patch isn't the cleanest. But I suggest we commit as is. The part that is not clean is that the offsetSize is passed to OTTableWriter first, then removed from it and put in the OffsetToWriter instead...
> 
> I might try cleaning it up.

Cleaned up.

anthrotype (approved):
LGTM thanks! And good catch!

PR #3248, opened 2023-08-07
Allow for merging of NULL anchors

### LINKED ISSUE #3247: [merger] Cursive attachment merging fails when all anchors are NULL
Carried over from https://github.com/googlefonts/fontmake/issues/1022.

When all anchors in all masters are of the form:

```
    pos cursive _0 <anchor NULL> <anchor 25 0>;
```

the merger creates an ExitEntryAnchor like this:

```python
        rec = ot.EntryExitRecord()
        rec.EntryAnchor = ot.Anchor()
        rec.EntryAnchor.Format = 1
        rec.ExitAnchor = ot.Anchor()
        rec.ExitAnchor.Format = 1
        self.EntryExitRecord.append(rec)
```

This means that deep inside `mergeThings`, when trying to merge the entry anchors, we hit this code path:

```python
            if allNone(lst):
                if out is not None:
                    raise FoundANone(self, got=lst)
```

All entry anchors are None, but the target entry anchor is an `Anchor` so merging fails.

simoncozens:
- A failing test for #3247
- Fix up NULL entry/exit records on merge, fixes #3247

### DIFF
--- a/Lib/fontTools/varLib/merger.py
+++ b/Lib/fontTools/varLib/merger.py
@@ -936,6 +936,15 @@ def merge(merger, self, lst):
     self.EntryExitCount = len(self.EntryExitRecord)
 
 
+@AligningMerger.merger(ot.EntryExitRecord)
+def merge(merger, self, lst):
+    if all(master.EntryAnchor is None for master in lst):
+        self.EntryAnchor = None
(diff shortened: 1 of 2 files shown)


anthrotype (approved):
LGTM!

behdad:
Oops. Thanks Simon.

Issue #3250, opened 2023-08-07
[colorLib] More optized LayerList builder


behdad:
Currently the code adds base paints to the layer list in order of the glyph index. It would share better if it did that by decreasing number of layers.  The code is in `buildColrV1`:
```
    for baseGlyph, paint in colorGlyphItems: 
        try: 
            baseGlyphs.append(buildBaseGlyphPaintRecord(baseGlyph, layerBuilder, paint)) 
    
        except (ColorLibError, OverflowError, ValueError, TypeError) as e: 
            errors[baseGlyph] = e 
```

PR #3251, opened 2023-08-09
[colorLib] Sort paints by decreasing number of layers when encoding

behdad:
Something like this. Untested. Trying to test now.

Fixes https://github.com/fonttools/fonttools/issues/3250

### DIFF
--- a/Lib/fontTools/colorLib/builder.py
+++ b/Lib/fontTools/colorLib/builder.py
@@ -629,22 +629,33 @@ def buildColrV1(
     *,
     allowLayerReuse: bool = True,
 ) -> Tuple[Optional[ot.LayerList], ot.BaseGlyphList]:
-    if glyphMap is not None:
-        colorGlyphItems = sorted(
-            colorGlyphs.items(), key=lambda item: glyphMap[item[0]]
-        )
-    else:
-        colorGlyphItems = colorGlyphs.items()
+    if glyphMap is None:
+        glyphMap = {k: i for i, k in enumerate(colorGlyphs)}
 
     errors = {}
     baseGlyphs = []
     layerBuilder = LayerListBuilder(allowLayerReuse=allowLayerReuse)
-    for baseGlyph, paint in colorGlyphItems:
+
+    for baseGlyph, paint in sorted(
+        colorGlyphs.items(),
+        key=lambda item: (
+            -len(item[1][1])
+            if isinstance(item[1], collections.abc.Sequence)
+            and item[1][0] == ot.PaintFormat.PaintColrLayers
+            else -len(item[1]["Layers"])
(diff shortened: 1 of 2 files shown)


behdad:
```
-rw-r--r--. 1 behdad behdad  4568244 Aug  9 13:10 NotoColorEmoji.ttf.after
-rw-r--r--. 1 behdad behdad  4590708 Aug  9 12:44 NotoColorEmoji.ttf.before
```

behdad:
Saves some 20kb.

behdad:
Wait. My measurement might be flawed.

behdad:
In my testing with NotoColorEmoji-Regular, it saved a meager 824 bytes. Still, it's the right thing to do.

Lib/fontTools/colorLib/builder.py
```diff
@@ -639,7 +639,11 @@ def buildColrV1(
     errors = {}
     baseGlyphs = []
     layerBuilder = LayerListBuilder(allowLayerReuse=allowLayerReuse)
-    for baseGlyph, paint in colorGlyphItems:
+    for baseGlyph, paint in sorted(
+        colorGlyphItems,
+        key=lambda item: len(getattr(item[0], "Layers", [])),
```
anthrotype, line 644: instead of reverse=True you could add a `-` minus sign to the `len`, no?

anthrotype:
The BaseGlyphList must be sorted by glyphID to be able to do the binary search, remember?

https://learn.microsoft.com/en-us/typography/opentype/spec/colr#baseglyph-and-layer-records

> The BaseGlyph records must be sorted in increasing glyphID order. It is assumed that a binary search can be used to find a matching BaseGlyph record for a specific glyphID.

If you really want to build the LayerList in that decreasing length order, you can do that but then you need to make sure that the baseGlyphs are sorted by glyphOrder at the end

behdad:
You are right. I'll fix up.

behdad:
@anthrotype PTAL.

anthrotype (approved):

anthrotype:
would also be nice to have a test that exercise this if not much trouble

behdad:
> would also be nice to have a test that exercise this if not much trouble

Good thing I did. Implementation was wrong. Building noto color emoji now to get new numbers.

behdad:
@anthrotype PTAL again. Sorry about that earlier.

behdad:
Humm. I don't see this making any difference in NotoColorEmoji. Investigating.

behdad:
> Humm. I don't see this making any difference in NotoColorEmoji. Investigating.

I fixed the culprit. However, now the font *grows* by a few hundred bytes. Maybe status quo is good enough.

PR #3258, opened 2023-08-16
Increase glyph name character limit to 127

eliheuer:
As discussed here: https://github.com/googlefonts/ufo2ft/issues/588

I tested this locally and confirm this change allowed `fontmake` to build an Arabic font with long names that was failing due to the 63 character limit.


khaledhosny:
Superseded by #3424

PR #3268, opened 2023-08-31
fix the order of prune_unused_regions

qxliu76:
It should be after each vardata optimization. 
Here's a testcase, I used this command `fonttools varLib.instancer --no-overlap-flag --no-recalc-timestamp --no-optimize --output=ft.ttf test/subset/data/fonts/NotoSans-VF.abc.ttf wght=200:600 wdth=80:90 CTGR=20:60`
File used is this: https://github.com/harfbuzz/harfbuzz/blob/main/test/subset/data/fonts/NotoSans-VF.abc.ttf
And generated MVAR table is like below:

    <VarStore Format="1">
      <Format value="1"/>
      <VarRegionList>
        
        
        <Region index="0">
      
      <VarData index="0">
        
        <NumShorts value="0"/>
        
        <VarRegionIndex index="0" value="0"/>
        <VarRegionIndex index="1" value="3"/>
        <Item index="0" value="[-7, 6]"/>
        <Item index="1" value="[-4, 4]"/>
      </VarData>
    </VarStore>

2 regions indexed at 1 and 2 are not used and not pruned. 
This PR will fix it.

### DIFF
--- a/Lib/fontTools/varLib/varStore.py
+++ b/Lib/fontTools/varLib/varStore.py
@@ -633,15 +633,15 @@ def VarStore_optimize(self, use_NO_VARIATION_INDEX=True, quantization=1):
     for k, v in front_mapping.items():
         varidx_map[k] = back_mapping[v] if v is not None else NO_VARIATION_INDEX
 
-    # Remove unused regions.
-    self.prune_regions()
-
     # Recalculate things and go home.
     self.VarRegionList.RegionCount = len(self.VarRegionList.Region)
     self.VarDataCount = len(self.VarData)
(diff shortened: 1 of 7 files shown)


anthrotype (approved):
thanks, makes sense, not sure why it was done that way to begin with. We need to also update the test expectations. I can take care of it if you want

qxliu76:
> thanks, makes sense, not sure why it was done that way to begin with. We need to also update the test expectations. I can take care of it if you want

I have updated tests accordingly. Thanks!

anthrotype:
Thanks 👍

Issue #3275, opened 2023-09-15
Commandline module to add compile avar(2) and add to existing font


behdad:
It can use designspaceLib to load a .designspace file, then use the code already in varLib to build the avar table and add to existing file. This makes tweaking and adjusting avar2 table much faster than having to recompile the font all the time.

PR #3276, opened 2023-09-15
[varLib.avar] New module to compile avar from .designspace file

behdad:
Fixes https://github.com/fonttools/fonttools/issues/3275

### DIFF
--- a/Lib/fontTools/varLib/avar.py
+++ b/Lib/fontTools/varLib/avar.py
@@ -1,0 +1,70 @@
+from fontTools.varLib import _add_avar, load_designspace
+from fontTools.misc.cliTools import makeOutputFileName
+import logging
+
+log = logging.getLogger("fontTools.varLib.avar")
+
+
+def main(args=None):
+    """Add `avar` table from designspace file to variable font."""
+
+    if args is None:
+        import sys
+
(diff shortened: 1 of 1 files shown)


Lib/fontTools/varLib/avar.py
```diff
@@ -0,0 +1,72 @@
...
+        "designspace", metavar="family.designspace", help="Designspace file."
+    )
+    parser.add_argument(
+        "-o",
+        "--output-file",
+        type=str,
+        help="Output font file name.",
+    )
+
+    logging_group = parser.add_mutually_exclusive_group(required=False)
```
anthrotype, line 36: mutually exclusive group but contains only one option as far as I can see..

anthrotype (approved):

Lib/fontTools/varLib/avar.py
behdad: Humm, yeah. There wasn't a point to add `--quiet`. So I just remove the `logging_group` completely?

anthrotype: ok

behdad: Fixed.

PR #3281, opened 2023-09-27
Pad the OS/2 achVendID value with spaces per the OpenType specification.

### LINKED ISSUE #3280: OpenType Tag padding
I discovered a strange discrepancy in the OS/2 `achVendID` value between a font generated through the fontmake toolchain (fontmake -> ufo2ft -> fontTools) and other font editors. When the vendor ID is shorted than 4 characters, fontTools pads it with `null` (0x00) bytes, whereas other editors pad the value with a space (0x20).

The OpenType specification has this to say on padding (emphasis mine):

> A Tag value is a uint8 array. Each byte within the array must have a value in the range 0x20 to 0x7E. This corresponds to the range of values of Unicode Basic Latin characters in UTF-8 encoding, which is the same as the printable ASCII characters. As a result, a Tag value can be re-interpreted as a four-character sequence, which is conventionally how they are referred to. Formally, however, the value is a byte array. When re-interpreted as characters, the Tag value is case sensitive. It must have one to four non-space characters, _padded with trailing spaces_ (byte value 0x20). A space character cannot be followed by a non-space character.

Based on that, I believe the fontTools behaviour to be incorrect. It appears the [vendor ID is packed](https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/ttLib/tables/O_S_2f_2.py#L70) with the following settings:

```python
	achVendID:     4s      # font vendor identification
```

The [Python documentation for `struct.pack`](https://docs.python.org/3/library/struct.html) explains that values are padded with `null` bytes:

> For packing, the string is truncated or padded with null bytes as appropriate to make it fit. 

There are a couple of other places in the code where this happens from what I can tell:

* The `sfnt` table directory entries
* The `fvar` tags
* The `sbix` glyph data
* The `meta` table

What is the preferred fix here (if any)? I could open a PR to fix just the vendor ID (by pre-padding the value with a space), but this feels like a larger problem.

bramstein:
Fixes #3280.

### DIFF
--- a/Lib/fontTools/ttLib/tables/O_S_2f_2.py
+++ b/Lib/fontTools/ttLib/tables/O_S_2f_2.py
@@ -165,6 +165,10 @@ class table_O_S_2f_2(DefaultTable.DefaultTable):
                 "OS/2 table version 4 and up: version %s",
                 self.version,
             )
+
+        if self.achVendID:
+            self.achVendID = self.padVendorID(self.achVendID)
+
         self.panose = sstruct.pack(panoseFormat, self.panose)
         if self.version == 0:
             data = sstruct.pack(OS2_format_0, self)
@@ -353,6 +357,11 @@ class table_O_S_2f_2(DefaultTable.DefaultTable):
                 avg_width = otRound(sum(widths) / len(widths))
         self.xAvgCharWidth = avg_width
         return avg_width
+
+    def padVendorID(self, achVendorID):
+        """Pad the Vendor ID with spaces as required for Tag types
+        in the OpenType specification."""
+        return achVendorID.ljust(4)
 
 
 # Unicode ranges data from the OpenType OS/2 table specification v1.7
--- a/Tests/ttLib/tables/O_S_2f_2_test.py
+++ b/Tests/ttLib/tables/O_S_2f_2_test.py
@@ -55,6 +55,11 @@ class OS2TableTest(unittest.TestCase):
             (set(range(123)) - {9, 57, 122}),
         )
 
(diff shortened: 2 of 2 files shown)


Lib/fontTools/ttLib/tables/O_S_2f_2.py
```diff
@@ -354,6 +358,11 @@ def recalcAvgCharWidth(self, ttFont):
         self.xAvgCharWidth = avg_width
         return avg_width
 
+    def padVendorID(self, achVendorID):
```
anthrotype, line 361: I don't think we need to have this staticmethod here, calling ljust inline should to the trick

anthrotype:
but what sets that field? should that be fixed to do the padding instead? In this case, I'd prefer if fonttools OS/2 table reader/writer keep whatever was there to begin with, rather than fix things up upon compile...

Lib/fontTools/ttLib/tables/O_S_2f_2.py
bramstein: That was mostly to make it easier to test rather than instantiating the entire table through FontBuilder.

anthrotype:
yeah, I think this is not the job of fonttools, I prefer we pad with spaces in ufo2ft instead

bramstein:
@anthrotype The field is set by [outlineCompiler.py in ufo2ft](https://github.com/googlefonts/ufo2ft/blob/main/Lib/ufo2ft/outlineCompiler.py#L541). I thought it would be the job of the library that generates the binary data to ensure it is correctly written.

PR #3286, opened 2023-10-05
Better OTL builder errors

simoncozens:
I'm forever getting errors like this:

```
fontmake: Error: In 'master_ufo/Foo.designspace': Generating fonts from Designspace failed: 'j'
```

which are not informative. This is slightly more informative:

```
fontmake: Error: In 'master_ufo/Foo.designspace': Generating fonts from Designspace failed:
 master_ufo/Foo-Regular-18pt.ufo/features.fea:1865:5: Could not find glyph 'j' in font
```

### DIFF
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -55,7 +55,11 @@ def buildCoverage(glyphs, glyphMap):
     if not glyphs:
         return None
     self = ot.Coverage()
-    self.glyphs = sorted(set(glyphs), key=glyphMap.__getitem__)
+    try:
+        self.glyphs = sorted(set(glyphs), key=glyphMap.__getitem__)
+    except KeyError as e:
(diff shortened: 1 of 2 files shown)


Lib/fontTools/otlLib/builder.py
```diff
@@ -55,7 +55,11 @@ def buildCoverage(glyphs, glyphMap):
     if not glyphs:
         return None
     self = ot.Coverage()
-    self.glyphs = sorted(set(glyphs), key=glyphMap.__getitem__)
+    try:
+        self.glyphs = sorted(set(glyphs), key=glyphMap.__getitem__)
+    except KeyError as e:
+        raise ValueError(f"Could not find glyph {e} in font")
```
anthrotype, line 61: I suggest to add `from e` so it gets chained instead of eaten up

anthrotype (approved):

Issue #3287, opened 2023-10-05
`TypeError` from `_intSecAtan` on previously valid arguments (valid before 4.40.0)


stenson:
Hi fontTools — long time fan, first time issue reporter — apologies if this has already been reported, but I've been using fontPen’s `FlattenPen` in coordination with ttf variable fonts for a number of years now with no issue, but when I recently updated a fontTools install (as a requirement of a fontmake install), I started getting this error whenever I tried to use the FlattenPen on glyphs with quadratic curves.

After some investigating, I narrowed the issue down to FlattenPen calling `calcQuadraticArcLength` and then, in turn, that function calling `_intSecAtan` — fontTools 4.39.0 works fine for all of my use cases, but 4.40.0 introduces the issue.

Here's an example traceback, with a minimal example (a direct call to `calcQuadraticArcLength`):

```
Traceback (most recent call last):
  File "/Users/robstenson/Type/flattentest/test_minimal.py", line 7, in <module>
    print(calcQuadraticArcLength((210, 333), (289, 333), (326.5, 290.5)))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "Lib/fontTools/misc/bezierTools.py", line 183, in fontTools.misc.bezierTools.calcQuadraticArcLength
  File "Lib/fontTools/misc/bezierTools.py", line 233, in fontTools.misc.bezierTools.calcQuadraticArcLengthC
  File "Lib/fontTools/misc/bezierTools.py", line 148, in fontTools.misc.bezierTools._intSecAtan
TypeError: must be real number, not complex
```

Not sure if it's of any value, but I put together a small repo to demonstrate the issue here: https://github.com/stenson/flattentest. My main issue is demonstrated in test.py, with this code snippet:

```python
from fontTools.ttLib import TTFont
from fontTools.pens.recordingPen import RecordingPen
from fontPens.flattenPen import FlattenPen

glyph = TTFont("MutatorSans.ttf").getGlyphSet()["B"]

rp = RecordingPen()
fp = FlattenPen(rp, approximateSegmentLength=5)
glyph.draw(fp)
```

As far as I know, I'm not doing anything out of the ordinary here, but I get the `TypeError` with all recent versions of fontTools.

anthrotype:
Seems like a regression, thanks for the report, I will take a look at it today

justvanrossum:
[...]

PR #3288, opened 2023-10-06
bezierTools: fix incorrectly typed cython local variable

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/3287


anthrotype:
@stenson I can also confirm that with this patch your https://github.com/stenson/flattentest passes

justvanrossum (approved):

Issue #3289, opened 2023-10-06
`FontBuilder` can't  `setupVerticalHeader()`


TakWolf:
fonttools==4.43.0

```python
builder = FontBuilder(2048, isTTF=is_ttf)

# some setup

builder.setupVerticalHeader(
    ascent=ascent,
    descent=descent,
)

# other setup

builder.save(file_path)
```

Cause error:

```
DEBUG:fontTools.ttLib.ttFont:Compiling 'maxp' table
DEBUG:fontTools.ttLib.ttFont:Writing 'maxp' table to disk
DEBUG:fontTools.ttLib.ttFont:Compiling 'CFF ' table
DEBUG:fontTools.ttLib.ttFont:Writing 'CFF ' table to disk
DEBUG:fontTools.ttLib.ttFont:Compiling 'head' table
DEBUG:fontTools.ttLib.ttFont:Writing 'head' table to disk
DEBUG:fontTools.ttLib.ttFont:Compiling 'hmtx' table
DEBUG:fontTools.ttLib.ttFont:Writing 'hmtx' table to disk
DEBUG:fontTools.ttLib.ttFont:Compiling 'hhea' table
DEBUG:fontTools.ttLib.ttFont:Writing 'hhea' table to disk
DEBUG:fontTools.ttLib.ttFont:Compiling 'OS/2' table
DEBUG:fontTools.ttLib.ttFont:Writing 'OS/2' table to disk
DEBUG:fontTools.ttLib.ttFont:Compiling 'name' table
DEBUG:fontTools.ttLib.ttFont:Writing 'name' table to disk
DEBUG:fontTools.ttLib.ttFont:Compiling 'cmap' table
DEBUG:fontTools.ttLib.ttFont:Writing 'cmap' table to disk
DEBUG:fontTools.ttLib.ttFont:Compiling 'post' table
DEBUG:fontTools.ttLib.ttFont:Writing 'post' table to disk
DEBUG:fontTools.ttLib.ttFont:Compiling 'vhea' table
Traceback (most recent call last):
  File "E:\Projects\Python\pixel-font-builder\examples\demo.py", line 183, in <module>
    main()
  File "E:\Projects\Python\pixel-font-builder\examples\demo.py", line 168, in main
    builder.save_otf(os.path.join(build_dir, 'demo.otf'))
  File "E:\Projects\Python\pixel-font-builder\src\pixel_font_builder\font.py", line 68, in save_otf
[...]

PR #3290, opened 2023-10-06
Fix UnbouldLocalError and let {h,v}hea be no-op when no {v,h}mtx is present

anthrotype:
I believe that was the spirit of this code, do have recalc() do nothing if the metrics tables are missing

Fixes https://github.com/fonttools/fonttools/issues/3289

### DIFF
--- a/Lib/fontTools/ttLib/tables/_h_h_e_a.py
+++ b/Lib/fontTools/ttLib/tables/_h_h_e_a.py
@@ -67,9 +67,11 @@ class table__h_h_e_a(DefaultTable.DefaultTable):
         return sstruct.pack(hheaFormat, self)
 
     def recalc(self, ttFont):
-        if "hmtx" in ttFont:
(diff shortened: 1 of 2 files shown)


anthrotype:
I wonder if fontBuilder should issue a warning (or even an error) if one attempts to call setup{Horizontal,Vertical}Header before setup{Horizontal,Vertical}Metrics?

anthrotype:
or maybe it's not about the order in which the methods are called but the requirement that both Header and Metrics methods are called before compiling the font (which is when the recalc() is attempted)

justvanrossum (approved):

justvanrossum:
Yes, I also don't think it's about the order. I'm not sure how fontbuilder could enforce such things.

anthrotype:
> I'm not sure how fontbuilder could enforce such things.

yeah, maybe just documenting it better would work

PR #3298, opened 2023-10-12
designspaceLib: Accept empty condition sets

madig:
Related: https://github.com/linebender/norad/issues/327


anthrotype (approved):

anthrotype:
evaluateConditions already returns True for empty conditions so this is just making sure the spec matches current implementation

https://github.com/fonttools/fonttools/blob/98242634c45cf4ca9f8aff6fd059de9dcf734471/Lib/fontTools/designspaceLib/__init__.py#L406-L422

OpenType FeatureVariations spec similarly says empty ConditionSet matches always

https://learn.microsoft.com/en-us/typography/opentype/spec/chapter2#conditionset-table

PR #3310, opened 2023-10-23
[varStore] Handle >65535 items per encoding

behdad:
By creating a new major for each 65535.

I'll think about how to add test...

### DIFF
--- a/Lib/fontTools/varLib/varStore.py
+++ b/Lib/fontTools/varLib/varStore.py
@@ -128,11 +128,11 @@ def VarData_addItem(self, deltas, *, round=round):
     countUs = self.VarRegionCount
     countThem = len(deltas)
     if countUs + 1 == countThem:
-        deltas = tuple(deltas[1:])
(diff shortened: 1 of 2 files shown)


Lib/fontTools/varLib/varStore.py
```diff
@@ -95,7 +95,8 @@ def storeMasters(self, master_values, *, round=round):
         return base, self.storeDeltas(deltas, round=noRound)
 
     def storeDeltas(self, deltas, *, round=round):
-        deltas = [round(d) for d in deltas]
+        if round != noRound:
```
anthrotype, line 98: why this? if round == noRound then it's no-op, right? not sure if this micro-optimization is worth it

anthrotype (approved):

Lib/fontTools/varLib/varStore.py
behdad: Sure I can drop that.

PR #3314, opened 2023-10-27
[subset] fix another null-offset error in MarkLigPos

qxliu76:
Similar to this commit: https://github.com/fonttools/fonttools/commit/1d0f3c27fcefa4f3265afc04eebf10360a0eed0c


Lib/fontTools/subset/__init__.py
```diff
@@ -890,6 +890,7 @@ def prune_post_subset(self, font, options):
             if m.MarkAnchor:
                 m.MarkAnchor.prune_hints()
         for l in self.LigatureArray.LigatureAttach:
+            if l is None: continue
```
anthrotype, line 893: our autoformatter doesn'rt like if statements in a single line

anthrotype (approved):
LGTM pending reformatting

PR #3319, opened 2023-10-30
[instancer] Check if an axis has values before requiring it to be found

### LINKED ISSUE #3318: [instancer] --update-name-tables unnecessarily strict?
In order to update the name tables, the instancer looks in the STAT table and expects to see an entry for every axis and every axis value which is being instantiated. I wonder if this is overly strict; my feeling is that if there is an axis which doesn't affect style naming at all, it does not need to be in the STAT table, and therefore the instancer should be able to ignore that axis when it comes to generating the style names. But maybe there is a good reason for this behaviour.

simoncozens:
Fixes #3318.


anthrotype (approved):
lgtm, thanks. bonus points if you also add a little test

Issue #3322, opened 2023-11-01
[instancer] Allow not specifying any of `min:default:max` values.


behdad:
It would be nice to be able to just change default, or just change max, etc, without having to find and spell out the rest.

Proposed syntax: allow not specifying any of them. Ie. `wght=:500:` will change the default only. `wght=::700` will change the max only. `wght=300::700` is the same as `wght=300:700` in which the default is not changed.

behdad:
cc @arrowtype

arrowtype:
That sounds nice! I could see it being especially useful, perhaps, for adjusting a collection of fonts (e.g. if a font publisher wanted to add a wght=400 default to a bunch of variable fonts).

> `wght=::700` will change the max only.

I would somewhat expect `wght=:700` to work as well (with only a single colon). Would that make sense, or not really?

behdad:
> I would somewhat expect `wght=:700` to work as well (with only a single colon). Would that make sense, or not really?

Good point. I'll make that work as well.

PR #3323, opened 2023-11-01
[instancer] Allow relaxed limits syntax

behdad:
Fixes https://github.com/fonttools/fonttools/issues/3322

cc @garretrieger @qxliu76

### DIFF
--- a/Lib/fontTools/varLib/instancer/__init__.py
+++ b/Lib/fontTools/varLib/instancer/__init__.py
@@ -139,26 +139,36 @@ def NormalizedAxisRange(minimum, maximum):
 class AxisTriple(Sequence):
     """A triple of (min, default, max) axis values.
 
-    The default value can be None, in which case the limitRangeAndPopulateDefault()
-    method can be used to fill in the missing default value based on the fvar axis
-    default.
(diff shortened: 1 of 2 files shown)


Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -1326,29 +1340,27 @@ def parseLimits(limits: Iterable[str]) -> Dict[str, Optional[AxisTriple]]:
...
             result[tag] = None
             continue
 
-        result[tag] = AxisTriple(lbound, default, ubound)
+        triple = match.group(3, 4, 5)
+
+        if triple[1] is None:  # "value" syntax
+            triple = (triple[0], triple[0], triple[0])
+        elif triple[2] is None:  # "min:max" syntax
+            triple = (triple[0], None, triple[1])
```
anthrotype, line 1359: this `if` is not immediately obvious to me..
maybe can you also add tests for the argument parsing?

anthrotype: ok I did https://github.com/fonttools/fonttools/pull/3323/commits/27c79679eb947a22a2ee79a60526cb0f95b9376b

anthrotype (approved):

PR #3340, opened 2023-11-16
[requirements] Require pycairo

behdad:


behdad:
Is this of any use? @anthrotype

anthrotype:
I don't know, is the new interpolatabelPlot.py tested? It doesn't look like to me. Also, the fact that pycairo could be installed successfully on the Github CI runners doesn't mean that it will work on a user's machine (without C compilers, libcairo with headers, pkg-config, etc.).
I'm ok to add this to requirements.txt if you do add some tests that exercise it

behdad:
Closing until I add tests for it.

PR #3341, opened 2023-11-17
Interpolatable mst

behdad:
This forms a minimum-spanning-tree over the master locations and compare two masters connected in the resulting tree. This reduces false-positives. Previously we were comparing each master to the first/base master.

### DIFF
--- a/Lib/fontTools/varLib/interpolatable.py
+++ b/Lib/fontTools/varLib/interpolatable.py
@@ -216,12 +216,13 @@ def test_gen(
         log.info("Parents: %s", parents)
 
     def grand_parent(i, glyphname, skips={}):
-        if i is None: return None
+        if i is None:
+            return None
         i = parents[i]
-        if i is None: return None
-        while (
-            parents[i] is not None and
-            (glyphsets[i][glyphname] is None or i in skips)
+        if i is None:
+            return None
+        while parents[i] is not None and (
+            glyphsets[i][glyphname] is None or i in skips
         ):
             i = parents[i]
         return i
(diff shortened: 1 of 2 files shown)


behdad:
I'm not sure if the last commit here is an improvement or not.

behdad:
Doesn't seem to work / not worth it.

behdad:
This seems to have some potential. At least for "wrong_start_point" reopening to reconsider.

behdad:
Reporting findings so far:
- Increases RobotoFlex issues from 38 to 41.
- Decreases Handjet from over 500 false-positives to 120.
- Decreases Honk from over 500 false-positives to 55.

I'm, still improving on it to see if I can RobotoFlex report back down to 38.
The rest of the false-positives mostly have to do start-point detection when things are rotating or otherwise skewing a lot. I'll work on an improved algorithm.

behdad:
> * Increases RobotoFlex issues from 38 to 41.
> * Decreases Handjet from over 500 false-positives to 120.
> * Decreases Honk from over 500 false-positives to 55.
> 
> I'm, still improving on it to see if I can RobotoFlex report back down to 38. The rest of the false-positives mostly have to do start-point detection when things are rotating or otherwise skewing a lot. I'll work on an improved algorithm.

Improving the start-point algorithm brought them down to:
- RobotoFlex 39
- Handjet 72
- Honk 51

behdad:
I think this is ready to be merged. @anthrotype?

behdad:
Going to merge this since @anthrotype typically says "It's your code".  Testing the improvements is rather cumbersome.

PR #3343, opened 2023-11-17
[interpolatable] WIP use uharfbuzz if available

behdad:
For faster drawing. However, I'm hitting a lot of
node-count-differs errors.  I have not tracked the cause down yet.

I thought this line might be:

https://github.com/harfbuzz/harfbuzz/blob/df635ab78ae1d397396e3a0fe688b8a4674b2e8a/src/hb-draw.hh#L148

But disabling it in uharfbuzz didn't help.

### DIFF
--- a/Lib/fontTools/varLib/interpolatable.py
+++ b/Lib/fontTools/varLib/interpolatable.py
@@ -428,13 +428,25 @@ def test(*args, **kwargs):
     return problems
 
 
-def recursivelyAddGlyph(glyphname, glyphset, ttGlyphSet, glyf):
+def recursivelyAddGlyph(glyphname, glyphset, ttGlyphSet, hbFont, glyf, reverseGlyphMap):
     if glyphname in glyphset:
         return
-    glyphset[glyphname] = ttGlyphSet[glyphname]
 
-    for component in getattr(glyf[glyphname], "components", []):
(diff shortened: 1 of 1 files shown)


behdad:
@khaledhosny Any help here is appreciated.

behdad:
Reopening to investigate so at least we know who's stealing a point from the outline!

anthrotype:
you need to explicitly emit the implied closing line otherwise a duplicate line segment at the end of a contour can get eaten up, see

https://github.com/fonttools/fonttools/blob/9c154e81c26dcd94622e01a2f984256778096055/Lib/fontTools/pens/pointPen.py#L231-L247

anthrotype:
and that's because SegmentToPointPen always pops the last node when path is closed and last == first

https://github.com/fonttools/fonttools/blob/9c154e81c26dcd94622e01a2f984256778096055/Lib/fontTools/pens/pointPen.py#L321-L323

so to compensate and make sure points are roundtripped when going points => segments => points you need to make sure the segment representation has explicit closing lineTos

behdad:
Yeah but doing that in HB didn't seem to help my problem. I'll investigate more.

behdad:
The fonttools `RecordingPen` and draw() work fine for me. HB is doing something different.

PR #3346, opened 2023-11-20
[interpolatable] Return master indices in addition to master names

behdad:
Using names to refer to masters is weird. Return the index and the user can look up the name.

This changes the API in two ways:
- `test` and `test_gen` do not accept a `names` parameter anymore. If needed I can put it back there unused.
- `master`, `master1`, `master2` in the problem reports are now indices, not names. If this is undesirable, I can add separate `master_idx`, `master1_idx`, `master2_idx` instead. Maybe that's better.


behdad:
Nevermind. I'll make a non-breaking change.

behdad:
PTAL!

anthrotype (approved):

PR #3348, opened 2023-11-21
[ttGlyphSet] Add option to not recalculate glyf bounds

behdad:
Avoids unnecessary work in the interpolator.

What do you think?


anthrotype:
yeah, better to avoid unnecessary work if you know in advance things are already correct (or assumed correct)

anthrotype (approved):

behdad:
In this case it's not correctness, just that I don't use the glyph xMin et al.

Thanks.

behdad:
One could say the glyphSet() API doesn't support xMin etc anyway.

Issue #3352, opened 2023-11-27
Curve-curve intersection edge cases


simoncozens:
These two "line-like" cubics have an intersection point at (0,0.5), but `curveCurveIntersections` does not find it:

```python
from fontTools.misc.bezierTools import lineLineIntersections, curveCurveIntersections

seg1 = [(0.0, 0.0), (0.0, 0.25), (0.0, 0.75), (0.0, 1.0)]
seg2 = [(0.0, 0.5), (0.25, 0.5), (0.75, 0.5), (1.0, 0.5)]

print(lineLineIntersections(seg1[0], seg1[-1], seg2[0], seg2[-1]))
# [Intersection(pt=(0.0, 0.5), t1=0.5, t2=0.0)]

print(curveCurveIntersections(seg1, seg2))
# []
```

simoncozens:
In this case, we are failing at the first test (bounding box intersection between the two curves), so we go home early and don't do any subdivision test:

```python
seg1 = [(0.0, 0.0), (0.0, 0.25), (0.0, 0.75), (0.0, 1.0)]
seg2 = [(0.0, 0.5), (0.25, 0.5), (0.75, 0.5), (1.0, 0.5)]

bounds1 = _curve_bounds(seg1)
bounds2 = _curve_bounds(seg2)

print(bounds1, bounds2)
# (0.0, 0.0, 0.0, 1.0) (0.0, 0.5, 1.0, 0.5)

print(sectRect(bounds1, bounds2))
# (False, (0, 0, 0, 0))
```

anthrotype:
good catch! would it be enough to special-case this edge case when both curves's bounds are flattened to perperdicular lines?

simoncozens:
What I would suggest is that if a curve is sufficiently line-like (for some definition...), it gets turned into a line, and then either `curveLineIntersections` or `lineLineIntersections` handles it. This'll be both faster and more correct.

anthrotype:
maybe it's our `sectRect` that's at fault.

In the classic "Graphics Gems" IV book, the chapter about curve intersections, they have the code below in the routine that checks two beziers' bounds intersection (in particular, see the comment ` // Not >= : need boundary case`)

https://web.archive.org/web/20130613230901/http://users.soe.ucsc.edu/~pang/160/f98/Gems/GemsIV/curve_isect/Bezier.cc

```c
    // Test bounding box of b against bounding box of a
[...]

PR #3354, opened 2023-11-27
[arrayTools] fix sectRect boundary case

anthrotype:
use '>' instead of '>=' when comparing intersection of two rectangles, so we don't skip boundary case when the two touch

Fixes https://github.com/fonttools/fonttools/issues/3352

### DIFF
--- a/Lib/fontTools/misc/bezierTools.py
+++ b/Lib/fontTools/misc/bezierTools.py
@@ -1395,7 +1395,7 @@ def curveCurveIntersections(curve1, curve2):
         >>> len(intersections)
         3
         >>> intersections[0].pt
-        (81.7831487395506, 109.88904552375288)
+        (81.7904225872997, 109.89939633675462)
     """
     if _is_linelike(curve1):
         line1 = curve1[0], curve1[-1]
@@ -1434,7 +1434,7 @@ def segmentSegmentIntersections(seg1, seg2):
         >>> len(intersections)
         3
         >>> intersections[0].pt
-        (81.7831487395506, 109.88904552375288)
+        (81.7904225872997, 109.89939633675462)
         >>> curve3 = [ (100, 240), (30, 60), (210, 230), (160, 30) ]
         >>> line  = [ (25, 260), (230, 20) ]
         >>> intersections = segmentSegmentIntersections(curve3, line)


justvanrossum (approved):
I notice I've done the same change in my JS port in Fontra, so that makes me +1.

anthrotype:
@simoncozens I noticed that when making this change to sectRect, from no intersections, we now find 4 intersections for the test curves (from https://github.com/fonttools/fonttools/pull/3354/commits/dad5bc6cdac12a2fe5ff90e424f12db613db8125) that touch exactly at t=0.5, but they look like duplicates.

```
[Intersection(pt=(49.853515811264515, 74.99971389770508), t1=0.4990234375, t2=0.4990234375),
 Intersection(pt=(50.146484188735485, 74.99971389770508), t1=0.5009765625, t2=0.4990234375),
 Intersection(pt=(49.853515811264515, 74.99971389770508), t1=0.4990234375, t2=0.5009765625),
 Intersection(pt=(50.146484188735485, 74.99971389770508), t1=0.5009765625, t2=0.5009765625)]
```

I see that `_curve_curve_intersections_t` function has a `precision` parameter set to 1e-3, which is used for two slightly different things:

1) when determining if an intersection was found, by comparing it against the area of the bounds rectangle

```python
    if rectArea(bounds1) < precision and rectArea(bounds2) < precision:
        return [(midpoint(range1), midpoint(range2))]
```

2) when rounding the `t`s in order to prune duplicates:

```python
    unique_key = lambda ts: (round(ts[0] / precision), round(ts[1] / precision))
    seen = set()
    unique_values = []

    for ts in found:
        key = unique_key(ts)
        if key in seen:
            continue
        seen.add(key)
        unique_values.append(ts)
```

Maybe we need to use precision*precision (squared) when checking the areas (given area itself is not a single-dimension dinstance but... an area), so that we then obtain more precise intersections whose ts will get rounded off with given precision and will become equal and deduped?

anthrotype:
basically this https://github.com/fonttools/fonttools/pull/3354/commits/bfd874a802f6d5c2ba46447e54c62cf4020ba08b

simoncozens:
This stuff (any kind of cubic-cubic intersection logic) is horribly complicated and full of floating-point imprecision and edge cases and I've basically hit the limit of my knowledge here. (Actually I hit the limit of my knowledge [*here*](https://github.com/linebender/kurbo/pull/199).) Raph has been working on a [new algorithm for Kurbo](https://github.com/linebender/kurbo/pull/258) but I'm not sure what state that's in.

PR #3356, opened 2023-11-28
Interpolatable structure area

behdad:


behdad:
This is a full day's of work worth. Going to squash and commit since history is dirty.

PR #3358, opened 2023-11-28
Add `public.fontInfo` to designSpaceLib specification

### LINKED ISSUE #2606: Set family name for variable fonts directly in designspace
Currently (as far as I know) any variable font generated from a designspace file will inherit its font family name from the default source. However, especially with the new ability to [generate multiple variable fonts](https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#variable-fonts-element) from a single designspace file, it would be very helpful to be able to explicitly set the name to be something different, similar to how it works for instances.

### LINKED ISSUE #2608: [designspace] Additional info for instances
Hi,

I wonder if this idea has come up before, or if there is a way to do it that I don't know of. If there is, please let me know!

I am wondering if we could extend the concept of [localized names for instances](https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#stylename-familyname-stylemapstylename-stylemapfamilyname-elements-localised-names-for-instances) sub elements of the `<instance>` tag to include more kinds of font info.

Say I have a designspace that I am using to generate static instances from. It has an axis other that Weight, Width, and Slope, so I want to make sure I set the WWS names for each of these instances. And more than that, I want to edit the Panose information, or make sure the trademark info has the individual style name in it. None of that stuff interpolates, so the generated instances would get that information from the source at the origin, which may be wrong in this case.

I can adjust these things after the fonts are generated, but it would be cleaner to have the option to declare UFO font info in the instance, maybe even as children of the `<info>` element to organize them, like so:

```xml
...
<instances>
    <instance familyname="My Font" stylename="Text Light" filename="instances/MyFont-TextLight.ufo" postscriptfontname="MyFont-TextLight" stylemapfamilyname="My Font Text Light" stylemapstylename="regular">
        <location>
            <dimension name="optical" xvalue="6"/>
            <dimension name="weight" xvalue="300"/>
        </location>
        <kerning/>
        <info>
            <openTypeNameWWSFamilyName>My Font Text</openTypeNameWWSFamilyName>
            <openTypeNameWWSFamilyName xml:lang="de">Meine Schrift Text</openTypeNameWWSFamilyName>
            <openTypeNameWWSFamilyName xml:lang="fr">Mon Police Texte</openTypeNameWWSFamilyName>
            <openTypeNameWWSSubfamilyName>Light</openTypeNameWWSSubfamilyName>
            <openTypeNameWWSSubfamilyName xml:lang="de">Leicht</openTypeNameWWSSubfamilyName>
            <openTypeNameWWSSubfamilyName xml:lang="fr">Léger</openTypeNameWWSSubfamilyName>
            <openTypeOS2Panose>...</openTypeOS2Panose>
            <trademark>My Font Text Light is a registered trademark...</trademark>
            <trademark xml:lang="de">Meine Schrift Text Leicht ist eine registrierte Marke...</trademark>
            <trademark xml:lang="fr">Mon Police Texte Léger est une marque enregistrée...</trademark>
        </info>
    </instance>
</instances>
...
```

The idea would be that build tools would take the info in the instance's `<info>` tag, and apply it after creating the UFO. Anything explicitly declared in the `<info>` tag would overwrite whatever info it inherits from the source at the origin. 

I have no idea if this has been brought up before, I just thought I would make this issue. If this idea seems feasible, I can make a more formal proposal, and/or take a crack at a PR.

Thanks!

Edit: Just saw #2606, maybe this could work with the `<variable-font>` element as well.

### LINKED ISSUE #3272: [designspace] Where to store varible font-specific data
It is common to set different family name and other related name data on variable fonts and static instances, but looking into DesignSpace v5, I see no designated place to store these.

There is `<variable-font>` but it seems to be designed mainly for creating subsets of designspace (to build a variable font, all axis settings needs to be duplicated as any missing axis will be sliced at its default value, if my reading of documentation is correct) and it does not seem to have any of `instance` attributes other than `name` and `filename`.

Should `<variable-font>` be extended to be able to handle this, or this not what it was designed for?

colinmford:
Fixes #2608, Fixes #2606, and Fixes #3272.

This PR was [requested](https://github.com/fonttools/fonttools/issues/2608#issuecomment-1820571580) by @belluzj.

This PR adds the proposed `public.fontInfo` lib item to the `designSpaceLib` specification, with examples as discussed in #2608.

@belluzj @khaledhosny @anthrotype I very much welcome a review and feedback. This is my first time contributing to the designSpaceLib specification and I'm sure I got something wrong here or there. I really believe in the power and flexibility this lib item could provide if it is supported by build tools, so I tried to be as specific as possible in its description.

Thanks!

### DIFF
--- a/Doc/source/designspaceLib/index.rst
+++ b/Doc/source/designspaceLib/index.rst
@@ -120,6 +120,34 @@ any of the UFOs. If the lib key is empty or not present in the Designspace, all
 glyphs should be exported, regardless of what the same lib key in any of the
 UFOs says.
 
+public.fontInfo
+-----------------------
+
+This lib key, when included in the ``<lib>`` element inside an ``<instance>`` or 
+``<variable-font>`` tag, allows for direct manipulation of font info data in 
+instances and variable fonts. The lib value should follow the 
+`UFO3 fontinfo.plist specification <https://unifiedfontobject.org/versions/ufo3/fontinfo.plist/>`_,
+and should functionally appear to be a property list dictionary with the same 
+structure as the ``fontinfo.plist`` file in a UFO.
+
+
+All font info items in the UFO fontinfo.plist specification should be able to 
+be defined in the ``public.fontInfo`` lib. Checking validity of the data using 
+``fontTools.ufoLib.validators`` is recommended but not required. 
+
+A tool generating the font should copy the font info directly from the 
+``public.fontInfo`` lib item to the instance or the variable font, superceding 
+all other settings. All info **not** supplied in ``public.fontInfo`` should be 
+inherited from the name attributes in the ``<instance>`` or ``<variable-font>`` 
+elements (first) and the UFO font info from the source at the origin (second). 
+Absence of a key from the ``public.fontInfo`` lib does **not** mean a that piece 
+of font info should be interpreted as being undefined unless it is also left 
+undefined in the other locations. In the case of information in ``public.fontInfo`` 
+conflicting with supplied name strings in the ``<instance>`` or ``<variable-font>`` 
+elements, or the info supplied from the source at the origin, the data in 
+``public.fontInfo`` should supercede that information because it is understood 
+that the designer is using ``public.fontInfo`` to gain more granular control 
+over the resulting font. 
 
 Implementation and differences
 ==============================
--- a/Doc/source/designspaceLib/xml.rst
+++ b/Doc/source/designspaceLib/xml.rst
@@ -790,7 +790,7 @@ The ``<variable-fonts>`` element contains one or more ``<variable-font>`` elemen
      but the design space is sliced at the given location. *Note:* While valid to have a
      specific value that doesn’t have a matching ``<source>`` at that value, currently there
      isn’t an implentation that supports this. See `this fontmake issue
-     <https://github.com/googlefonts/fontmake/issues/920>`.
+     <https://github.com/googlefonts/fontmake/issues/920>`_.
 
      .. code:: xml
 
@@ -841,6 +841,38 @@ Arbitrary data about this variable font.
 .. versionadded:: 5.0
 
 .. seealso:: :ref:`lib`
+
+Here is an example of using the ``public.fontInfo`` lib key to gain more granular
+control over the font info of a variable font, in this case setting some names to
+reflect the fact that this is a Narrow variable font subset from the larger designspace. 
+This lib key allows font info in variable fonts to be more specific than the font 
+info of the sources.
+
+.. rubric:: Example
+
+.. code:: xml
+
+    <variable-font name="MyFontNarrVF">
+      <axis-subsets>
+        <axis-subset name="Weight"/>
+        <axis-subset name="Width" uservalue="75"/>
+      </axis-subsets>
+      <lib>
+        <dict>
+          <key>public.fontInfo</key>
+          <dict>
+            <key>familyName</key>
+            <string>My Font Narrow VF</string>
+            <key>styleName</key>
+            <string>Regular</string>
+            <key>postscriptFontName</key>
+            <integer>MyFontNarrVF-Regular</integer>
+            <key>trademark</key>
+            <integer>My Font Narrow VF is a registered trademark...</integer>
+          </dict>
+        </dict>
+      </lib>
+    </variable-font>
 
 
 Instances included in the variable font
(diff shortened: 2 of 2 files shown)


Doc/source/designspaceLib/index.rst
```diff
@@ -120,6 +120,34 @@ any of the UFOs. If the lib key is empty or not present in the Designspace, all
 glyphs should be exported, regardless of what the same lib key in any of the
 UFOs says.
 
+public.fontInfo
+-----------------------
+
+This lib key, when included in the ``<lib>`` element inside an ``<instance>`` or 
+``<variable-font>`` tag, allows for direct manipulation of font info data in 
+instances and variable fonts. The lib value should follow the
```
benkiel, line 128: 'should' should be 'must'

benkiel: anything that is a hard requirement should use "must". Should implies that a thing can be optional. @colinmford see https://unifiedfontobject.org/versions/ufo3/#terminology and the linked RFC 2119 there.

benkiel (commented):
See notes about language

colinmford (commented):
changed some instances of "should" to "must" in compliance with the UFO spec and rfc2119

Doc/source/designspaceLib/index.rst
```diff
@@ -120,6 +120,34 @@ any of the UFOs. If the lib key is empty or not present in the Designspace, all
 glyphs should be exported, regardless of what the same lib key in any of the
 UFOs says.
 
+public.fontInfo
+-----------------------
+
+This lib key, when included in the ``<lib>`` element inside an ``<instance>`` or 
+``<variable-font>`` tag, allows for direct manipulation of font info data in 
+instances and variable fonts. The lib value must follow the
```
belluzj, line 128: Should we allow this lib key to also be present in a third place: the root `<lib>` element of the whole designspace?

That would support two use-cases:
* it would work for DS version 4, in that case it would specify the fontinfo of the VF that covers the whole designspace (as in DS v4, one DS == one VF)
* it would one more level of priority when looking up a value:
  1. use the value from `<variable-font>` or `<instance>`'s lib public.fontInfo
  2. use the value from the `<varaiable-font>` or `<instance>` XML attributes for names
  3. (NEW) use the value from the designspace's root `<lib>`'s public.fontInfo
  4. use the value from the source UFO's fontinfo.plist

behdad:
I recently needed to find the UPEM of a designspace. Does this address that? Or I still have to look into the UFOs?

Doc/source/designspaceLib/index.rst
anthrotype: > Should we allow this lib key to also be present in a third place: the root <lib> element of the whole designspace?

yeah +1 on that

Doc/source/designspaceLib/index.rst
```diff
@@ -120,6 +120,34 @@ any of the UFOs. If the lib key is empty or not present in the Designspace, all
...
+and should functionally appear to be a property list dictionary with the same 
+structure as the ``fontinfo.plist`` file in a UFO.
+
+
+All font info items in the UFO fontinfo.plist specification should be able to 
+be defined in the ``public.fontInfo`` lib. Checking validity of the data using 
+``fontTools.ufoLib.validators`` is recommended but not required. 
+
+A tool generating the font should copy the font info directly from the 
+``public.fontInfo`` lib item to the instance or the variable font, superceding
```
anthrotype, line 139: If I am not mistaken, the correct English spelling should be "supersede" with the "s", not "supercede" with "c"

anthrotype:
> I recently needed to find the UPEM of a designspace. Does this address that? Or I still have to look into the UFOs?

in theory, it could, yes. If the public.fontInfo lib key is present and defines unitsPerEm, it would override the one in the sources, like other fontinfo attributes defined in there. However it's unlikely that one would want to use the public.fontInfo key to override unitsPerEm (strictly override i.e. use a different value than the UFO sources have), because that would change the relative scale of all the outlines coordinates. This new proposed lib key is mostly for other kind of metadata that one would like to differ from the default UFO source fontinfo (e.g. names), or for when the same default UFO source is referenced in different variable-font elements.

Anyway, the answer is yes, technically one should check the new public.fontInfo first, following the precendence order i.e. in the global DS.lib (as Jany proposed), then inside the specific variable-font lib, finally inside the default UFO source fontinfo.

We shall probably add some methods to lookup fontinfo attributes for the whole designspace or particular VF/instances within it that take into account this inheritance logic.

colinmford:
@behdad with this you could change the UPM value of the generated fonts to all be a chosen value, I suppose, but as for looking up what it currently is, no I think you would still have to go into the source UFOs

colinmford (commented):
Incorporated feedback into a new commit

colinmford:
@belluzj @anthrotype I updated the lib definition, have a look!

belluzj (approved):
Looks very clear to me, thanks!

Doc/source/designspaceLib/index.rst
```diff
@@ -120,6 +120,41 @@ any of the UFOs. If the lib key is empty or not present in the Designspace, all
...
+be defined in the ``public.fontInfo`` lib. Checking validity of the data using 
+``fontTools.ufoLib.validators`` is recommended but not required. 
+
+All font info items for a variable font or an instance must be inherited using 
+the following order, in order of descending priority:
+
+#. The ``public.fontInfo`` key in the ``<lib>`` element of the ``<variable-font>``
+   or ``<instance>`` elements.
+#. The ``<varaiable-font>`` or ``<instance>`` XML attributes for names (i.e.
+   ``familyname``, ``stylename``, etc.)
```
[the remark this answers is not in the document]
colinmford: @anthrotype a good point, fixed

anthrotype:
Thanks @colinmford - Now we'll need to think about how to implement support for this, in ufo2ft, designspaceLib and varLib

Issue #3004, opened 2023-02-23
"rules process last" should by default assign the conditional subs to both rclt & calt


twardoch:
Do we know why DesignSpace "rules process last" assigns the conditional glyph substitutions only to one feature (rclt) and you can use a lib key to choose another feature like calt? 

It could easily produce a Feature variations lookup and then assign it to several features (like rclt & calt), to maximize engine compatibility. The lookup will be processed once anyway if it's just one lookup.

behdad:
What's the point of adding them to `rclt` if you're also adding to `calt`?

behdad:
Any engine that supports `rclt` also supports `calt`. I suppose the only benefit of adding to both is that on engines that support `rclt`, you wouldn't be able to turn these off by `calt`. So yeah it might be considered an improvement.

anthrotype:
> What's the point of adding them to rclt if you're also adding to calt?

I presume some buggy software refuses to apply rclt but works if feature is named calt.. not making names here.

anthrotype:
yeah, there's a duplicate issue here which explains the rationale of registering to both:

https://github.com/fonttools/fonttools/issues/2050

anthrotype:
this was the conclusion last time we discussed this

> So (if I'm getting this straight), the desire to do both calt and rclt simultaneously is to work around an Adobe bug. I don't see why the designspace format should even be accomodating this. This is a typical case for a gftools-like afterburner tool.

I propose to close unless Adam puts forward more convincing arguments?

twardoch:
Please point me to the resource that indicates that `rclt` should be on by default for all scripts. 

So far I see the [`rclt` description](https://learn.microsoft.com/en-us/typography/opentype/spec/features_pt#rclt) saying: 

> Application interface: In recommended usage, this feature triggers substitutions required for correct display of certain scripts. It should be applied in the appropriate contexts, as determined by script-specific processing requirements. When processing lookups, context before or after the glyph sequence to which the feature is applied must be considered.

The [MS spec for Arabic Script](https://learn.microsoft.com/en-us/typography/script-development/arabic) does list `rclt` among the on-by default features but the [MS spec for Standard Scripts](https://learn.microsoft.com/en-us/typography/script-development/standard) does not. 

If Adobe does not enable `rclt` for all scripts — what’s the basis for claiming that Adobe’s implementation is »buggy«?

behdad:
[...]

Issue #2050, opened 2020-08-22
Feature request: allow designspace files to control both rvrn & rclt, rather than forcing an either/or option


arrowtype:
Currently, the `rules` element in a designspace must _either_ make all substitutions before or after all other opentype feature substitutions. As @anthrotype explained it in https://github.com/fonttools/fonttools/issues/2040:

> the rules in `rvrn` are processed first before any others, that's how `rvrn` works. You should add `processing="last"` attribute to your `<rules>` element, this will make fonttools use a different feature tag (`rclt`) that respects the lookup order and will be applied last.

In that issue, the conclusion presented by @ctrlcctrlv is:

> it seems strange that either all the rules must be processed last or none of the rules must be; unless there can be multiple <rules>? 🤔 But designspaceLib doesn't support that

@justvanrossum responded with:

> What .designspace supports is a high level simplification of what is theoretically possible with FeatureVariations, based on use cases we were aware of at the time. While what you suggest is possible, it would make everything a lot more complex.
> 
> Do you have a specific use case in mind where this could be handy?

I agree with @ctrlcctrlv, and in respond to Just’s question, I do have a use case:

In Recursive (as of [`dc36c9e`](https://github.com/arrowtype/recursive/tree/dc36c9e450a6115e4bd832f62578526902e45e4b)), there is a relatively complex set of `rules` to activate cursive italics & mono or sans-specific glyphs at various axis locations. Many of these glyph substitutions are then used in logic by other OT features, so by now it would very likely be disruptive to the family to now change the rules with `processing="last"`.

_But_, I am now trying to make arbitrary fractions possible, where the fraction bar is split in the default proportional/“Sans” styles (`MONO < 0.5`) but solid in the monospaced/“Mono” styles (`MONO >= 0.5`). 

The intent is for the default to use `fraction.split`, and the mono to use `fraction`.

This`frac` feature relies on detecting the `slash` and subbing it for a `fraction.split` (plus subbing in numerators/denominators for numbers). 

[...]

PR #3360, opened 2023-11-30
[featureVars] allow to assign same subst lookup to several feature tags

anthrotype:
this can be useful to maximise layout engine compatibility or simply to have same feature variations substitutions be accessible via mutliple feature tags for whatever reason.

Fixes https://github.com/fonttools/fonttools/issues/2050
Fixes https://github.com/fonttools/fonttools/issues/3004

/cc @twardoch @arrowtype


anthrotype:
I changed the "com.github.fonttools.varLib.featureVarsFeatureTag" key to take a comma-separated list of feature tags.
The current public functions continue to work like before and accept a single string or an iterable of strings.

anthrotype:
I modified one of the existing tests so it exercises this change. As it's relatively small and backward compatible, I'm going to merge this as is, but feel free to raise any issues.

PR #3363, opened 2023-12-01
[featureVars] don't overwrite FeatureVariations, append new records

anthrotype:
Currently, addFeatureVariationsRaw always deletes existing GSUB.FeatureVariations and overwrites them with the newly built records.

https://github.com/fonttools/fonttools/blob/0b05cec0a345fc9d912170f1fa811215ae54579c/Lib/fontTools/varLib/featureVars.py#L352

It doesn't have to, though. If the features for which we are adding feature variations are not already "variable", we can simply append the new DS-rules-generated records to the existing FeatureVariations list.
We can raise an error if the existing FeatureVariations already reference the same feature tags used by the DS rules, as that could generate ambiguity/undefined logic.

### DIFF
--- a/Lib/fontTools/varLib/featureVars.py
+++ b/Lib/fontTools/varLib/featureVars.py
@@ -69,6 +69,14 @@ def addFeatureVariations(font, conditionalSubstitutions, featureTag="rvrn"):
     )
     if "GSUB" not in font:
         font["GSUB"] = buildGSUB()
+    else:
+        existingTags = _existingVariableFeatures(font["GSUB"].table).intersection(
+            featureTags
+        )
+        if existingTags:
+            raise VarLibError(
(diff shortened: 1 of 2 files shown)


anthrotype:
Also note, feaLib itself is using `addFeatureVariationsRaw` method in a `for` loop, for each feature tag for which the FEA has conditionsets defined:

https://github.com/fonttools/fonttools/blob/0b05cec0a345fc9d912170f1fa811215ae54579c/Lib/fontTools/feaLib/builder.py#L972-L975

that means currently only the last one will prevail, since `addFeatureVariationsRaw` always resets the GSUB.FeatureVariations = None before creating the new records.

With this PR, it will no longer do so.

justvanrossum (approved):

anthrotype:
Thanks Just, I'm writing some tests now

PR #3373, opened 2023-12-06
Fix #3369: sources should not allow anisotropic location

### LINKED ISSUE #3369: [designspaceLib] sources should not allow anisotropic location
This [diagram for the designspace format](https://fonttools.readthedocs.io/en/latest/_images/v5_class_diagram.png) indicates it is possible to allow anisotropic locations for sources.

no_anisotropy_for_sources

- Mutatormath does not support anisotropic locations for **sources**. (it does for instances)
- It would surprise me if Varlib supports it (but love to hear)
- I can't think of any practical applications for this (but love to hear)
- If there are no strong arguments in favor, we could remove it from the requirements?

belluzj:
Fix https://github.com/fonttools/fonttools/issues/3369 in documentation and type hints


anthrotype (approved):

belluzj:
I pushed the black formatting for the lint fail

PR #3380, opened 2023-12-07
[varLib.models] Add getMasterScalars

behdad:
Undocumented and untested, I know.

### DIFF
--- a/Lib/fontTools/varLib/models.py
+++ b/Lib/fontTools/varLib/models.py
@@ -472,6 +472,22 @@ class VariationModel(object):
             for support in self.supports
         ]
 
+    def _getMasterScalarsRecurse(self, out, j, weights, scalar=1):
+        for i, weight in weights.items():
+            influence = -weight * scalar
+            out[i] += influence
+            self._getMasterScalarsRecurse(out, i, self.deltaWeights[i], influence)
+
+    def getMasterScalars(self, targetLocation):
+        out = []
+        for i, (weights, support) in enumerate(zip(self.deltaWeights, self.supports)):
+            scalar = supportScalar(targetLocation, support)
+            out.append(scalar)
+            self._getMasterScalarsRecurse(out, i, weights, scalar)
+
+        out = [out[self.mapping[i]] for i in range(len(out))]
+        return out
+
     @staticmethod
     def interpolateFromDeltasAndScalars(deltas, scalars):
         v = None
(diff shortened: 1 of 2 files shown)


behdad:
@schriftgestalt needed this.

anthrotype:
What is it doing? What's it for?

behdad:
The logic is sound I think. Added test. Documenting all of varLib.models soon.

behdad:
> This may be equivalent to this addition on my JS port of var model:
> https://github.com/googlefonts/fontra/blob/ca8a2cec01a133d6328e730ac9fbdb211f20df10/src/fontra/client/core/var-model.js#L181-L193

Certainly looks like the same goal. My first implementation was exactly like yours, but in testing it gave me wrong results. Let me try again.

behdad:
You are right. Your implementation is simpler. Thanks. I updated.

behdad:
@anthrotype Please take another look. Tested and documented now.

justvanrossum (commented):
LGTM! I'm glad you confirm my approach to be correct.

justvanrossum:
FWIW, here are my JS tests, not sure if too redundant:

https://github.com/googlefonts/fontra/blob/ca8a2cec01a133d6328e730ac9fbdb211f20df10/test-js/test-var-model.js#L346-L437

behdad:
> FWIW, here are my JS tests, not sure if too redundant:
> 
> https://github.com/googlefonts/fontra/blob/ca8a2cec01a133d6328e730ac9fbdb211f20df10/test-js/test-var-model.js#L346-L437

Thanks. Ported.

anthrotype:
I still don't see how/where this could be used, what do you do with those master scalars in practice?
Is it to visualize the relative contribution from each source to the interpolation at a particular location?

If you have the master values and a location and you want to interpolate, why can't one just use the existing interpolateFromMasters method (which in turns computes the deltas and the delta scalars at location and interpolates those)?

anthrotype (approved):

justvanrossum:
I currently use it to visualize the contribution factors, but yeah, these same scalars can be used to just multiply and add the source values, instead of computing delta’s first, and multiplying and adding those. The two models should be completely equivalent.

behdad:
> I wonder if we should add `interpolateFromMastersAndMasterScalars`? It's just a `sum(i*j for i,j in zip(master_values, master_scalars)`.

That would be faster than current `interpolateFromMastersAndScalars` which first calculates deltas. For a GUI use I suppose it would be nice to avoid extra work.

Issue #3383, opened 2023-12-08
`varLib.interpolatable` doesn’t create a required folder path for PDF output, has confusing error


arrowtype:
When running `varLib.interpolatable` (which I love, btw), I tend to hit a couple of snags. One is the installation issue which (I think) is listed at https://github.com/fonttools/fonttools/issues/3382. Another is this:

```
▶ fonttools varLib.interpolatable < fontname>.ttf --pdf <desired_PDF_output_folder>/<fontname>.pdf       
ERROR: error while writing to output stream
Traceback (most recent call last):
  File "/Users/<project_directory>/venv/bin/fonttools", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/<project_directory>/venv/lib/python3.11/site-packages/fontTools/__main__.py", line 31, in main
    runpy.run_module(mod, run_name="__main__")
  File "<frozen runpy>", line 229, in run_module
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/<project_directory>/venv/lib/python3.11/site-packages/fontTools/varLib/interpolatable.py", line 1155, in <module>
    problems = main()
               ^^^^^^
  File "/Users/<project_directory>/venv/lib/python3.11/site-packages/fontTools/varLib/interpolatable.py", line 1083, in main
    with InterpolatablePDF(args.pdf, glyphsets=glyphsets, names=names) as pdf:
  File "/Users/<project_directory>/venv/lib/python3.11/site-packages/fontTools/varLib/interpolatablePlot.py", line 1135, in __enter__
    self.surface = cairo.PDFSurface(self.out, self.width, self.height)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
fontTools.varLib.interpolatablePlot.cairo.IOError: error while writing to output stream
```

The fix is that I just need to create the directory at `< desired_PDF_output_folder >`, and then the PDF generation will work.

Would it make sense to just create that folder if it doesn’t exist?

If not, would it be sensible to check for that folder, and make a clearer message that it doesn’t exist?

behdad:
Most tools don't do this. But since Cosimo already fixed it i think we're good.

@arrowtype how do you like the new tests? I tried to sort the report by decreasing importance to help filter out the noise.

arrowtype:
Oh wow, that was super quick! Thanks so much to both of you.

[...]

PR #3384, opened 2023-12-08
[interpolatable] ensure parent directory for various output paths

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/3383


arrowtype:
Awesome, thanks Cosimo!

PR #3386, opened 2023-12-11
[varLib.models] Add interpolateFromValuesAndScalars()

behdad:
Faster than the alternative way to calculate them.

### DIFF
--- a/Lib/fontTools/varLib/models.py
+++ b/Lib/fontTools/varLib/models.py
@@ -526,8 +526,8 @@ class VariationModel(object):
 
     def interpolateFromMasters(self, loc, masterValues, *, round=noRound):
         """Interpolate from master-values, at location loc."""
-        deltas = self.getDeltas(masterValues, round=round)
-        return self.interpolateFromDeltas(loc, deltas)
+        scalars = self.getMasterScalars(loc)
+        return self.interpolateFromMastersAndMasterScalars(masterValues, scalars)
 
     def interpolateFromMastersAndScalars(self, masterValues, scalars, *, round=noRound):
         """Interpolate from master-values, and scalars fetched from
--- a/Lib/fontTools/varLib/models.py
+++ b/Lib/fontTools/varLib/models.py
@@ -486,7 +486,7 @@ class VariationModel(object):
         """Return multipliers for each master, for the given location.
         If interpolating many master-values at the same location,
         this function allows speed up by fetching the scalars once
-        and using them with interpolateFromMastersAndMasterScalars().
+        and using them with interpolateFromValuesAndScalars().
 
         Note that the scalars used in interpolateFromMastersAndScalars(),
         are *not* the same as the ones returned here. They are the result
@@ -501,7 +501,15 @@ class VariationModel(object):
 
     @staticmethod
     def interpolateFromValuesAndScalars(values, scalars):
-        """Interpolate from values and scalars coefficients."""
+        """Interpolate from values and scalars coefficients.
+
+        If the values are master-values, then the scalars should be
+        fetched from getMasterScalars().
+
+        If the values are deltas, then the scalars should be fetched
+        from getScalars(); in which case this is the same as
+        interpolateFromDeltasAndScalars().
+        """
         v = None
         assert len(values) == len(scalars)
         for value, scalar in zip(values, scalars):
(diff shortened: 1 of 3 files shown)


behdad:
In theory I can use this in `varLib.merger`. Let me know if you want me to do that.

F944:
Good

justvanrossum:
Should `interpolateFromMasters()` then be implemented in terms of `interpolateFromMastersAndMasterScalars()`? That should be more efficient.

Lib/fontTools/varLib/models.py
```diff
@@ -530,6 +531,22 @@ def interpolateFromMastersAndScalars(self, masterValues, scalars, *, round=noRou
         deltas = self.getDeltas(masterValues, round=round)
         return self.interpolateFromDeltasAndScalars(deltas, scalars)
 
+    @staticmethod
+    def interpolateFromMastersAndMasterScalars(masterValues, masterScalars):
```
justvanrossum, line 535: This function is literally the same as `interpolateFromDeltasAndScalars()`, except for the variable names. Do we really need both?

anthrotype:
I dont like the method name, it contains twice a name "master" which in theory we would like to avoid.
And if it is indeed faster (has anybody actually measured?), why do we need two methods to interpolate from sources and precomputed scalars? Ok backward compat sure, but if one of the two is definitely more efficient, then we should deprecate one and suggest to use the other.

behdad:
All good points. Maybe I can just add a boolean optional argument to the existing API?

behdad:
> I dont like the method name, it contains twice a name "master" which in theory we would like to avoid. And if it is indeed faster (has anybody actually measured?), why do we need two methods to interpolate from sources and precomputed scalars? Ok backward compat sure, but if one of the two is definitely more efficient, then we should deprecate one and suggest to use the other.

I think `interpolateFromMastersAndScalars` can be deprecated. I don't have a situation in mind where one has delta scalars but want to interpolate masters. I was going to merge the two with a bool parameter, but realized that one is a staticmethod and one isn't.

Lib/fontTools/varLib/models.py
```diff
@@ -499,29 +500,34 @@ def getMasterScalars(self, targetLocation):
...
                 continue
-            contribution = delta * scalar
+            contribution = value * scalar
             if v is None:
                 v = contribution
             else:
                 v += contribution
         return v
 
+    @staticmethod
```
anthrotype, line 517: maybe make it a @classmethod that takes `cls` as first param. it wouldn't change the way one one interacts with it vs staticmethod

Lib/fontTools/varLib/models.py
```diff
@@ -530,6 +536,14 @@ def interpolateFromMastersAndScalars(self, masterValues, scalars, *, round=noRou
         deltas = self.getDeltas(masterValues, round=round)
         return self.interpolateFromDeltasAndScalars(deltas, scalars)
 
+    @staticmethod
+    def interpolateFromMastersAndMasterScalars(masterValues, masterScalars):
+        """Interpolate from master-values and master-scalars fetched
+        from getMasterScalars()."""
+        return VariationModel.interpolateFromValuesAndScalars(
```
anthrotype, line 543: why can't simply add this `interpolateFromValuesAndScalars` staticmethod? What's the use of `interpolateFromMastersAndMasterScalars` -- besides the questionable name itself?
I think "value" is generic enough to mean "source" or "master" or even "delta", it's just the responsibility of the caller to pass in type of scalars that matches the kind of "values".

Lib/fontTools/varLib/models.py
```diff
@@ -499,29 +500,34 @@ def getMasterScalars(self, targetLocation):
...
-            contribution = delta * scalar
+            contribution = value * scalar
             if v is None:
                 v = contribution
             else:
                 v += contribution
         return v
 
+    @staticmethod
+    def interpolateFromDeltasAndScalars(deltas, scalars):
```
anthrotype, line 518: yes, this one we can keep as an alias of `interpolateFromValuesAndScalars` for backward compat, but I don't think we also need `interpolateFromMastersAndMasterScalars`

Lib/fontTools/varLib/models.py
anthrotype: or actually just get rid of it altogether (see my other comment)

Lib/fontTools/varLib/models.py
behdad: Sounds good to me. I'll adjust.

behdad:
PTAL.

I didn't change staticmethod to classmethod. I don't understand the need for that.

anthrotype:
> I don't understand the need for that.

I think the only difference/advantage is that the classmethod would work with subclasses, but it's ok if you prefer not to use

anthrotype (approved):

F944:
H

PR #3392, opened 2023-12-14
[interpolatablePlot] Draw summary / index / table of contents

behdad:


behdad:
Sample output:
[out.pdf](https://github.com/fonttools/fonttools/files/13668020/out.pdf)

MariannaPaszkowska:
HI Behdad, when I try to select/copy/paste text from the PDF (legend or anywhere else) I'm getting mirrored output: "ruotnoc eht ni knik gnisuac tnioP", ":dnegeL"

MariannaPaszkowska:
I think that a list of all glyphs with issues would be helpful to print out. Then, it will be easy to copy those and put them in a text editor to test them all together). I suggest this format: "asterisk 002A *".

MariannaPaszkowska:
Also, the legend is very helpful, but it would be even better to print in each issue, somewhere next to the "midway interpolation", the list of all issues in this glyph. Example: "midway interpolation", then "Kink, Contur start point issue, Wrong contour order", etc.

behdad:
> I think that a list of all glyphs with issues would be helpful to print out. Then, it will be easy to copy those and put them in a text editor to test them all together). I suggest this format: "asterisk 002A *".

That's what I put in the Summary page. Or do you mean print out to the terminal?

behdad:
> Also, the legend is very helpful, but it would be even better to print in each issue, somewhere next to the "midway interpolation", the list of all issues in this glyph. Example: "midway interpolation", then "Kink, Contur start point issue, Wrong contour order", etc.

But that's written on top of each page.

behdad:
> HI Behdad, when I try to select/copy/paste text from the PDF (legend or anywhere else) I'm getting mirrored output: "ruotnoc eht ni knik gnisuac tnioP", ":dnegeL"

Yeah this seems to be an issue with the Apple PDF Preview app. Try opening the PDF in Chrome for example and observe that it copies correctly. You can also `brew install evince` and use `evince out.pdf`.

PR #3406, opened 2024-01-07
[otlLib] Pass ttFont to name.addName in buildStatTable

khaledhosny:
Avoids ltag warning similar to: ae31d056051a28c39bc60fa2b827bc14efcef287


anthrotype:
thanks

PR #3408, opened 2024-01-08
Merge tool improvements

miloush:
- proper usage help
- print key names during merging when `--verbose` (so that failing fields of tables can be identified)

New options:
- `--input-file` to read file names from file instead of command line
- `--output-file` moved from `main` to `Options`
- `--import-file` to import TTX after merge


behdad:
Thanks! LGTM.

Probably better to port to argparse...

PR #3413, opened 2024-01-10
[featureVars] Re-use FeatureVariationRecord's when possible

khaledhosny:
If a FeatureVariationRecord with the same ConditionTable exists re-use it and append FeatureTableSubstitutionRecord’s.

Without this, in the following feature code only the first lookup will be applied since there will be two FeatureVariationRecord with the same ConditionTable, so the first will be matched and the other will be skipped:

    conditionset test {
        wght 600 1000;
        wdth 150 200;
    } test;

    variation ccmp test {
        sub e by a;
    } ccmp;

    variation rlig test {
        sub b by c;
    } rlig;

With this change only one FeatureVariationRecord will be created with two FeatureTableSubstitutionRecord’s.

### DIFF
--- a/Lib/fontTools/varLib/featureVars.py
+++ b/Lib/fontTools/varLib/featureVars.py
@@ -635,15 +635,7 @@ def findFeatureVariationRecord(featureVariations, conditionTable):
         )
 
     for fvr in featureVariations.FeatureVariationRecord:
-        if fvr.ConditionSet.ConditionCount != len(conditionTable):
-            continue
-        if all(
-            ct1.Format == ct2.Format
-            and ct1.AxisIndex == ct2.AxisIndex
-            and ct1.FilterRangeMinValue == ct2.FilterRangeMinValue
-            and ct1.FilterRangeMaxValue == ct2.FilterRangeMaxValue
-            for ct1, ct2 in zip(conditionTable, fvr.ConditionSet.ConditionTable)
-        ):
+        if conditionTable == fvr.ConditionSet.ConditionTable:
             return fvr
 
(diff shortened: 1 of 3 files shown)


Lib/fontTools/varLib/featureVars.py
```diff
@@ -614,6 +626,29 @@ def buildConditionTable(axisIndex, filterRangeMinValue, filterRangeMaxValue):
...
+    for fvr in featureVariations.FeatureVariationRecord:
+        if fvr.ConditionSet.ConditionCount != len(conditionTable):
+            continue
+        for ct1, ct2 in zip(conditionTable, fvr.ConditionSet.ConditionTable):
+            if (
+                ct1.Format == ct2.Format
+                and ct1.AxisIndex == ct2.AxisIndex
+                and ct1.FilterRangeMinValue == ct2.FilterRangeMinValue
+                and ct1.FilterRangeMaxValue == ct2.FilterRangeMaxValue
+            ):
```
anthrotype: I believe (though haven't tested) that you could simply compare ct1 == ct2 because BaseTable implements `__eq__` that compares attributes by value, maybe try see if that works

anthrotype: yes, I think it should work, see https://github.com/fonttools/fonttools/blob/e3cde46e69074f75462f3ca1bfeec736f2b559ee/Lib/fontTools/ttLib/tables/otBase.py#L1166-L1173

anthrotype (approved):
LGTM! (with minor simplification suggested in comments above)

PR #3415, opened 2024-01-11
[CI] adapt to new upload/download-artifact changes 

anthrotype:
https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes

> Due to how Artifacts are created in this new version, it is no longer possible to upload to the same named Artifact multiple times. You must either split the uploads into multiple Artifacts with different names, or only upload once. Otherwise you will encounter an error.

This caused the PyPI upload job to fail earlier. Hopefully this should fix it.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -110,7 +110,7 @@ jobs:
     # only run if the commit is tagged...
     if: startsWith(github.ref, 'refs/tags/')
     steps:
-    - uses: actions/download-artifact@v3
+    - uses: actions/download-artifact@v4
       with:
         name: artifact
         path: dist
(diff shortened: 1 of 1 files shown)


anthrotype:
(we'll only know if this actually works when we will try to release again)

khaledhosny (approved):
LGTM, but indeed we will find out if it works by the time we cut a new release.

anthrotype:
> indeed we will find out if it works by the time we cut a new release.

well, deploying 4.48.0 actually failed :)

https://github.com/fonttools/fonttools/actions/runs/7799843107

I hope that this will fix it:
https://github.com/fonttools/fonttools/commit/8e7a4ebb45aedb99b3066d0bc196ccc1e1d41950

Issue #3418, opened 2024-01-15
lxml 5.1


frenzymadness:
We are testing version 5.1.0 of lxml and it seems that fonttools builds fine and all tests pass with it so it might be a good idea to remove the version limit from the following files:

https://github.com/fonttools/fonttools/blob/bf37f5721106afddeec197af5343027e4f487d55/requirements.txt#L19
and
https://github.com/fonttools/fonttools/blob/bf37f5721106afddeec197af5343027e4f487d55/setup.py#L100

anthrotype:
good idea, thanks. Would you like to send a PR?

PR #3420, opened 2024-01-15
don't restrict lxml<5, and run tests with latest 5.1.0

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/3418


frenzymadness (approved):
Thanks for the quick PR. Looks good to me!

Issue #3421, opened 2024-01-18
HashPointPen fails with ufoLib2


jenskutilek:
I'm using the HashPointPen to check glyphs from a UFO (opened with ufoLib2) against glyphs from a TTF.

Both produce different hashes for identical glyphs when the glyph is a composite.

```
TTF: w2671[...|(+1+0+0+1+0+0)][l1008+3162l459+3736l942+3736l1335+3162|(+1+0+0+1+180+0)]
UFO: w2671[...|(+1.0+0.0+0.0+1.0+0+0)][l1008+3162l459+3736l942+3736l1335+3162|(+1.0+0.0+0.0+1.0+180+0)]
```

As you can see, the difference is that the transformation (scale) is stored as an int when the pen is fed by the TTF, but as a float when it is fed by the UFO.

https://github.com/fonttools/fonttools/blob/679a5c0c252fd5b983de0266b0a250fc7cbb49fc/Lib/fontTools/pens/hashPointPen.py#L68-L69

I guess the HashPointPen should be changed to even out this difference, but what would be preferred? The scales in the transformation are stored as F2Dot14 values in a TTF, so for maximum precision, we should probably always write out the full F2Dot14?

jenskutilek:
Using the format `{t:+g}` instead of `{t:+}` would even out the difference for integer scales.

But the F2Dot14 precision comes into play once the components are scaled by a fractional number.

Example: Scaled in the font editor by 91.3 %, represented in the two formats:

```
UFO: (+0.913+0+0+0.913+516+272)
TTF: (+0.913025+0+0+0.913025+516+272)
```

justvanrossum:
> Example: Scaled in the font editor by 91.3 %, represented in the two formats:

But that's literally a difference between the two components... TTF necessarily has to store an approximation, so that is lossy, and therefore I would expect a different hash.

jenskutilek:
Formatting the values with `{floatToFixed(t, 14):+g}` is not pretty, but solves the problem:

Scale 1.0, 1.0:

```
UFO: (+16384+0+0+16384-1.16818e+07+0)
TTF: (+16384+0+0+16384-1.16818e+07+0)
```

Scale 0.913, 0.913:

```
UFO: (+14959+0+0+14959+8.45414e+06+4.45645e+06)
TTF: (+14959+0+0+14959+8.45414e+06+4.45645e+06)
```

justvanrossum:
[...]

PR #3422, opened 2024-01-19
Support rounding the scale values of components in HashPointPen

jenskutilek:
See #3421

### DIFF
--- a/Lib/fontTools/pens/hashPointPen.py
+++ b/Lib/fontTools/pens/hashPointPen.py
@@ -1,10 +1,28 @@
-# Modified from https://github.com/adobe-type-tools/psautohint/blob/08b346865710ed3c172f1eb581d6ef243b203f99/python/psautohint/ufoFont.py#L800-L838
+from __future__ import annotations
 import hashlib
 
-from fontTools.misc.fixedTools import floatToFixedToFloat
+from fontTools.misc.fixedTools import floatToFixedToStr
 from fontTools.pens.basePen import MissingComponentError
 from fontTools.pens.pointPen import AbstractPointPen
-from functools import partial
+from typing import Callable
+
+
+def ttScaleRound(value: float) -> float:
+    """
+    Rounds to values that can be represented by F2Dot14, in case of the HashPointPen
+    the scale values of component transformations. This should be used when comparing a
+    UFO glyph against a TrueType glyph from a glyf table.
+
+    Args:
+        value (float): The value to round.
+
+    Returns:
+        float: The rounded value.
+    """
+    return float(floatToFixedToStr(value, 14))
+
+
+# Modified from https://github.com/adobe-type-tools/psautohint/blob/08b346865710ed3c172f1eb581d6ef243b203f99/python/psautohint/ufoFont.py#L800-L838
 
 
 class HashPointPen(AbstractPointPen):
@@ -36,10 +54,7 @@ class HashPointPen(AbstractPointPen):
     """
 
     def __init__(
-        self,
-        glyphWidth=0,
-        glyphSet=None,
-        roundFunction=partial(floatToFixedToFloat, precisionBits=14),
+        self, glyphWidth=0, glyphSet=None, roundFunction: Callable | None = None
     ):
         self.glyphset = glyphSet
         self.roundFunction = roundFunction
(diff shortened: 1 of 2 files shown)


Lib/fontTools/pens/hashPointPen.py
```diff
@@ -33,8 +35,14 @@ class HashPointPen(AbstractPointPen):
     >    pass
     """
 
-    def __init__(self, glyphWidth=0, glyphSet=None):
+    def __init__(
+        self,
+        glyphWidth=0,
+        glyphSet=None,
+        roundFunction=partial(floatToFixedToFloat, precisionBits=14),
```
justvanrossum, line 42: TBH I don't think the floatToFixedToFloat behavior should be the default. Since this is specific to your use case, I think you have to enable it explicitly.

jenskutilek: I thought that it was useful for all cases, but of course you are right.

jenskutilek:
I made the rounding function optional and added some more tests. I'll check again tomorrow if it now does what I need, as I don't have my test data here right now.

Lib/fontTools/pens/hashPointPen.py
```diff
@@ -33,8 +53,11 @@ class HashPointPen(AbstractPointPen):
     >    pass
     """
 
-    def __init__(self, glyphWidth=0, glyphSet=None):
+    def __init__(
+        self, glyphWidth=0, glyphSet=None, roundFunction: Callable | None = None
```
anthrotype, line 57: instead of optional callable, you could set the default to `fontTools.misc.roundTools.noRound`

jenskutilek: I figured it would be better to check for None once, instead of rounding 4 values with a no-op function :)

anthrotype: have you tested it? I think there will likely be a negligible difference between the approaches, since modern Python interpreters are highly optimized. We use the `noRound` approach in other parts of the code like varLib.varStore which is arguably more perf critical than this pen. In general I think it's more elegant to use a no-op function.

Lib/fontTools/pens/hashPointPen.py
```diff
@@ -1,8 +1,28 @@
...
+    the scale values of component transformations. This should be used when comparing a
+    UFO glyph against a TrueType glyph from a glyf table.
+
+    Args:
+        value (float): The value to round.
+
+    Returns:
+        float: The rounded value.
+    """
+    return float(floatToFixedToStr(value, 14))
```
anthrotype, line 22: floatToFixedToFloat is what you want

Lib/fontTools/pens/hashPointPen.py
```diff
@@ -1,8 +1,28 @@
...
+from __future__ import annotations
 import hashlib
 
+from fontTools.misc.fixedTools import floatToFixedToStr
 from fontTools.pens.basePen import MissingComponentError
 from fontTools.pens.pointPen import AbstractPointPen
+from typing import Callable
+
+
+def ttScaleRound(value: float) -> float:
```
anthrotype, line 10: the name of this function is unclear, `partial(floatToFixedToFloat(precisionBits=14))` does what it says on the tin

Lib/fontTools/pens/hashPointPen.py
```diff
@@ -33,8 +53,11 @@ class HashPointPen(AbstractPointPen):
     >    pass
     """
 
-    def __init__(self, glyphWidth=0, glyphSet=None):
+    def __init__(
+        self, glyphWidth=0, glyphSet=None, roundFunction: Callable | None = None
+    ):
         self.glyphset = glyphSet
+        self.roundFunction = roundFunction
```
anthrotype, line 60: I find a bit strange that immediately below this there is a call to built-in `round` to round the advance width, a user might immediately wonder why not use the roundFunction that I just provided?

anthrotype: if this function is only meant to round the components' 2x2 transform, then it should be renamed accordingly, because as it stands it suggests that it's going to be used to round in general which is not the case (advance width and glyph points or translation dx,dy are not being rounded).

I think overall that this HashPointPen could stay as it is and not round anything, and instead you create a separate FilterPen subclass that only does what you want, that is round the component's transform as desired.

anthrotype: we have already a RoundingPen and RoundingPointPen here:
https://github.com/fonttools/fonttools/blob/306f40a2aea237ca98d2d207847164c6784bc79f/Lib/fontTools/pens/roundingPen.py#L9

these only apply the rounding to point coords and dx/dy offsets, but maybe you could add an additional parameter to round the component transform floats.
Then you could use this in conjunction to the HashPointPen to achieve the desired effect

jenskutilek: I just found another complication: the HashPointPen does not conform [to the UFO spec](https://github.com/unified-font-object/ufo-spec/issues/224) at all anymore ...

jenskutilek: Alright, I forgot about RoundingPointPen.

I think calculating a hash to compare source data to TTF data is a common use case, so maybe there should be a readymade solution for it? (I thought that ufo2ft didn't compare the UFO hash against TTF glyphs, but actually it does) So maybe I could add another pen that combines HashPointPen and RoundingPointPen to make comparing with TTF easy?

anthrotype: first of all we could add the optional `transformRoundFunction` parameter to the rounding pens, that defaults to `noRound`. Then you could already start using them in combination with the HashPointPen in ufo2ft or wherever you need them. It's just one additional line to construct the combined pen, I don't think we need to add anything here. You could add to the docstring of HashPointPen an example.

jenskutilek: Ok, I'll start a new PR for that.

PR #3424, opened 2024-01-22
feaLib: Drop glyph and class names length limit

khaledhosny:
These were implemented to follow FEA spec, but makeotf does not seem to have a length limit on class names, and the glyph names limit (if it still has it, I didn’t check) seems like an implementation detail we don’t have to enforce.

Fixes https://github.com/googlefonts/ufo2ft/issues/588
See also https://github.com/googlefonts/ufo2ft/pull/811#discussion_r1461667058

### DIFF
--- a/Lib/fontTools/feaLib/lexer.py
+++ b/Lib/fontTools/feaLib/lexer.py
@@ -111,10 +111,6 @@ class Lexer(object):
             glyphclass = text[start + 1 : self.pos_]
             if len(glyphclass) < 1:
                 raise FeatureLibError("Expected glyph class name", location)
-            if len(glyphclass) > 63:
-                raise FeatureLibError(
-                    "Glyph class names must not be longer than 63 characters", location
-                )
             if not Lexer.RE_GLYPHCLASS.match(glyphclass):
                 raise FeatureLibError(
                     "Glyph class names must consist of letters, digits, "
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -2071,13 +2071,7 @@ class Parser(object):
     def expect_glyph_(self):
         self.advance_lexer_()
         if self.cur_token_type_ is Lexer.NAME:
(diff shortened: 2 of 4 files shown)


anthrotype:
> makeotf does not seem to have a length limit on class names

how did check that?

khaledhosny:
> > makeotf does not seem to have a length limit on class names
> 
> how did you check that?

I tested with this feature file:
```fea
@aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa = [a];

feature liga {
 sub @aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa by b;
} liga;
```

anthrotype:
> the glyph names limit (if it still has it, I didn’t check)

it would be nice to confirm if makeotf still actually enforces that or not

anthrotype (approved):

khaledhosny:
> > the glyph names limit (if it still has it, I didn’t check)
> 
> it would be nice to confirm if makeotf still actually enforces that or not

I just checked, and the same long class name is also accepted as glyph name (using both UFO and OTF files as input).

PR #3426, opened 2024-01-23
Add transformRoundFunc parameter to RoundingPens

jenskutilek:
From discussion in #3422:

Add option to supply a function to RoundingPen and RoundingPointPen which is used to round the scale components of transforms.

The scenario to use this to make sure scale values conform to F2Dot14, as used in TrueType composite glyphs, is described in the docstring of the related HashPointPen.

### DIFF
--- a/Lib/fontTools/pens/hashPointPen.py
+++ b/Lib/fontTools/pens/hashPointPen.py
@@ -31,6 +31,20 @@ class HashPointPen(AbstractPointPen):
     >    # The hash values are identical, the outline has not changed.
     >    # Compile the hinting code ...
     >    pass
+
+    If you want to compare a glyph from a source format which supports floating point
+    coordinates and transformations against a glyph from a format which has restrictions
+    on the precision of floats, e.g. UFO vs. TTF, you must use an appropriate rounding
+    function to make the values comparable. For TTF fonts with composites, this
+    construct can be used to make scale values conform to F2Dot14:
+
+    > ttf_hash_pen = HashPointPen(ttf_glyph_width, ttFont.getGlyphSet())
+    > ttf_round_pen = RoundingPointPen(ttf_hash_pen, transformRoundFunc=partial(floatToFixedToFloat, precisionBits=14))
+    > ufo_hash_pen = HashPointPen(ufo_glyph.width, ufo)
+    > ttf_glyph.drawPoints(ttf_round_pen, ttFont["glyf"])
+    > ufo_round_pen = RoundingPointPen(ufo_hash_pen, transformRoundFunc=partial(floatToFixedToFloat, precisionBits=14))
+    > ufo_glyph.drawPoints(ufo_round_pen)
+    > assert ttf_hash_pen.hash == ufo_hash_pen.hash
     """
 
     def __init__(self, glyphWidth=0, glyphSet=None):
--- a/Lib/fontTools/pens/hashPointPen.py
+++ b/Lib/fontTools/pens/hashPointPen.py
@@ -36,7 +36,7 @@ class HashPointPen(AbstractPointPen):
     coordinates and transformations against a glyph from a format which has restrictions
     on the precision of floats, e.g. UFO vs. TTF, you must use an appropriate rounding
     function to make the values comparable. For TTF fonts with composites, this
-    construct can be used to make scale values conform to F2Dot14:
+    construct can be used to make the transform values conform to F2Dot14:
 
     > ttf_hash_pen = HashPointPen(ttf_glyph_width, ttFont.getGlyphSet())
     > ttf_round_pen = RoundingPointPen(ttf_hash_pen, transformRoundFunc=partial(floatToFixedToFloat, precisionBits=14))
--- a/Lib/fontTools/pens/roundingPen.py
+++ b/Lib/fontTools/pens/roundingPen.py
@@ -9,8 +9,8 @@ __all__ = ["RoundingPen", "RoundingPointPen"]
 class RoundingPen(FilterPen):
     """
     Filter pen that rounds point coordinates and component XY offsets to integer. For
-    rounding the component scale values, a separate round function can be passed to the
-    pen.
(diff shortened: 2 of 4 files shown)


Lib/fontTools/pens/roundingPen.py
```diff
@@ -8,7 +8,9 @@
 
 class RoundingPen(FilterPen):
     """
-    Filter pen that rounds point coordinates and component XY offsets to integer.
+    Filter pen that rounds point coordinates and component XY offsets to integer. For
+    rounding the component scale values, a separate round function can be passed to the
```
anthrotype, line 12: technically it's not just "scale" values (but also shear or rotation anything that 2x2 matrix can do), i'd just call it "component transform" values

anthrotype (approved):

jenskutilek:
Hm, when the recording pen gets the Transformation from the rounding pen, it is stored as is.

```
        roundpen.addComponent("a", (0.913, 0, 0, -1, 10.5, -10.5))
        # The 0.913 should have been rounded to F2Dot14 precision
>       assert recpen.value == [
            ("addComponent", ("a", (0.913025, 0, 0, -1, 11, -10)), {}),
        ]
E       AssertionError: assert [('addCompone...1 -10]>), {})] == [('addCompone...1, -10)), {})]
E         At index 0 diff: ('addComponent', ('a', <Transform [0.913025 0 0 -1 11 -10]>), {}) != ('addComponent', ('a', (0.913025, 0, 0, -1, 11, -10)), {})
```

It's probably preferable that the recording pen always stores the transformation as a tuple? @anthrotype could you take another look?

anthrotype:
wait.. we don't want to do convert the transform to a plain tuple in the recording pen.
The Transform class is a NamedTuple (which is still a tuple if I am not mistaken) and it should compare equal to a regular tuple. 

```
>>> from fontTools.misc.transform import Transform
>>> Transform() == (1, 0, 0, 1, 0, 0)
True
```

jenskutilek:
I reverted the commit. Any idea why the comparison fails? Just int vs. float problems again?

anthrotype:
> Just int vs. float problems again?

i don't think so, will take a look

jenskutilek:
```
E       AssertionError: assert [('addComponent', ('a', <Transform [0.913025 0 0 -1 11 -10]>), {})] == [('addComponent', ('a', (0.913025, 0, 0, -1, 11, -10)), {})]
E         At index 0 diff: ('addComponent', ('a', <Transform [0.913025 0 0 -1 11 -10]>), {}) != ('addComponent', ('a', (0.913025, 0, 0, -1, 11, -10)), {})
E         Full diff:
E         - [('addComponent', ('a', (0.913025, 0, 0, -1, 11, -10)), {})]
E         ?                         ^        -  -  -   -   -    ^
E         + [('addComponent', ('a', <Transform [0.913025 0 0 -1 11 -10]>), {})]
E         ?                         ^^^^^^^^^^^^                      ^^
```

Tests/pens/roundingPen_test.py
```diff
@@ -0,0 +1,67 @@
+from fontTools.misc.fixedTools import floatToFixedToFloat
+from fontTools.pens.recordingPen import RecordingPen, RecordingPointPen
+from fontTools.pens.roundingPen import RoundingPen, RoundingPointPen
+from functools import partial
+
+
+tt_scale_round = partial(floatToFixedToFloat, precisionBits=14)
```
anthrotype, line 7: the test rightly fails. floatToFixedToFloat simply does `otRound(value * scale) / scale` (with `scale = 1 << precisionBits`), so it comes out with a lot of decimal positions. It's meant to give you a float that is quantized to the fractional portions of the F2Dot14, to do computations with it. Instead you want to use `floatToFixedToStr` here (and convert the str back to float()) so you obtain a normalized, short representation suitable for pretty strings, like the hashpointpen

anthrotype: > Does pytest compare string representations, or is that just done for the diff output?

no, pytest compares the actual values just using the built-in `assert`; the computed and expected values _are_ different!

anthrotype: you were misled into thinking that the Transform object actually contain `0.913025` because its `repr()` prints that string rounded to 6 decimal positions, but it _actually_ contains a longer float

behdad: > Instead you want to use `floatToFixedToStr` here (and convert the str back to float()) so you obtain a normalized, short representation suitable for pretty strings, like the hashpointpen

That's why I originally suggested that ;).

anthrotype: my fault, thanks behdad

jenskutilek:
Thanks. Hope I got it now :)

Lib/fontTools/pens/hashPointPen.py
```diff
@@ -31,6 +31,20 @@ class HashPointPen(AbstractPointPen):
...
     >    pass
+
+    If you want to compare a glyph from a source format which supports floating point
+    coordinates and transformations against a glyph from a format which has restrictions
+    on the precision of floats, e.g. UFO vs. TTF, you must use an appropriate rounding
+    function to make the values comparable. For TTF fonts with composites, this
+    construct can be used to make the transform values conform to F2Dot14:
+
+    > ttf_hash_pen = HashPointPen(ttf_glyph_width, ttFont.getGlyphSet())
+    > ttf_round_pen = RoundingPointPen(ttf_hash_pen, transformRoundFunc=partial(floatToFixedToFloat, precisionBits=14))
```
anthrotype, line 42: I think using `floatToFixedToFloat` should still work when you hit the assertion below (it should pass, does it not?), but maybe you want to amend the example here in the docstring to use the approach you went with in the tests i.e. float(floatToFixedToStr)?

anthrotype (approved):

jenskutilek:
I switched to `floatToFixedToFloat` again, because it can be used as a partial, which is more elegant ;)

PR #3427, opened 2024-01-23
HashPointPen: Output decimals as specced in UFO

jenskutilek:
The implementation of HashPointPen has deviated from the UFO spec. This PR rounds the decimals as specified.

https://github.com/unified-font-object/ufo-spec/issues/224


anthrotype:
the rounding should depend on the usage of this hash. You want different roundings (or none at all) depending on whether you are computing this as the id for PostScript hints vs TrueType instructions. Forcing the use of the same HashPointPen for both doesn't make sense to me.
Also, detailing exactly in the UFO spec which sort of rounding to apply to what and when is too (p)restrictive, better to leave it out altogether and leave it to the tools to coordinate with each other.

Issue #2937, opened 2022-12-22
feaLib: incorrect ordering of alternates in aalt lookups


cmyr:
I've run into something that I believe is a bug in the tables generated for the `aalt` feature.

According to the spec, in the case there are multiple alternates for a given glyph,

> These alternate glyphs will be sorted in the order that the source features are named in the aalt definition, not the order of the feature definitions in the file. Alternates defined explicitly, as in step 2 above, will precede all others.

However looking at the output of feaLib, it does not respect this ordering. I think [it *attempts* to](https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/feaLib/builder.py#L285) (at least, it is going out of its way to add them in the order they are referenced?), but using a set to store the alternates makes this impossible, and I suspect from the output that they are getting sorted again later on.

Issue #836, opened 2017-02-11
[feaLib] Put glyph alternates in the same order as makeotf


brawer:
feaLib currently sorts GSUB glyph alternates by glyph order. We should figure out what exactly makeotf does, and emulate that behavior.

khaledhosny:
It should be just the same order as in the feature file, isn’t it?

anthrotype:
I think Sascha meant for the auto generated aalt feature.
I think glyph order is good enough.

behdad:
Duplicate of https://github.com/fonttools/fonttools/issues/2937

PR #3430, opened 2024-01-25
[feaLib] fix ordering of alternates in aalt lookups

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/2937 and https://github.com/fonttools/fonttools/issues/836

alternate glyphs are now sorted in the order that the source features are named in the aalt definition, as the FEA spec says, and not by glyph order.


behdad:
Lgtm

PR #3432, opened 2024-01-25
fix bug in instancer iup optimization

qxliu76:
Fix tolerance value. Previously boolean value is passed to tolerance which is of float type.


behdad:
Ouch!

PR #3433, opened 2024-01-26
Support sbix Glyphs with graphicType "flip"

PoomSmart:
Since iOS 17.4 beta, Apple introduced the "flip" graphic type for the sbix glyphs to reference the another glyph to be flipped by their render engine. Their use case is for directional emojis.

Example emoji: https://emojipedia.org/woman_walking_facing_right#technical

This is the example of the output XML after this PR:

```
<glyph graphicType="flip" name="u1F3C3.0.M.u27A1" originOffsetX="0" originOffsetY="0">
    <ref glyphname="u1F3C3.0.M" />
</glyph>
```

Before, it was just <hexdata> with some data in it.

### DIFF
--- a/Lib/fontTools/ttLib/tables/sbixGlyph.py
+++ b/Lib/fontTools/ttLib/tables/sbixGlyph.py
@@ -71,7 +71,7 @@ class Glyph(object):
                 sbixGlyphHeaderFormat, self.rawdata[:sbixGlyphHeaderFormatSize], self
             )
 
-            if self.graphicType == "dupe":
+            if self.graphicType == "dupe" or self.graphicType == "flip":
                 # this glyph is a reference to another glyph's image data
                 (gid,) = struct.unpack(">H", self.rawdata[sbixGlyphHeaderFormatSize:])
                 self.referenceGlyphName = ttFont.getGlyphName(gid)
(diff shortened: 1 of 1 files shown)


Lib/fontTools/ttLib/tables/sbixGlyph.py
```diff
@@ -71,7 +71,7 @@ def decompile(self, ttFont):
                 sbixGlyphHeaderFormat, self.rawdata[:sbixGlyphHeaderFormatSize], self
             )
 
-            if self.graphicType == "dupe":
+            if self.graphicType == "dupe" or self.graphicType == "flip":
```
anthrotype, line 74: maybe you could define a method `is_reference_type` that simply `return self.graphicType == "dupe" or self.graphicType == "flip"` and use everywhere

PoomSmart: @anthrotype Done!

anthrotype:
I haven't seen official documentation of this yet. E.g. https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6sbix.html does not mention this.
We confirmed this by experimentation.

radarhere:
Ok, thanks very much.

Issue #2934, opened 2022-12-21
[feaLib] getCompiledSize_ produces misleading results for contextual lookups


cmyr:
When compiling contextual lookups, otlLib attempts to construct the various formats (glyph-based, class-based, and coverage-based) and then compiles them to determine their size; this size is then used to pick the appropriate format.

This function takes a list of subtables, which it measures independently and then sums the resulting sizes. This means that any objects that could be shared between items in the list are not accounted for in the size computation.

This is a problem because format 3 generally produces more subtables than the other formats; it requires a subtable per rule, but the other formats can store multiple rules. In the final binary, these will all be deduped together, but this doesn't happen during measurement. This means that we may choose not to compile to format 3 even when it is the best representation.

I suspect this is happening frequently; in my own measurements, format 3 is often better than format 2, but the way this measurement is implemented it is very unlikely to get selected.

I believe this is a bug.

Fixing this should not be too difficult; it would only require stuffing all the subtables into some dummy table that has no other contents, but which ensures that all the subtables are compiled together for the purposes of deduplication.

PR #3439, opened 2024-02-05
[otlLib] choose most compact contextual lookup format

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/2934

/cc @cmyr


behdad:
LGTM.

cmyr:
Awesome, thanks! This is the biggest source of diffs between feaLib and fea-rs. :)

PR #3446, opened 2024-02-13
[otlLib] Support building MATH table

khaledhosny:


Lib/fontTools/otlLib/builder.py
```diff
@@ -2906,3 +2907,294 @@ def _addName(ttFont, value, minNameID=0, windows=True, mac=True):
...
+    are the glyph names, and the values are tuples of glyph name and full
+    advance width.
+
+    'vertGlyphAssembly' is a dictionary of vertical glyph assemblies. The keys
+    are the glyph names, and the values are tuples of assembly parts and italics
+    correction. The assembly parts are tuples of glyph name, flags, start
+    connector length, end connector length, and full advance height.
+
+    'horizGlyphAssembly' is a dictionary of horizontal glyph assemblies. The
+    keys are the glyph names, and the values are tuples tuples of assembly parts
```
anthrotype, line 2962:
```suggestion
    keys are the glyph names, and the values are tuples of assembly parts
```

anthrotype (approved):
I don't know well the innards of MATH table but this looks good at first glance, thanks Khaled!

I also wonder if this piece of code could be moved in its own a submodule e.g. fontTools.otlLib.builder.math or something

khaledhosny:
> I also wonder if this piece of code could be moved in its own a submodule e.g. fontTools.otlLib.builder.math or something

Code for other tables (even not really layout tables like STAT) is all kept in the builder.py file, but this file is getting big indeed so it is fine by me either way. So if you feel strongly about it, I’ll move the code.

anthrotype:
it;s ok we can do separately later

PR #3449, opened 2024-02-17
code formatting and some comparisons

schriftgestalt:


behdad:
Did some tool flag these?

Some of the isinstance changes, like in the merge module,  might be wrong.

PR #3460, opened 2024-03-05
add filter pens that decompose components

anthrotype:
This PR adds a bunch of filter and recording pens to do composite glyph decomposition (I need to use them for some ufo2ft filter).


behdad:
Does this really need to be a class? It sounds trivial / no-op:
```
class DecomposingRecordingPointPen(DecomposingPointPen, RecordingPointPen):
```

PR #3464, opened 2024-03-19
Optimization on load

simoncozens:
Some OT tables rely on "aux" fields which are conditions written as strings; a classic example is the Device field which has the condition "DeltaFormat in (1,2,3)". This is currently stored as a string in the Convertor object, and is `eval`ed for each Device table it decompiles.

`eval` is very slow. You can get a small speedup - quite a large speedup for fonts with lots of Device/VariationIndex tables - by compiling the string and `eval`ing the compiled version later.

### DIFF
--- a/Lib/fontTools/ttLib/tables/otConverters.py
+++ b/Lib/fontTools/ttLib/tables/otConverters.py
@@ -153,6 +153,8 @@ class BaseConverter(object):
         self.name = name
         self.repeat = repeat
         self.aux = aux
+        if self.aux and not self.repeat:
+            self.aux = compile(self.aux, "<string>", "eval")
         self.tableClass = tableClass
         self.isCount = name.endswith("Count") or name in [
             "DesignAxisRecordSize",


simoncozens:
Thank you! Here's a font with a particularly large GPOS table:

```
font = TTFont("Bigfont.ttf")
count = len(font["GPOS"].table.LookupList.Lookup)
print(f"I have {count} GPOS lookups")
```

Before:

```
$ time python3 open.py
I have 46 GPOS lookups
python3 open.py  3.07s user 0.06s system 98% cpu 3.161 total
```

After:

```
$ time PYTHONPATH=Lib python3 open.py
I have 46 GPOS lookups
PYTHONPATH=Lib python3 open.py  2.29s user 0.05s system 99% cpu 2.355 total
```

anthrotype:
thanks. A test would also have been nice, to confirm this trick works reliably on all the pythons we support.

simoncozens:
That's what the existing test suite is for...

anthrotype:
thanks for confirming that the changes are covered by our current test suite

Issue #3469, opened 2024-03-28
avar-2 support for TTFont.getGlyphSet()?


justvanrossum:
It would be nice if `TTFont.normalizeLocation()` (as used by `getGlyphSet()`) would support the avar-2 format:

https://github.com/fonttools/fonttools/blob/430e364db7baf5494d1bdfe077154edc4dbd6b71/Lib/fontTools/ttLib/ttFont.py#L774-L804

Apart from the usefulness of having getGlyphSet() support avar-2, this would be very useful as a reference / example.

behdad:
Yep, I think there's a couple other places too that need updating.

behdad:
I'm working on this. Might take a few days to update all places.

behdad:
I'm working on this right now. Should have ttFont done today for sure. WIP will go into:
https://github.com/fonttools/fonttools/pull/3473

PR #3473, opened 2024-04-02
[avar2] Implement avar2 support in ttGlyphSet.

behdad:
Implementing avar2 support in ttGlyphSet

Fixes https://github.com/fonttools/fonttools/issues/3469


behdad:
This seems to be working now!  You can grab avar2 fonts from:

https://drive.google.com/drive/folders/1OjZFadiSKwzPnGWWCmNwscJN8bsjfqR0

behdad:
Will work on other modules too. But this is ready. We don't have subsetter for it though, so can't produce a test right now.

behdad:
> We don't have subsetter for it though

Nevermind. Subsetter doesn't change avar. Instancer does.

behdad:
Test added!

behdad:
Humm. Let's merge this and continue on other modules separately. I see three modules that need updating:

  - varLib.instancer. The hardest. I have jotted some notes here:
    https://docs.google.com/document/d/14elERu9Y_umFMg_lxnw7qtpskNlFvKhVtg2RNQteOek/edit

  - varLib.interpolatable. Not exactly sure what needs to be done, but I'll take a look.

  - designspaceLib. Also hard, because it uses single-axis `map_forward` and `map_backward` a lot. In other word, some of the functions there need to be deprecated and cannot work in a avar2 model.

behdad:
@justvanrossum Can I get a review from you please?

anthrotype (approved):
LGTM

justvanrossum:
It works beautifully! Thank you so much. Playing with svgPathPen as you suggested.

For more thorough code reviewing I'd need more time, but since Cosimo approved, feel free to merge.

Issue #2838, opened 2022-10-01
[varLib.cff] Suboptimal packing of deltas, or flaw in CFF2?


justvanrossum:
With a CFF2 VF built from the attached designspace project (modified from MutatorSans), the `S` rendered at `wdth=1000`, `wght=595` has a bump on the right hand side, even though its corresponding source does not:

![letter_S](https://user-images.githubusercontent.com/4246121/193420150-b9315322-f0c0-4b7f-8c0c-d2ecf056c1cd.svg)

The bump is at the start/end point of the contour.

The design space has five sources, one on each corner, and a support master between Light Wide and Bold Wide.

    Light    Medium   Bold
    Wide     Wide     Wide
    o--------o--------o
    |                 |
    |                 |
    |                 |
    |                 |
    |                 |
    |                 |
    O-----------------o
    Light             Bold
    Condensed         Condensed
    (neutral)

The _Medium Wide_ source is at the location shown (`wdth=1000`, `wght=595`)

What I _think_ happens is:

- The delta for that location is relative to the interpolation between _Light Wide_ and _Bold Wide_
- This interpolation has fractional coordinates
- The delta's are calculated for _relative_ coordinates (the differences between one point and the next), rounded to integers
- Leading to an accumulation of rounding errors
- To the extent that the final point is no longer coinciding with the starting point.
- (The difference between start and end point is ≈3 units horizontally, and ≈1.5 units vertically.)

On the one hand, in theory the varLib CFF2 packer could do a better job, but looking at the code this seems far from trivial. On the other hand this shows that CFF2 has pretty terrible properties for cases like this.

[mutator_sans_S.zip](https://github.com/fonttools/fonttools/files/9690800/mutator_sans_S.zip)

justvanrossum:
Btw. props to @behdad for the wonderful main prog in `svgPathPen.py`:
```
python -m fontTools.pens.svgPathPen MutatorSans-VF.otf S --variations "wght=595 wdth=1000" > letter_S.svg
```

behdad:
This fixes it:
```diff
diff --git a/Lib/fontTools/varLib/cff.py b/Lib/fontTools/varLib/cff.py
index b2fb51f1..2dac5123 100644
--- a/Lib/fontTools/varLib/cff.py
+++ b/Lib/fontTools/varLib/cff.py
@@ -502,7 +502,7 @@ class CFF2CharStringMergePen(T2CharStringPen):
        """
        def __init__(
[...]

PR #3481, opened 2024-04-13
[varLib.mutator] Improve CharString rounding

khaledhosny:
Since CharString point coordinates are relative, rounding each point independently can accumulate rounding errors causing to the last point to not match the first one, which leads to rendering artifacts.

This is a hack to round the absolute point coordinates, but using pens to convert the relative coordinates to absolute ones and back.

This is a hack since it drops hinting operators, and it can’t preserve the original operators, in addition to being slower.

There is probably a more clever way to doing this.

### DIFF
--- a/Lib/fontTools/pens/t2CharStringPen.py
+++ b/Lib/fontTools/pens/t2CharStringPen.py
@@ -47,7 +47,7 @@ class T2CharStringPen(BasePen):
     def _endPath(self):
         pass
 
-    def getCharString(self, private=None, globalSubrs=None, optimize=True):
+    def getProgram(self, optimize=True):
         commands = self._commands
         if optimize:
             maxstack = 48 if not self._CFF2 else 513
@@ -62,6 +62,10 @@ class T2CharStringPen(BasePen):
             program.insert(0, otRound(self._width))
         if not self._CFF2:
             program.append("endchar")
+        return program
+
+    def getCharString(self, private=None, globalSubrs=None, optimize=True):
+        program = self.getProgram(optimize=optimize)
         charString = T2CharString(
             program=program, private=private, globalSubrs=globalSubrs
         )
--- a/Lib/fontTools/varLib/mutator.py
+++ b/Lib/fontTools/varLib/mutator.py
@@ -69,6 +69,8 @@ def interpolate_cff2_PrivateDict(topDict, interpolateFromDeltas):
 
 
 def interpolate_cff2_charstrings(topDict, interpolateFromDeltas, glyphOrder):
+    from fontTools.pens.t2CharStringPen import T2CharStringPen
+
     charstrings = topDict.CharStrings
     for gname in glyphOrder:
         # Interpolate charstring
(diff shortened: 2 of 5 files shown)


justvanrossum:
Smells like a variant of this: https://github.com/fonttools/fonttools/issues/2838

khaledhosny:
> Smells like a variant of this: #2838

I think so (here at the instancing not merging side, though).

Not rounding at all works as well, but it gives larger file size and can expose bugs in implementations not prepared to handle fractional CFF coordinates.

behdad:
> Interestingly, harfbuzz instancing produces the same artifacts, but AFDKO’s tx does not.

@qxliu76

anthrotype:
thanks for spotting this Khaled

> There is probably a more clever way to doing this.
> harfbuzz instancing produces the same artifacts, but AFDKO’s tx does not.

would be nice to find out how tx manages to do this while, if I understand correctly, also preserving hinting and other operators. Maybe @skef could help undestand this?

skef:
`tx` translates the various PostScript-like CharString formats into callbacks of this form:

```
struct abfGlyphCallbacks_ {
    void *direct_ctx;
    void *indirect_ctx;
    abfGlyphInfo *info;
    int (*beg)(abfGlyphCallbacks *cb, abfGlyphInfo *info);
    void (*width)(abfGlyphCallbacks *cb, float hAdv);
    void (*move)(abfGlyphCallbacks *cb, float x0, float y0);
    void (*line)(abfGlyphCallbacks *cb, float x1, float y1);
    void (*curve)(abfGlyphCallbacks *cb,
                  float x1, float y1,
                  float x2, float y2,
                  float x3, float y3);
    void (*stem)(abfGlyphCallbacks *cb,
                 int flags, float edge0, float edge1);
    void (*flex)(abfGlyphCallbacks *cb, float depth,
                 float x1, float y1,
                 float x2, float y2,
                 float x3, float y3,
                 float x4, float y4,
                 float x5, float y5,
                 float x6, float y6);
    void (*genop)(abfGlyphCallbacks *cb, int cnt, float *args, int op);
    void (*seac)(abfGlyphCallbacks *cb,
                 float adx, float ady, int bchar, int achar);
    void (*end)(abfGlyphCallbacks *cb);
    void (*moveVF)(abfGlyphCallbacks *cb, abfBlendArg *x0, abfBlendArg *y0);
    void (*lineVF)(abfGlyphCallbacks *cb, abfBlendArg *x1, abfBlendArg *y1);
    void (*curveVF)(abfGlyphCallbacks *cb,
                    abfBlendArg *x1, abfBlendArg *y1,
                    abfBlendArg *x2, abfBlendArg *y2,
                    abfBlendArg *x3, abfBlendArg *y3);
    void (*stemVF)(abfGlyphCallbacks *cb,
                   int flags, abfBlendArg *edge0, abfBlendArg *edge1);
};
```

Note that the hints are represented with absolute positions and are (currently) in "hint substitution" style rather than "hintmask" style (for CFF variants they're translated back into hintmask style after processing).

behdad:
Random thing: if the test is CFF2, why does it have `endchar` operators?

justvanrossum:
> I think this PR predates CFF2 support in varLib.instancer which was added around May 2024.

Ah, that makes sense. I can confirm that the same issue exists in varLib.instancer, though.

behdad:
We should try converting all coordinates to absolute, do the instantiation, then revert.

behdad:
> We should try converting all coordinates to absolute, do the instantiation, then revert.

In the form of a charstring still I mean.

PR #3490, opened 2024-04-27
small fix to fontTools Documentation heading

rrawatt:


rrawatt:
Changes also made to optional.rst for easy and concise readbility

Doc/source/optional.rst
```diff
@@ -1,142 +1,112 @@
-:orphan:
-
 Optional Dependencies
 =====================
 
-The fonttools PyPI distribution also supports so-called "extras", i.e. a
-set of keywords that describe a group of additional dependencies, which can be
-used when installing via pip, or when specifying a requirement.
-For example:
+The FontTools PyPI distribution supports "extras", which are groups of additional dependencies that can be installed alongside FontTools via pip to enable specific features. For example:
```
n8willis, line 4: I know that there is already a lack of total consistency in the docs regarding the capitalization of "FontTools" when it is used as the name of the project (versus, for example, referring to the modules). I see that this PR uses "FontTools" here, but uses "fontTools" in index.rst. Might as well be consistent within the PR, IMO.

Personally, I think starting with the capital F is the way to go, both because it more clearly differentiates between "project name" and "token" use cases, but also because of the project logo....

If there are other strong opinions, though, I'm happy to hear what they all are.

Doc/source/optional.rst
```diff
@@ -1,142 +1,112 @@
...
+*Extra:* `lxml`
 
 :py:mod:`fontTools.ufoLib`
-^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-Package for reading and writing UFO source files; it requires:
+----------------
 
-* `fs <https://pypi.org/pypi/fs>`__: (aka ``pyfilesystem2``) filesystem abstraction layer.
+Package for reading and writing UFO source files. Dependencies include:
```
n8willis, line 29: My preference would be to spell out "Unified Font Object (UFO)" here, on first usage at this entry point. It's clear if the reader already knows what UFO is, but you are always going to get people who are diving in without a lot of prior background reading, because they've got some ultra-specific problem they're trying to fix.

Doc/source/optional.rst
```diff
@@ -1,142 +1,112 @@
...
 
+*Extra:* `woff`
 
 :py:mod:`fontTools.unicode`
-^^^^^^^^^^^^^^^^^^^^^^^^^^^
+-----------------
 
-To display the Unicode character names when dumping the ``cmap`` table
-with ``ttx`` we use the ``unicodedata`` module in the Standard Library.
+Displays Unicode character names when working with the `cmap` table dumps,with `ttx` we use the `unicodedata` module in the Standard Library.
```
n8willis, line 48: There's a space missing between "," and "with". I would also prefer to move towards "FontTools uses the `unicodedata` [...] over "we use [...]" but that might be a bigger discussion. Certainly it's not _wrong_ to say "FontTools" instead of "we" though. Also, I would drop the "the" before `cmap`; pretty minor though.

Doc/source/optional.rst
```diff
@@ -1,142 +1,112 @@
...
-To get the file creator and type of Macintosh PostScript Type 1 fonts
-on Python 3 you need to install the following module, as the old ``MacOS``
-module is no longer included in Mac Python:
+Retrieves information about Macintosh PostScript Type 1 fonts on Python 3. Requires:
 
-* `xattr <https://pypi.python.org/pypi/xattr>`__: Python wrapper for extended filesystem attributes
-  (macOS platform only).
-
-*Extra:* ``type1``
+- `xattr <https://pypi.python.org/pypi/xattr>`__: Python wrapper for extended filesystem attributes, macOS only.
```
n8willis, line 89: Feel free to disagree here, but I always worry about ambiguity when saying "macOS only" or the like unless it's totally spelled out. As in, are we saying that the module can only be installed on macOS? Or just that the xattrs are only found on macOS (different question how you'd be accessing those filesystems on another platform, though).

n8willis: (same "only"ness question on the CocoaPen bit below, I suppose, although I think mounting HFS/HFS+ disks is still possible on Linux, so the xattrs stuff *might* be usable there. I'd certainly be happy to learn for sure)

Doc/source/optional.rst
```diff
@@ -1,142 +1,112 @@
...
 
 :py:mod:`fontTools.varLib.interpolatable`
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+-------------------------------
 
-Module for finding wrong contour/component order between different masters.
-It requires one of the following packages in order to solve the so-called
-"minimum weight perfect matching problem in bipartite graphs", or
-the Assignment problem:
+Module for resolving contour or component order between different masters. Requires following packages to solve 'minimum weight perfect matching problem in bipartite graphs':
```
n8willis, line 59: Needs a "the" before "following packages" ... or "one of the"?

n8willis: I think that, if the Assignment Problem is well-defined enough to have a name, then it'd be ideal to drop a link about it where the name is mentioned. Wikipedia maybe?

Doc/source/optional.rst
```diff
@@ -1,142 +1,112 @@
...
+----------------
 
-* `fs <https://pypi.org/pypi/fs>`__: (aka ``pyfilesystem2``) filesystem abstraction layer.
+Package for reading and writing UFO source files. Dependencies include:
 
-* `enum34 <https://pypi.org/pypi/enum34>`__: backport for the built-in ``enum`` module (only required on Python < 3.4).
-
-*Extra:* ``ufo``
+- `fs <https://pypi.org/pypi/fs>`__: (aka ``pyfilesystem2``) filesystem abstraction layer
+- `enum34 <https://pypi.org/pypi/enum34>`__ backport for the built-in ``enum`` module (module for Python versions < 3.4)
```
n8willis, line 32: I'd cut out the second "module" there.

n8willis (changes requested):
Thanks for this!

I see a couple of things needing further fixing, in optional.rst, and as long as there is some real fixing, I decided to also note a few stylistic questions. A couple are just questions.

Doc/source/optional.rst
n8willis: I should maybe add that I don't think it's necessary to *always* spell out UFO and other abbreviations, but this is intro material, dealing with step-one installation stuff, so I think the need is higher here....

PR #3491, opened 2024-04-28
[feaLib.variableScalar] fix value_at_location() method

justvanrossum:
VariableScalar's value_at_location() method is untested and unused, but I find it useful (if only for playing and testing) and found that it's broken. I fixed it, but that makes an assumption about the intended call signature: I assume the intention was to feed it a location dict in user coordinates.

### DIFF
--- a/Lib/fontTools/feaLib/variableScalar.py
+++ b/Lib/fontTools/feaLib/variableScalar.py
@@ -75,10 +75,11 @@ class VariableScalar:
         return self.values[key]
 
     def value_at_location(self, location, model_cache=None, avar=None):
-        loc = location
+        loc = Location(location)
         if loc in self.values.keys():
             return self.values[loc]
         values = list(self.values.values())
+        loc = dict(self._normalized_location(loc))
         return self.model(model_cache, avar).interpolateFromMasters(loc, values)
 
     def model(self, model_cache=None, avar=None):


behdad:
Tangent: It really sucks that Adobe people decided to add user-space variable values to the .fea format, since those can't work with avar2.

justvanrossum:
I also don't quite like it, but given that `fvar` is all that's guaranteed to available during compile time, I can understand it.

I'm also not quite sure what the fundamental problem is: these coords get normalized without applying any avar, and that allows us to express everything that's needed, even if the coords aren't in designspace coords. These coordinates are _effectively_ source coordinates, even if they are expressed in user coordinates. Not sure I'm expressing myself very clearly :)

behdad:
> I'm also not quite sure what the fundamental problem is: these coords get normalized without applying any avar, and that allows us to express everything that's needed, even if the coords aren't in designspace coords. These coordinates are _effectively_ source coordinates, even if they are expressed in user coordinates. Not sure I'm expressing myself very clearly :)

Well if avar is not applied, then the "user" values mentioned in the .fea file won't actually match the observed user value coordinates. Eg. you say "switch to this other shape at wght=700" but then avar maps wght=700 somewhere different from what the .fea compiler does...

behdad:
But the fundamental issue is that with avar2, any point in the user space can be mapped to a point in the normalized space. But the same is not necessarily true about an orthogonal _plane_. That is to say, there might not be a simple formulation of 'wght >= 700' if eg. the font is a parametric design.

justvanrossum:
Any thoughts on this PR, @simoncozens, @anthrotype?

simoncozens (approved):

justvanrossum:
Thanks!

anthrotype:
> It really sucks that Adobe people decided to add user-space variable values to the .fea format, since those can't work with avar2.

to be fair, the variable scalar proposal came from @simoncozens, who also implemented it in fonttools, not from "Adobe people", and it's not even officially in the FEA spec (https://github.com/adobe-type-tools/afdko/issues/153, https://github.com/adobe-type-tools/afdko/pull/1350).

PR #3492, opened 2024-04-28
[psCharStrings] Shorten output of `encodeFloat`

dscorbett:
This optimizes `fontTools.misc.psCharStrings.encodeFloat` to return shorter outputs. There are three independent optimizations:

* Use exponential notation for multiples of 1000. For example, 10000 was encoded as `1e 10 00 0f` but is now `1e 1b 4f`.
* Never use a decimal point and an exponent in the same number. For example, 12345678000 was encoded as `1e 1a 23 45 67 8b 10 ff` but is now `1e 12 34 56 78 b3 ff`.
  * Furthermore, don’t use an exponent at all if this optimization would otherwise produce an exponent of 1. For example, 909876540 was encoded as `1e 9a 09 87 65 4b 08 ff` but is now `1e 90 98 76 54 0f`.
* Remove leading zeros from exponents. For example, 100000000 was encoded as `1e 1b 08 ff` but is now `1e 1b 8f`.

Sometimes the new output is not shorter (e.g. `1e e1 00 0f` vs. `1e e1 b3 ff`) but it is never longer.


behdad:
Thanks. Can you do the same in HarfBuzz's `encode_num_tp` function?

PR #3500, opened 2024-05-07
Better failure mode for sstruct

simoncozens:
Currently if you give a field value - any field value - a number it isn't expecting for whatever reason (float when uint is expected, number out of range, etc.), you get this:

```
font["OS/2"].usWinDescent = -600
font.save("foo.ttf")

...

  File "/opt/homebrew/lib/python3.11/site-packages/fontTools/ttLib/tables/O_S_2f_2.py", line 173, in compile
    data = sstruct.pack(OS2_format_2, self)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/fontTools/misc/sstruct.py", line 75, in pack
    data = struct.pack(*(formatstring,) + tuple(elements))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
struct.error: argument out of range
```

Which at least leads you to the right table, but nothing more than that. With this PR, you get:

```
    data = sstruct.pack(OS2_format_2, self)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/simon/others-repos/fonttools/Lib/fontTools/misc/sstruct.py", line 82, in pack
    raise ValueError(
ValueError: Value -600 does not fit in format H for usWinDescent
```

### DIFF
--- a/Lib/fontTools/misc/sstruct.py
+++ b/Lib/fontTools/misc/sstruct.py
@@ -64,7 +64,10 @@ def pack(fmt, obj):
     elements = []
     if not isinstance(obj, dict):
         obj = obj.__dict__
-    for name in names:
+    string_index = formatstring
+    if formatstring.startswith(">"):
+        string_index = formatstring[1:]
+    for ix, name in enumerate(names.keys()):
         value = obj[name]
         if name in fixes:
             # fixed point conversion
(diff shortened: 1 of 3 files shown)


simoncozens:
The only code I've found on GitHub that isn't a vendoring of fontTools is Inter's "fontinfo.py". Amusingly the reason it is reaching for those `names` is to turn them into a dict...

Also we probably won't be able to write GPKG tables correctly any more, but I reckon we should just leave that and see if anyone notices. :-)

anthrotype:
What even is a GPKG table?!

simoncozens:
SING Glyphlet stuff. Adobe idea for embeddable single-glyph fonts, abandoned in 2009.

Issue #1835, opened 2020-02-22
CFFFontSet.convertCFFToCFF2() vs. varLib.cff.lib_convertCFFToCFF2()


khaledhosny:
We seem to have to implementations for converting `CFF` table to `CFF2`, one in `CFFFontSet` and seems to be unused (and untested?) and one in `varLib.cff`. Any idea what is the difference between the two and if we can use the `CFFFontSet` one in varLib?

anthrotype:
I was also wondering today about this. Any news?

josh-hadley:
No. Started investigating this a while back when I first took this assignment, but ran into some blocks. I was trying to establish whether both things do the same thing or "close enough" but wasn't able to get that going easily and got onto more urgent matters.

@anthrotype do you have a suggestion for how to compare the results of both? I was discussing privately with @khaledhosny and came up with a simple script (basically just decompile a CFF, pass to both methods) but each one failed, for different reasons.

anthrotype:
One thing I noticed is that, after converting a CFF table with any of these two functions, the resulting CFF2 charstrings still contain the width as first token and thus are invalid.

Because of this, attempting to draw them onto a pen (which uses the T2OutlineExtractor) will then fail with the AssertionError about "CFF2 CharStrings must not have an initial width value".

I also noticed that fontTools.varLib.cff module contains a [CFFToCFF2OutlineExtractor](https://github.com/fonttools/fonttools/blob/74867c6ddbb09c929141fb1db9bda29ae7576824/Lib/fontTools/varLib/cff.py#L435-L446) that -- as the docstrings say -- "is used to remove the initial width from the CFF charstring without trying to add the width to self.nominalWidthX, which is None".

I just wanted to convert a CFF font to CFF2, but it doesn't seem that fonttools currently exposes an API for that that works end-to-end.

How does one go about converting a T2CharString that was originally contained in a CFF 1.0 table, to one that can be stored in a valid CFF2 table -- thus dropping width, endchar and return ops -- but _without_ having to draw the charstring onto a new T2CharStringPen?

Drawing onto a new T2CharStringPen and finally call `pen.getCharStrings()` kind of works, provided I override the charstring objects' `outlineExtractor` attribute to this above-mentioned `CFFToCFF2OutlineExtractor`, which ignores the widths.

[...]

PR #3506, opened 2024-05-14
[instancer] CFF2 instancing

behdad:
WIP. Almost there.


behdad:
I put the blends into the VarStore, instantiate that, and rebuild the blends from it.

Since we have code to instantiate VarStore, it works nicely.

behdad:
hmtx/vmtx need to be updated as well. The side-bearings are very hard to update (and unused for CFF).

behdad:
Did the advance widths.

behdad:
Oh, and pruning the VarStore; and dropping vsindex when not needed and VarStore when not needed.

behdad:
> Oh, and pruning the VarStore; and dropping vsindex when not needed and VarStore when not needed.

Done.

behdad:
> > * The font private program needs to be instantiated too. Donno how hard that one is.

I know how to do it. But can't find any fonts using those. So, for now, I left it unimplemented error.

I think this PR is ready. Volunteers for tests needed.

behdad:
We might also want to downgrade to CFF if static instancing.

behdad:
> > > * The font private program needs to be instantiated too. Donno how hard that one is.
> 
> I know how to do it. But can't find any fonts using those. So, for now, I left it unimplemented error.

This done now as well. I manually constructed a font.

behdad:
With test now.

behdad:
Looks like we only have code to upgrade CFF to CFF2, but not the other way around. Shouldn't be hard to write.

behdad:
It occurred to me that if we downgrade to CFF, we also have to remove overlaps. Is that desirable?

punchcutter:
Thanks for working on this. I've been waiting a long time to be able to do this with fontTools. Definitely would be nice to optionally remove overlaps with a flag.

behdad:
> Looks like we only have code to upgrade CFF to CFF2, but not the other way around. Shouldn't be hard to write.

I implemented a very rudimentary CFF2ToCFF. Doesn't remove overlaps, but seems to do enough that ots passes and both svgPathPen and hb-view show text. My confidence is rather low though, I'm sure there are incomplete corners.

behdad:
> > 
> 
> I implemented a very rudimentary CFF2ToCFF. Doesn't remove overlaps, but seems to do enough that ots passes and both svgPathPen and hb-view show text. My confidence is rather low though, I'm sure there are incomplete corners.

While writing that I found that CFFToCFF2 is missing a lot actually. It doesn't add endchar and return ops to charstrings for example.

behdad:
> We have two convertCFFToCFF2 #1835

Thanks. Very timely. I'm working on it.

behdad:
> > We have two convertCFFToCFF2 #1835
> 
> Thanks. Very timely. I'm working on it.

Humm. M'okay. It's a bit more work. I'll do it in a separate PR. Can I get a review on this one?

Lib/fontTools/cffLib/CFFToCFF2.py
```diff
@@ -101,9 +102,66 @@ def _convertCFFToCFF2(cff, otFont):
...
+        program = cs.program
+        if fdIndex == None:
+            fdIndex = 0
+
+        # Intentionally use None for nominalWidthX, such that any
+        # CharString that has an explicit width encoded will throw.
+        extractor = T2WidthExtractor(localSubrs[fdIndex], globalSubrs, None, 0)
+        try:
+            extractor.execute(cs)
+        except TypeError:
```
anthrotype: TypeError seems a bit too broad here, I wonder if the T2WidthExtractor could throw some more specific exception that you can catch here?

behdad: Yeah I don't like it either. I'll make a custom object that throws a custom exception on addition. No need to change the `T2WidthExtractor`.

behdad: The merger uses a custom object as defaultWidth instead and detects if that was returned as width. But this way using exceptions stops as soon as the question is answered so I went with this model this time.

Lib/fontTools/cffLib/CFFToCFF2.py
```diff
@@ -101,9 +102,66 @@ def _convertCFFToCFF2(cff, otFont):
...
+            # Program has explicit width. We want to drop it, but can't
+            # just pop the first number since it may be a subroutine call.
+            # Instead, when seeing that, we embed the subroutine and recurse.
+            # This has the problem that some subroutines might become unused.
+            # We don't currently prune those. Subset module has code for this
+            # kind of stuff, possibly plug it in here if pruning becomes needed.
```
anthrotype: +1 on moving the subset's subrs pruning to a place where can be reused here

Lib/fontTools/cffLib/CFFToCFF2.py
anthrotype: i'm simply concerned that a TypeError may trigger inadvertently for unrelated bugs, activating the subrs-embeddindg-width-popping logic below

Lib/fontTools/cffLib/CFFToCFF2.py
```diff
@@ -0,0 +1,258 @@
...
+
+    # TODO(behdad): What does the following comment even mean? Both CFF and CFF2
+    # use the same T2Charstring class.
+
+    # At this point, the Subrs and Charstrings are all still T2Charstring class
+    # easiest to fix this by compiling, then decompiling again
+    file = BytesIO()
+    cff.compile(file, otFont, isCFF2=True)
+    file.seek(0)
+    cff.decompile(file, otFont, isCFF2=True)
```
anthrotype, line 177: could it be that this block is no longer needed after you added the T2WidthExtractor thing?

Lib/fontTools/cffLib/CFF2ToCFF.py
```diff
@@ -0,0 +1,166 @@
+"""CFF to CFF2 converter."""
```
anthrotype, line 1:
```suggestion
"""CFF2 to CFF converter."""
```

Lib/fontTools/cffLib/CFF2ToCFF.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+log = logging.getLogger("fontTools.cffLib")
+
+
+def _convertCFF2ToCFF(cff, otFont):
+    """Converts this object from CFF2 format to CFF format. This conversion
+    is done 'in-place'. The conversion cannot be reversed.
+
+    The CFF2 font cannot be variable. (TODO Accept those and convert to the
```
anthrotype, line 21: is the input CFF2 also assumed to _not_ be already subroutinized? Or it doesn't matter in this case?

anthrotype: oh I see you do seem to handle GlobalSubrs and Private.Subrs in here. I was simply wondering if inserting the optimal width at index 0 can be done unconditionally even in presence of subroutines. Ignore me if it doesn't make any sense

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -765,22 +1024,52 @@ def _remapVarIdxMap(table, attrName, varIndexMapping, glyphOrder):
...
+            metrics = metricsTable.metrics
+            for glyphName, (advanceWidth, sb) in metrics.items():
+                if advMapping:
+                    varIdx = advMapping.mapping[glyphName]
+                else:
+                    varIdx = varfont.getGlyphID(glyphName)
+                metrics[glyphName] = (advanceWidth + round(defaultDeltas[varIdx]), sb)
+
+            # TODO if tableTag == "VVAR" and getattr(vhvar, tableFields.vOrigMapping),
+            # update VORG table as well
```
anthrotype, line 1066: how about raising some error if VORG is present? a user might not notice this is not implemented

anthrotype (approved):

Lib/fontTools/cffLib/CFF2ToCFF.py
behdad: Yeah there's no problem with subroutines. I first compute optimal default and nominal widths from the hmtx table data, then just insert at index 0 the width for those glyphs that need it. It works with subroutines as well.

Lib/fontTools/cffLib/CFFToCFF2.py
behdad: Humm. I'm not sure. I'll check. Maybe. But even with the block I was getting ots fail re the width without explicitly implementing it.

It talks about T2Charstring class... @readroberts maybe is still around?

I'll go ahead and delete it I think.

behdad: I added a comment re what I think it's doing. It's fine for now. If it's slow we can look into removing and implement what I suggest in the comment.

Lib/fontTools/varLib/instancer/__init__.py
behdad: I logged a warning.

Lib/fontTools/cffLib/CFFToCFF2.py
behdad: https://github.com/fonttools/fonttools/issues/3511

Issue #3511, opened 2024-05-17
[CFFToCFF2.py] Prune unused subrs?


behdad:
When we go about removing the encoded width in CFF charstrings, we might desubroutinizes a bit. Move unused-subrs pruning from subset module to cffLib and call it here.

PR #3517, opened 2024-05-21
Move CFF transforms from subset module to cffLib

behdad:
`remove_hints` and `remove_unused_subroutines`.

To be used eg. in https://github.com/fonttools/fonttools/issues/3511


behdad:
This is just a mechanical moving of code.

khaledhosny (approved):

PR #3518, opened 2024-05-23
[CFFToCFF2] Remove unused subroutines if necessary

behdad:
The transforms.py changes are for them to work with fonts without font.charset. I'm not sure how we didn't catch these before.

Ideally needs test.

Fixes https://github.com/fonttools/fonttools/issues/3511

### DIFF
--- a/Lib/fontTools/cffLib/CFFToCFF2.py
+++ b/Lib/fontTools/cffLib/CFFToCFF2.py
@@ -65,6 +65,7 @@ def _convertCFFToCFF2(cff, otFont):
             program[min(i, j) :] = []
 
     # Clean up glyph charstrings
+    removeUnusedSubrs = False
     nominalWidthXError = _NominalWidthUsedError()
     for glyphName in charStrings.keys():
         cs, fdIndex = charStrings.getItemAndSelector(glyphName)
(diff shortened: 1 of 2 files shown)


Lib/fontTools/cffLib/CFFToCFF2.py
```diff
@@ -90,6 +91,7 @@ def _convertCFFToCFF2(cff, otFont):
             # We don't currently prune those. Subset module has code for this
             # kind of stuff, possibly plug it in here if pruning becomes needed.
```
khaledhosny, line 92: This comment is outdated now?

Lib/fontTools/cffLib/CFFToCFF2.py
```diff
@@ -103,6 +105,9 @@ def _convertCFFToCFF2(cff, otFont):
         if program and program[-1] == "endchar":
             program.pop()
 
+    if removeUnusedSubrs:
+        cff.remove_unused_subroutines()
```
khaledhosny, line 109: Codecov suggests that existing tests do not cover this.

Lib/fontTools/cffLib/transforms.py
```diff
@@ -362,8 +362,7 @@ def remove_hints(cff):
         #    we have seen any non-hint operators so far and do the right
         #    thing, recursively... Good luck understanding that :(
         css = set()
-        for g in font.charset:
-            c, _ = cs.getItemAndSelector(g)
+        for c in cs.values():
```
khaledhosny, line 365: Might warrant a test, if possible, since this was not caught earlier.

khaledhosny (approved):

Lib/fontTools/cffLib/CFFToCFF2.py
behdad: Yeah I have to dig which glyph of NotoSansCJK or another font I was testing triggers this and subset it to a test.

Lib/fontTools/cffLib/transforms.py
behdad: Will add. Is probably enough to run on a CFF2.

behdad:
Thanks Cosimo. I'll still try to add tests.

Lib/fontTools/cffLib/CFFToCFF2.py
behdad: Adding that here: https://github.com/fonttools/fonttools/pull/3525

PR #3524, opened 2024-05-24
really drop python<=3.7 support

kloczek:
Filtrer all code over `pyupgrade --py38-pus`.


anthrotype:
thanks. I don't oppose these kind of automatic code cleanups in principle, but at least we should try to make the CI pass. I'll have another look if/when it goes green

kloczek:
As you know which one tool can be used feel free to close this PR and apply those changes other way 😋

anthrotype:
thanks, but there are now some merge conflicts apparently

Issue #3521, opened 2024-05-23
`CFFToCFF2` (and possibly `CFF2ToCFF`) is very brittle


behdad:
Many fonts crash it currently. Keeping this issue to track my progress mostly.

behdad:
Looks like now the main issue is broken files generated by `CFF2ToCFF`:

https://github.com/fonttools/fonttools/issues/3534#issuecomment-2134012246

liZe:
I’ve tried many tools, including PDF readers (my use case is to include fonts in PDF files):
- Firefox’s PDF reader prints the "CFF : Failed to parse Top DICT Data"
- Other browsers I’ve tried are OK with the PDF
- No browser can use the font in a HTML/CSS document, with various useless error messages
- Evince prints "some font thing failed"
- Poppler prints "Couldn't create a font for 'GHXLSD+Cantarell-Light'" (that’s the internal name of the font in the PDF)
- Ghostscript and PDF validators I’ve tried are happy with the PDF
- Fontforge crashes when opening the font file
- Gnome Fonts prints "Could not display font"

Don’t hesitate to ping me if I can do anything to help.

behdad:
Thanks. I'm debugging.

behdad:
Working here now: https://github.com/fonttools/fonttools/pull/3536

Still no luck with CFF2ToCFF though.

behdad:
@liZe please give https://github.com/fonttools/fonttools/pull/3536 a try. I think it should be good now.

liZe:
> @liZe please give #3536 a try. I think it should be good now.

I get this stacktrace:
```
Traceback (most recent call last):
  File "/tmp/subset_font.py", line 7, in <module>
    output_font.save('output.otf')
  File "/home/lize/.local/share/venv/weasyprint/lib/python3.12/site-packages/fontTools/ttLib/ttFont.py", line 216, in save
    writer_reordersTables = self._save(tmp)
                            ^^^^^^^^^^^^^^^
  File "/home/lize/.local/share/venv/weasyprint/lib/python3.12/site-packages/fontTools/ttLib/ttFont.py", line 263, in _save
    self._writeTable(tag, writer, done, tableCache)
  File "/home/lize/.local/share/venv/weasyprint/lib/python3.12/site-packages/fontTools/ttLib/ttFont.py", line 713, in _writeTable
    tabledata = self.getTableData(tag)
                ^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3525, opened 2024-05-24
CFF converters fixes

behdad:
To fix https://github.com/fonttools/fonttools/issues/3521


khaledhosny (approved):
I’m going to cut a new release now, do you want to merge this before or after the release?

behdad:
> I’m going to cut a new release now, do you want to merge this before or after the release?

Preferably yes. It clearly fixes non-FDArray fonts which crashed before.

PR #3526, opened 2024-05-24
Ensure that scipy/numpy output is JSON-serializable

### LINKED ISSUE #3522: varLib.interpolatable --json with scipy installed: Object of type int64 is not JSON serializable
Hello.

It seems that, the moment I install `scipy` into my project, `varLib.interpolatable` starts using it, and then fails outputting JSON with:

```shell
TypeError: ('Object of type int64 is not JSON serializable')
```

To clarify, the above happens when running the following with `scipy` installed:
```shell
fonttools varLib.interpolatable font1.ttf font2.ttf --json
```

I assume this may be due to some conditional `scipy` imports? E.g.:
https://github.com/fonttools/fonttools/blob/90fed08964bba2b81f0e1ec9222c95b0bee05384/Lib/fontTools/varLib/interpolatableHelpers.py#L174-L188
    
https://github.com/fonttools/fonttools/blob/90fed08964bba2b81f0e1ec9222c95b0bee05384/Lib/fontTools/varLib/interpolatableHelpers.py#L306-L308

Happy to dig deeper, and maybe provide a minimal reproduction, with some guidance on how to best supply that?

Thank you :)

PeterDekkers:
Resolves #3522

### DIFF
--- a/Lib/fontTools/varLib/interpolatableHelpers.py
+++ b/Lib/fontTools/varLib/interpolatableHelpers.py
@@ -143,6 +143,9 @@ def min_cost_perfect_bipartite_matching_scipy(G):
     n = len(G)
     rows, cols = linear_sum_assignment(G)
     assert (rows == list(range(n))).all()
+    # Convert numpy array and integer to Python types,
+    # to ensure that this is JSON-serializable.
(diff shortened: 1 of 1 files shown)


behdad:
Thanks.

Issue #3512, opened 2024-05-17
[CFF2ToCFF] remove-overlaps


behdad:

behdad:
@khaledhosny Any chance you can do this?

khaledhosny:
I can work on it. We have `fontTools.ttLib.removeOverlaps` but it does not support CFF table, so I’ll try to add it there and use it in the instancer (which already uses it to remove overlaps for `glyf` table).

ftCLI:
If can be of any help, I’ve already adapted removeOverlaps to work with CFF.

https://github.com/ftCLI/FoundryTools-CLI-2/blob/main/foundrytools_cli_2/lib/skia/skia_tools.py

It also corrects contours direction and removes tiny paths (like AFDKO’s checkoutlinesufo) but you can easily strip those parts

khaledhosny:
This is already fixed.

PR #3528, opened 2024-05-25
[ttLib.removeOverlaps] Support CFF table

khaledhosny:
Part of #3512.

### DIFF
--- a/Lib/fontTools/ttLib/removeOverlaps.py
+++ b/Lib/fontTools/ttLib/removeOverlaps.py
@@ -7,11 +7,14 @@ import itertools
 import logging
 from typing import Callable, Iterable, Optional, Mapping
 
-from fontTools.misc.roundTools import otRound
+from fontTools.cffLib import CFFFontSet
 from fontTools.ttLib import ttFont
 from fontTools.ttLib.tables import _g_l_y_f
 from fontTools.ttLib.tables import _h_m_t_x
+from fontTools.misc.psCharStrings import T2CharString
+from fontTools.misc.roundTools import otRound, noRound
 from fontTools.pens.ttGlyphPen import TTGlyphPen
+from fontTools.pens.t2CharStringPen import T2CharStringPen
 
 import pathops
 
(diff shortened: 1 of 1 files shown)


behdad:
Thanks. We probably should add an option (default on?) to remove unused subroutines if anything was changed.

behdad:
Please make the `remove_unused_subroutines` optional, since eg. `CFF2ToCFF` does it itself too...

khaledhosny:
> Please make the `remove_unused_subroutines` optional, since eg. `CFF2ToCFF` does it itself too...

Done.

behdad:
Please also add `*, ` before most of the options so they can only be called by name.

khaledhosny:
> Please also add `*, ` before most of the options so they can only be called by name.

I can do that for the new functions, but wouldn’t doing so for the old ones potentially break user code?

behdad:
> I can do that for the new functions,

Do it for new arguments.

anthrotype:
could we add some tests in https://github.com/fonttools/fonttools/blob/main/Tests/ttLib/removeOverlaps_test.py ?

khaledhosny:
> could we add some tests in https://github.com/fonttools/fonttools/blob/main/Tests/ttLib/removeOverlaps_test.py ?

I wondered about this as well, but the existing code does not have much tests either, so I thought it might be implicitly testes by the instancer, so this code would get some testing when it get integrated into the instancer as well.

anthrotype:
oh you're right actually.

Lib/fontTools/ttLib/removeOverlaps.py
```diff
@@ -225,27 +219,170 @@ def removeOverlaps(
...
+    log.debug("Removed overlaps for %s glyphs:\n%s", len(modified), " ".join(modified))
+
+
+def removeOverlaps(
+    font: ttFont.TTFont,
+    glyphNames: Optional[Iterable[str]] = None,
+    removeHinting: bool = True,
+    ignoreErrors: bool = False,
+    *,
+    removeUnusedSubroutines: bool = True,
```
anthrotype, line 285: why would one wish to keep unused subroutines?

anthrotype (approved):

Lib/fontTools/ttLib/removeOverlaps.py
khaledhosny: Behdad requested it, since code calling this might have removed unused subroutines already.

behdad: Yeah, in `CFF2ToCFF` we have done that already.

PR #3536, opened 2024-05-27
CFF convertors more fixes

behdad:
So far in `CFFToCFF2`. Hoping to find what's so broken with `CFF2ToCFF` as well.

### DIFF
--- a/Lib/fontTools/cffLib/CFF2ToCFF.py
+++ b/Lib/fontTools/cffLib/CFF2ToCFF.py
@@ -32,7 +32,7 @@ def _convertCFF2ToCFF(cff, otFont):
 
     cff.major = 1
 
-    topDictData = TopDictIndex(None, isCFF2=True)
+    topDictData = TopDictIndex(None)
     for item in cff.topDictIndex:
         # Iterate over, such that all are decompiled
         topDictData.append(item)
--- a/Lib/fontTools/cffLib/CFF2ToCFF.py
+++ b/Lib/fontTools/cffLib/CFF2ToCFF.py
@@ -100,6 +100,21 @@ def _convertCFF2ToCFF(cff, otFont):
         if width != private.defaultWidthX:
             cs.program.insert(0, width - private.nominalWidthX)
 
+    mapping = {
+        name: ("cid" + str(n) if n else ".notdef")
+        for n, name in enumerate(topDict.charset)
+    }
+    topDict.charset = [
+        "cid" + str(n) if n else ".notdef" for n in range(len(topDict.charset))
+    ]
+    charStrings.charStrings = {
+        mapping[name]: v for name, v in charStrings.charStrings.items()
+    }
+
+    # I'm not sure why the following is *not* necessary. And it breaks
+    # the output if I add it.
+    # topDict.ROS = ("Adobe", "Identity", 0)
+
 
 def convertCFF2ToCFF(font, *, updatePostTable=True):
     cff = font["CFF2"].cff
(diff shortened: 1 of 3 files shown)


behdad:
Humm. `CFF2ToCFF` still fails with `recalcBBoxes=False`. Is rather hard to fix, since the CFF table has cid names and the rest of the font have the original names...

behdad:
> Humm. `CFF2ToCFF` still fails with `recalcBBoxes=False`. Is rather hard to fix, since the CFF table has cid names and the rest of the font have the original names...

Since we are converting the CFF2 to a CID-keyed CFF, the glyphnames are going to be dropped. We might not be able to just change create the `CFF ` table and compile the font :(.  I'll see if I can hack `cffLib` to work though.

behdad:
We really need a CFF lawyer. Maybe I can get help from Ken Lunde.

punchcutter:
Since a CFF2 font from CID-keyed sources doesn't have any recollection of the ROS there's no way to know the CID mapping unless that can be passed along as another parameter and understood.

behdad:
> Since a CFF2 font from CID-keyed sources doesn't have any recollection of the ROS there's no way to know the CID mapping unless that can be passed along as another parameter and understood.

Humm. I'm encoding a charset in the CFF file. What I don't understand is how all software seems to accept CID-keyed fonts (ie. FDSelect etc) without ROS. Or am I wrong and that's a valid CFF?

punchcutter:
I'm not sure how you can have CID-keyed without a ROS. For any ROS other than Adobe-Identity-0 it's THE thing that tells you what's what.

behdad:
> I'm not sure how you can have CID-keyed without a ROS. For any ROS other than Adobe-Identity-0 it's THE thing that tells you what's what.

punchcutter:
There's also no guarantee of a post 2.0 table with glyph names. So it really feels like we need some extra data to properly cover all possible CID situations.

khaledhosny:
AFDKO’s tx can convert CFF2 to CFF, may be try it to see what it is doing in these situations?

punchcutter:
tx doesn't do anything. I think it has a bug and sets the ROS to Adobe-Identity-1 which isn't a thing. I use tx for this, but only for the outline data and then I need to build the new OTF with makeotf. tx can't pull a finished OTF instance like fonttools can.

behdad:
> sets the ROS to Adobe-Identity-1 which isn't a thing

What do you mean it isn't a thing?

behdad:
Oh, I didn't realize there's Adobe-Identity-0 and Adobe-Identity-1. That said, HB code doesn't do anything with the ROS. So it's basically assuming an identity map?

behdad:
> Oh, I didn't realize there's Adobe-Identity-0 and Adobe-Identity-1. That said, HB code doesn't do anything with the ROS. So it's basically assuming an identity map?

Or said another way, are these relevant in OT fonts? Or cmap skips the ROS / charset mapping?

behdad:
I'm not sure what to do here, but I suggest merging. The command-line `CFF2ToCFF` seems to work well; not so much the function, since TTFont's recalcBBoxes=True fails. I'll keep thinking about a solution.

behdad:
Not sure what to do about this. Land? The code is still a mystery to me, but I don't see what letting this rot will buy us.

anthrotype:
maybe merge but file an issue summarizing the remaining problems?

khaledhosny (approved):
I second what @anthrotype said above.

behdad:
Thanks. The issue remains to be https://github.com/fonttools/fonttools/issues/3521 I'll comment there.

PR #3541, opened 2024-05-29
[VARC] Update to reinterpretation of RESET_UNSPECIFIED_AXES

behdad:
Implements:
https://github.com/harfbuzz/boring-expansion-spec/commit/3f5cdc45d2373d274290ab34cf4fc7d48bcf01ca

### DIFF
--- a/Lib/fontTools/ttLib/ttGlyphSet.py
+++ b/Lib/fontTools/ttLib/ttGlyphSet.py
@@ -25,13 +25,7 @@ class _TTGlyphSet(Mapping):
     def __init__(self, font, location, glyphsMapping, *, recalcBounds=True):
         self.recalcBounds = recalcBounds
         self.font = font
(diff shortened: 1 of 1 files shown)


justvanrossum (commented):
I now remember better what "rawLocation" is about: it should perhaps be called "sparseLocation". 

It is needed to preserve the information which axes are specified for addVarComponent(). Without the rawLocation mechanism, it is not possible to *extract* variable components correctly, you can only *render* them.

I will experiment a bit, but I think instead of this PR, all we need to do is to change the reset of rawLocation to self.originalLocation.copy(). The `self.rawLocation = {}` in the else is (I think) a crucial thing to keep. I'll get back about this later.

behdad:
Okay let's continue in https://github.com/fonttools/fonttools/pull/3549

PR #3543, opened 2024-05-30
Revert "[svgPen] Write two digits after decimal by default"

madig:
This reverts commit 42d6b6b4fedf51dd741d3134da74df04339335b4.

As discussed with @anthrotype, this changed default behavior.


behdad:
Fine to revert. I'll open an issue for it. I think adding a new argument to do compact writing is preferred for me than making all callers provide their custom `ntos`. My use-case for this was when we pass `--variations` to `svgPathPen` cmdline, everything suddenly becomes float, and written to many digits.

PR #3548, opened 2024-05-30
[instancer] Fix output filename decision-making

behdad:
All limits are tuples now when not None. The old logic was broken and for the following command:

$ fonttools varLib.instancer  AdobeVFPrototype.otf CNTR=50:80 wght=900

it was saving the output with the name suffix `-instance`, whereas it's clearly a partial instantiation. This fixes that.


behdad:
Sorry I don't have tests for this.

anthrotype:
LGTM. Tested locally with wght="drop", wght=900 or wght=900:900 and all work

PR #3549, opened 2024-05-30
[VARC] Special-case hidden axes when applying RESET_UNSPECIFIED_AXES to pen.addVarComponent()

justvanrossum:
Alternative to #3541.

As explained in https://github.com/fonttools/fonttools/pull/3541#pullrequestreview-2088359154, the rawLocation mechanism is needed to meaningfully extract variable components via addVarComponent().

But this gives a bit of a conundrum with the new meaning of RESET_UNSPECIFIED_AXES.

Old meaning: when RESET_UNSPECIFIED_AXES is set, reset unspecified axes to default
New meaning: when RESET_UNSPECIFIED_AXES is set, reset unspecified axes the current font location

- to extract variable components for a glyph, we instantiate the glyph at all locations for which the glyph has variations
- this means we also need to specify values for the "private" axes used for variable components
- but _these_ values should not "leak" into descendants
- for normal rendering this is no problem, as the private axes are not user-settable (they are hidden), so the "global font location" for these axes is at default anyway

Long story short, to implement RESET_UNSPECIFIED_AXES in a useful way for addVarComponent(), we need to special case hidden axes.

### DIFF
--- a/Lib/fontTools/ttLib/ttGlyphSet.py
+++ b/Lib/fontTools/ttLib/ttGlyphSet.py
@@ -25,14 +25,22 @@ class _TTGlyphSet(Mapping):
     def __init__(self, font, location, glyphsMapping, *, recalcBounds=True):
         self.recalcBounds = recalcBounds
         self.font = font
-        self.defaultLocationNormalized = (
+        defaultLocationNormalized = (
             {axis.axisTag: 0 for axis in self.font["fvar"].axes}
             if "fvar" in self.font
             else {}
         )
+        hiddenAxisTags = {
+            axis.axisTag
+            for axis in self.font["fvar"].axes
+            if not axis.flags & 0x0001  # HIDDEN_AXIS
+        }
         self.location = location if location is not None else {}
         self.rawLocation = {}  # VarComponent-only location
         self.originalLocation = location if location is not None else {}
+        self.originalLocationNonHidden = defaultLocationNormalized | {
+            k: v for k, v in self.originalLocation.items() if k in hiddenAxisTags
+        }
         self.depth = 0
         self.locationStack = []
         self.rawLocationStack = []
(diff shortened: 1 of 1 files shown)


behdad:
> we need to special case hidden axes.

I'm a bit uncomfortable with that. Can we change the protocol and send two locations down?

justvanrossum:
> I'm a bit uncomfortable with that. Can we change the protocol and send two locations down?

Hm, then we burden the addVarComponent signature with this problem. Not sure which is better.

The way we currently compile with RESET_UNSPECIFIED_AXES already depends on those axes always being at the default: as soon as you set one to a non-default value, the font breaks, as can be seen in (my private build of) FontGoggles:

https://github.com/fonttools/fonttools/assets/4246121/02b00d83-b20d-49dd-bdea-3102bd097207

So those axes are already special, and to me special-casing them here doesn't make things worse.

behdad:
> Hm, then we burden the addVarComponent signature with this problem. Not sure which is better.

What if we add a lower-level `addVARCComponent` or something that takes exactly what's in the format? The location as well as the flag?

Let me also think more. In my compiler I drop encoding axes if they have the same value as the parent glyph. Let me think more about your requirements and get back to you.

justvanrossum:
I'm having a hard time to come up with a use case for pen.addVarComponent() that is other than mine... There is of course the option that I don't use the pen protocol for this at all, and just read VARC directly. Maybe that makes more sense.

There's also your earlier idea if having _two_ flags:
1. reset to font location
2. reset to default

The first is handy when we pass down (user) axes to deeply nested components. The second is what we really need for private axes.

justvanrossum:
(Hm, additionally, pen.addVarComponent() doesn't support extracting conditions (which Fontra doesn't yet support), but that also suggests that the pen protocol is perhaps not the right tool to extract variable components.)

Issue #3546, opened 2024-05-30
AttributeError: 'table__a_v_a_r' object has no attribute 'majorVersion'


anthrotype:
@madig wrote in chat

> fonttools 4.52.(0|1) gives me an AttributeError: 'table__a_v_a_r' object has no attribute 'majorVersion' when i try to varfont.getGlyphSet(location=coordinate) on a compiled but not yet written to disk font
> it works with 4.51.0
> here's a repro with an up to date mutatorsans checkout:

```python
from fontTools.designspaceLib import DesignSpaceDocument
import ufo2ft
import ufoLib2

d = DesignSpaceDocument.fromfile("/home/nikolaus/mutatorSans/MutatorSans.designspace")
d.axes[1].minimum = 100
d.axes[1].default = 100
d.axes[1].maximum = 900
d.axes[1].map = [(100, 0), (500, 400), (900, 1000)]
d.loadSourceFonts(ufoLib2.Font.open)
ts = ufo2ft.compileVariableTTFs(d)
t = ts["MutatorSans_All_Variable"]
assert "avar" in t
gs = t.getGlyphSet(location={"wght": 500, "wdth": 500})Traceback (most recent call last):
  File "/home/nikolaus/lubalin-recast/test/test.py", line 14, in <module>
    gs = t.getGlyphSet(location={"wght": 500, "wdth": 500})
  File "/home/nikolaus/lubalin-recast/.venv/lib/python3.10/site-packages/fontTools/ttLib/ttFont.py", line 771, in getGlyphSet
    location = self.normalizeLocation(location)
  File "/home/nikolaus/lubalin-recast/.venv/lib/python3.10/site-packages/fontTools/ttLib/ttFont.py", line 801, in normalizeLocation
    location = self["avar"].renormalizeLocation(location, self)
  File "/home/nikolaus/lubalin-recast/.venv/lib/python3.10/site-packages/fontTools/ttLib/tables/_a_v_a_r.py", line 146, in renormalizeLocation
    if self.majorVersion not in (1, 2):
AttributeError: 'table__a_v_a_r' object has no attribute 'majorVersion'
```

PR #3550, opened 2024-05-30
[avar] Fix assertion

behdad:
Fixes https://github.com/fonttools/fonttools/issues/3546


madig:
That fixes the issue for me :)

madig (approved):

PR #3561, opened 2024-06-03
Check unicode elements have required hex attribute

drj11:
I created a UFO that erroneously had a `<unicode>` element without a `hex` attribute. Naturally instead of just fixing this in my own code, i poked about inside fonttools to work out how to improve the validation, run the tests, and blah blah blah.

### DIFF
--- a/Lib/fontTools/ufoLib/glifLib.py
+++ b/Lib/fontTools/ufoLib/glifLib.py
@@ -1191,12 +1191,12 @@ def _readGlyphFromTreeFormat1(
             haveSeenAdvance = True
             _readAdvance(glyphObject, element)
         elif element.tag == "unicode":
-            if element.get("hex") is None:
+            v = element.get("hex")
+            if v is None:
                 raise GlifLibError(
                     "A unicode element is missing its required hex attribute."
                 )
             try:
-                v = element.get("hex")
                 v = int(v, 16)
                 if v not in unicodes:
                     unicodes.append(v)
@@ -1258,12 +1258,12 @@ def _readGlyphFromTreeFormat2(
             haveSeenAdvance = True
             _readAdvance(glyphObject, element)
         elif element.tag == "unicode":
-            if element.get("hex") is None:
+            v = element.get("hex")
+            if v is None:
                 raise GlifLibError(
                     "A unicode element is missing its required hex attribute."
                 )
             try:
-                v = element.get("hex")
                 v = int(v, 16)
                 if v not in unicodes:
                     unicodes.append(v)
--- a/Lib/fontTools/ufoLib/glifLib.py
+++ b/Lib/fontTools/ufoLib/glifLib.py
@@ -1191,6 +1191,10 @@ def _readGlyphFromTreeFormat1(
             haveSeenAdvance = True
             _readAdvance(glyphObject, element)
         elif element.tag == "unicode":
+            if element.get("hex") is None:
+                raise GlifLibError(
+                    "A unicode element is missing its required hex attribute."
+                )
             try:
                 v = element.get("hex")
                 v = int(v, 16)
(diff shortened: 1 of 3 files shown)


simoncozens:
I'm not clear about this. I mean, you're probably right, but there's nothing actually in the UFO spec which says this attribute is required. (The UFO spec is a bit wooly on whether attributes are required by default or optional by default.)

drj11:
I agree with your assessment that the UFO spec is not terribly clear on whether this attribute is required or not. For reference, link to spec https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#unicode

This patch was mostly motivated by the fact that a <unicode> element without a `hex` attribute makes the code fail with a fairly inscrutable error message. So i conclude that no existing UFO file has a <unicode> element without a `hex` attribute.

The fact that the UFO spec does not give a default for this attribute was one of the things that tipped my scales towards me thinking that the attribute should be required when the element is present. And therefore making the code more explicit about checking, instead of failing more or less by accident.

I don't really see a use case for a <unicode> element without an attribute, and presumably all existing tools (except mine) never output such a thing.

benkiel:
`hex` is a required attribute in the UFO spec. If something has a defined default, it is [optional unless stated otherwise](https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#specification). `hex` neither has a defined default nor is it called out explicitly as being optional; so it is required.

simoncozens:
> hex is a required attribute in the UFO spec. 

No, it isn't, as the spec is currently written.

> hex does not have a defined default, nor is it explicitly called out as optional, 

Right.

> so it is required.

This is [denying the antecedent](https://en.wikipedia.org/wiki/Denying_the_antecedent). ("All cats are animals; my rabbit is not a cat; therefore my rabbit is not an animal.") The spec does say that attributes with a defined default are optional but is *silent* as to whether or not attributes *without* a defined default are required - which is why you have offered to add that clarification. I think it's a helpful clarification. I know I'm being picky, but this is the nature of specs; they need to be clear and explicit, to avoid these kinds of assumptions which can differ between interpretations.

drj11:
The spec is one thing, this patch another.

Currently the `Exception` that this patch raises has the message "A unicode element is missing its required hex attribute."; i think that's ok, even if the requirement is by common understanding, not by legal application of the spec rules. We could of course change it to "A unicode element is missing its hex attribute, which is required by common understanding", and then later—after the spec is changed—to "A unicode element its missing its hex attribute, which is required by the UFO specification".

But honestly i don't think that's an improvement.

Lib/fontTools/ufoLib/glifLib.py
```diff
@@ -1191,6 +1191,10 @@ def _readGlyphFromTreeFormat1(
             haveSeenAdvance = True
             _readAdvance(glyphObject, element)
         elif element.tag == "unicode":
+            if element.get("hex") is None:
```
justvanrossum, line 1194: While we're nitpicking:

suggest to take `v = element.get("hex")` out of the try/except and place it here, then test for `if v is None: ...`. This avoids calling `.get()` twice, and now that we've checked it, the second `.get()` should otherwise become `element["hex"]`.

justvanrossum:
> "A unicode element is missing its required hex attribute."

I think that's fine, as that is (obviously, in hindsight) the _intention_ of the spec. Improving the spec would be good.

Lib/fontTools/ufoLib/glifLib.py
drj11: i've done that.

justvanrossum (approved):

drj11:
is this okay to merge now? I'm happy to rebase to a more recent commit if you want.

justvanrossum (approved):

PR #3566, opened 2024-06-11
[otlLib.builder._addName] raise exception if nameID does not exist

m4rc1e:
`buildStatTable` in otlLib builds a STAT table from a python dictionary. AxisValue names can either be a string or an int. If an int is used, it'll simply find a name table name record with the same nameID as the int. This pr will raise an exception if a user tries to get a name record that doesn't exist.

I uncovered this issue while working on Merriweather. The family's opsz axis has the following yaml:

```YAML
  - name: Weight
    ...
  - name: Optical Size
    tag: opsz
    values:
    - name: 12
      value: 12
    - name: 14
      value: 14
    - name: 16
      value: 16
    - name: 24
      value: 24
    - name: 48
      value: 48
    - name: 64
      value: 64
    - name: 72
      value: 72
    - name: 144
      value: 144
```

if i ttx the table, I get:

```XML
 <AxisValue index="13" Format="1">
        <AxisIndex value="2"/>
        <Flags value="0"/>
        <ValueNameID value="24"/>  <!-- missing from name table -->
        <Value value="24.0"/>
      </AxisValue>
      <AxisValue index="14" Format="1">
        <AxisIndex value="2"/>
        <Flags value="0"/>
        <ValueNameID value="48"/>  <!-- missing from name table -->
        <Value value="48.0"/>
      </AxisValue>
      <AxisValue index="15" Format="1">
        <AxisIndex value="2"/>
        <Flags value="0"/>
        <ValueNameID value="64"/>  <!-- missing from name table -->
        <Value value="64.0"/>
      </AxisValue>
      <AxisValue index="16" Format="1">
        <AxisIndex value="2"/>
        <Flags value="0"/>
        <ValueNameID value="72"/>  <!-- missing from name table -->
        <Value value="72.0"/>
      </AxisValue>
      <AxisValue index="17" Format="1">
        <AxisIndex value="2"/>
        <Flags value="0"/>
        <ValueNameID value="144"/>  <!-- missing from name table -->
        <Value value="144.0"/>
```

What the author actually wanted is for the names to be strings such as "12pt", "144pt" etc.

Happy to fix tests if current approach is accepted.

### DIFF
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -2881,8 +2881,10 @@ def _buildAxisValuesFormat4(locations, axes, ttFont, windowsNames=True, macNames
 
 def _addName(ttFont, value, minNameID=0, windows=True, mac=True):
     nameTable = ttFont["name"]
+    # Already a nameID
     if isinstance(value, int):
-        # Already a nameID
+        if not any(nr.nameID == value for nr in nameTable.names):
+            raise ValueError(f"NameID {value} not found in name table")
         return value
     if isinstance(value, str):
         names = dict(en=value)


anthrotype:
> What the author actually wanted is for the names to be strings such as "12pt", "144pt" etc.

then, should they actually have written them as string literals (e.g. within double quotes) instead of as integers? Or can't yaml distinguish those?

Note that I'm not opposed to have a check for existing nameIDs which could be useful in its own right (even though, one may wish to first build a STAT and then a name table that matches those nameIDs, which this PR would prevent); it's more like this is not a direct fix for your problem, which was that the caller were holding the function wrong, so to speak: i.e. passing in integers (which are assumed to be nameIDs) instead of strings (for which new name records are created).

anthrotype:
(clicked the wrong button sorry)

m4rc1e:
> then, should they actually have written them as string literals (e.g. within double quotes) instead of as integers? Or can't yaml distinguish those?

The designer should've written them in double quotes.

> one may wish to first build a STAT and then a name table that matches those nameIDs, which this PR would prevent

Good point!

I don't mind closing this PR and doing the validation in gftools gen-stat instead.

m4rc1e:
Going to close this and do the check in gftools instead. Thanks Cosimo!

Issue #2605, opened 2022-04-28
glyf table `__setitem__` append to glyphOrder but doesn't invalidate _reverseGlyphOrderDict


anthrotype:
The `table__g_l_y_f` class holds a mutable reference to the `TTFont.glyphOrder` list, in its `decompile` method it calls `ttFont.getGlyphOrder()` like this:

https://github.com/fonttools/fonttools/blob/41bd3a143d3182ccb5604b7b01d1cf088da75e94/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L92

Then its `__setitem__` method appends the new glyph name to self.glyphOrder, which at the same time automatically updates the global TTFont.glyphOrder (they are the same list object):

https://github.com/fonttools/fonttools/blob/41bd3a143d3182ccb5604b7b01d1cf088da75e94/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L299-L302

However, if one calls `ttFont.getGlyphID` or `ttFont.getReverseGlyphMap` after setting a new glyph in the glyf table, one gets an incorrect result because the `_reverseGlyphOrderDict` cached value is not invalidated and `getReverseGlyphMap` returns the one computed before setting the new glyph.

```python
from fontTools.ttLib import TTFont, newTable
font = TTFont()
font.setGlyphOrder([".notdef"])
glyphMap = font.getReverseGlyphMap()  # this caches _reverseGlyphOrderDict

glyf = font["glyf"] = newTable("glyf")
glyf.glyphs = {".notdef": Glyph()}
glyf.glyphOrder = font.getGlyphOrder()

# now we add a new glyph via __setitem__
glyf["a"] = Glyph()

print(font.getGlyphOrder())  # => [".notdef", "a"]

print(font.getReverseGlyphMap())  # => {".notdef": 0}   ... whereas it should have been {".notdef": 0, "a": 1} 

print(font.getGlyphID("a"))  # => None ... instead of 1! 
```

This is a but problematic. On the one hand, glyf's `__setitem__` modifes the TTFont's own global `glyphOrder` (indirectly via mutable ref) without invalidating the reverse glyph map cache. On the other hand, it doesn't have a ref to the parent TTFont to be able to invalidate the said cache..

[...]

PR #3582, opened 2024-07-04
[ttFont] Automatically rebuild reverse glyph map if needed

behdad:
Fixes https://github.com/fonttools/fonttools/issues/2605

"Regression" from ca06c9530c6a196ad5b2570afc1f19734fb756bc


Lib/fontTools/ttLib/ttFont.py
```diff
@@ -666,7 +666,12 @@ def getGlyphNameMany(self, lst):
     def getGlyphID(self, glyphName):
         """Returns the ID of the glyph with the given name."""
         try:
-            return self.getReverseGlyphMap()[glyphName]
+            gid = self.getReverseGlyphMap()[glyphName]
+            if self.getGlyphName(gid) == glyphName:
+                return gid
+            self._buildReverseGlyphOrderDict()
```
xadxura, line 672: @behdad — thank you for the quick turn around. With this change as is, it still fails for me, but if I move 672 before 669 then it works fine, i.e., rebuild first.

behdad:
Thanks Andrew. I made another change, can you please test again?

xadxura:
> Thanks Andrew. I made another change, can you please test again?

Thanks, Behdad, I tested again — the code runs without error, but my code's output fails because no glyph ID was returned for cases where I appended new glyphs to the font. These return None but I would expect to get back the gid for the added glyphs. It seems the gids for the new glyphs can only be obtained by calling _buildReverseGlyphOrderDict(). The new try/except test doesn't cover this case. I found I can call _buildReverseGlyphOrderDict() preemptively from my code and this works, so that may be the way forward.

Lib/fontTools/ttLib/ttFont.py
anthrotype: fails how exactly? have you got a test case that reproduces this?
also it isn't clear to me what you mean by 'move 672 before 669'

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -666,7 +666,17 @@ def getGlyphNameMany(self, lst):
     def getGlyphID(self, glyphName):
         """Returns the ID of the glyph with the given name."""
         try:
-            return self.getReverseGlyphMap()[glyphName]
+            gid = None
+            try:
+                gid = self.getReverseGlyphMap()[glyphName]
```
anthrotype, line 671: I think you just return directly here, in the happy case, and not bother with the additional `self.getGlyphName(gid) == glyphName`. The glyf `__setitem__` only appends at the end of the glyphOrder when it's a _new_ glyphName, otherwise it overwrites an existing glyph in place and keeps the current glyphOrder, nothing ever 'moves' an existing named glyph. And glyf `__setitem__` is I think the only place where a TTFont.glyphOrder list is ever modified (or rather allowed to modify), apart from the TTFont.setGlyphOrder() itself. Supporting changing the glyph index of existing named glyphs is trickier than appending at the end because many glyph-related tables need modification, and I think we should not support this use case in the getGlyphID API. If one intentionally wishes to reshuffle glyphs, they should explicitly call font.setGlyphOrder() and rebuild or invalidate the cache (e.g. font.getReverseGlyphMap(rebuild=True) or deleting `_reversedGlyphOrderDict` attribute). Quicker/easier to assume that if a glyphName is present in the current `_reverseGlyphOrderDict` cache, then the cached index is the good one; otherwise if absent (i.e. the KeyError case) go on to rebuild the cache.

anthrotype: > glyf __setitem__ is I think the only place where a TTFont.glyphOrder list is ever modified (or rather allowed to modify)

oh.. actually that's not true unfortunately. In addition to `__setitem__`, the glyf table object also has a `__delitem__` which removes from the glyphOrder without updating the TTFont's `_reversedGlyphOrderDict`, basically screwing the glyph index of all the glyphs that follows the removed glyph... :(

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -666,7 +666,17 @@ def getGlyphNameMany(self, lst):
...
+            except KeyError:
+                if len(self.getGlyphOrder()) == len(self._reverseGlyphOrderDict):
+                    raise
+            if gid is not None and self.getGlyphName(gid) == glyphName:
+                return gid
+            self._buildReverseGlyphOrderDict()
+            return self._reverseGlyphOrderDict[glyphName]
+
         except KeyError:
             if glyphName[:5] == "glyph":
```
anthrotype, line 681: wouldn't this patch now slow down the "virtual GID" code path? Glyphs named e.g. 'glyph00123' would initially raise KeyError because they are absent by definition (hence 'virtual') and thus trigger a rebuild of the glyph map cache each single time, and only after the second KeyError their name's 5-character tail would be turned into an integer.

Lib/fontTools/ttLib/ttFont.py
anthrotype: @xadxura I just saw your other comment now

> no glyph ID was returned for cases where I appended new glyphs to the font. These return None

I think the latest @behdad's patch does not return a None gid any more. Are you sure you tested with the latest latest commit? Try to git fetch origin && git reset --hard origin/getGlyphID (where `origin` in my case is the upstream remote)

behdad:
I've lost track of all the requirements. @anthrotype can you take this over by any chance?

PR #3585, opened 2024-07-15
Parallelize gvar generation

simoncozens:
Generating gvar variations is embarrassingly parallel, and `concurrent.futures` is right there. We now require Python >3.8, so this shouldn't be an issue.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -19,6 +19,7 @@ Then you can make a variable-font this way:
 API *will* change in near future.
 """
 
+from concurrent.futures import ThreadPoolExecutor
 from typing import List
 from fontTools.misc.vector import Vector
 from fontTools.misc.roundTools import noRound, otRound
@@ -308,7 +309,7 @@ def _add_gvar(font, masterModel, master_ttfs, tolerance=0.5, optimize=True):
         for m in master_ttfs
     ]
 
-    for glyph in font.getGlyphOrder():
+    def compute_a_variation(glyph):
         log.debug("building gvar for glyph '%s'", glyph)
         isComposite = glyf[glyph].isComposite()
 
@@ -332,11 +333,10 @@ def _add_gvar(font, masterModel, master_ttfs, tolerance=0.5, optimize=True):
         control = allControls[0]
         if not models.allEqual(allControls):
             log.warning("glyph %s has incompatible masters; skipping" % glyph)
-            continue
+            return glyph, []
         del allControls
 
         # Update gvar
-        gvar.variations[glyph] = []
         deltas = model.getDeltas(
             allCoords, round=partial(GlyphCoordinates.__round__, round=round)
         )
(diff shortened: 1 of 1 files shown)


justvanrossum:
Would be cool to try with the Python 3.13 free threading beta.

simoncozens:
OK, I...

1) Installed a free-threaded 3.13 using pyenv (`PYTHON_CONFIGURE_OPTS='--disable-gil' pyenv install 3.13-dev`)
2) Upgraded to git HEAD Cython
3) Tweaked `openstep-plist` to remove a `long` on line 191 of `writer.pyx`
4) Built lxml, pyclipper, compreffor, uharfbuzz from source
5) Installed fontmake.
6) Let's go.

Compiling Noto Sans with stock fontTools: 24 seconds
Compiling Noto Sans with this patch: 20 seconds (-16%)

Compiling a large private font (8 master, 7000 glyph) with stock fontTools: 68s
Compiling a large private font (8 master, 7000 glyph) with this patch: 57s (-16%)

Compiling Roboto Flex with stock fontTools: 32s
Compiling Roboto Flex with this patch: 25s (-20%)

So this will be useful, just in the future...

justvanrossum:
Thanks! It is great to know that (with some effort), fontmake can run at all with free-threaded Python. This will be a game changer for Python.

jenskutilek:
Would multiprocessing help in this case? I've used it in a couple of places, like to speed up generating a bunch of woffs in parallel. It was very effective.

simoncozens:
No.

simoncozens:
Sorry, that was a bit curt because I was tired and Brandolini's Law applies (generating ideas is easy, explanations take work). Here's something I wrote in another context which explains a bit more.

> ....parallelisation inside Python has historically been a horrible mess, because you either use (a) multi-threading, which to a first approximation *doesn't work* because Python has a Global Interpreter Lock so only one thread can do any work at once (except waiting on IO), or (b) multi-processing, which *often* doesn't work (and usually doesn't help) because you have to create another little Python process, bundle up all the data and code you want it to handle, then unbundle all the results when it's finished. Not all code can be sent across in this way, it has to copy every bit of state that it needs, and often the copying/bundling/unbundling undoes any speed benefit. So currently multithreading is only useful for jobs which do a mix of computation and disk IO, and multiprocessing is only useful for easily separable (not much code/data to copy) but also long-running jobs, where the overhead of copying/bundling/unbundling the data is minimal compared to the time taken by the actual work. Neither are useful for doing a small task in parallel.
> With Python 3.13, you can turn off the Global Interpreter Lock and truly run different code in different threads at the same time. The best use of this is to spread easily-separable computation jobs across different CPU cores. So generating a gvar table is a fantastic use of this - the only state you need is the glyph and the variation model, and each gvar table entry can be generated independently. I think there are probably other potential wins in fontmake as well, such as the fact that we compile each master into a TTF before merging it into a VF - that is a separable job which doesn't depend on anything outside of the master, so in theory we should be able to compile n(=number of cores) masters at the same time.

simoncozens:
So here's an interesting question: if this currently *does nothing* in a backwards compatible way, but will magically give a speedup once free-threading lands... should we merge it?

benkiel:
I do not see any harm in doing so, other than needing to add a test environment for it to the CI (not sure how complicated that would be, having not looked into it)

PR #3588, opened 2024-07-18
Minor fixes to documentation for merge

drj11:
Mostly fixing the explanation of how glyphs get renamed (they get `.n` added, not `#n`), but couldn't resist a couple of other minor fixups too.

### DIFF
--- a/Lib/fontTools/merge/__init__.py
+++ b/Lib/fontTools/merge/__init__.py
@@ -27,7 +27,7 @@ class Merger(object):
 
     This class merges multiple files into a single OpenType font, taking into
     account complexities such as OpenType layout (``GSUB``/``GPOS``) tables and
-    cross-font metrics (e.g. ``hhea.ascent`` is set to the maximum value across
+    cross-font metrics (for example ``hhea.ascent`` is set to the maximum value across
     all the fonts).
 
     If multiple glyphs map to the same Unicode value, and the glyphs are considered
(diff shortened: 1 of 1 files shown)


behdad:
> By the way, i'm not sure that the sentence "If duplicate glyph disambiguation takes place as described above then the fonts must have a GSUB table." is true (any more?). For example the following works and the fontfile does not have a GSUB table.

We do the disambiguation only if the two glyphs have different shapes. I'll try to update the documentation.

behdad:
> > By the way, i'm not sure that the sentence "If duplicate glyph disambiguation takes place as described above then the fonts must have a GSUB table." is true (any more?). For example the following works and the fontfile does not have a GSUB table.
> 
> We do the disambiguation only if the two glyphs have different shapes. I'll try to update the documentation.

Looks like the docu already mentions that.

drj11:
Oh yes, i did actually read that. It doesn't work, or there is something more subtle going on.

The merged TTF has double the number of glyphs (TTF is from https://github.com/google/fonts/tree/main/apache/craftygirls by the way)

```
; otfinfo --glyphs CraftyGirls-Regular.ttf | wc
     230     230    1604
; otfinfo --glyphs merged.ttf | wc
     460     460    3668
```

behdad:
The merge tool is very incomplete and unmaintained unfortunately...

I suspect we add the duplicate glyph, just never use it...

drj11:
Fair enough. i was poking around because i was thinking of writing a merging tool. :)

behdad:
All yours to take. As we say: "you touch it you own it". :D

Issue #3503, opened 2024-05-08
`varLib.interpolatable` suggests fixing functioning interpolations


ddaanniiieeelll:
Hello,

I recently tested a bunch of VF’s and I got some curios test reports.

As you can see in the attached screenshot it suggests moving the startpoint of a contour to a different place. Interestingly the midway interpolation used to look good and the suggested fix would turn the glyphs into spaghetti.

In my opinion the startpoint is in an unusual position anyway, but I don’t think that should make a difference for the test?

![Screenshot 2024-05-08 at 10 02 02](https://github.com/fonttools/fonttools/assets/9609886/0e3867f4-d5e5-4e51-9395-706ccf0e2ff5)

ddaanniiieeelll:
Info:
This was with fonttools 4.51.0
Downgrading to 4.45.0 results in a report without those "false" errors.

behdad:
To be fair, the suggestion looks cool! 

Thanks for using the tool. Can you email me the font? behdad at behdad dot org.  Thanks

ddaanniiieeelll:
Sent you an email with the font and the reports.

I also made a new report with version 4.53.0, but the issue persists.

behdad:
I understand what's happening here: a *reverse* isomorphism is deemed a better match.

I tried to only match isomorphisms of the same winding direction, which should fix this. However, I'm approximating winding direction by the sign of the area of the contour, which fails in varfont designs when the sign of the area of the contour legitimately changes across masters (because of the corner triangles).

So I probably need to (finally) find the true winding of a contour. I'm hoping that it can be done without ray casting. Let me think.

Lorp:
What do you think of this “Contour direction finder”?

https://codepen.io/lorp/pen/OJeWpJG

Font editors should display white vs black vs figure-of-8 while editing IMO :)

Lorp:
Now with outline editing :)

https://github.com/user-attachments/assets/c44f06f8-78ac-4645-b189-c6670e94b747

behdad:
> I tried to only match isomorphisms of the same winding direction, which should fix this. However, I'm approximating winding direction by the sign of the area of the contour, which fails in varfont designs when the sign of the area of the contour legitimately changes across masters (because of the corner triangles).

Picture showing how the area sign can change while the contour direction is the same:
![image](https://github.com/user-attachments/assets/5f095d33-55d4-4e2f-be40-35d04cd99865)

behdad:
[...]

PR #3594, opened 2024-07-29
[interpolatable] In starting-point test, consider contour sign as well

behdad:
Only try matching to an isomorphic rotation if the area signs match

Fixes https://github.com/fonttools/fonttools/issues/3503

Still testing it more broadly.


behdad:
This approach is broken. See https://github.com/fonttools/fonttools/issues/3503#issuecomment-2256773142

PR #3598, opened 2024-08-02
[varLib.avar] Reconstruct mappings from binary

behdad:
https://github.com/Lorp/fencer/issues/25

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -869,7 +869,7 @@ def _add_COLR(font, model, master_fonts, axisTags, colr_layer_reuse=True):
         colr.VarIndexMap = builder.buildDeltaSetIndexMap(varIdxes)
 
 
-def load_designspace(designspace, log_enabled=True):
+def load_designspace(designspace, log_enabled=True, *, require_sources=True):
     # TODO: remove this and always assume 'designspace' is a DesignSpaceDocument,
     # never a file path, as that's already handled by caller
     if hasattr(designspace, "sources"):  # Assume a DesignspaceDocument
@@ -878,7 +878,7 @@ def load_designspace(designspace, log_enabled=True):
         ds = DesignSpaceDocument.fromfile(designspace)
 
     masters = ds.sources
-    if not masters:
+    if require_sources and not masters:
         raise VarLibValidationError("Designspace must have at least one source.")
     instances = ds.instances
 
@@ -978,7 +978,7 @@ def load_designspace(designspace, log_enabled=True):
                     "More than one base master found in Designspace."
                 )
             base_idx = i
-    if base_idx is None:
+    if require_sources and base_idx is None:
         raise VarLibValidationError(
             "Base master not found; no master at default location?"
         )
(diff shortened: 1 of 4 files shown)


behdad:
Just wrote this to @Lorp:

So here's the thing:
lemme see how to explain it.
Everything in normalized numbers.
Imagine a master at wght=1. driving stem=1
And wdth=1 also driving stem=1
And then a middle master at wght=.5,wdth=.5 with start/end's at 0,1 on both axes.
We generate a corner for wght=1,wdth=1
Then we sample the varStore at that point, for stem axis. We would get 2.
Then we clamp it to 1 and that changes the geometry.
I think we should NOT clamp.
And then when we drop "unnecessary" poles, the problem will probably go away.
That's my current thinking.

behdad:
Run the tool like:
```
fonttools varLib.avar RobotoFlex-avar2.ttf
```
It will write the avar1 segments and avar2 mappings to console.

behdad:
The tricky part: removing unnecessary pins, is untested by me as I don't have a font for it right now. But otherwise I think this is ready for review.

behdad:
With tests now for the location-pruning. My confidence is rather high now.

behdad:
Next step, I suppose, is to print out .designspace-like output.

behdad:
Here's it in action. With Raqq's build, the mappings:
https://github.com/aliftype/raqq/blob/8d8ca9e4e41ac660113204fd82fc9029a06335a8/scripts/build.py#L876-L885
```
mappings = [
        [{"Justification": -100}, {"Spacing": -100, "Mashq": 0}],
        [{"Justification": -50}, {"Spacing": 0, "Mashq": 0}],
        [{"Justification": 0}, {"Spacing": 0, "Mashq": 10}],
        [{"Justification": 90}, {"Spacing": 0, "Mashq": 100}],
        [{"Justification": 100}, {"Spacing": 125, "Mashq": 100}],
    ]
```
the tool output:
```
$ ./fonttools varLib.avar Raqq.ttf 
{}
[({'jstf': -50.0}, {'MSHQ': 0.0}),
 ({'jstf': -100.0}, {'MSHQ': 0.0, 'SPAC': -100.0}),
 ({'jstf': 90.00244140625}, {'MSHQ': 100.0}),
 ({'jstf': 100.0}, {'MSHQ': 100.0, 'SPAC': 125.0})]
4 mappings
```

behdad:
I added a roundtrip test, and it's failing, because:

https://github.com/fonttools/fonttools/pull/3601

Or we can change the tests.

behdad:
With https://github.com/fonttools/fonttools/pull/3601 fixed, this is now ready.

@Lorp how does it look to you?
@justvanrossum @anthrotype Can you take a look possibly? Thanks.

behdad:
I'll make it print .designspace snippet instead.

behdad:
> I'll make it print .designspace snippet instead.

Done. At some point we'll add `--output-file` etc. But I think this can be merged now.

behdad:
Since this just adds new functionality to an obscure module (`varLib.avar`), I'm leaning towards merging it if no one wants to review. Thanks.

anthrotype (approved):

Issue #3597, opened 2024-08-02
[varLib.interpolatable] Does not correctly handle discrete axes


colinmford:
Hi @behdad @anthrotype,

I love the `varLib.interpolatable` tool, and I would love to use it on a designspace with a discrete axis. But, when I try to run one through, it gives the following error:

```
  File "/Users/colinmford/Git/fonttools/Lib/fontTools/varLib/interpolatable.py", line 718, in main
    axis_triples = {
                   ^
  File "/Users/colinmford/Git/fonttools/Lib/fontTools/varLib/interpolatable.py", line 719, in <dictcomp>
    a.name: (a.minimum, a.default, a.maximum) for a in designspace.axes
             ^^^^^^^^^
AttributeError: 'DiscreteAxisDescriptor' object has no attribute 'minimum'
```

Larger context:
https://github.com/fonttools/fonttools/blob/ba20aa2f56c90da06000aac7110a99ccfacb8239/Lib/fontTools/varLib/interpolatable.py#L710-L723

I suppose it should split up the designspace into its subspaces, (probably using `designspaceLib.split.splitInterpolatable`) and then run the interpolatable tests on each subspace, though I can see this being a can of worms.

I can attempt a PR to enable discrete axes with `interpolatable`, but I wanted to check first with the both of you to see that something wasn't already in progress, or if you had any suggestions to start.

Thanks!

behdad:
> I can attempt a PR to enable discrete axes with `interpolatable`, but I wanted to check first with the both of you to see that something wasn't already in progress, or if you had any suggestions to start.

It hasn't been on my radar. Let me think about it. I think it can be done without splitting the designspace. Right now I collect all masters, and for each one, pick a closest master to compare with. I just need to have multiple groups of masters.

behdad:
If you don't mind sending me a reproducer, I can work on it right away. You can email me. Thanks.

colinmford:
@behdad I don't have your email, here it is attached:

[Discrete Axis.zip](https://github.com/user-attachments/files/16474547/Discrete.Axis.zip)

behdad:
> @behdad I don't have your email address, here it is attached. It's just a simple 2-axis set up, with one axis discrete.

Thanks. first.last@gmail works...

behdad:
Thanks. This should fix it: https://github.com/fonttools/fonttools/pull/3599

colinmford:
@behdad Thanks so much!

PR #3599, opened 2024-08-03
[varLib.interpolatable] Support discrete axes in .designspace

behdad:
Fixes https://github.com/fonttools/fonttools/issues/3597

Hummm. Only manually tested.

### DIFF
--- a/Lib/fontTools/varLib/interpolatable.py
+++ b/Lib/fontTools/varLib/interpolatable.py
@@ -135,6 +135,7 @@ def test_gen(
     kinkiness=DEFAULT_KINKINESS,
     upem=DEFAULT_UPEM,
     show_all=False,
+    discrete_axes=[],
 ):
     if tolerance >= 10:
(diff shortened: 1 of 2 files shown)


colinmford:
@behdad looks like it works for me!

A minor issue is that when the error is just for one source (like if a glyph is missing), the source name gets "joined" together:
```
Glyph A was not compatible:
  Masters: T, e, s, t, F, a, m, i, l, y, -, S, o, u, r, c, e, _, 1, 0, 0, 0, _, 0:
    Glyph was missing in master TestFamily-Source_1000_0
```

The issue seems to be here:
https://github.com/fonttools/fonttools/blob/64853921508940c3788e215d4aef7907b05648dc/Lib/fontTools/varLib/interpolatable.py#L947-L953

behdad:
> A minor issue is that when the error is just for one source (like if a glyph is missing), the source name string gets "joined" together:

Fixed. Thanks!

PR #3601, opened 2024-08-05
[varLib.models] By default, assume OpenType-like normalized space

behdad:
See:
https://github.com/fonttools/fonttools/pull/2846#issuecomment-2267750076

I *think* this is an improvement, and no one should have been relying on the broken existing behavior.

We do have multiple tests failing because of it. I can go through them and confirm in each case that the new tent is preferred. But I welcome feedback.

Docs need updating.
Tests need fixing.

### DIFF
--- a/Lib/fontTools/varLib/models.py
+++ b/Lib/fontTools/varLib/models.py
@@ -209,10 +209,14 @@ def supportScalar(location, support, ot=True, extrapolate=False, axisRanges=None
 class VariationModel(object):
     """Locations must have the base master at the origin (ie. 0).
 
+    If axis-ranges are not provided, values are assumed to be normalized to
+    the range [-1, 1].
+
     If the extrapolate argument is set to True, then values are extrapolated
     outside the axis range.
 
       >>> from pprint import pprint
+      >>> axisRanges = {'wght': (-180, +180), 'wdth': (-1, +1)}
       >>> locations = [ \
       {'wght':100}, \
       {'wght':-100}, \
@@ -224,7 +228,7 @@ class VariationModel(object):
       {'wght':+180,'wdth':.3}, \
       {'wght':+180}, \
       ]
-      >>> model = VariationModel(locations, axisOrder=['wght'])
+      >>> model = VariationModel(locations, axisOrder=['wght'], axisRanges=axisRanges)
       >>> pprint(model.locations)
       [{},
        {'wght': -100},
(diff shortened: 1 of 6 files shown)


behdad:
For:
```
fonttools varLib.models --locations 0 .5
```
Before:
```
Sorted locations:
[{}, {'A': 0.5}]
Supports:
[{}, {'A': (0, 0.5, 0.5)}]
```
After:
```
Sorted locations:
[{}, {'A': 0.5}]
Supports:
[{}, {'A': (0, 0.5, 1)}]
```

anthrotype:
> no one should have been relying on the broken existing behavior.

could you please elaborate why the existing behavior is broken?

anthrotype:
>  If I have only a non-default master at wght=.5, I expect the tent to be 0,.5,1. But with this change we now get 0,.5,.5. We basically made it a requirement that axis extremes have a master.

well... yeah to me that feels right. I think this is how the VariationModel always worked. You need an additional master at 1.0 if  you don't want the font to snap back to the default between 0.5 and 1.0.
Has it ever worked the way you describe?

behdad:
> > If I have only a non-default master at wght=.5, I expect the tent to be 0,.5,1. But with this change we now get 0,.5,.5. We basically made it a requirement that axis extremes have a master.
> 
> well... yeah to me that feels right. I think this is how the VariationModel always worked. You need an additional master at 1.0 if you don't want the font to snap back to the default between 0.5 and 1.0. Has it ever worked the way you describe?

I don't know. I have to go back in time to check.

That said, I would expect VariationModel output to be a smooth surface. It normally is within the range of on-axis masters. What I'm proposing is to recognize that in the OpenType model, the range is always -1,+1.

Let me test without this bug's change and see.

behdad:
The most relevant commit / issue I can find is this:
https://github.com/googlefonts/fontmake/issues/473

behdad:
FWIW I stumbled on this issue when working on avar2-decompiler.

behdad:
Actually `supportScalar` already has a `ot=True` argument. I wonder if we should also add that to `VariationModel`. Then the main question is: is it safe to change the default...

behdad:
> > no one should have been relying on the broken existing behavior.
> 
> could you please elaborate why the existing behavior is broken?

This is why I find the current behavior surprising to the user, and hence broken:

Let's call a master a pole. Each pole, when added to the system, will receive a tent of influence. These tents normally result in a "bump" in the shape of the manifold (the function surface in multiple dimensions). The exception to this generic rule is that the poles at the extremes of each dimension, have an abrupt falling-off. I agree it has always worked that way.

However, in an OpenType setup, we *know* the range of axes are -1 to +1. So I think instead of abruptly falling, we should consider the -1..+1 cube as the starting point of the manifold and each pole simply makes a bump into it.

Now, as I said, I can write this in a way to convince you is useful to have but not default. But I suggest we make this the new default. As you said, so far we have always said you need a master at the axis extremes. Above, you ask you can't understand how eg. having a Bold but not Black would work...

The problem is, the same VariationModel is also used for cases like building the `avar2` system. There, pretty much anything goes, and since an axis's delta is *added* to its value, it's quite conceivable to have eg. just a delta of 0.1 at axis=0.5 and nothing else. This should result in a tent of 0,0.5,1.

We already consider the value of 0 specially, as does OpenType. What I'm suggesting is to acknowledge the special value of -1 and +1 as well.

behdad:
I talked to @anthrotype and he agreed my proposed model (that VariationModel output is always a smooth surface) is desirable, and it's unlikely that any fonts are relying on the existing model. I've fixed tests and documentation. I'm going to merge.

Issue #3606, opened 2024-08-14
[docs] Sphinx >=8.0.0 breaks RTD theme


n8willis:
The update in #3604 set the `sphinx` dependency in Docs/requirements.txt to `==8.0.2`.

This conflicts with `sphinx_rtd_theme` 2.0, which is set to require sphinx `<8`.

It does look like an update to sphinx_rtd_theme v2.1 is coming (currently in rc2...), but in the meantime, the dependencies are unresolvable because of this, which makes it impossible for people to build local docs.

Downgrading the sphinx requirement to the latest pre-8.0.x works fine.

PR #3611, opened 2024-08-20
Small docs cleanups

n8willis:
These are pretty minor, but need to come before some reorg work on the autodoc configuration.

It does make a change to docs-requirements.txt, as described in #3606, which I think is necessary for the time being. I can only debug the output locally as long as the RTD theme is broken. I think a fix to the theme package is imminent, though.

### DIFF
--- a/Doc/docs-requirements.txt
+++ b/Doc/docs-requirements.txt
@@ -1,4 +1,4 @@
-sphinx==8.0.2
+sphinx==7.4.3
 sphinx_rtd_theme==2.0.0
 reportlab==4.2.2
 freetype-py==2.4.0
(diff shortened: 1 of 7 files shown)


n8willis:
I'm not clear what the linter is unhappy with [here](https://github.com/fonttools/fonttools/actions/runs/10474434543/job/29008694193?pr=3611).

behdad:
> I'm not clear what the linter is unhappy with [here](https://github.com/fonttools/fonttools/actions/runs/10474434543/job/29008694193?pr=3611).

Run `black fonttools/fonttools/Doc/source/conf.py`

behdad:
LGTM but I'm no expert in the docs. Perhaps @anthrotype or @justvanrossum set this up?

anthrotype:
> Perhaps anthrotype or justvanrossum set this up?

I believe @chrissimpkins did

Thanks @n8willis, I may take a look at this later but do not wait for my LGTM

n8willis:
Also, if you're cool with me merging in small stuff, like stuff that doesn't extend outside the `Doc` subdirectory, I can save review-requests for things that alter or add content.....

anthrotype (approved):

PR #3612, opened 2024-08-20
[Tests] Do not require fonttools command to be available

behdad:
I typically run tests like:

$ python setup.py build_ext -i && PYTHONPATH=Lib pytest

Previously, this particular test and only this, required that a `pip install -e .` has had happened. Not anymore.

### DIFF
--- a/Tests/otlLib/optimize_test.py
+++ b/Tests/otlLib/optimize_test.py
@@ -2,7 +2,6 @@ import contextlib
 import logging
 import os
 from pathlib import Path
-from subprocess import run
 from typing import List, Optional, Tuple
 
 import pytest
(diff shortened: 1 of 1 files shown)


behdad:
> not opposed to this kind of simplification (the use of subprocess in this case isn't strictly required), but may I ask why in general do you prefer to _not_ do `pip install -e .` while developing?

It's just that from the C background I come from, I'm used to developing in-tree and leaving the system-installed version intact.

anthrotype:
> leaving the system-installed version intact.

yeah, I never mean to say pip install -e . globally, I meant from within a venv. Anyway LGTM

anthrotype:
> developing in-tree

that only works for simple self-contained python packages that have no dependencies. Even the test runner you had to pip install beforehand :)

behdad:
Thanks. :)

Issue #3616, opened 2024-08-27
[subset] NameRecordVisitor only visits FirstParamUILabelNameID, should visit the whole sequence starting at it


anthrotype:
when we collect the nameIDs used in cvXX features, the NameRecordVisitor is simply visiting FirstParamUILabelNameID, but this is only the first in a range of name records:

https://github.com/fonttools/fonttools/blob/770917d89e9151023cce011c7f7223f4ce84169e/Lib/fontTools/subset/__init__.py#L2995-L2997

https://github.com/fonttools/fonttools/blob/770917d89e9151023cce011c7f7223f4ce84169e/Lib/fontTools/ttLib/tables/_n_a_m_e.py#L1176-L1197

https://learn.microsoft.com/en-us/typography/opentype/spec/features_ae

> If numNamedParameters is non-zero, then firstParamUiLabelNameId and numNamedParameters specify a sequence of consecutive name IDs in the 'name' table. These are used to provide user-interface strings for individual variants. 

harfbuzz does this:

https://github.com/harfbuzz/harfbuzz/blob/2ae90ed3274a67720de1c074115b428a4b098ef8/src/hb-ot-layout-common.hh#L645-L651

thanks @qxliu76 for the heads up

PR #3617, opened 2024-08-27
NameRecordVisitor: include whole sequence of character variants' UI labels, not just the first

### LINKED ISSUE #3516: Bloated subset when using --retain-gids
Hello team.

I have a font collection file `Noto Sans CJK` which has a size of ~18MB

When I subset the font using the following comand. `pyftsubset notosanscjksc.ttf --unicodes="0-FF" --output-file="notosanscjksc-2.ttf"  --font-number=2` , it creates a ubset of size 24KB.

But when I use the arg `--retain-gids` as follows `pyftsubset notosanscjksc.ttf --unicodes="0-FF" --output-file="notosanscjksc-1.ttf"  --font-number=2 --retain-gids`, it creates a subset of size 762KB.

I have attached all 3 files here. Can you please help us understand why this flag causes the subset to be so bloated, and is there a way we can reduce the size with this flag?
[Documents.zip](https://github.com/fonttools/fonttools/files/15393489/Documents.zip)

Thanks.

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/3616

### DIFF
--- a/Lib/fontTools/ttLib/tables/_n_a_m_e.py
+++ b/Lib/fontTools/ttLib/tables/_n_a_m_e.py
@@ -1196,9 +1196,12 @@ def visit(visitor, obj):
     # also include the sequence of UI strings for individual variants, if any
     if obj.FirstParamUILabelNameID == 0 or obj.NumNamedParameters == 0:
         return
-    last_name_id = obj.FirstParamUILabelNameID + obj.NumNamedParameters - 1
-    if last_name_id >= 256 and last_namd_id <= 0x7FFF:
-        visitor.seen.update(range(obj.FirstParamUILabelNameID, last_name_id + 1))
+    visitor.seen.update(
+        range(
+            obj.FirstParamUILabelNameID,
+            obj.FirstParamUILabelNameID + obj.NumNamedParameters,
+        )
+    )
 
 
 @NameRecordVisitor.register(ttLib.getTableClass("fvar"))
--- a/Lib/fontTools/ttLib/tables/_n_a_m_e.py
+++ b/Lib/fontTools/ttLib/tables/_n_a_m_e.py
@@ -1177,15 +1177,6 @@ class NameRecordVisitor(TTVisitor):
     (
         (otTables.FeatureParamsSize, ("SubfamilyID", "SubfamilyNameID")),
         (otTables.FeatureParamsStylisticSet, ("UINameID",)),
-        (
-            otTables.FeatureParamsCharacterVariants,
-            (
-                "FeatUILabelNameID",
-                "FeatUITooltipTextNameID",
-                "SampleTextNameID",
-                "FirstParamUILabelNameID",
-            ),
-        ),
         (otTables.STAT, ("ElidedFallbackNameID",)),
         (otTables.AxisRecord, ("AxisNameID",)),
         (otTables.AxisValue, ("ValueNameID",)),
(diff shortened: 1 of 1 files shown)


behdad:
Lgtm.

anthrotype:
@behdad thanks. I also added [f51a874](https://github.com/fonttools/fonttools/pull/3617/commits/f51a874611e2d067a7b9cbd62a8c0972870c5590)

thanks to Ginger for spotting

behdad:
Ah, thanks.

Lib/fontTools/ttLib/tables/_n_a_m_e.py
```diff
@@ -1197,6 +1188,19 @@ def visit(visitor, obj, attr, value):
...
+def visit(visitor, obj):
+    for attr in ("FeatUILabelNameID", "FeatUITooltipTextNameID", "SampleTextNameID"):
+        value = getattr(obj, attr)
+        visitor.seen.add(value)
+    # also include the sequence of UI strings for individual variants, if any
+    if obj.FirstParamUILabelNameID == 0 or obj.NumNamedParameters == 0:
+        return
+    last_name_id = obj.FirstParamUILabelNameID + obj.NumNamedParameters - 1
+    if last_name_id >= 256 and last_namd_id <= 0x7FFF:
+        visitor.seen.update(range(obj.FirstParamUILabelNameID, last_name_id + 1))
```
khaledhosny, line 1201: Is this check necessary, from the [spec](https://learn.microsoft.com/en-us/typography/opentype/spec/features_ae#tag-cv01--cv99) (emphasis mine):
> The values of featUiLabelNameId, featUiTooltipTextNameId and firstParamUiLabelNameId are **expected** to be in the font-specific name ID range (256–32767), though that is **not a requirement** in this Feature Parameters specification. The value of firstParamUiLabelNameId + numNamedParameters - 1 should not exceed 32767.

anthrotype: > Is this check necessary 

which check exactly? `last_name_id >= 256` ?

behdad: I agree with Khaled here. No need to check these things...

khaledhosny: The whole check. Do we check elsewhere that name IDs are less than 0x7FFF before using them?

anthrotype: ok.. I simply copied the code verbatim from hb

anthrotype: and there's even a typo! 🤣 

`last_namd_id <= 0x7FFF`

khaledhosny: We should drop the check from HarfBuzz then.

khaledhosny:
The issue to be fixed by this should be https://github.com/fonttools/fonttools/issues/3616 not https://github.com/fonttools/fonttools/issues/3516

anthrotype:
thank you Khaled for double checking!

Lib/fontTools/ttLib/tables/_n_a_m_e.py
anthrotype: I have removed the check. Ideally we'd have some test font to exercise this with. I know I could make one ad-hoc, but do you happen to know a font that contains these which I simply use?

khaledhosny: No, I don’t have or know such a font.

behdad: @qxliu76 Can you remove this check from HB too? We don't enforce these kinds of things unless necessary.

PR #3618, opened 2024-08-28
add EBSC

### LINKED ISSUE #685: EBSC support missing
The font added to issue #684 contains an `EBSC` table – it's the only sample I know of...

It would be great if ttx could produce a readable version of it.

NightFurySL2001:
Fix #685


behdad:
I'm fine with this in principle. But at least a test would be nice.

PR #3621, opened 2024-08-29
allow passing in TTFont, change to use argparse

NightFurySL2001:
Implements #3614


behdad:
I have a vague memory that the merger doesn't take TTFonts, because it has to load fonts multiple times. But I'm probably wrong. I don't have time to review this currently.

NightFurySL2001:
From what I saw  the fonts are loaded multiple times to build a new GlyphOrder. Tried to sidestep it by using `deepcopy` (although I believe it's not required at all).

behdad:
> From what I saw the fonts are loaded multiple times to build a new GlyphOrder. Tried to sidestep it by using `deepcopy` (although I believe it's not required at all).

I don't think that would work. Se want to reload the fonts using the *new* GlyphOrder.

NightFurySL2001:
Pretty sure the reload happens *before* setting the new GlyphOrder. No other files call for `_openFonts()` in merge module.
https://github.com/fonttools/fonttools/blob/770917d89e9151023cce011c7f7223f4ce84169e/Lib/fontTools/merge/__init__.py#L79-L91

behdad:
The font's GlyphOrder is loaded lazily. After the first `_openFonts`, the GlyphOrder of the font is used, which means `cmap` and some other tables are loaded.

We load the font a second time, and set GlyphOrder on it *before* accessing any tables, so all tables are loaded using the new GlyphOrder...

PR #3622, opened 2024-08-30
subset_test: add failing test to reproduce issue #3616

anthrotype:
If we subset this test font (a subset of Google Fonts' Andika-Regular.ttf) and request to keep 'cv43', only the FirstParaUILabelNameID (324) is currently kept, the other two (325 amd 326) get incorectly dropped. All referenced nameIDs should be kept.

This will be fixed with https://github.com/fonttools/fonttools/pull/3617

```
>       assert nameIDs == keepNameIDs
E       assert {0, 1, 2, 3, 4, 5, 6, 321, 322, 323, 324} == {0, 1, 2, 3, 4, 5, 6, 321, 322, 323, 324, 325, 326}
E         Extra items in the right set:
E         325
E         326
E         Full diff:
E         - {0, 1, 2, 3, 4, 5, 6, 321, 322, 323, 324, 325, 326}
E         ?                                         ----------
E         + {0, 1, 2, 3, 4, 5, 6, 321, 322, 323, 324}
```

### DIFF
--- a/Tests/subset/data/Andika-Regular.subset.ttx
+++ b/Tests/subset/data/Andika-Regular.subset.ttx
@@ -1,0 +1,733 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ttFont sfntVersion="\x00\x01\x00\x00" ttLibVersion="4.53">
+
+  <GlyphOrder>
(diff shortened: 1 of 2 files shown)


anthrotype:
i'll merge this then rebase #3617 on top of `main` to confirm it passes with the fix applied

PR #3625, opened 2024-09-02
Docs: fix Sphinx builds

n8willis:
This changes the Sphnix config to keep the napoleon and autodoc extensions from interfering with each other, and it moves the module docstrings for designspaceLib and ufoLib (in the respective `__init__.py`s) to the start of the file. That's more in line with PEP257, but more practically, autodoc will not see those docstrings unless they are absolutely first.


behdad:
lgtm

PR #3627, opened 2024-09-03
Docs: cleanup code blocks

n8willis:
Fixes several code-block and `Example::` stanzas, plus a couple of bullet lists, in various docstrings.

### DIFF
--- a/Lib/fontTools/pens/freetypePen.py
+++ b/Lib/fontTools/pens/freetypePen.py
@@ -289,7 +289,7 @@ class FreeTypePen(BasePen):
                 >> type(a), a.shape
                 (<class 'numpy.ndarray'>, (1000, 500))
         """
-        
+
         import numpy as np
 
         buf, size = self.buffer(
@@ -333,7 +333,7 @@ class FreeTypePen(BasePen):
         Example:
             .. code-block:: pycon
 
-                >>> 
+                >>>
                 >> pen = FreeTypePen(None)
                 >> glyph.draw(pen)
                 >> pen.show(width=500, height=1000)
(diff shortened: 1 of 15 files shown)


n8willis:
These Fails seem to be caused by the test incorrectly parsing the indented REPL blocks in the docstrings [example](https://github.com/fonttools/fonttools/actions/runs/10705868539/job/29682258348?pr=3627#step:5:3498) as if they are _not_ part of the quoted string....

The quote matching does not look wrong to me (nor to Emacs highlighting or the Pygments parsing if building the docs locally). Am I missing something subtle?

n8willis:
So this is evidently a known problem, without a solution or a great set of options (IMO). Boils down to `doctest` vs Sphinx (which has is own, unrelated `doctest` directive feature, a point on which I was confused....).

Basically, you can't have the `doctest` module, Sphinx's syntax highlighting, and a console/REPL example block that incorporates any dummy/placeholder objects. There are a bunch of those in the code, like "`path/to/my_font.ttf`" or stuff like that.

Option A: syntax highlighting on the example console code that matches the highlighting on all non-interactive code-snippet examples, which causes `doctest` errors:
Screenshot 2024-09-11 at 15-26-27 instancer — fontTools Documentation

Option B: syntax highlighting that looks different for interactive console examples than for everything else, but is at least wrapped in a Sphinx "this is a code example" box and does not trigger `doctest` fails:
Screenshot 2024-09-11 at 15-40-11 instancer — fontTools Documentation

that's via the workaround of prepending a blank-line `>>>` to each block.

Option C would be to lose all the "this is a code example" box wrappers and all highlighting  on interactive console code, which would make `doctest` happy but perhaps be harder to read, and still look different from other sample snippets.

Option D would be to rewrite all the console examples to pass `doctest` testing. I'm not sure that's doable.

n8willis:
(this PR is currently using Option B)

behdad:
I'd go with what you recommend. @justvanrossum

n8willis:
Option B is the smoothest for the time being, for sure.

Long-term, maybe rewriting some of the console examples would be nice, but then again I also might figure out some workarounds within Sphinx itself.

PR #3633, opened 2024-09-10
Small tweaks to pyftsubset documentation

RoelN:
Two small things I noticed in the docs:

1. Remove `./` from example (32617523eebf6723c24bbd0126bbc4f7df181e4d)
2. Add note to escape the `?` from arguments (41854749d7f2eedf860b9ec082aca3565be641ea)

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -128,9 +128,9 @@ Examples::
 
     $ pyftsubset --glyph-names?
     Current setting for 'glyph-names' is: False
-    $ ./pyftsubset --name-IDs=?
+    $ pyftsubset --name-IDs=?
     Current setting for 'name-IDs' is: [0, 1, 2, 3, 4, 5, 6]
-    $ ./pyftsubset --hinting? --no-hinting --hinting?
+    $ pyftsubset --hinting? --no-hinting --hinting?
     Current setting for 'hinting' is: True
     Current setting for 'hinting' is: False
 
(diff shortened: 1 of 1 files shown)


behdad:
Thanks. We probably should promote `fonttools subset` name instead of `pyftsubset`.

n8willis:
> Thanks. We probably should promote `fonttools subset` name instead of `pyftsubset`.

Is this a general principle, that we want to migrate away from specialty / one-off CLI utility names?

Asking because there are some others I'm currently hacking at docs on, like the CFF 1/2 ones....

behdad:
> > Thanks. We probably should promote `fonttools subset` name instead of `pyftsubset`.
> 
> Is this a general principle, that we want to migrate away from specialty / one-off CLI utility names?

Yes, except for ttx.

> Asking because there are some others I'm currently hacking at docs on, like the CFF 1/2 ones....

There shouldn't be. The only ones we have are `ttx`, `pyftsubset`, and `pyftmerge`. Everything else is invoked through the `fonttools moduleName` syntax.

n8willis:
> There shouldn't be. The only ones we have are `ttx`, `pyftsubset`, and `pyftmerge`. Everything else is invoked through the `fonttools moduleName` syntax.

Right right; I was misremembering the CFF ones having only glanced at them. Eyes will stay peeled for others....

Shifting focus away from `pyftsubset` towards `fonttools subset` will affect the top-level info page & the subset module page only, I believe; I'll add it to the to-do list.

PR #3635, opened 2024-09-11
Fix visual artefacts with partial L2 instancing

### LINKED ISSUE #3634: Large visual artefacts when L2 instancing a partial VF with IUP optimisation
To produce inferred deltas that will be correct given [OpenType's `gvar` semantics](https://learn.microsoft.com/en-us/typography/opentype/spec/gvar#inferred-deltas-for-un-referenced-point-numbers), fontTool's IUP optimisation module [checks the equality of some points](https://github.com/fonttools/fonttools/blob/11343ed64c10d3905f4fb4c8d2810c9612ee8acd/Lib/fontTools/varLib/iup.py#L57). However, this happens **before** the points are rounded, whereas the point comparison that happens at runtime will occur **after** the points are rounded (as is necessary to serialise `glyf`), which leads to diverging semantics and so diverging and incorrect implied deltas.

https://github.com/fonttools/fonttools/blob/11343ed64c10d3905f4fb4c8d2810c9612ee8acd/Lib/fontTools/varLib/iup.py#L57-L63

This leads to significant visual artefacts, e.g. where large deltas that should be inferred based on previous values are instead interpreted as 0 at runtime.

I suspect this has gone undetected as the subsetter normally works with rounded points; in the rarer case that partial VF instancing is occurring with a different default position, however, `varLib.instancer` will calculate and [apply the relevant deltas to the font's original coordinates](https://github.com/fonttools/fonttools/blob/11343ed64c10d3905f4fb4c8d2810c9612ee8acd/Lib/fontTools/varLib/instancer/__init__.py#L881) to effect the new default position, which leads to unrounded points in memory.

https://github.com/fonttools/fonttools/blob/11343ed64c10d3905f4fb4c8d2810c9612ee8acd/Lib/fontTools/varLib/instancer/__init__.py#L875-L881

---

I have a patch that rounds earlier and fixes the problem in a test font, and will start a PR this morning :)

Hoolean:
Fixes #3634

The issue above emerges when assumptions are made about point equality based on unrounded points, whereas the comparisons that happen at runtime will be based on rounded points. In particular, unrounded points are present when `varLib.instancer` calculates and applies deltas to the font's original coordinates to effect a new default position, and so this PR rounds immediately after this happens to ensure that the coordinates match those in the final `glyf` table.

---

### To undraft

- [x] Check tests
- [x] Add regression test with minimal reproducer

#### Further

- [x] See if deltas need rounding too?
- [x] See if CFF2 code path is affected?
- [x] See if `iup.py` should instead work with `int`s for point or delta comparison (potentially still using the original `double`s to evaluate tolerance?)
- [x] Confirm comment on unrounded coordinates below

### DIFF
--- a/Lib/fontTools/varLib/instancer/__init__.py
+++ b/Lib/fontTools/varLib/instancer/__init__.py
@@ -897,7 +897,18 @@ def _instantiateGvarGlyph(
         return
 
     if optimize:
+        # IUP semantics depend on point equality, and so round prior to
+        # optimization to ensure that comparisons that happen now will be the
+        # same as those that happen at render time. This is especially needed
+        # when floating point deltas have been applied to the default position.
+        #     See https://github.com/fonttools/fonttools/issues/3634
+        # Rounding must happen only after calculating glyf metrics above, to
+        # preserve backwards compatibility.
+        #     See 0010a3cd9aa25f84a3a6250dafb119743d32aa40
+        coordinates.toInt()
+
         isComposite = glyf[glyphname].isComposite()
+
         for var in tupleVarStore:
             var.optimize(coordinates, endPts, isComposite=isComposite)
 
--- a/Tests/varLib/instancer/data/3634-VF.ttx
+++ b/Tests/varLib/instancer/data/3634-VF.ttx
@@ -1,0 +1,294 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ttFont sfntVersion="\x00\x01\x00\x00" ttLibVersion="4.53">
+
+  <GlyphOrder>
+    <!-- The 'id' attribute is only for humans; it is ignored when parsed. -->
+    <GlyphID id="0" name=".notdef"/>
+    <GlyphID id="1" name="A"/>
+  </GlyphOrder>
+
+  <head>
+    <!-- Most of this table will be recalculated by the compiler -->
+    <tableVersion value="1.0"/>
+    <fontRevision value="0.0"/>
+    <checkSumAdjustment value="0x139cb6b8"/>
+    <magicNumber value="0x5f0f3cf5"/>
+    <flags value="00000000 00000011"/>
+    <unitsPerEm value="1000"/>
+    <created value="Wed Sep 11 11:44:03 2024"/>
+    <modified value="Wed Sep 11 11:44:03 2024"/>
+    <xMin value="0"/>
+    <yMin value="-200"/>
+    <xMax value="1000"/>
+    <yMax value="1000"/>
+    <macStyle value="00000000 00000000"/>
+    <lowestRecPPEM value="6"/>
+    <fontDirectionHint value="2"/>
+    <indexToLocFormat value="0"/>
+    <glyphDataFormat value="0"/>
+  </head>
+
+  <hhea>
+    <tableVersion value="0x00010000"/>
+    <ascent value="1000"/>
+    <descent value="-200"/>
+    <lineGap value="0"/>
+    <advanceWidthMax value="1000"/>
+    <minLeftSideBearing value="0"/>
+    <minRightSideBearing value="0"/>
+    <xMaxExtent value="1000"/>
+    <caretSlopeRise value="1"/>
+    <caretSlopeRun value="0"/>
+    <caretOffset value="0"/>
+    <reserved0 value="0"/>
+    <reserved1 value="0"/>
+    <reserved2 value="0"/>
+    <reserved3 value="0"/>
+    <metricDataFormat value="0"/>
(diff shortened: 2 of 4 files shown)


Hoolean:
Some tests are failing because they expect unrounded coordinates; these were added based on 0010a3cd9aa25f84a3a6250dafb119743d32aa40, and so we'll have to be careful that we preserve the semantics this was intended to preserve.

My loose understanding is that rounding after the deltas are already applied to `glyf` is late enough and will match the implicit rounding that will happen when calculating metrics, but we should confirm.

behdad:
The analysis is spot-on. Lemme digest the proposed solution for any side-effects.

behdad:
> #### Further
> * [ ]  See if deltas need rounding too?

See https://github.com/fonttools/fonttools/pull/2214/commits/68004b8fec94729b7cfe9755eb956fa39feff5de.  Not sure what should happen during instancing.

> * [ ]  See if CFF2 code path is affected?

Good question.

> * [ ]  See if `iup.py` should instead work with `int`s for point or delta comparison (potentially still using the original `double`s to evaluate tolerance?)

Maybe we can add a round parameter there that defaults to noRound to preserve existing behavior. But I think what you have is fine, and letting the tolerance take care of it...

> * [ ]  Confirm comment on unrounded coordinates below

Hoolean:
I've stepped through again with a debugger and it looks like the deltas are always rounded before they reach IUP optimisation, so we should not be at risk of the same rounding issue that affected coordinates:

https://github.com/fonttools/fonttools/blob/11343ed64c10d3905f4fb4c8d2810c9612ee8acd/Lib/fontTools/varLib/instancer/__init__.py#L866-L878

https://github.com/fonttools/fonttools/blob/11343ed64c10d3905f4fb4c8d2810c9612ee8acd/Lib/fontTools/varLib/instancer/__init__.py#L559-L561

I've ticked that point off in the PR description 🚀

behdad:
> * [ ]  See if CFF2 code path is affected?

CFF2 doesn't have IUP.  So we should be good I think.

Hoolean:
> - [ ] Confirm comment on unrounded coordinates below
> > Some tests are failing because they expect unrounded coordinates; these were added based on [0010a3c](https://github.com/fonttools/fonttools/commit/0010a3cd9aa25f84a3a6250dafb119743d32aa40), and so we'll have to be careful that we preserve the semantics this was intended to preserve.
> >
> > My loose understanding is that rounding after the deltas are already applied to `glyf` is late enough and will match the implicit rounding that will happen when calculating metrics, but we should confirm.

I revisited Cosimo's commit (0010a3cd9aa25f84a3a6250dafb119743d32aa40) and `glyf`'s metrics implementation, and now believe that it's important that we pass `glyf` unrounded coordinates. `glyf` uses the rounded diff of unrounded coordinates, which can be different to the rounded diff of rounded coordinates.

https://github.com/fonttools/fonttools/blob/9563608df9eeadef487d4143bed34f796ad15421/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L480-L493

With this in mind, I've made the patch round directly prior to IUP optimisation instead of directly after applying deltas to the default coordinates.

This ensures that all existing uses of the coordinates will continue to use unrounded coordinates, unchanged, with only IUP opimisation receiving the rounded coordinates that it requires. This preserves the side-bearings backwards compatibility with `varLib.mutator` that Cosimo's original commit intends, and so also allowed me to revert the changes we had to make to expected values in existing tests.

(this change was made with a rebase to avoid git history having two directly contradicting rationales)

Hoolean:
> > * [ ]   See if `iup.py` should instead work with `int`s for point or delta comparison (potentially still using the original `double`s to evaluate tolerance?)
> 
> Maybe we can add a round parameter there that defaults to noRound to preserve existing behavior. But I think what you have is fine, and letting the tolerance take care of it...

Having thought on this some more, I think that any invocation of `iup.py` with coordinates or deltas with fractional parts is at risk of being unsound, as the fractional parts can never be represented in the final font in any context, and so coordinate or delta equality would always be at risk of being incorrect after rounding.

L2 instancing is the only place where we have found a real-world issue with this so far though, and so I will tick this off here to avoid holding our fix back with something that might just be hypothetical. I'll make a spin-off issue to cover this though, in case we are calling the `iup.py` with fractional parts anywhere else and it hasn't surfaced an issue yet :)

Hoolean:
Ready for review now!

@behdad Thanks again for the guidance Behdad :)

behdad:
> Ready for review now!
> 
> @behdad Thanks again for the guidance Behdad :)

Thank *you*! Given that @anthrotype is on leave, let's merge this. LGTM.

PR #3641, opened 2024-09-23
Add TTF2OTF snippet

ftCLI:
This pull request adds a snippet to convert TTF fonts to OTF.


behdad:
Thanks for this.

My only feedback is that it would be nice to do it without click.

behdad:
Well, and move it to an actual module, instead of Snippets.

ftCLI:
> Well, and move it to an actual module, instead of Snippets.

In which module should I move it? `qu2cu` seems the natural location.

Ok to rewrite the CLI with argparse. Is it ok to keep the `cffsubr` dependency or should it be removed too?

behdad:
> > Well, and move it to an actual module, instead of Snippets.
> 
> In which module should I move it? `qu2cu` seems the natural location.

I think `ttLib` is best.

> Ok to rewrite the CLI with argparse. Is it ok to keep the `cffsubr` dependency or should it be removed too?

`cffsubr` can stay, as an optional dependency.

ftCLI:
Hi,

I've replaced `click` with `argparse`. Regarding moving the snippet to `ttLib`, I'm not sure. It does what the `Snippets/otf2ttf.py` does, but in the opposite direction.

behdad:
Can you please run `black` on it, to fix the lint test? Looks good to me otherwise.

PR #3645, opened 2024-09-23
Escape `\` in subset docs

mark-thm:
In Python 3.12, it's a syntax error to use invalid escape sequences. `\?` is not a valid escape sequence but is used in a docs variable in `subset/__init__.py`. This PR escapes the backslash.

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -123,7 +123,7 @@ Other options
 
 For the other options listed below, to see the current value of the option,
 pass a value of '?' to it, with or without a '='. In some environments,
-you might need to escape the question mark, like this: '--glyph-names\?'.
+you might need to escape the question mark, like this: '--glyph-names\\?'.
 
 Examples::
 


behdad:
Thanks.

bersbersbers:
While this seems innocent, it has the potential to break CI, e.g., like this:

```
pip install typeguard fontTools==4.54.0
python -W error -c "import typeguard; typeguard.install_import_hook(); import fontTools.subset"
```

(Note that `python -W error -c "import fontTools.subset"` does not seem to trigger the syntax warning.)

I think this maybe warrants a hotfix release.

kkom:
Yes, +1 to a hotfix please! It did break CI for us:

```
.venv/lib/python3.12/site-packages/ddtrace/internal/module.py:309: in _exec_module
    self.loader.exec_module(module)
.venv/lib/python3.12/site-packages/ddtrace/internal/module.py:287: in get_code
    code = _get_code(fullname)
E     File "/app/.venv/lib/python3.12/site-packages/fontTools/subset/__init__.py", line 44
E       + """
E
E   SyntaxError: invalid escape sequence '\?'
```

Might be related to using `ddtrace`, that's why I pasted a stack frames above the actual error.

behdad:
Okay let me clean this up.

PR #3650, opened 2024-09-27
[TupleVariation] Optimize for loading speed, not size

behdad:
Not worth it. Just leaving here for future reference.


behdad:
This makes VARC+gvar fonts slightly larger but slightly faster. Slightly being about 20%. I think it's worth having the code in, since it makes no change to the font format, just how we encode things.

behdad:
Not hooked up anywhere right now, just the code that needs to somehow be enabled.

behdad:
The idea here is that instead of encoding deltas run-length encoded as 0, 1, 2, or 4 bytes each, just use the widest necessary and encode all with that width.

PR #3653, opened 2024-09-30
[TupleVariation] Optimize for loading speed, not size

behdad:
This makes VARC+gvar fonts slightly larger but slightly faster. Slightly being about 20%. I think it's worth having the code in, since it makes no change to the font format, just how we encode things. I'll try to wire it up further.

Originally posted as https://github.com/fonttools/fonttools/pull/3650


behdad:
Thanks. Added some tests and wired it up to gvar table.

PR #3656, opened 2024-10-07
Build, test, and label support for Python 3.13

Hoolean:
[Python 3.13 has a stable release!](https://www.python.org/downloads/release/python-3130/) :tada:

This PR repeats at least some of the steps we took for 3.12:

- eda6fa5cfbdfaf1d54cf391ed9c86b72288882a2
- 9b1dae4765b419778f8721caf9d0a6e05c2758eb

## Next Steps

From limited local testing, we will probably have to bump some pinned versions in `requirements.txt` to successfully build Cython-based dependencies. Alternatively, this could be made unnecessary if they ship binary wheels for the new Python release.

I am not sure how we decide which subset of Python versions to continue to run tests for after a new release, and so for now the PR makes a best guess based on the previous semantics.

<sub>Note: this is a personal contribution independent of my employer, and so I've submitted from a fork under my personal profile and email to make this distinction ✔️</sub>


Hoolean:
The [pyclipper build error](https://github.com/fonttools/fonttools/actions/runs/11224471626/job/31201256633?pr=3656#step:5:352) on `test (3.13, ubuntu-latest)` matches what I saw when installing `requirements.txt` locally.

This compilation error is fixed upstream in cython (cython/cython#6256), and I could not replicate when installing pyclipper directly, and so I suspect a pin may need to be bumped or relaxed somewhere in the file.

Alternatively - and probably preferably, as the project is also under fonttools' umbrella - pyclipper could ship a binary wheel for 3.13 :rocket: It looks like this is in part tracked upstream: fonttools/pyclipper#56

anthrotype:
this now passes. however it's taking longer than needed because it's building from source zopfli, skia-pathops and openstep-plist. I need to also update those to build and upload pre-compiled wheels for 3.13.

anthrotype:
i'll continue from here, thanks

PR #3657, opened 2024-10-09
[glyf] Add optimizeSize option

behdad:
Set to True by default. Can be turned to False on the table.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -713,7 +713,9 @@ class Glyph(object):
         else:
             self.decompileCoordinates(data)
 
-    def compile(self, glyfTable, recalcBBoxes=True, *, boundsDone=None):
+    def compile(
+        self, glyfTable, recalcBBoxes=True, *, boundsDone=None, optimizeSize=None
+    ):
         if hasattr(self, "data"):
             if recalcBBoxes:
                 # must unpack glyph in order to recalculate bounding box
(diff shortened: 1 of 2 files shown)


khaledhosny (approved):
LGTM, but probably needs some tests.

Lorp:
To confirm, this allows the glyph to be written as a u8[numPoints] flags, i16[numPoints] xCoords, i16[numPoints] yCoords?

behdad:
> To confirm, this allows the glyph to be written as a u8[numPoints] flags, i16[numPoints] xCoords, i16[numPoints] yCoords?

Something like that. Might be i8 for xCoords / yCoords permitting... And of course runs are limited to 256 entries at a time.

Lorp:
What runs in glyf? Ah, this is the gvar deltas thing? I assumed you were talking glyf because of "[glyf]" in the subject :) Writing uncompressed glyf is quicker, and 100% beneficial if immediately converting to WOFF2.

behdad:
> What runs in glyf? Ah, this is the gvar deltas thing? I assumed you were talking glyf because of "[glyf]" in the subject :) Writing uncompressed glyf is quicker, and 100% beneficial if immediately converting to WOFF2.

It *is* glyf. The flags repeat goes up to 255 only, so every 256 items need a new flag...

Lorp:
If writing uncompressed, why set the repeat bit at all?

behdad:
> If writing uncompressed, why set the repeat bit at all?

Oh compressing the repeat is faster I think to process, since you can have one loop per repeat, instead of loop iterations on each coordinate.

Lorp:
In Samsa I found quite significant speedups in writing glyf entirely uncompressed, since I know in advance where to write xCoords and yCoords.

behdad:
> In Samsa I found quite significant speedups in writing glyf entirely uncompressed, since I know in advance where to write xCoords and yCoords, and can do so in a single loop.

I saw about 5% speedup decoding the results of this PR. That's without any special coding...

Lorp:
Ah ok, I was concerned with improving encoding speed.

Issue #3658, opened 2024-10-11
ttLib.removeOverlaps shifts glyphs on the left in CFF fonts


ftCLI:
Running the following command on IBMPlexSans-Bold.otf produces a strange effect (the script finds an overlap in the `OE` glyph)

`fonttools ttLib.removeOverlaps IBMPlexSans-Bold.otf IBMPlexSans-Bold_1.otf`

The `OE` glyph in the original font (dumped with `tx -pdf`):

![image](https://github.com/user-attachments/assets/9e746674-3a8a-4af9-8719-996cd513d2e5)

The `OE` glyph in the modified font

![image](https://github.com/user-attachments/assets/e3fa5a29-b2a7-45ac-af75-17e9bc531110)

More: if I remove overlaps from the modified font, the glyph is shifted again

`fonttools ttLib.removeOverlaps IBMPlexSans-Bold_1.otf IBMPlexSans-Bold_2.otf`

![image](https://github.com/user-attachments/assets/17e13da5-111e-4ab5-8f92-d87ea347f505)

It happens with all CFF fonts, but here's the one I used: [IBMPlexSans-Bold.zip](https://github.com/user-attachments/files/17340639/IBMPlexSans-Bold.zip)

I think[1] the issue is in this line: 

https://github.com/fonttools/fonttools/blob/afceebcda515b779faabe0acdec13018bab0c246/Lib/fontTools/ttLib/removeOverlaps.py#L236

[1] I almost went mad for this last year, and the only workaround I found is to [rebuild the whole charstrings dict](https://github.com/ftCLI/FoundryTools-CLI-2/blob/add614d9fbe561885aba68b6350255ed09e8e77b/foundrytools_cli_2/lib/skia/skia_tools.py#L311) and [passing it to `fontBuilder`](https://github.com/ftCLI/FoundryTools-CLI-2/blob/add614d9fbe561885aba68b6350255ed09e8e77b/foundrytools_cli_2/lib/font/font.py#L825)

PR #3659, opened 2024-10-11
[removeOverlaps] Fix CFF CharString width

khaledhosny:
The width argument of `T2CharStringPen()` is inserted directly into the CharString program, so it must be relative to Private.nominalWidthX, but CharString.width is a calculated absolute value.

Some implementations, notably Adobe’s, will use the width from the CFF CharString instead of the one from hmtx table.

- Fixes https://github.com/fonttools/fonttools/issues/3658


behdad:
While there... if width is equal to `defaultWidthX`, pass `None` to the pen and it won't encode it.

khaledhosny:
> While there... if width is equal to `defaultWidthX`, pass `None` to the pen and it won't encode it.

Done.

behdad:
LGTM FWIW.

PR #3667, opened 2024-10-21
Update shebang

luckman212:
Shebang was hardcoded to `/usr/bin/python` which will fail on many modern OSes.

### DIFF
--- a/Lib/fontTools/mtiLib/__init__.py
+++ b/Lib/fontTools/mtiLib/__init__.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 
 # FontDame-to-FontTools for OpenType Layout tables
 #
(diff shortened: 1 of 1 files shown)


anthrotype:
I don't think anybody uses those and we should probably remove them all.
You are not expected to run this (or similar modules) like scripts `./Lib/fontTools/mtiLib/__init__.py`, but with either `python -m fontTools.mtiLib` (where `python` is whatever is the python executable fonttools is importable from) or `fonttools mtiLib` (using the `fonttools` cli helper, which does the same as python -m fontTools.<submodule>).

/cc @behdad

luckman212:
Got it. My linter caught this which is the only reason I flagged it. If it's a non-issue sorry to have bothered.

anthrotype:
No problem and thanks for flagging this

behdad:
Should be removed.

khaledhosny:
I removed the shebang completely.

PR #3669, opened 2024-10-21
Set isfinal to true in XML parser for proper resource cleanup

allrob23:
### Summary

Set `isfinal=1` in `Parse()` calls to ensure proper resource finalization after parsing.

### Reference:

[Python XML Parser Documentation](https://docs.python.org/3/library/pyexpat.html#xml.parsers.expat.xmlparser.Parse)

### DIFF
--- a/Lib/fontTools/misc/testTools.py
+++ b/Lib/fontTools/misc/testTools.py
@@ -38,7 +38,7 @@ def parseXML(xmlSnippet):
             % type(xmlSnippet).__name__
         )
     xml += b"</root>"
-    reader.parser.Parse(xml, 0)
+    reader.parser.Parse(xml, 1)
     return reader.root[2]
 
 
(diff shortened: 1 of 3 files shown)


behdad:
Probably better to spellout the argument, like `Parse(text, isfinal=True)` perhaps?

allrob23:
Thanks for the feedback. I agree with you; I implemented this way for consistency with the implementation here

https://github.com/allrob23/fonttools/blob/19e6a2c18fad7f2b6207dbe63f40ec5a54641fe6/Lib/fontTools/misc/xmlReader.py#L70-L79

What do you think of changing all of them to the spelled-out version?

behdad:
> What do you think of changing all of them to the spelled-out version?

I think that would be nice. Thanks.

behdad:
Test failures seem to be unrelated.

allrob23:
No, I think it is related. I believe we need to revert back the spelled out version.

image

behdad:
Ooh. I was seeing some other errors. You're right, it's failing here as well. I'll revert.

Issue #3666, opened 2024-10-21
`varLib.interpolatable` not supporting CFF2 VFs?


frankrolf:
I’ve been playing around with `varLib.interpolatable` after receiving [a few useful reports](https://github.com/adobe-fonts/source-serif/issues/131) – now, I am experimenting with a different font, and it seems like CFF2 VFs are not supported at all.

I would not be too surprised if I’d get an error message communicating as much – instead, the PDF claims “Your font’s good – have a cupcake”. The same command for the same VF in TTF format will produce a 76-page PDF with kink reports.

Is this deliberate, or an oversight?

justvanrossum:
> Is this deliberate, or an oversight?

Option three: no one got around building it. It's always option three.

frankrolf:
The only real complaint I have here is that a CFF2 VF is passed through as “fine”. An error message might be clearer (and doesn’t take that long to build).

behdad:
> The only real complaint I have here is that a CFF2 VF is passed through as “fine”. An error message might be clearer (and doesn’t take that long to build).

It should be rather trivial to fix this. Which font can I use with?

frankrolf:
Didn’t get to say “thank you“ here!
Just used it on a CFF2 VF, and it works impeccably! :-)

behdad:
You're welcome.
Glad it works.

Issue #3665, opened 2024-10-21
Should drop support for python 3.8 EOL


anthrotype:
3.8 reached end of life a week ago, https://devguide.python.org/versions/

we could now bump the mimimum python and make use of [3.9-only features](https://docs.python.org/3/whatsnew/3.9.html)

anthrotype:
however, if we stick to the [NEP deprecation policy](https://numpy.org/neps/nep-0029-deprecation_policy.html) (which our README states we do), we could even go futher and drop 3.9 as well and go full in on Python 3.10+...

The latter has even more juicy new features like structural pattern matching (Rust-like `match` expressions)
https://docs.python.org/3/whatsnew/3.10.html

behdad:
I sugget bumping the library major version each time dumping a Python version.

behdad:
> we have only bumped the major version once during my involvement with fonttools maintenance and it was when we dropped support for python 2. When we dropped support for EOL pythons, we only bumped the minor version. A major version bump signals a break in backward compatibility of the public API (which we hardly ever do, at least on purpose). I'm not sure. Maybe this is an opportunity to bump major more frequently and get rid of lots of legacy stuff for good.

That works for libraries that do break back-compat often. In HarfBuzz we never do, so we switched to bumping major version on major changes, not on back-compat breakage. It feels like fonttools major version will remain at what it is for a loooooooooooooong time if we follow semver.

Hoolean:
Not user-facing but partially relevant: looks like [our current `dev-requirements.txt`](https://github.com/fonttools/fonttools/blob/7e374c53da9a7443d32b31138a0e5be478bcbab9/dev-requirements.txt) no longer installs on Python 3.8:

> the current Python version (3.8.20) does not satisfy Python>=3.9 and black==24.10.0 depends on Python>=3.9

lianghai:
> The latter has even more juicy new features like structural pattern matching (Rust-like `match` expressions) https://docs.python.org/3/whatsnew/3.10.html

Another advantage of 3.10 is “[Allow writing union types as `X | Y`](https://peps.python.org/pep-0604/)”. I feel with that it’ll be a good time to [type annotate feaLib.ast](https://github.com/fonttools/fonttools/pull/3735#issuecomment-2574567142).

[...]

PR #3670, opened 2024-10-22
[`varLib.interpolatable`] Support CFF2 fonts

behdad:
It's rather straightforward. But testing with `source-serif-4.005_Desktop/VAR/SourceSerif4Variable-Italic.otf` vs `source-serif-4.005_Desktop/VAR/SourceSerif4Variable-Italic.ttf` in:

  https://github.com/adobe-fonts/source-serif/releases/download/4.005R/source-serif-4.005_Desktop.zip

I'm not satisfied that it's all working. The OTF version of the font reports more problems than the TTF version. Which is surprising. So, testing is appreciated. @frankrolf

Fixes https://github.com/fonttools/fonttools/issues/3666


frankrolf:
Thank you for this, I can confirm it’s working.

The reason for different scope of problems reported could be:
* everything is decomposed in this CFF2, while the TTF still has components. Hence the CFF reporting on X, Chi, and Ha (XΧХ) while the TTF only reports on the base component glyph
* curve conversion may be responsible for new kinking possibilities being introduced

behdad:
> * everything is decomposed in this CFF2, while the TTF still has components. Hence the CFF reporting on X, Chi, and Ha (XΧХ) while the TTF only reports on the base component glyph

Good point.

> * curve conversion may be responsible for new kinking possibilities being introduced

The "J" for example, is kinky in both TTF and OTF. But only reported on the OTF. So it's a fals negative on the TTF side. I'll file another issue for that.

behdad:
I have no clue why the test is failing on Python 3.8 (not finding the kink).

behdad:
> > * curve conversion may be responsible for new kinking possibilities being introduced
> 
> The "J" for example, is kinky in both TTF and OTF. But only reported on the OTF. So it's a fals negative on the TTF side. I'll file another issue for that.

So it looks like the kink is not reported in TTF because it's deemed not bad enough to be reported. This is not good. Investigating more.

behdad:
I don't know why Python 3.8 is failing. Otherwise, this is ready I think.

anthrotype:
I really don't see why this should fail on python3.8 but pass on >3.12 or 3.13...

maybe it is time to drop it https://github.com/fonttools/fonttools/issues/3665

anthrotype:
I suggest for now you add a `@pytest.mark.skipif(sys.version_info[:2] == (3, 8))` decorator to the failing test, until we actually go and do drop support for 3.8 (I don't have the time right now)

anthrotype:
feel free to merge if @frankrolf is happy with the changes

anthrotype:
looks like you _must_ also specify a reason `pytest.mark.skipif(condition, reason="whatever")`

frankrolf:
Thanks all!

behdad:
> Thanks all!

You're welcome. Let me know if you hit any issues with it.

Issue #3673, opened 2024-10-25
[`CFF2`] Lift `uint16 VariationStore.length` limitation


behdad:
See this for details:
https://github.com/harfbuzz/boring-expansion-spec/issues/159

I suggest we implement that already. That is, if the `VariationStore` struct's `length` field is 65,535, then allow the embedded `ItemVariationStore` to reach out to the rest of the font file.

@justvanrossum has been building a font that hit that limit.

We are going to also implement this change in OTS. HarfBuzz already renders such fonts fine, and I will test FreeType and report.

OTS report: https://github.com/harfbuzz/boring-expansion-spec/issues/159

PR #3674, opened 2024-10-25
[`CFF2`] Lift `uint16 VariationStore.length` limitation

behdad:
Fixes https://github.com/fonttools/fonttools/issues/3673

### DIFF
--- a/Lib/fontTools/cffLib/__init__.py
+++ b/Lib/fontTools/cffLib/__init__.py
@@ -867,7 +867,11 @@ class VarStoreData(object):
         if self.file:
             # read data in from file. Assume position is correct.
             length = readCard16(self.file)
-            self.data = self.file.read(length)
+            # https://github.com/fonttools/fonttools/issues/3673
+            if length == 65535:
+                self.data = self.file.read()
+            else:
+                self.data = self.file.read(length)
             globalState = {}
             reader = OTTableReader(self.data, globalState)
             self.otVarStore = ot.VarStore()
@@ -1956,7 +1960,8 @@ class VarStoreCompiler(object):
         self.parent = parent
         if not varStoreData.data:
             varStoreData.compile()
-        data = [packCard16(len(varStoreData.data)), varStoreData.data]
+        varStoreDataLen = min(0xFFFF, len(varStoreData.data))
+        data = [packCard16(varStoreDataLen), varStoreData.data]
         self.data = bytesjoin(data)
 
     def setPos(self, pos, endPos):


Lib/fontTools/cffLib/__init__.py
```diff
@@ -867,7 +867,12 @@ def decompile(self):
         if self.file:
             # read data in from file. Assume position is correct.
             length = readCard16(self.file)
-            self.data = self.file.read(length)
+            # https://github.com/fonttools/fonttools/issues/3673
+            if length == 65535:
+                length = 65536 * 16  # *This* out to be enough for anyone.
+                self.data = self.file.read()
```
justvanrossum, line 873: I assume you meant this?
```suggestion
                self.data = self.file.read(length)
```

justvanrossum: Did you check the failures we're getting if we unconditionally do `self.file.read()`? I couldn't immediately make sense of them, but it does make me worry that leaving the file pointer at the wrong location may have consequences elsewhere.

And since this block isn't covered by tests, we don't know, really.

justvanrossum: > I assume you meant this?

If so, then the file read can move out of the if/else.

behdad: Done.

> Did you check the failures we're getting if we unconditionally do self.file.read()? I couldn't immediately make sense of them, but it does make me worry that leaving the file pointer at the wrong location may have consequences elsewhere.

You mean test failures? I don't get any.

behdad:
Also, 16x wasn't enough! So I made it 256x.

Lib/fontTools/cffLib/__init__.py
justvanrossum: I think have messed up something, I can't reproduce now.

behdad:
Should we land this?

anthrotype:
> Also, 16x wasn't enough! So I made it 256x.

then why not simply let it read the whole file?

behdad:
> > Also, 16x wasn't enough! So I made it 256x.
> 
> then why not simply let it read the whole file?

I guess we can do that... I thought 16MB is still better than 50MB. But can remove it.

anthrotype:
it's just that the magic number feels a bit arbitrary and may come back biting us again in the future. If the font is so big that Python ends up OOM, well... we don't protect ourselves from that e.g. when we read the TTFont upfront (unless lazy=True)

behdad:
> it's just that the magic number feels a bit arbitrary and may come back biting us again in the future. If the font is so big that Python ends up OOM, well... we don't protect ourselves from that e.g. when we read the TTFont upfront (unless lazy=True)

Done.

PR #3677, opened 2024-11-02
Update Black min python version

benkiel:
Matches min version now


benkiel:
@anthrotype ok to merge? Or should i say 3.9?

khaledhosny:
#3815 changed it to py39

PR #3683, opened 2024-11-05
[ttGlyphSet] Support VARC CFF2 fonts

behdad:


behdad:
@justvanrossum @anthrotype May I commit this?

justvanrossum:
It looks fine to me, but I didn't properly test it.

PR #3697, opened 2024-11-22
[DecomposedTransform] Document and implement always skewY == 0

behdad:
Spotted by Cosimo. I convinced myself he is right, since a and b are zero in that branch.

### DIFF
--- a/Lib/fontTools/misc/transform.py
+++ b/Lib/fontTools/misc/transform.py
@@ -437,8 +437,20 @@ class DecomposedTransform:
 
     @classmethod
     def fromTransform(self, transform):
+        """Return a DecomposedTransform() equivalent of this transformation.
+        The returned solution always has skewY = 0, and angle in the (-180, 180].
+
+        :Example:
+                >>> DecomposedTransform.fromTransform(Transform(3, 0, 0, 2, 0, 0))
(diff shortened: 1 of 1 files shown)


Lib/fontTools/misc/transform.py
```diff
@@ -450,21 +462,20 @@ def fromTransform(self, transform):
...
             scaleX, scaleY = (r, delta / r)
-            skewX, skewY = (math.atan((a * c + b * d) / (r * r)), 0)
+            skewX = math.atan((a * c + b * d) / (r * r))
         elif c != 0 or d != 0:
             s = math.sqrt(c * c + d * d)
             rotation = math.pi / 2 - (
                 math.acos(-c / s) if d >= 0 else -math.acos(c / s)
             )
             scaleX, scaleY = (delta / s, s)
-            skewX, skewY = (0, math.atan((a * c + b * d) / (s * s)))
```
justvanrossum, line 467: Are you saying it's never possible for this math.atan() expression to return anything but zero?

behdad: Yes. No?

justvanrossum: Ah, it is because we know that both a and b are zero at this point, right?

behdad: Exactly.

justvanrossum (approved):

Issue #3703, opened 2024-11-27
varLib.iup_delta_optimize can produces different outputs when using `cython`


cmyr:
Still digging into this, but feels worth opening an issue at this point:

The following python program produces different output depending on whether or not varLib is using `cython`:

```python
from fontTools.varLib import iup_delta_optimize


def main():
    deltas = [
        (40, 0),
        (24, -104),
        (15, -104),
        (3, -104),
        (-19, -92),
        (-32, -68),
        (0, 0),
        (0, 0),
        (0, 0),
        (0, 0),
    ]
    orig = [
        (206, 0),
        (174, 458),
        (152, 458),
        (124, 458),
        (74, 486),
        (42, 541),
        (0, 0),
        (0, 0),
        (0, 0),
        (0, 0),
    ]

    result = iup_delta_optimize(deltas, orig, [5], 0.5)
    print(result)

if __name__ == "__main__":
    main()
```

with cython:
```
[(40, 0), (24, -104), None, (3, -104), (-19, -92), (-32, -68), None, None, None, None]
```

without cython:
```
[(40, 0), (24, -104), None, None, (-19, -92), (-32, -68), None, None, None, None]
```

I suspect this has something to do with how the cython code is handling the `tolerance` param; a very good clue is that the output is identical if I change the tolerance from `0.5` to `0.50001`.

edit: right, this strongly echoes #3282

behdad:
Ugh. Not again!

behdad:
It sounds like eg. a fused mul/sum or some other result-changing optimization in the Cython compiled code...

behdad:
> edit: right, this strongly echoes #3282

Yeah I was just looking for that...

behdad:
> > edit: right, this strongly echoes #3282
> 
> Yeah I was just looking for that...

Despite the similarity, I think this is just a float-precision issue.

cmyr:
"fixed" in #3704

PR #3704, opened 2024-11-28
[varLib] "Fix" cython iup issue?

cmyr:
See #3703.

In some cases we were seeing different output from iup depending on whether or not we were running cython code.

I've tracked this particular issue down to the line that is changed in this diff, and the change introduced in this diff does (locally, for me, on one machine with one architecture and one compiler) seem to suppress the problem.

However... it feels pretty bad??

I'm not sure how motivated I am to try and generate a proper minimal test case and try to get this fixed upstream. I guess I'm.. medium motivated? But at the very least it would be nice to figure out a more robust way to prevent this optimization from happening, and at the very _very_ least it would be nice to figure out away to test this.

The solution I was hoping for was some way to write some actual hand-written C so we could have finer-grained control over what's going on, and use that just for this one little bit of arithmetic, but I didn't see an easy way to do that.

In any case, I'm going to PR this as a conversation starter at least, and we can try to figure out what the best solution looks like?

### DIFF
--- a/Lib/fontTools/varLib/iup.py
+++ b/Lib/fontTools/varLib/iup.py
@@ -77,7 +77,14 @@ def iup_segment(
                 d = d2
             else:
                 # Interpolate
-                d = d1 + (x - x1) * scale
+                #
+                # NOTE: we assign an explicit intermediate variable here in
+                # order to disable a fused mul-add optimization. See:
+                #
+                # - https://godbolt.org/z/YsP4T3TqK,
+                # - https://github.com/fonttools/fonttools/issues/3703
+                nudge = (x - x1) * scale
+                d = d1 + nudge
 
             out.append(d)
 


behdad:
Ah, so it was a fused mul-add after all! Yeah the best thing would be to add a compiler barrier there. I'll take a look later, I don't think it's possible. The nudge forces this to go to a Python object and back, which is annoying.

behdad:
Seems like verbatim C can be added:
https://stackoverflow.com/questions/57965780/cython-openmp-single-barrier

behdad:
I'm working on it...

cmyr:
> Ah, so it was a fused mul-add after all! Yeah the best thing would be to add a compiler barrier there. I'll take a look later, I don't think it's possible. The nudge forces this to go to a Python object and back, which is annoying.

I looked at the generated C and it doesn't seem to, though? Everything is a `double`:

```
  079:                 # Interpolate
+080:                 nudge = (x - x1) * scale
      /*else*/ {
        __pyx_v_nudge = ((__pyx_v_x - __pyx_v_x1) * __pyx_v_scale);
+081:                 d = d1 + nudge
        __pyx_v_d = (__pyx_v_d1 + __pyx_v_nudge);
      }
      __pyx_L10:;
```

and so I'm surprised this breaks the optimization? Initially I had a `print` statement that printed intermediate results, but then I saw that it was working even without that..

behdad:
Ugh. I'm doing some refactoring and now gcc/clang hang on compiling the modules!

cmyr:
Updated this to explain what's going on in a comment, and add a link to godbolt showing the optimization on & off. It shouldn't _hurt_ to merge...

behdad:
LGTM. Thanks.

PR #3706, opened 2024-11-28
[cython] Simplify import more

behdad:
We're not accessing any attribute so it doesn't make sense to me. This was probably changed in:

  e037cea7261857d44ab6b797b3de7d493fb1927c

But even then, I'm not sure why cython module might not have the compiled attribute. At any rate, I think this is safe.


behdad:
Should I just close this?

PR #3712, opened 2024-12-03
[Docs] update Sphinx config

n8willis:
This is part of refactoring the doc system to realign the different layouts found in different packages and make them consistent.

Small change to the overall config, plus updates to several of the single-page modules as test cases.

### DIFF
--- a/Doc/source/afmLib.rst
+++ b/Doc/source/afmLib.rst
@@ -2,7 +2,14 @@
 afmLib: Read and write Adobe Font Metrics files
 ###############################################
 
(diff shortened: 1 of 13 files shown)


n8willis:
Also adding `qu2cu` pages to the docs structure; that package had been missing previously and, from conversations, that seemed to be an unintentional gap.

n8willis:
@anthrotype If you don't object, I'd like to merge in the `Doc/`-only bits and get on to staging other things that might warrant a little real-human review.

behdad:
LGTM.

PR #3716, opened 2024-12-04
Docs: fill out ttLib table section

n8willis:
This is primarily repetitive stuff, adding some basic information to each module page under `ttLib/tables` in the Docs hierarchy and some basic docstring content to the modules themselves — at least, for the ones that didn't already have it.

I think there are a number of table modules that could use more, mainly as a "newcomer-friendly" feature that can point them in the right direction if they're needing to access or manipulate something specific. This .rst+docstring setup should allow that to go either place.

### DIFF
--- a/Doc/source/ttLib/tables.rst
+++ b/Doc/source/ttLib/tables.rst
@@ -1,13 +1,131 @@
-###############################
-TrueType/OpenType Table Modules
-###############################
+#######################################
+tables: Access TrueType/OpenType tables
+#######################################
+
+.. contents:: On this page
+    :local:
+
+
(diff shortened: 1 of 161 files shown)


n8willis:
@behdad I think this is fairly cut and dried, but if you have any feedback on the wording parts, especially in the top-level overview pages, that would be appreciated.

As per our discussion of entry points, I started marking helper/utility-only submodule pages with explicit notes about what they're designed for, to point people in the right direction, and putting the TOCtrees to them in page sections that designate them as helpers. It seems easier to navigate to me, anyhow.

PR #3717, opened 2024-12-04
[feaLib] Don’t modify variable anchors in place

khaledhosny:
When passing a parsed feature file that has variable anchors to addOpenTypeFeatures(), builder would modify the anchors in place and discard the variations, which break any subsequent use of the feature file.

I encountered this building a font that has variable cursing anchors with ufo2ft. The cursFeatureWriter would write the variable anchors, but then when kernFeatreWriter compiles the file to get GSUB closure, the variation would be dropped from the anchors, and later when when the feature data is compiled into the font, the anchors would be compiled without variations.

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -1658,38 +1658,31 @@ class Builder(object):
 
         return default, device
 
+    def makeAnchorPos(self, varscalar, deviceTable, location):
+        device = None
+        if not isinstance(varscalar, VariableScalar):
+            if deviceTable is not None:
+                device = otl.buildDevice(dict(deviceTable))
+            return varscalar, device
+        default, device = self.makeVariablePos(location, varscalar)
+        if device is not None and deviceTable is not None:
+            raise FeatureLibError(
+                "Can't define a device coordinate and variable scalar", location
+            )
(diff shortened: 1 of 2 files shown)


Lib/fontTools/feaLib/builder.py
```diff
@@ -1658,38 +1658,31 @@ def makeVariablePos(self, location, varscalar):
...
-                    deviceX = device
-                else:
-                    deviceY = device
-                variable = True
-
-        otlanchor = otl.buildAnchor(
-            anchor.x, anchor.y, anchor.contourpoint, deviceX, deviceY
-        )
-        if variable:
-            otlanchor.Format = 3
```
khaledhosny, line 1692: This checks was redundant, since `otl.buildAnchor()` will set format to 3 if either `deviceX` or `deviceY` is not None, so I took the opportunity to simplify the code.

simoncozens:
Sorry, my fault; I guess I didn't expect the builder to be run more than once, but of course why not?

khaledhosny:
Passing a parsed feature file to the builder is probably something only ufo2ft does. It also reveals that we are doing work that will be thrown away, since ufo2ft asks to build GSUB table only but we are building positioning statements regardless.

PR #3720, opened 2024-12-06
[Docs] ttLib: higher-level navigation and helper documentation

n8willis:
Adds text to help orient newcomers at ttLib and at ttLib/ttFont, and fills in helper/utility module pages.

This also links the idea of TTGlyphs (or ttGlyphSet members) to the pen protocol, hopefully clarifying that relationship.


n8willis:
@khaledhosny Thanks for taking a look at this. I've been away from the keyboard for a bit, but I am revisiting these. If you had any feedback, it'd be most appreciated.

PR #3724, opened 2024-12-09
[Docs] Improvements to Pens and pen-module documentation

n8willis:


benkiel:
You might want to look at the old robofab documentation on pens; may be some things from that worth bringing over

benkiel:
See:
- https://robofab.org/howtos/use_pens.html
- https://robofab.org/objects/pens.html

n8willis:
Great; thanks for the suggestion! I have definitely seen the second of those links (but found it a bit terse). The first one is a lot more useful.

If you don't mind, what sections on the FT pens section do you think need the most help? Like, is handling the "protocol" stuff where it currently is, in basePen, useful? I have kind of been going top-down, but I don't want to sound like I think that's the one true approach. It's just what has worked for me understanding it....

benkiel:
I think that what you have (quick read) is good for the base level of understanding at a developer level, but maybe a 'using pens' section like the first link at a top level is good — Pens are something that a lot of folks get confused by for sure

jenskutilek:
FWIW, I gave a workshop/presentation about pens at TypoLabs 2016. Feel free to salvage any interesting parts from it :)

https://github.com/jenskutilek/TypoLabs2016

n8willis:
Yeah, I have been puzzling over how to fit the different flavors of documentation into a single structure. There are kind of three levels that all get munged together by the Sphinx setup:
- explicitly written stuff in the .rst files
- docstrings on the public members
- module/package level docstrings

and the build system only gives you partial control over what it (or, autodoc, that is) inserts where. In a perfect world, I think that every pen could probably use a "how to use" / examples block. It seems like that would be best done in the .rst files, and — separately from that — the actual pen docstrings would give you the signature / definitions for output etc. So there's a cleaner separation between what a developer would see in the Python code and what a user would want to start with.

Right now, those things are really intertwined, and intertwined differently in different pens. Module-level could maybe provide an overview, but there isn't really a bright-line distinction to make in some cases.

Probably the "protocol" needs a page of its own, disentangled from the BasePen/AbstractPen/PointPen classes. The separation there is definitely not a clean one.

PR #3728, opened 2024-12-10
Add `GVAR` table

behdad:


behdad:
@justvanrossum @anthrotype Can I merge this?

Lib/fontTools/ttLib/tables/_g_v_a_r.py
```diff
@@ -23,19 +23,24 @@
...
+# of different size: 2 bytes for gvar, and 3 bytes for GVAR.
+GVAR_HEADER_FORMAT_TAIL = """
+	> # big endian
 	flags:				H
 	offsetToGlyphVariationData:	I
 """
 
-GVAR_HEADER_SIZE = sstruct.calcsize(GVAR_HEADER_FORMAT)
+GVAR_HEADER_SIZE_HEAD = sstruct.calcsize(GVAR_HEADER_FORMAT_HEAD)
+GVAR_HEADER_SIZE_TAIL = sstruct.calcsize(GVAR_HEADER_FORMAT_TAIL)
```
khaledhosny, line 43:
```suggestion
GVAR_HEADER_HEAD_SIZE = sstruct.calcsize(GVAR_HEADER_FORMAT_HEAD)
GVAR_HEADER_TAIL_SIZE = sstruct.calcsize(GVAR_HEADER_FORMAT_TAIL)
```

Lib/fontTools/ttLib/tables/_g_v_a_r.py
```diff
@@ -23,19 +23,24 @@
...
 	offsetToSharedTuples:		I
-	glyphCount:			H
+"""
+# In between the HEAD and TAIL lies the glyphCount, which is
+# of different size: 2 bytes for gvar, and 3 bytes for GVAR.
+GVAR_HEADER_FORMAT_TAIL = """
+	> # big endian
 	flags:				H
 	offsetToGlyphVariationData:	I
 """
```
khaledhosny, line 40: I feel that duplicating `GVAR_HEADER_FORMAT` for each table would make the code simpler. May be have a `GVAR_HEADER_FORMAT_TEMPLATE` or some such that has `glyphCount: %s`, then the table class would have something like `GVAR_HEADER_FORMAT = GVAR_HEADER_FORMAT_TEMPLATE % gid_size`.

behdad: I think the problem is that there's no way to read 24bit using `sstruct`. No?

khaledhosny: I didn’t know that. OK then.

PR #3729, opened 2024-12-10
[otBase] Use int.from_bytes() instead of struct.unpack()

behdad:
Should be faster but I haven't measured. Is definitely simpler.


behdad:
I have no clue why the test is failing in 3.8 only.

behdad:
> I have no clue why the test is failing in 3.8 only.

It was a bug in the test actually! Fixed now.

behdad:
Actually running with `%timeit` in IPython suggests that this slows things down. Closing.

behdad:
For reference:

https://bugs.python.org/issue27638#msg271501

PR #3735, opened 2025-01-06
Correct docstring of AlternateSubstStatement

lianghai:
The current docstring suggests that both `.glyph` and `.replacement` should be **lists** of “glyph-containing objects”:

https://github.com/fonttools/fonttools/blob/ef07e3b2cc6987a35ba32d8dcbb79a435965ba3b/Lib/fontTools/feaLib/ast.py#L595-L604

But the `asFea()` implementation actually expects `.glyph` and `.replacement` to be “glyph-containing objects”, not nested in lists:

https://github.com/fonttools/fonttools/blob/ef07e3b2cc6987a35ba32d8dcbb79a435965ba3b/Lib/fontTools/feaLib/ast.py#L616-L629

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -595,8 +595,8 @@ class MarkClassDefinition(Statement):
 class AlternateSubstStatement(Statement):
     """A ``sub ... from ...`` statement.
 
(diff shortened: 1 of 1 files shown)


simoncozens:
*sigh* We should really get this stuff fully typed. Thanks for the PR!

lianghai:
I’ll be happy to work on type annotation one day when the Python requirement has been raised to 3.10! 🫠

PR #3740, opened 2025-01-14
Default to not including entries for mac (platformID=1)

driehuis:
Default to not including entries for mac (platformID=1), to match deprecation in fontbakery.

Without this change, statmake would add platformID=1 names for the axis labels, which is inconsistent with the rest of the names in the font if it started out without any platformID=1 names. In general, any caller of buildStatTable that does not set macNames explicity would get deprecated platformID=1 names.

Applications that bother to pass a value for macNames would see their preference honored, so as far as I can see this is a safe fix.

### DIFF
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -2708,7 +2708,7 @@ AXIS_VALUE_POSITIVE_INFINITY = fixedToFloat(0x7FFFFFFF, 16)
 
 
 def buildStatTable(
-    ttFont, axes, locations=None, elidedFallbackName=2, windowsNames=True, macNames=True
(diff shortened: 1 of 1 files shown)


driehuis:
Hmmm, from the looks of it I can't fix the readthedocs failure because it looks like it can't cope with pull requests from outside branches.

jenskutilek:
Hm, shouldn't the adding of Mac platform name entries be synced with the existing name table of the font? Or is this decision up to the caller of the `buildStatTable` function anyway?

driehuis:
It's an optional parameter, which should not default to true for a deprecated feature.

I'd be totally fine with it if otlLib did a check to see if there are any deprecated name records already. But I doubt its value, especially given that any caller who cares can just specify the macNames parameter. Default parameters should be sensible.

driehuis:
Statmake 1.0.0 will address my concern. Thanks!

PR #3747, opened 2025-01-21
Come back from the future to work around #3744

### LINKED ISSUE #3744: JSON ufo deserialization broken?
It appears that I cannot load JSON ufo sources in ufoLib2 when using current fonttools HEAD. I did a git bisect and it pointed to the innocuous looking c00bc6fec; the error i'm seeing is originating in `cattrs` and seems to be related to an unexpected type showing up in a converter function? I'm not sure how any of that works.

To reproduce, run the tests for `ufoLib2` after installing `fonttools@c00bc6f`.

Hoolean:
Fixes #3744

In theory, forward references are harmless, but until we can tackle the bigger issue of supporting them in ufoLib2's cattrs-based serialisation it is more practical for us to represent unions with the old syntax.

When testing locally, this commit helped ufoLib2's serde test suite pass while retaining transform's current behaviour in type checkers .

### DIFF
--- a/Lib/fontTools/misc/transform.py
+++ b/Lib/fontTools/misc/transform.py
@@ -52,10 +52,8 @@ translate, rotation, scale, skew, and transformation-center components.
 	>>>
 """
 
-from __future__ import annotations
-
 import math
-from typing import NamedTuple
+from typing import NamedTuple, Optional
 from dataclasses import dataclass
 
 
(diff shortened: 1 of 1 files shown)


Hoolean:
Closing, we landed a better fix upstream at fonttools/ufoLib2#367 🚀

PR #3750, opened 2025-01-23
[glyf] fix bbox of transformed components with un-rounded coordinates

anthrotype:
The first commit reproduces the issue found in https://github.com/googlefonts/fontc/issues/1206, whereby the bounding box of a composite glyph with non-trivial transforms that references a base glyph in turn containing un-rounded floating-point coordinates, may end up with off-by-ones 

The fix (in a follow up commit) is to compute their bounding box on the rounded integer coordinates.

context: when building a VF, ufo2ft passes to fonttools TTGlyphs with unrounded coordinates, so that the latter can effectively drop implied off-curve glyf points (when that optional feature is enabled). We want to keep the coords unrounded (they will be eventually rounded anyway upon saving, after dropImpliedOncurvePoints has done its thing), but at the same time we want ufo2ft to compute the _final_ glyf bounding boxes from the to-be-rounded integer coordinates. We don't want that a bunch of composite glyphs' bboxes get "fixed" e.g. if one dumps the VF to xml and recompiles with ttx, for example.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -1187,7 +1187,7 @@ class Glyph(object):
         ):
             return
         try:
-            coords, endPts, flags = self.getCoordinates(glyfTable)
+            coords, endPts, flags = self.getCoordinates(glyfTable, round=otRound)
             self.xMin, self.yMin, self.xMax, self.yMax = coords.calcIntBounds()
         except NotImplementedError:
             pass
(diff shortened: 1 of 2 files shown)


anthrotype:
and the computation of the final bounding boxes is crucial also because the hmtx left sidebearings are set to glyph.xMin by the compiler, so we want the latter to be right

anthrotype:
as expected, test fails with: 

```
>       assert (glyph.xMin, glyph.yMin, glyph.xMax, glyph.yMax) == bounds
E       assert (73, 680, 329, 834) == (74, 680, 329, 834)
E         
E         At index 0 diff: 73 != 74
```

https://github.com/fonttools/fonttools/actions/runs/12927979510/job/36054284702?pr=3750

PR #3752, opened 2025-01-24
[recalcBounds] fix rounding of bbox for nested transformed components

anthrotype:
This is a regression introduced with https://github.com/fonttools/fonttools/pull/3750

The test currently fails because after that PR we are rounding the coordinates extracted from each component of a composite glyph *before* these are transformed; with nested components of components with non trivial transforms, the rounding errors compound leading to incorrect bounds.

The correct way to do this (in a way that gives the same result whether the simple glyphs' original coordinates were rounded to begin with or not) is to ensure the rounding occurs only at the leaves (on the simple glyphs), and
at the very end of the chain after all the nested component trasforms have been concatenated and applied; but not at intermediate levels of the component tree.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -1206,9 +1206,7 @@ class Glyph(object):
         Return True if bounds were calculated, False otherwise.
         """
         for compo in self.components:
-            if hasattr(compo, "firstPt") or hasattr(compo, "transform"):
-                return False
-            if not float(compo.x).is_integer() or not float(compo.y).is_integer():
+            if not compo._hasOnlyIntegerTranslate():
                 return False
 
         # All components are untransformed and have an integer x/y translate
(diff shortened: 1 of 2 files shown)


anthrotype:
test fails like so

```
    def assertGlyphBoundsEqual(self, glyph, bounds):
>       assert (glyph.xMin, glyph.yMin, glyph.xMax, glyph.yMax) == bounds
E       assert (85, 163, 267, 345) == (84, 163, 267, 345)
E         
E         At index 0 diff: 85 != 84
```

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -1258,7 +1258,11 @@ def getCoordinates(self, glyfTable, round=noRound):
         """
 
         if self.numberOfContours > 0:
-            return self.coordinates, self.endPtsOfContours, self.flags
+            coordinates = self.coordinates
+            if round is not noRound:
```
rsheeter, line 1262: I suggest a comment

rsheeter (approved):

Lib/fontTools/ttLib/tables/_g_l_y_f.py
anthrotype: added a comment, and made the changes more lightweight

PR #3756, opened 2025-01-29
[unicodedata] Fix bug in buildUCD.py

cmyr:
The method that was being used to read the header from the input was inadvertently dropping the first non-header line on the floor; although this happens to be okay in some cases (where there is an empty line after the header) in the case of newer versions of the ScriptExtensions.txt file, this was causing the generated code to be missing the first entry, for U+00B7 (MIDDLE DOT)


anthrotype:
Good catch, thanks!

PR #3758, opened 2025-01-30
[otData] Fix DeltaValue repeat value

behdad:
The "" marker is used when a custom converter expects a list. For DeltaValues, we expect a list as a single value, not a list to be enumerated in XML. So, None is appropriate here. Cryptic, I know...

Fixes https://github.com/fonttools/fonttools/pull/3757

### DIFF
--- a/Lib/fontTools/misc/testTools.py
+++ b/Lib/fontTools/misc/testTools.py
@@ -58,6 +58,9 @@ class FakeFont:
         self.tables = {}
         self.cfg = Config()
 
+    def __contains__(self, tag):
+        return tag in self.tables
+
     def __getitem__(self, tag):
         return self.tables[tag]
 
--- a/Lib/fontTools/ttLib/tables/otData.py
+++ b/Lib/fontTools/ttLib/tables/otData.py
@@ -385,7 +385,7 @@ otData = [
             (
                 "DeltaValue",
                 "DeltaValue",
-                "",
+                None,
                 "DeltaFormat in (1,2,3)",
                 "Array of compressed data",
             ),
--- a/Tests/ttLib/tables/otTables_test.py
+++ b/Tests/ttLib/tables/otTables_test.py
@@ -4,6 +4,7 @@ from fontTools.misc.xmlWriter import XMLWriter
 from fontTools.ttLib.tables.otBase import OTTableReader, OTTableWriter
 import fontTools.ttLib.tables.otTables as otTables
 from io import StringIO
+from textwrap import dedent
 import unittest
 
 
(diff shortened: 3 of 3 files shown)


anthrotype:
wow.. the DeltaValue's conv.repeat field has been an empty string for 23 years and nobody noticed? 🤔 
Is it like nobody ever used it?

Just to confirm I understand, seeting it to `None` basically is to take the `else` branch of this `if` statement, in `BaseTable.fromXML`?

https://github.com/fonttools/fonttools/blob/b467579c8d5a8fd642bb81d5922d650ea267d1b6/Lib/fontTools/ttLib/tables/otBase.py#L1148-L1159

behdad:
> wow.. the DeltaValue's conv.repeat field has been an empty string for 23 years and nobody noticed? 🤔
> Is it like nobody ever used it?

I probably broke it in 2013: 7d130307e6e6094e40a841a3c41fc74b203b7e8e

But yeah, wow!

> Just to confirm I understand, seeting it to `None` basically is to take the `else` branch of this `if` statement, in `BaseTable.fromXML`?

Correct.

anthrotype:
I'll work on a test tomorrow

behdad:
> I probably broke it in 2013: [7d13030](https://github.com/fonttools/fonttools/commit/7d130307e6e6094e40a841a3c41fc74b203b7e8e)

Oh. Probably more recently: ec78b572c9ea1516f54a281524e08a1a9cc5137d

behdad:
> > I probably broke it in 2013: [7d13030](https://github.com/fonttools/fonttools/commit/7d130307e6e6094e40a841a3c41fc74b203b7e8e)
> 
> Oh. Probably more recently: [ec78b57](https://github.com/fonttools/fonttools/commit/ec78b572c9ea1516f54a281524e08a1a9cc5137d)

I just checked, and DeltaValues are the only suspect of breakage from my change. The rest are intentional I believe:
```

>>> from otData import otData 
>>> for name,attrs in otData:
...     for typ,attrName,repeat,aux,desc in attrs:
...         if not repeat and repeat is not None:
...             print(name, attrName)
...             
Device DeltaValue
VarIdxMap mapping
DeltaSetIndexMapFormat0 mapping
DeltaSetIndexMapFormat1 mapping
MultiVarData Item
AxisIndicesList Item
VarCompositeGlyphs VarCompositeGlyph
```

behdad:
LGTM. THanks.

anthrotype:
I force-pushed so that the test now does a full roundtrip (XML->binary->XML)

Issue #3478, opened 2024-04-12
varLib.mutator Font instancing produces invalid font file for certain fonts


mitterdoo:
Hiya,

I'm trying to create static font files (TTF) for the font Bahnschrift (now included with Windows) since the variable font does not play nice with Autodesk Fusion. However, when I run this command:
```powershell
fonttools varLib.mutator .\bahnschrift.ttf wght=300 wdth=100
```
then I try to open the new font file, Windows gives me the following error:
![The requested file bahnschrift-instance.ttf is not a valid font file.](https://github.com/fonttools/fonttools/assets/4317493/dc56867e-36cf-4fa1-8995-389edcc36b9d)

I'm not sure what the cause is, and I tried several different weight/width values to no avail. Below is the output with the verbose flag enabled:
```
Location: {'wght': 300.0, 'wdth': 100.0}
Loading variable font
Reading 'fvar' table from disk
Decompiling 'fvar' table
Reading 'avar' table from disk
Decompiling 'avar' table
Normalized location: {'wght': -1.0, 'wdth': 0.0}
Mutating glyf/gvar tables
Reading 'gvar' table from disk
Decompiling 'gvar' table
Reading 'post' table from disk
Decompiling 'post' table
Reading 'maxp' table from disk
Decompiling 'maxp' table
Reading 'glyf' table from disk
Decompiling 'glyf' table
Reading 'loca' table from disk
Decompiling 'loca' table
Reading 'head' table from disk
Decompiling 'head' table
Reading 'hmtx' table from disk
Decompiling 'hmtx' table
Reading 'hhea' table from disk
Decompiling 'hhea' table
Reading 'vmtx' table from disk
Mutating MVAR table
Reading 'MVAR' table from disk
Decompiling 'MVAR' table
Reading 'OS/2' table from disk
Decompiling 'OS/2' table
Mutating FeatureVariations
Reading 'GSUB' table from disk
Decompiling 'GSUB' table
Reading 'GPOS' table from disk
Decompiling 'GPOS' table
Reading 'GDEF' table from disk
Decompiling 'GDEF' table
Mutating GDEF/GPOS/GSUB tables
Pruning name table
Reading 'name' table from disk
Decompiling 'name' table
Removing variable tables
Saving instance font .\bahnschrift-instance.ttf
Compiling 'glyf' table
Writing 'glyf' table to disk
Compiling 'maxp' table
Writing 'maxp' table to disk
Compiling 'loca' table
[...]

PR #3765, opened 2025-02-04
Check for name ID in other platforms if needed

arrowtype:
I’m running into a similar issue as is described at https://github.com/fonttools/fonttools/issues/3478#issuecomment-2051372683

I’m attempting to create instances from a variable font that has all name entries for `platformID="3" platEncID="1"`, but only a limited set for `platformID="1" platEncID="0"` (basically, just names 0–14). From what I’ve observed, this is fairly common.

The instancing fails because the name handling comes across a name entry that doesn’t exist, and tries to convert that NoneType to Unicode. I think this happens because the name handling is trying to update names across each platform, but doesn’t account for some platforms having only a subset of entries.

This solution first tries the previous approach, but if an `AttributeError` is thrown, it then goes and checks for that entry in other platforms, and carries on once it finds a value.

In my current use case, this seems to work well. However, my understanding of the full picture here is somewhat limited, so I may be missing something. I’m happy for any feedback or suggestions!

Thanks.

UPDATE:

Oh no... this doesn’t quite work yet. I’m getting a doubling of style names. Not sure if that is from this adjustment, or from something else. Will adjust this if I find the solution.

### DIFF
--- a/Lib/fontTools/varLib/instancer/names.py
+++ b/Lib/fontTools/varLib/instancer/names.py
@@ -221,9 +221,18 @@ def _updateNameRecords(varfont, axisValues):
             getName(n, *platform).toUnicode() for n in ribbiNameIDs
         )
         if nonRibbiNameIDs:
-            typoSubFamilyName = " ".join(
-                getName(n, *platform).toUnicode() for n in axisValueNameIDs
-            )
+            try:
+                typoSubFamilyName = " ".join(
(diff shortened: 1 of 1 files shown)


arrowtype:
I’m abandoning this, at least for now, as it proved a bit more complex than anticipated, and I just had to get the work done in another way. I’ll close this PR.

anthrotype:
it's still worth filing an issue, it's not good we crash while instancer is updating the name table if some aren't present in all platforms

PR #3767, opened 2025-02-06
[varStore] use the same sorting for both input/output

anthrotype:
When the VarStore.optimize() algorithm was improved with the addition of a priority queue, specifically with this commit https://github.com/fonttools/fonttools/commit/47ec18f788, a pre-sorting step of the todo list (before the queue itself was populated) was kept to make sure the algorithm continued to produce stable results no matter the order of the inputs.
The optimizer's output is itself sorted, again for stability, but using a different key function with_sort_key() from the one used on the input todo list gain_sort_key().
The rationale for a distinct gain_sort_key, where encodings get sorted by maximum theoretical gain (decreasing, initally, when reverse=True was set, then incrasing as reverse was somehow dropped), is no longer needed now that a priority queue is used (this sorts by actual gains from merging specific pairs of encodings) and as Behdad confirmedf in chat, is probably just a remnant of the previous algorithm.

I propose we do keep some pre-sorting to ensure stability (in case the priority queue initially contains multiple pairs with the same gain), but we reuse the same width_sort_key that is used at the end to sort the output.

Note this doesn't change the overall level of compression achieved by the optimizer, but makes the algorithm a bit less complicated, and easier to match in our alternative Rust implementation.

### DIFF
--- a/Lib/fontTools/varLib/varStore.py
+++ b/Lib/fontTools/varLib/varStore.py
@@ -412,25 +412,6 @@ class _Encoding(object):
     def extend(self, lst):
         self.items.update(lst)
 
-    def get_room(self):
-        """Maximum number of bytes that can be added to characteristic
-        while still being beneficial to merge it into another one."""
-        count = len(self.items)
-        return max(0, (self.overhead - 1) // count - self.width)
-
-    room = property(get_room)
-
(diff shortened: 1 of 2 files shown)


behdad:
LGTM!

PR #3770, opened 2025-02-14
[feaLib] More efficient inline multisub compilation

### LINKED ISSUE #3769: [feaLib] more suboptimal compilation of inline multiple sub rules?
See #3551 for context.

I'm running into something where it looks like feaLib is missing an optimization?

Given the following FEA,

```fea
@abc = [a-c];
@def = [d-f];

feature psts {
    lookup test {
        sub @abc' @def @abc by NULL;
    } test;
} psts;
```

Although feaLib is reusing the contextual lookup, it is creating a new 'rule' for each item in the input class (https://github.com/fonttools/fonttools/blob/8a89f4f81b0068d7dabf7b421587787a49d3737c/Lib/fontTools/feaLib/builder.py#L1395) even though the items should be able to all share the same rule, since they all have the same target.

I think there's a reasonable solution, here, which is to look for a reusable rule in the same way that we look for a reusable lookup, and I'm going to see if this is trickier than it sounds.

cmyr:
Specifically we were always creating a new ChainContextualRule for each glyph in an input class, even when their replacements were identical, which means we would later miss the chance to use a format3 encoding, which would have been smaller.

The implementation here feels... fine, I guess? In fea-rs we don't decompose the input set, but we also don't support the syntax,

`sub a [b c]' d by e f` which is supported by fontmake (but has not yet appeared in any font we've handled) which might motivate the decomposition.

- fixes #3769


Lib/fontTools/otlLib/builder.py
```diff
@@ -786,6 +786,39 @@ def getAlternateGlyphs(self):
...
+            lookup = rule.lookups[0]
+            # if the replacements already exist for another glyph, add us?
+            if (
+                any(replacements == value for value in lookup.mapping.values())
+                and lookup.mapping.get(glyph, replacements) == replacements
+            ):
+                lookup.mapping[glyph] = replacements
+                rule.glyphs[0].add(glyph)
+
+                print("reused a mapping")
```
anthrotype, line 818: remove stray print - or if you like to keep, turn it into a log.debug perhaps

cmyr: oops!

cmyr:
huh let's hold off on this for now, I'm seeing some regressions on crater that I will need to investigate..

PR #3771, opened 2025-02-14
add --no-recalc-timestamp to fonttools ttLib CLI

anthrotype:
does the same as ttx homonymous option


behdad:
How do you like the tool?

anthrotype:
I like to use it when I need to do two ttx commands back to back

PR #3776, opened 2025-02-20
[`ttLib.__main__`] improvements

behdad:


behdad:
Humm. This has the side-effect of loading all tables by default now. I find it useful though. Thoughts?

Lib/fontTools/ttLib/__main__.py
```diff
@@ -108,10 +108,16 @@ def main(args=None):
...
 
-    if lazy is False:
-        for font in fonts:
-            for table in tables if "*" not in tables else font.keys():
-                font[table]  # Decompiles
+    if tables is None:
+        if lazy is True:
+            tables = []
+        else:
+            tables = ["*"]
```
anthrotype, line 115: I think lazy=None by default, so if you want to avoid decompile all tables by default, when neither `-t '*' or `--no-lazy` options are passed, then you could do:

```suggestion
        elif lazy is False:
            tables = ["*"]
```

behdad: What to set in the case of lazy None? I need to set *something*.

anthrotype: Empty list?

behdad: Yes, that's what my latest change does. Thanks.

PR #3777, opened 2025-02-22
[sfnt] Use memoryview instead of bytes for loading tables

behdad:
And some adjustments. Still ~60 tests fail. Not sure this is worth it. Just leaving here for posterity.


behdad:
This, and https://github.com/fonttools/fonttools/pull/581 if implemented, will significantly reduce font loading times...

behdad:
I'm actually not sure about this resulting in any meaningful speedup.

Issue #3778, opened 2025-02-24
Add HVAR synthesis


yanone:
As discussed with @behdad in a separate chat, please add a tool into fonttools that can synthesize a missing `HVAR` table from existing `glyf`&`gvar` tables.

Behdad said it shouldn't be too complicated to make.

This is needed for fonts exported directly out of Glyphs.app as Georg doesn't want to implement `HVAR` as he is of the position that `glyf`&`gvar` is sufficient despite being aware of the rendering speed impairment.

Thank you.

yisibl:
I've carried @behdad's post here for more background.

> ## FontTools `varLib.hvar` module

> The HVAR table in a variable font is essentially a cache of vertical glyph advance widths. A variable font will perform functionally correctly without it, but with (vastly) slowed-down shaping. For that reason, the Glyphs app authors see it unnecessary to generate an HVAR table in the exported fonts.
To work around the Glyphs lack of this export feature, some designers have turned to surgically adding an HVAR table to their fonts in post-processing. The HVAR will be generated by exporting a font binary using fontmake with the Glyphs font source, then the TTX appended to the font exported natively by Glyphs app. This is a tedious and completely unnecessary burden.
While I cannot convince Georg to implement HVAR export in Glyphs, I could help desperate designers by creating a tool to automate the HVAR generation and addition to binary fonts. That is exactly what the new fonttools varLib.hvar command does. This will be in the next FontTools release.
> https://docs.google.com/document/d/1RUZ0QDwSQVxlz1-pYJdUan7hoRzmoLz2cVmtCHUoqUI/view?tab=t.0#heading=h.3si0s2hgh6c7


Georg from the Glyphs.app team said:

> There can be a tiny performance advantage with an HVAR table. But it usually contains value for the width (this is the useful part) but also values for the LSB. Relying on the variable LSB is bad and can lead to quite bad spacing (I can explain that in more detail if needed). So I’m reluctant to add the HVAR table.
> https://forum.glyphsapp.com/t/variable-font-breaks-spacing-on-export-problems-variate-from-software-to-software/8268/23


@behdad In particular, is the LSB issue mentioned here addressed in PR #3780?

> Relying on the variable LSB is bad and can lead to quite bad spacing

We used to add the HVAR table to our fonts, but did run into some problems with it, so we've since removed it.
If FontTools solves the problem well, I'm happy to add it again.

behdad:
> Georg from the Glyphs.app team said:
> 
[...]

PR #3780, opened 2025-02-24
Add `varLib.hvar` module

behdad:
Fixes https://github.com/fonttools/fonttools/issues/3778

Mildly tested. In particular, on RobotoFlex VF, it produces smaller HVAR table. I have not investigated. The code is heavily lifted from `varLib.__init__`.


Lib/fontTools/varLib/hvar.py
```diff
@@ -0,0 +1,167 @@
...
+            assert len(phantoms) == 2
+            phantoms[0] = phantoms[0][phantomIndex] if phantoms[0] is not None else 0
+            phantoms[1] = phantoms[1][phantomIndex] if phantoms[1] is not None else 0
+            deltas.append(phantoms[1] - phantoms[0])
+
+        allDeltas[glyphName] = deltas[1:]
+        storeBuilder.setSupports(supports)
+        advMapping[glyphName] = storeBuilder.storeDeltas(deltas, round=noRound)
+
+    singleModel = all(advMapping[g] >> 16 == 0 for g in glyphOrder)
```
anthrotype, line 63: shouldn't this check that all the supports are the same (like we do in the varLib _get_advance_metrics)?

Lib/fontTools/varLib/hvar.py
```diff
@@ -0,0 +1,167 @@
...
+        for tv in variations:
+            supports.append(tv.axes)
+            phantoms = tv.coordinates[-4:]
+            phantoms = phantoms[phantomIndex * 2 : phantomIndex * 2 + 2]
+            assert len(phantoms) == 2
+            phantoms[0] = phantoms[0][phantomIndex] if phantoms[0] is not None else 0
+            phantoms[1] = phantoms[1][phantomIndex] if phantoms[1] is not None else 0
+            deltas.append(phantoms[1] - phantoms[0])
+
+        allDeltas[glyphName] = deltas[1:]
```
anthrotype, line 59: why skip the first one? an inline comment might help the reader

anthrotype:
I wish we could reuse more of the code that's already in varLib.__init__ (e.g. `_get_advance_metrics`) instead of maintaining two very similar copies

Lib/fontTools/varLib/hvar.py
behdad: That's one way. My reasoning here was that if the OnlineVarStoreBuilder put them all in the same group (ie. all major's match), then we can do this. I'll try to revert to other logic.

Lib/fontTools/varLib/hvar.py
behdad: First "delta" is the actual advance itself.

Lib/fontTools/varLib/hvar.py
```diff
@@ -0,0 +1,167 @@
...
+        storeBuilder.setSupports(supports)
+        advMapping[glyphName] = storeBuilder.storeDeltas(deltas, round=noRound)
+
+    singleModel = all(advMapping[g] >> 16 == 0 for g in glyphOrder)
+
+    directStore = None
+    if singleModel:
+        # Build direct mapping
+        supports = supports[1:]
```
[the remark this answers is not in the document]
behdad: Ah. I think I messed up both of these. I'll fix. I used to have {} as first item, but I think I removed it.

behdad:
> I wish we could reuse more of the code that's already in `varLib.__init__` (e.g. `_get_advance_metrics`) instead of maintaining two very similar copies

Me too. I'll work on it.

behdad:
I addressed reviews.

One thing occurs to me that in a post-64k world, we should be more careful building the direct mapping since not all will fit in a VarData.

I'll see about more code sharing. And needs some kinds of tests.

behdad:
I haven't found a font yet that this fully roundtrips to the same XML...

behdad:
> I haven't found a font yet that this fully roundtrips to the same XML...

We changed the sort algorithm a bit. Maybe it's that.

behdad:
> > I haven't found a font yet that this fully roundtrips to the same XML...
> 
> We changed the sort algorithm a bit. Maybe it's that.

Oh it's probably because we don't go through the VariationModel, just directly to deltas. Anyway, I've convinced myself that the code is correct. Now onto refactoring.

behdad:
VORG is not done in this module. VVAR is done but untested.

Lib/fontTools/varLib/__init__.py
```diff
@@ -657,10 +600,71 @@ def _get_advance_metrics(
...
+                else None
+            )
+            for metrics in advMetricses
+        ]
+        vhAdvanceDeltasAndSupports[glyph] = masterModel.getDeltasAndSupports(
+            vhAdvances, round=round
+        )
+
+    if vOrigMetricses:
+        singleModel = False
```
anthrotype, line 654: local variable no longer used within this scope

anthrotype:
> I haven't found a font yet that this fully roundtrips to the same XML...

maybe because we end up ordering the VarRegions differently in the VarRegionList?

> And needs some kinds of tests.

yes, ideally

behdad:
> > I haven't found a font yet that this fully roundtrips to the same XML...
> 
> maybe because we end up ordering the VarRegions differently in the VarRegionList?

I built a fresh NotoSansArabic, and HVAR roundtrips perfectly.

Issue #3779, opened 2025-02-24
Debg table not supported for ttx roundtrip


n7s:
I noticed this when running the FontBakery ttf_roundtrip check, the Degb table is not fully supported: 

```
ERROR: Unhandled exception has occurred
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/dist-packages/fontTools/ttx.py", line 452, in main
    process(jobs, options)
  File "/usr/local/lib/python3.12/dist-packages/fontTools/ttx.py", line 434, in process
    action(input, output, options)
  File "/usr/local/lib/python3.12/dist-packages/fontTools/misc/loggingTools.py", line 375, in wrapper
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/fontTools/ttx.py", line 334, in ttCompile
    ttf.save(output)
  File "/usr/local/lib/python3.12/dist-packages/fontTools/ttLib/ttFont.py", line 226, in save
    writer_reordersTables = self._save(tmp)
                            ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/fontTools/ttLib/ttFont.py", line 273, in _save
    self._writeTable(tag, writer, done, tableCache)
  File "/usr/local/lib/python3.12/dist-packages/fontTools/ttLib/ttFont.py", line 721, in _writeTable
    tabledata = self.getTableData(tag)
                ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/fontTools/ttLib/ttFont.py", line 743, in getTableData
    return self.tables[tag].compile(self)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/fontTools/ttLib/tables/D__e_b_g.py", line 11, in compile
    return json.dumps(self.data).encode("utf-8")
                      ^^^^^^^^^
AttributeError: 'table_D__e_b_g' object has no attribute 'data'

```

please consider providing support for this table. Thanks.

anthrotype:
this is strange, how was the Debg table created to begin with? It should have a `data` attribute if it was decompiled from binary font, or if it was parsed from a ttx file.

From the traceback, it seems you are running ttx command to compile a .ttx file to .ttf, I don't see why `self.data` should not be set in this case.

[...]

PR #3781, opened 2025-02-25
[Debg] fix parsing from XML; add roundtrip tests

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/3779

note this changes the ttx dump of 'Debg' table in that it wraps the CDATA block inside a `<json>` sub-element, which is required for the XMLReader to be able to parse it back from TTX.

Since parsing was silently broken until now, we shouldn't worry about backward compatibility but treat this as a bug fix.

It's also a private table which shouldn't be in production fonts so we are probably free to change as needed.


anthrotype:
note, API or the binary serialization has not changed, only the xml serialization

PR #3783, opened 2025-02-26
[fealib] Support *Base.MinMax tables

simoncozens:
Nobody told me this was missing!

### DIFF
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -1306,6 +1306,20 @@ class Parser(object):
                         location=self.cur_token_location_,
                     )
                 )
+            elif self.is_cur_keyword_("HorizAxis.MinMax"):
+                if len(statements) < 1 or not isinstance(
+                    statements[-1], self.ast.BaseAxis
+                ):
+                    raise FeatureLibError(
+                        "MinMax must be preceded by BaseScriptList",
+                        self.cur_token_location_,
+                    )
+                if statements[-1].vertical:
+                    raise FeatureLibError(
+                        "HorizAxis.MinMax must be preceded by HorizAxis statements",
+                        self.cur_token_location_,
+                    )
+                statements[-1].minmax.append(self.parse_base_minmax_())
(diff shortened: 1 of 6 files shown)


anthrotype:
nice, thank you. So to confirm, this currently supports BaseCoord format 1 (no contour point nor device), only minmax for scripts/languages but not for specific features; and only static minmax values, no variations (I don't think we support variable BASE at all).

anthrotype (approved):
LGTM

anthrotype:
i wonder (not that we should block on this), does makeotf support compiling BASE MinMax statements?

simoncozens:
>  this currently supports BaseCoord format 1 (no contour point nor device), only minmax for scripts/languages but not for specific features; and only static minmax values, no variations (I don't think we support variable BASE at all).

Correct, yes.

> does makeotf support compiling BASE MinMax statements?

No, it doesn't.

Lib/fontTools/feaLib/parser.py
```diff
@@ -1306,6 +1306,20 @@ def parse_table_BASE_(self, table):
                         location=self.cur_token_location_,
                     )
                 )
+            elif self.is_cur_keyword_("HorizAxis.MinMax"):
+                if len(statements) < 1 or not isinstance(
+                    statements[-1], self.ast.BaseAxis
```
belluzj, line 1311: I was testing this and had comments in my feature file, like this:

```fea
   HorizAxis.BaseTagList romn;
   HorizAxis.BaseScriptList latn romn 0;
   HorizAxis.MinMax latn dflt -800, 1200;
   # Some comment
   HorizAxis.MinMax latn VIT  -800, 2000;
```
      
    

      
    

    
  
and I got the error "MinMax must be preceded by BaseScriptList". I guess it's because of the -1 lookup, probably it finds the comment instead of the BaseScriptList? Do we want to support comments in this place? If so I can do a PR. For now I just moved my comment out of the way.

belluzj: I'll make a small PR with a test and bugfix

PR #3784, opened 2025-02-27
[config] Add OPTIMIZE_FONT_SPEED

behdad:
Use it for compiling faster-to-load glyf/gvar/VARC.

### DIFF
--- a/Lib/fontTools/config/__init__.py
+++ b/Lib/fontTools/config/__init__.py
@@ -73,3 +73,18 @@ Config.register_option(
     parse=Option.parse_optional_bool,
     validate=Option.validate_optional_bool,
 )
+
+Config.register_option(
+    name="fontTools.ttLib:OPTIMIZE_FONT_SPEED",
+    help=dedent(
+        """\
+        Enable optimizations that prioritize speed over file size. This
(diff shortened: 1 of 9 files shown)


anthrotype:
out of curiosity, I tried to compile a glyf+gvar heavy font like Google Sans Flex with OPTIMIZE_FONT_SPEED=True, and the resulting ttf is 1,065,644 bytes bigger (4,823,984 vs 3,758,340) as expected, but interestingly when compressing both fonts with WOFF2, the speed-optimized font is actually 41,932 bytes _smaller_ than the normal font which is supposed to be the size-optimized one.

anthrotype:
how is VARC compilation specifically affected by this?

I see `VarComponent.compile` calls `TupleVariation.compileDeltaValues_` but I don't see the new flag being used in there:
https://github.com/fonttools/fonttools/blob/c93b45bc3a4fb6d61a45976fee8f692457619e0a/Lib/fontTools/ttLib/tables/otTables.py#L261C32-L261C51

anthrotype (approved):

behdad:
> how is VARC compilation specifically affected by this?

Mostly by change in TupleValues compilation. But I fixed the one you pointed out either.

PR #3788, opened 2025-03-08
[feaLib] Correct ChainContextPosStatement and ChainContextSubstStatement

lianghai:
They don’t really have `self.glyphs`. The erroneous `self.glyph` has been hiding in a branch of edge case (no prefix, no suffix, and no nested lookups).

It’s unclear if such an edge case should be serialized as a special case (without the `'` mark), therefore I’m only proposing a minimal fix for now.

### DIFF
--- a/Lib/fontTools/feaLib/ast.py
+++ b/Lib/fontTools/feaLib/ast.py
@@ -753,7 +753,7 @@ class ChainContextPosStatement(Statement):
             if len(self.suffix):
                 res += " " + " ".join(map(asFea, self.suffix))
         else:
-            res += " ".join(map(asFea, self.glyph))
+            res += " ".join(map(asFea, self.glyphs))
         res += ";"
         return res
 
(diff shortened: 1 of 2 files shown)


khaledhosny:
Since this code is not covered by any tests, can you add some?

khaledhosny (approved):

lianghai:
Uh, I don’t really understand the intended behavior of this part of code (this PR just fixed some apparent typos), and https://github.com/fonttools/fonttools/tree/main/Tests/feaLib seems rather complicated. I’ll leave the test case to you guys…

khaledhosny:
I thought you hit this with some feature code, but it is alright if you don’t have a test. I’m also not sure about this special case.

lianghai:
Oh I hit this when I generated FEA code from the AST. I accidentally created a ChainContextSubstStatement instance with “no prefix, no suffix, and no nested lookups”. Parsing from a sub rule without the `'` mark (as suggested in the `.asFea()` code I touched) probably won’t ever end up with a ChainContextSubstStatement…

khaledhosny:
> Oh I hit this when I generated FEA code from the AST. I accidentally created a ChainContextSubstStatement instance with “no prefix, no suffix, and no nested lookups”. Parsing from a sub rule without the `'` mark (as suggested in the `.asFea()` code I touched) probably won’t ever end up with a ChainContextSubstStatement…

Yes, I guessed it would be something like this. Too niche for a test.

PR #3798, opened 2025-04-09
[ttLib] Apply rounding more often in getCoordinates

cmyr:
Regardless of whether or not a component transform is simple, we still have to round the points before returning, or we risk the error propogating up to another component that might have a non-trivial transform.


behdad:
Sounds about right. I had done that for optimization purposes I believe...

anthrotype:
LGTM thanks!

PR #3799, opened 2025-04-09
[ttLib] Ignore component bounds if empty

cmyr:
That is, if the bounds of a component have zero area, do not consider it when computing the bounds of the parent.

This handles the case of a composite glyph where all the components are themselves empty, which if it sounds weird you're probably new around here.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -1225,7 +1225,7 @@ class Glyph(object):
                 if boundsDone is not None:
                     boundsDone.add(glyphName)
             # empty components shouldn't update the bounds of the parent glyph
-            if g.numberOfContours == 0:
+            if g.yMin == g.yMax and g.xMin == g.xMax:
                 continue
 
             x, y = compo.x, compo.y
--- a/Tests/ttLib/tables/_g_l_y_f_test.py
+++ b/Tests/ttLib/tables/_g_l_y_f_test.py
@@ -567,17 +567,75 @@ class GlyphTest:
         pen = TTGlyphPen(glyphSet)
         # empty simple glyph
         foo = glyphSet["foo"] = pen.glyph()
+
+        pen.moveTo((5, 5))
+        pen.lineTo((10, 10))
+        pen.lineTo((10, 5))
+        pen.lineTo((5, 5))
+        pen.closePath()
+
+        # non-empty simple glyph
+        doo = glyphSet["doo"] = pen.glyph()
+
+        pen = TTGlyphPen(glyphSet)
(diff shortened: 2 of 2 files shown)


behdad:
@qxliu76 Is this something HB needs fixing as well?

qxliu76:
do we have a testcase that shows the problem?

qxliu76:
ah I found this one: https://github.com/harfbuzz/harfbuzz/pull/4111, we might need a fix in HB after this PR is merged?

anthrotype:
good catch! thanks

anthrotype:
@qxliu76 @behdad I have just distilled a test font from @cmyr's unit test, you'll find it attached below:

[RecalcBoundsEmptyComponentsNested-Regular.ttf.zip](https://github.com/user-attachments/files/19851038/RecalcBoundsEmptyComponentsNested-Regular.ttf.zip)

If you rountrip the font through ttx (with this PR merged on main), you'll notice that the bounds of glyph "bar" have been updated (i.e. inheriting the bounds of its non-empty component, ignoring the empty one)  

```diff
ttx -q -o - "/Users/clupo/oss/fonttools/master_ttf/RecalcBoundsEmptyComponentsNested-Regular.ttf" 2>/dev/null
ttx -q -o - "/Users/clupo/oss/fonttools/master_ttf/RecalcBoundsEmptyComponentsNested-Regular#1.ttf" 2>/dev/null
--- /dev/fd/63  2025-04-22 15:01:11
+++ /dev/fd/62  2025-04-22 15:01:11
@@ -14,12 +14,12 @@
     <!-- Most of this table will be recalculated by the compiler -->
     <tableVersion value="1.0"/>
     <fontRevision value="1.0"/>
-    <checkSumAdjustment value="0xf5e5411a"/>
+    <checkSumAdjustment value="0xf5df4094"/>
     <magicNumber value="0x5f0f3cf5"/>
-    <flags value="00000000 00000011"/>
+    <flags value="00000000 00000001"/>
     <unitsPerEm value="1000"/>
     <created value="Tue Apr 22 13:50:52 2025"/>
-    <modified value="Tue Apr 22 13:59:53 2025"/>
+    <modified value="Tue Apr 22 14:00:55 2025"/>
     <xMin value="0"/>
     <yMin value="-200"/>
     <xMax value="450"/>
@@ -166,7 +166,7 @@
       <instructions/>
     </TTGlyph>

-    <TTGlyph name="bar" xMin="0" yMin="0" xMax="10" yMax="10">
+    <TTGlyph name="bar" xMin="5" yMin="5" xMax="10" yMax="10">
       <component glyphName="foofoo" x="0" y="0" flags="0x4"/>
       <component glyphName="doo" x="0" y="0" flags="0x4"/>
     </TTGlyph>
```

PR #3800, opened 2025-04-11
[feaLib] Support subtable statements in more lookups

khaledhosny:
Add support for subtable statements to cursive and mark to base lookups.


behdad:
I don't feel qualified to review this. Simon maybe?

khaledhosny:
I have convinced myself this is a low risk, uncontroversial change, so I’m going to merge it.

PR #3801, opened 2025-04-11
[subset] close unicodes over bidi mirror variants.

garretrieger:
Matching a change to harfbuzz, for context: https://github.com/harfbuzz/harfbuzz/pull/5282 and https://github.com/harfbuzz/harfbuzz/issues/5281

Draft as it still needs tests and we need to sort out the dep on python-bidi.


Lib/fontTools/subset/__init__.py
```diff
@@ -2870,6 +2872,15 @@ def prune_post_subset(self, font, options):
 def closure_glyphs(self, s):
     tables = [t for t in self.tables if t.isUnicode()]
 
+    # Closure unicodes, which for now is pulling in bidi mirrored variants
+    if s.options.bidi_closure:
+        additional_unicodes = set()
+        for u in s.unicodes_requested:
+            if mirrored(chr(u)):
```
behdad, line 2879: Faster to forgo this step I think.

behdad: Just `MIRRORED.get(chr(u))` would return None.

behdad:
LGTM. Let's see if Cosimo has any input.

Lib/fontTools/unicodedata/__init__.py
```diff
@@ -45,6 +44,439 @@
     "ot_tag_to_script",
 ]
 
+# Derived from http://www.unicode.org/Public/UNIDATA/BidiMirroring.txt
```
behdad, line 47: Please put this in it's own file, like `Scripts.py` and import it like that one. Thanks.

Lib/fontTools/unicodedata/__init__.py
```diff
@@ -46,6 +45,10 @@
 ]
 
 
+def mirrored(code):
```
khaledhosny, line 48: Needs a doc string here. LGTM otherwise.

khaledhosny (approved):

PR #3804, opened 2025-04-19
[feaLib] Remove duplicates from class pair pos classes

khaledhosny:
AFDKO’s makeotf removes the duplicates and produces the following message:

    NOTE: [line     3 char  29] Removing duplicate glyph <A>
    NOTE: [line     3 char  29] Removing duplicate glyph <A>

While feaLib fails with:

      File "/fonttools/Lib/fontTools/feaLib/builder.py", line 875, in buildLookups_
        otLookups.append(l.build())
                         ^^^^^^^^^
      File "/fonttools/Lib/fontTools/otlLib/builder.py", line 1460, in build
        builder.addPair(glyphclass1, value1, glyphclass2, value2)
      File "/fonttools/Lib/fontTools/otlLib/builder.py", line 1340, in addPair
        self.classDef1_.add(gc1)
      File "/fonttools/Lib/fontTools/otlLib/builder.py", line 2656, in add
        raise OpenTypeLibError(
    fontTools.otlLib.error.OpenTypeLibError: Glyph A is already present in class.

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -1471,7 +1471,9 @@ class Builder(object):
         lookup = self.get_lookup_(location, PairPosBuilder)
         v1 = self.makeOpenTypeValueRecord(location, value1, pairPosContext=True)
         v2 = self.makeOpenTypeValueRecord(location, value2, pairPosContext=True)
-        lookup.addClassPair(location, glyphclass1, v1, glyphclass2, v2)
+        cls1 = tuple(sorted(set(glyphclass1)))
+        cls2 = tuple(sorted(set(glyphclass2)))
+        lookup.addClassPair(location, cls1, v1, cls2, v2)
 
     def add_specific_pair_pos(self, location, glyph1, value1, glyph2, value2):
         if not glyph1 or not glyph2:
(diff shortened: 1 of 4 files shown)


Lib/fontTools/feaLib/builder.py
```diff
@@ -1471,7 +1471,9 @@ def add_class_pair_pos(self, location, glyphclass1, value1, glyphclass2, value2)
         lookup = self.get_lookup_(location, PairPosBuilder)
         v1 = self.makeOpenTypeValueRecord(location, value1, pairPosContext=True)
         v2 = self.makeOpenTypeValueRecord(location, value2, pairPosContext=True)
-        lookup.addClassPair(location, glyphclass1, v1, glyphclass2, v2)
+        cls1 = tuple(sorted(set(glyphclass1)))
```
anthrotype, line 1474: these glyphclass1/glyphclass2 are they supposed to be of type `set` or `tuple`? The docstring of `addClassPair` method says "set"

anthrotype (approved):
LGTM (but see comment/question)

Lib/fontTools/feaLib/builder.py
khaledhosny: If I use set, I get an error down the line that it it is unhashable type. They are used as a dict key.

PR #3805, opened 2025-04-19
[feaLib] Upgrade single substitutions mixed with ligature ones

khaledhosny:
If a lookup has only single and ligature substitutions, we now upgrade single substitutions to ligature ones. Previously this would cause an error in named lookup, and would result in splitting lookups in anonymous ones.

This is similar to what we have done with multiple substitutions in https://github.com/fonttools/fonttools/pull/1133, but back then I incorrectly convinced myself that ligature substitutions can’t have a single source glyph, which is obviously not true.


behdad:
> I incorrectly convinced myself that ligature substitutions can’t have a single source glyph, which is obviously not true.

We even have code in HB to treat them as single substitutions not ligatures, for further processing purposes.

anthrotype (approved):

PR #3809, opened 2025-04-22
[ttLib] Change the separator for duplicate names from "#" to "."

khaledhosny:
The default behaviour of suffixing them with "#" + a number can sometimes be problematic, e.g. it is not valid if feature files ("#" starts a comment).


justvanrossum (commented):
I was confused by the option name, as I don’t see anything being dropped, just mapped differently.

Wouldn’t it be better to fix the default behavior so it emits a more compatible glyph name? Perhaps leaving out the # is enough?

khaledhosny:
> I was confused by the option name, as I don’t see anything being dropped, just mapped differently.

It is confusing indeed. The idea is that a glyph with a duplicate name is now treated as a glyph with no name at all.

> Wouldn’t it be better to fix the default behavior so it emits a more compatible glyph name? Perhaps leaving out the # is enough?

If that is acceptable change it would be even better. I was afraid that changing such an ancient FontTools convention might be problematic.

justvanrossum:
(Or replace # with a dot.)

I think this is not a common thing, and we generally expect the post table to contain unique names. While I may be missing something, I can’t imagine much breakage over such a change.

khaledhosny:
One complication with using a common separator like a dot, is that now if the font has `["A", "A", "A.1"]` names, we will rename the second glyph to `A.1`, and then the third glyph will be seen as duplicate and gets renamed to `A.1.1`. I guess I’ll have to re-work the existing code a bit to avoid this.

khaledhosny:
cffLib has similar code, so I applied the same change there as well.

justvanrossum (approved):

Lib/fontTools/ttLib/tables/_p_o_s_t.py
```diff
@@ -122,13 +122,16 @@ def build_psNameMapping(self, ttFont):
             glyphName = psName = self.glyphOrder[i]
             if glyphName == "":
                 glyphName = "glyph%.5d" % i
+
             if glyphName in allNames:
                 # make up a new glyphName that's unique
                 n = allNames[glyphName]
-                while (glyphName + "#" + str(n)) in allNames:
+                # check if the exists in any of the seen names or later ones
+                names = set(allNames.keys()) | set(self.glyphOrder)
```
jenskutilek, line 130: Now we are building and unioning those sets for each duplicate glyph, which makes the method slow when there are many identical names. This is the case in the font in #3921.

I'm struggling to imagine a case where we would need to take `allNames` into account. Wouldn't it be enough to build a set from `self.glyphOrder` and check against that? See #3923 for my suggestion.

PR #3811, opened 2025-04-24
[feaLib] Support creating extension lookups

khaledhosny:
Respect `useExtension` keyword instead of silently ignoring it.


khaledhosny:
I'm working on a VOLT compiler that goes VOLT -> FEA -> binary, see the volt-build branch.

I initially thought it would be easier than compiling to binary directly since we already had voltToFea, but I keep hitting these little issues.

Anyway, this still useful in case one wants to convert from VOLT to FEA permanently and make it the canonical source.

khaledhosny:
This still does not handle implicitly created lookups.

PR #3813, opened 2025-04-27
fix Read supplement encoding

shimondoodkin:


behdad:
Thanks! Can you please run black on the code to fix the lint check?

I haven't checked the spec, but LGTM.

shimondoodkin:
@behdad  I did

PR #3815, opened 2025-04-29
Minor housekeeping

madig:
* Fix a warning about license descriptors when building packages
* Build docs through a tox command for easier management (`tox -e docs`)
* Remove Makefile. Was anyone using it? @behdad?
* Use uv  to build packages
* Declare that black should use Python 3.9 as the minimum version.
* Remove unused tox commands `bdist` and `pypi`. The former is faster done by `uv build`, the latter was out of date, given that it was referencing .zip files instead of .tar.gz.

A future housekeeping PR could use uv more thoroughly, i.e. through tox-uv, using the caching mechanisms described in https://docs.astral.sh/uv/guides/integration/github/#caching.

Note that while tox will use uv under the hood, you don't have to use it directly. Unless you want to build packages locally, then you can use `uv build` or any other established build tool like https://pypi.org/project/build/.


behdad:
`Makefile` change LGTM.

Issue #3816, opened 2025-04-29
Make Cythonized build nicer?


madig:
Some time has passed, maybe there's a nicer way to get optionally Cython-ized fontTools packages than a sprawling setup.py and unmanaged requirements.txt files. A quick google turned up this: https://github.com/aio-libs/yarl/blob/master/packaging/README.md.

PR #3817, opened 2025-04-30
Use pyproject.toml and PEP 517 build packend

madig:
My attempt at https://github.com/fonttools/fonttools/issues/3816 :) 

TODO:
- [ ] Would need to switch tox over to tox-uv maybe.
- [ ] Find a replacement for the code that updates the Changelog on release.


benkiel:
Very interested in this outcome, need to do this for all the robotools repositories

madig:
Can't make use of https://github.com/tox-dev/tox/blob/4a8e50e2cdd420d0b058b747092dbfc72192db51/tests/tox_env/python/virtual_env/package/test_package_pyproject.py#L362 in tox.ini to get rid of the env var completely, the setting never makes it into the custom backend?!

madig:
Apparently tox has fixed https://github.com/tox-dev/tox/issues/3125, which would hopefully help to make passing options more idiomatic/simpler.

PR #3818, opened 2025-05-04
Fixes to voltLib and VoltToFea

khaledhosny:
This PR add some missing functionality and fixes to voltLib and VoltToFea, making the conversion to feature files more robust. Together with the recent feaLib changes, this allowed me to compile a large corpus of VOLT sources (which I can not share). The PR also adds a script that uses voltLib and feaLib to compile VOLT sources directly (doing an intermediate fea conversion internally) to be used as `fonttools voltLib` from command line.

### DIFF
--- a/Lib/fontTools/voltLib/__main__.py
+++ b/Lib/fontTools/voltLib/__main__.py
@@ -1,0 +1,206 @@
+import argparse
+import logging
+import sys
+from io import StringIO
+from pathlib import Path
+
+from fontTools import configLogger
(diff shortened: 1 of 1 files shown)


khaledhosny:
Python 3.8 does not have graphlib module, I tried adding `graphlib-backport==1.1.0; python_version <= '3.8'` to `requirements.txt`, but we run the tests once with requirements and once without, so that one is failing on 3.8.

Suggestions are welcome. 3.8 is EOL’d already, if dropping support for it is an option.

PR #3821, opened 2025-05-05
Make PointPen argument names consistent

jenskutilek:
Has been discussed before in #2919. Consistently use `glyphName`, `transformation` as argument names instead of a mixture (`baseGlyphName`, `glyphName`, `transform`, `transformation`).

Maybe the imminent drop of Python 3.8 support is a good point to make those args consistent.

### DIFF
--- a/Lib/fontTools/pens/pointPen.py
+++ b/Lib/fontTools/pens/pointPen.py
@@ -54,7 +54,7 @@ class AbstractPointPen:
 
     def addComponent(
         self,
-        baseGlyphName: str,
+        glyphName: str,
         transformation: Tuple[float, float, float, float, float, float],
         identifier: Optional[str] = None,
         **kwargs: Any,
@@ -259,10 +259,10 @@ class PointToSegmentPen(BasePointToSegmentPen):
         else:
             pen.endPath()
 
-    def addComponent(self, glyphName, transform, identifier=None, **kwargs):
+    def addComponent(self, glyphName, transformation, identifier=None, **kwargs):
         del identifier  # unused
         del kwargs  # unused
-        self.pen.addComponent(glyphName, transform)
+        self.pen.addComponent(glyphName, transformation)
 
 
 class SegmentToPointPen(AbstractPen):
@@ -337,10 +337,10 @@ class SegmentToPointPen(AbstractPen):
         self._flushContour()
         self.contour = None
 
-    def addComponent(self, glyphName, transform):
+    def addComponent(self, glyphName, transformation):
         if self.contour is not None:
             raise PenError("Components must be added before or after contours")
-        self.pen.addComponent(glyphName, transform)
+        self.pen.addComponent(glyphName, transformation)
 
 
 class GuessSmoothPointPen(AbstractPointPen):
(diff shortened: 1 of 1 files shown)


anthrotype:
I don't recall the old discussion but looking at it now, I am wondering if the change is actually needed. What is this change making argument names consistent with? The base AbstractPointPen originally had `baseGlyphName` instead of `glyphName` and it seems the various sub-classes that you are changing were all duly using the same parameter name. What exactly is wrong with `baseGlyphName`? Is it because the segment pens' addComponent methods have `glyphName`? So what? One could argue that, given the two sets of pens have a different somewhat parallel history (and completely different API anyway), it's fine they have slightly different parameter names as long as they are consistent within each set.

jenskutilek:
`PointToSegmentPen`, `GuessSmoothPointPen`, `ReverseContourPointPen`, all inheriting from `AbstractPointPen`, already used `glyphName` instead of `baseGlyphName` in `addComponent`. I often stumble over this because my type checker complains that the method signatures are different. Sometimes a whole class is marked, see the attached screenshot. Changing the argument name in the base class is the least invasive fix IMHO.

There are also some mixins (e.g. `_DecomposingFilterPenMixin`) which are used in both segment and point pens, so there will be warnings about different argument names in them as well.

The newer `addVarComponent` uses ' glyphName` and `transformation` consistently.

But of course you're right, there is no functional reason to change anything.

Bildschirmfoto 2025-05-06 um 14 19 16

Hoolean:
We hit a real-world issue this week that this PR would solve: [RoundingPointPen calls its target pen with an explicitly named `baseGlyphName`](https://github.com/fonttools/fonttools/blob/e5c9b1e7b04c0f06d52fa5cb42ae6e7f8f13fe6a/Lib/fontTools/pens/roundingPen.py#L119), but [ufoLib2 only accepts `baseName`](https://github.com/fonttools/ufoLib2/blob/440b2dd247ee75095a8b6c2132b471dffb92e8da/src/ufoLib2/pointPens/glyphPointPen.py#L58), so we get a crash.

A workaround is possible, e.g., passing the argument positionally, but it would be nice to achieve proper typing and eliminate this whole class of error forever.

anthrotype:
technically that's a bug in ufoLib2 because the GlyphPointPen subclasses AbstractPointPen (just like RoundingPointPen) so it should really have `baseGlyphName` like its base class..
in any case, the presence of mixins that are used by both APIs is a stronger argument for giving the two sets consistent parameter names. I just hope that we won't make other client code crash suddenly because they happened to use keyword argument with the old name..

anthrotype:
@Hoolean are you sure this PR would actually solve that bug you pointed at? ufoLib2 GlyphPointPen still calls its parameter `baseName` whereas in @jenskutilek's PR the RoundingPointPen is calling it `glyphName` (instead of the previous `baseGlyphName`, in both cases you'd get a TypeError, unless you change ufoLib2 as well..

Lib/fontTools/pens/roundingPen.py
```diff
@@ -113,10 +112,10 @@ def addPoint(
             **kwargs,
         )
 
-    def addComponent(self, baseGlyphName, transformation, identifier=None, **kwargs):
+    def addComponent(self, glyphName, transformation, identifier=None, **kwargs):
         xx, xy, yx, yy, dx, dy = transformation
         self._outPen.addComponent(
-            baseGlyphName=baseGlyphName,
+            glyphName=glyphName,
```
anthrotype, line 118: I suggest you just pass `glyphName` as a positional argument here

knutnergaard:
Commented on this problem in #4037. Basically it's what keeps those pen modules from passing all mypy checks.

PR #3826, opened 2025-05-13
Type hints

schriftgestalt:
Add some type hints.

### DIFF
--- a/Lib/fontTools/ttLib/tables/C_F_F_.py
+++ b/Lib/fontTools/ttLib/tables/C_F_F_.py
@@ -1,6 +1,11 @@
 from io import BytesIO
 from fontTools import cffLib
 from . import DefaultTable
+from typing import Any, Dict, List, Optional, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from fontTools.ttLib import TTFont
+    from fontTools.misc.xmlWriter import XMLWriter
 
 
 class table_C_F_F_(DefaultTable.DefaultTable):
@@ -19,27 +24,30 @@ class table_C_F_F_(DefaultTable.DefaultTable):
     See also https://learn.microsoft.com/en-us/typography/opentype/spec/cff
     """
 
-    def __init__(self, tag=None):
+    cff: cffLib.CFFFontSet
+    _gaveGlyphOrder: bool
+
+    def __init__(self, tag: str | None = None) -> None:
         DefaultTable.DefaultTable.__init__(self, tag)
         self.cff = cffLib.CFFFontSet()
         self._gaveGlyphOrder = False
 
-    def decompile(self, data, otFont):
+    def decompile(self, data: bytes, otFont: 'TTFont') -> None:
         self.cff.decompile(BytesIO(data), otFont, isCFF2=False)
         assert len(self.cff) == 1, "can't deal with multi-font CFF tables."
 
-    def compile(self, otFont):
+    def compile(self, otFont: 'TTFont') -> bytes:
         f = BytesIO()
         self.cff.compile(f, otFont, isCFF2=False)
         return f.getvalue()
 
-    def haveGlyphNames(self):
+    def haveGlyphNames(self) -> bool:
         if hasattr(self.cff[self.cff.fontNames[0]], "ROS"):
             return False  # CID-keyed font
         else:
             return True
 
-    def getGlyphOrder(self):
+    def getGlyphOrder(self) -> List[str]:
         if self._gaveGlyphOrder:
             from fontTools import ttLib
 
@@ -47,15 +55,15 @@ class table_C_F_F_(DefaultTable.DefaultTable):
         self._gaveGlyphOrder = True
         return self.cff[self.cff.fontNames[0]].getGlyphOrder()
 
-    def setGlyphOrder(self, glyphOrder):
+    def setGlyphOrder(self, glyphOrder: List[str]) -> None:
         pass
         # XXX
         # self.cff[self.cff.fontNames[0]].setGlyphOrder(glyphOrder)
 
-    def toXML(self, writer, otFont):
+    def toXML(self, writer: 'XMLWriter', otFont: 'TTFont') -> None:
         self.cff.toXML(writer)
 
-    def fromXML(self, name, attrs, content, otFont):
+    def fromXML(self, name: str, attrs: Dict[str, str], content: List[Any], otFont: 'TTFont') -> None:
         if not hasattr(self, "cff"):
             self.cff = cffLib.CFFFontSet()
         self.cff.fromXML(name, attrs, content, otFont)
--- a/Lib/fontTools/ttLib/tables/DefaultTable.py
+++ b/Lib/fontTools/ttLib/tables/DefaultTable.py
@@ -1,11 +1,13 @@
 from fontTools.misc.textTools import Tag
-from fontTools.ttLib import getClassTag
 
 
 class DefaultTable(object):
     dependencies = []
 
     def __init__(self, tag=None):
+
+        from fontTools.ttLib import getClassTag
+
         if tag is None:
             tag = getClassTag(self.__class__)
         self.tableTag = Tag(tag)
@@ -40,7 +42,7 @@ class DefaultTable(object):
         return "<'%s' table at %x>" % (self.tableTag, id(self))
 
     def __eq__(self, other):
-        if type(self) != type(other):
+        if not isinstance(self, type(other)):
             return NotImplemented
         return self.__dict__ == other.__dict__
 
--- a/Lib/fontTools/ttLib/tables/otConverters.py
+++ b/Lib/fontTools/ttLib/tables/otConverters.py
@@ -691,6 +691,8 @@ class StructWithLength(Struct):
         assert writer.items[lengthIndex] == b"\xde\xad\xbe\xef"[: conv.staticSize]
         writer.items[lengthIndex] = lengthWriter.getAllData()
 
+    def __repr__(self):
+        return "StructWithLength of " + repr(self.tableClass)
 
 class Table(Struct):
     staticSize = 2
@@ -725,6 +727,8 @@ class Table(Struct):
             writer.writeSubTable(subWriter, offsetSize=self.staticSize)
             value.compile(subWriter, font)
 
+    def __repr__(self):
+        return "Table of " + repr(self.tableClass)
 
 class LTable(Table):
     staticSize = 4
@@ -735,6 +739,9 @@ class LTable(Table):
     def writeNullOffset(self, writer):
         writer.writeULong(0)
 
+    def __repr__(self):
+        return "LTable of " + repr(self.tableClass)
+    
 
 # Table pointed to by a 24-bit, 3-byte long offset
 class Table24(Table):
@@ -746,6 +753,8 @@ class Table24(Table):
     def writeNullOffset(self, writer):
         writer.writeUInt24(0)
 
+    def __repr__(self):
+        return "Table24 of " + repr(self.tableClass)
 
 # TODO Clean / merge the SubTable and SubStruct
 
@@ -758,6 +767,9 @@ class SubStruct(Struct):
     def xmlWrite(self, xmlWriter, font, value, name, attrs):
         super(SubStruct, self).xmlWrite(xmlWriter, font, value, None, attrs)
 
+    def __repr__(self):
+        return "SubStruct of " + repr(self.tableClass)
+
 
 class SubTable(Table):
     def getConverter(self, tableType, lookupType):
@@ -767,17 +779,26 @@ class SubTable(Table):
     def xmlWrite(self, xmlWriter, font, value, name, attrs):
         super(SubTable, self).xmlWrite(xmlWriter, font, value, None, attrs)
 
+    def __repr__(self):
+        return "SubTable of " + repr(self.tableClass)
+
 
 class ExtSubTable(LTable, SubTable):
     def write(self, writer, font, tableDict, value, repeatIndex=None):
         writer.Extension = True  # actually, mere presence of the field flags it as an Ext Subtable writer.
         Table.write(self, writer, font, tableDict, value, repeatIndex)
 
+    def __repr__(self):
+        return "ExtSubTable of " + repr(self.tableClass)
+
 
 class FeatureParams(Table):
     def getConverter(self, featureTag):
         tableClass = self.featureParamTypes.get(featureTag, self.defaultFeatureParams)
         return self.__class__(self.name, self.repeat, self.aux, tableClass)
+
+    def __repr__(self):
+        return "FeatureParams of " + repr(self.tableClass)
 
 
 class ValueFormat(IntValue):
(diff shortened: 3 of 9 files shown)


Lib/fontTools/ttLib/tables/DefaultTable.py
```diff
@@ -40,7 +42,7 @@ def __repr__(self):
         return "<'%s' table at %x>" % (self.tableTag, id(self))
 
     def __eq__(self, other):
-        if type(self) != type(other):
+        if not isinstance(self, type(other)):
```
justvanrossum, line 45: This is unrelated to type hints and changes the meaning of the code. The current code is written as intended: only compare when the types match exactly.

Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -691,6 +691,8 @@ def write(self, writer, font, tableDict, value, repeatIndex=None):
         assert writer.items[lengthIndex] == b"\xde\xad\xbe\xef"[: conv.staticSize]
         writer.items[lengthIndex] = lengthWriter.getAllData()
 
+    def __repr__(self):
```
justvanrossum, line 694: I recomment to make a separate PR for changes other than type annotations.

schriftgestalt: You are right. I tried to avoid changes like this …

Lib/fontTools/ttLib/tables/DefaultTable.py
schriftgestalt: You are right. I tried to avoid changes like this. I’ll remove it.

schriftgestalt:
I fixed a few flake issues I missed. But tox is not happy at all when I run it locally. Is there anything I can do or can I ignore issues in files I didn’t touch? I have fixes some easy ones. But that is probably another PR?

jenskutilek:
I think you need to `from __future__ import annotations` for the tests to pass in Python 3.9.

And you should format the code with `black` to get rid of the linter fails.

Lib/fontTools/ttLib/tables/C_F_F_.py
```diff
@@ -1,6 +1,11 @@
 from io import BytesIO
 from fontTools import cffLib
 from . import DefaultTable
+from typing import Any, Dict, List, Optional, TYPE_CHECKING
```
jorenham, line 4: There's no need for `Dict`, `List` anymore since Python 3.9 :)
https://peps.python.org/pep-0585/

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -102,25 +105,46 @@ class TTFont(object):
...
-        allowVID=NotImplemented,
-        ignoreDecompileErrors=False,
-        recalcTimestamp=True,
-        fontNumber=-1,
-        lazy=None,
-        quiet=None,
-        _tableCache=None,
-        cfg={},
-    ):
+        file: str | IO[bytes] | Any | None = None,  # Using Any for generic file-like objects
```
jorenham: Does `os.PathLike` ([docs](https://docs.python.org/3/library/os.html#os.PathLike)) fit the bill for "generic file-like objects"?

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -102,25 +105,46 @@ class TTFont(object):
...
+        res_name_or_index: str | int | None = None,
+        sfntVersion: bytes = b"\000\001\000\000",
+        flavor: str | None = None,
+        checkChecksums: int = 0,
+        verbose: bool | None = None,  # Deprecated
+        recalcBBoxes: bool = True,
+        allowVID: Any = NotImplemented,  # Deprecated/Unused
```
jorenham, line 139: Once Python 3.9 support is no longer required, `[types.NotImplementedType](https://docs.python.org/3/library/types.html#types.NotImplementedType)` (docs) could be used for this. Otherwise, you could import it from `typing_extensions` from within a `if TYPE_CHECKING:` block (which magically doesn't require `typing_extensions` to be installed), and stringify-annotate it.

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -102,25 +105,46 @@ class TTFont(object):
...
+        ignoreDecompileErrors: bool = False,
+        recalcTimestamp: bool = True,
+        fontNumber: int = -1,
+        lazy: bool | None = None,
+        quiet: bool | None = None,  # Deprecated
```
jorenham, line 144: If you want to avoid people from using this, perhaps it's an idea to annotate it as 

```suggestion
        quiet: None = None,  # Deprecated
```

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -102,25 +105,46 @@ class TTFont(object):
...
+        _tableCache: dict[tuple[Tag, bytes], DefaultTable] | None = None,
```
jorenham, line 145: Keep in mind that both type parameters of `dict` are invariant. So if you have

```py
class DefaultTableSubclass(DefaultTable): ...

my_table_cache: dict[tuple[Tag, bytes], DefaultTableSubclass] = ...
```

then you can't pass `my_table_cache` as `_tableCache` here. So the value types must be exactly the same.
This is also the case for the key type, so `Tag` subclasses won't work.

So if that's not what you want, then you could use `collections.abc.Mapping`, which has a covariant value type, but the key type is unfortunately also invariant (which is a known issue). To get around that, you could use a "free" type parameter, e.g. `_TagT = TypeVar("_TagT", bound=Tag)`, and use that instead of `Tag`. Since it's not associated to anything else, it can materialize to anything assignable to `Tag`, effectively emulating covariant behavior ;)

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -131,78 +155,101 @@ def __init__(
...
+        # Now current_file is either the original seekable file-like object,
+        # an opened file handle, or a BytesIO stream.
+        from fontTools.ttLib.sfnt import SFNTReader  # Import here or top with guard
+        self.reader = SFNTReader(current_file, checkChecksums, fontNumber=fontNumber)
         self.sfntVersion = self.reader.sfntVersion
         self.flavor = self.reader.flavor
         self.flavorData = self.reader.flavorData
 
-    def __enter__(self):
+    def __enter__(self) -> 'TTFont':
```
jorenham: For subclasses of `TTFont` this won't be as accurate as it could be. Usually, `typing[_extensions].Self` is used in these kinds of situations, but it would theoretically be equivalent to use a `_T = TypeVar("_T")` to write `(self: _T) -> _T`.

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -131,78 +155,101 @@ def __init__(
...
         return self
 
-    def __exit__(self, type, value, traceback):
+    def __exit__(
+        self,
+        exc_type: Type[BaseException] | None,
```
jorenham, line 246:
```suggestion
        exc_type: type[BaseException] | None,
```

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -268,39 +315,66 @@ def _save(self, file, tableCache=None):
...
             self._writeTable(tag, writer, done, tableCache)
 
         writer.close()
 
         return writer.reordersTables()
 
-    def saveXML(self, fileOrPath, newlinestr="\n", **kwargs):
+    def saveXML(
+        self,
+        fileOrPath: str | IO[str] | Any,  # Use Any for generic text stream
```
jorenham, line 337: maybe `os.PathLike[str]` could replace `Any` here?

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -359,19 +431,21 @@ def _saveXML(
...
-        if tag in self:
-            table = self[tag]
-            report = "Dumping '%s' table..." % tag
-        else:
-            report = "No '%s' table found." % tag
-        log.info(report)
+
         if tag not in self:
+            report = "No '%s' table found." % tag
+            log.info(report)
```
jorenham, line 450: I realize this is (also) out of scope, but the recommended way of formatting logging calls would in this case look something like this:

```suggestion
            log.info( "No %r table found.", tag)
```

The `%r` effectively does a `repr(_)`, which gives you free quotes.
And to be honest I can't remember exactly why the format arguments should be passed like this, just that there's a good reason for doing so haha.

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -499,48 +575,50 @@ def _readTable(self, tag):
...
+        tag = Tag(tag)
         if tag not in self:
             raise KeyError("'%s' table not found" % tag)
         if tag in self.tables:
             del self.tables[tag]
         if self.reader and tag in self.reader:
             del self.reader[tag]
 
-    def get(self, tag, default=None):
+    def get(self, tag: str, default: Any | None = None) -> DefaultTable | Any | None:
```
jorenham, line 602: it might be nice to `@overload` this, and use a `TypeVar` for the default, similar to `Mapping.get` in https://github.com/python/typeshed/blob/aa52024657e7aecfe03d08a7e3719a2ded9a907a/stdlib/typing.pyi#L768-L773

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -748,8 +832,12 @@ def getTableData(self, tag):
             raise KeyError(tag)
 
     def getGlyphSet(
-        self, preferCFF=True, location=None, normalized=False, recalcBounds=True
-    ):
+        self,
+        preferCFF: bool = True,
+        location: dict[str, float] | None = None,
```
jorenham, line 848: Note that this won't allow you to pass `my_location: dict[str, int]`, but I'm not sure if that's such a bad thing here.

jorenham (commented):
Here's a completely unsolicited review; I hope you don't mind :)

Lib/fontTools/ttLib/ttFont.py
madig: PathLike is for paths, like strings or Path objects. https://docs.python.org/3/library/typing.html#typing.IO is for File-like objects.

Lib/fontTools/ttLib/ttFont.py
madig: Hrm, I'd prefer a https://docs.python.org/3/library/exceptions.html#DeprecationWarning to be emitted if this parameter is used, rather than the type being lied about.

Lib/fontTools/ttLib/ttFont.py
madig: Hrm, I'd prefer a https://docs.python.org/3/library/exceptions.html#DeprecationWarning to be emitted if this parameter is used, rather than the type/default value being lied about.

Lib/fontTools/ttLib/ttFont.py
madig: Rather, `os.PathLike[str] | IO[str]` instead of `str | IO[str] | Any`.

Lib/fontTools/ttLib/ttFont.py
madig: I believe the string interpolation will only happen when the log happens, that's why they recommend doing it the way you suggest. https://docs.astral.sh/ruff/rules/logging-percent-format/

Lib/fontTools/ttLib/ttFont.py
madig: One could turn this into `dict[str, float] | dict[str, int] | None`. Not sure if there's a more elegant way.

jorenham: You could use the `dict[str, T]` trick, with `T = TypeVar("T", bound=float)`

Lib/fontTools/ttLib/ttFont.py
jorenham: `str` won't be assignable to that

madig: Ugh. I forgot that PathLike is a special snowflake.

jorenham: Well, I suppose there's [`_typeshed.StrPath`](https://github.com/python/typeshed/blob/fe11cc058c59679573b08c5f4de48012fa05ca22/stdlib/_typeshed/__init__.pyi#L182) you could use

schriftgestalt:
I currently have no time to work on this. Is there a way to work on this together?

jorenham:
> I currently have no time to work on this. Is there a way to work on this together?

I wouldn't mind helping out. What do you propose?

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -103,24 +106,48 @@ class TTFont(object):
...
-        lazy=None,
-        quiet=None,
-        _tableCache=None,
-        cfg={},
-    ):
+        file: (
+            str | IO[bytes] | Any | None
+        ) = None,  # Using Any for generic file-like objects
+        res_name_or_index: str | int | None = None,
+        sfntVersion: bytes = b"\000\001\000\000",
```
madig, line 134: This is more correct, but there is code that checks for a str, not bytes...

schriftgestalt:
there are a lot of good suggestions. Someone needs to put them in. 
I’m not sure how that works git wise. Can you add commits to my fork?

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -131,78 +158,106 @@ def __init__(
...
             # SFNTReader wants the input file to be seekable.
             # SpooledTemporaryFile has no seekable() on < 3.11, but still can seek:
             # https://github.com/fonttools/fonttools/issues/3052
-            if hasattr(file, "seekable"):
-                seekable = file.seekable()
-            elif hasattr(file, "seek"):
+            if hasattr(current_file, "seekable"):
+                seekable = current_file.seekable()  # type: ignore
+            elif hasattr(current_file, "seek"):
                 try:
```
madig, line 201: What is the reason behind this change?

jorenham:
> there are a lot of good suggestions. Someone needs to put them in. I’m not sure how that works git wise. Can you add commits to my fork?

I'm not a maintainer here, so not directly. I could PR to your PR branch on your fork I suppose.

madig:
I can push to the branch of your fork. I'm currently scratching my head a bit though, because there are various code changes that go way beyond adding type hints and break tests. I'm contemplating breaking off pieces of this PR and focussing on adding type hints without changing anything else.

anthrotype:
> focussing on adding type hints without changing anything else.

strong yes please. We already had a regression which slipped through another similar big type-hinting releated PR (for ufoLib) recently

Lib/fontTools/ttLib/ttFont.py
madig: Meaning that TTFont would need to use generics like so: `class TTFont(Generic[_TagT])`?

@anthrotype I don't think we ever subclass Tag, no? I see no reason to...

Issue #3827, opened 2025-05-14
CFF2 instantiator fail in fetchBlendsFromVarStore


simoncozens:
Here is a very small font containing only the letter 'e' in a CFF2 table: [dummy2.otf.gz](https://github.com/user-attachments/files/20206943/dummy2.otf.gz)

When attempting to instantiate this font, I get:

```
$ python3 -m fontTools.varLib.instancer dummy2.otf wght=400
Restricting axes: {'wght': (400, 400, 400)}
Loading variable font
Restricted limits: {'wght': (400, 400, 400)}
Normalized limits: {'wght': (0.25, 0.25, 0.25, 0, 400)}
Instantiating CFF2 table
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/opt/homebrew/lib/python3.13/site-packages/fontTools/varLib/instancer/__main__.py", line 5, in <module>
    sys.exit(main())
             ~~~~^^
  File "/opt/homebrew/lib/python3.13/site-packages/fontTools/varLib/instancer/__init__.py", line 1915, in main
    instantiateVariableFont(
    ~~~~~~~~~~~~~~~~~~~~~~~^
        varfont,
        ^^^^^^^^
    ...<5 lines>...
        downgradeCFF2=options.downgrade_cff2,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/opt/homebrew/lib/python3.13/site-packages/fontTools/varLib/instancer/__init__.py", line 1651, in instantiateVariableFont
    instantiateCFF2(varfont, normalizedLimits, downgrade=downgradeCFF2)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.13/site-packages/fontTools/varLib/instancer/__init__.py", line 816, in instantiateCFF2
    value = fetchBlendsFromVarStore(value)
  File "/opt/homebrew/lib/python3.13/site-packages/fontTools/varLib/instancer/__init__.py", line 730, in fetchBlendsFromVarStore
    deltas = varData.Item[minor]
             ~~~~~~~~~~~~^^^^^^^
IndexError: list index out of range
```

behdad:
Weird. I get a crash here:
```
[...]

PR #3828, opened 2025-05-14
[instantiate/cff2] Fix vsindex of Private dicts when instantiating

behdad:
Also add a TODO for a bug in that code.

Fixes https://github.com/fonttools/fonttools/issues/3827


simoncozens:
I don't understand it, but this works for me...

PR #3830, opened 2025-05-15
[merge] handle cmap format=14 subtable

JLHwung:
In this PR we add cmap format=14 table support to the merge tool.

For duplicate UVS mapping entries, the tool prefer the previous font, aligned with how we deal with cmap duplicates.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_c_m_a_p.py
+++ b/Lib/fontTools/ttLib/tables/_c_m_a_p.py
@@ -221,9 +221,7 @@ class table__c_m_a_p(DefaultTable.DefaultTable):
                 chunk = table.compile(ttFont)
                 offset = done.get(chunk)
                 if offset is None:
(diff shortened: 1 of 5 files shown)


Lib/fontTools/ttLib/tables/_c_m_a_p.py
```diff
@@ -350,6 +349,10 @@ def isUnicode(self):
             self.platformID == 3 and self.platEncID in [0, 1, 10]
         )
 
+    def isUVS(self):
+        """Returns true if it is a UVS subtable"""
+        return self.platformID == 0 and self.platEncID == 5
```
anthrotype, line 354: shouldn't this also check that self.format == 14? Or is platEncID=5 only used for cmap14?

anthrotype (approved):
the changes to the merge.py and corresponding tests look good to me. But see my other question about `_c_m_a_p.py`

PR #3831, opened 2025-05-16
Instances should only reuse name ID 2 or 17 if they are at the default location across all axes

### LINKED ISSUE #3825: [name] Instances should only reuse name ID 2 or 17 if they are at the default location across all axes
Currently, the code tries to reuse name ID 2 or 17 for the "Regular" variable font instance here: https://github.com/fonttools/fonttools/blob/e5c9b1e7b04c0f06d52fa5cb42ae6e7f8f13fe6a/Lib/fontTools/varLib/__init__.py#L133-L135

The [OT spec says this, though](https://learn.microsoft.com/en-us/typography/opentype/spec/fvar#instancerecord), emphasis mine:

> The subfamilyNameID field provides a name ID that can be used to obtain strings from the 'name' table that can be treated as equivalent to name ID 17 (typographic subfamily) strings for the given instance. Values of 2 or 17 can be used; otherwise, values must be greater than 255 and less than 32768. **The values 2 or 17 should only be used if the named instance corresponds to the font’s default instance.**

Imagine a Designspace where the default weight is a Medium (wght=500) and there is a Regular instance at wght=400. The Regular instance should not use name ID 2 or 17 but something >= 255.

This is probably an irrelevant violation of the spec, but there are tools out there that check for it...

(Thanks to @Hoolean for identifying the issue internally.)

madig:
Closes #3825

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -109,6 +109,8 @@ def _add_fvar(font, axes, instances: List[InstanceDescriptor]):
         axis.flags = int(a.hidden)
         fvar.axes.append(axis)
 
+    default_coordinates = {axis.axisTag: axis.defaultValue for axis in fvar.axes}
+
     for instance in instances:
         # Filter out discrete axis locations
         coordinates = {
@@ -130,16 +132,26 @@ def _add_fvar(font, axes, instances: List[InstanceDescriptor]):
         psname = instance.postScriptFontName
 
         inst = NamedInstance()
+        inst.coordinates = {
+            axes[k].tag: axes[k].map_backward(v) for k, v in coordinates.items()
+        }
+
+    	# Instances should only reuse name ID 2 or 17 if they are at the default
+        # location across all axes. See
+        # https://github.com/fonttools/fonttools/issues/3825.
+        styleNameExists = nameTable.findMultilingualName(
+            localisedStyleName, windows=True, mac=macNames, minNameID=0
+        )
+        if styleNameExists in {2, 17} and inst.coordinates != default_coordinates:
+            minNameId = 256
+        else:
+            minNameId = 0
         inst.subfamilyNameID = nameTable.addMultilingualName(
-            localisedStyleName, mac=macNames
+            localisedStyleName, mac=macNames, minNameID=minNameId
         )
         if psname is not None:
             psname = tostr(psname)
             inst.postscriptNameID = nameTable.addName(psname, platforms=platforms)
-        inst.coordinates = {
-            axes[k].tag: axes[k].map_backward(v) for k, v in coordinates.items()
-        }
-        # inst.coordinates = {axes[k].tag:v for k,v in coordinates.items()}
         fvar.instances.append(inst)
 
     assert "fvar" not in font


anthrotype:
the existing test files don't have a default instance and you're testing that nameID 2/17 are no longer being used; maybe you should also add a test with an explicit default instance that checks the nameID2/17 do get used

Lib/fontTools/varLib/__init__.py
```diff
@@ -130,16 +132,26 @@ def _add_fvar(font, axes, instances: List[InstanceDescriptor]):
...
 
         inst = NamedInstance()
+        inst.coordinates = {
+            axes[k].tag: axes[k].map_backward(v) for k, v in coordinates.items()
+        }
+
+        # Instances should only reuse name ID 2 or 17 if they are at the default
+        # location across all axes. See
+        # https://github.com/fonttools/fonttools/issues/3825.
+        styleNameExists = nameTable.findMultilingualName(
```
anthrotype, line 142: if findMultilingualName finds something that is either 2 or 17 and instance.coordinates == default, you can just set that to the instance.subfamilyNameID, no? Why call addMultilingualName again in that case? The latter will again try to find the same nameID you already found (the one in `styleNameExists`)

Lib/fontTools/varLib/__init__.py
```diff
@@ -130,16 +132,27 @@ def _add_fvar(font, axes, instances: List[InstanceDescriptor]):
         psname = instance.postScriptFontName
 
         inst = NamedInstance()
-        inst.subfamilyNameID = nameTable.addMultilingualName(
-            localisedStyleName, mac=macNames
+        inst.coordinates = {
+            axes[k].tag: axes[k].map_backward(v) for k, v in coordinates.items()
+        }
+
+        subfamilyNameID = nameTable.addMultilingualName(
```
anthrotype, line 139: I find it a bit confusing that that you call a method named `addMultilingualName` twice. I know this only adds if it one is not already there, but wouldn't it read better if you did findMultilingualName and then add... (the latter only once)?

madig: Maybe I'm being dense, but how? `Find` can also return None, so you'd have to have two calls to `add` either way? Like so:

```py
subfamilyNameID = findMultilingualName(...)
if subfamilyNameID in {2, 17} and inst.coordinates != default_coordinates:
    inst.subfamilyNameID = addMultilingualName(..., minNameID=256)
elif subfamilyNameID is None:
    inst.subfamilyNameID = addMultilingualName(...,  minNameID=0)
else:
    inst.subfamilyNameID = subfamilyNameID
```

anthrotype: would this not work?

```python
subfamilyNameID = findMultilingualName(...)
if subfamilyNameID in {2, 17} and inst.coordinates == default_coordinates:
    inst.subfamilyNameID = subfamilyNameID
else:
    # subfamilyNameID is None, or neither 2 or 17, or it is but instance coordinates != default
    inst.subfamilyNameID = nameTable.addMultilingualName(
        localisedStyleName, windows=True, mac=macNames, minNameID=256
    )
```

PR #3832, opened 2025-05-20
[instancer/CFF2] Inherit vsindex from private dict

behdad:
The CharString vsindex defaults to that of its private dict, not 0.


anthrotype (approved):
LGTM. I assume you tested this to work on a real font?

behdad:
> LGTM. I assume you tested this to work on a real font?

I'm yet to find a font that uses `vsindex` in Private dict. So, in a sense, no I haven't. But it replicates the current results for all fonts around basically.

PR #3838, opened 2025-05-23
Create a dflt LangSys in a ScriptRecord when adding variations later

madig:
We found this one during the development of a font. The GSUB table builder seems to discard defined language systems when they have no lookups. This is a problem when we add feature variations in a later step, like we do when we compile a Designspace with rules.

When like in the test code no lookup is defined for `latn` and `dflt`, just for the `CAT ` language, the GSUB table won't have a `dflt` LangSysRecord that https://github.com/fonttools/fonttools/blob/e89d7db4f457742a03d7a07121fd7348ca3ee1ca/Lib/fontTools/varLib/featureVars.py#L394 needs to attach itself to.

TODO:

- [x] Come up with a fix.

### DIFF
--- a/Lib/fontTools/misc/symfont.py
+++ b/Lib/fontTools/misc/symfont.py
@@ -235,6 +235,7 @@ if __name__ == '__main__':
 
 if __name__ == "__main__":
     import sys
+
     if sys.argv[1:]:
         penName = sys.argv[1]
         funcs = [(name, eval(f)) for name, f in zip(sys.argv[2::2], sys.argv[3::2])]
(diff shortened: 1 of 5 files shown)


behdad:
A no-lookup langsys is overriding the dflt langsys, so it changes font behavior if you drop it.

behdad:
And for dropping dflt, yes, indeed, you need to either never drop it, or only drop it if it has no data whatsoever, including variations.

behdad (approved):

anthrotype (approved):
LGTM

pls fix the lint as well thanks

anthrotype:
> The GSUB table builder seems to discard defined language systems when they have no lookups. This is a problem when we add feature variations in a later step, like we do when we compile a Designspace with rules.

@madig Have you tested if this also happens with fontc as well?

madig:
I have not, can do soon.

By the way, Harry and me were reading https://learn.microsoft.com/en-us/typography/opentype/spec/chapter2#scriptlist-table and wondered if a lazily-created "latn dflt" LangSys for feature variations should carry all features of "DFLT dflt"? Our reading of the spec says "no".

anthrotype:
I don't think so either

Issue #3839, opened 2025-05-23
CFF specializer error with fontmake variable-cff2 output


justvanrossum:
I have a font that had a glyph with a stray point that gives an error within the CFF specializer.

This happens only when outputting with `-o variable-cff2`, not with `-o variable`.

The glyph contains one "normal" shape (a line and a curve) and a stray (single, loose) point.

If I remove the stray point the compile works. But if I delete the other shape, the compile *also* works, so the problem is not *just* with the stray point.

It seems to have to do with the exact coordinates of the shape: I have tried to draw a shape with the same topology (but different coordinates), but failed to reproduce the problem.

Now, of course the source data is buggy, but it's at least odd that this only fails for cff2 output.

I have attached a zip with a .designspace file and two UFOs, each containing a single glyph.

```
Traceback (most recent call last):
  File "xxx/venv/lib/python3.12/site-packages/fontmake/font_project.py", line 1208, in run_from_designspace
    self._run_from_designspace_interpolatable(
  File "xxx/venv/lib/python3.12/site-packages/fontmake/font_project.py", line 1326, in _run_from_designspace_interpolatable
    self.build_variable_fonts(
  File "xxx/venv/lib/python3.12/site-packages/fontmake/font_project.py", line 470, in build_variable_fonts
    fonts = ufo2ft.compileVariableCFF2s(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "xxx/venv/lib/python3.12/site-packages/ufo2ft/__init__.py", line 295, in compileVariableCFF2s
    return VariableCFF2sCompiler(**kwargs).compile_variable(designSpaceDoc)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "xxx/venv/lib/python3.12/site-packages/ufo2ft/_compilers/baseCompiler.py", line 420, in compile_variable
    vfNameToTTFont = self._merge(designSpaceDoc, excludeVariationTables)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "xxx/venv/lib/python3.12/site-packages/ufo2ft/_compilers/interpolatableOTFCompiler.py", line 32, in _merge
    return varLib.build_many(
           ^^^^^^^^^^^^^^^^^^
[...]

PR #3840, opened 2025-05-23
[cffLib.specializer] Fix rmoveto merging when blends used

behdad:
Fixes https://github.com/fonttools/fonttools/issues/3839


justvanrossum (approved):
I can confirm that the fix works, thanks! I'm not familiar enough with the code for a proper review, though.

behdad:
> I can confirm that the fix works, thanks! I'm not familiar enough with the code for a proper review, though.

Basically, the items in the operands are either a number, or a list if they are a variable number. To add two operands we need to call `_addArgs` instead of using the `+` operator.

Issue #3846, opened 2025-05-27
[feaLib] _upgrade_mixed_subst_statements does not account for all possible lookup breaks


cmyr:
Another little issue with #3805.

Consider the following FEA:

```fea
feature derp {
    sub x by y;
    lookupflag 4;
    sub f i by one;
} derp;
```

because the logic in `up` (https://github.com/fonttools/fonttools/blob/b21e538f2fb04f2397adde0c52fc7fd0d9e831c7/Lib/fontTools/feaLib/ast.py#L340) only considers the sub statements, it ends up turning the `sub x by y` rule into a ligature lookup, even though the subsequent lookupflag statement means it is not in the same lookup as the subsequent ligature sub rule.

The same miscompilation can also occur, for instance, with FEA like:

```fea
lookup ther {
sub l by j;
} ther ;

feature clig {
    sub a by b;
    sub e l l' lookup lookup_1;
    sub e l f by z;
} clig;
```

where the contextual rule in the middle forces a new lookup, and we end up giving the first rule the wrong lookup type.

khaledhosny:
I think we need to move this logic further down into the builder. Last time I checked this seemed rather complicated.

khaledhosny:
https://github.com/fonttools/fonttools/pull/3849

PR #3849, opened 2025-05-28
[feaLib] Improve single substitution promotion

khaledhosny:
Move the logic down into the builder so that it is applied after lookups are split due to e.g. different lookup flags or interspersed contextual substitution rules. Arguably the builder was the right place for this all along, but I took the either route (twice!) before.

Fixes https://github.com/fonttools/fonttools/issues/3846

### DIFF
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -170,6 +170,9 @@ class LookupBuilder(object):
             and self.extension == other.extension
         )
 
+    def promote_lookup_type(self, is_named_lookup):
+        return [self]
+
     def inferGlyphClasses(self):
         """Infers glyph glasses for the GDEF table, such as {"cedilla":3}."""
         return {}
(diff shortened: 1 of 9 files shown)


Lib/fontTools/otlLib/builder.py
```diff
@@ -1324,6 +1327,151 @@ def add_subtable_break(self, location):
...
+
+        # sub x by y;
+        builder.mapping[("x",)] = ("y",)
+        # sub a by b c;
+        builder.mapping[("a",)] = ("b", "c")
+        # sub f i by f_i;
+        builder.mapping[("f", "i")] = ("f_i",)
+
+    Then call `promote_lookup_type()` to convert this builder into the
+    appropriate type of substitution lookup builder. This would prompote single
```
anthrotype, line 1344:
```suggestion
    appropriate type of substitution lookup builder. This would promote single
```

Tests/otlLib/maxContextCalc_test.py
```diff
@@ -30,7 +30,6 @@ def test_max_ctx_calc_features():
         sub A B C by c;
         sub [A B] C by c;
         sub [A B] C [A B] by c;
-        sub A by A B;
```
anthrotype, line 33: was this change intentional?

anthrotype (approved):

Issue #3773, opened 2025-02-20
[subset] `--obfuscate-names` can drop or scramble names used outside of the `name` table


Hoolean:
The fonttools subsetter has an option for obfuscating names to make web fonts unusable as system fonts without modification.

This only targets specific standard names, with font-specific names (IDs 256+) left alone:

https://github.com/fonttools/fonttools/blob/598b974f87f35972da24e96e45bd0176d18930a0/Lib/fontTools/subset/__init__.py#L3013-L3025

When building dependent tables such as `fvar` and `STAT`, however, fonttools will seek to reuse existing name records: for example, an `fvar` instance named `Regular` may reuse the 'Font Subfamily' name (ID 2) instead of creating an identical but font-specific entry (ID 256+) in order to save space. 

This is desirable and spec-compliant, but as a side-effect means that later fonttools scrambling will reach outside of the `name` entry into any `fvar` and `STAT` entries that reused the name ID, and scramble their names too as an unintended consequence.

Where this happens, should we make duplicates in the font-specific area (ID 256+) and redirect other tables to them, or are we confident that none of these name IDs are exposed in the environments in which names may wish to be obfuscated (e.g. browsers)?

Hoolean:
Example scenario:

1. TTF is built with `STAT` or `fvar` referencing name ID 17
2. TTF is converted to WOFF2, with names obfuscated, which drops ID 17
3. `STAT` and `fvar` now point to a name ID that does not exist, and so are invalid according to the spec
4. OTS and applications that depend on it or perform their own validation will reject these tables

Hoolean:
The right fix would be to insert font-specific names (IDs 256+) with equivalent contents for these tables to refer to.

PR #3853, opened 2025-06-02
Copy name IDs in use before scrapping or scrambling them for webfonts

madig:
Preserve select nameIDs in `fvar` and `STAT` when obfuscating names, to avoid dangling name IDs in subset fonts.

Closes https://github.com/fonttools/fonttools/issues/3773.

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -3031,7 +3031,7 @@ def prune_post_subset(self, font, options):
         # parts of the font don't break.
         needRemapping = visitor.seen.intersection(NAME_IDS_TO_OBFUSCATE)
         if needRemapping:
-            remap_select_name_ids(font, needRemapping)
+            _remap_select_name_ids(font, needRemapping)
         for n in self.names:
             if n.nameID in [1, 4]:
                 n.string = ".\x7f".encode("utf_16_be") if n.isUnicode() else ".\x7f"
@@ -3046,11 +3046,17 @@ def prune_post_subset(self, font, options):
     return True  # Required table
 
 
-def remap_select_name_ids(font: ttLib.TTFont, needRemapping: set[int]) -> None:
+def _remap_select_name_ids(font: ttLib.TTFont, needRemapping: set[int]) -> None:
     """Remap a set of IDs so that the originals can be safely scrambled or
-    dropped."""
+    dropped.
 
-    if "fvar" not in font:
+    For each name record whose name id is in the `needRemapping` set, make a copy
+    and allocate a new unused name id in the font-specific range (> 255).
+
+    Finally update references to these in the `fvar` and `STAT` tables.
+    """
+
+    if "fvar" not in font and "STAT" not in font:
         return
 
     name = font["name"]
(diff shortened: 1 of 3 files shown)


madig:
Hm, how to get `CPAL` labels and the `feat` table into a test font?

anthrotype:
colorLib.buildCPAL can build palette labels, there's a bunch of tests in Tests/colorLib/builder_test.py, e.g.

https://github.com/fonttools/fonttools/blob/e857fe5ef7b25e92fd829a445357e45cde16eb04/Tests/colorLib/builder_test.py#L151

as to [feat](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6feat.html), it's defined in otData.py and it doesn't look like we test it anywhere..

madig:
I limited the test to STAT, fvar, ss and cv feature names and CPAL labels for now.

Lib/fontTools/subset/__init__.py
```diff
@@ -470,7 +473,7 @@ def _log_glyphs(self, glyphs, font=None):
 
 
 def _dict_subset(d, glyphs):
-    return {g: d[g] for g in glyphs}
+    return {g: d[g] for g in glyphs if g in d}
```
anthrotype: this particular change seems unrelated

madig: The test I added will crash because there are no variations, so you get:

```
d = {'.notdef': []}, glyphs = frozenset({'.notdef', 'A', 'A.something', 'B', 'B.cv01', 'C', ...})

    def _dict_subset(d, glyphs):
>       return {g: d[g] for g in glyphs}
                   ^^^^
E       KeyError: 'a'
```

anthrotype: this could hide other bugs, let's revert it and find an alternative way to make the test pass

madig: It passes when I delete the empty gvar table from the TTX file. Acceptable?

anthrotype: probably better to adapt the test file insead of changing lib code just to make one single test pass

Tests/subset/data/PreserveSillyNamesTest.ttx
```diff
@@ -0,0 +1,553 @@
...
+  </HVAR>
+
+  <STAT>
+    <Version value="0x00010001"/>
+    <DesignAxisRecordSize value="8"/>
+    <!-- DesignAxisCount=2 -->
+    <DesignAxisRecord>
+      <Axis index="0">
+        <AxisTag value="wght"/>
+        <AxisNameID value="16"/>  <!-- Weight -->
```
anthrotype, line 443: you probably took a ttx file and manually edited the nameIDs because the xml comments don't seem to match the content of the respective nameIDs (e.g. here shouldn't nameID=16 be "Blu" if I read that correctly from the name table?).

Tests/subset/subset_test.py
```diff
@@ -2122,5 +2123,24 @@ def test_cvXX_feature_params_nameIDs_are_retained():
...
+    assert visitor.seen.intersection(subset.NAME_IDS_TO_OBFUSCATE)
+
+    options = subset.Options(obfuscate_names=True)
+    subsetter = subset.Subsetter(options)
+    subsetter.populate(glyphs=font.getGlyphOrder())
+    subsetter.subset(font)
+
+    visitor = NameRecordVisitor()
+    visitor.visit(font)
+    assert not visitor.seen.intersection(subset.NAME_IDS_TO_OBFUSCATE)
```
anthrotype: the names as 'seen' no longer intersect the obfuscated names after the subset operation; ok, but this doesn't show the names have actually been kept and remapped (they could have just been nuked).

Tests/subset/data/PreserveSillyNamesTest.ttx
madig: Yep. Edited them for max impact. Found me a bug even!

Lib/fontTools/subset/__init__.py
```diff
@@ -3036,6 +3046,67 @@ def prune_post_subset(self, font, options):
     return True  # Required table
 
 
+def remap_select_name_ids(font: ttLib.TTFont, needRemapping: set[int]) -> None:
+    """Remap a set of IDs so that the originals can be safely scrambled or
+    dropped."""
+
+    if "fvar" not in font:
```
anthrotype: technically STAT could be present even if fvar is not...

Lib/fontTools/subset/__init__.py
```diff
@@ -3036,6 +3046,67 @@ def prune_post_subset(self, font, options):
     return True  # Required table
 
 
+def remap_select_name_ids(font: ttLib.TTFont, needRemapping: set[int]) -> None:
```
anthrotype: I suggest you make this private `_remap_select_name_ids`

Issue #71, opened 2013-12-10
[ttx] Control chars in xml


behdad:
Currently, we write out control chars to XML.  That's invalid XML.  Need to escape those.  Can be in a feature tag, or name table, or CFF, etc.  Test for example with Salsa-Regular.otf from googlefontdirectory.

anthrotype:
Reminder to self: see if we can define extra markup to escape invalid control chars in XML. 

some relevant links:
https://www.w3.org/International/questions/qa-controls
http://stackoverflow.com/questions/4513672/python-escaping-non-ascii-characters-in-xml

shwetamar85:
Has there  been a solution to this?

anthrotype:
No, not yet I'm afraid. Any suggestions?

anthrotype:
I'm thinking of using a regular expression to escape any occurrence of invalid control characters in character data with <u>0001</u>, as in this SO answer http://stackoverflow.com/a/4519925

WDYT?

anthrotype:
`<u>0001</u>`

anthrotype:
i think this is a legitimate feature request, even if arguably low priority, let's keep open, maybe one day someone will get to it

anthrotype:
I consider this fixed by #3868, even though we aren't roundtripping the illegal characters from/to XML, but merely replacing them with "?" and issuing a warning. I think it's enough for the purpose of ttx'ing buggy fonts.

PR #3868, opened 2025-06-22
Escape control characters in `xmlWriter` and test.

knutnergaard:
This hopefully resolves issue #71.

### DIFF
--- a/Lib/fontTools/misc/xmlWriter.py
+++ b/Lib/fontTools/misc/xmlWriter.py
@@ -5,18 +5,21 @@ import sys
 import os
 import string
 import logging
+import itertools
 
 INDENT = "  "
 TTX_LOG = logging.getLogger("fontTools.ttx")
 REPLACEMENT = "?"
-ILLEGAL_CHARS = dict.fromkeys(
-    [chr(c) for c in range(0x00, 0x09)]
-    + [chr(0x0B), chr(0x0C)]
-    + [chr(c) for c in range(0x0E, 0x20)]
-    + [chr(0xFFFE), chr(0xFFFF)],
+ILLEGAL_XML_CHARS = dict.fromkeys(
+    itertools.chain(
+        range(0x00, 0x09),
+        (0x0B, 0x0C),
+        range(0x0E, 0x20),
+        range(0xD800, 0xE000),
+        (0xFFFE, 0xFFFF),
+    ),
     REPLACEMENT,
 )
-ILLEGAL_TRANS = str.maketrans(ILLEGAL_CHARS)
 
 
 class XMLWriter(object):
@@ -179,13 +182,14 @@ class XMLWriter(object):
 
 
 def escape(data):
+    """Escape characters not allowed in `XML 1.0 <https://www.w3.org/TR/xml/#NT-Char>`_."""
     data = tostr(data, "utf_8")
     data = data.replace("&", "&amp;")
     data = data.replace("<", "&lt;")
     data = data.replace(">", "&gt;")
     data = data.replace("\r", "&#13;")
 
-    newData = data.translate(ILLEGAL_TRANS)
+    newData = data.translate(ILLEGAL_XML_CHARS)
     if newData != data:
         maxLen = 10
         preview = repr(data)


Tests/ttx/ttx_test.py
```diff
@@ -69,6 +72,86 @@ def read_file(file_path):
...
+        name_table.setName("Control\x01Char", 1, 3, 1, 0x409)
+        font["name"] = name_table
+
+        # Save to XML
+        self.temp_dir()
+        ttx_path = Path(self.tempdir) / "test.ttx"
+        font.saveXML(str(ttx_path))
+
+        xml_content = ttx_path.read_text(encoding="utf-8")
+        assert "&#x01;" in xml_content
```
justvanrossum: I think you may want to test that the `""Control\x01Char"` string survives round-tripping, ie. after importing the font from the .ttx data.

Tests/ttx/ttx_test.py
```diff
@@ -69,6 +72,86 @@ def read_file(file_path):
     # Tests
     # -----
 
+    def test_saveXML_escapes_control_characters(self):
+        # create font
+        font = TTFont()
+
+        font.setGlyphOrder([".notdef"])
```
justvanrossum: Potentially the setup of this dummy font can be made (much) more compact by using [FontBuilder](https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/fontBuilder.py). I'm not sure you even need to add a `glyf` table. `head` and `name` may be enough.

Tests/ttx/ttx_test.py
knutnergaard: I tried this , but couldn't make the test pass (see commented-out test code in the last commit).

Tests/ttx/ttx_test.py
knutnergaard: Done.

Tests/ttx/ttx_test.py
```diff
@@ -69,6 +74,41 @@ def read_file(file_path):
...
+        fb = FontBuilder(unitsPerEm=1000, isTTF=True)
+        fb.setupGlyphOrder([".notdef"])
+        fb.setupCharacterMap({})
+        fb.setupGlyf({".notdef": Glyph()})
+        fb.setupHorizontalMetrics({".notdef": (600, 0)})
+        fb.setupHorizontalHeader(ascent=800, descent=-200)
+        fb.setupOS2()
+        fb.setupPost()
+
+        control_string = "Control\x01Char"
```
justvanrossum: Equivalent:
```python
        fb.setupNameTable({"familyName": "Control\x01Char"})
```

Tests/ttx/ttx_test.py
justvanrossum: Ah, when parsing `"&#x01;"` we get:
```
xml.parsers.expat.ExpatError: reference to invalid character number: line 149, column 13
```
This suggests that escaping that way is not enough to create valid XML. Not sure how to proceed.

Tests/ttx/ttx_test.py
knutnergaard: Fixed.

Tests/ttx/ttx_test.py
knutnergaard: Seems to me like the characters must either be replaced by some agreed upon, safe alternative, or the issue should be closed with no solution.

knutnergaard: @justvanrossum Perhaps we could add some sort of safe mode to `saveXML`, which users may opt into, that strips the characters or replaces them safely? This could then be tied to a boolean flag in `ttx`.

BTW, I think `ttx` might benefit from employing `argparse` rather than `getopt`. Any objections to refactoring the module?

justvanrossum: I don't really know, I'm all out of knowledge and optinions for this. Maybe others want to chime in.

About ttx argparse/getopt: please open a discussion or issue where you can explain what the benefits are.

anthrotype:
no this won't work. Most control characters are not allowed in XML, even if encoded as hexadecimal character references.
We control our schema so we could wrap these with a new element tag of our chosing which is what I hinted at in my old comment at https://github.com/fonttools/fonttools/issues/71#issuecomment-277435178

Alternatively, the XMLWriter could fall back to writing strings that contain invalid control characters using dumphex method, same we use for serializing binary data like bytecode and such.

knutnergaard:
@anthrotype I see. I'll look into that. I don't have a thorough understanding of the conversion logic yet. Is it necessary to handle this on a table-by-table basis, or can it be centralized? `writeRaw` looks like such an option when writing to xml, but I'm not sure if there is a reliable equivalent when reading.

anthrotype:
The problem is that, even if we define a new element to handle illegal XML characters occurring inside text content (e.g. name records), we can't use the same technique for illegal characters that may occur inside element attributes (e.g. feature tags).

The other thing I was thinking was to use escape sequences such as the ones in python string repr (e.g. "\x01" or "\ufffe") but then we need to double up the backslash character itself and document this to any other consumers of ttx files.

knutnergaard:
@anthrotype Appending an `encoding` attribute and converting the whole string containing invalid characters to hex seems to work fine for name records: 

```python
def toXML(self, writer, ttFont):
        try:
            unistr = self.toUnicode()
        except UnicodeDecodeError:
            unistr = None
        attrs = [
            ("nameID", self.nameID),
            ("platformID", self.platformID),
            ("platEncID", self.platEncID),
            ("langID", hex(self.langID)),
        ]

        def getHex():
            raw = (
                self.string
                if isinstance(self.string, bytes)
                else self.string.encode(self.getEncoding())
            )
            return raw.hex()

        hexstr = None
        if unistr is None:
            attrs.append(("unicode", False))
            attrs.append(("encoding", "hex"))
            hexstr = getHex()
        elif hasInvalidXMLChars(unistr):
            attrs.append(("encoding", "hex"))
            hexstr = getHex()
        elif not self.encodingIsUnicodeCompatible():
            attrs.append(("unicode", True))

        writer.begintag("namerecord", attrs)
        writer.newline()
        if hexstr is not None:
            writer.write(hexstr)
        elif unistr is not None:
            writer.write(unistr)
        else:
            writer.write8bit(self.string)
        writer.newline()
        writer.endtag("namerecord")
        writer.newline()
```

Shouldn't this work for element attributes as well?

anthrotype:
But an element like the name record can have an attribute that tells us whether the string content is encoded as hex or not; for an element attribute string itself, how can we tell that it is a literal string or one that was encoded as hex? You'd have to pick an escape character like the backslash and escape that as well to disambiguate from a literal backslash.
I'm not sure it's worth it.

How about initially we just focus on handling the illegal characters that may occur inside the name record strings using an approach like the one you suggested, as I believe these are the most common?

knutnergaard:
@anthrotype Can't we just add a custom attribute, e.g., `<feature tag-hex="630001" />` or am I missing something?

knutnergaard:
That said, `fea` likely does not run an actual risk of containing control characters, at least not to my understanding. `ltag`, `meta`, and `cmap` are more vulnerable. Are there other tables which would befefit from this?

knutnergaard:
Also, come to think of it, couldn't hex conversion be used for strings containing any illegal character, not just control characters? If so, name record conversion could be simplified as follows:
```python
def toXML(self, writer, ttFont):
        try:
            unistr = self.toUnicode()
        except UnicodeDecodeError:
            unistr = None
        
        attrs = [
            ("nameID", self.nameID),
            ("platformID", self.platformID),
            ("platEncID", self.platEncID),
            ("langID", hex(self.langID)),
        ]

        if unistr is None or hasInvalidXMLChars(unistr):
            attrs.append(("encoding", "hex"))
            raw = (
                self.string
                if isinstance(self.string, bytes)
                else self.string.encode(self.getEncoding())
            )
            text = raw.hex()
        else:
            text = unistr

        writer.begintag("namerecord", attrs)
        writer.newline()
        writer.write(text)
        writer.newline()
        writer.endtag("namerecord")
        writer.newline()
```

This change would mean that a lot of test would have to be modified, although not particularly laboriously.

knutnergaard:
@anthrotype This approach to hex conversion in `table__l_t_a_g` seems to work fine:

```python
def toXML(self, writer, ttFont):
        writer.simpletag("version", value=self.version)
        writer.newline()
        writer.simpletag("flags", value=self.flags)
        writer.newline()
        for tag in self.tags:
            tagHex = tag.encode("ascii").hex()
            if hasInvalidXMLChars(tag):
                writer.simpletag("LanguageTag", **{"tag-hex": tagHex})
            else:
                writer.simpletag("LanguageTag", tag=tag)
            writer.newline()
```

anthrotype:
Hm I'm not particularly fond of this approach.. Before you jump on working on this, may I ask is there any particular font that you wish to dump to TTX that currently fails? Or is this only like a theoretical interest of yours?
I believe the added complexity may not be justified considering the relatively little gains. Besides broken or fuzzed fonts, I don't see how one would like to include control characters or non-characters in their text strings that are meant for humans to read..

knutnergaard:
@anthrotype I've just been working with the xml module in another context lately, saw the long standing issue, and thought I'd try  to solve it. IOW, this is purely theoretical from my point of view.

anthrotype:
ok thanks. Maybe the reason the issue is so long standing is there hasn't been enough interest in supporting this. The font where this was first noticed (Salsa-Regular.ttf from GF) seems to work just fine and was probably just a font bug.

Perhaps what would be useful is stripping these illegal characters (with some warnings) or replacing them with "?" so that the buggy fonts (because that's what these are in the end) can at least be dumped to XML. How about you do that instead?

knutnergaard:
@anthrotype  If this is purely a matter of handling buggy fonts, and the interest is minimal, the script failing seems like not such a bad thing. However, the substitution logic is basically already done in the last commit, so I can change the result and submit it if desired.

Warnings will have to be implemented in `xmlWriter`, or else `ttx` would need a special hook to log this correctly. Maybe not worth it if indeed "nobody cares"? :)

anthrotype:
> the substitution logic is basically already done in the last commit, so I can change the result and submit it if desired.

yes please

> Warnings will have to be implemented in xmlWriter

indeed, which is exactly what you touched in this PR. In addition to replacing with `"?"` you can also issue a logging.warning when the regex matches. The logging message could contain the offending repr(string), shortened to max 10-20 characters perhaps

knutnergaard:
@anthrotype Done!

Lib/fontTools/misc/xmlWriter.py
```diff
@@ -173,6 +175,30 @@ def escape(data):
     data = data.replace("<", "&lt;")
     data = data.replace(">", "&gt;")
     data = data.replace("\r", "&#13;")
+
+    log = logging.getLogger("fontTools.ttx")
+    illegalXML = re.compile("[\x00-\x08\x0b\x0c\x0e-\x1f\ufffe\uffff]")
```
anthrotype: this regex should be compiled outside of the body of the `escape` function, like a constant

Lib/fontTools/misc/xmlWriter.py
```diff
@@ -173,6 +175,30 @@ def escape(data):
     data = data.replace("<", "&lt;")
     data = data.replace(">", "&gt;")
     data = data.replace("\r", "&#13;")
+
+    log = logging.getLogger("fontTools.ttx")
```
anthrotype: also getLogger() can be called once outside the function, usually at the top of the module

anthrotype:
I was thinking, since this is now a simple character by character translation, it should be faster to use str.maketrans + str.translate instead of re.compile/re.sub. You'd first use maketrans to create the translation table from invalid XML characters to "?" and then call translate with it to obtain the new string:

https://docs.python.org/3/library/stdtypes.html#str.maketrans
https://docs.python.org/3/library/stdtypes.html#str.translate

knutnergaard:
Good idea. Changed. I also changed all immutable values to public constants, as they may be useful in testing.

Lib/fontTools/misc/xmlWriter.py
```diff
@@ -4,8 +4,19 @@
...
+TTX_LOG = logging.getLogger("fontTools.ttx")
+REPLACEMENT = "?"
+ILLEGAL_CHARS = dict.fromkeys(
+    [chr(c) for c in range(0x00, 0x09)]
+    + [chr(0x0B), chr(0x0C)]
+    + [chr(c) for c in range(0x0E, 0x20)]
+    + [chr(0xFFFE), chr(0xFFFF)],
+    REPLACEMENT,
+)
+ILLEGAL_TRANS = str.maketrans(ILLEGAL_CHARS)
```
anthrotype: we don't actually need to call str.maketrans, that just returns a plain dict with integers (unicode codepoint) as keys and the replacement string(s) as values, so you could simply do

```suggestion
ILLEGAL_XML_CHARS = dict.fromkeys(
    itertools.chain(
        range(0x00, 0x09),
        (0x0B, 0x0C),
        range(0x0E, 0x20),
        (0xFFFE, 0xFFFF),
    ),
    REPLACEMENT,
)
```

and then you'd use it below like

```
newData = data.translate(ILLEGAL_XML_CHARS)
```

anthrotype: Could you maybe also add a link to the spec? 

https://www.w3.org/TR/xml/#charsets

anthrotype:
@knutnergaard looking at https://www.w3.org/TR/xml/#charsets, I realised that we also need to add the surrogate block `range(0xD800, 0xE000)` to our list of illegal XML characters. Even though a well-formed unicode string in python3 should never contain those if it was decoded properly, it's possible to construct a malformed one e.g. `"This string is not ok " + chr(0xD800)` and then attempting later to encode to utf-8 would give a UnicodeEncodeError.

Lib/fontTools/misc/xmlWriter.py
knutnergaard: > we don't actually need to call str.maketrans

Doh! :)

> Could you maybe also add a link to the spec?
> 
> https://www.w3.org/TR/xml/#charsets

Added in new `escape` docstring.

knutnergaard:
> @knutnergaard looking at https://www.w3.org/TR/xml/#charsets, I realised that we also need to add the surrogate block `range(0xD800, 0xE000)` to our list of illegal XML characters. Even though a well-formed unicode string in python3 should never contain those if it was decoded properly, it's possible to construct a malformed one e.g. `"This string is not ok " + chr(0xD800)` and then attempting later to encode to utf-8 would give a UnicodeEncodeError.

Done. Forgot to add this to the commit message, so had to force push.

Lib/fontTools/misc/xmlWriter.py
anthrotype: thanks, I think it's ready for merge now, will do as soon as CI passes

Issue #3620, opened 2024-08-29
pyfilesystem2 (fs) is no longer maintained; find a replacement or avoid dependency?


anthrotype:
the `fs` module that fontTools.ufoLib uses to implement .ufoz (UFO inside a .zip file) hasn't received an update in more than 2 years, it's basically unmaintained, see: https://github.com/PyFilesystem/pyfilesystem2/issues/571

There is a new module called https://github.com/fsspec/filesystem_spec which may be a suitable replacement and is more actively maintained. Maybe we should try to use that.
It does have a ZipFileSystem implementation:
https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/zip.py

Or alternatively we could reimplement our own filesystem-like wrapper around a zip file.

I'd prefer option 1), or whichever is easier.

justvanrossum:
I just had a brief look at fsspec: it looks very good. Also in favor of option 1).

a-detiste:
Hi, keeping `fs` alive is a liability for the distribution. In Debian I see that `fonttools` will soon be the very last user of `fs`. Replacing it's usage with a maintained library is hugely appreciated.

behdad:
I remember when I prototyped using the `fs` module first in `ufoLib`, I also implemented a dummy fallback. Maybe we can find that and use it if `fs` is not available for now, until port to `fsspec` happens.

behdad:
> I remember when I prototyped using the `fs` module first in `ufoLib`, I also implemented a dummy fallback. Maybe we can find that and use it if `fs` is not available for now, until port to `fsspec` happens.

Looks like I was wrong: https://github.com/behdad/ufoLib/commit/acfe65b312c3c7fe5e31b2040b406f145e396ebb

skef:
@behdad looks like Python 3.13 is raising inconvenient (for AFDKO) warnings from fs now:

```
../../py/lib/python3.13/site-packages/fs/__init__.py:4
  /home/skef/py/lib/python3.13/site-packages/fs/__init__.py:4: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
    __import__("pkg_resources").declare_namespace(__name__)  # type: ignore

../../py/lib/python3.13/site-packages/fs/__init__.py:4
[...]

PR #3871, opened 2025-06-23
Use fsspec instead of pyfilesystem2

madig:
Closes https://github.com/fonttools/fonttools/issues/3620.


madig:
I believe fsspec can write zip files in place, so UFOWriter could be simplified.

Edit: Or not, the writer has to make a zip file in a different location because it could be opened read-only by the reader, so writing to it would immediately truncate the file. Needs more thinking.

madig:
There is another solution: Drop UFOZ support and fs/fsspec along with it. It seems to be a very niche feature.

PR #3875, opened 2025-06-27
Annotate ufoLib

knutnergaard:


knutnergaard:
@anthrotype @madig I'm running into a type mismatch between the `glyphSet` parameter in `ufoLib.converters.convertUFO1OrUFO2KerningToUFO3Kerning`, which appear to expect Iterable[str] (or something along those lines), given the default value of `()` and `ufoLib.UFOReader.getGlyphSet`, which returns a `GlyphSet`. The conflict results in an error in this call:
```python
class UFOReader(_UFOBaseIO):
    def _upConvertKerning(self, validate: bool) -> None:
        if self._upConvertedKerningData:
            ...
        else:
            ...
            # convert kerning and groups
            kerning, groups, conversionMaps = convertUFO1OrUFO2KerningToUFO3Kerning(
                self._upConvertedKerningData["originalKerning"],
                deepcopy(self._upConvertedKerningData["originalGroups"]),
                self.getGlyphSet(),
            )
```
Should the last argument in the call to  `convertUFO1OrUFO2KerningToUFO3Kerning` actually be `self.getGlyphSet().keys()`? If not, and `glyphSet` is supposed to accept a `GlyphSet`, should't the default value for `glyphSet` be `None`?

jenskutilek:
`convertUFO1OrUFO2KerningToUFO3Kerning` only checks for the presence of glyph names in the glyph set, so I think the last argument should indeed be Iterable[str].

knutnergaard:
@jenskutilek I see, but this means that the last argument in the call should be `self.getGlyphSet().keys()` rather than the current `self.getGlyphSet()`.

justvanrossum:
The point of this argument is membership testing. This works for sets, dicts and lists (although porentially not efficiently for the latter), as well as anything that implements `__contains__`, such as `GlyphSet`. Therefore I think `collections.abc.Set` the the most appropriate annotation.

knutnergaard:
> `collections.abc.Set`

I Take it you mean `typing.Set`, to be able to specify items type (`Set[str]`), or am I missing something?

If using such a strict annotation, we'll need to cast the `glyphSet` argument to `set` in addition to ensuring it's an iterable of glyph names, i.e., `set(self.getGlyphSet().keys())`. Maybe this is a bit much?

Correction: Off course, dicts can be coerced to sets, so it's a bit simpler than the above, but I'm not sure that's what we should do.

jenskutilek:
`GlyphSet.keys()` is a bit weird, it casts to list instead of just returning `GlyphSet.contents.keys()`. Maybe a remnant from Python 2?

https://github.com/fonttools/fonttools/blob/75af36804fb595cabd768d2bc3ccf9b64445a014/Lib/fontTools/ufoLib/glifLib.py#L565-L568

knutnergaard:
@jenskutilek As you suspect, `dict.keys()` in Python 2 returned a list, so this is a way to ensure consistency across versions. I see no reason not to update it now, though.

justvanrossum:
> I Take it you mean typing.Set, to be able to specify items type (Set[str])

No I do not. We don't want the concrete type `set` (which `typing.Set` is a deprecated alias for), we want to check "does this object support membership testing", aka, does it implement `__contains__`.

https://docs.python.org/3/library/typing.html#typing.Set says:

> Note that to annotate arguments, it is preferred to use an abstract collection type such as [collections.abc.Set](https://docs.python.org/3/library/collections.abc.html#collections.abc.Set) rather than to use [set](https://docs.python.org/3/library/stdtypes.html#set) or [typing.Set](https://docs.python.org/3/library/typing.html#typing.Set).

justvanrossum:
Maybe `collections.abc.Set` is still too narrow. We really only want to check that the object supports `__contains__`.

knutnergaard:
> > I Take it you mean typing.Set, to be able to specify items type (Set[str])
> 
> No I do not. We don't want the concrete type `set` (which `typing.Set` is a deprecated alias for), we want to check "does this object support membership testing", aka, does it implement `__contains__`.
> 
> https://docs.python.org/3/library/typing.html#typing.Set says:
> 
> > Note that to annotate arguments, it is preferred to use an abstract collection type such as [collections.abc.Set](https://docs.python.org/3/library/collections.abc.html#collections.abc.Set) rather than to use [set](https://docs.python.org/3/library/stdtypes.html#set) or [typing.Set](https://docs.python.org/3/library/typing.html#typing.Set).

Thanks, You're absolutely right! My brain was stuck on Python 3.8.
If `__contains__` is the only requirement, a union type might be the way to go.

knutnergaard:
@justvanrossum There is always `collections.abc.Container` if we want to be totally generic.

knutnergaard:
Incidentally `GlyphSet` presents another type conflict in the `getUnicodes` and `getComponentReferences` methods, as `text` is defined as bytes  via `self.getGLIF(glyphName)`, before it's passed to `_fetchUnicodes`, which takes a string. At least that is what I can deduce from the code. 

Any idea how to solve this?

justvanrossum:
> There is always collections.abc.Container if we want to be totally generic.

Ah yes, that’s the perfect one for this argument.

Lib/fontTools/ufoLib/glifLib.py
```diff
@@ -597,7 +631,9 @@ def getUnicodes(self, glyphNames=None):
             unicodes[glyphName] = _fetchUnicodes(text)
```
knutnergaard, line 629: `text` is defined as `bytes` while `_fetchUnicodes` expects `str`. Not sure how to solve this.

Lib/fontTools/ufoLib/glifLib.py
```diff
@@ -612,7 +648,9 @@ def getComponentReferences(self, glyphNames=None):
             components[glyphName] = _fetchComponentBases(text)
```
knutnergaard, line 646: See [line 631](https://github.com/fonttools/fonttools/pull/3875/files#diff-a2a63773eb6cfb5dcd03686b70c892d6221ed12cde4fe4f9b8c9626641a7b57aR631).

knutnergaard:
Didn't bother with the deprecated `plistLib` module. Let me know if that should be annotated as well.

knutnergaard:
Still a few kinks to iron out. Reconverting to draft.

knutnergaard:
All Mypy checks and tests passing.

Lib/fontTools/ufoLib/glifLib.py
```diff
@@ -627,8 +626,9 @@ def getUnicodes(
         if glyphNames is None:
             glyphNames = self.contents.keys()
         for glyphName in glyphNames:
-            text = self.getGLIF(glyphName)
-            unicodes[glyphName] = _fetchUnicodes(text)
+            bytesText = self.getGLIF(glyphName)
+            strText = bytesText.decode(encoding="utf-8", errors="strict")
```
jenskutilek: Maybe `getGLIF()` should be changed to return a string directly?
It reads the plist file, which in theory could be another encoding than utf-8, but that case isn't handled in the current code either.
My hunch is that the current reading XML as bytes is another leftover from Python 2 to 3 migration that has not been given a second thought.

justvanrossum: I would not change the return type of `getGLIF()` as this is pretty much a public API.

I don't understand this change to begin with. The code worked before, so there must be some misunderstanding.

My guess is that the parser (as invoked by `_fetchUnicodes`) takes both `str` and `bytes`. Either we should annotate it as a union type, *or* annotate it as `bytes`, since that's what we appear to use.

justvanrossum: Indeed `_fetchUnidoes()` accepts both `str` and `bytes`:
```python
from fontTools.ufoLib.glifLib import _fetchUnicodes

print(_fetchUnicodes("<a></a>"))
print(_fetchUnicodes(b"<a></a>"))
```
Output:
```
[]
[]
```

justvanrossum: (I know for sure *my* code will break if `getGLIF()` starts returning `str`...)

justvanrossum: `bytes` actually makes sense, since the encoding is specified in the XML itself, and the parser will take care of that. `getGLIF()` simply returns the raw source data. I'm not sure the UFO standard requires .glif files to be UTF8.

knutnergaard: > I don't understand this change to begin with. The code worked before, so there must be some misunderstanding.
> 
> My guess is that the parser (as invoked by `_fetchUnicodes`) takes both `str` and `bytes`. Either we should annotate it as a union type, _or_ annotate it as `bytes`, since that's what we appear to use.

The change was due to the fact `xml.parsers.expat.ParserCreate.Parse`, called in `_fetchUnicodes` via `_BaseParser`, only accepts `str`, so Mypy was reporting this as a type conflict.

jenskutilek: > `bytes` actually makes sense, since the encoding is specified in the XML itself, and the parser will take care of that. `getGLIF()` simply returns the raw source data. I'm not sure the UFO standard requires .glif files to be UTF8.

Ah, you're right. I was confused by `_glifTreeFromString(aString)` seeminigly ignoring the encoding by calling `data = tobytes(aString, encoding="utf-8")`, but actually `tobytes` will pass bytes right through, so indeed the XML parser handles the encoding.

justvanrossum: > The change was due to the fact xml.parsers.expat.ParserCreate.Parse, called in _fetchUnicodes via _BaseParser, only accepts str, 

Where would that be declared? Because it wouldn't match reality.

knutnergaard: I thought I got it from the xml docs, but looking at it again, the `data` input type of `Parse` is `str | ReadableBuffer`, so I guess I must have been confused by the alias – which obviously represents a bytes-like object – as well as the variable name `text`, which in my mind would indicate a string. 

So, should the relevant types in `glifLib` be `bytes` or `Union[str, bytes]`?

justvanrossum: If `_fetchUnicodes()` only ever takes output of `getGLIF()`, then `bytes`.

knutnergaard: Done.

Lib/fontTools/ufoLib/utils.py
```diff
@@ -35,41 +43,61 @@ def wrapper(*args, **kwargs):
...
+    if isinstance(value, int):
+        return cls((value, 0))
+    if isinstance(value, tuple) and len(value) == 2:
+        return cls(value)
+    raise ValueError(f"Unsupported format version: {value!r}")
+
+
+# Base class for UFOFormatVersion and GLIFFormatVersion
+class BaseFormatVersion(tuple[int, int], enum.Enum):
+    value: Union[tuple[int, int]]
```
jenskutilek: Is Union necessary here?

knutnergaard: No, thanks, fixed.

Lib/fontTools/ufoLib/__init__.py
```diff
@@ -1644,12 +1719,13 @@ def renameGlyphSet(self, layerName, newLayerName, defaultLayer=False):
                 newLayerName, existing=existing, prefix="glyphs."
```
jenskutilek, line 1719: `newLayerName` may still be `None` at this point. The condition that `layerName` or `newLayerName` are `None` in the method call are not covered in the existing tests, so it's a bit hard to tell what the best solution is.

Lib/fontTools/ufoLib/__init__.py
```diff
@@ -1801,18 +1893,19 @@ def validateFontInfoVersion2ValueForAttribute(attr, value):
...
+        if valueOptions is not None:
+            isValidValue = validator(value, valueOptions)
+        # no specific options
         else:
-            isValidValue = validator(value)
+            if validator == genericTypeValidator:
+                isValidValue = validator(value, valueType)
+            else:
+                isValidValue = validator(value)
     return isValidValue
```
jenskutilek, line 1904: If `dataValidationDict.get("valueValidator")` returned `None`, `isValidValue` is unbound.

Where `fontInfoAttributesVersion2ValueData` is constructed, `genericValidator` is added to all entries that don't already have a `valueValidator`, so we could use it as a fallback here:

```py
validator = dataValidationDict.get("valueValidator", genericValidator)
```

Lib/fontTools/ufoLib/__init__.py
```diff
@@ -1847,18 +1940,19 @@ def validateFontInfoVersion3ValueForAttribute(attr, value):
...
```
jenskutilek, line 1950: Same as comment for `validateFontInfoVersion2ValueForAttribute`

jenskutilek (commented):
Thank you for your patience! I've added just a few comments.

Lib/fontTools/ufoLib/glifLib.py
```diff
@@ -12,19 +12,21 @@
 
 from __future__ import annotations
 
-import enum
 import logging
 from collections import OrderedDict
+from typing import TYPE_CHECKING, cast, Any, Optional, Union
+from lxml import etree
```
anthrotype, line 18: we don't want to unconditionally import lxml, it's an optional extra, that's why we have fontTools.misc.etree module, to fall back to the stdlib ElementTree when lxml is not installed!

anthrotype:
Thank you @knutnergaard for this, and also @jenskutilek for reviewing. However we have to make sure that these sweeping changes (which are meant to make the life of fonttools developers easier) do not impact the behavior of fonttools at runtime for regular users.

I noted there was one such unwanted change here: https://github.com/fonttools/fonttools/pull/3875#discussion_r2318551677 (the tests _should_ have caught that but apparently they didn't?)

I don't have the time to re-review this PR right now. I see Jens has already merged it to main. If Jens or someone else can take another look at this to make sure the typing-related changes only affect typing proper and there are no other changes that are not strictly related to typing, that would be great. Otherwise, I think it's safer that we revert this until we are confident there's no other runtime changes.

jenskutilek:
I did a thorough review and remember noticing that `lxml` import. But maybe I forgot to push my fix for it. I was on vacation for two weeks, and am working from another office today, so let me check tonight/tomorrow when I'm back at the machine where I reviewed this PR.

anthrotype:
thank you! There's no rush, we just need to fix things before a new release, which is not imminent from my side at least.

knutnergaard:
> Thank you @knutnergaard for this, and also @jenskutilek for reviewing. However we have to make sure that these sweeping changes (which are meant to make the live of fonttools developers easier) do not impact the behavior of fonttools at runtime for regular users.

I tried my best to be careful about this, and give notice of potential problems, but apparently I missed the import.

anthrotype:
it's ok!

the CI **is** also running `tox -e py-cov-noextra` which should _not_ install `lxml` among the rest of the `requirements.txt`, see:
https://github.com/fonttools/fonttools/blob/7928b8054febc40f3d365f310746eb7275422db1/.github/workflows/test.yml#L57-L58
https://github.com/fonttools/fonttools/blob/7928b8054febc40f3d365f310746eb7275422db1/tox.ini#L24

I actually just tried to run that command on my machine, and indeed it did not fail as it should have. pytest is simply skipping the test altogether:

```
SKIPPED [22] Tests/ufoLib/__init__.py:3: could not import 'fontTools.ufoLib': No module named 'lxml'
```

anthrotype:
this is too broad and should be narrowed down a bit:

https://github.com/fonttools/fonttools/blob/7928b8054febc40f3d365f310746eb7275422db1/Tests/ufoLib/__init__.py#L1-L3

jenskutilek:
> this is too broad and should be narrowed down a bit:

I guess that was put in because of the `ufo` extra (which installed `fs` as a dependency). So now that `fs` has been replaced, shouldn't the `importorskip` removed completely?

anthrotype:
> shouldn't the importorskip removed completely

i suppose, yes. If there are lxml-specific tests elsewhere in ufoLib, those should have their own conditional skip decorators

jenskutilek:
I found my missing additions to the PR ... I had made a PR to @knutnergaard's fork (https://github.com/knutnergaard/fonttools/pull/1), but that is a mess to merge now. I collected the changes and also another fix for missing `lxml`, and removed the `importorskip` from the ufoLib tests in a new PR in #3925.

Issue #3639, opened 2024-09-17
Support hvgl table


moi15moi:
macOS 15 Sequoia includes a new version of `PingFangUI.ttc` that contains a table with the tag `hvgl`. As far as I know, Apple has not yet documented this table, so there’s not much we can do about it **at the moment**.

For reference, here’s the corresponding [FreeType issue](https://gitlab.freedesktop.org/freetype/freetype/-/issues/1281).

justvanrossum:
Via @bungeman, https://github.com/googlefonts/fontations/issues/1369

> The hvgl table was introduced by Apple with a new PingFang font family (PingFangUI.ttc). There isn't yet any official documentation of the format.

> See also [FreeType issue](https://gitlab.freedesktop.org/freetype/freetype/-/issues/1281) and [Chromium issue](https://issues.chromium.org/issues/380313555).

davelab6:
https://developer.apple.com/documentation/hvf?language=objc is official documentation :)

PR #3876, opened 2025-07-03
Implement Apple `hvgl` table

behdad:
https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6hvgl.html


Lib/fontTools/misc/transform.py
```diff
@@ -216,7 +216,7 @@ def transformVectors(self, vectors):
         xx, xy, yx, yy = self[:4]
         return [(xx * dx + yx * dy, xy * dx + yy * dy) for dx, dy in vectors]
 
-    def translate(self, x: float = 0, y: float = 0):
+    def translate(self, x: float = 0, y: float = 0, before=False):
```
justvanrossum: The same could be achieved by using the transform objects differently. I don’t quite like the added complexity.

behdad: I'll remove and move to callsites. I personally prefer it this way because the callsites are then simplified.

behdad: Thanks for the review.

Lib/fontTools/ttLib/sfnt.py
```diff
@@ -269,33 +269,43 @@ def setEntry(self, tag, entry):
...
         entry = self.DirectoryEntry()
         entry.tag = tag
-        entry.offset = self.nextTableOffset
         if tag == "head":
             entry.checkSum = calcChecksum(data[:8] + b"\0\0\0\0" + data[12:])
             self.headTable = data
             entry.uncompressed = True
         else:
             entry.checkSum = calcChecksum(data)
+        from fontTools.ttLib import getTableClass
```
anthrotype: I presume this import is placed in here to avoid import cycles; perhas move it to the first line of the `__setitem__` method where I'd expect to find method-related imports

Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -488,3 +488,36 @@ def draw(self, pen):
...
+
+    # Load font and draw all glyphs
+    from fontTools.ttLib import TTFont
+    from fontTools.pens.basePen import NullPen
+
+    if len(sys.argv) < 2:
+        print("Usage: ttGlyphSet.py [-y fontNumber] [-n numGlyphs] fontfile.ttf")
+        sys.exit(1)
+
+    # Convert to argparse reading
```
anthrotype: leftovers?

Lib/fontTools/varLib/hvar.py
```diff
@@ -56,7 +56,7 @@ def add_HVAR(font):
 def add_VVAR(font):
     if "VVAR" in font:
         del font["VVAR"]
-    getAdvanceMetrics = partial(_get_advance_metrics, font, axisTags, HVAR_FIELDS)
+    getAdvanceMetrics = partial(_get_advance_metrics, font, axisTags, VVAR_FIELDS)
```
anthrotype: this should arguably go into a separate PR

Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -427,12 +428,10 @@ def toXML(self, xmlWriter, font, attrs, name):
...
-        content = [c for c in content if isinstance(c, tuple)]
-        for name, attrs, content in content:
-            assert name == "VarComponent"
-            component = VarComponent()
-            component.fromXML(name, attrs, content, font)
-            self.components.append(component)
+        assert name == "VarComponent"
+        component = VarComponent()
+        component.fromXML(name, attrs, content, font)
+        self.components.append(component)
```
anthrotype, line 434: I don't understand this change -- besides the fact that it doesn't seem to belong to HVF proper...
Deserialization of VariableCompositeGlyph elements was working before, or not? What else changed?

anthrotype: oh nevermind, I see in otConverters the IndexBase.xmlRead changed (VarCompositeGlyphList is a special CFF2Index)

NightFurySL2001:
Will closes #3639 .

Lib/fontTools/varLib/hvar.py
behdad: Done: https://github.com/fonttools/fonttools/pull/3884

Lib/fontTools/ttLib/ttGlyphSet.py
behdad: Haha. Yeah, prompt to get Copilot to do the work...

Lib/fontTools/ttLib/sfnt.py
behdad: Done. Although, the rest of the file just imports close to use.

Lib/fontTools/misc/transform.py
behdad: Do you think I should add `Rotation` and `RotationAroundCenter` free functions?

behdad: > Do you think I should add `Rotation` and `RotationAroundCenter` free functions?

Currently we have `Offset` (which I can also redefine as `Translation` if desired) and `Scale`. I can add `Skew` as well for completeness.

justvanrossum: Maybe yes (for Translation and Skew). 

Not sure about RotationAroundCenter, but if that helps you, why not.

behdad: Thanks. I'm gonna add Rotation that takes optional `tx` and `ty`.

behdad:
The composite rendering is wrong in certain cases. I'm working on it. A simple example is the ASCII asterisk `*`.

behdad:
> The composite rendering is wrong in certain cases. I'm working on it. A simple example is the ASCII asterisk `*`.

Fixed. Now onto addressing Just's request.

Lib/fontTools/misc/transform.py
behdad: Fixed. I ended up adding `pre_translate` since it was used in multiple places in HVF code and is much faster to code instead of doing `Translation(x, y).transform(t)`.

behdad:
I think this is ready for further testing / review now. Thanks.

Lib/fontTools/ttLib/ttGlyphSet.py
```diff
@@ -488,3 +488,35 @@ def draw(self, pen):
...
+
+    import sys
+
+    # Load font and draw all glyphs
+    from fontTools.ttLib import TTFont
+    from fontTools.pens.basePen import NullPen
+
+    if len(sys.argv) < 2:
+        print("Usage: ttGlyphSet.py [-y fontNumber] [-n numGlyphs] fontfile.ttf")
+        sys.exit(1)
```
khaledhosny: You should not need this block now that you are using argparse.

behdad: Thanks.

behdad:
HB branch merged: https://github.com/harfbuzz/harfbuzz/pull/5379

behdad:
@anthrotype We landed the HarfBuzz PR. May I merge this?

anthrotype:
Sorry I didn't have the time to review but sure, feel free to merge if you believe it's ready, we can address any issues as they come up. Thanks

behdad:
Thanks!

PR #3879, opened 2025-07-03
Micro optimizations in ttLib

jenskutilek:
Mainly simplified for-loops.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -1,37 +1,34 @@
 """_g_l_y_f.py -- Converter classes for the 'glyf' table."""
 
-from collections import namedtuple
-from fontTools.misc import sstruct
-from fontTools import ttLib
-from fontTools import version
-from fontTools.misc.transform import DecomposedTransform
-from fontTools.misc.textTools import tostr, safeEval, pad
-from fontTools.misc.arrayTools import updateBounds, pointInRect
-from fontTools.misc.bezierTools import calcQuadraticBounds
-from fontTools.misc.fixedTools import (
-    fixedToFloat as fi2fl,
-    floatToFixed as fl2fi,
-    floatToFixedToStr as fl2str,
-    strToFixedToFloat as str2fl,
-)
-from fontTools.misc.roundTools import noRound, otRound
-from fontTools.misc.vector import Vector
-from numbers import Number
-from . import DefaultTable
-from . import ttProgram
-import sys
-import struct
 import array
 import logging
 import math
 import os
-from fontTools.misc import xmlWriter
-from fontTools.misc.filenames import userNameToFileName
-from fontTools.misc.loggingTools import deprecateFunction
+import struct
+import sys
+from collections import namedtuple
 from enum import IntFlag
 from functools import partial
+from numbers import Number
 from types import SimpleNamespace
 from typing import Set
+
+from fontTools import ttLib, version
+from fontTools.misc import sstruct, xmlWriter
+from fontTools.misc.arrayTools import pointInRect, updateBounds
+from fontTools.misc.bezierTools import calcQuadraticBounds
+from fontTools.misc.filenames import userNameToFileName
+from fontTools.misc.fixedTools import fixedToFloat as fi2fl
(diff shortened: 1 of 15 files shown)


justvanrossum:
The formatting and import changes distract a lot from the actual changes, making review harder than needed.

madig:
Agreed. Though, if a separate formatting and import sorting PR (preferably of the whole codebase to get it over with) is too much trouble, you can at least go through the changes commit by commit.

jenskutilek:
I will revert the sorting/formatting changes tomorrow. Thank you for your patience :)

Lib/fontTools/ttLib/tables/_g_l_y_f.py
```diff
@@ -2037,8 +2036,8 @@ def toInt(self, *, round=otRound):
         if round is noRound:
             return
         a = self._a
-        for i in range(len(a)):
-            a[i] = round(a[i])
+        for i, value in enumerate(a):
+            a[i] = round(value)
```
anthrotype, line 2040: note that GlyphsCoordinate has been heavily perf-optimized by behdad, I'm not sure if using this seemingly more idiomatic style of iteration is actually better here.
I think for a tight loop, the `for i in range(len(a))` approach would be typically faster for a Python `array.array`, as the enumerate() function introduces some small overhead by creating an enumerator object and yielding the tuple (index, value). Maybe don't touch GlyphCoordinates methods.

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -260,7 +260,7 @@ def _save(self, file, tableCache=None):
             ]  # make sure 'head' is loaded so the recalculation is actually done
 
         tags = list(self.keys())
-        if "GlyphOrder" in tags:
+        if "GlyphOrder" in self.keys():
```
anthrotype: TTFont.keys() does a bunch of things that you don't need here to check if a tag is present, please revert to the old approach here and elsewhere you used this patterm

Lib/fontTools/ttLib/woff2.py
```diff
@@ -642,9 +641,8 @@ def reordersTables(self):
 
 def getKnownTagIndex(tag):
     """Return index of 'tag' in woff2KnownTags list. Return 63 if not found."""
-    for i in range(len(woff2KnownTags)):
-        if tag == woff2KnownTags[i]:
-            return i
+    if tag in woff2KnownTags:
+        return woff2KnownTags.index(tag)
```
anthrotype: or you could except ValueError (raised by index method) and return woff2UnknownTagIndex

Lib/fontTools/ttLib/tables/_g_l_y_f.py
jenskutilek: Hm, I ran some (limited) timing tests and `enumerate` seemed a little bit faster than `range` + `__getitem__` for lists and arrays alike, but it also seems to largely depend on what actually happens in the loop.

[compare.py.txt](https://github.com/user-attachments/files/21054492/compare.py.txt)

jenskutilek:
```sh
$ python compare.py
list range: 0.10156416893005371
list enum:  0.09562277793884277
list in:    0.06502509117126465
array range: 0.10119390487670898
array enum:  0.09602022171020508
array in:    0.0670781135559082
```

Lib/fontTools/ttLib/ttFont.py
jenskutilek: Actually "GlyphOrder" is always present, as TTFont.keys() ensures that it is the first item in the returned list.

I've adapted the code to take advantage of this.

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -259,9 +259,7 @@ def _save(self, file, tableCache=None):
                 "head"
             ]  # make sure 'head' is loaded so the recalculation is actually done
 
-        tags = list(self.keys())
-        if "GlyphOrder" in self.keys():
-            tags.remove("GlyphOrder")
+        tags = self.keys()[1:]  # skip GlyphOrder tag
```
anthrotype: or simply `tags.pop(0)` to avoid creating a new list

PR #3881, opened 2025-07-04
Autoformat imports with isort

jenskutilek:
Automatic application of coding styles makes my life easier a lot.

Split out from #3879, this applies import sorting for everything. After that I've run black again to iron out any formatting differences (we run black already).

I've applied both tools with their default settings:

```sh
$ isort .
$ black .
```

### DIFF
--- a/Lib/fontTools/misc/etree.py
+++ b/Lib/fontTools/misc/etree.py
@@ -14,7 +14,6 @@ iterwalk.
 
 from fontTools.misc.textTools import tostr
 
-
 XML_DECLARATION = """<?xml version='1.0' encoding='%s'?>"""
 
 __all__ = [
@@ -208,7 +207,7 @@ except ImportError:
     #   Char ::= #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]
     # Here we reversed the pattern to match only the invalid characters.
     _invalid_xml_string = re.compile(
-        "[\u0000-\u0008\u000B-\u000C\u000E-\u001F\uD800-\uDFFF\uFFFE-\uFFFF]"
+        "[\u0000-\u0008\u000b-\u000c\u000e-\u001f\ud800-\udfff\ufffe-\uffff]"
     )
 
     def _tounicode(s):
--- a/Lib/fontTools/__init__.py
+++ b/Lib/fontTools/__init__.py
@@ -1,4 +1,5 @@
 import logging
+
 from fontTools.misc.loggingTools import configLogger
 
 log = logging.getLogger(__name__)
--- a/Lib/fontTools/afmLib.py
+++ b/Lib/fontTools/afmLib.py
@@ -7,36 +7,36 @@ implement the whole Adobe AFM specification [#f1]_ but, it should read most
 
 Here is an example of using `afmLib` to read, modify and write an AFM file:
 
-	>>> from fontTools.afmLib import AFM
-	>>> f = AFM("Tests/afmLib/data/TestAFM.afm")
-	>>>
-	>>> # Accessing a pair gets you the kern value
-	>>> f[("V","A")]
-	-60
-	>>>
-	>>> # Accessing a glyph name gets you metrics
-	>>> f["A"]
-	(65, 668, (8, -25, 660, 666))
-	>>> # (charnum, width, bounding box)
-	>>>
-	>>> # Accessing an attribute gets you metadata
-	>>> f.FontName
-	'TestFont-Regular'
-	>>> f.FamilyName
-	'TestFont'
-	>>> f.Weight
-	'Regular'
-	>>> f.XHeight
-	500
-	>>> f.Ascender
-	750
-	>>>
-	>>> # Attributes and items can also be set
-	>>> f[("A","V")] = -150 # Tighten kerning
-	>>> f.FontName = "TestFont Squished"
-	>>>
-	>>> # And the font written out again (remove the # in front)
-	>>> #f.write("testfont-squished.afm")
+    >>> from fontTools.afmLib import AFM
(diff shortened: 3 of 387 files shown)


justvanrossum:
A big downside is that this makes merging open PRs a lot harder.

Additionally we need to set up pre-commit or something like it to ensure things keep the expected formatting.

On the one hand I'd be highly in favor of reformatting, but I think we need a broad concensus among the regular committers.

anthrotype:
we have a `tox -e lint` task that is run by the CI on PRs which already runs `black --check --diff` to check that (new) things are properly formatted, so we could do the same with isort I guess? Although this probably would only apply to the newly added/modified files. This PR on the other hand is applying these tools to the entire codebase, some very old files which haven't been touched in ages. By the way, didn't we run black on all files before setting up the CI job? I think we did, but of course black itself has changed..

There's the issue of open PRs becoming unmergeable that Just mentioned, but it could be worked around by running the same automatic tools on those PR branches to fix the merge conflicts I presume.

Running black should be safe, but for isort we need to make sure that the reodering doesn't mess up the functionality for module imports that have side effects, e.g. importing fontTools.subset adds the subset methods to table classes, or importing otTables defines the otData classes etc.

anthrotype:
I think you also need to tell isort to follow black's profile otherwise they'll step over each other:

https://pycqa.github.io/isort/docs/configuration/black_compatibility.html

anthrotype:
@madig added these to ufoLib2's pyproject.toml, have a look

https://github.com/fonttools/ufoLib2/blob/5db1b88ca7894cc41215af664a00be39ab90b08c/pyproject.toml#L53-L57

jenskutilek:
>By the way, didn't we run black on all files before setting up the CI job? I think we did, but of course black itself has changed..

Yes, we did. I ran black after isort again only to clean up some things that isort formats differently, which black would correct then again, e.g. isort’s default style for wrapping long import lines is

```python
from fontTools.cffLib import (TopDictIndex, buildDefaults, buildOrder,
                              privateDictOperators, topDictOperators)
```

while black does

```python
from fontTools.cffLib import (
    TopDictIndex,
    buildDefaults,
    buildOrder,
    privateDictOperators,
    topDictOperators,
)
```

I see now that I could have avoided that by using the isort black profile. So this is really only about adding isort.

anthrotype:
> Although this probably would only apply to the newly added/modified files.

turns out I was wrong, I thought black's --diff flag which we use in tox -e lint would only run black on the diff (i.e. added/modified) but it actually means "show the unified diff in the console output".

If we go with isort, we should add something like `isort --profile black --skip-gitignore --check-only --diff .` to the `[testenv:lint]` section of tox.ini so that it gets run on PRs, similar to the exiting `black --check --diff .`

madig:
If you're going to change tools around, I highly recommend replacing black and isort with just ruff. It is just better and faster.

I also see the existing PR problem pointed out by Just, but unless someone goes through the trouble of reviewing and merging all existing PRs, I don't see how to make it go away. The only alternative I can think of would be to not autoformat. I'd see it as a change for all future PRs.

Oh, also, don't forget to add the formatting commit to https://github.com/fonttools/fonttools/blob/main/.git-blame-ignore-revs to keep the Blame view working well.

anthrotype:
> I highly recommend replacing black and isort with just ruff. It is just better and faster.

+1 to just switch to ruff (I just realized that's what my vscode is also using to autoformat on save)

anthrotype:
technically ruff is not 100% identical to black (here's the list of known deviation https://docs.astral.sh/ruff/formatter/black/) but it's close enough for me not to even realise I was using that instead of black

anthrotype:
> The only alternative I can think of would be to not autoformat. I'd see it as a change for all future PRs.

we are already autoformatting with black, as Jens said, this PR only adds isort (PR title should be renamed for clarity). He ran black after isort because he had not set up isort to output black-compatible output.

Running isort only on future PRs for newly added or modified files is a bit impractical. The GH action would have to first identify all the newly added or changed files (e.g. using https://github.com/tj-actions/changed-files) and then run isort on those only.

Lib/fontTools/misc/etree.py
```diff
@@ -208,7 +207,7 @@ def tostring(
     #   Char ::= #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]
     # Here we reversed the pattern to match only the invalid characters.
     _invalid_xml_string = re.compile(
-        "[\u0000-\u0008\u000B-\u000C\u000E-\u001F\uD800-\uDFFF\uFFFE-\uFFFF]"
+        "[\u0000-\u0008\u000b-\u000c\u000e-\u001f\ud800-\udfff\ufffe-\uffff]"
```
khaledhosny, line 210: Is there away to disable this hex lower casing? I find `\udfff` to be less readable than `\uDFFF`.

anthrotype: ugh.. looks like it's a new thing they added with https://github.com/psf/black/pull/2916

and black being black, you can't disable it :(

anthrotype: ruff does the same so that's the canonical way we have to get used to i'm afraid
https://github.com/astral-sh/ruff/issues/13859

khaledhosny: 😢

anthrotype: they probably wanted to make it consistent with the `repr()` of these strings which is all lowercase. But it's inconsistent with the way they make hex digits all uppercase:

```python
0xffff  # == black ==> 0xFFFF
"\uFFFF"  # == black ==> "\uffff"
```

madig:
Here's the ruff version of this PR: https://github.com/fonttools/fonttools/compare/main...format-everything-mk2.

PR #3885, opened 2025-07-10
remove hard-dependency on pyfilesystem2, add a minimal drop-in replacement

anthrotype:
This adds a new fontTools.misc.filesystem package that exposes the same API as pyfilesystem2 but only for the subset that fontTools.ufoLib uses: i.e. OSFS, SubFS, ZipFS, TempFS and other little helpers.

the requirement in setup.py was removed, so pip install fonttools[ufo] will no longer instal fs. However, the code continues to work with fs if one installs it separately and wishes to pass actual FS objects to ufoLib.

Similarly to what we do with lxml, I try to import fs and if that fails I fall back to this new fontTools.misc.filesystem. No other changes inside the fontTools.ufoLib code itself besides the try/import.

The CI runs test with and without fs. I think it's ok to continue supporting fs as a soft/optional dependency given that it still works (despite the pkg_resources deprecation warning), and it may be resurrected at some point.

But if one doesn't care about fancy fileststems but just UFO as local directories or zip, the fs module is no longer required basically and the warning shoud stop showing up.


anthrotype:
OSFS is just a thin wrapper around pathlib.Path, whereas ZipFS uses zipfile to open and read specific files within the archive (without unpacking it all); for the zip writer, it's just a TempFS that calls shutil.make_archive on close() (not dissimilar from the original fs.ZipFS).

anthrotype:
ok this is ready for review now. I tested ufoLib2 and defcon test suite and they just work with minor modification, notably stuff like

```diff
- import fs
+ from fontTools.misc import filesystem as fs
```

Lib/fontTools/misc/filesystem/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+when this is available. Otherwise it falls back to the replacement modules within this package.
+
+Client code that wishes to do the same could replace direct `import fs` with e.g.
+`from fontTools.misc import filesystem as fs` or equivalent.
+"""
+
+from __future__ import annotations
+
+try:
+    __import__("fs")
```
madig, line 22: Why not `import fs`?

Lib/fontTools/misc/filesystem/_base.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def isclosed(self) -> bool:
+        return self._closed
+
+    def __enter__(self) -> Self:
+        return self
+
+    def __exit__(self, exc_type, exc, tb):
+        self.close()
+        return False  # never swallow exceptions
```
madig, line 80: Why, what happened?

Lib/fontTools/misc/filesystem/__init__.py
anthrotype: Because I don't want to bind a variable named `fs` in the current scope

Lib/fontTools/misc/filesystem/_base.py
anthrotype: Just a more explicit way to tell context manager to not suppress exceptions occurring within the with statement (would work the same without return statement actually)

Lib/fontTools/misc/filesystem/_errors.py
```diff
@@ -0,0 +1,54 @@
+class FSError(Exception):
```
madig, line 1: Random thought: client code that tries to catch fs exceptions may need to be updated to import errors in a way that imports these ones when `fs` is not available 🤔

anthrotype: yes that's exactly why to support this they'll have to change all their `import fs` with `from fontTools.misc import filesystem as fs`

anthrotype: there is no other way I can think of to make this work smoothly

anthrotype: note that if client code is catching errors from the upstream `fs` module, they can continue to support this without any changes by simply making sure they install `fs` alongside fonttools (it'll be no longer required with `fonttools[ufo]` extra)

Lib/fontTools/misc/filesystem/_osfs.py
```diff
@@ -0,0 +1,140 @@
...
+            if e.errno == errno.ENOTEMPTY:
+                raise DirectoryNotEmpty(f"Directory not empty: {path!r}")
+            else:
+                raise ResourceError(f"unable to remove {path!r}: {e}")
+
+    def removetree(self, path: str):
+        shutil.rmtree(self._abs(path))
+
+    def movedir(self, src: str, dst: str, create: bool = False):
+        # TODO create
```
madig: You mean the move-to dir must be created if it doesn't exist?

Lib/fontTools/misc/filesystem/_errors.py
anthrotype: basically, clients can either add `fs` to their own direct dependencies and keep the code unchanged, or modify the import statements and then they can work with/without `fs` just like fonttools itself

Lib/fontTools/misc/filesystem/_path.py
```diff
@@ -0,0 +1,71 @@
...
+
+def forcedir(path: str) -> str:
+    # Ensure the path ends with a trailing forward slash.
+    if not path.endswith("/"):
+        return path + "/"
+    return path
+
+
+def abspath(path: str) -> str:
+    # Since FS objects have no concept of a *current directory*, this
```
madig: Want to turn these into docstrings?

Lib/fontTools/misc/filesystem/_osfs.py
anthrotype: good catch. I didn't implement support for the create flag because it's only used once in ufoLib with create=True and it already works for our own use case. It's a minor thing, I can implement if needed (basically create=False would fail if `dst` is not already there)

Lib/fontTools/misc/filesystem/_path.py
anthrotype: I don't want to encourage third party use of this so no

requirements.txt
```diff
@@ -7,6 +7,8 @@ scipy==1.13.1; platform_python_implementation != "PyPy" and python_version <= '3
 scipy==1.14.1; platform_python_implementation != "PyPy" and python_version >= '3.10'
 munkres==1.1.4; platform_python_implementation == "PyPy"
 zopfli==0.2.3.post1
+# fs is no longer required by the 'ufo' extra (only used if present) but we keep it in
+# here so that tox continues to install it for the tests
```
madig, line 11: But we also need a run without fs, right?

anthrotype: we do. the CI runs tox py-noextra as well

madig:
Where did you get all the implementation code from? Did you copy-paste and prune fs's code?

Lib/fontTools/misc/filesystem/_errors.py
madig: Fair. You probably want to spell this out in the init.

anthrotype:
well not literal copy paste, more like taking "inspiration" while aiming for the most concise approach to keep things working as before while using stdlib-only methods.. I could add citation links to original source code if you like.

EDIT: I added original copyright and license notice to LICENSE.external file: https://github.com/fonttools/fonttools/commit/6580a4283d4adc51a41f6aaf44ee398869ebfcd8

Lib/fontTools/misc/filesystem/_osfs.py
madig: Hm, no need then. Maybe just immortalise your comment in the code.

madig:
No need, I just wondered how you hammered all that code out so quickly.

madig:
Oh BTW, you may want to run `uv run ruff check --select I --fix . ; uv run ruff check --fix ; uv run ruff format` over this code, just in case.

Lib/fontTools/ufoLib/__init__.py
```diff
@@ -32,20 +32,21 @@
...
 from fontTools.misc import filesystem as fs
 from fontTools.misc import plistlib
-from fontTools.ufoLib.validators import *
-from fontTools.ufoLib.filenames import userNameToFileName
 from fontTools.ufoLib.converters import convertUFO1OrUFO2KerningToUFO3Kerning
 from fontTools.ufoLib.errors import UFOLibError
-from fontTools.ufoLib.utils import numberTypes, _VersionTupleEnumMixin
+from fontTools.ufoLib.filenames import userNameToFileName
+from fontTools.ufoLib.utils import _VersionTupleEnumMixin, numberTypes
+from fontTools.ufoLib.validators import *
```
madig, line 49: Actually, maybe just import pngValidator?

anthrotype: yeah ruff didn't like the star-import..

Lib/fontTools/misc/filesystem/_osfs.py
anthrotype: @madig fwiw I ended up implementing the TODO https://github.com/fonttools/fonttools/pull/3885/commits/4c4474b8972d91aa98556966277d936942127791

anthrotype: of course it failed on windows, I should have left it alone lol
https://github.com/fonttools/fonttools/actions/runs/16299042215/job/46028533829#step:6:451

PR #3886, opened 2025-07-17
Cff2cff improvements

behdad:
Improve CFF2ToCFF a bit.

Doesn't touch CFFToCFF2 yet. That's much more broken. https://github.com/fonttools/fonttools/issues/3521

### DIFF
--- a/Lib/fontTools/cffLib/CFF2ToCFF.py
+++ b/Lib/fontTools/cffLib/CFF2ToCFF.py
@@ -111,9 +111,7 @@ def _convertCFF2ToCFF(cff, otFont):
         mapping[name]: v for name, v in charStrings.charStrings.items()
     }
 
-    # I'm not sure why the following is *not* necessary. And it breaks
-    # the output if I add it.
-    # topDict.ROS = ("Adobe", "Identity", 0)
+    topDict.ROS = ("Adobe", "Identity", 0)
 
 
 def convertCFF2ToCFF(font, *, updatePostTable=True):
--- a/Lib/fontTools/cffLib/CFF2ToCFF.py
+++ b/Lib/fontTools/cffLib/CFF2ToCFF.py
@@ -101,11 +101,11 @@ def _convertCFF2ToCFF(cff, otFont):
             cs.program.insert(0, width - private.nominalWidthX)
 
     mapping = {
-        name: ("cid" + str(n) if n else ".notdef")
+        name: ("cid" + str(n).zfill(5) if n else ".notdef")
         for n, name in enumerate(topDict.charset)
     }
     topDict.charset = [
-        "cid" + str(n) if n else ".notdef" for n in range(len(topDict.charset))
+        "cid" + str(n).zfill(5) if n else ".notdef" for n in range(len(topDict.charset))
     ]
     charStrings.charStrings = {
         mapping[name]: v for name, v in charStrings.charStrings.items()
--- a/Lib/fontTools/cffLib/CFF2ToCFF.py
+++ b/Lib/fontTools/cffLib/CFF2ToCFF.py
@@ -8,6 +8,7 @@ from fontTools.cffLib import (
     buildDefaults,
     topDictOperators,
     privateDictOperators,
+    FDSelect,
 )
 from .specializer import specializeProgram
 from .width import optimizeWidths
@@ -52,8 +53,13 @@ def _convertCFF2ToCFF(cff, otFont):
             if hasattr(topDict, key):
                 delattr(topDict, key)
 
-    fdArray = topDict.FDArray
     charStrings = topDict.CharStrings
+
+    fdArray = topDict.FDArray
+    if not hasattr(topDict, "FDSelect"):
+        # FDSelect is optional in CFF2, but required in CFF.
+        fdSelect = topDict.FDSelect = FDSelect()
+        fdSelect.gidArray = [0] * len(charStrings.charStrings)
 
     defaults = buildDefaults(privateDictOperators)
     order = buildOrder(privateDictOperators)
--- a/Lib/fontTools/cffLib/CFF2ToCFF.py
+++ b/Lib/fontTools/cffLib/CFF2ToCFF.py
@@ -9,6 +9,7 @@ from fontTools.cffLib import (
     topDictOperators,
     privateDictOperators,
 )
+from .specializer import specializeProgram
 from .width import optimizeWidths
 from collections import defaultdict
 import logging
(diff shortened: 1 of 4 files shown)


behdad:
One thing is, we don't account for the lower stack size of CFF. Also the return/endchar. So:
- We should desubroutinize,
- Add endchar
- Specialize (via generalization first) to get the stack size right.

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -1720,6 +1727,36 @@ def instantiateVariableFont(
...
+
+        # Save these
+        recalcTimestamp = (varfont.recalcTimestamp,)
+        recalcBBoxes = varfont.recalcBBoxes
+
+        # Disable them
+        varfont.recalcTimestamp = False
+        varfont.recalcBBoxes = False
+
+        # Save to memory, reload, downgrade and save again, reload.
```
anthrotype: why do you need to do that?

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -1929,10 +1966,15 @@ def main(args=None):
         if limit is None or limit[0] == limit[2]
     }.issuperset(axis.axisTag for axis in varfont["fvar"].axes)
 
-    instantiateVariableFont(
+    inplace = True
+
+    if options.downgrade_cff2 and isFullInstance:
+        inplace = False
```
anthrotype: why the downgrading cannot be done inplace?

behdad: Because the glyph names are gonna all change from whatever they are, to `cid00XXX`. So we need all other tables to be in binary form and not loaded and only loaded with new glyph names. I don't know of an easier way to do it.

It can't be done inplace because this is Python and I need a `varfont = TTFont(...)` so it will not modify the input varfont.

Lib/fontTools/varLib/instancer/__init__.py
behdad: Because the glyph names are gonna all change from whatever they are, to cid00XXX. So we need all other tables to be in binary form and not loaded and only loaded with new glyph names. I don't know of an easier way to do it.

Lib/fontTools/varLib/instancer/__init__.py
behdad: I can, of course, copy the __dict__ of the final varfont into the original one if `inplace`. Let me know if you prefer that.

anthrotype: > I can, of course, copy the dict of the final varfont into the original one if inplace

you mean, work temporarily on the copy and at the end, if was inplace, overwrite the original TTFont? I guess from a user stantpoint if I tell you to modify the input font I expect it to do so regardless of other options so yeah

Lib/fontTools/varLib/instancer/__init__.py
anthrotype: maybe add a comment about that?

Lib/fontTools/varLib/instancer/__init__.py
behdad: PTAL.

Lib/fontTools/varLib/instancer/__init__.py
behdad: Done. PTAL.

Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -1377,6 +1383,44 @@ def _isValidAvarSegmentMap(axisTag, segmentMap):
...
+    stream.seek(0)
+    varfont = TTFont(stream, recalcTimestamp=False, recalcBBoxes=False)
+
+    convertCFF2ToCFF(varfont)
+
+    stream = io.BytesIO()
+    log.info("Saving downgraded CFF font to memory")
+    varfont.save(stream)
+    stream.seek(0)
+    varfont = TTFont(stream, recalcTimestamp=False, recalcBBoxes=False)
```
khaledhosny, line 1415: It would be nice to verify now that the second save is not needed.

khaledhosny (approved):

Lib/fontTools/varLib/instancer/__init__.py
behdad: Okay. I'll do some testing.

behdad: Did some testing. It's currently necessary. See comments.

khaledhosny: Good to know.

behdad: I might fight it again, but I think this is better than current situation at least.

khaledhosny: I think it is fine. Slow but correct is certainly better than the current broken state.

Lib/fontTools/cffLib/CFF2ToCFF.py
```diff
@@ -69,14 +76,15 @@ def _convertCFF2ToCFF(cff, otFont):
                 if hasattr(privateDict, key):
                     delattr(privateDict, key)
 
+    cff.desubroutinize()
```
anthrotype: So downgrading requires to de-subroutinize (remove and expand subroutines), and the resulting CFF 1.0 is only 'specialized' but not fully subroutinized. Maybe we should make that clearer in the fonttools varLib.instancer --help section

behdad: Maybe I can do something better here:

- Walk all charstrings and compute max stack use of each; only if it exceeds CFF1's stack limit (48), then: desubroutinize & specialize that charstring, to address the stack use.
- If any charstring went through such transformation, do a round of unused-subroutine-elimination.

I like it. A bit more work and needs more testing, but worth it IMO.

behdad: Humm. Looks like we were desubroutinizing in instancer anyway. That's necessary because we rewrite the charstrings to remove / modify the blend ops. So, this is not a new limitation for the instancer, just for the downgrader. I see about still implementing the new logic.

behdad: I swear somewhere I had written code to calculate stack use of a charstring. But can't find it :((

behdad: Done. Lightly tested.

behdad:
@anthrotype Is this good to merge? Thanks.

PR #3894, opened 2025-07-29
Recompute `OS/2.usMaxContext` when adding feature variations

madig:
A font might not have had a GSUB or GPOS table before, but does after addFeatureVariations, so we need to recompute OS/2.usMaxContext.

To do: 
- [x] ~~Don't assume `OS/2` exists?~~ https://learn.microsoft.com/en-us/typography/opentype/spec/otff#required-tables says it's required, so I'll assume it.
- [x] Write test.

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -926,6 +926,11 @@ class Builder(object):
                     l.lookup_index for l in lookups if l.lookup_index is not None
                 )
             )
+            # order doesn't matter, but lookup_indices preserves it.
+            # We want to combine identical sets of lookups (order doesn't matter)
+            # but also respect the order provided by the user (although there's
+            # a reasonable argument to just sort and dedupe, which fontc does)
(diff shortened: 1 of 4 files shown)


Lib/fontTools/varLib/featureVars.py
```diff
@@ -95,6 +96,10 @@ def addFeatureVariations(font, conditionalSubstitutions, featureTag="rvrn"):
 
     addFeatureVariationsRaw(font, font["GSUB"].table, conditionsAndLookups, featureTags)
 
+    # Recompute usMaxContext in case the font didn't have features before, but
+    # does now.
+    font["OS/2"].usMaxContext = maxCtxFont(font)
```
behdad: Since we only add 1:1 features, maybe only set `usMaxContext` to the max of 1 and it's current value?

Also, I see you say `OS/2` table is required, but we deal with all kinds of non-OpenType-standard fonts all the time, so my preference would be to add a conditional while at it. Thanks.

PR #3895, opened 2025-07-29
[feaLib] Combine duplicate features

cmyr:
The order of lookups within a feature doesn't matter, and so features that have the same set of lookups can reuse the same feature record, regardless of the order in which those lookups are added to the feature.


anthrotype (approved):
LGTM

(personally I'd also be ok to just sort and dedup like fontc does, if the order of feature lookup indices is indeed meaningless, but this at at least would reduce the amount of needless changes)

cmyr:
yea I wanted to just keep this patch minimal.

behdad:
I'm in for sorting as well.

Issue #2680, opened 2022-07-06
deprecate varLib.mutator in favour of varLib.instancer


anthrotype:
The `fonttools varLib.instancer` does all the things that `varLib.mutator` does plus more. The reason both exist is historical, with the former derived from the latter, but being a more general tool that doesn't only cut static instances but also partial ones.
I think this duality only serve to confuse users so I propose we officially deprecate and later on remove `varLib.mutator` module.
Alternatively we could keep both entry points but make mutator simply be an alias to instancer.
We need to double-check whether the latter is in fact a feature-superset of the former.

khaledhosny:
`varLib.mutator` supports CFF2 table, `varLib.instancer` does not.

anthrotype:
ah thanks for reminding me of that, I had completely forgotten. Ok, then it's premature to drop, but not good either that if we keep this duality for too long.

behdad:
Also, deprecate where? Via a warning?

behdad:
varLib.instancer instantiates CFF2 now. https://github.com/fonttools/fonttools/pull/3506

behdad:
We should proceed with this I think.

benkiel:
(a big assumption here based on the `mutator` name, that it does extrapolation — not checked code/docs)

Does `instancer` support extrapolation? If not, that would be an argument for not getting rid of `mutator`. 

Please ignore this if it does support, or if `mutator` doesn't support this.

behdad:
> (a big assumption here based on the `mutator` name, that it does extrapolation — not checked code/docs)
> 
> Does `instancer` support extrapolation? If not, that would be an argument for not getting rid of `mutator`.
> 
> Please ignore this if it does support, or if `mutator` doesn't support this.

It's just a full-instancer. OpenType variation model doesn't support extrapolation in the sense that you mean, so `mutator` is no exception. I just chose the name `mutator` after the old tool that MutatorMath was allegedly named after.

Now, it *could* potentially be changed to extrapolate, for experimentation purposes. It's a much simpler tool than the instancer, which supports partial-instancing.

At any rate, I don't think anyone's suggesting we remove it, at least not for now. But it would be good to warn to let people they should be using the instancer for any real work.

PR #3902, opened 2025-08-13
issue deprecation warning when calling mutator.instantiantiateVariableFont

anthrotype:
Fixes https://github.com/fonttools/fonttools/issues/2680


anthrotype:
one big user-facing difference between the two is that mutator will automatically drop (i.e. pin to default) axes that aren't explicitly provided in the input, whereas the instancer keeps those that aren't explicitly pinned. Should we add a parameter and CLI flag for the instancer that activates the same behavior as mutator? E.g. `fonttools varLib.instancer --full-instance` or `--static`?

behdad:
I agree, but no idea what's a better name. Maybe `--static` if we properly document it.

behdad:
LGTM.

PR #3903, opened 2025-08-13
[cu2qu] Handle degenerate case of 3 equal points

cmyr:
I'm dealing with a situation where open corner erasure is producing a cubic segment where all the points are equal, which is subsequently falling through the conversion to quadratics.

This handles that situation naively, with a branch; actually improving the robustness of the numerical code is an exercise for someone else.

### DIFF
--- a/Lib/fontTools/cu2qu/cu2qu.py
+++ b/Lib/fontTools/cu2qu/cu2qu.py
@@ -385,6 +385,10 @@ def cubic_approx_spline(cubic, n, tolerance, all_quadratic):
     """
 
     if n == 1:
+        if cubic[1] == cubic[2] and (cubic[0] == cubic[1] or cubic[3] == cubic[1]):
+            # a degenerate case: if control points are equal and fall on start or end,
+            # just drop a control point.
+            return [cubic[0], cubic[1], cubic[3]]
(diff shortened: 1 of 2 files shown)


cmyr:
i'm not sure about this patch, it's causing a bunch of other failures I need to investigate...

behdad:
> which is subsequently falling through the conversion to quadratics.

How does it fail?

behdad:
Maybe add a failing test first.

cmyr:
the test case included here fails if you run it against main.

behdad:
I suggest the following fix:
```diff
diff --git a/Lib/fontTools/cu2qu/cu2qu.py b/Lib/fontTools/cu2qu/cu2qu.py
index c0d4cf947..42c8f3983 100644
--- a/Lib/fontTools/cu2qu/cu2qu.py
+++ b/Lib/fontTools/cu2qu/cu2qu.py
@@ -273,6 +273,8 @@ def calc_intersect(a, b, c, d):
     try:
         h = dot(p, a - c) / dot(p, cd)
     except ZeroDivisionError:
+        if a == b == c == d:
+            return a
         return complex(NAN, NAN)
     return c + cd * h
 ```

anthrotype:
closing in favour of #3904

PR #3904, opened 2025-08-14
[cu2qu] handle case when all points in cubic segments are equal

anthrotype:
Fixes https://github.com/fonttools/fonttools/pull/3903 and supersedes it

I applied behdad's suggestion


cmyr:
Okay, note that kurbo also has this quirk, so merging this will actually break a bunch of diffs. I'll open an issue there to discuss.

anthrotype:
> note that kurbo also has this quirk

probably because kurbo's cu2qu code is a port of fonttools

cmyr (approved):
great, thanks!

anthrotype:
@cmyr I'm not going to rush releasing a new fonttools with this given it's relatively minor. But when we do push this, we'll have to concurrently do kurbo as well

PR #3908, opened 2025-08-20
[varLib.instancer] Add --static to pin unspecified axes

behdad:


anthrotype:
this works for the CLI but not for the instantiateVariableFont API. I think we should add a `static=False` parameter to that function instead so it works for both use cases. I'll make a new PR

PR #3909, opened 2025-08-20
[instancer] Implement fully-instantiating avar2 fonts

behdad:


Lib/fontTools/varLib/instancer/__init__.py
```diff
@@ -435,7 +435,27 @@ def normalize(self, varfont, usingAvar=True) -> "NormalizedAxisLimits":
...
+                # TODO: Merge this with the main codepath.
+
+                # Full instancing of avar2 font. Use avar table to normalize location and return.
+                location = self.pinnedLocation()
+                location = {
+                    tag: normalize(value, axes[tag], avarSegments.get(tag, None))
+                    for tag, value in location.items()
+                }
+                return NormalizedAxisLimits(
+                    **avar.renormalizeLocation(location, varfont, False)
```
anthrotype:
```suggestion
                    **avar.renormalizeLocation(location, varfont, dropZeroes=False)
```

Lib/fontTools/ttLib/tables/_a_v_a_r.py
```diff
@@ -143,7 +143,7 @@ def fromXML(self, name, attrs, content, ttFont):
         else:
             super().fromXML(name, attrs, content, ttFont)
 
-    def renormalizeLocation(self, location, font):
+    def renormalizeLocation(self, location, font, dropZeroes=True):
```
anthrotype, line 146: i've always used "zeros" for the plural of "zero" but apparently both are acceptable

PR #3912, opened 2025-08-22
[varLib] Clear inconsistent USE_MY_METRICS flags across masters

anthrotype:
A VF's glyf table is copied from the default master's glyf table. If a composite glyph's component has USE_MY_METRICS set in there (because the composite happens to have the same advance width as one of its components) but the flag is not set consistently among the other masters, that's probably because the composite and component metrics aren't meant to be the same throughout the variation space. Leaving the flag set, means ignoring variations of glyf/gvar phantom points of that composite glyph (as these gets simply copied from the component glyph with USE_MY_METRICS flag on).
The issue with composite glyph metrics doesn't arise normally unless the optional HVAR table (which provides deltas for the glyph metrics irrespective of the USE_MY_METRICS_FLAG) is removed from the VF. The absence of HVAR forces font renderers to use the glyf/gvar phantom points directly for the glyph metrics.

Our main provider or master ttf fonts (ufo2ft) should probably stop setting these flags by default when the intention is to build VF that aren't going to be hinted anyway (USE_MY_METRICS only make sense for hinted fonts).

From varLib's point of view, the safest approach is to check that the flags are set consistently on all the master glyf tables, and un-set them when that is not the case.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -30,7 +30,11 @@ from fontTools.misc.fixedTools import floatToFixed as fl2fi
 from fontTools.misc.textTools import Tag, tostr
 from fontTools.ttLib import TTFont, newTable
 from fontTools.ttLib.tables._f_v_a_r import Axis, NamedInstance
-from fontTools.ttLib.tables._g_l_y_f import GlyphCoordinates, dropImpliedOnCurvePoints
+from fontTools.ttLib.tables._g_l_y_f import (
+    GlyphCoordinates,
+    dropImpliedOnCurvePoints,
+    USE_MY_METRICS,
+)
 from fontTools.ttLib.tables.ttProgram import Program
 from fontTools.ttLib.tables.TupleVariation import TupleVariation
 from fontTools.ttLib.tables import otTables as ot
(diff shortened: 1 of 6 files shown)


anthrotype:
writing tests for varLib is tricky, they are mostly end-to-end. Here I made a test VF with 1 weight axis and 2 masters, Regular and Bold. The masters contain two glyphs, "zero" and "zero.tf" (composite glyph using zero as component). In Regular (default master), "zero.tf" has USE_MY_METRICS flag set because it has the same advance width as "zero". In the Bold, it has not the flag set because the advance widths are different. The generated VF (defaulting at Regular weight) should have the flag un-set.

Lib/fontTools/varLib/__init__.py
```diff
@@ -489,6 +493,77 @@ def _merge_TTHinting(font, masterModel, master_ttfs):
...
+        # and unset the flag if not.
+        expected_num_components = len(glyph.components)
+        if any(
+            _has_inconsistent_use_my_metrics_flag(
+                master_glyf, glyph_name, flagged_components, expected_num_components
+            )
+            for master_glyf in master_glyfs
+        ):
+            comp_names = [name for _, name in flagged_components]
+            log.info(
```
anthrotype, line 556: we could possibly demote this to DEBUG if we deem it too noisy

madig: After having fontmake/ufo2ft do the right thing, I don't expect this to ever trigger outside people manually using `public.truetype.useMyMetrics` :)

anthrotype: yes, either that, or when one explicitly passes the new `--auto-use-my-metrics` flag in fontmake
https://github.com/googlefonts/fontmake/pull/1154

anthrotype: I'll keep logging.INFO especially considering this won't be as frequent going forward

PR #3915, opened 2025-08-25
[subset] Fix shaping behaviour when pruning empty mark sets

khaledhosny:
See https://github.com/harfbuzz/harfbuzz/issues/5499 and the corresponding HarfBuzz PR https://github.com/harfbuzz/harfbuzz/pull/5502


behdad:
Looks good. Thanks.

PR #3918, opened 2025-08-27
Avoid negative hmtx values when creating font from variable

liZe:
This fix is the same as #2827, but in `varLib.instancer`.


anthrotype:
thanks!

PR #3923, opened 2025-09-01
Speed up post.build_psNameMapping

jenskutilek:
See https://github.com/fonttools/fonttools/pull/3809/files#r2314221887

### DIFF
--- a/Lib/fontTools/ttLib/tables/_p_o_s_t.py
+++ b/Lib/fontTools/ttLib/tables/_p_o_s_t.py
@@ -118,6 +118,7 @@ class table__p_o_s_t(DefaultTable.DefaultTable):
     def build_psNameMapping(self, ttFont):
         mapping = {}
         allNames = {}
+        glyphOrderNames = set(self.glyphOrder)
         for i in range(ttFont["maxp"].numGlyphs):
             glyphName = psName = self.glyphOrder[i]
             if glyphName == "":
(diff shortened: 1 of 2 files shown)


Lib/fontTools/ttLib/tables/_p_o_s_t.py
```diff
@@ -126,16 +127,15 @@ def build_psNameMapping(self, ttFont):
             if glyphName in allNames:
                 # make up a new glyphName that's unique
                 n = allNames[glyphName]
-                # check if the exists in any of the seen names or later ones
-                names = set(allNames.keys()) | set(self.glyphOrder)
-                while (glyphName + "." + str(n)) in names:
+                # check if the glyph name exists in the glyph order
+                while f"{glyphName}.{n}" in glyphOrderNames:
                     n += 1
                 allNames[glyphName] = n + 1
```
behdad, line 133: Shouldn't you also insert the new name in `glyphOrderNames`?

jenskutilek: Intuitively I thought so as well, but I failed to come up with an example where it matters. Do you happen to have one?

E.g. in the test I added there are 5 glyphs `[".notdef", "XY", "XY", "XY", "XY.2"]` so the method will disambiguate the second XY to XY.1, then for the third XY find that XY.2 is present in the original glyph order, so the third XY will be named XY.3.

behdad:
Humm. I think you're right.

Issue #3928, opened 2025-09-09
Different number of points in outline in variable font vs if static


cmyr:
I have a funny case that I'm hoping is weird enough that when I explain it an immediate culprit jumps out at somebody?

This is occuring in a glyph in Ojuju.glyphs (I'll include two test files below for the two conditions).

Basically, when I compile this glyph statically (in this case by just trimming the other master) I end up with an outline with `N` points; but if I compile it as a variable font, I end up with `N + 1` points.

(fontc is always producing `N` points, and that is the reason I'm investigating this).

Is it possible that qu2cu is behaving slightly differently when there are multiple outlines? iirc we do consider multiple outlines when we do that conversion, to ensure that the results are compatible. Are there any other obvious candidates?

[Ojuju glyf diff.zip](https://github.com/user-attachments/files/22242728/Ojuju.glyf.diff.zip)

behdad:
cu2qu does convert all masters together indeed, so they are compatible. If it's one master it can result in different output. That's expected.

anthrotype:
> cu2qu does convert all masters together

but so does fontc (via kurbo's cu2qu equivalent, a direct port of the fonttools one). I'll check why fontc and fontmake end up with different outputs

anthrotype:
there must be some rounding issues going on in the --drop-implied-oncurves feature. The extra point that fontmakes adds in  Ojuju is an oncurve that is 48 units apart from previous/next off-curve points...

```
--- /Users/clupo/oss/fontc/build/default/fontc.glyf.ttx	2025-09-10 16:09:37
+++ /Users/clupo/oss/fontc/build/default/fontmake.glyf.ttx	2025-09-10 16:09:37
@@ -3945,6 +3945,7 @@
         <pt x="90" y="101" on="0"/>
         <pt x="90" y="162" on="0"/>
         <pt x="90" y="229" on="0"/>
+        <pt x="90" y="277" on="1"/>
         <pt x="90" y="325" on="0"/>
         <pt x="90" y="392" on="0"/>
         <pt x="90" y="453" on="0"/>
```

i'll keep digging

anthrotype:
ah! it's another one of those differences when cu2qu is run in pure-python vs compiled with cythonized... :(

if I install fonttools in editable mode from a local source checkout _without_ exporting `FONTTOOLS_WITH_CYTHON=1`, then running

```
[...]

PR #3930, opened 2025-09-11
Fix complex division differences in cython vs pure cu2qu

anthrotype:
Reproduces and fixes https://github.com/fonttools/fonttools/issues/3928

### DIFF
--- a/Lib/fontTools/cu2qu/cu2qu.py
+++ b/Lib/fontTools/cu2qu/cu2qu.py
@@ -61,6 +61,23 @@ def dot(v1, v2):
 
 
 @cython.cfunc
+@cython.locals(z=cython.complex, den=cython.double)
+@cython.locals(zr=cython.double, zi=cython.double)
+def _complex_div_by_real(z, den):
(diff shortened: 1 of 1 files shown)


anthrotype:
the reproducer passed LOL 😅 

I think because we only run tests for fonttools with cython on linux, where this doesn't manifest itself. I need to test cython on macos to trigger this

anthrotype:
and it failed (on macos only)!

```python
>       assert result[1][5][1] == 326.49999999999994
E       assert 326.5 == 326.49999999999994
```

https://github.com/fonttools/fonttools/actions/runs/17650235132/job/50159050845?pr=3930#step:5:4006

anthrotype:
now it passes after the fix.

Ojuju.glyphs (and hopefully more) now gets "output is identical" in ttx_diff.py

behdad:
Pure luck, eh?
LGTM. Thanks.

behdad:
> Pure luck, eh?

Oh I see your explanation in the other issue. LGTM

PR #3934, opened 2025-09-16
DecomposePointPen: add reverseFlipped="on_curve_first" to match fontc's flipped component decomposition

anthrotype:
Fixes the discrepancy in starting points when decomposing flipped components between fonttools and fontc (googlefonts/fontc#1633).

When decomposing composite glyphs with flipped components (negative determinant), fonttools and fontc handle the contour starting point differently:
  - fonttools (via ReverseContourPointPen): Reverses the contour but keeps the same starting point (even if it's off-curve)
  - fontc (via kurbo's `BezPath::reverse_subpaths()`: Reverses the contour and rotates it to start with the first on-curve point (BezPath being a collection of curve segments rather than a closed list of points; the rotation occurs well before `reverse_subpaths` while constructing the BezPaths from the UFO/Glyphs sources' outline points).

ufo2ft's uses fonttools' `DecomposingFilterPointPen` to decompose composite glyphs; when `reverseFlipped=True`, this uses the `ReverseContourPointPen` to reverse contour direction and compensate for the flip (negative determinant) in the component transformation.

This PR changes `reverseFlipped` parameter to take an enum (or StrEnum actually) to give users control over the behaviour when reversing contour direction of to-be-decomposed flipped components:

  - `ReverseFlipped.NO` - Don't reverse the contour (same as the current `False` default)
  - `ReverseFlipped.KEEP_START` - Reverse but keep original starting point (same as the current `True`)
  - `ReverseFlipped.ON_CURVE_FIRST` - Rotate to first on-curve point before reversing (new behavior, matching fontc).

The implementation includes:
  1. A new `OnCurveFirstPointPen` filter pen that ensures closed contours start with on-curve points.
  2. Integration in `DecomposingPointPen` (abstract base class of both `DecomposingFilterPointPen` and `DecomposingRecordingPointPen`)
  3.  Backward compatibility with the existing `reverseFlipped: bool` parameter, so exising code is unaffected.

For fontmake/ufo2ft to match fontc's output they will have to be modified to do something like

```python
pen = DecomposingFilterPointPen(outPen, glyphSet, reverseFlipped=ReverseFlipped.ON_CURVE_FIRST, ...)
```

### DIFF
--- a/Lib/fontTools/misc/enumTools.py
+++ b/Lib/fontTools/misc/enumTools.py
@@ -1,0 +1,23 @@
+"""Enum-related utilities, including backports for older Python versions."""
+
+from __future__ import annotations
+
+from enum import Enum
+
+
+__all__ = ["StrEnum"]
+
+# StrEnum is only available in Python 3.11+
+try:
+    from enum import StrEnum
+except ImportError:
+
+    class StrEnum(str, Enum):
+        """
+        Minimal backport of Python 3.11's StrEnum for older versions.
+
+        An Enum where all members are also strings.
+        """
+
+        def __str__(self) -> str:
+            return self.value
--- a/Lib/fontTools/pens/filterPen.py
+++ b/Lib/fontTools/pens/filterPen.py
@@ -1,7 +1,11 @@
 from __future__ import annotations
 
 from fontTools.pens.basePen import AbstractPen, DecomposingPen
-from fontTools.pens.pointPen import AbstractPointPen, DecomposingPointPen
+from fontTools.pens.pointPen import (
+    AbstractPointPen,
+    DecomposingPointPen,
+    ReverseFlipped,
+)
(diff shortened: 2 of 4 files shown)


anthrotype:
as I noted in https://github.com/googlefonts/fontc/issues/1633#issuecomment-3299360590, this is not just to match fontc (which may not in itself be a sufficient argument for the added complexity).

The current behavior of the decomposing point pens when faced with reversing flipped components of keeping the original starting point (which potentially can be an off-curve point) means that subsequent transformations of the same paths that rely on a segment-oriented representation (e.g. converting cubic curves to quadratic splines) end up rotating the points to the subsequent on-curve point _anyway_; fixing the starting point beforehand ensures that the same on-curve starting point is used for both the base glyphs and the (decomposed, reversed) flipped components.

Issue #3935, opened 2025-09-16
[designspaceLib] Proposal: `public.skipExportGlyphs` should be possible to set in UFOs, not only designspace, even when compiling from a designspace


arrowtype:
## Common scenario

`public.skipExportGlyphs` gets set in UFOs, with the intention and expectation that these glyphs won’t be built into output fonts. This is behavior that is useful and common in font editors like RoboFont and GlyphsApp. I use it for two main things:
1. Marking glyphs that are only used as components, such as the /caron.alt, but which don’t have Unicodes to make them accesible. I want these components to decompose on build, and for the parent glyph to be removed.
2. Marking alternate glyphs that were made as "draft" or experimental designs, but not intended for inclusion in output fonts. But, I want to leave them in the source, in case I return to that idea later.

When building directly from a UFO, `public.skipExportGlyphs` works exactly as expected – those glyphs don’t appear in the output font.

## Problem:

As stated, the output from a single UFO is as expected. However, when building from a designspace – even if this lib item is set in all UFO sources – this lib key is completely disregarded. It _only_ works if the lib key is set within the designspace itself. I would argue this is highly unintuitive.

Yet, this counter-intuitive requirement is very explicitly clearly [specified in the DS spec](https://github.com/googlefonts/ufo2ft/issues/854):

> If the lib key is empty or not present in the Designspace, all glyphs should be exported, regardless of what the same lib key in any of the UFOs says.

This requirement was put in place for good reasons – namely, trying to avoid complexity and out-of-sync data, but I’m hoping that this is something we might consider again.

## Proposal

- If a designspace does have its own lib item `public.skipExportGlyphs`, this is used as the single record to be used.
- If a designspace does _not_ have its own lib item `public.skipExportGlyphs`, it would refer to the origin/default source UFO. For simplicity’s sake, this should be the case whether the fonts being exported are variable or static interpolations.

I’ll make a PR for this, next.

## Reasoning

If I had to give a reason of the logic of _why_ it is unintuitive, I would argue that (almost) nothing else about glyph-level data is set up in a designspace. We don’t set up glyph names, or outlines, or spacing, or kerning, or kern groups, or anchors, etc etc, in the designspace. That data is all set up along with the glyphs themselves, within the source documents.

This is discussed further in a ufo2ft issue: https://github.com/googlefonts/ufo2ft/issues/854, with several good points made. A couple of highlights:
[...]

PR #3936, opened 2025-09-16
[designspaceLib] Adjust spec for `public.skipExportGlyphs`

arrowtype:
This proposal is described further at https://github.com/fonttools/fonttools/issues/3935, as well as the earlier issue thread at https://github.com/googlefonts/ufo2ft/issues/854.

In this PR, I am slightly more confident in the [first commit](https://github.com/fonttools/fonttools/commit/11c0a38b1c7d447103cd4d4434c3abb1608b3cc3) and slightly less confident about the [second commit](https://github.com/fonttools/fonttools/commit/481ce3d6ae9fd6a1bb91cba7baa9fa8c6fc35631), so I left them deliberately separate. However, I think this arrangement probably makes sense, assuming others agree that an empty lib key should be treated as significant data.

I am open to critique of language or formatting here, and direct edits from maintainers. I’m also happy to answer questions, if there are any that I can speak to.

Thank you for your consideration here!

### DIFF
--- a/Doc/source/designspaceLib/index.rst
+++ b/Doc/source/designspaceLib/index.rst
@@ -116,7 +116,9 @@ public.skipExportGlyphs
 This lib key works the same as the UFO lib key with the same name. The
 difference is that applications using a Designspace as the corner stone of the
 font compilation process should use the lib key in that Designspace instead of
-any of the UFOs, if the lib key is present in the Designspace.
+any of the UFOs, if the lib key is present in the Designspace. This should be
+true even if the lib key is present but empty in the Designspace, so that it is
+possible to use the Designspace lib key as an override to UFO-level lib keys.
 
 If the lib key is not present in the Designspace but it is present in the origin
 source of the Designspace, the font compiler should use the lib key from that
--- a/Doc/source/designspaceLib/index.rst
+++ b/Doc/source/designspaceLib/index.rst
@@ -116,9 +116,13 @@ public.skipExportGlyphs
 This lib key works the same as the UFO lib key with the same name. The
 difference is that applications using a Designspace as the corner stone of the
 font compilation process should use the lib key in that Designspace instead of
-any of the UFOs. If the lib key is empty or not present in the Designspace, all
-glyphs should be exported, regardless of what the same lib key in any of the
-UFOs says.
+any of the UFOs, if the lib key is present in the Designspace.
+
+If the lib key is not present in the Designspace but it is present in the origin
+source of the Designspace, the font compiler should use the lib key from that
+origin UFO, and it should apply that set of skipped glyphs to any fonts 
+compiled from the Designspace.
+
 
 public.fontInfo
 -----------------------


lianghai:
Will it be simpler to actually specify that the inheritance logic should be the same for `public.skipExportGlyphs` and `public.fontInfo`?

> However, I think this arrangement probably makes sense, assuming others agree that an empty lib key should be treated as significant data.

Yes, for an optional key in a dictionary-like data structure, the absence of this key should be analyzed as having a different meaning from the value being falsy (`[]` in this case).

Doc/source/designspaceLib/index.rst
```diff
@@ -116,9 +116,15 @@ public.skipExportGlyphs
 This lib key works the same as the UFO lib key with the same name. The
 difference is that applications using a Designspace as the corner stone of the
 font compilation process should use the lib key in that Designspace instead of
-any of the UFOs. If the lib key is empty or not present in the Designspace, all
-glyphs should be exported, regardless of what the same lib key in any of the
-UFOs says.
+any of the UFOs, if the lib key is present in the Designspace. This should be
+true even if the lib key is present but empty in the Designspace, so that it is
+possible to use the Designspace lib key as an override to UFO-level lib keys.
```
anthrotype:
```suggestion
any of the UFOs. If the lib key is present but empty in the Designspace, all
glyphs should be exported, regardless of what the same lib key in any of the
UFOs says.
```

Doc/source/designspaceLib/index.rst
```diff
@@ -116,9 +116,15 @@ public.skipExportGlyphs
...
+
+If the lib key is not present in the Designspace but it is present in the origin
```
anthrotype:
```suggestion
If the lib key is not present in the Designspace but it is present in the default
```

Doc/source/designspaceLib/index.rst
```diff
@@ -116,9 +116,15 @@ public.skipExportGlyphs
...
+source of the Designspace, the font compiler should use the lib key from that
+origin UFO, and it should apply that set of skipped glyphs to any fonts
```
anthrotype:
```suggestion
UFO, and it should apply that set of skipped glyphs to any fonts 
```

anthrotype:
> Will it be simpler to actually specify that the inheritance logic should be the same for public.skipExportGlyphs and public.fontInfo?

that's actually a good idea. E.g. one may want to define different skipExportGlyphs for different VF or instance targets.

Doc/source/designspaceLib/index.rst
arrowtype: Seems more straightforward. Thanks!

arrowtype:
> Will it be simpler to actually specify that the inheritance logic should be the same for public.skipExportGlyphs and public.fontInfo?

Is there specific language you might suggest adding/adjusting in this PR?

I’m not super familiar with the inheritance logic of public.fontInfo, offhand. I suppose you mean that it treats some fields as applying to everything, but lets other fields apply on to individual instances? Is there a record of what goes where? There must be...

>  E.g. one may want to define different skipExportGlyphs for different VF or instance targets.

YES, this would be handy. It’s pretty frequent that static exports end up with unreachable glyphs that are there simply because they are required for the variable font’s font variations glyph substitution, such as an alt /Q or barless /dollar.

Would this automatically be covered by specifying that “the inheritance logic should be the same for public.skipExportGlyphs and public.fontInfo”?

lianghai:
> that's actually a good idea. E.g. one may want to define different skipExportGlyphs for different VF or instance targets.

To be clear, my intention is more about reducing special cases in the spec rather than expanding the functionality.

> Is there specific language you might suggest adding/adjusting in this PR?

I’m still pondering over this idea. I’m not familiar with this spec either (I didn’t even know about its existence before, as I thought everything was in [the source code](https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/designspaceLib/__init__.py)…), but will try drafting something later.

> I suppose you mean that it treats some fields as applying to everything, but lets other fields apply on to individual instances?

Oh I meant the inheritance chain specified for [public.fontInfo](https://fonttools.readthedocs.io/en/stable/designspaceLib/index.html#public-fontinfo), for example:

```py

InstanceDescriptor.lib["public.fontInfo"]["familyName"]  # Only applicable to `InstanceDescriptor` and `VariableFontDescriptor`
InstanceDescriptor.familyName  # Only applicable to `InstanceDescriptor`
DesignSpaceDocument.lib["public.fontInfo"]["familyName"]
DesignSpaceDocument.default.font.info.familyName
```

Conceptually I guess public.skipExportGlyphs should actually be treated as one of the font info keys instead of the whole public.fontInfo, as each key of the latter has its value inherited separately.

arrowtype:
> Oh I meant the inheritance chain specified for [public.fontInfo](https://fonttools.readthedocs.io/en/stable/designspaceLib/index.html#public-fontinfo), for example:

I wasn’t familiar with / hadn’t remembered this explicit hierarchy, but that is so helpful to see again. Thanks!

I wonder if we might specify something like...

> With a hierarchy similar to [public.fontInfo](https://fonttools.readthedocs.io/en/stable/designspaceLib/index.html#public-fontinfo), `skipExportGlyphs` should be inherited using the following order, in order of descending priority::
> 
> 1. The `public.skipExportGlyphs` key in the `<lib>` element of the `<variable-font>` or `<instance>` elements.
> 
> 2. The `public.skipExportGlyphs` key found in the `<lib>` element of the designspace document’s root.
> 
> 3 The `lib.plist` in the UFO source at the origin of the interpolation space.

Would item 1 be a reasonable way to specify functionality to “define different skipExportGlyphs for different VF or instance targets”? Is this something we can reasonably add to the spec?

anthrotype:
@arrowtype yes, sounds good to me

arrowtype:
Okay, added! The previous information still seemed to provide useful detail, but felt a bit hard to easily read, especially when placed before the hierarchy. So, I moved the previous information to bullets that follow.

It’s possible that this now provides more information than needed, but I do think it helps clarify potential areas of uncertainty.

@anthrotype what are your thoughts on this?

arrowtype:
(I’ve just rebased this branch to update it.)

Doc/source/designspaceLib/index.rst
```diff
@@ -113,12 +113,28 @@ Common Lib Key Registry
...
 
-This lib key works the same as the UFO lib key with the same name. The
-difference is that applications using a Designspace as the corner stone of the
-font compilation process should use the lib key in that Designspace instead of
-any of the UFOs. If the lib key is empty or not present in the Designspace, all
-glyphs should be exported, regardless of what the same lib key in any of the
-UFOs says.
+This lib key works the same as the UFO lib key with the same name.
+
+With a hierarchy similar to
```
lianghai, line 118: “Hierarchy” is a new concept, and it’s unclear what it refers to. Better avoid it.

```suggestion
Similar to
```

Doc/source/designspaceLib/index.rst
```diff
@@ -113,12 +113,28 @@ Common Lib Key Registry
...
-difference is that applications using a Designspace as the corner stone of the
-font compilation process should use the lib key in that Designspace instead of
-any of the UFOs. If the lib key is empty or not present in the Designspace, all
-glyphs should be exported, regardless of what the same lib key in any of the
-UFOs says.
+This lib key works the same as the UFO lib key with the same name.
+
+With a hierarchy similar to 
+`public.fontInfo <https://fonttools.readthedocs.io/en/stable/designspaceLib/index.html#public-fontinfo>`, 
+``public.skipExportGlyphs`` should be inherited using the following order, in
```
lianghai, line 120: Align with the language about `public.fontInfo`:

```suggestion
``public.skipExportGlyphs`` must be inherited using the following order, in 
```

arrowtype:
@lianghai Thanks so much for your edits!

> consider adding cross-references ... from all the affected .lib’s

Seems like a good idea. I’ll have to come back to this aspect, hopefully over the weekend.

anthrotype:
@arrowtype would you like to also work on updating ufo2ft to match this? 😛

arrowtype:
@lianghai I’ve implemented you specific example, but I’m not familiar enough with this library’s structure to know where _all_ affected libs would be, and why it might make sense to link from `InstanceDescriptor.familyName` and `.styleName` docstrings. (I guess for those, it’s because `public.fontInfo` can override them?) How might I best determine a full list of docstrings which should have the cross reference?

@anthrotype I would like to help further if I can, but I’d have to ask how this might fit into the priorities at The Type Founders. Also, I probably wouldn’t be particularly efficient at making such changes. Of course, I also understand if this isn’t a high priority on anyone’s docket, and I very much appreciate the expertise being lent in this PR, even just as a way I’m learning more about the designspaceLib.

lianghai:
> @lianghai I’ve implemented you specific example, but I’m not familiar enough with this library’s structure to know where _all_ affected libs would be, …

> How might I best determine a full list of docstrings which should have the cross reference?

Uh I thought the spec itself you’re editing now is supposed to explain which exactly attributes across the DesignSpaceDoc hierarchy participate in this inheritance logic? Then I expect all participating Python attributes need to also point back to the spec, otherwise whoever using the Python API doesn’t know how to interpret the attributes properly. I, for one, didn’t realize there was explicitly documented inheritance logic that has no reference in designspaceLib/__init__.py‎.

But I’m quite confused by how https://fonttools.readthedocs.io/en/latest/designspaceLib/ is organized, with the Python API and XML API documented separately… @anthrotype, who’s supposed to be the owner of this chapter of documentation https://fonttools.readthedocs.io/en/latest/designspaceLib/? We need the owner to make a decision about how to document such info properly.

> …, and why it might make sense to link from `InstanceDescriptor.familyName` and `.styleName` docstrings. (I guess for those, it’s because `public.fontInfo` can override them?)

If docstrings of `.lib` attributes should be updated to cross-reference to the spec, `InstanceDescriptor.familyName`, etc, should too, because otherwise it’s unclear how it affects those font info values. The key is not the concept of a lib, but how font info values should be interpreted given various values in the libs and other attributes.

Now I’m thinking such an inheritance logic should actually captured in a function…

Btw, if you add me as a collaborator, I can try drafting something.

arrowtype:
@lianghai Thanks, sounds good! Done: https://github.com/arrowtype/fonttools/invitations

PR #3938, opened 2025-09-19
Fix interpolation of variable scalars, and allow on pre-compiled fonts 

Hoolean:
(aims to facilitate https://github.com/googlefonts/ufo2ft/issues/945)

Previously, calculating an individual interpolated value did not account
 for avar mapping. This commit fixes this, as well as allowing for
 interpolation and other common operations to be performed on fonts prior
 to compilation by decoupling from the exact ttLib representations of
 axes.

Some refactoring has been undergone to assist with this, including
 adding some more types and docs, and splitting the builder into a
 separate class to limit the amount of state in the VariableScalar class
 itself.

The fonttools and ufo2ft test suites pass with these changes, although
 there is no coverage for the designspace-based interpolation yet. This
 should be introduced before merging.

## TODOs

- [ ] Improve test coverage

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -8,7 +8,7 @@ from fontTools.feaLib.lookupDebugInfo import (
 )
 from fontTools.feaLib.parser import Parser
 from fontTools.feaLib.ast import FeatureFile
-from fontTools.feaLib.variableScalar import VariableScalar
+from fontTools.feaLib.variableScalar import VariableScalar, VariableScalarBuilder
 from fontTools.otlLib import builder as otl
 from fontTools.otlLib.maxContextCalc import maxCtxFont
 from fontTools.ttLib import newTable, getTableModule
@@ -124,6 +124,7 @@ class Builder(object):
             self.varstorebuilder = OnlineVarStoreBuilder(
                 [ax.axisTag for ax in self.axes]
             )
+            self.scalar_builder = VariableScalarBuilder.from_ttf(font)
         self.default_language_systems_ = set()
         self.script_ = None
         self.lookupflag_ = 0
@@ -180,10 +181,6 @@ class Builder(object):
         self.stat_ = {}
         # for conditionsets
         self.conditionsets_ = {}
-        # We will often use exactly the same locations (i.e. the font's masters)
-        # for a large number of variable scalars. Instead of creating a model
-        # for each, let's share the models.
-        self.model_cache = {}
 
     def build(self, tables=None, debug=False):
         if self.parseTree is None:
@@ -805,7 +802,6 @@ class Builder(object):
                 gdef.remap_device_varidxes(varidx_map)
                 if "GPOS" in self.font:
                     self.font["GPOS"].table.remap_device_varidxes(varidx_map)
-            self.model_cache.clear()
         if any(
             (
                 gdef.GlyphClassDef,
@@ -1719,19 +1715,24 @@ class Builder(object):
 
         self.conditionsets_[key] = value
 
-    def makeVariablePos(self, location, varscalar):
-        if not self.varstorebuilder:
+    def makeVariablePos(
+        self, location, varscalar: VariableScalar
+    ) -> tuple[int, int | None]:
+        """Make a pos statement from a VariableScalar, returning the default
+        value, and optionally the variation index if the scalar genuinely
+        requires variation too."""
+
+        if self.varstorebuilder is None or self.scalar_builder is None:
             raise FeatureLibError(
                 "Can't define a variable scalar in a non-variable font", location
             )
 
-        varscalar.axes = self.axes
         if not varscalar.does_vary:
-            return varscalar.default, None
+            return self.scalar_builder.default_value(varscalar), None
 
         try:
-            default, index = varscalar.add_to_variation_store(
-                self.varstorebuilder, self.model_cache, self.font.get("avar")
+            default, index = self.scalar_builder.add_to_variation_store(
+                varscalar, self.varstorebuilder
             )
         except VarLibError as e:
             raise FeatureLibError(
--- a/Lib/fontTools/feaLib/variableScalar.py
+++ b/Lib/fontTools/feaLib/variableScalar.py
@@ -1,18 +1,31 @@
+from collections.abc import Mapping
+from dataclasses import dataclass
+from typing import Self
+
+from fontTools.designspaceLib import DesignSpaceDocument
+from fontTools.ttLib.ttFont import TTFont
 from fontTools.varLib.models import VariationModel, normalizeValue, piecewiseLinearMap
 
+LocationTuple = tuple[tuple[str, float], ...]
+"""A hashable location."""
 
-def Location(loc):
-    return tuple(sorted(loc.items()))
+
+def Location(location: Mapping[str, float]) -> LocationTuple:
+    """Create a hashable location from a dictionary-like location."""
+    return tuple(sorted(location.items()))
 
 
 class VariableScalar:
     """A scalar with different values at different points in the designspace."""
 
+    values: dict[LocationTuple, int]
+    """The values across various user-locations. Must always include the default
+    location by time of building."""
+
     def __init__(self, location_value={}):
-        self.values = {}
-        self.axes = {}
-        for location, value in location_value.items():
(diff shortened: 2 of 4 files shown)


Hoolean:
I've added coverage for pre-compiled interpolation and interpolating a single value now 🚀 

@anthrotype Is there anything we want to do to confirm that the full stack is still building `GPOS` correctly aside from running the ufo2ft test suite? Otherwise, ready for review.

Lib/fontTools/feaLib/builder.py
```diff
@@ -805,7 +804,6 @@ def buildGDEF(self):
                 gdef.remap_device_varidxes(varidx_map)
                 if "GPOS" in self.font:
                     self.font["GPOS"].table.remap_device_varidxes(varidx_map)
-            self.model_cache.clear()
```
Hoolean, line 808: Is this important to retain? Would be easy to add a helper function on the builder class that does the same if we expect feaLib builder re-use

Lib/fontTools/feaLib/variableScalar.py
```diff
@@ -27,92 +46,186 @@ def __repr__(self):
...
 
-    def add_value(self, location, value):
-        if self.axes:
-            location = self.fix_location(location)
+        return Location(full)
 
-        self.values[Location(location)] = value
+    def normalised_values(self, scalar: VariableScalar) -> dict[LocationTuple, int]:
+        """Get the values of a variable scalar with fully-specified, normalized
+        and validated user-locations."""
```
anthrotype, line 139: these are no longe "user" locations, are they? The `normalise_location` method would turn them into fully normalized and avar-mapped locations 

```suggestion
        and validated locations."""
```

Hoolean: `avar`-mapping happens as a step after `normalise_location` so that we we don't have to calculate it before the lookup in `model_cache`; maybe the complexity of the half-way approach isn't worth it though

Hoolean: I think this convinces me that `model_cache` should use one of the following as a key instead:

- Pre-normalized but unmapped user locations
    - Cheap to compute cache key, but more misses
    - This is what was used before
- Fully-normalized and mapped user locations
    - More expensive to compute cache key, but fewer misses

I suspect the cache was added to optimise for the case where there is a large machine-generated feature code file (e.g. variable kerning), and so I lean towards the former

anthrotype: oh, so here the "normalized user" coordinates are the user coordinates normalized using the fvar's user-space axis triplets only, _before_ doing the avar normalization. Usually when I see "normalized" I think about the end result, i.e. the internal normalized coordinates _after_ the avar mapping. The user coordinates pre-normalized are just an intermediate step to avar-normalized coordinates, so it's strange to seem them used here like this..

Lib/fontTools/feaLib/builder.py
anthrotype: i don't think the builder is meant to be reused, so no

Lib/fontTools/feaLib/variableScalar.py
```diff
@@ -27,92 +46,186 @@ def __repr__(self):
...
+            for location in full_values.keys()
+        ]
+        model = self.model_cache[cache_key] = VariationModel(design_locations)
+
+        return model
+
+    def get_deltas_and_supports(self, scalar: VariableScalar):
+        """Calculate deltas and supports from this scalar's variation model."""
+        values = list(scalar.values.values())
+        return self.model(scalar).getDeltasAndSupports(values)
```
anthrotype, line 217: it might be a contrived example but `scalar.values` could in theory contain duplicate locations e.g. if repeated twice with implicit/explicit defaults:

```fea
feature kern {
    pos two <0 (wght=200:10 wght=200,wdth=100:20) 0 0>;
} kern;
```

```python
scalar = VariableScalar()
scalar.add_value({"wght": 200.0}, 10)  # omits wdth (defaults to 100)
scalar.add_value({"wght": 200.0, "width": 100.0}, 20)  # explicitly sets wdth to default
```

That in turn would make VariationModel.getDeltas() trigger an assertion at `assert len(masterValues) == len(self.deltaWeights)`.

Did you use the raw `scalar.values` intentionally? Or maybe you meant to do `self.normalised_values(scalar)` which would deduplicate keys (effectively the last wins):

```suggestion
        values = list(self.normalised_values(scalar).values())
        return self.model(scalar).getDeltasAndSupports(values)
```

In the `main` branch, a situation like the one above would trigger FeatureLibError("Failed to compute deltas for variable scalar") because constructing the VariationModel throws "Location must be unique". Maybe that's what we actually want here, instead of silently overwriting.
In your code, constructing the VariationModel doesn't fail like that with duplicate locations, because you first call `full_values = self.normalised_values(scalar)` which deduplicates them, but then we get the assertion error in `get_deltas_and_supports`.

anthrotype: this passes on `main` but throws `AssertionError: (2, 1)` in your branch:

```diff
diff --git a/Tests/feaLib/builder_test.py b/Tests/feaLib/builder_test.py
index 05f645af3..5cc744b56 100644
--- a/Tests/feaLib/builder_test.py
+++ b/Tests/feaLib/builder_test.py
@@ -1308,6 +1308,22 @@ class BuilderTest(unittest.TestCase):
         assert self.get_region(var_region_axis_wght) == (0.0, 0.90625, 1.0)
         assert self.get_region(var_region_axis_wdth) == (0.0, 0.5, 1.0)
 
+    def test_variable_scalar_duplicate_location(self):
+        features = """
+            languagesystem DFLT dflt;
+
+            feature kern {
+                pos two <0 (wght=200:10 wght=200,wdth=100:20) 0 0>;
+            } kern;
+        """
+
+        font = self.make_mock_vf()
+
+        with self.assertRaisesRegex(
+            FeatureLibError, "Failed to compute deltas for variable scalar"
+        ):
+            addOpenTypeFeaturesFromString(font, features)
+
     def test_ligatureCaretByPos_variable_scalar(self):
         """Test that the `avar` table is consulted when normalizing user-space
         values."""

```

feel free to reuse the test if you like. I think it should still fail and with a more meaninful message like the one we had on `main`, rather than AssertionError.

Lib/fontTools/feaLib/variableScalar.py
anthrotype: I think before the model_cache was keyed on pre-normalized _user_ locations, the same that VariableScalar values are, not "pre-normalized _design_ locations" as you said above. Isn't VariableScalar specified in user-space coordinates? If not, why would avar be needed?

anthrotype:
> Previously, calculating an individual interpolated value did not account for avar mapping. This commit fixes this

not exactly clear to me where the bug was. would be nice to see a test that would fail on `main` and which would be fixed by this PR

Lib/fontTools/feaLib/variableScalar.py
Hoolean: Sorry, that was my typo above, I meant 'user' everywhere I said 'design' 🔀

Lib/fontTools/feaLib/variableScalar.py
```diff
@@ -27,92 +46,186 @@ def __repr__(self):
...
-            # I *guess* we could interpolate one, but I don't know how.
-        return self.values[key]
 
-    def value_at_location(self, location, model_cache=None, avar=None):
-        loc = Location(location)
-        if loc in self.values.keys():
-            return self.values[loc]
-        values = list(self.values.values())
-        loc = dict(self._normalized_location(loc))
-        return self.model(model_cache, avar).interpolateFromMasters(loc, values)
```
anthrotype, line 88: I suspect the bug was here, where it was passing non-avar-mapped normalized location to interpolateFromMasters?

Hoolean: Yep, that was it exactly

Hoolean: The newly added test should fail on `main` in kind

Hoolean: (Although it uses the new designspace instantiation and builder, so will not run without some adapting to the old API)

anthrotype:
Superseded by #4054

PR #3939, opened 2025-09-22
Use trusted publishing

madig:
In light of the recent npm supply chain attacks and also https://blog.pypi.org/posts/2025-09-16-github-actions-token-exfiltration/, I'm combing through our font stack to see if all them Py projects are using the trusted publisher mechanism as recommended by PyPI. See https://docs.pypi.org/trusted-publishers/ and https://docs.astral.sh/uv/guides/integration/github/#publishing-to-pypi.

Someone needs to do two things for this PR to work:

* Create an environment called "publish-to-pypi" in this GitHub repository under Settings -> Environments.
* Follow https://docs.pypi.org/trusted-publishers/adding-a-publisher/ to set up the other side on PyPI.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -21,9 +21,9 @@ jobs:
   build_pure:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
     - name: Install uv
-      uses: astral-sh/setup-uv@v5
+      uses: astral-sh/setup-uv@v6
     - name: Build source distribution and pure-python wheel
       run: |
         uv build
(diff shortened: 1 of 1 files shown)


.github/workflows/wheels.yml
```diff
@@ -83,13 +83,24 @@ jobs:
...
+      id-token: write # IMPORTANT: mandatory for trusted publishing
+
     steps:
+    - uses: actions/checkout@v5
     - uses: actions/download-artifact@v4
       with:
         # so that all artifacts are downloaded in the same directory specified by 'path'
         merge-multiple: true
         path: dist
-    - uses: pypa/gh-action-pypi-publish@v1.13.0
```
anthrotype, line 92: but the official pypa/gh-action-pypi-publish should support trusted publishing, no? I don't think we _need_ to use uv publish

anthrotype:
> Create an environment called "publish-to-pypi" in this GitHub repository under Settings -> Environments

created, but didn't do anything more than that

anthrotype:
the pypi docs say

> In addition, you _may optionally_ provide the name of a [GitHub Actions environment](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment).

i'm not sure what that adds, but I did it anyway

anthrotype:
@madig please double check I did it correctly:

image

madig:
I believe that's correct! I can undo the uv thing to keep it minimal.

PR #3944, opened 2025-09-25
Add docs for the visitor module

Hoolean:
I've been using the `visitor` module and finding it excellent, and so thought it deserved some love ❤️ This PR adds an initial draft of documentation, generates the automated docs, and includes a bit of example code too.

### DIFF
--- a/Lib/fontTools/misc/visitor.py
+++ b/Lib/fontTools/misc/visitor.py
@@ -79,12 +79,12 @@ class Visitor(object):
     def visitObject(self, obj, *args, **kwargs):
         """Called to visit an object. This function loops over all non-private
         attributes of the objects and calls any user-registered (via
-        @register_attr() or @register_attrs()) visit() functions.
+        ``@register_attr()`` or ``@register_attrs()``) ``visit()`` functions.
 
         If there is no user-registered visit function, of if there is and it
-        returns True, or it returns None (or doesn't return anything) and
-        visitor.defaultStop is False (default), then the visitor will proceed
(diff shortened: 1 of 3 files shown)


Lib/fontTools/misc/visitor.py
```diff
@@ -79,12 +79,12 @@ def _visitorsFor(celf, thing, _default={}):
     def visitObject(self, obj, *args, **kwargs):
         """Called to visit an object. This function loops over all non-private
         attributes of the objects and calls any user-registered (via
-        @register_attr() or @register_attrs()) visit() functions.
+        ``@register_attr()`` or ``@register_attrs()``) ``visit()`` functions.
```
behdad, line 83: There's a typo "of" should be "or".

behdad (approved):

behdad:
Thanks!

behdad:
One comment I forgot in my review. Should the intro docs mention TTVisitor?

Lib/fontTools/misc/visitor.py
Hoolean: Oops, I must admit I didn't proof-read the existing docstrings... I've corrected this now and edited them to use RST bulleted lists, so they read a little nicer on the net

Hoolean:
> Should the intro docs mention TTVisitor?

Yes! I didn't realise this existed, so definitely worthy of documentation 😁

I've added it now, along with its equivalent auto-generated reference; I've also added a few commits to address the above review feedback too.

PR #3949, opened 2025-09-29
Pass BASE table through subsetter by default

RickyDaMa:
I don't have push access to the repo, but I've rebased #1866 without conflicts so hopefully we can get this merged 🙏

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -272,7 +272,7 @@ Font table options
   Specify (=), add to (+=) or exclude from (-=) the comma-separated
   set of tables that will be be dropped.
   By default, the following tables are dropped:
-  'BASE', 'JSTF', 'DSIG', 'EBDT', 'EBLC', 'EBSC', 'PCLT', 'LTSH'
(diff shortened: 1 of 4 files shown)


anthrotype:
LGTM but black is angry, can you fix? i'll merge it afterwards

anthrotype:
tests also fail with `ValueError: too many values to unpack (expected 2)`

Lib/fontTools/subset/__init__.py
```diff
@@ -3323,6 +3322,7 @@ class UnknownOptionError(OptionError):
     _drop_tables_default += ["Feat", "Glat", "Gloc", "Silf", "Sill"]  # Graphite
     _no_subset_tables_default = [
         "avar",
+        "BASE",
```
RickyDaMa, line 3325: General thought: we may want to implement something better than this in the future, because for example if a user is subsetting out scripts, it could be useful to remove no longer relevant BASE min/max records etc.

For now though I guess it's fine to keep the whole table around, since it's not going to do any harm

behdad:
LGTM.

PR #3952, opened 2025-10-08
Break off typing part of #3826

madig:
This tries to port only the type hints (plus some import sorting and asserts to please the type checkers) from https://github.com/fonttools/fonttools/pull/3826 and nothing else.

The type checker will complain about various function bodies, but that can be tackled in another PR.

---

@jorenham In case you're still interested :)

Closes https://github.com/fonttools/fonttools/pull/3954.

### DIFF
--- a/Lib/fontTools/ttLib/ttFont.py
+++ b/Lib/fontTools/ttLib/ttFont.py
@@ -371,7 +371,10 @@ class TTFont(object):
         ):
             if reorderTables is False:
                 # sort tables using the original font's order
-                assert self.reader is not None
+                if self.reader is None:
+                    raise TTLibError(
+                        "The original table order is unavailable because there isn't a font to read it from."
+                    )
                 tableOrder = list(self.reader.keys())
             else:
                 # use the recommended order from the OpenType specification
@@ -489,7 +492,10 @@ class TTFont(object):
         if not splitTables:
             writer.newline()
         else:
-            assert writer.filename is not None
+            if writer.filename is None:
+                raise TTLibError(
+                    "splitTables requires the file name to be a file system path, not a stream."
+                )
             path, ext = os.path.splitext(writer.filename)
 
         for tag in tables:


Lib/fontTools/ttLib/ttFont.py
```diff
@@ -1058,31 +1259,45 @@ def xmlToTag(tag):
...
+    Tag("maxp"),
+    Tag("OS/2"),
+    Tag("name"),
+    Tag("cmap"),
+    Tag("post"),
+    Tag("CFF "),
+]
 
+DSIGTableTag = Tag("DSIG")
+GlyphOrderTag = Tag("GlyphOrder")
```
jorenham: can we do without these runtime changes?

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -1126,8 +1346,10 @@ def reorderFontTables(inFile, outFile, tableOrder=None, checkChecksums=False):
     writer.close()
 
 
-def maxPowerOfTwo(x):
-    """Return the highest exponent of two, so that
+def maxPowerOfTwo(x: int) -> int:
+    """Return floor(log2(x)).
+
+    Return the highest exponent of two, so that
```
jorenham: non-typing change

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -1007,10 +1208,10 @@ def tagToIdentifier(tag):
     return ident
 
 
-def identifierToTag(ident):
+def identifierToTag(ident: str) -> Tag:
     """the opposite of tagToIdentifier()"""
     if ident == "GlyphOrder":
-        return ident
+        return Tag(ident)
```
jorenham: maybe we could introduce a `TagLike: TypeAlias = Tag | str` and use it here to avoid having to touch the implementation

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -444,14 +618,22 @@ def ensureDecompiled(self, recurse=None):
...
-    def __getitem__(self, tag):
+    @overload
+    def __getitem__(self, tag: Literal["cmap"]) -> "_c_m_a_p.table__c_m_a_p": ...
+
+    # TODO: ...
+
+    @overload
+    def __getitem__(self, tag: str) -> DefaultTable | GlyphOrder: ...
+
+    def __getitem__(self, tag: str) -> DefaultTable | GlyphOrder:
```
jorenham: Nit: It can help with readability to have no blank newlines in between overloads

```suggestion
   # TODO: more overloads
    @overload
    def __getitem__(self, tag: Literal["cmap"]) -> "_c_m_a_p.table__c_m_a_p": ...
    @overload
    def __getitem__(self, tag: str) -> DefaultTable | GlyphOrder: ...
    def __getitem__(self, tag: str) -> DefaultTable | GlyphOrder:
```

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -421,20 +595,20 @@ def has_key(self, tag):
 
     __contains__ = has_key
 
-    def keys(self):
+    def keys(self) -> list[Tag]:
```
jorenham: perhaps we can avoid we runtime changes below by changing this to `-> Sequence[Tag | str]`?

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -9,16 +11,128 @@
...
+        _v_h_e_a,
+        _v_m_t_x,
+    )
+
+    from fontTools.ttLib.tables.DefaultTable import DefaultTable
 
 log = logging.getLogger(__name__)
 
 
+Number = TypeVar("Number", bound=float)
```
jorenham: The `_*` helps prevent mistaking it as public API, and the `*T` suffix is convention for type parameters

```suggestion
_NumberT = TypeVar("_NumberT", bound=float)
```

Lib/fontTools/ttLib/ttFont.py
dscorbett: `Tag` is a subclass of `str`, so the type `Tag | str` is equivalent to `str`.

jorenham: Ahh I had no idea! Well that simplifies things then :)

Lib/fontTools/ttLib/ttFont.py
jorenham: Ah nevermind; feel free to hit "resolve" on this one (cus I can't)

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -972,7 +1173,7 @@ def _escapechar(c):
         return hex(byteord(c))[2:]
 
 
-def tagToIdentifier(tag):
+def tagToIdentifier(tag: str | Tag) -> str:
```
jorenham:
```suggestion
def tagToIdentifier(tag: str) -> str:
```

ref: https://github.com/fonttools/fonttools/pull/3952/files#r2414777868

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -1028,7 +1229,7 @@ def identifierToTag(ident):
     return Tag(tag)
 
 
-def tagToXML(tag):
+def tagToXML(tag: str | Tag) -> str:
```
jorenham:
```suggestion
def tagToXML(tag: str) -> str:
```

ref: https://github.com/fonttools/fonttools/pull/3952/files#r2414777868

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -946,21 +1146,22 @@ def getTableClass(tag):
...
+def getClassTag(klass: type[DefaultTable]) -> Tag:
     """Fetch the table tag for a class object."""
     name = klass.__name__
     assert name[:6] == "table_"
     name = name[6:]  # Chop 'table_'
     return identifierToTag(name)
 
 
-def newTable(tag):
+def newTable(tag: str | Tag) -> DefaultTable:
```
jorenham:
```suggestion
def newTable(tag: str) -> DefaultTable:
```

ref: https://github.com/fonttools/fonttools/pull/3952/files#r2414777868

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -722,7 +914,7 @@ def _writeTable(self, tag, writer, done, tableCache=None):
         if tableCache is not None:
             tableCache[(Tag(tag), tabledata)] = writer[tag]
 
-    def getTableData(self, tag):
+    def getTableData(self, tag: str | Tag) -> bytes:
```
jorenham:
```suggestion
    def getTableData(self, tag: str) -> bytes:
```

ref: https://github.com/fonttools/fonttools/pull/3952/files#r2414777868

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -9,16 +11,128 @@
...
+    TypedDict,
+    Unpack,
+)
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping, MutableMapping
+    from typing import overload
+    from typing_extensions import Self
+
+    from fontTools.ttLib.tables import (
```
jorenham, line 31: How about `from fontTools.ttLib import tables`?

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -9,16 +11,128 @@
...
+    Sequence,
+    TYPE_CHECKING,
+    TypeVar,
+    TypedDict,
+    Unpack,
+)
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping, MutableMapping
+    from typing import overload
```
jorenham: `@overload` is needed at runtime unfortunately, even with `__future__.annotations`

jorenham:
Would it help if I'd open a PR to this branch with my proposed changes?

Lib/fontTools/ttLib/ttFont.py
dscorbett: For any function that starts with `tag = Tag(tag)`, the type of `tag` is not `str | Tag` (which is equivalent to `str`): it is anything that the `Tag` constructor accepts, which is `object`, because anything can be stringified. So the type of the `tag` parameter, here and in some other functions, should be `object`. In practice, it is probably generally true that `isinstance(tag, (bytes, str))`.

madig:
> Would it help if I'd open a PR to this branch with my proposed changes?

Feel free!

@dscorbett @anthrotype Maybe I'm confused about Tag and str. What should I do here? Do we expect the Tag class, which extends str? Should we allow both? Or just str?

Lib/fontTools/ttLib/ttFont.py
madig: Then with `... -> "tables._c_m_a_p.table__c_m_a_p": ...` I get `"_c_m_a_p" is not a known attribute of module "fontTools.ttLib.tables" Pylance`. Unless you know a trick, I have explicitly import all of them :(

madig: One could maybe change MetaTools/buildTableList.py to place an explicit `__all__ = [ ... ]` into the generated tables module.

Lib/fontTools/ttLib/ttFont.py
madig: I changed everythign to be `str | bytes` instead of `Tag` until we decide otherwise.

madig:
@jorenham I'm confused about the `_NumberT` thing; At https://github.com/fonttools/fonttools/blob/515fa761c60276ae9fb02ad574b04fc24a0e2d58/Lib/fontTools/ttLib/ttFont.py#L1346 I tried to add to `renormalizeLocation` the return type `dict[str, float]`, but the type checker didn't like it: 

```
Type "dict[str, float]*" is not assignable to declared type "Mapping[str, _NumberT@normalizeLocation]"
  "dict[str, float]*" is not assignable to "Mapping[str, _NumberT@normalizeLocation]"
    Type parameter "_VT_co@Mapping" is covariant, but "float" is not a subtype of "_NumberT@normalizeLocation"
      Type "float" is not assignable to type "_NumberT@normalizeLocation"
```

madig:
I have changed all typings of Tag to just `str | bytes`. There are some functions that don't do `tag = Tag(tag)` when they probably should, like in https://github.com/fonttools/fonttools/blob/515fa761c60276ae9fb02ad574b04fc24a0e2d58/Lib/fontTools/ttLib/ttFont.py#L854, but that can be fixed later.

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -1,24 +1,140 @@
...
+import traceback
+from io import BytesIO, StringIO, UnsupportedOperation
+from types import ModuleType, TracebackType
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    BinaryIO,
+    Literal,
+    Mapping,
+    Sequence,
```
jorenham: These could go under `TYPE_CHECKING`

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -1,24 +1,140 @@
+from __future__ import annotations
+
+import logging
+import os
+import traceback
+from io import BytesIO, StringIO, UnsupportedOperation
+from types import ModuleType, TracebackType
```
jorenham: seems like a good `TYPE_CHECKING` candidate

Lib/fontTools/ttLib/ttFont.py
jorenham: Hmm yea annoying... I suppose an `__all__` would be best, but that's probably out of scope

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -1106,7 +1657,12 @@ def sortedTagList(tagList, tableOrder=None):
     return orderedTables
 
 
-def reorderFontTables(inFile, outFile, tableOrder=None, checkChecksums=False):
+def reorderFontTables(
+    inFile: BinaryIO,  # Takes file-like object as per original
+    outFile: BinaryIO,  # Takes file-like object
+    tableOrder: list[Tag] | None = None,
```
jorenham: This won't accept `list[str]` now; is that ok?

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -1082,7 +1631,9 @@ def xmlToTag(tag):
 OTFTableOrder = ["head", "hhea", "maxp", "OS/2", "name", "cmap", "post", "CFF "]
 
 
-def sortedTagList(tagList, tableOrder=None):
+def sortedTagList(
+    tagList: Sequence[str], tableOrder: Sequence[str] | None = None
+) -> list[str | bytes]:
```
jorenham: `list[str | bytes]` can be pretty annoying to have to deal with for users, seeing as it's invariant. A `Sequence` (paradoxically), or, if realistic, `list[str] | list[bytes]`, might be easier

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -1046,7 +1595,7 @@ def tagToXML(tag):
         return tagToIdentifier(tag)
 
 
-def xmlToTag(tag):
+def xmlToTag(tag: str) -> Tag | str:
```
jorenham:
```suggestion
def xmlToTag(tag: str) -> str:
```
equiv

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -1007,7 +1556,7 @@ def tagToIdentifier(tag):
     return ident
 
 
-def identifierToTag(ident):
+def identifierToTag(ident: str) -> Tag | str:
```
jorenham:
```suggestion
def identifierToTag(ident: str) -> str:
```

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -492,25 +844,206 @@ def _readTable(self, tag):
...
+    def __delitem__(self, tag: str | bytes) -> None:
         if tag not in self:
             raise KeyError("'%s' table not found" % tag)
         if tag in self.tables:
             del self.tables[tag]
         if self.reader and tag in self.reader:
             del self.reader[tag]
 
-    def get(self, tag, default=None):
+    @overload
```
jorenham, line 857: Nice!

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -444,10 +620,185 @@ def ensureDecompiled(self, recurse=None):
                 table.ensureDecompiled(recurse=recurse)
         self.lazy = False
 
-    def __len__(self):
+    def __len__(self) -> int:
         return len(list(self.keys()))
 
-    def __getitem__(self, tag):
+    @overload
```
jorenham, line 625: Also very nice

Lib/fontTools/ttLib/ttFont.py
madig: Oh, oops. This should probably also be `list[str | bytes]`.

jorenham: I'd go with something like `list[Tag] | list[str] | list[bytes]`, or use the typevar trick

dscorbett: Can it be `Sequence[str | bytes] | None`?

jorenham: Yea I guess that would help. It would prohibit mutating it, which I'm guessing would be fine for most users, but for those that append to it, this might not be ideal. They'd have to `typing.cast` it to a `list` or something then, but that's not the end of the world.

dscorbett: The caller can pass a `list[str]` to a function that takes a `Sequence[str | bytes]`. No casting is needed and it doesn’t prevent the caller from mutating it. A simplified example:
```console
$ cat >x.py <<'# EOF'
from collections.abc import Sequence

def reorderFontTables(tableOrder: Sequence[str | bytes] | None = None) -> None:
    return None

tableOrder: list[str] = ["head", "GSUB"]
reorderFontTables(tableOrder)
tableOrder.append("GPOS")
# EOF

$ mypy x.py
Success: no issues found in 1 source file
```

jorenham: I meant the other way around, i.e. the user would have to cast the `Sequence` to a `list` if the user wants to append to the return type of `reorderFontTables` (in the scenario where we would've changed it to return a `Sequence`)

jorenham: Oh woops, nevermind: For some reason I thought we we talking about the `sortedTagList` return type. Apologies for the confusion

Lib/fontTools/ttLib/ttFont.py
madig: Some code tracing suggests that it's really just `list[str]`, so I did that.

madig:
@jorenham The number type issue remains, otherwise I tried to simplify a bit.

jorenham:
> I'll try to work on this this weekend, so consider the comments "notes to self". But I really like what you've done so far.

I wasn't able to find the time to work on this, sorry about that.

> @jorenham The number type issue remains, otherwise I tried to simplify a bit.

I'll take a look at it soon

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -780,7 +1313,9 @@ def getGlyphSet(
             glyphSet = _TTGlyphSetVARC(self, location, glyphSet)
         return glyphSet
 
-    def normalizeLocation(self, location):
+    def normalizeLocation(
+        self, location: Mapping[str, _NumberT]
+    ) -> dict[str, _NumberT]:
```
jorenham: This should to the trick:

```suggestion
    ) -> dict[str, float]:
```

the reasoning here is that, as far as I can tell, `fontTools.varLib.models.normalizeLocation` always returns a `dict[str, float]`

jorenham:
> @jorenham I'm confused about the `_NumberT` thing; At
> 
> https://github.com/fonttools/fonttools/blob/515fa761c60276ae9fb02ad574b04fc24a0e2d58/Lib/fontTools/ttLib/ttFont.py#L1346
> 
> I tried to add to `renormalizeLocation` the return type `dict[str, float]`, but the type checker didn't like it:
> ```
> Type "dict[str, float]*" is not assignable to declared type "Mapping[str, _NumberT@normalizeLocation]"
>   "dict[str, float]*" is not assignable to "Mapping[str, _NumberT@normalizeLocation]"
>     Type parameter "_VT_co@Mapping" is covariant, but "float" is not a subtype of "_NumberT@normalizeLocation"
>       Type "float" is not assignable to type "_NumberT@normalizeLocation"
> ```

The re-assignment to `location` is the main problem here, which can easily be worked around by using a different name for it.

madig:
@jorenham Thanks! I guess I'll deal with renaming variables in another PR.

@anthrotype This is ready :)

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -1313,9 +1313,7 @@ def getGlyphSet(
             glyphSet = _TTGlyphSetVARC(self, location, glyphSet)
         return glyphSet
 
-    def normalizeLocation(
-        self, location: Mapping[str, _NumberT]
-    ) -> dict[str, _NumberT]:
+    def normalizeLocation(self, location: Mapping[str, _NumberT]) -> dict[str, float]:
```
jorenham: the `_NumberT` trick shouldn't be necessary here because the value type parameter of `Mapping` is covariant

Lib/fontTools/varLib/models.py
```diff
@@ -1,5 +1,7 @@
 """Variation fonts interpolation models."""
 
+from __future__ import annotations
+
```
jorenham, line 4: this suggests that you intended to import the `Mapping` and `Sequence` in a `if TYPE_CHECKING: ` block

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -191,23 +319,31 @@ def __init__(
...
+        traceback: TracebackType | None,
+    ) -> None:
         self.close()
 
-    def close(self):
+    def close(self) -> None:
         """If we still have a reader object, close it."""
         if self.reader is not None:
             self.reader.close()
+            self.reader = None
```
anthrotype, line 340: this is a behavioral change, not purely typing related... although it's sensible in that prevents use-after-close. We can keep

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -1106,7 +1645,12 @@ def sortedTagList(tagList, tableOrder=None):
     return orderedTables
 
 
-def reorderFontTables(inFile, outFile, tableOrder=None, checkChecksums=False):
+def reorderFontTables(
+    inFile: BinaryIO,  # Takes file-like object as per original
+    outFile: BinaryIO,  # Takes file-like object
+    tableOrder: Sequence[Tag] | None = None,
```
anthrotype: I think the types are inconsistent here. `tableOrder` is annotated `Sequence[Tag] | None` and is passed straight to `sortedTagList`, where it is annotated `Sequence[str]`. Even though `Tag` is a `str` subclass, `Sequence[Tag]` isn't assignable to `Sequence[str]` because generic collection types are ["invariant"](https://peps.python.org/pep-0484/#covariance-and-contravariance).
You could make both `Sequence[str]` and coerce the Tag list to a list of str before passing it along. This would preserve runtime behavior while keeping typechecker happy.

Lib/fontTools/misc/xmlWriter.py
```diff
@@ -25,17 +28,19 @@
...
+        encoding: str = "utf_8",
+        newlinestr: str | bytes = "\n",
+    ) -> None:
         if encoding.lower().replace("-", "").replace("_", "") != "utf8":
             raise Exception("Only UTF-8 encoding is supported.")
         if fileOrPath == "-":
             fileOrPath = sys.stdout
+        self.filename: str | os.PathLike[str] | None
         if not hasattr(fileOrPath, "write"):
+            assert isinstance(fileOrPath, (str, os.PathLike))
```
anthrotype: in theory someone could pass any object that lacks a write method (like an int or None), and the assert would fail with an unhelpful AssertionError. I think a TypeError would be better here.

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -142,6 +268,7 @@ def __init__(
             return
         seekable = True
         if not hasattr(file, "read"):
+            assert isinstance(file, (str, os.PathLike))
```
anthrotype: same here, assertion error would trigger on abritrary non-file-like inputs, I'd prefer TypeError

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -232,6 +368,7 @@ def save(self, file, reorderTables=True):
         ):
             if reorderTables is False:
                 # sort tables using the original font's order
+                assert self.reader is not None
```
anthrotype: a freshly built TTFont() with no reader which gets saved with `reorderTable=False` would raise AssertionError. I suggest instead a proper `TTLibError("original table order unavailable")`

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -329,6 +486,7 @@ def _saveXML(
         if not splitTables:
             writer.newline()
         else:
+            assert writer.filename is not None
```
anthrotype: `splitTables=True` with an XMLWriter initialized with a stream or file-like object will assert here; better to raise TTLibError("splitTables requires a filesystem path, not a stream").

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -444,10 +610,185 @@ def ensureDecompiled(self, recurse=None):
...
 
-    def __len__(self):
+    def __len__(self) -> int:
         return len(list(self.keys()))
 
-    def __getitem__(self, tag):
+    @overload
+    def __getitem__(self, tag: Literal["BASE"]) -> B_A_S_E_.table_B_A_S_E_: ...
+    @overload
+    def __getitem__(self, tag: Literal["CBDT"]) -> C_B_D_T_.table_C_B_D_T_: ...
```
anthrotype, line 628: I wish we could automate all this (write some built-time script to generate all these overloads based on the table classes that are actually present, e.g. in case new ones get added) ... but perhaps in another PR

madig: Yeah, that'd be a separate effort :) I'd have to ask around if there was a magic way to fill this stuff in.

jorenham: As tempting as it might sound; code generation is almost never worth it in my experience.

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -1,24 +1,130 @@
...
-import traceback
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping, MutableMapping
+    from types import ModuleType, TracebackType
+    from typing import Any, BinaryIO, Literal, Sequence, TextIO
+
+    from typing_extensions import Self, Unpack
+
+    from fontTools.ttLib.tables import (
```
khaledhosny, line 31: Can this import be move to `fontTools/ttLib/tables/__init__.py` and table modules accessed bellow as `tables.B_A_S_E_` etc. to avoid this very long import here? Adding it there might also increase the chance of keeping the list up to date.

khaledhosny: OK, I should have read the comment above.

PR #3954, opened 2025-10-10
Add type hint for TTFont.get

op200:
Added `@overload` type hint for `get` to reduce static check errors, as `get("name")` is frequently used.


madig:
Already part of https://github.com/fonttools/fonttools/pull/3952, which deals with all tables.

op200:
It's so coincidental, amazing

PR #3958, opened 2025-10-17
Error when condition sets have the same name

Hoolean:
This is never valid, and raising an error with the location early can avoid some lengthy debugging -- I've not checked this yet though, and so it should be left as a draft until I've run it properly and added a test case.


Hoolean:
Added a test, and fixed the linting issue; this is now ready for review.

PR #3963, opened 2025-10-21
[bezierTools] Fix logic bug in curveCurveIntersections

### LINKED ISSUE #3962: small logic bug in bezierTools curveCurveIntersections
Specifically in the case where `is_linelike` is `true` for `curve`` but not for `curve2`: we end up calling `curveLineIntersections` switching the segment order, which means that `t1` and `t2` are incorrect in the result:

https://github.com/googlefonts/fonttools/blob/03a3c8ed9e7ab35b5f219a6d25ee8081107ff2bb/Lib/fontTools/misc/bezierTools.py#L1406-L1412

I have a fix, just wanted an issue to reference in the test case.

cmyr:
closes #3962

### DIFF
--- a/Tests/misc/bezierTools_test.py
+++ b/Tests/misc/bezierTools_test.py
@@ -1,5 +1,7 @@
+import math
 import fontTools.misc.bezierTools as bezierTools
 from fontTools.misc.bezierTools import (
+    _is_linelike,
     calcQuadraticBounds,
     calcQuadraticArcLength,
     calcCubicBounds,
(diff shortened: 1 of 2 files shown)


Tests/misc/bezierTools_test.py
```diff
@@ -233,3 +235,22 @@ def test_intersections_linelike():
     seg2 = [(0.0, 0.5), (0.25, 0.5), (0.75, 0.5), (1.0, 0.5)]
     pt = curveCurveIntersections(seg1, seg2)[0][0]
     assert pt == (0.0, 0.5)
+
+
+# see https://github.com/fonttools/fonttools/issues/3962
+def test_linelink_curve_curve_regression():
```
anthrotype:
```suggestion
def test_linelike_curve_curve_regression():
```

anthrotype (approved):

PR #3978, opened 2025-11-04
Bump unicodedata2 >= 17

### LINKED ISSUE #3971: Update to Unicode 17

anthrotype:
fixes #3971

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -112,7 +112,7 @@ extras_require = {
     # of the Unicode Character Database instead of the built-in unicodedata
     # which varies between python versions and may be outdated.
     "unicode": [
-        "unicodedata2 >= 17.0.0",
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -2,7 +2,7 @@
 # extension 'brotlipy' on PyPy
 brotli==1.1.0; platform_python_implementation != "PyPy"
 brotlicffi==1.1.0.0; platform_python_implementation == "PyPy"
-unicodedata2==15.1.0; python_version <= '3.11'
+unicodedata2==17.0.0
```
khaledhosny, line 5: The next Python version (3.15?) will have Unicode 17 data, should we preemptively check for it here?

anthrotype: ok, i'll keep the <= '3.14'

PR #3986, opened 2025-11-19
[feaLib] Sort BasLangSysRecords by tag

cmyr:
As required by the spec.

(@anthrotype pointed out in https://github.com/googlefonts/fontc/pull/1763/files#r2542066755, where I was also failing to sort these; fixing here so that I can reuse the test case)


cmyr:
I don't have merge rights here, but anyone should feel free.

PR #3990, opened 2025-11-27
Also CI-test on Python 3.14

madig:


moi15moi:
To fix https://github.com/fonttools/fonttools/actions/runs/19741206304/job/56565254249, bump (https://github.com/fonttools/fonttools/blob/9f810def1e4ae2ca97a8e2dd92ea4a95901388ff/requirements.txt#L13) to 0.9.0

anthrotype:
I think the last one is fixing pypy3 build. I'll take a look at https://github.com/fonttools/pyclipper/pull/65 now

Issue #3996, opened 2025-12-06
[FontBuilder] `setupOS2()` default params `_OS2Defaults.panose` could be globally polluted


TakWolf:
https://github.com/fonttools/fonttools/blob/4.61.0/Lib/fontTools/fontBuilder.py#L492-L511

```python
    def setupOS2(self, **values):
        """Create a new `OS/2` table and initialize it with default values,
        which can be overridden by keyword arguments.
        """
        self._initTableWithValues("OS/2", _OS2Defaults, values)
        ....
```

https://github.com/fonttools/fonttools/blob/4.61.0/Lib/fontTools/fontBuilder.py#L267-L309

```python
_panoseDefaults = Panose()

_OS2Defaults = dict(
    ....
    panose=_panoseDefaults,
    ....
)
```

Example:

```python
builder_1 = FontBuilder(...)
builder_1.setupOS2()
builder_1.font['OS/2'].panose.bProportion = 9

builder_2 = FontBuilder(...)
builder_2.setupOS2()
value = builder_2.font['OS/2'].panose.bProportion  # value is 9
````

PR #3997, opened 2025-12-06
Fix FontBuilder `setupOS2()` default params `_OS2Defaults.panose` global polluted

TakWolf:
Fix https://github.com/fonttools/fonttools/issues/3996

### DIFF
--- a/Lib/fontTools/fontBuilder.py
+++ b/Lib/fontTools/fontBuilder.py
@@ -264,49 +264,49 @@ _nameIDs = dict(
 # to insert in setupNameTable doc string:
 # print("\n".join(("%s (nameID %s)" % (k, v)) for k, v in sorted(_nameIDs.items(), key=lambda x: x[1])))
 
-_panoseDefaults = Panose()
 
-_OS2Defaults = dict(
-    version=3,
-    xAvgCharWidth=0,
-    usWeightClass=400,
-    usWidthClass=5,
-    fsType=0x0004,  # default: Preview & Print embedding
-    ySubscriptXSize=0,
-    ySubscriptYSize=0,
-    ySubscriptXOffset=0,
-    ySubscriptYOffset=0,
-    ySuperscriptXSize=0,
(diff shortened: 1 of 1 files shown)


behdad:
Another option is to use `deepcopy`. I let Just decide.

justvanrossum:
I'd slightly prefer deepcopy. Or maybe a helper func that returns a fresh copy of the os2 defaults:

```python
def _getOS2Defaults():
    return dict(
        version=3,
        ...
        panose=Panose(),
        ...
    )
```

justvanrossum:
Sorry, we weren't very explicit about the deepcopy: I meant to add a deepcopy in this line:

```python
    self._initTableWithValues("OS/2", deepcopy(_OS2Defaults), values)
```

That's prettier than the `if "panose" not in values` special case.

But now that I see it, I slightly prefer the `_getOS2Defaults()` approach sketched above.

TakWolf:
> Sorry, we weren't very explicit about the deepcopy: I meant to add a deepcopy in this line:
> 
> ```python
>     self._initTableWithValues("OS/2", deepcopy(_OS2Defaults), values)
> ```
> 
> That's prettier than the `if "panose" not in values` special case.
> 
> But now that I see it, I slightly prefer the `_getOS2Defaults()` approach sketched above.

Whether `_getOS2Defaults()` or `deepcopy(_OS2Defaults)`, both methods create an additional object (Although this performance loss can be ignored). And this default parameter is only used for iteration, not actually assigned to the target, the allocation is actually unnecessary.

justvanrossum:
> the allocation is actually unnecessary

You're not wrong, but in this case I would choose simplicity of the code (by avoiding the special case) over efficiency.

But... I'll let you decide.

TakWolf:
Change to: `def _getOS2Defaults()`

I will prioritize the opinions of the official maintainers.

anthrotype:
thanks. can you please run black to fix the `lint` CI job failing?

PR #4001, opened 2025-12-09
[designspaceLib] Preserve empty conditionsets when serializing

### LINKED ISSUE #4000: [designspaceLib] Empty conditionsets are lost when serialising to XML
designspaceLib drops empty conditionsets when serialising, but these are the only semantically correct way to represent a rule that always applies, and so they should be preserved instead.

## Motivation

This is a meaningful edge-case, as it is necessary for projects that contain multiple designspaces, where one may be a subset of the other. E.g. to represent an italic alternate for a project spanning `wght`+`ital`, a normal condition set would be used in the full designspace, whereas an empty one would be required in the sub-designspace representing the built VF pinned at `ital`.

This is covered by the designspace spec:

> [`<rule>` element](https://fonttools.readthedocs.io/en/stable/designspaceLib/xml.html#rule-element)
> - [...]
> - Each `<rule>` element contains one or more <conditionset> elements
> - Only one <conditionset> needs to be true to trigger the rule (logical OR). An empty condition set is considered to be true, as in, the rule will be always-on.
> - [...]

, compiled correctly by varLib:

https://github.com/fonttools/fonttools/blob/f87cc70a9fcef392ad06dc2266e4830546bcfa9a/Lib/fontTools/varLib/featureVars.py#L611-L616

, and in compliance with OpenType too:

> [If the ConditionSet offset is 0, there is no condition set table. This is treated as the universal condition: all contexts are matched.](https://github.com/fonttools/fonttools/blob/f87cc70a9fcef392ad06dc2266e4830546bcfa9a/Lib/fontTools/varLib/featureVars.py#L616)

and so is something we should preserve when round-tripping to and from XML.

## Reproducing

```python
# /// script
# requires-python = ">=3.10"
# dependencies = [
#     "fonttools",
# ]
# ///

from fontTools.designspaceLib import DesignSpaceDocument

doc = """\
<?xml version='1.0' encoding='UTF-8'?>
<designspace format="5.0">
  <rules>
    <rule>
        <conditionset></conditionset>
    </rule>
</rules>
</designspace>
"""

print(len(doc.rules))
print(DesignSpaceDocument.fromstring(doc).tostring())
```

### Expected

```
len(doc.rules)=1
<?xml version='1.0' encoding='UTF-8'?>
<designspace format="5.0">
  <rules>
    <rule>
        <conditionset></conditionset>
    </rule>
</rules>
</designspace>
```

### Actual

```
len(doc.rules)=1
<designspace format="5.0">
  <rules/>
</designspace>
```

## Fix

This line currently prunes them, and so it is probably sufficient for us to remove it and the comment above:

https://github.com/fonttools/fonttools/blob/f87cc70a9fcef392ad06dc2266e4830546bcfa9a/Lib/fontTools/designspaceLib/__init__.py#L1583-L1584

I'll open a draft PR shortly :)

Hoolean:
Fixes #4000

(just a draft for now, tests incoming)

### DIFF
--- a/Tests/designspaceLib/designspace_test.py
+++ b/Tests/designspaceLib/designspace_test.py
@@ -1189,3 +1189,20 @@ def test_get_axes(datadir: Path) -> None:
 
     assert ds.getAxis("Width") is ds.getAxisByTag("wdth")
     assert ds.getAxis("Italic") is ds.getAxisByTag("ital")
+
+
+# https://github.com/fonttools/fonttools/issues/4000
+def test_preserveEmptyConditionSet():
+    ds = DesignSpaceDocument()
+    ds.rules = [
+        RuleDescriptor(
+            conditionSets=[{}]  # Empty conditionset (implicit always enabled)
+        ),
+    ]
+
+    ds_str = ds.tostring()
+    assert b"<conditionset/>" in ds_str
+
+    roundtripped_ds = DesignSpaceDocument.fromstring(ds_str)
+    assert len(roundtripped_ds.rules) == 1
+    assert len(roundtripped_ds.rules[0].conditionSets) == 1


behdad:
Tests seem to be failing.

Hoolean:
@behdad It was still in the oven, but Ricky has kindly tidied, updated the relevant snapshots and furnished it with some tests, and it's ready for review proper now 🚀

PR #4002, opened 2025-12-10
[instancer] bug fix in avar2 full instancing

qxliu76:
Avoid double piecewiseLinearMap:
in normalize() and avar renormalizeLocation()


behdad:
Good catch!

PR #4004, opened 2025-12-12
[otlLib] buildCoverage: return empty Coverage instead of None

### LINKED ISSUE #4003: fonttools can write null offset in GDEF MarkGlyphSets coverage array
The spec does not say that these offsets are nullable: https://learn.microsoft.com/en-us/typography/opentype/spec/GDEF#mark-glyph-sets-table.

Maybe they _should_ be nullable, but I think it is best to just conform with the spec in this case?

anthrotype:
Per the OpenType spec, coverage offsets that are not explicitly documented as nullable should not be NULL.

Here I changed otlLib.builder.buildCoverage() to always return a Coverage object (with empty glyphs list if no glyphs supplied) instead of returning None.

This fixes null offsets being written in GDEF MarkGlyphSetsDef coverage arrays, and everywhere else buildCoverage is used.

The only nullable Coverage in MATH table is already handled specially, see: https://github.com/fonttools/fonttools/issues/4003#issuecomment-3645992252

Fixes #4003

### DIFF
--- a/Tests/feaLib/data/empty_filter_sets_and_mark_classes.ttx
+++ b/Tests/feaLib/data/empty_filter_sets_and_mark_classes.ttx
@@ -6,7 +6,8 @@
     <MarkGlyphSetsDef>
       <MarkSetTableFormat value="1"/>
       <!-- MarkSetCount=1 -->
-      <Coverage index="0" empty="1"/>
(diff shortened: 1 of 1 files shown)


cmyr:
Thanks!

Issue #1190, opened 2018-02-22
feature request: add `fonttools diff` to compare fonts with ttx


anthrotype:
I often find myself using ttx in combination with some diffing tool, so much that I wrote a `ttdiff` shell function that takes two fonts (and optionally a list of table tags), and redirect the output of ttx to `diff` (pipes it to `colordiff` to get nice colors and `less` for pagination): 

https://gist.github.com/anthrotype/6c3dce176314ec6ffb116b0f8561bee0

I guess many of us have their own variant of this. So I wonder, should we add a new sub-command, `fonttools diff` that uses python to do the same, so it works everywhere fonttools work and not just in a unix shell?

I think this is a nice first-time-contributor kind of PR. Any takers?

khaledhosny:
I use `vim -d` instead of `diff`, so if this can allow me to choose the diffing program, that would be super cool :wink:

anthrotype:
I was thinking more about using python own difflib for the diffing instead of an external program, for portability.. But I imagine it's gonna be much slower than the standard diff or vimdiff or similar programs. 
You know, the more I think about it, the more I think it may not be worth it, and a shell snippet just works fine. At least it does for me, as I don't work from a Windows command prompt.
But then sometimes I wish I can tell someone else "compare these fonts" without me having to tell them how to set up their shell, install this or that, and so forth.

anthrotype:
I'm sure someone is thinking, "why just textual diff, we are in python so let's compare the objects themselves and print out differences in a structured way"...
That'd be cool too, but it's a completely different thing and a bit more involved I believe. 

Just for reference, here is module to recursively compare python objects:
https://github.com/seperman/deepdiff
https://pypi.python.org/pypi/deepdiff

chrissimpkins:
Thoughts about a separate, single repository to collect fonttools based scripts (Python, shell, and others) under open licensing?  It would be useful to have a single repository where the scripts could be easily referenced for direct use or modification based upon project-specific needs.  I'm sure that there are many scripts out there that can be generalized and would be extremely helpful to others who share similar needs.

anthrotype:
I feel that's a bit out of scope. (And sorry I'm not volunteering to support another repository).

chrissimpkins:
[...]

PR #4007, opened 2025-12-16
Import `fdiff` as `fonttools diff`

behdad:
Fixes https://github.com/fonttools/fonttools/issues/1190

### DIFF
--- a/Lib/fontTools/diff/__main__.py
+++ b/Lib/fontTools/diff/__main__.py
@@ -30,6 +30,7 @@ except ImportError:
 
 
 def main() -> None:  # pragma: no cover
+    """Compare two fonts for differences"""
     # try/except block rationale:
     # handles "premature" socket closure exception that is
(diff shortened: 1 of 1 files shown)


behdad:
TODO:
- [ ] Make the external diff the default. It requires running `diff -u` and detect failure and fallback to diffLib,
- [ ] Currently the whole thing is too slow. I think the following logic should work: if GlyphOrder of the two fonts is the same, just compare each table's binary and dump to XML  and diff only those tables that are different...
- [ ] Add `--summary` to just list tables that are different, ala `fbdiff` https://github.com/justvanrossum/fbdiff
- [ ] At that point, remove the `rich` dependency, which is just used to show a "Processing..." progress bar.
- [ ] Implement the `less -R`.

behdad:
@anthrotype @justvanrossum This works now. WDYT about merging and addressing the TODO list in upcoming PRs?

PR #4012, opened 2025-12-16
[feaLib] write 0xFFFF instead of '0' for missing nameIDs

### LINKED ISSUE #4010: cvXX FeatureParamaters tables should not write zeros for missing name IDs
[The spec here is very strangely worded](https://learn.microsoft.com/en-us/typography/opentype/spec/features_ae#:~:text=The%20name%20ID%20provided%20by%20featUiLabelNameId%20is%20intended%20to%20provide%20a%20user%2Dinterface%20string%20for%20the%20feature;%20for%20example%2C%20“Capital%2Deng%20variants”.%20If%20set%20to%20NULL%2C%20no%20'name'%20table%20string%20is%20used%20for%20the%20feature%20name.), and talks about 'NULL', but 'NULL' is not conceptually valid for a non-offset, and `0` is a valid/reserved nameID. If no name is specified for fields in this table, we should be writing `0xFFFF`, which is the conventual way of representing 'no name ID'.

See https://github.com/googlefonts/fontc/pull/1839 for background

cmyr:
Specifically in FeatureParams for the cv features; the spec is poorly worded but a 0 value here may be misinterpreted by clients.

- fixes #4010

### DIFF
--- a/Tests/feaLib/data/cvparam_null.fea
+++ b/Tests/feaLib/data/cvparam_null.fea
@@ -1,6 +1,6 @@
 feature cv02 {
     # the missing fields are allowed, and the compiler should generate a FeatureParameters
-    # table where the missing name ids are `0`
+    # table where the missing name ids are `0xFFFF`
     cvParameters {
         FeatUILabelNameID {
             name "Tall W";


Tests/feaLib/data/cvparam_null.fea
```diff
@@ -0,0 +1,10 @@
+feature cv02 {
+    # the missing fields are allowed, and the compiler should generate a FeatureParameters
+    # table where the missing name ids are `0`
```
anthrotype:
```suggestion
    # table where the missing name ids are `0xFFFF`
```

cmyr:
filed https://github.com/MicrosoftDocs/typography-issues/issues/1227

cmyr:
I don't have permissions here, feel free to merge :)

PR #4015, opened 2025-12-17
Fix SegmentToPointPen edge case

### LINKED ISSUE #4014: Quad blob vs SegmentToPointPen woes
`SegmentToPointPen` does the wrong thing with a quad blob (a contour with only off-curve points) when the first and last points are the same.

```python
from fontTools.pens.recordingPen import RecordingPointPen
from fontTools.pens.pointPen import SegmentToPointPen

coordsGood = [(0, 0), (0, 0), (0, 100), (100, 100)]
coordsBuggy = [(0, 0), (0, 100), (100, 100), (0, 0)]  # same, but shifted by one

for coords in [coordsGood, coordsBuggy]:
    rpp = RecordingPointPen()
    sp = SegmentToPointPen(rpp)
    sp.qCurveTo(*coords, None)
    sp.closePath()

    outCoords = [args[0] for cmd, args, kwargs in rpp.value if cmd == "addPoint"]
    print(len(outCoords), outCoords)
```

Output:
```
4 [(0, 0), (0, 0), (0, 100), (100, 100)]
3 [(0, 0), (0, 100), (100, 100)]
```

Both outputs should be four long, so the second one is missing a point.

This happened in a real source font with a quad blob circle approximation of eight off-curve points that interpolates to a square, made of four pairs of off-curve points on the corners.

justvanrossum:
This fixes #4014 

Fix SegmentToPointPen pen edge case: only remove a duplicate final point during closePath() if it is an on-curve point.

### DIFF
--- a/Lib/fontTools/pens/pointPen.py
+++ b/Lib/fontTools/pens/pointPen.py
@@ -344,6 +344,11 @@ class SegmentToPointPen(AbstractPen):
     def closePath(self):
         if self.contour is None:
             raise PenError("Contour missing required initial moveTo")
+
+        # Remove the last point if it's a duplicate of the first, but only if both
+        # are on-curve points (segmentType is not None); for quad blobs
+        # (all off-curve) every point must be preserved:
+        # https://github.com/fonttools/fonttools/issues/4014
         if (
             len(self.contour) > 1
             and (self.contour[0][0] == self.contour[-1][0])


Lib/fontTools/pens/pointPen.py
```diff
@@ -344,7 +344,12 @@ def qCurveTo(self, *pts):
     def closePath(self):
         if self.contour is None:
             raise PenError("Contour missing required initial moveTo")
-        if len(self.contour) > 1 and self.contour[0][0] == self.contour[-1][0]:
```
anthrotype, line 347: I suggest you add a comment here like:

```
      # Remove the last point if it's a duplicate of the first, but only if both
      # are on-curve points (segmentType is not None); for quad blobs
      # (all off-curve) every point must be preserved:
      # https://github.com/fonttools/fonttools/issues/4014
```

PR #4024, opened 2026-01-06
[feaLib] Add shorthand for value at default location in variable scalar

khaledhosny:
Writing a number-only without an axis name and colon is used to mean this is the value for default location. For example, in a font whose `wght` default is 400:

```
(100 wght=900:120)
```

is shorthand for:

```
(wght=400:100 wght=900:120)
```

### DIFF
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -2190,10 +2190,17 @@ class Parser(object):
     def expect_variable_scalar_(self):
         self.advance_lexer_()  # "("
         scalar = VariableScalar()
+        default_seen = False
         while True:
             if self.cur_token_type_ == Lexer.SYMBOL and self.cur_token_ == ")":
                 break
             if self.cur_token_type_ == Lexer.NUMBER:
+                if default_seen:
+                    raise FeatureLibError(
+                        "Duplicate value for the default location",
+                        self.cur_token_location_,
+                    )
+                default_seen = True
                 location, value = {}, self.cur_token_
                 self.advance_lexer_()
             else:
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -2193,7 +2193,11 @@ class Parser(object):
         while True:
             if self.cur_token_type_ == Lexer.SYMBOL and self.cur_token_ == ")":
                 break
-            location, value = self.expect_master_()
+            if self.cur_token_type_ == Lexer.NUMBER:
+                location, value = {}, self.cur_token_
+                self.advance_lexer_()
+            else:
+                location, value = self.expect_master_()
             scalar.add_value(location, value)
         return scalar
 
--- a/Tests/feaLib/builder_test.py
+++ b/Tests/feaLib/builder_test.py
@@ -1344,6 +1344,28 @@ class BuilderTest(unittest.TestCase):
         ):
             addOpenTypeFeaturesFromString(font, features)
 
+    def test_variable_scalar_default(self):
+        """Test that missing axis name(s) in variable scalar means default location."""
+
+        features = """
+            feature kern {
+                pos two <0 (wght=900:22 12 wdth=150,wght=900:42) 0 0>;
+            } kern;
+        """
+
+        font = self.make_mock_vf()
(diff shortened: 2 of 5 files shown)


khaledhosny:
This was suggested by @simoncozens in https://github.com/adobe-type-tools/afdko/pull/1350#discussion_r636008312, and the [proposed Adobe variable sealers syntax](https://adobe-type-tools.github.io/feature_file_change_review/Variable_Feature_Guide#single-variable-values) has it, and [Glyphs’ syntax](https://handbook.glyphsapp.com/layout/variable-gpos/) has its own way to add value at default location. Having this supported by feaLib simplifies converting between them (in particular converting Glyphs’ syntax to feaLib’s).

Lib/fontTools/feaLib/parser.py
```diff
@@ -2193,7 +2193,11 @@ def expect_variable_scalar_(self):
         while True:
             if self.cur_token_type_ == Lexer.SYMBOL and self.cur_token_ == ")":
                 break
-            location, value = self.expect_master_()
+            if self.cur_token_type_ == Lexer.NUMBER:
+                location, value = {}, self.cur_token_
+                self.advance_lexer_()
+            else:
+                location, value = self.expect_master_()
             scalar.add_value(location, value)
```
anthrotype, line 2208: Could we reject a second bare value here? Both bare values use `{}` as the key, so the second overwrites the first e.g. `(10 20 wght=900:30)` is silently reduced to `(20 wght=900:30)`.

The AFDKO 5.0.1's FEA specification [permits at most one bare/default value](https://github.com/adobe-type-tools/afdko/blob/cef990e059a71644478d19c3c5fbb7c9e01dc0c6/docs/OpenTypeFeatureFileSpecification.md?plain=1#L454-L465), and AFDKO reports duplicate defaults as an [error](https://github.com/adobe-type-tools/afdko/blob/cef990e059a71644478d19c3c5fbb7c9e01dc0c6/c/shared/include/varsupport.h#L312-L329). A parser test for this case would be useful too.

anthrotype:
FWIW the PR description's example above is malformed and doesn't parse:

"""
For example, writing:

(100 wght:100,wdth=1)

as a shorthand for:

(wght:400,wdth:5=100 wght:100,wdth=10)
"""

I suggest replacing with something like:

"""
For example, in a font whose wght default is 400:

```fea
(100 wght=900:120)
```

is shorthand for:

```fea
(wght=400:100 wght=900:120)
```
"""

khaledhosny:
> FWIW the PR description's example above is malformed and doesn't parse:
> 
> """ For example, writing:
> 
> (100 wght:100,wdth=1)
> 
> as a shorthand for:
> 
> (wght:400,wdth:5=100 wght:100,wdth=10) """
> 
> I suggest replacing with something like:
> 
> """ For example, in a font whose wght default is 400:
> 
> ```opentype
> (100 wght=900:120)
> ```
> 
> is shorthand for:
> 
> ```opentype
> (wght=400:100 wght=900:120)
> ```
> 
> """

Thanks. I might have been using Glyphs syntax here subconsciously.

Lib/fontTools/feaLib/parser.py
khaledhosny: Done

Tests/feaLib/builder_test.py
```diff
@@ -1344,6 +1344,28 @@ def test_variable_scalar_duplicate_location(self):
         ):
             addOpenTypeFeaturesFromString(font, features)
 
+    def test_variable_scalar_default(self):
+        """Test that missing axis name(s) in variable scaler means default location."""
```
anthrotype:
```suggestion
        """Test that missing axis name(s) in variable scalar means default location."""
```

anthrotype (approved):
LGTM

this would need an equivalent change in fea-rs.. Are you volunteering? ;)

khaledhosny:
> LGTM
> 
> this would need an equivalent change in fea-rs.. Are you volunteering? ;)

I can give it a try (once I re-work the VARC PR).

anthrotype:
> I can give it a try

I got distracted.. https://github.com/googlefonts/fontc/pull/2084

PR #4027, opened 2026-01-08
Remove vestigial workaround for optimized variable composite delta problem on old macOS

madig:
End what #1788 started, a small cleanup :)

### DIFF
--- a/Lib/fontTools/varLib/instancer/__init__.py
+++ b/Lib/fontTools/varLib/instancer/__init__.py
@@ -954,11 +954,8 @@ def _instantiateGvarGlyph(
         # preserve backwards compatibility.
         #     See 0010a3cd9aa25f84a3a6250dafb119743d32aa40
         coordinates.toInt()
-
-        isComposite = glyf[glyphname].isComposite()
-
         for var in tupleVarStore:
-            var.optimize(coordinates, endPts, isComposite=isComposite)
+            var.optimize(coordinates, endPts)
 
 
 def instantiateGvarGlyph(varfont, glyphname, axisLimits, optimize=True):


Lib/fontTools/varLib/__init__.py
```diff
@@ -382,25 +382,7 @@ def _add_gvar(font, masterModel, master_ttfs, tolerance=0.5, optimize=True):
...
-                    )  # Shouldn't matter that this is different from fvar...?
-                    tupleData, auxData = var.compile(axis_tags)
-                    unoptimized_len = len(tupleData) + len(auxData)
-                    tupleData, auxData = var_opt.compile(axis_tags)
-                    optimized_len = len(tupleData) + len(auxData)
-
-                    if optimized_len < unoptimized_len:
-                        var = var_opt
-
+                var.optimize(origCoords, endPts, tolerance=tolerance)
```
behdad, line 385: I *think* we still need the size comparison. I might be wrong, but the `optimize` method work based on heuristics, like encoding the point numbers as shared or not, that can make the glyph's variation data become larger after optimize... I'm not 100% sure though.

Lib/fontTools/ttLib/tables/TupleVariation.py
```diff
@@ -620,17 +620,19 @@ def calcInferredDeltas(self, origCoords, endPts):
     def optimize(self, origCoords, endPts, tolerance=0.5, isComposite=False):
         from fontTools.varLib.iup import iup_delta_optimize
 
+        if isComposite:
+            warnings.warn(
+                "The isComposite argument is deprecated and may be removed in a future version",
+                DeprecationWarning,
+            )
+
```
Hoolean, line 628: As we're deprecating it, let's stop passing it internally too:

https://github.com/fonttools/fonttools/blob/d9e23c49e44f7bc157cfd6ee2158acf09e96dbda/Lib/fontTools/varLib/instancer/__init__.py#L958-L961

Hoolean: Maybe we should make the new default `None` and show the warning if it is anything other than that too, in case someone is passing a variable arg that might be `False` in explicitly

Lib/fontTools/varLib/__init__.py
madig: What do you mean? Looking at the removed code and the code in optimize(), they look identical to me if `support` is `self.axes`? Am I missing something?

behdad: Ooh. My bad. Thanks. LGTM.

behdad:
Oof. Nice. Thanks!

PR #4037, opened 2026-02-06
[pens] Type annotate pens and related modules

knutnergaard:
This adds type annotation to all  unannotated `pen` modules and improves annotation in the rest.

It also adds type annotations to related `misc` modules:
- arrayTools
- bezierTools
- enumTools
- roundTools
- transform

`annotations` is also expanded and organized.

`mypy.ini` is bumped to python 3.10 and annotations follow 3.10+ syntax.

The `addComponent` method has conflicting signatures between certain base classes and subclasses, which should be addressed to make mypy happy. These are marked with `type: ignore[override]`.


madig:
@anthrotype mentioned that the newly annotated modules these past few typing PRs weren't being checked by mypy actually, which causes problems for him internally.

mypy.ini currently is set up to only check `Lib/fontTools/misc/plistlib`. Can you please add all of the modules you added typing for as well and look what mypy stumbles over?

knutnergaard:
> @anthrotype mentioned that the newly annotated modules these past few typing PRs weren't being checked by mypy actually, which causes problems for him internally.
> 
> mypy.ini currently is set up to only check `Lib/fontTools/misc/plistlib`. Can you please add all of the modules you added typing for as well and look what mypy stumbles over?

Setting `files = Lib/fontTools/`  in `mypy.ini` and running mypy on `pens` and `misc` reveals a lot more errors than what was initially flagged – almost none of which originate in the modules I've been working on, and none of which are flagged in VsCode. Why is mypy set up this way? How to proceed?

knutnergaard:
@madig Seems like `follow_imports = silent` in combination with files being partially untyped  is the culprit. Setting that option to `skip` might be a good idea, as long as typing is spotty.

At least with that option set, all is well with regard to this PR.

knutnergaard:
Took care of the errors originating from the files in this PR, which showed up after modifying `mypy.ini`. Should be ready for review, unless other files need to be addressed as well.

madig:
I set mypy up this way initially because I only added typing to plistlib (or someone else did? I can't remember and there is no way to find out), so I only wanted to check that. Did you intend to commit your changes to mypy.ini?

knutnergaard:
Yes, but my only commit was the Python version bump. Since FontTools have moved beyond 3.9, I thought it a good idea to move annotations along with it.

madig:
I mean, we want mypy to check more modules by default, meaning you could add your newly annotated modules to the list of files to check :) If that presents too many problems, let's see how to fix or work around them.

knutnergaard:
> I mean, we want mypy to check more modules by default, meaning you could add your newly annotated modules to the list of files to check :) If that presents too many problems, let's see how to fix or work around them.

Sure, it just feels very awkward to add each individual file for mypy to check, but maybe there's no other way, given the circumstances.

Anyway, I added the files and fixed a couple of errors which surfaced after the edit to `mypy.ini`. 

I also inadvertently closed the PR, which evidently caused tests to fail after reopening. Are there caches to clear or something else to do to fix this?

Edit: Test failures were due to an ImportError, which VSCode didn't flag. Solved.

knutnergaard:
Added a few more pens and made som small improvements, pending a review.

knutnergaard:
Setting to draft until all pens are done.

knutnergaard:
All done. Ready for review.

ihsandeniz:
While going through `TeePen` I noticed this PR changes its behaviour beyond typing, so here is what I measured on current `main` (718b61b55) and on this PR's head (b60384d8), calling `moveTo` on each construction with `RecordingPen`s:

| construction | `main` | this PR |
|---|---|---|
| `TeePen(a, b)` | ✅ | ✅ |
| `TeePen(a)` (single pen) | ❌ `TypeError: 'RecordingPen' object is not iterable` | ✅ |
| `TeePen([a, b])` (documented: *"or TeePen(iterableOfPens)"*) | ✅ | ❌ `AttributeError: 'list' object has no attribute 'moveTo'` |
| `TeePen(pen for pen in ...)` | ⚠️ first call reaches both pens, every later call reaches none (the generator is exhausted) | ❌ `AttributeError: 'generator' object has no attribute 'moveTo'` |

So the PR fixes the single-pen case, which has been broken on `main` since the class was added, but `_iterPens` now treats the documented iterable form as a single pen. Nothing inside `Lib/` constructs a `TeePen`, and `Tests/` has no test for it, which is probably why this didn't show up in CI.

One way to keep all four working, in case it's useful:

```python
def __init__(self, *pens):
    if len(pens) == 1 and not hasattr(pens[0], "moveTo"):
        pens = pens[0]
    self.pens = tuple(pens)
```

I tried it against the same four constructions (plus `TeePen()`), and each pen receives every call. Materializing into a tuple also fixes the generator case, but it means that appending to a list after construction no longer affects the `TeePen`. I don't know whether anyone relies on that, so it's your call. If it helps, I'm happy to send a small `Tests/pens/teePen_test.py` covering these cases, either to this branch or as a separate PR after this one lands, whichever creates less churn for you.

PR #4038, opened 2026-02-13
Fix formatting of skia-pathops link in README

wzjoriv:
The link for skia-pathops library pointed to `https://pypi.python.org/pypy/skia-pathops` instead of teh correct link `https://pypi.python.org/pypi/skia-pathops`.


benkiel:
Nice catch, thank you!

PR #4045, opened 2026-02-27
[instancer] Implement avar2 partial-instancing

behdad:
For detailed design, see:

  https://github.com/fonttools/fonttools/blob/avar2-instancing/partial-instancing-avar2-design.md

TL;DR:

Whereas my previous attempt [0] at avar2 partial-instancing design had failed, this new PR goes a different way completely: When partial-instancing avar2 fonts, thee avar2 ItemVariationStore is tasked to adjust the designspace post-avar1 to look exactly like before the partial-instancing. It then computes the range of reachable normalized ranges for all axes, and culls the unreachable tents from all variation-stores.

This works through a very narrow hole in the standard: whereas normally it is assumed that all variation axes being at default result at the default instance, with avar2 that requirement was lifted and never codified. That allows us to use the avar2 ItemVariationStore to encode a "bias" (a region with no active axes), which allows for changing the default instance of the font without touching the glyf/gvar data at all.

One upside of this approach is that, unlike partial-instancing non-avar2 fonts, the font size will *never* blow up as a result of the partial-instancing. It can only shrink. 🥳

IMPORTANT NOTE: For the partial-instancer to be able to reduce file size, you want the parametric & otherwise internal axes to be marked private in the fvar table. If they are _not_, then no file size reduction is possible since it is assumed that the full range of the parametric axes should still be exposed to the user.

cc @lorp @qxliu76 @anthrotype @rsheeter @davelab6

[0] https://docs.google.com/document/d/1iVp0cRZx4co5hg97y19JD7ERc51WmOfwu1qFB4DyDwo/edit


behdad:
@jenskutilek Mind testing this please?

behdad:
@qxliu76 @garretrieger We should port this to hb-subset at some point.

behdad:
[Crispy-Regular.ttf.gz](https://github.com/user-attachments/files/26036949/Crispy-Regular.ttf.gz)

This is a version of Crispy font with the default dragged to wght=400 wdth=100 using this branch. Currently, the default instance renders fine with HarfBuzz, FreeType, and Skrifa, but not with CoreText and fontTools.ttGlyphSet. I'll fix the latter soon.

behdad:
I threw the latest Claude with maximum power on this, and it unearthed two real bugs, which are now fixed. It also did some exhaustive testing. My confidence in this work is very high now. @anthrotype

behdad:
@anthrotype Any objections to merging? We just merged the HarfBuzz counterpart, which was reviewed by Ginger.

behdad:
Addressed Fable review by @anthrotype and @m4rc1e.

behdad:
@anthrotype Actually in harfbuzz I decided if font is avar2, do the avar2 partial-instancing even if VarStore is empty. This is better default, No?

anthrotype:
Sorry for the slow reply, as you know I was on vacation. Each time I look at this branch, it takes me half a day just to warm up 😅  

Ok so, I re-reviewed at the current head (bf6f827) and found four remaining ways the culling can delete live variation data, plus a crash:

1) `getExtremes` misses the F2Dot14 point adjacent to one-sided tents (start==peak / peak==end); 
2) reachable bounds don't account for the final `otRound`, so a tuple one F2Dot14 unit away gets culled; 
3) one-sided endpoint clamping can potentially invert a range that overshoots +/-1;
4) and full instancing raises IndexError on out-of-range implicit VarStore indices (partial avar2 instancing treats them as zero deltas, as do HarfBuzz and fontations)

See the branch https://github.com/fonttools/fonttools/tree/avar2-instancing-regression-fixes which contains regression tests first followed by four fixes that turn them green. Feel free to merge/cherry-pick into your PR.

No objections to merging from me with those in.

> in harfbuzz I decided if font is avar2, do the avar2 partial-instancing even if VarStore is empty

agreed, never downgrading is the better default (my branch is built on top of that commit).

PR #4063, opened 2026-03-12
[ufoLib] Add description and annotation to fontInfoAttributes value data

### LINKED ISSUE #3862: Add type annotations and descriptions to `ufoLib.fontInfoAttributesVersion...ValueData`
This would be great for documentation and type hinting of font info attributes.

KrZ-W:
## Summary

Adds `description` and `annotation` fields to all entries in `fontInfoAttributesVersion2ValueData` and `fontInfoAttributesVersion3ValueData`, as discussed in #3862.

- **description**: Concise text from the [UFO specification](https://unifiedfontobject.org/versions/ufo3/fontinfo.plist/)
- **annotation**: Python type hint string using standard `typing` types (`Union`, `List`, `Dict`, `Any`)

This enables better documentation and type hinting of font info attributes for IDE autocomplete and static analysis.

### Design decisions per maintainer guidance

- Kept `dict(a=b)` constructor style (per @justvanrossum's [comment](https://github.com/fonttools/fonttools/issues/3862#issuecomment-2697461729))
- Used UFO spec descriptions verbatim where concise, condensed where long
- Used standard `typing` module types (not `annotated-types`)
- `"integerList"` types that per UFO spec may contain floats (e.g. PostScript blues/stems) use `"List[Union[int, float]]"`; pure-integer lists (flags, ranges) use `"List[int]"`

### Test results

All 299 existing ufoLib tests pass. The 5 pre-existing failures in GLIF2/glifLib tests (unrelated to this change) remain unchanged.

Closes #3862

### DIFF
--- a/Lib/fontTools/ufoLib/__init__.py
+++ b/Lib/fontTools/ufoLib/__init__.py
@@ -41,7 +41,7 @@ import zipfile
 from collections import OrderedDict
 from copy import deepcopy
 from os import fsdecode
-from typing import IO, TYPE_CHECKING, Any, Optional, Union, cast
+from typing import IO, TYPE_CHECKING, Any, Dict, List, Optional, Union, cast
 
 from fontTools.misc import filesystem as fs
 from fontTools.misc import plistlib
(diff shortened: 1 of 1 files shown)


knutnergaard:
Is there a benefit to defining annotations as strings rather than `typing` objects, except for  avoiding the extra imports? 

If Strings are indeed most appropriate, I guess testing them for validity would be in order.

KrZ-W:
Good question! The issue author (@anthrotype) gave specific guidance in #3862 to use string annotations and to avoid `annotated-types` as a dependency. The examples in the issue itself use strings like `"int"`, `"str"`, `"Optional[bool]"` etc.

String annotations have a few practical benefits here:
- No extra imports needed (the data structures are plain dicts)
- Forward-compatible — they don't constrain the runtime type system
- They match the existing pattern in fontTools where type info is informational rather than enforced

That said, I agree that validation of the string format could be useful. Happy to add that if the maintainers want it — perhaps a simple check that each annotation string is a valid Python type expression.

knutnergaard:
@KrZ-W Sorry, the link to the issue authored by @anthrotype you're referencing appears to be wrong. Could you edit at your convenience?

anthrotype:
What issue did I author?

PR #4064, opened 2026-03-13
Build abi3 wheels for Cython extensions

anthrotype:
Build the optional Cython extensions as cp310-abi3 wheels instead of per-CPython-version wheels.

This should substantially reduce the number of published binary wheels (49 currently!) while keeping support for Python 3.10+.

The free-threaded CPython 3.14t wheel builds are skipped since abi3 wheels are [currently incompatible](https://docs.python.org/3.14/howto/free-threading-extensions.html#limited-c-api-and-stable-abi) there, and we don't actually test those specifically either.
fontTools still publishes the pure-Python wheel, so unsupported environments can continue to install that one.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -9,7 +9,8 @@ on:
 
 env:
   # skip binary wheels for pypy (preferable to use pure-python) and 32-bit Linux
-  CIBW_SKIP: pp* cp*linux_i686
+  # skip free-threaded CPython too: abi3 wheels are currently incompatible there
+  CIBW_SKIP: pp* cp*linux_i686 cp31?t-*
   CIBW_ENVIRONMENT: FONTTOOLS_WITH_CYTHON=1
   CIBW_TEST_REQUIRES: tox
   # only test core fonttools library without extras for now, stuff like lxml or scipy
--- a/Lib/fontTools/cu2qu/cu2qu.py
+++ b/Lib/fontTools/cu2qu/cu2qu.py
@@ -1,6 +1,4 @@
 # cython: language_level=3
-# distutils: define_macros=CYTHON_TRACE_NOGIL=1
-
 # Copyright 2015 Google Inc. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/Lib/fontTools/qu2cu/qu2cu.py
+++ b/Lib/fontTools/qu2cu/qu2cu.py
@@ -1,6 +1,4 @@
 # cython: language_level=3
-# distutils: define_macros=CYTHON_TRACE_NOGIL=1
-
 # Copyright 2023 Google Inc. All Rights Reserved.
 # Copyright 2023 Behdad Esfahbod. All Rights Reserved.
 #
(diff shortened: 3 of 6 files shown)


anthrotype:
it worked! https://github.com/fonttools/fonttools/actions/runs/23055286315

8 wheels in 13m 13s (vs 49 wheels in 24m 23s, 54% wall-clock time reduction)

```
fonttools-4.62.2.dev0-cp310-abi3-macosx_10_9_x86_64.whl
fonttools-4.62.2.dev0-cp310-abi3-macosx_10_9_universal2.whl
fonttools-4.62.2.dev0-cp310-abi3-win_amd64.whl
fonttools-4.62.2.dev0-cp310-abi3-win32.whl
fonttools-4.62.2.dev0-cp310-abi3-manylinux1_x86_64.manylinux2014_x86_64.manylinux_2_17_x86_64.manylinux_2_5_x86_64.whl
fonttools-4.62.2.dev0-cp310-abi3-musllinux_1_2_x86_64.whl
fonttools-4.62.2.dev0-cp310-abi3-manylinux2014_aarch64.manylinux_2_17_aarch64.manylinux_2_28_aarch64.whl
fonttools-4.62.2.dev0-cp310-abi3-musllinux_1_2_aarch64.whl
```

anthrotype:
cibuildwheel builds the wheel once on 3.10 and then runs the tests on all the supported pythons (3.10-3.14).

madig:
Out of curiosity, have you done performance tests?

anthrotype:
Actually.. I haven't. I probably should have before merging. I'll see if I can make some benchmarks later today.
Do you believe this would slow things down considerably? I don't believe cu2qu or bezier math in various ufo2ft filters are the bottleneck in fontmake builds, at least from what I recall. But it's worth testing before committing to this.

anthrotype:
So.. I ran benchmarks comparing regular cpython-version-specific wheels vs abi3 on macOS (M1 Pro, Python 3.14.2). Scripts are in abi3-benchmarks branch if anyone wants to reproduce.

In the micro-benchmarks (which directly perf the Cython-accelerated functions in hot loops), the `abi3` modules are about ~27% slower. The stable ABI requires C API access to go through function calls rather than inlined macros (see [c-api docs](https://docs.python.org/3/c-api/stable.html#limited-api-scope-and-performance)). Also [Cython's own docs](https://cython.readthedocs.io/en/latest/src/userguide/limited_api.html) note the performance penalty is most pronounced for code interacting heavily with Python objects, such as the case for fonttools (unlike, say, uharfbuzz where most of the work is in the external C library).

In end-to-end fontmake compilation (-o variable), which exercises cu2qu + iup + feaLib lexer:

| Font | Regular | ABI3 | Diff |
|---|---|---|---|
| Oswald (small, .glyphs) | 1.47s | 1.53s | +3.8% |
| Google Sans (large, .designspace) | 181s | 187s | +3.3% |

The ~27% micro overhead shrinks to ~3-4% in practice because Cython-accelerated code is a small fraction of the overall pipeline. On a 3-minute build that's ~6 seconds.

Whether that trade-off is worth the reduced wheel matrix, I don't know... The convenience of the wheels just working when new minor CPython is released is nice. On the other hand, the wheel building CI workflow is only run on tagged commits, so we aren't paying a huge cost to build 49 of them. We might incur in PyPI storage limits at some point (like uharfbuzz did).

I guess I'll revert it for now. Thanks for having me double check on this.

PR #4065, opened 2026-03-14
Add support for Apple Color Emoji `bgcl` table

PoomSmart:
Add support for Apple Color Emoji `bgcl` table (JSON payload).

Purpose: stores Apple Color Emoji background palettes as a UTF‑8 JSON blob (palette colors + per‑emoji palette indices). The system font reads it at runtime to style emoji backgrounds. It is used in iOS 16+ lock screen emoji wallpaper.

### DIFF
--- a/Tests/ttLib/tables/_b_g_c_l_test.py
+++ b/Tests/ttLib/tables/_b_g_c_l_test.py
@@ -1,0 +1,27 @@
+import importlib
+import unittest
+
+
+class Test_b_g_c_l(unittest.TestCase):
+    """Unit tests for the `bgcl` table implementation."""
+
+    def test_compile_decompile_roundtrip(self):
+        mod = importlib.import_module("fontTools.ttLib.tables._b_g_c_l")
+        cls = getattr(mod, "table__b_g_c_l")
+
+        sample = {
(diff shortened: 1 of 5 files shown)


behdad:
Can you provide more information about this table and what it is used for?

PoomSmart:
@behdad Hi, done.

Tests/ttLib/tables/_b_g_c_l_test.py
```diff
@@ -0,0 +1,26 @@
+import importlib
+import unittest
+
+
+class Test_b_g_c_l(unittest.TestCase):
+    """Unit tests for the `bgcl` table implementation."""
+
+    def test_compile_decompile_roundtrip(self):
+        mod = importlib.import_module("fontTools.ttLib.tables._b_g_c_l")
+        cls = getattr(mod, "table__b_g_c_l")
```
behdad: Can't you use `fontTools.ttLib.getTableClass`?

behdad (approved):
Minor comment. The rest looks good, assuming that it works for you.

Tests/ttLib/tables/_b_g_c_l_test.py
PoomSmart: Yes, we can. Updated.

PR #4071, opened 2026-03-23
[ci] add riscv64 wheel builds using RISE RISC-V native runners

anthrotype:
supersedes #4069

/cc @gounthar

https://github.com/apps/rise-risc-v-runners

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -54,6 +54,8 @@ jobs:
             arch: auto32
           - os: ubuntu-24.04-arm
             arch: aarch64
+          - os: ubuntu-24.04-riscv
+            arch: riscv64
     steps:
     - uses: actions/checkout@v6
       with:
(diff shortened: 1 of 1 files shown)


anthrotype:
unfortunately even with native runners, it takes 2 hours and 43 minutes, which is still too long...

https://github.com/fonttools/fonttools/actions/runs/23442692981/job/68197789166

that's for 12 wheels: (3.10, 3.11, 3.12, 3.13, 3.14, 3.14t) x (manylinux, musllinux)

even if we drop 3.14t (free-threaded, not really tested on our end) and musllinux (is that even used on riscv64?), it would still take more than an hour to complete...

gounthar:
Dropping musllinux makes sense - the RISE runners are Ubuntu-based so manylinux is the natural fit. And starting with just cp312 + cp313 (skipping 3.10/3.11 and free-threaded) would bring it under 1 hour.

Can always expand later once RISE runner hardware gets faster (current gen is Scaleway EM-RV1 which isn't the quickest).

Should I close my PR #4069 since this one supersedes it?

Issue #4070, opened 2026-03-22
Add support for IFT and IFTX tables


wmedrano:
This is an upcoming font format described at https://www.w3.org/TR/IFT/#font-format-extensions

- `ttx` support would be good for debugging

wmedrano:
FYI: I'm working on this. Is there any preference between using `otData` vs a custom implementation?

The weirdest part of the IFT spec is the use of a union type



Type | Field | Description
-- | -- | --
uint8 / uint16 | entryIndex[glyphCount - firstMappedGlyph] | The entry index for glyph $i$ is stored in entryIndex[i - firstMappedGlyph]. Array members are uint8 if maxEntryIndex is less than 256; otherwise, they are uint16.

behdad:
`otData` is highly preferred.

For the union type you mention, I think a custom type makes sense. Just keep them minimal to what's absolutely necessary.

You can even use `otData` to declare two members: `entryIndex8` and `entryIndex16` and conditionalize one on `mapEntryIndex < 256` and the other on `mapEntryIndex >= 256`, then add a `postRead` method to combine them. Likewise, use `preWrite` to convert back.

wmedrano:
# Current Status

- Format2 supported for IFT and IFTX

# Next Steps

- Play around with it and monitor for any issues
- Implement Format1
  - Low priority (I'm not actively working on it) as we don't have any Format1 fonts yet.

PR #4072, opened 2026-03-25
Add IFT table

wmedrano:
For #4070 

- Will help us debug the upcoming IFT format
  - Introduces the "IFT " and "IFTX" tables which use the PatchMapFormat2 schema.
- Only Format2 was implemented as it is the only format we have implemented
- Verified across all the fonts in https://github.com/garretrieger/ift-demo/tree/main/docs/fonts
- MappingEntriesConverter uses custom code since it has a lot of custom logic

### DIFF
--- a/Lib/fontTools/ttLib/tables/otConverters.py
+++ b/Lib/fontTools/ttLib/tables/otConverters.py
@@ -141,6 +141,11 @@ class BaseConverter(object):
             "BaseGlyphRecordCount",
             "LayerRecordCount",
             "AxisIndicesList",
+            # IFT - propagated to GlyphMap/FeatureMap/MappingEntries subtables
+            "NumGlyphs",
+            "MaxGlyphMapEntryIndex",
+            "MaxEntryIndex",
+            "NumEntries",
         ]
         self.description = description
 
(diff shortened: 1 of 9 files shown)


Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -2017,6 +2022,278 @@ class CompositeMode(_UInt8Enum):
     enumClass = _CompositeMode
 
 
+class MappingEntryFormat:
+    def __init__(self, value):
+        self.value = value
+
+    @property
+    def has_subset_def(self):
+        return bool(self.value & 0x01)
```
behdad: These values maybe should be a flags enum, at least some time in the future.

wmedrano: Done, I think

Lib/fontTools/ttLib/tables/otConverters.py
```diff
@@ -2017,6 +2023,262 @@ class CompositeMode(_UInt8Enum):
...
+                    encoded_value = (delta << 1) | has_more
+                    writer.writeUInt24(encoded_value & 0xFFFFFF)
+                    lastId = eid
+            else:
+                lastId += 1
+
+            if formatFlags & MappingEntryFormat.HAS_PATCH_FORMAT:
+                writer.writeUInt8(entry.get("patchFormat", 0))
+
+            if formatFlags & MappingEntryFormat.CODE_POINT_BIAS_BITS:
```
wmedrano, line 2177: It's better to recompute the formatFlags based on the value of codePointsBias

wmedrano:
Ready for review!

Lib/fontTools/ttLib/tables/otConverters.py
wmedrano: Added a TODO at the top of the function for this.

behdad:
Looks good to me. Thanks.

Please run `MetaTools/buildTableList.py` and commit the changes as well.

wmedrano:
Tool is run, ready to submit

PR #4078, opened 2026-04-05
[ttLib] Add support for AppleColorEmoji CCF table, sbix glyph type "iden" and Avoid error parsing `hdmx` table when `maxp.numGlyphs` doesn't match the actual record

### LINKED ISSUE #962: AppleColorEmoji ccf format
Hi everyone,

I am looking for info on the new ccf font format that was included around iOS 9 (not sure on this, @vXBaKeRXv might know). 

Originally, Apple stored all their emojis in the `sbix` table in a ttf file. Now they appear to have switched to a combination of ttf and ccf, or ttc and ccf. The ccf file contains all the actual image data, but not in any format that I can determine. 

The `sbix` table now uses a `graphicType` of `iden`, which may be short for "identifier." Its hex data contains lots of patterns, but I can't figure out what they refer to. I assume they are some sort of offset:

image

I've shared fonts from iOS 9.1, including the ttf and ccf files. 

https://1drv.ms/f/s!ArbWhCPYeIx2gYZEuRu47_Ecxfaf6Q

Any information on this file type would be really appreciated. Maybe if we can get enough information on it, we can add it into fonttools.

PoomSmart:
## Summary

Adds a complete implementation of Apple's private `accf` table, used in colour-emoji fonts on iOS 8.3–9.3 (e.g. `AppleColorEmoji@2x.ccf`). The table stores glyph images with a proprietary palette + run-length compression scheme.

LLM with IDA Pro MCP was used to help implement the features.

Both known wire formats are supported:

| | Version 1 (iOS 8.x) | Version 2 (iOS 9.x) |
|---|---|---|
| Magic | `\x40\x30\x20\x10` | `fcca` |
| Header size | 48 112 bytes | 105 332 bytes |
| LUT entries/tier | 2 000 | 2 500 |
| Sizes order | largest first | smallest first |

The image-record format (palette, runs table, bitstream) is identical between versions.

## Changes

- `Lib/fontTools/ttLib/tables/_h_d_m_x.py` — fix `IndexError` for out-of-range glyph IDs:
  - `_GlyphnamedList.__getitem__` now returns `0` instead of raising `IndexError` when a
    glyph ID exceeds the number of entries in the `hdmx` record (can happen when `maxp.numGlyphs`
    disagrees with the actual record length)
  - _This happens exclusively on iOS 8.x AppleColorEmoji font_

- `Lib/fontTools/ttLib/tables/sbixGlyph.py` — extend `iden` graphic type support:
  - `is_reference_type()` now recognises `"iden"` alongside `"dupe"` and `"flip"`
  - `iden` payload is an ASCII glyph-name string (e.g. `0023 20E3_40`), not a uint16 GID
  - `decompile` / `compile` / `fromXML` handle `iden` payloads correctly
  - _`iden` graphic type is found on iOS 9.x AppleColorEmoji sbix (on iOS 8.x it would be just `????`)_

- `Lib/fontTools/ttLib/tables/_a_c_c_f.py` — new table module:
  - `_decode_palette` / `_encode_palette` — 21-bit packed RGB codec
  - `decode_image_record` / `encode_image_record` — CCF ↔ raw RGBA
  - `_rgba_to_png` / `_png_to_rgba` — helpers with Pillow fallback to pure-Python PNG
  - `CcfGlyph` — lazy PNG decode from raw record on first access
  - `CcfStrike` — one resolution tier
  - `table__a_c_c_f` — `decompile` / `compile` / `toXML` / `fromXML`
  - `_rawHeader` preserved for lossless round-trips when no edits are made
  - `rawRecordData` shortcut in `compile()` — unedited glyphs re-emit their original bytes verbatim
  - Image deduplication in `compile()` — identical records share one physical copy (matches Apple)

## Usage

```python

# Open the .ccf sidecar (works for both v1 and v2)
font = TTFont(sfVersion="ccff")
font["accf"] = font.newTable("accf")
with open("AppleColorEmoji@2x.ccf", "rb") as f:
    font["accf"].decompile(f.read(), font)

accf = font["accf"]

# List available pixel sizes
print(accf.strikes.keys())  # e.g. dict_keys([40, 60, ...])

# Extract one glyph image as PNG bytes
strike = accf.strikes[40]
glyph  = strike.glyphs["uni1F600"]  # grinning face
png_bytes = glyph.imageData          # decoded on first access

with open("grinning_40px.png", "wb") as f:
    f.write(png_bytes)

# Re-compile back to CCF (round-trip)
with open("out.ccf", "wb") as f:
    f.write(font["accf"].compile(font))
```

## Notes

- Re-encoded files are **~105–106%** of Apple-produced sizes.
  The encoder uses frequency-sorted palettes, delta opcodes (keep/±1/absolute), Mode-A for
  uniform-alpha spans, Mode-B for anti-aliasing borders, and skips fully-transparent pixels.
- Unedited round-trips use the original raw record bytes directly; the only difference from the
  source file is glyph-offset-table ordering for deduplicated images (~1 kB).
- The `accf` table only ever appears as a sidecar `.ccf` file, not embedded in the `.ttf` itself.

image

Fixes #962


PoomSmart:
@behdad Maybe could you help review?

PoomSmart:
Can be merged after https://github.com/fonttools/fonttools/pull/4080

behdad:
What does the `iden` do?

PoomSmart:
> What does the `iden` do?

It's basically a reference type but encoded as glyph name in ASCII rather than glyph ID in hex. @behdad

PoomSmart:
@behdad The PR is merge-ready.

behdad:
It looks fine to me. I can't review the CCF table since I don't know anything about it. Maybe someone else can take a look.

PoomSmart:
@behdad It's alright. CCF is a lesser known format. It only ever appeared on iOS, and iOS doesn't use it anymore. It's a legacy format for iOS 8.3 - 9.3.

Who do you suggest reviewing this?

PoomSmart:
Maybe @anthrotype ?

PoomSmart:
Who else should I tag?

PoomSmart:
Maybe @benkiel or @jenskutilek ?

benkiel:
Sorry, I don't know much about this table.

PoomSmart:
It's going to be very difficult to find anyone that knows CCF table anywhere and anytime. It's Apple's proprietary table. What are the options?

anthrotype:
thanks but...

> It only ever appeared on iOS, and iOS doesn't use it anymore. It's a legacy format for iOS 8.3 - 9.3.

so.. why should fonttools support this proprietary, private, undocumented, legacy format?

> It's going to be very difficult to find anyone that knows CCF table anywhere and anytime.

you've answered yourself.

PoomSmart:
@anthrotype To address your concerns

1. There does not appear to be a formal written policy statement like “we do/do not support proprietary tables” in the main docs.
2. Speaking of proprietary, fonttools does already support many of Apple AAT proprietary tables e.g. `ankr`, `bsln`, `mort`, `morx`, `gcid`, `cidg`, `lcar`, `opbd`, `prop`, `feat`, `trak`, `ltag`.
3. `bgcl` table is supported by fonttools and this table is proprietary, private and undocumented. I was the one who created the PR and it got merged.
4. If this PR was created 9-10 years ago, there would be no argument about supporting legacy tables. Right? And let's say this PR was merged 9-10 years ago, are we planning to deprecate and remove the table support in present days? I don't think that's the policy.
5. I don't see anything that wrong for supporting legacy tables. It's similar to talking about lesser known tables where less than 1% of developers would actually use them. :)

behdad:
I'm okay we commit this. My only hesitation is:

> * The `accf` table only ever appears as a sidecar `.ccf` file, not embedded in the `.ttf` itself.

So, it doesn't belong in a table-reader whatsoever! It would deserve a ccfLib in that case, but given that it's obscurely legacy and not commonly used, I don't think we want to keep it in fontTools. I suggest you make a tiny repo with the reader/writer.

behdad:
We removed a lot of legacy-format support when I took over the project ~2013.

PoomSmart:
> We removed a lot of legacy-format support

Fair, but I'd suggest you to make this explicitly clear so that we didn't have to spend time talking all about this in the first place. :D

> So, it doesn't belong in a table-reader whatsoever! It would deserve a ccfLib in that case

Fair.

PR #4079, opened 2026-04-06
[ttLib] Add support for Apple's 'cntr' (Emoji Centering Data) table

PoomSmart:
## What

Adds support for Apple's proprietary `cntr` table found in AppleColorEmoji font (macOS).

## Background

The `cntr` table payload is a UTF-8 JSON blob decoded on-device by the `EmojiFoundation` framework via `CTFontCopyTable('cntr')` + `JSONDecoder` into the `EmojiCenteringData` Swift type.  Its schema:

```json
{
  "version": 1,
  "size": [160, 160],
  "profiles": {
    "<glyphID>": [[left, bottom, right, top], [dx, dy]],
    …
  }
}
```

According to `EmojiFoundation` framework, the `cntr` table is used for centering the emoji image. However, it is currently unknown which framework on macOS does actually utilize this.


PoomSmart:
Can be merged after https://github.com/fonttools/fonttools/pull/4080

PoomSmart:
@behdad Following https://github.com/fonttools/fonttools/pull/4078, what do you think of adding support to this table. It's proprietary, private, undocumented, simple enough but not legacy?

behdad:
If it's a font table and in active use, I'm supportive.

PR #4083, opened 2026-04-22
[cu2qu] Use _complex_div_by_real in split_cubic_into_three

### LINKED ISSUE #3928: Different number of points in outline in variable font vs if static

anthrotype:
Apply the same fix from PR #3930 to the two remaining `/ 3.0` expressions in `split_cubic_into_three` which I missed in the original `calc_cubic_points` fix.

Cython's C complex division (`__divdc3`) uses multiply-by-reciprocal internally, producing a 1-ULP difference from Python's scalar division. This sometimes causes ±1 off-curve coordinate rounding e.g. on fonts with fractional default-master coordinates like JosefinSans-Italic.

By doing `_complex_div_by_real` we use the same division operator that pure-Python fontTools (and Rust fontc) use

Fixes https://github.com/fonttools/fonttools/issues/3928

### DIFF
--- a/Lib/fontTools/cu2qu/cu2qu.py
+++ b/Lib/fontTools/cu2qu/cu2qu.py
@@ -240,9 +240,9 @@ def split_cubic_into_three(p0, p1, p2, p3):
     mid2 = (p0 + 6 * p1 + 12 * p2 + 8 * p3) * (1 / 27)
     deriv2 = (4 * p3 - 3 * p1 - p0) * (1 / 27)
     return (
-        (p0, (2 * p0 + p1) / 3.0, mid1 - deriv1, mid1),
+        (p0, _complex_div_by_real(2 * p0 + p1, 3.0), mid1 - deriv1, mid1),
         (mid1, mid1 + deriv1, mid2 - deriv2, mid2),
-        (mid2, mid2 + deriv2, (p2 + 2 * p3) / 3.0, p3),
+        (mid2, mid2 + deriv2, _complex_div_by_real(p2 + 2 * p3, 3.0), p3),
     )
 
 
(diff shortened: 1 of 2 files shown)


anthrotype:
only fails on macos cython job (all the pure python ones pass):

https://github.com/fonttools/fonttools/actions/runs/24789313141/job/72542060199?pr=4083#step:5:4078
```
        result = curves_to_quadratic(curves, max_errors=[1.0] * len(curves))
    
>       assert result[0][3][1] == 504.5000000000001
E       assert 504.49999999999994 == 504.5000000000001
```

(unother unrelated `test_deprecated_add_to_variation_store` is failing on Ubuntu py313 but that's pre-existing, it expects exactly 1 warning but gets 2. Worth a separate follow up fix to assert on the warning message/category rather than `len(w)`. Edit: fixed on main https://github.com/fonttools/fonttools/commit/74c193fec73b5a193462046712954325076ce77c)

behdad:
lgtm

PR #4085, opened 2026-05-01
[designspaceLib] Fix map_backward for many-to-one axis maps

### LINKED ISSUE #978: SUSE-Italic GPOS kerning variation region mismatch
leaving here verbatim the output of me asking claude to investigate a diff:

# SUSE-Italic GPOS kerning variation region mismatch

## 0. Source compiled and reproduction command

**Source:** `https://github.com/SUSE/suse-font?7159afb255#sources/SUSE-Italic.glyphs`

```bash
python3 -m ttx_diff 'https://github.com/SUSE/suse-font?7159afb255#sources/SUSE-Italic.glyphs'
```

The font source has a conflict between the master's `Axis Location` custom parameter and the instance-derived axis mapping:

- **Master** "ExtraBold Italic": design=800, `Axis Location` user=800
- **Instance** "ExtraBold Italic": design=780, `Axis Location` user=800

glyphsLib builds the axis map from instances: `user=800 -> design=780`. This means the master at design=800 does NOT correspond to user=800 in the axis map. Its user-space position is ~818.18 (via `designspace.map_backward(800)`).

### HVAR/gvar path (correct)
Built by `fonttools.varLib`, which normalizes source locations from design space directly:
`normalizeValue(800, [100, 100, 1000]) = 0.7778`

### GPOS kerning path (incorrect)

**File:** `ufo2ft/Lib/ufo2ft/featureWriters/kernFeatureWriter.py`, lines 510-511
```python
location = VariableScalarLocation(
    get_userspace_location(designspace, source.location)
)
```

**File:** `ufo2ft/Lib/ufo2ft/util.py`, lines 663-666
```python
def get_userspace_location(designspace, location):
    location_user = designspace.map_backward(location)
    return {designspace.getAxis(k).tag: v for k, v in location_user.items()}
```

This maps the master's design location (800) backward through the axis map to get user=818.18. This user-space location is then passed to `VariableScalar`.

**File:** `fonttools/Lib/fontTools/feaLib/variableScalar.py`, lines 164-178
```python
def _normalize_location(self, location):
    for axtag, value in location:
        axis_min, axis_default, axis_max = self.axis_triples[axtag]
        normalized = normalizeValue(value, (axis_min, axis_default, axis_max))
        mapping = self.axis_mappings.get(axtag)
        if mapping is not None:
            normalized = piecewiseLinearMap(normalized, mapping)
        result[axtag] = normalized
```

This normalizes user=818.18 to 0.7980, then applies the avar mapping: `0.7980 -> 0.8000`. This 0.8 value ends up as the GDEF VarStore region peak.

The axis map inversion is also lossy due to the non-invertible mapping (both user=900 and user=1000 map to design=1000, but the `{d: u for u, d in map}` dict only keeps the last one).

## 2. Root cause

The fontmake pipeline uses **two different normalization paths** for the same master position:

1. **varLib** (for HVAR, gvar): design=800 -> normalize directly -> 0.7778
2. **ufo2ft kern writer + feaLib** (for GPOS kerning): design=800 -> `map_backward` -> user=818.18 -> normalize -> 0.7980 -> avar -> 0.8000

These produce different normalized coordinates (0.7778 vs 0.8) for the same master, resulting in **internally inconsistent VarStore regions** within the same font. The GDEF VarStore (used by GPOS kerning) has different region peaks than the HVAR VarStore, even though they describe the same variation space.

The root cause is triggered by a conflict in the font source: the master's `Axis Location` says user=800 -> design=800, but the instance-derived axis map says user=800 -> design=780. The `map_backward()` inversion of this conflicting map produces user=818.18 for design=800, which after normalization and avar mapping gives 0.8 instead of 0.7778.

## 3. Which side is wrong

**fontmake is wrong.** fontc is correct.

fontmake produces an internally inconsistent font: the GDEF VarStore (backing GPOS kerning) uses region peak 0.8 while the HVAR VarStore and gvar use region peak 0.7778 for the same master. All variation data in a font should use the same coordinate system. At runtime, the rasterizer computes a single set of coordinates and applies them uniformly to all variation data; having different regions means kerning deltas are evaluated at a different point than glyph outline and advance width deltas, producing incorrect kerning at non-master positions.

fontc consistently uses 0.7778 (the design-to-normalized value) for all VarStores, which is correct and matches what varLib produces for HVAR/gvar.

## 4. Suggested fix

This is a **fontmake/ufo2ft bug**, not a fontc bug. No fix is needed in fontc.

The fix in ufo2ft would be to avoid the design -> user -> normalize -> avar round-trip in `kernFeatureWriter.py`. Instead of:
```python
location = VariableScalarLocation(
    get_userspace_location(designspace, source.location)
)
```

The kern feature writer should either:
1. Use design-space locations and normalize them the same way varLib does (matching the HVAR/gvar path), or
2. Use the source's actual user-space location from the master's `Axis Location` parameter instead of computing it via `map_backward()`.

Alternatively, the fix could be in `fonttools/feaLib/variableScalar.py`'s `VariableScalarBuilder` to accept design-space locations directly and normalize them without the avar round-trip.

anthrotype:
`AxisDescriptor.map_backward()` used a dict comprehension (`{d: u for u, d in map}`) to invert the axis map, which silently dropped entries when multiple user-space values mapped to the same design-space value (a "flat segment" is valid per the [avar spec](https://learn.microsoft.com/en-us/typography/opentype/spec/avar#table-formats), which allows `toCoordinate` to be "greater than or equal to" the preceding one).

This produced wrong interpolation for nearby design values, leading to incorrect normalized coordinates for any code relying on `map_backward` (e.g. ufo2ft feature writers when remapping design -> user coordinates for VariableScalars). 

Replace with a sorted list of `(design, user)` pairs that preserves both endpoints of flat segments, so the inverse interpolation is correct on both sides of the flat region.

Fixes https://github.com/googlefonts/ufo2ft/issues/978

### DIFF
--- a/Lib/fontTools/designspaceLib/__init__.py
+++ b/Lib/fontTools/designspaceLib/__init__.py
@@ -1041,13 +1041,25 @@ class AxisDescriptor(AbstractAxisDescriptor):
 
     def map_backward(self, v):
         """Maps value from axis mapping's output (design) to input (user)."""
-        from fontTools.varLib.models import piecewiseLinearMap
-
         if isinstance(v, tuple):
             v = v[0]
         if not self.map:
             return v
-        return piecewiseLinearMap(v, {v: k for k, v in self.map})
+        # Build (design, user) pairs sorted by design then user, keeping
+        # both endpoints of any many-to-one (flat) segments so we can
+        # invert them and interpolation is correct on both sides.
+        # https://github.com/googlefonts/ufo2ft/issues/978
+        backward = sorted((design, user) for user, design in self.map)
+        design0, user0 = backward[0]
+        if v <= design0:
+            return v + user0 - design0
+        for (design1, user1), (design2, user2) in zip(backward, backward[1:]):
+            if design1 <= v <= design2:
+                if design1 == design2:
+                    return user1
+                return user1 + (user2 - user1) * (v - design1) / (design2 - design1)
+        designN, userN = backward[-1]
+        return v + userN - designN
 
 
 class DiscreteAxisDescriptor(AbstractAxisDescriptor):
--- a/Tests/designspaceLib/designspace_test.py
+++ b/Tests/designspaceLib/designspace_test.py
@@ -1206,3 +1206,29 @@ def test_preserveEmptyConditionSet():
     roundtripped_ds = DesignSpaceDocument.fromstring(ds_str)
     assert len(roundtripped_ds.rules) == 1
     assert len(roundtripped_ds.rules[0].conditionSets) == 1
+
+
+def test_map_backward_many_to_one():
+    """Test that map_backward handles many-to-one (flat segment) axis maps.
+
+    Regression test for https://github.com/googlefonts/ufo2ft/issues/978
+    """
+    ax = AxisDescriptor()
+    ax.name = "Weight"
+    ax.tag = "wght"
+    ax.minimum = 100
+    ax.default = 100
+    ax.maximum = 1000
+    # Many-to-one: both user=900 and user=1000 map to design=1000
+    ax.map = [
+        (100, 100),
+        (800, 780),
+        (900, 1000),
+        (1000, 1000),
+    ]
+
+    # map_backward(800) should interpolate between (780, 800) and (1000, 900),
+    # NOT between (780, 800) and (1000, 1000).
+    result = ax.map_backward(800)
+    expected = 800 + (800 - 780) / (1000 - 780) * (900 - 800)  # ~809.09
+    assert result == pytest.approx(expected)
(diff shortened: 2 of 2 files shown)


anthrotype:
Repro here:

https://github.com/fonttools/fonttools/actions/runs/25217674832/job/73941813830?pr=4085#step:5:4125

```
>       assert result == pytest.approx(expected)
E       assert 818.1818181818181 == 809.0909090909091 ± 8.1e-04
E         
E         comparison failed
E         Obtained: 818.1818181818181
E         Expected: 809.0909090909091 ± 8.1e-04
```

cmyr (approved):
looks good!

PR #4092, opened 2026-05-20
[feaLib] Fix name-table parsing for multibyte Mac encodings (#1196)

### LINKED ISSUE #1196: [feaLib] feaLib may fail to parse name table's entry encoded by x_mac_japanese_ttx
It is not a serious problem but I would like to share a bug.
A name table's entry encoded by x_mac_japanese_ttx caused UnicodeDecodeError as below:
```
UnicodeDecodeError: 'x_mac_japanese_ttx' codec can't decode byte 0x8f in position 0: incomplete multibyte sequence
```
by a snippet like:
```python
parser = Parser(sys.argv[1], [])
doc = parser.parse()
for statement in doc.statements:
    if isinstance(statement, fontTools.feaLib.ast.TableBlock):
        if hasattr(statement, "statements"):
            for statement2 in statement.statements:
                print(statement2.string)
```

A sample of such a name table can be got from page 8 of:
https://www.adobe.com/content/dam/acom/en/devnet/font/pdfs/5149.OTFname_Tutorial.pdf
i.e.
```
table name {
  nameid 9 1 1 11 "Masahiko Kozuka \8f\ac\92\cb\8f\b9\95F";
} name;
```

miguelsousa:
The `unescape_string_` helper used to decode each `\XX` escape as a single byte, which broke multibyte encodings (`x_mac_japanese_ttx` and friends) whenever one character's bytes were split across escapes or across an escape/literal-ASCII boundary (e.g. `\95F` = Shift-JIS `0x95` `0x46` = 彦).

Build the full byte sequence from literal chars + `\XX` escapes and decode it in one pass.

Fixes #1196

### DIFF
--- a/Lib/fontTools/feaLib/parser.py
+++ b/Lib/fontTools/feaLib/parser.py
@@ -2,7 +2,7 @@ from fontTools.feaLib.error import FeatureLibError
 from fontTools.feaLib.lexer import Lexer, IncludingLexer, NonIncludingLexer
 from fontTools.feaLib.variableScalar import VariableScalar
 from fontTools.misc.encodingTools import getEncoding
-from fontTools.misc.textTools import bytechr, tobytes, tostr
+from fontTools.misc.textTools import tobytes, tostr
 import fontTools.feaLib.ast as ast
 import logging
 import os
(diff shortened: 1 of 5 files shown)


Lib/fontTools/feaLib/parser.py
```diff
@@ -1273,8 +1273,18 @@ def unescape_string_(self, string, encoding):
...
+            # multibyte encodings (e.g. x_mac_japanese_ttx) where bytes
+            # belonging to a single character may straddle literal/escape
+            # boundaries — e.g. "\95F" is one Shift-JIS char (0x95 0x46).
+            out = bytearray()
+            for token in re.split(r"(\\[0-9a-fA-F]{2})", string):
+                if len(token) == 3 and token[0] == "\\":
+                    out.append(int(token[1:], 16))
+                elif token:
+                    out.extend(token.encode(encoding))
+            s = bytes(out).decode(encoding)
```
justvanrossum: `bytesarray` has a `decode` method, so there should be no need to convert to `bytes`

justvanrossum (approved):

Lib/fontTools/feaLib/parser.py
miguelsousa: thanks Just. Updated in 58cf4c8e465706f1ee6290f4809f69ac79bc3812

PR #4096, opened 2026-06-03
[subset] bug fix for MATH table

qxliu76:
No need to keep constructions for glyphs that are only added during math closure.
See also: https://github.com/harfbuzz/harfbuzz/issues/5998


anthrotype:
Thanks Ginger! I added a regression test here https://github.com/fonttools/fonttools/commit/0dbc9eab01ccc6f285350c82e2e35a4f2e52919b

qxliu76:
> Thanks Ginger! I added a regression test here [0dbc9ea](https://github.com/fonttools/fonttools/commit/0dbc9eab01ccc6f285350c82e2e35a4f2e52919b)

Thank you! Sorry I missed to add a test for this change.

PR #4097, opened 2026-06-05
Implement beyond-64k from the proposed spec changes

behdad:
This PR adds initial fontTools support for proposed beyond-64k companion tables.

  - Adds beyond-64k uppercase companion table support.
  - Adds ttLib.beyond64k upper/lower APIs and CLI.
  - Implements Layout/core companion schemas and table classes.
  - Updates FontBuilder and merger to create/use uppercase companions.
  - Updates tools/helpers: qu2cu, removeOverlaps, scaleUpem, svgPathPen.
  - Updates varLib helpers/build paths for GVAR/GLYF/HMTX.
  - Confirms uppercase companions round-trip through WOFF2 as generic tables.
  - Adds broad tests for conversion, build, merge, round-trip, and rendering paths.
  - Defers subset, instancer, DMAP, VARC subsetting, and best-effort lowering.

Supersedes https://github.com/fonttools/fonttools/pull/2678

HarfBuzz counterpart: https://github.com/harfbuzz/harfbuzz/pull/5655


behdad:
So there was one spec divergence I needed to make, to make merging NotoSans fonts work: spec doesn't widen LookupList-to-Lookup offsets in the extended LookupList2 field and keeps them 16bit, which defeats the whole purpose. So I widened them to 32bit and merge succeeds quickly with the fontTools.merge tool.

Here's a merged NotoSans font with all scripts except for emoji & CJK:

https://drive.google.com/file/d/13pK4xKRheLEuUhQB2xvkS6pqZSSB7_iO/view

```
Unicode count: 26984
Glyph count: 106958
```

```
Table information:

Tag	Size
------------
GDEF    57051 bytes
GLYF 13001268 bytes
GPOS  1587490 bytes
GSUB  1660634 bytes
HHEA       38 bytes
HMTX   427830 bytes
LOCA   427836 bytes
MAXP       33 bytes
OS/2       96 bytes
cmap   115950 bytes
cvt       300 bytes
fpgm     3605 bytes
gasp        8 bytes
head       54 bytes
name     2568 bytes
post       32 bytes
prep      261 bytes
```

Doc/source/ttLib/tables/G_V_A_R_.rst
```diff
@@ -1,9 +1,9 @@
 ``GVAR``: 24-bit Glyph Variation table
 --------------------------------------
 
-The ``GVAR`` table is an Apple-extended variant of the ``gvar`` table that
-uses 24-bit glyph IDs instead of 16-bit ones, allowing it to support fonts
-with more than 65,535 glyphs.
+The ``GVAR`` table is the beyond-64k companion to the ``gvar`` table. It uses
```
khaledhosny, line 4: Do we have a more official and user-friendly name to these extension tables other than “beyond-64k”?

behdad: I'm afraid not, unless we make up something better.

davelab6: Once the mpeg open font format 5th edition is out, anyone can propose any copy edit to become the 5th edition first amendment. 

But I'm not aware of a better name for this part so far

simoncozens:
LGTM. Nice and clean!

simoncozens (approved):

Doc/source/ttLib/tables/G_V_A_R_.rst
khaledhosny: I guess it is as a good name as any.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -256,6 +256,21 @@
             ),
         ],
     ),
+    (
+        "SeqLookup",
+        [
+            FieldSpec(
+                "uint16",
+                "SequenceIndex",
+                description="Index into current glyph sequence-first glyph = 0",
```
khaledhosny, line 284: Is it intentional that these descriptions don’t use the exact spec text?

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -343,6 +404,47 @@
...
+            FieldSpec(
+                "GlyphID24",
+                "StartGlyph",
+                description="First GlyphID of the ClassValueArray",
+            ),
+            FieldSpec(
+                "uint24", "GlyphCount", description="Size of the ClassValueArray"
+            ),
+            FieldSpec(
+                "uint16",
```
khaledhosny, line 441: The spec has `uint24 | classValues[glyphCount]`

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1004,6 +1153,56 @@
             ),
         ],
     ),
+    (
+        "ContextPosFormat4",
```
khaledhosny, line 1157: I’m having trouble finding this in the spec, I only found SequenceContextFormat4, is it the same thing?

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -2012,6 +2829,173 @@
...
+    (
+        "ChainedSeqRuleSet2",
+        [
+            FieldSpec(
+                "uint16",
+                "ChainedSeqRuleCount",
+                description="Number of ChainedSeqRule2 tables",
+            ),
+            FieldSpec(
+                "OffsetTo(ChainedSeqRule2)",
```
khaledhosny, line 2866: The spec has `Offset24 | chainedSeqRuleOffsets[chainedSeqRuleCount]` here.

Lib/fontTools/ttLib/tables/otData.py
behdad: Yes. Old spec had diff SubstContext / PosContext. Peter unified them under SequenceContext. I kept dropped the Sequence because that's how we call them in HB, but can change.

behdad: Err sorry. I might have misspoken. Lemme check later and get back to you.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1356,6 +1561,201 @@
...
+    (
+        "PairPosFormat3",
+        [
+            FieldSpec("uint16", "PosFormat", description="Format identifier-format = 3"),
+            FieldSpec("LOffset", "Coverage"),
+            FieldSpec("uint16", "ValueFormat1"),
+            FieldSpec("uint16", "ValueFormat2"),
+            FieldSpec("uint24", "PairSetCount"),
+            FieldSpec(
+                "LOffsetTo(PairSet2)", "PairSet", repeat="PairSetCount", aux=0
```
khaledhosny: The spec has `Offset24 | pairSetOffsets[pairSetCount]` here.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1356,6 +1561,201 @@
...
+        [
+            FieldSpec("uint16", "PosFormat", description="Format identifier-format = 4"),
+            FieldSpec("LOffset", "Coverage"),
+            FieldSpec("uint16", "ValueFormat1"),
+            FieldSpec("uint16", "ValueFormat2"),
+            FieldSpec("LOffset", "ClassDef1"),
+            FieldSpec("LOffset", "ClassDef2"),
+            FieldSpec("uint16", "Class1Count"),
+            FieldSpec("uint16", "Class2Count"),
+            FieldSpec("Class1Record", "Class1Record", repeat="Class1Count", aux=0),
```
khaledhosny, line 1657: `Class2Record` is missing?

khaledhosny: `PairPosFormat2` does not have `Class2Record`, so I guess this is handled in some special way.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1356,6 +1561,201 @@
...
+            FieldSpec("LOffset", "Coverage"),
+            FieldSpec("uint24", "EntryExitCount"),
+            FieldSpec("EntryExit2", "EntryExitRecord", repeat="EntryExitCount", aux=0),
+        ],
+    ),
+    (
+        "EntryExit2",
+        [
+            FieldSpec("LOffsetTo(Anchor)", "EntryAnchor"),
+            FieldSpec("LOffsetTo(Anchor)", "ExitAnchor"),
```
khaledhosny: The spec has `Offset24` for both of these.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1356,6 +1561,201 @@
...
+        [
+            FieldSpec("uint24", "BaseCount"),
+            FieldSpec("BaseRecord2", "BaseRecord", repeat="BaseCount", aux=0),
+        ],
+    ),
+    (
+        "BaseRecord2",
+        [
+            FieldSpec(
+                "LOffsetTo(Anchor)",
```
khaledhosny: The spec has `Offset24 | baseAnchorOffset[markClassCount]` here.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1356,6 +1561,201 @@
...
+            FieldSpec("LOffsetTo(MarkArray2)", "MarkArray"),
+            FieldSpec("LOffsetTo(LigatureArray2)", "LigatureArray"),
+        ],
+    ),
+    (
+        "LigatureArray2",
+        [
+            FieldSpec("uint24", "LigatureCount"),
+            FieldSpec(
+                "LOffsetTo(LigatureAttach2)",
```
khaledhosny: The spec has `Offset24 | ligatureAttachOffsets[ligatureCount]` here.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1356,6 +1561,201 @@
...
+                repeat="ComponentCount",
+                aux=0,
+            ),
+        ],
+    ),
+    (
+        "ComponentRecord2",
+        [
+            FieldSpec(
+                "LOffsetTo(Anchor)",
```
khaledhosny: The spec has `Offset24 | ligatureAnchorOffsets[markClassCount]` here.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1356,6 +1561,201 @@
...
+        [
+            FieldSpec("uint16", "Mark2Count"),
+            FieldSpec("Mark2Record2", "Mark2Record", repeat="Mark2Count", aux=0),
+        ],
+    ),
+    (
+        "Mark2Record2",
+        [
+            FieldSpec(
+                "LOffsetTo(Anchor)",
```
khaledhosny: The spec has `Offset24 | mark2AnchorOffsets[markClassCount]` here.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1356,6 +1561,201 @@
...
+        [
+            FieldSpec("uint24", "MarkCount"),
+            FieldSpec("MarkRecord2", "MarkRecord", repeat="MarkCount", aux=0),
+        ],
+    ),
+    (
+        "MarkRecord2",
+        [
+            FieldSpec("uint16", "Class"),
+            FieldSpec("LOffsetTo(Anchor)", "MarkAnchor"),
```
khaledhosny: The spec has `Offset24 | markAnchorOffset` here.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1475,6 +1936,48 @@
...
+                "Coverage",
+                description="Offset to Coverage table-from beginning of Substitution table",
+            ),
+            FieldSpec(
+                "uint24",
+                "SequenceCount",
+                description="Number of Sequence2 table offsets in the Sequence array",
+            ),
+            FieldSpec(
+                "LOffsetTo(Sequence2)",
```
khaledhosny: The spec has `Offset24 | sequenceOffsets[sequenceCount]` here,

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1517,6 +2020,48 @@
...
+                "Coverage",
+                description="Offset to Coverage table-from beginning of Substitution table",
+            ),
+            FieldSpec(
+                "uint24",
+                "AlternateSetCount",
+                description="Number of AlternateSet2 tables",
+            ),
+            FieldSpec(
+                "LOffsetTo(AlternateSet2)",
```
khaledhosny: The spec has `Offset24 | alternateSetOffsets[alternateSetCount]` here.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1575,6 +2120,64 @@
...
+            FieldSpec(
+                "LOffset",
+                "Coverage",
+                description="Offset to Coverage table-from beginning of Substitution table",
+            ),
+            FieldSpec(
+                "uint24", "LigSetCount", description="Number of LigatureSet2 tables"
+            ),
+            FieldSpec(
+                "LOffsetTo(LigatureSet2)",
```
khaledhosny: The spec has `Offset24 | ligatureSetOffsets[ligatureSetCount]` here.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1575,6 +2120,64 @@
...
+        ],
+    ),
+    (
+        "LigatureSet2",
+        [
+            FieldSpec(
+                "uint16", "LigatureCount", description="Number of Ligature2 tables"
+            ),
+            FieldSpec(
+                "LOffsetTo(Ligature2)",
```
khaledhosny: The spec has `Offset24 | ligatureOffsets[ligatureCount]` here.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1723,37 +2326,208 @@
...
+                "Coverage",
+                description="Offset to Coverage table-from beginning of Substitution table",
+            ),
+            FieldSpec(
+                "uint24",
+                "SeqRuleSetCount",
+                description="Number of SeqRuleSet2 tables",
+            ),
+            FieldSpec(
+                "LOffsetTo(SeqRuleSet2)",
```
khaledhosny: The spec has `Offset24 | seqRuleSetOffsets[seqRuleSetCount]` here.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1004,6 +1134,50 @@
...
         ],
     ),
+    (
+        "ContextPosFormat4",
+        [
+            FieldSpec("uint16", "PosFormat", description="Format identifier-format = 4"),
+            FieldSpec("LOffset", "Coverage"),
+            FieldSpec("uint24", "SeqRuleSetCount"),
+            FieldSpec(
+                "LOffsetTo(SeqRuleSet2)",
```
khaledhosny: The spec has `Offset24 | seqRuleSetOffsets[seqRuleSetCount]` here.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1723,37 +2326,208 @@
...
+                "ClassDef",
+                description="Offset to glyph ClassDef table-from beginning of Substitution table",
+            ),
+            FieldSpec(
+                "uint24",
+                "ClassSeqRuleSetCount",
+                description="Number of ClassSeqRuleSet2 tables",
+            ),
+            FieldSpec(
+                "LOffsetTo(ClassSeqRuleSet2)",
```
khaledhosny: The spec has `Offset24 | classSeqRuleSetOffsets[classSeqRuleSetCount]` here

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1004,6 +1134,50 @@
...
+    ),
+    (
+        "ContextPosFormat5",
+        [
+            FieldSpec("uint16", "PosFormat", description="Format identifier-format = 5"),
+            FieldSpec("LOffset", "Coverage"),
+            FieldSpec("LOffset", "ClassDef"),
+            FieldSpec("uint24", "ClassSeqRuleSetCount"),
+            FieldSpec(
+                "LOffsetTo(ClassSeqRuleSet2)",
```
khaledhosny: The spec has `Offset24 | classSeqRuleSetOffsets[classSeqRuleSetCount]` here

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -1723,37 +2326,208 @@
...
+    (
+        "ClassSeqRuleSet2",
+        [
+            FieldSpec(
+                "uint16",
+                "ClassSeqRuleCount",
+                description="Number of ClassSeqRule tables",
+            ),
+            FieldSpec(
+                "LOffsetTo(ClassSeqRule)",
```
khaledhosny: The spec has `Offset24 | classSeqRuleOffsets[classSeqRuleCount]` here.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -2012,6 +2786,173 @@
             ),
         ],
     ),
+    (
+        "ChainContextSubstFormat4",
```
khaledhosny, line 2833: FWIW, the spec calls this `ChainedSequenceContextFormat4`

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -2012,6 +2786,173 @@
...
+                "Coverage",
+                description="Offset to Coverage table-from beginning of Substitution table",
+            ),
+            FieldSpec(
+                "uint24",
+                "ChainedSeqRuleSetCount",
+                description="Number of ChainedSeqRuleSet2 tables",
+            ),
+            FieldSpec(
+                "LOffsetTo(ChainedSeqRuleSet2)",
```
khaledhosny: The spec has `Offset24 | chainedSeqRuleSetOffsets[chainedSeqRuleSetCount]` here.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -2012,6 +2786,173 @@
...
+    (
+        "ChainedSeqRuleSet2",
+        [
+            FieldSpec(
+                "uint16",
+                "ChainedSeqRuleCount",
+                description="Number of ChainedSeqRule2 tables",
+            ),
+            FieldSpec(
+                "LOffsetTo(ChainedSeqRule2)",
```
khaledhosny: The spec has `Offset24 | chainedSeqRuleOffsets[chainedSeqRuleCount]`

khaledhosny:
Some of my comments are probably on outdated code, sorry for that. I tried to review the full diff but it was unusable with GitHub UI, so I had to review individual commits, so I might be reviewing code that a later commit fixed.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -2012,6 +2786,173 @@
...
+                "SeqLookup",
+                "SeqLookupRecord",
+                repeat="SeqLookupCount",
+                aux=0,
+                description="Array of SeqLookup records-in design order",
+            ),
+        ],
+    ),
+    (
+        "ChainContextSubstFormat5",
```
khaledhosny, line 2928: FWIW, the spec apparently calls this `ChainedSequenceContextFormat5`.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -2012,6 +2786,173 @@
...
+            FieldSpec("LOffset", "BacktrackClassDef"),
+            FieldSpec("LOffset", "InputClassDef"),
+            FieldSpec("LOffset", "LookAheadClassDef"),
+            FieldSpec(
+                "uint16",
+                "ChainedClassSeqRuleSetCount",
+                description="Number of ChainedClassSeqRuleSet2 tables",
+            ),
+            FieldSpec(
+                "LOffsetTo(ChainedClassSeqRuleSet2)",
```
khaledhosny: The spec has `Offset24 | chainedClassSeqRuleSetOffsets[chainedClassSeqRuleSetCount]` here.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -2012,6 +2786,173 @@
...
+    (
+        "ChainedClassSeqRuleSet2",
+        [
+            FieldSpec(
+                "uint16",
+                "ChainedClassSeqRuleCount",
+                description="Number of ChainedClassSeqRule tables",
+            ),
+            FieldSpec(
+                "LOffsetTo(ChainedClassSeqRule)",
```
khaledhosny: The spec has `Offset24 | chainedSeqRuleOffsets[chainedSeqRuleCount]` here.

Lib/fontTools/ttLib/tables/otData.py
```diff
@@ -2184,6 +3207,26 @@
...
+            FieldSpec(
+                "LOffset",
+                "Coverage",
+                description="Offset to Coverage table - from beginning of LigCaretList2 table",
+            ),
+            FieldSpec(
+                "uint24", "LigGlyphCount", description="Number of ligature glyphs"
+            ),
+            FieldSpec(
+                "LOffsetTo(LigGlyph)",
```
khaledhosny, line 3265: The spec has `Offset24 | ligGlyphOffsets[ligGlyphCount]` here.

Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -1187,17 +1187,18 @@ def populateDefaults(self, propagator=None):
...
+            outGIDS = [(glyphID + delta) % modulus for glyphID in inputGIDS]
             outNames = font.getGlyphNameMany(outGIDS)
             for inp, out in zip(input, outNames):
                 mapping[inp] = out
-        elif self.Format == 2:
+        elif self.Format in (2, 4):
             assert (
                 len(input) == rawTable["GlyphCount"]
-            ), "invalid SingleSubstFormat2 table"
+            ), "invalid SingleSubst table"
```
khaledhosny: What about `f"invalid SingleSubstFormat{self.Format} table"`?

Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -1238,8 +1245,10 @@ def preWrite(self, font):
         subst = [item[1][1] for item in sortableItems]
         cov.glyphs = input
         rawTable["Coverage"] = cov
-        if format == 1:
+        if format in (1, 3):
             assert delta is not None
+            if format == 3 and delta >= 0x800000:
+                delta -= 0x1000000
```
khaledhosny, line 1263: I’m probably missing something, but what is this block doing?

Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -1614,10 +1625,17 @@ def preWrite(self, font):
...
             for lig in set:
-                ligs.append(lig)
+                ligatureClass = Ligature2 if extended else Ligature
+                if isinstance(lig, ligatureClass):
+                    out = lig
+                else:
+                    out = ligatureClass()
+                    out.LigGlyph = lig.LigGlyph
+                    out.Component = lig.Component
+                ligs.append(out)
```
khaledhosny, line 1659: A comment here that this is converting between `Ligature2`/`Ligature` depending on the format might be nice. It wasn’t obvious the first time I read it.

Lib/fontTools/otlLib/builder.py
```diff
@@ -2645,14 +2649,15 @@ def buildSinglePosSubtable(values, glyphMap):
         ValueRecord(src=values[g], valueFormat=valueFormat)
         for g in self.Coverage.glyphs
     ]
+    formatOffset = 2 if _glyphMapHasExtendedGlyphIDs(glyphMap) else 0
     if all(v == valueRecords[0] for v in valueRecords):
-        self.Format = 1
+        self.Format = 1 + formatOffset
```
khaledhosny, line 2722: This is clever, but I think the `self.Format = 3 if extended else 1` used elsewhere is more readable.

Lib/fontTools/otlLib/builder.py
```diff
@@ -1274,17 +1274,18 @@ def build(self):
...
             st = ot.MarkMarkPos()
-            st.Format = 1
+            extended = _glyphMapHasExtendedGlyphIDs(self.glyphMap)
+            st.Format = 2 if extended else 1
             st.ClassCount = len(markClasses)
             st.Mark1Coverage = buildCoverage(marks, self.glyphMap)
             st.Mark2Coverage = buildCoverage(subtable[1], self.glyphMap)
             st.Mark1Array = buildMarkArray(marks, self.glyphMap)
-            st.Mark2Array = ot.Mark2Array()
+            st.Mark2Array = ot.Mark2Array2() if extended else ot.Mark2Array()
```
khaledhosny, line 1345: I’m wondering if there is some clever way to keep using the none-v2 classes etc elsewhere and handle this selection at compile time. The duplication looks rather ugly and too much code churn (and I worry also about downstream code that would need to adapt to this), but I have no alternative to suggest.

Lib/fontTools/feaLib/builder.py
```diff
@@ -220,10 +220,19 @@ def build(self, tables=None, debug=False):
             table = self.makeTable(tag)
             if self.feature_variations_:
                 self.makeFeatureVariations(table, tag)
+            if self.font.hasExtendedGlyphIDs():
+                self.promoteLayoutTable_(table)
```
khaledhosny, line 224: I’d move the check inside `promoteLayoutTable_()`, and make it return the three *Lists, or inline `promoteLayoutTable_()`. I don’t like how this work is split over two places.

Lib/fontTools/feaLib/builder.py
```diff
@@ -817,21 +817,35 @@ def buildGDEF(self):
...
             (
                 gdef.GlyphClassDef,
                 gdef.AttachList,
                 gdef.LigCaretList,
                 gdef.MarkAttachClassDef,
                 gdef.MarkGlyphSetsDef,
             )
-        ) or hasattr(gdef, "VarStore"):
+        ) or hasattr(gdef, "VarStore")
+        if self.font.hasExtendedGlyphIDs():
```
khaledhosny, line 829: Wouldn’t removing the `if` check above make it try to promote the table if it has no data?

Lib/fontTools/ttLib/tables/_m_a_x_p.py
khaledhosny, line 5: This is now unused.

Lib/fontTools/ttLib/tables/_m_a_x_p.py
```diff
@@ -36,24 +36,31 @@ class table__m_a_x_p(DefaultTable.DefaultTable):
...
+    glyphTableTag = "glyf"
+    metricsTableTag = "hmtx"
+    numGlyphsSize = 2
 
     def decompile(self, data, ttFont):
-        dummy, data = sstruct.unpack2(maxpFormat_0_5, data, self)
+        headerSize = 4 + self.numGlyphsSize
+        self.tableVersion = int.from_bytes(data[:4], "big")
+        self.numGlyphs = int.from_bytes(data[4:headerSize], "big")
+        data = data[headerSize:]
```
khaledhosny, line 47: Maybe `maxpFormat_0_5` can be a class member (instead of `numGlyphsSize`) and the old code wouldn’t need to change much?

Lib/fontTools/ttLib/tables/H_H_E_A_.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class table_H_H_E_A_(table__h_h_e_a):
+    dependencies = ["HMTX", "GLYF"]
+    tableFormat = hheaFormat.replace(
+        "numberOfHMetrics:       H", "numberOfHMetrics:       L"
+    )
+    glyphTableTag = "GLYF"
+    metricsTableTag = "HMTX"
+    outlineTags = ("GLYF",)
```
khaledhosny, line 11: Does’t `CFF2` table support 32-bit GIDs, or using it with `HHEA` is not supported?

Lib/fontTools/ttLib/tables/V_H_E_A_.py
```diff
@@ -0,0 +1,9 @@
+from ._v_h_e_a import table__v_h_e_a, vheaFormat
+
+
+class table_V_H_E_A_(table__v_h_e_a):
+    dependencies = ["VMTX", "GLYF"]
+    tableFormat = vheaFormat.replace("numberOfVMetrics:\tH", "numberOfVMetrics:\tL")
+    glyphTableTag = "GLYF"
+    metricsTableTag = "VMTX"
+    outlineTags = ("GLYF",)
```
khaledhosny, line 9: Same question about `CFF2`.

Lib/fontTools/ttLib/tables/H_H_E_A_.py
```diff
@@ -0,0 +1,11 @@
+from ._h_h_e_a import hheaFormat, table__h_h_e_a
+
+
+class table_H_H_E_A_(table__h_h_e_a):
+    dependencies = ["HMTX", "GLYF"]
+    tableFormat = hheaFormat.replace(
+        "numberOfHMetrics:       H", "numberOfHMetrics:       L"
```
khaledhosny, line 7: in `VHEA` `\t` is used. Either way is fine, but it needs to be consistent.

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -1325,12 +1331,12 @@ def getGlyphSet(
...
+        ``GLYF`` or ``glyf`` table.
 
-        If the font contains both a ``CFF ``/``CFF2`` and a ``glyf`` table, you
-        can use the ``preferCFF`` argument to specify which one should be taken.
-        If the font contains both a ``CFF `` and a ``CFF2`` table, the latter is
-        taken.
+        If the font contains both a ``CFF ``/``CFF2`` and a ``GLYF``/``glyf``
+        table, you can use the ``preferCFF`` argument to specify which one
+        should be taken. If the font contains both a ``CFF `` and a ``CFF2``
+        table, the latter is taken.
```
khaledhosny, line 1339: It should also say which wins when both `glyf` and `GLYF` are present.

Lib/fontTools/ttLib/ttFont.py
```diff
@@ -1346,10 +1352,14 @@ def getGlyphSet(
...
+        glyfTag = "GLYF" if "GLYF" in self else "glyf"
+        hasGlyf = glyfTag in self
+        if ("CFF " in self or "CFF2" in self) and (preferCFF or not hasGlyf):
             glyphSet = _TTGlyphSetCFF(self, location)
-        elif "glyf" in self:
-            glyphSet = _TTGlyphSetGlyf(self, location, recalcBounds=recalcBounds)
+        elif hasGlyf:
+            glyphSet = _TTGlyphSetGlyf(
+                self, location, glyfTag, recalcBounds=recalcBounds
+            )
```
khaledhosny, line 1362: `_TTGlyphSetCFF` does the `CFF`/`CFF2` selection itself, maybe `_TTGlyphSetGlyf` should do similarly.

Lib/fontTools/ttLib/beyond64k.py
```diff
@@ -0,0 +1,202 @@
...
+_TABLE_PAIRS = {
+    "glyf": "GLYF",
+    "loca": "LOCA",
+    "maxp": "MAXP",
+    "hhea": "HHEA",
+    "hmtx": "HMTX",
+    "vhea": "VHEA",
+    "vmtx": "VMTX",
+    "gvar": "GVAR",
+}
```
khaledhosny, line 48: Shouldn’t it list one set, and uppercase/lowercase to get the other?

Lib/fontTools/ttLib/beyond64k.py
```diff
@@ -0,0 +1,202 @@
...
+
+def _validate_lowering(font: TTFont, conversions: dict[str, str]) -> None:
+    if "MAXP" in conversions and "MAXP" in font and font["MAXP"].numGlyphs > 0xFFFF:
+        raise ValueError("MAXP.numGlyphs does not fit in maxp")
+    if "HHEA" in conversions and "HHEA" in font:
+        if font["HHEA"].numberOfHMetrics > 0xFFFF:
+            raise ValueError("HHEA.numberOfHMetrics does not fit in hhea")
+    if "VHEA" in conversions and "VHEA" in font:
+        if font["VHEA"].numberOfVMetrics > 0xFFFF:
+            raise ValueError("VHEA.numberOfVMetrics does not fit in vhea")
```
khaledhosny, line 517: Shouldn’t this also check other tables?

Lib/fontTools/ttLib/beyond64k.py
```diff
@@ -33,6 +35,23 @@ class _TableConversion:
...
+    otTables.AlternateSubst,
+    otTables.LigatureSubst,
+)
+
+
+def _force_layout_formats(table, extended):
+    for path in dfs_base_table(table):
+        subtable = path[-1].value
+        if isinstance(subtable, _AUTO_FORMAT_TABLES):
+            subtable._forceExtended = extended
```
khaledhosny: Do we need this to be per subtable, wouldn’t a global table attribute make more sense?

Lib/fontTools/merge/__init__.py
```diff
@@ -79,6 +97,9 @@ def merge(self, fontfiles):
         fonts = self._openFonts(fontfiles)
         glyphOrders = [list(font.getGlyphOrder()) for font in fonts]
         computeMegaGlyphOrder(self, glyphOrders)
+        self.beyond64k = len(self.glyphOrder) > 0x10000 or any(
```
khaledhosny: Other places use `hasExtendedGlyphIDs`, so maybe use something similar here for consistency.

Lib/fontTools/fontBuilder.py
```diff
@@ -310,7 +310,9 @@ def _getOS2Defaults():
 
 
 class FontBuilder(object):
-    def __init__(self, unitsPerEm=None, font=None, isTTF=True, glyphDataFormat=0):
+    def __init__(
+        self, unitsPerEm=None, font=None, isTTF=True, glyphDataFormat=0, beyond64k=False
```
khaledhosny: Maybe `extendedGlyphIDs` instead of `beyond64k`.

Lib/fontTools/fontBuilder.py
```diff
@@ -368,6 +387,34 @@ def _updateTableWithValues(self, tableTag, values):
         for k, v in values.items():
             setattr(table, k, v)
 
+    def _tableTag(self, lower, upper):
+        return upper if self.beyond64k else lower
+
+    def _glyfTableTag(self):
+        return self._tableTag("glyf", "GLYF")
```
khaledhosny, line 384: What about passing a single tag and letting `_tableTag()` return its `.lower()` or `.upper()`?

Lib/fontTools/fontBuilder.py
```diff
@@ -709,13 +758,8 @@ def setupAvar(self, axes, mappings=None):
         _add_avar(self.font, axes, mappings, axisTags)
 
     def setupGvar(self, variations):
-        gvar = self.font["gvar"] = newTable("gvar")
-        gvar.version = 1
-        gvar.reserved = 0
-        gvar.variations = variations
-
-    def setupGVAR(self, variations):
```
khaledhosny, line 717: Keep `setupGVAR()` for backward compatibility?

Lib/fontTools/ttLib/beyond64k.py
```diff
@@ -1,4 +1,27 @@
-"""Convert between OpenType beyond-64k companion tables."""
+"""Convert between OpenType beyond-64k companion tables.
+
+The OpenType beyond-64k specification adds uppercase companion tables and
```
khaledhosny, line 3: Maybe instead of beyond-64k mention the (upcoming) OpenType version that will have these tables?

Lib/fontTools/ttLib/removeOverlaps.py
```diff
@@ -293,8 +295,8 @@ def removeOverlaps(
 
     Overlapping components are first decomposed to simple contours, then merged.
 
-    Currently this only works for fonts with 'glyf' or 'CFF ' tables.
-    Raises NotImplementedError if 'glyf' or 'CFF ' tables are absent.
+    Currently this only works for fonts with 'glyf'/'GLYF' or 'CFF ' tables.
+    Raises NotImplementedError if 'glyf'/'GLYF' or 'CFF ' tables are absent.
```
khaledhosny, line 299: An old issue, but looks like the code supports `CFF2` table, but the doc string mentions `CFF ` only.

Lib/fontTools/qu2cu/cli.py
khaledhosny, line 39: This is from the old `glyf1` proposal, right?

Lib/fontTools/ttLib/beyond64k.py
behdad: It is uncertain whether these make it into OpenType. We can mention OFF5 (OpenFontFormat ...) but that's illegible to most people.

Lib/fontTools/fontBuilder.py
behdad: No one should be using it. I added it such that @justvanrossum can test making a varfont where `gvar` was overflowing offsets. It was never used beyond such testing.

Lib/fontTools/ttLib/beyond64k.py
behdad: In the latest commit I removed this forcing completely.

Lib/fontTools/merge/__init__.py
```diff
@@ -188,12 +188,22 @@ def _preMerge(self, font):
...
+        self._dropPostGlyphNamesIfNeeded(font)
 
         if "OS/2" in font:
             # https://github.com/fonttools/fonttools/issues/2538
             # TODO: Add an option to disable this?
             font["OS/2"].recalcAvgCharWidth(font)
 
+    def _dropPostGlyphNamesIfNeeded(self, font):
+        if "post" not in font or len(font.getGlyphOrder()) <= 0xFFFF:
+            return
```
khaledhosny, line 201: So there is no `POST` table? Nice!

Lib/fontTools/ttLib/tables/otBase.py
```diff
@@ -119,6 +119,11 @@ def compile(self, font):
         # or it errors out; False, don't use it, even if you can.
         use_hb_repack = font.cfg[USE_HARFBUZZ_REPACKER]
         if self.tableTag in ("GSUB", "GPOS"):
+            if (
+                use_hb_repack is None
+                and getattr(self.table, "Version", 0) >= 0x00010002
+            ):
+                use_hb_repack = False
```
khaledhosny, line 126: Are we skipping hb.repack because repacking is not needed or because the new table version is not supported by HarfBuzz?

Lib/fontTools/ttLib/tables/_c_m_a_p.py
```diff
@@ -1545,6 +1551,20 @@ def compile(self, ttFont):
...
 
+    def compileNonDefaultUVSRecord(self, uv, gid):
+        return struct.pack(">3sH", cvtFromUVS(uv), gid)
+
+
+class cmap_format_15(cmap_format_14):
+    nonDefaultUVSRecordSize = 6
+
+    def decompileNonDefaultUVSRecord(self, data):
+        uv, gid = struct.unpack(">3s3s", data)
```
khaledhosny, line 1562: The spec (incorrectly I think) still uses uint16 for the glyphID in format 15.

anthrotype:
is https://github.com/harfbuzz/boring-expansion-spec/blob/main/beyond-64k.md the latest version of the proposed spec?

behdad:
> is https://github.com/harfbuzz/boring-expansion-spec/blob/main/beyond-64k.md the latest version of the proposed spec?

No. Bizarre as it sounds, latest version is spread in a 900 page draft of the ISO Open Font Format that is not public. I sent you a copy privately.

anthrotype:
This should be extended to contain also MAXP and LOCA 

https://github.com/fonttools/fonttools/blob/c5d5359f7ab9989306ee0ff5897d834d0f1b5c29/Lib/fontTools/ttLib/tables/_h_e_a_d.py#L47

otherwise `head.indexToLocFormat` could get compiled with its default 0 while LOCA is written long and the font can't be read back and the following crashes with: `fontTools.ttLib.TTLibError: not enough 'glyf' table data`

```python
from io import BytesIO
from fontTools.fontBuilder import FontBuilder
from fontTools.pens.ttGlyphPen import TTGlyphPen

pen = TTGlyphPen(None)
pen.moveTo((0, 0))
pen.lineTo((0, 700))
pen.lineTo((500, 700))
pen.closePath()
triangle = pen.glyph()

glyphs = [".notdef"] + ["g%d" % i for i in range(1, 7000)]

fb = FontBuilder(1000, isTTF=True, beyond64k=True)
fb.setupGlyphOrder(glyphs)
fb.setupGlyf({name: triangle for name in glyphs})
fb.setupHorizontalMetrics({name: (600, 0) for name in glyphs})
fb.setupHorizontalHeader(ascent=800, descent=-200)
fb.setupCharacterMap({0x41: "g1"})
fb.setupNameTable({"familyName": "Test", "styleName": "Regular"})
fb.setupOS2()
fb.setupPost(keepGlyphNames=False)

buf = BytesIO()
fb.save(buf)

font = TTFont(buf)
print("head.indexToLocFormat:", font["head"].indexToLocFormat)
print("LOCA length:", len(font.reader["LOCA"]), "bytes for", len(glyphs) + 1, "entries")
glyph = font["GLYF"][font.getGlyphOrder()[1]]
print("glyph 1 bbox:", (glyph.xMin, glyph.yMin, glyph.xMax, glyph.yMax))
```

Lib/fontTools/ttLib/tables/otData.py
anthrotype: Maybe the spec is the one that's off here, instead of the code?

The non-chained `SequenceRuleSet2` (also glyph-based) keeps `seqRuleOffsets` at Offset16 in the spec, and `ChainedSequenceRuleSet2` is the only glyph-based ruleset the spec bumps to Offset24. Everywhere else it's consistently glyph-based -> Offset16, class-based -> Offset24.

Note that HarfBuzz keeps this Offset16 too (ChainRuleSet is hardcoded to SmallTypes):

https://github.com/harfbuzz/harfbuzz/blob/882ce8e6f381c45c9bf129aad37b284d5abcff73/src/hb-ot-layout-gsubgpos.hh#L3706-L3709

So following the spec would de-sync us from HB...

The class-keyed rulesets, `ClassSequenceRuleSet2` and `ChainedClassSequenceRuleSet2`, are Offset24 throughout, so the glyph-keyed chained table is the only one out of step. Is this inconsistency intentional?

Lib/fontTools/ttLib/tables/otData.py
anthrotype: looks like this one slipped the Offset24 conversion. The other extended offset arrays Khaled flagged (PairSet2, LigatureSet2, Sequence2/AlternateSet2, the Mark/Anchor arrays, etc.) all got switched to `Offset24To()`.

anthrotype:
@behdad PTAL https://github.com/fonttools/fonttools/pull/4099

anthrotype:
Two more issues arise when compiling masters with differing hand-written per-master .fea (`_featuresCompatible` false or `variableFeatures=False`) or when invoking the varLib CLI on master TTFs with pre-compiled GPOS:

* varLib.merger raises `UnsupportedFormat` on the beyond-64k extended formats (PairPos 3/4, Single 3/4, Mark Format 2 etc.).
* `_merge_OTL` asserts GDEF.Version <= 0x00010002 while feaLib promotes a >64k master’s GDEF to v1.4.

Simple repro below (for PairPos only):

```python
from fontTools.fontBuilder import FontBuilder
from fontTools.ttLib.tables._g_l_y_f import Glyph
from fontTools.feaLib.builder import addOpenTypeFeaturesFromString
from fontTools.designspaceLib import (
    DesignSpaceDocument,
    AxisDescriptor,
    SourceDescriptor,
)
from fontTools import varLib

order = [".notdef"] + ["g%05d" % i for i in range(1, 0x10001)]  # 65537 glyphs


def make_master(kern):
    fb = FontBuilder(unitsPerEm=1000)
    fb.setupGlyphOrder(order)
    fb.setupGlyf({g: Glyph() for g in order})
    fb.setupHorizontalMetrics({g: (500, 0) for g in order})
    fb.setupHorizontalHeader(ascent=800, descent=-200)
    fb.setupNameTable({"familyName": "T", "styleName": "R"})
    fb.setupPost(keepGlyphNames=False)
    addOpenTypeFeaturesFromString(
        fb.font, "feature kern { pos g00001 g00002 %d; } kern;" % kern
    )
    return fb.font


doc = DesignSpaceDocument()
doc.addAxis(AxisDescriptor(tag="wght", name="Weight", minimum=0, default=0, maximum=1000))
for loc, kern in [(0, -50), (1000, -120)]:
    doc.addSource(SourceDescriptor(font=make_master(kern), location={"Weight": loc}))

varLib.build(doc)  # raises UnsupportedFormat in varLib/merger.py
```

Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -2283,7 +2358,7 @@ def fixLookupOverFlows(ttf, overflowRecord):
                 extSubTable.Format = 1
                 extSubTable.ExtSubTable = subTable
                 lookup.SubTable[si] = extSubTable
-    ok = 1
+                ok = 1
```
anthrotype: @behdad why this change here? It looks independent from this PR

behdad: You're right. I was getting into a situation with some version of the PR where extension was tried but didn't succeed, and it resulted in infinite loop. This addresses that to raise an error. I'll submit separate.

behdad: https://github.com/fonttools/fonttools/pull/4107

PR #4099, opened 2026-06-15
[beyond64k] Support > 64k glyphs in HVAR/VVAR and fix head/LOCA save order

anthrotype:
Two fixes on top of @behdad's `beyond-64k` branch, found while reviewing and building variable fonts with more than 0xFFFF glyphs.

varLib couldn't build HVAR/VVAR past 64k glyphs because `_add_VHVAR` raised `struct.error` from two uint16 ceilings. 

The direct (no-map) advance store packs one VarData item per glyph, but VarData.ItemCount is uint16, so it can't index more than 0xFFFF glyphs; an indirect store is needed when > 64k.

And the HVVAR's `VarIdxMap` had a fixed uint16 MappingCount: we need to make VarIdxMap a subclass of the
format-switching `DeltaSetIndexMap` so it bumps to Format 1 (uint32 MappingCount) once the map exceeds 0xFFFF entries, while keeping its glyph-name-keyed mapping (the `DeltaSetIndexMap` class is also used by COLRv1 and avar2 and the mapping keys there are not glyph names, so we need a distinct class even though the serialized table is the same).
I don't see another way to avoid breaking existing users of either VarIdxMap or DeltaSetIndexMap.

Separately, head must depend on the uppercase MAXP/LOCA companions. The latter sort last in the save order, so a beyond-64k font with long loca wrote head before LOCA and froze a stale `indexToLocFormat`, corrupting the saved offsets. Listing the two uppercase tables as `head.dependencies` makes head write after them, matching post which already depends on both maxp and MAXP.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_h_e_a_d.py
+++ b/Lib/fontTools/ttLib/tables/_h_e_a_d.py
@@ -44,7 +44,7 @@ class table__h_e_a_d(DefaultTable.DefaultTable):
     See also https://learn.microsoft.com/en-us/typography/opentype/spec/head
     """
 
-    dependencies = ["maxp", "loca", "CFF ", "CFF2"]
+    dependencies = ["maxp", "MAXP", "loca", "LOCA", "CFF ", "CFF2"]
 
     def decompile(self, data, ttFont):
         dummy, rest = sstruct.unpack2(headFormat, data, self)
--- a/Lib/fontTools/ttLib/tables/otData.py
+++ b/Lib/fontTools/ttLib/tables/otData.py
@@ -4304,10 +4304,26 @@ otData = [
     ),
     # Variation helpers
     (
-        "VarIdxMap",
+        "VarIdxMapFormat0",
         [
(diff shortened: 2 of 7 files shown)


behdad:
> The direct (no-map) advance store packs one VarData item per glyph, but VarData.ItemCount is uint16, so it can't index more than 0xFFFF glyphs; an indirect store is needed when > 64k.

My original design of the VarStore had varIdx as a 32bit number constructed as `(major << 16) + minor`. In that world, glyph indices > 64k would translate to valid varIdx'es. I wonder if we should use that and update the spec for it.

behdad:
> > The direct (no-map) advance store packs one VarData item per glyph, but VarData.ItemCount is uint16, so it can't index more than 0xFFFF glyphs; an indirect store is needed when > 64k.
> 
> My original design of the VarStore had varIdx as a 32bit number constructed as `(major << 16) + minor`. In that world, glyph indices > 64k would translate to valid varIdx'es. I wonder if we should use that and update the spec for it.

Looks good for now. We can switch to the above design later.

PR #4100, opened 2026-06-16
Implement support for TrueType Collection v2

jenskutilek:
TrueType Collections with version 2 can contain a DSIG table in the TTC structure. This is currently ignored when reading or saving a `TTCollection` with FontTools.

This PR adds support for reading and writing version 2 TTC headers and TTC DSIG tables.

References:
- [TTC Header spec](https://learn.microsoft.com/en-us/typography/opentype/spec/otff#ttc-header)
- [DSIG table in TTC](https://learn.microsoft.com/en-us/typography/opentype/spec/dsig#signatures-for-font-collection-files)

A real-world example for a TTC with DSIG is `cambria.ttc` in Windows.

```py
from fontTools.ttLib import TTCollection
ttc = TTCollection('CAMBRIA.TTC')
ttc.dsig.decompile(ttc.dsig.data, None)
ttc.dsig.ulVersion  # 1
ttc.dsig.usNumSigs  # 1
ttc.dsig.usFlag  # 1
ttc.dsig.signatureRecords
# [<SignatureRecord: {'ulFormat': 1, 'ulLength': 6750, 'ulOffset': 20, 'usReserved1': 0, 'usReserved2': 0, 'cbSignature': 6742, 'pkcs7': b'0\x82\x1aR ...
```


jenskutilek:
Right now you have to explicitly save as version 2 to keep the version and DSIG, even if the TTCollection was opened from a version 2 file. Maybe there is a better solution without breaking backwards compatibility.

Lib/fontTools/ttLib/sfnt.py
```diff
@@ -433,6 +433,18 @@ def reordersTables(self):
 
 ttcHeaderSize = sstruct.calcsize(ttcHeaderFormat)
 
+ttcHeaderFormat2 = """
```
behdad: I think calling this Tail instead of Header makes more sense.

behdad:
> Right now you have to explicitly save as version 2 to keep the version and DSIG, even if the TTCollection was opened from a version 2 file. Maybe there is a better solution without breaking backwards compatibility.

Since before the object never had a `dsig` member, why not just write it if available? That would write the wrong signature though, if one opens, modifies the fonts, and save?

Tests/ttLib/ttCollection_test.py
```diff
@@ -25,3 +28,112 @@ def test_lazy_open_file(lazy):
         assert len(collection) == 2
         assert collection[0]["maxp"].numGlyphs == 6
         assert collection[1]["maxp"].numGlyphs == 6
+
+
+def test_save_ttc_v2_empty_dsig(lazy):
```
anthrotype: I don't understand what this is testing. Why would one try to set `ttc.dsig` to an empty placeholder without any of the required DSIG header attributes and expect our code to silently skip and swallow that with that overly broad exception (that may silently mask other compilation bugs):

```python
                except KeyError:
                    # Incomplete DSIG
                    write_dsig = False
```

who does that? The 'dummy' dsig that people insert is not empty (0 bytes), it has no signature records, but the header is there.

https://github.com/googlefonts/gftools/blob/20f79823a1641c477549111b8e9680e91e9ce212/Lib/gftools/fix.py#L168-L184

jenskutilek: I've changed the code so that compiling an invalid DSIG table will raise an exception.

jenskutilek:
Updated logic:

## Reading a TTC which has version 2

- If a `DSIG` table exists (tag, length, and offset fields are populated), it is read into `table_D_S_I_G_.data` and assigned to `TTCollection.dsig`
- If no `DSIG` table exists (tag, length, and offset fields are 0), `TTCollection.dsig` is set to `None`

## Writing a TTC

- If `TTCollection` doesn't have the attribute `dsig`, the `TTCollection` is written as version 1
- If `TTCollection` has the attribute `dsig` and it is `None`, the `TTCollection` is written as version 2, with the tag, length, and offset fields set to 0
- If `TTCollection` has the attribute `dsig` and it is a valid `table_D_S_I_G_` (compiled or uncompiled), the `TTCollection` is written as version 2, with the tag, length, and offset fields populated. If the `table_D_S_I_G_` is uncompiled, compilation is attempted, and if it fails, an error is raised.

The case where the DSIG fields are written as 0 is useful for actually signing the TTC, because if those additional fields are present, the offsets in the member fonts are already correct and the content digest can be directly calculated from the binary TTC.

Makes sense?

anthrotype:
Yes. But what's an "uncompliled" DSIG?

anthrotype:
Oh ok I see now, ignore my previous comment.

jenskutilek:
> But what's an "uncompliled" DSIG?

Like every other table, one that doesn't have a `data` attribute, and in the case of DSIG, has the attributes `ulVersion`, `usNumSigs`, `usFlag`. and `signatureRecords`.

jenskutilek:
> That would write the wrong signature though, if one opens, modifies the fonts, and save?

That happens for DSIG tables in non-collection fonts as well. The user has to take care of it, I'd say.

jenskutilek:
I've successfully generated signed TTCs based on this code now.

Related: Here's a [Hex Fiend template for TrueType Collection](https://gist.github.com/jenskutilek/153e7b115cb10d82e88588b61d0736b4) files.

behdad (approved):

behdad (approved):

anthrotype:
Thank you!

PR #4101, opened 2026-06-16
[varLib.merger] Support beyond-64k extended formats in GPOS merger

anthrotype:
Building a variable font from masters whose GPOS/GDEF use the beyond-64k extended layouts (PairPos/SinglePos format 3/4, mark/cursive format 2, GDEF v1.4) raised `UnsupportedFormat` in the varLib.merger, which only handles format 1/2.
feaLib promotes a master's layout tables to the extended formats by glyph count, so every master of a font with more than 65535 glyphs is already extended.

I opted for demoting the extended GPOS subtables to format 1/2 before merging and promoting the merged output back afterwards, so that the existing mergers continue to work as is. The conversion is lossless since Coverage and ClassDef pick their own format at compile time. It lives in the shared Lookup merger, so it covers both varLib.build and the instancer. Contextual positioning, mark-to-ligature and GSUB merge through the generic object walker, which leaves the extended subtables intact, so they need no special handling.

I also changed `_merge_OTL` to attach the VarStore to a beyond-64k v1.4 GDEF instead of failing on the v1.2 assertion.

Two tests build a >64k VF through this path: one round-trips the whole font, checks the lookup types come out extended, and GDEF VarStore is preserved; the other checks PairPos coverage alignment and backfill across masters that kern against disjoint >64k glyphs.

@behdad if you had a different idea, feel free to take only the tests and discard/rework the implementation.

### DIFF
--- a/Lib/fontTools/varLib/__init__.py
+++ b/Lib/fontTools/varLib/__init__.py
@@ -888,7 +888,9 @@ def _merge_OTL(font, model, master_fonts, axisTags):
         return
     try:
         GDEF = font["GDEF"].table
-        assert GDEF.Version <= 0x00010002
+        # v1.4 is the beyond-64k GDEF version; bump to >= v1.3 to carry the VarStore
+        assert GDEF.Version <= 0x00010004
+        GDEF.Version = max(GDEF.Version, 0x00010003)
     except KeyError:
         GDEFTable = font["GDEF"] = newTable("GDEF")
         GDEF = GDEFTable.table = ot.GDEF()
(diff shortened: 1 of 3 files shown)


anthrotype:
I pushed a small follow-up (dd4e9bc8) for a GDEF edge case that this PR exposed: the GlyphClassDef merger was keyed on the `GlyphClassDef` attribute name only, so a beyond-64k GDEF v1.4's `GlyphClassDef2` fell through to the generic merge and raised `ShouldBeConstant` whenever masters classified different glyphs. Registering the merger for both restores the union semantics.

PR #4102, opened 2026-06-17
[ufoLib] Make glyph-to-group construction accessible outside of lookup function

Hoolean:
This improves ergonomics for constructing glyph-to-group mappings prior to repeated kern lookups: you get the convenience of fonttools constructing the mappings, but avoid the repeated cost on each call.

### DIFF
--- a/Lib/fontTools/ufoLib/kerning.py
+++ b/Lib/fontTools/ufoLib/kerning.py
@@ -32,7 +32,7 @@ def lookupKerningValue(
       kerning:
           A dictionary of kerning pairs.
       groups:
-          A set of kerning groups.
+          A dictionary mapping each group name to its glyph names.
       fallback:
           The fallback value to return if no kern is found between
           the elements in ``pair``. Defaults to 0.
@@ -137,7 +137,7 @@ def glyphsToGroups(groups: KerningGroups) -> tuple[StrDict, StrDict]:
 
     Args:
       groups:
-          A dictionary of kerning groups.
+          A dictionary mapping each group name to its glyph names.
 
     Returns:
       A tuple of two dictionaries, respectively mapping glyph names to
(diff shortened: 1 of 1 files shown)


Lib/fontTools/ufoLib/kerning.py
```diff
@@ -135,6 +127,40 @@ def lookupKerningValue(
...
 
 
+def glyphsToGroups(groups: KerningGroups) -> tuple[StrDict, StrDict]:
+    """Build dictionaries mapping glyph names to their kerning groups.
+
+    This is a reverse-mapping of the UFO kerning groups data structure.
+
+    Args:
+      groups:
+          A set of kerning groups.
```
anthrotype:
```suggestion
          A dictionary of kerning groups.
```

Lib/fontTools/ufoLib/kerning.py
```diff
@@ -135,6 +127,40 @@ def lookupKerningValue(
     return fallback
 
 
+def glyphsToGroups(groups: KerningGroups) -> tuple[StrDict, StrDict]:
+    """Build dictionaries mapping glyph names to their kerning groups.
+
+    This is a reverse-mapping of the UFO kerning groups data structure.
```
anthrotype, line 136: maybe document here the non-kerning groups are ignored and the input is expected to have already passed UFO group validation? The helper currently does last group wins if a glyph appears in more than one.

anthrotype (approved):

Lib/fontTools/ufoLib/kerning.py
Hoolean: Thanks, I copied this over without checking in error; I've updated and clarified it for this and the original function too now

Lib/fontTools/ufoLib/kerning.py
Hoolean: I've documented this now :+1: I was tempted to start raising a `ValueError` instead, but that felt too radical for a refactor; I've tried to leave the door open a little with the new documentation anyway

Hoolean:
Ta for speedy review; I've applied all feedback and added a new accelerated example to the docs

PR #4103, opened 2026-06-17
[glyf] Use reverse glyph map for O(1) __setitem__ membership

anthrotype:
`table__g_l_y_f.__setitem__` tested `glyphName not in self.glyphOrder`, a linear scan of the glyphOrder list on every assignment. So repeatedly assigning glyphs over a pre-populated glyphOrder (like ufo2ft outlineCompiler currently does) was O(n^2) in the glyph count, which means several seconds of pure scanning for a 60k-glyph font.

We can instead use the existing reverse glyph map (`self._reverseGlyphOrder`, the same we added to speed up `getGlyphID`) for O(1) membership, rebuilding it when it is missing or its length no longer matches `self.glyphOrder`, and invalidating it on `__delitem__`.

Reordering goes through `setGlyphOrder` as usual, which already clears the map. And editing glyphOrder directly in-place behind the table's back was never supported.

This is somewhat related to #2605, a separate correctness issue about the TTFont-level `_reverseGlyphOrderDict` going stale (see #3582), I have not changed it here.

It's also relevant for #4097 because that's exactly how I stumbled into this perf issue: by writing unit tests for ufo2ft's beyond-64k PR (https://github.com/googlefonts/ufo2ft/pull/987).

In ufo2ft I ended up circumventing the `glyf.__setitem__` altogether and directly assigning the inner `glyf.glyphs` dictionary (like our own fontBuilder does): https://github.com/googlefonts/ufo2ft/pull/991

But It'd still be good to have this optimization in case someone else may benefit.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -338,12 +338,23 @@ class table__g_l_y_f(DefaultTable.DefaultTable):
 
     def __setitem__(self, glyphName, glyph):
         self.glyphs[glyphName] = glyph
-        if glyphName not in self.glyphOrder:
+        # Use the reverse glyph map for O(1) membership so that building a font by
+        # repeatedly assigning glyphs stays linear instead of quadratic. Rebuild
+        # the map if it is missing or no longer matches the length of glyphOrder;
+        # reordering goes through setGlyphOrder, which clears it.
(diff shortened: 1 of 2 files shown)


anthrotype:
e.g. qu2cu's cli is also affected directly by this

https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/qu2cu/cli.py#L30-L35

behdad:
Lgtm.

PR #4106, opened 2026-06-18
[beyond64k] Drop post 2.0 names at 65536 glyphs, not 65537

anthrotype:
post 2.0 numberOfGlyphs is uint16, so a glyph count of exactly 65536 already overflows it even though every glyph id still fits. _drop_beyond64k_post_glyph_names gated on > 0x10000 (i.e. >= 65537) instead of > 0xFFFF (>= 65536), leaving such a font with an unencodable post 2.0 table.


anthrotype:
the merger had the same issue, fixed in the same go

Issue #4091, opened 2026-05-16
GPOS pack fails (or hangs with uharfbuzz) when LookupList overflow needs splitting a SinglePos (type 1) subtable


AmitMY:
## Summary

When the GPOS table contains enough large `SinglePos Format 2` subtables to overflow the `LookupList`'s uint16 offsets, fontTools' compile fails because the dispatch table for overflow recovery has no entry for lookup type 1 — `splitSinglePos` is commented out at [`Lib/fontTools/ttLib/tables/otTables.py:2514`](https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/ttLib/tables/otTables.py#L2514).

The user-visible failure mode depends on whether `uharfbuzz` is installed, but the underlying gap is the same:

- **Without `uharfbuzz`** — `TTFont.save()` raises `OTLOffsetOverflowError` (after the splitter logs `Don't know how to split GPOS lookup type 1` and `tryResolveOverflow` gives up). At least it terminates with a clear failure.
- **With `uharfbuzz` installed** — fontTools delegates to `hb.repack` first, which raises `RepackerError`. fontTools falls back to the same (incomplete) splitter, but the outer retry loop in `BaseTTXConverter.compile` never gives up: it keeps re-trying with the same unsplittable overflow record. **`save()` hangs indefinitely.**

Same shape as #1328 (the type 6 / `MarkMarkPos` variant, open since 2020 — `splitMarkMarkPos` is also missing from the dispatch table).

## Verification matrix

Both arms verified on fresh venvs.

| fontTools | `uharfbuzz` | Result with the reproducer below |
|---|---|---|
| 4.62.1 | not installed | `OTLOffsetOverflowError` after ~70 s ❌ |
| 4.62.1 | 0.54.1 | hangs indefinitely ❌ |
| 4.63.0 | not installed | `OTLOffsetOverflowError` after ~70 s ❌ |
| 4.63.0 | 0.54.1 | hangs indefinitely ❌ |

The bare-fontTools packer does handle smaller GPOS that hb.repack would refuse (e.g. N=800 in this repro), so swapping in/out `uharfbuzz` is **not** a real workaround — past the bare-packer threshold the only difference is whether you get a clean exception or a hang.

## Reproducer (self-contained)

```python
"""N SinglePos Format 2 lookups, each with M glyphs in coverage.
N=3500 saves cleanly without uharfbuzz. N=4000 fails as described."""
import time
import fontTools
[...]

Issue #1328, opened 2018-09-27
Support splitting GPOS lookup type 6 (MarkMarkPos)


miguelsousa:
```
ERROR:fontTools.ttLib.tables.otTables:Don't know how to split GPOS lookup type 6
```
I'm getting this error while trying to build the upcoming version of Source Sans variable.
The source files are at https://github.com/adobe-fonts/source-sans-pro/tree/fontmake

Build command used
`fontmake -m Roman/Masters/SourceSansVariable-Roman.designspace -o variable`

behdad:
Cosimo, you can fix this, right?

anthrotype:
I can have a look. If the mkmk feature was generated by ufo2ft, then each mkmk lookup contains only one mark class, and therefore we can’t use the same approach that we used for fixing overflows in MarkToBase subtables (i.e. splitting half of the mark classes to the new subtable), as there is only one per lookup.
Maybe we should split the coverage tables, e.g. half of the mark2 glyphs go to the new subtable, but shares the same mark1?

kontur:
I've encountered these warnings for type 5 and 6 with some variable fonts that have their mkmk also created with ufo2ft. 

What causes those overflows? And what does this error mean for the resulting font and is there something in the feature file or ufo source I can do to amend it?

PR #4108, opened 2026-06-21
Implement splitSinglePos for overflow recovery (#4091)

### LINKED ISSUE #1: Subsetting CFFs without subroutines
When testing the subsetter script, I came across a very complex font that apparently doesn’t use subroutines. The subsetter chokes on that ... I think the script has to check if there are any subroutines:

``` patch
--- subset.py   2013-09-03 12:08:32.000000000 +0200
+++ subset.py   2013-09-03 12:47:09.000000000 +0200
@@ -1404,7 +1404,8 @@
     if hasattr(font, 'FDSelect'):
       all_subrs.extend(fd.Private.Subrs for fd in font.FDArray if hasattr(fd.Private, 'Subrs'))
     else:
-      all_subrs.append(font.Private.Subrs)
+      if hasattr(font.Private, 'Subrs'):
+        all_subrs.append(font.Private.Subrs)
     # Prepare
     for subrs in all_subrs:
       if not subrs: continue
```

AmitMY:
## Summary

Implements `splitSinglePos`, the missing GPOS lookup-type-1 entry in the overflow-recovery `splitTable` dispatch (it was commented out at `otTables.py`). Until now, a `SinglePos Format 2` subtable whose internal `Coverage` offset overflowed uint16 could not be split, so `fixSubTableOverFlows` logged *"Don't know how to split GPOS lookup type 1"* and packing failed.

This is the type-1 sibling of the already-present `splitPairPos` / `splitMarkBasePos`, and the same class of gap as #1328 (type 6).

## Implementation

`SinglePos Format 2`'s `Value` array is 1:1 with its `Coverage`, so the subtable splits cleanly along the coverage — the same shape as `splitPairPos`. `Format 1` shares a single `Value` across the whole coverage and has nothing to split, so it returns `False`.

Verified end-to-end: a single `SinglePos Format 2` with 40 000 glyphs (whose `Coverage` offset overflows) now compiles, splitting into two subtables that round-trip back to all 40 000 glyphs.

## Note on the #4091 reproducer

The N=4000 reproducer in the issue actually hits a **different**, architectural limit: the `LookupList`'s `Offset16` cap. With ~4000 lookups, the offset to lookup ~3596 (≈ 8000-byte offset array + 3596×16 bytes per extension-wrapped lookup) exceeds 65535 even when every lookup is promoted to Extension — no subtable split can recover it (the fix there is fewer/merged lookups). `splitSinglePos` is still a genuine missing feature and resolves the *splittable* case this issue named as fix #1.

## Test

`test_splitSinglePos` (Format 2 splits along coverage) and `test_splitSinglePos_format1` (Format 1 is unsplittable) in `Tests/ttLib/tables/otTables_test.py`.

### DIFF
--- a/Lib/fontTools/ttLib/tables/otTables.py
+++ b/Lib/fontTools/ttLib/tables/otTables.py
@@ -2368,6 +2368,33 @@ def splitLigatureSubst(oldSubTable, newSubTable, overflowRecord):
     return ok
 
 
+def splitSinglePos(oldSubTable, newSubTable, overflowRecord):
+    # Only Format 2 (one Value per covered glyph) is worth splitting; a Format 1
+    # subtable shares a single Value across its whole Coverage, so there is
+    # nothing worth splitting.
+    if oldSubTable.Format != 2 or len(oldSubTable.Coverage.glyphs) <= 1:
+        return False
+
+    newSubTable.Format = oldSubTable.Format
+    newSubTable.ValueFormat = oldSubTable.ValueFormat
+
+    coverage = oldSubTable.Coverage.glyphs
+    values = oldSubTable.Value
+    oldCount = len(coverage) // 2
+
+    newSubTable.Coverage = oldSubTable.Coverage.__class__()
+    newSubTable.Coverage.glyphs = coverage[oldCount:]
+    newSubTable.Value = values[oldCount:]
+
+    oldSubTable.Coverage.glyphs = coverage[:oldCount]
+    oldSubTable.Value = values[:oldCount]
+
+    oldSubTable.ValueCount = len(oldSubTable.Value)
+    newSubTable.ValueCount = len(newSubTable.Value)
+
+    return True
+
+
 def splitPairPos(oldSubTable, newSubTable, overflowRecord):
     st = oldSubTable
     ok = False
@@ -2511,7 +2538,7 @@ splitTable = {
         # 					8: splitReverseChainSingleSubst,
     },
     "GPOS": {
-        # 					1: splitSinglePos,
+        1: splitSinglePos,
         2: splitPairPos,
         # 					3: splitCursivePos,
         4: splitMarkBasePos,
--- a/Tests/ttLib/tables/otTables_test.py
+++ b/Tests/ttLib/tables/otTables_test.py
@@ -785,6 +785,78 @@ def test_splitSinglePos_format1():
     assert not otTables.splitSinglePos(subTable, otTables.SinglePos(), None)
 
 
+def test_splitSinglePos_overflow_end_to_end():
+    # A SinglePos Format 2 whose ValueArray pushes its own Coverage offset past
+    # uint16 must be split when packing. With ValueFormat=1 (2 bytes per Value)
+    # the Coverage offset is ~8 + 2*N bytes, so it overflows above ~32763 glyphs.
+    # https://github.com/fonttools/fonttools/issues/4091
+    from fontTools.ttLib import TTFont, getTableClass
+
+    N = 40000
+    glyphs = [f"g{i}" for i in range(N)]
+    font = TTFont()
+    font.setGlyphOrder([".notdef"] + glyphs)
+    # exercise the pure-fontTools overflow recovery, not the harfbuzz repacker
+    font.cfg["fontTools.ttLib.tables.otBase:USE_HARFBUZZ_REPACKER"] = False
+
+    subTable = otTables.SinglePos()
+    subTable.Format = 2
+    subTable.Coverage = otTables.Coverage()
+    subTable.Coverage.glyphs = list(glyphs)
+    subTable.ValueFormat = 0x0001  # XPlacement
+    subTable.Value = []
+    for i in range(N):
+        value = otTables.ValueRecord()
+        value.XPlacement = -(i % 100 + 1)
+        subTable.Value.append(value)
+
+    lookup = otTables.Lookup()
+    lookup.LookupType = 1
+    lookup.LookupFlag = 0
+    lookup.SubTable = [subTable]
+
+    feature = otTables.Feature()
+    feature.FeatureParams = None
+    feature.LookupListIndex = [0]
+    featureRecord = otTables.FeatureRecord()
+    featureRecord.FeatureTag = "kern"
+    featureRecord.Feature = feature
+    langSys = otTables.DefaultLangSys()
+    langSys.LookupOrder = None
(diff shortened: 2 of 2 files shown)


AmitMY:
Now that #4109 is merged, this is the orthogonal half of #4091: #4109 made the LookupList-level dead-end fail with an actionable error, and this PR adds the actual `splitSinglePos` so an oversized SinglePos Format 2 subtable can be split (the type-1 sibling of `splitPairPos`/`splitMarkBasePos`, same gap as #1328).

No conflicts with the merge and CI is green. Whenever you have a moment for a look — thanks!

Lib/fontTools/ttLib/tables/otTables.py
```diff
@@ -2368,6 +2368,32 @@ def splitLigatureSubst(oldSubTable, newSubTable, overflowRecord):
     return ok
 
 
+def splitSinglePos(oldSubTable, newSubTable, overflowRecord):
+    # Format 1 shares a single Value across the whole Coverage, so there is
+    # nothing to split; only Format 2 (one Value per covered glyph) can shrink.
```
anthrotype: Technically one could split the Coverage into two Format 1 subtables and duplicate the same shared ValueRecord. Though I don’t see an overflow-recovery case where it helps. Maybe the comment could just say "there is nothing worth splitting".

anthrotype:
Thanks. Overall the changes look good. May I suggest you remove that "Fixes" verb from the commit message (Github would auto-close the issue once the commit is merged). You can keep the link or say "Closes part of #4091" or similar.

The second part of the issue, the infinite retry / hang triggered by the 4000-lookup test font with uharfbuzz installed, where it seems to reach an unrecoverable LookupList -> Lookup Offset16 overflow, may still be outstanding I believe.
 
Worth checking if @behdad's recently merged https://github.com/fonttools/fonttools/pull/4107 might fix that.

anthrotype:
I can’t reproduce the infinite retry/hang with the exact reproducer on either current main (ca105fa4b) or the commit immediately before #4107 (c5c6b5c8a): with uharfbuzz installed, both eventually do fail with the same unrecoverable LookupList -> Lookup Offset16 overflow at lookup 3596, and I don’t see repeated `Don't know how to split GPOS lookup type 1` messages either. So I’d treat the 4000-lookup case as an unencodable LookupList limit, while this PR fixes the separate SinglePos Format 2 subtable-splitting gap.

AmitMY:
Thanks! Done in 69ec7fe4f:

- Dropped the auto-close verb — the commit message now says "Part of #4091" (and the PR body has no closing keyword either).
- Reworded the comment to "nothing worth splitting" — you're right, a Format 1 subtable could technically be split by duplicating the shared ValueRecord; it's just never worth it for overflow recovery.

And agreed on the 4000-lookup case: it's an unencodable LookupList → Lookup Offset16 limit (overflows around lookup 3596 even when every lookup is already Extension), not something subtable splitting can fix. #4109 now surfaces that exact dead-end with an actionable error instead of failing silently, so I think between the two PRs #4091 is well covered.

AmitMY:
Right — and to be clear, the 4000-lookup font in #4091 never reaches `splitSinglePos` (it dies at the LookupList Offset16 limit, as you found), so it isn't a reproducer for *this* PR. The case this PR fixes is a single `SinglePos Format 2` whose own ValueArray pushes its internal `Coverage` offset past uint16 — here's a self-contained reproducer:

```python
from fontTools.ttLib import TTFont, getTableClass

N = 40000
font = TTFont()
font.setGlyphOrder([".notdef"] + [f"g{i}" for i in range(N)])
glyphs = [f"g{i}" for i in range(N)]

sp = ot.SinglePos(); sp.Format = 2
sp.Coverage = ot.Coverage(); sp.Coverage.glyphs = glyphs
sp.ValueFormat = 1
sp.Value = [ot.ValueRecord() for _ in range(N)]
for i, v in enumerate(sp.Value):
    v.XPlacement = -(i % 100 + 1)

lk = ot.Lookup(); lk.LookupType = 1; lk.LookupFlag = 0; lk.SubTable = [sp]
feat = ot.Feature(); feat.FeatureParams = None; feat.LookupListIndex = [0]
frec = ot.FeatureRecord(); frec.FeatureTag = "kern"; frec.Feature = feat
fl = ot.FeatureList(); fl.FeatureRecord = [frec]
ls = ot.DefaultLangSys(); ls.LookupOrder = None; ls.ReqFeatureIndex = 0xFFFF; ls.FeatureIndex = [0]
sc = ot.Script(); sc.DefaultLangSys = ls; sc.LangSysRecord = []
srec = ot.ScriptRecord(); srec.ScriptTag = "DFLT"; srec.Script = sc
sl = ot.ScriptList(); sl.ScriptRecord = [srec]
ll = ot.LookupList(); ll.Lookup = [lk]
gpos = ot.GPOS(); gpos.Version = 0x00010000
gpos.ScriptList = sl; gpos.FeatureList = fl; gpos.LookupList = ll
GPOS = getTableClass("GPOS")(); GPOS.table = gpos
font["GPOS"] = GPOS
font.cfg["fontTools.ttLib.tables.otBase:USE_HARFBUZZ_REPACKER"] = False  # deterministic, exercise the fontTools splitter

data = GPOS.compile(font)
GPOS2 = getTableClass("GPOS")(); GPOS2.decompile(data, font)
subs = GPOS2.table.LookupList.Lookup[0].SubTable
print(f"{len(data)} bytes, {len(subs)} subtables, {sum(len(s.Coverage.glyphs) for s in subs)} glyphs")
```

- **main:** `Don't know how to split GPOS lookup type 1` → `OTLOffsetOverflowError ('GPOS', 0, 0, 'Coverage', None)`
- **this PR:** `80094 bytes, 2 subtables, 40000 glyphs` (round-trips intact)

Happy to fold this into a regression test if you'd prefer an end-to-end one alongside the unit test.

anthrotype:
Yes, thanks, since you have one, let's add it.

AmitMY:
Added in fda6bfb32 — `test_splitSinglePos_overflow_end_to_end` compiles a GPOS with that oversized SinglePos Format 2 (repacker off), and asserts the subtable is split and all coverage/values round-trip. Runs in ~0.07s.

Lib/fontTools/ttLib/tables/otTables.py
AmitMY: Reworded to "there is nothing worth splitting" in 69ec7fe4f — thanks.

PR #4109, opened 2026-06-22
Give an actionable error when a LookupList offset overflow is unrecoverable

AmitMY:
## Summary

When a **LookupList → Lookup offset overflows** (`OTLOffsetOverflowError` with `SubTableIndex is None`) and *every* preceding lookup is already an Extension lookup (`LookupType` 9 for GPOS / 7 for GSUB), `fixLookupOverFlows` has nothing left to promote. It walks its index back past 0 and returns `0` **silently**, so `BaseTTXConverter.compile` re-raises a bare `OTLOffsetOverflowError` with no indication of *why* recovery failed.

This change logs a clear, actionable error at that dead-end instead of failing silently:

```
GPOS LookupList offset overflowed and all lookups are already Extension lookups,
so the overflow can't be resolved by promotion. Enable the HarfBuzz repacker
(install uharfbuzz; it can pack this via subtable sharing/duplication) or reduce
the number of lookups.
```

## Rationale

Fonts with thousands of (context-)positioning lookups — e.g. `volt2ttf` output, or axis-decomposed positioning where each coordinate emits a chained-context + inner lookup pair — can push the LookupList past what uint16 offsets can address. Once that happens *and* every lookup is already wrapped in an Extension, the existing recovery (promote the previous lookup to Extension) has no move left.

I checked whether fontTools could actually *recover* here rather than just report better, and concluded a better error is the correct, in-scope fix:

- **The data is representable, and HarfBuzz already packs it.** With `uharfbuzz` installed (the default), `hb.repack` serializes a ~7200-all-Extension-lookup GPOS in well under 2s and it round-trips correctly. Users on a default install don't hit this path at all.
- **The pure-fontTools packer cannot, and it isn't merely a sharing gap.** It overflows regardless; forcing cross-extension sharing (`shareExtension=True`) in `getAllData` still overflows. fontTools uses a fixed depth-first layout with no overflow-driven reordering/duplication — only HarfBuzz's repacker does that. Reimplementing the repacker is out of scope.
- **You can't split a LookupList.** There is exactly one LookupList per GSUB/GPOS, and lookups are referenced by index (FeatureList, `PosLookupRecord`/`SubstLookupRecord`), so they can't be merged or split away to reduce the offset. So the actionable guidance — use the HarfBuzz repacker, or reduce lookup count — is genuinely the right answer for this dead-end.

The behavior is otherwise unchanged: the `OTLOffsetOverflowError` still propagates (it is genuinely unrecoverable in the pure-FT packer); we just no longer fail without telling the user why or what to do.

## Test

`test_fixLookupOverFlows_all_extension_logs_error`: a LookupList overflow where all lookups are already Extension now returns false **and** logs the actionable error.

### DIFF
--- a/Tests/ttLib/tables/otTables_test.py
+++ b/Tests/ttLib/tables/otTables_test.py
@@ -8,6 +8,8 @@ from types import SimpleNamespace
 from textwrap import dedent
 import unittest
 
+import pytest
+
 
 def makeCoverage(glyphs):
     coverage = otTables.Coverage()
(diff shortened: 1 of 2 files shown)


anthrotype:
The error message assumes this is always about LookupList->Lookup offset overflow, but `fixLookupOverFlows` is also called as a fallback for subtable overflows. 

https://github.com/fonttools/fonttools/blob/5e84a4c24/Lib/fontTools/ttLib/tables/otBase.py#L235

`tryResolveOverflow` in otBase.py calls it when `fixSubTableOverFlows` failed and in that path `overflowRecord.SubTableIndex is not None`:

https://github.com/fonttools/fonttools/blob/5e84a4c24/Lib/fontTools/ttLib/tables/otTables.py#L2258

The message "LookupList offset overflowed" would be wrong then, because it's a subtable offset that overflowed.
I suggest you guard the log on `overflowRecord.SubTableIndex is None` so it only fires for actual LookupList->Lookup overflows.

Also, "Install uharfbuzz" is wrong/confusing when uharfbuzz is already installed. When state is HB_FT, [`tryPackingHarfbuzz`](https://github.com/fonttools/fonttools/blob/5e84a4c24/Lib/fontTools/ttLib/tables/otBase.py#L184-L204) already tried the HB repacker. If it threw `RepackerError` or `MemoryError` (which can still happen in rare cases), the fallback [writer.getAllData(remove_duplicate=False)](https://github.com/fonttools/fonttools/blob/5e84a4c24/Lib/fontTools/ttLib/tables/otBase.py#L204) runs, and that overflow may also reach `fixLookupOverFlows`. Maybe soften to "if uharfbuzz is not installed, installing it may help" or check the import inside the function.

AmitMY:
Thanks, both spot on — fixed in 737d38c9f:

- Guarded the `log.error` on `overflowRecord.SubTableIndex is None`, so it only fires for a genuine LookupList→Lookup overflow. On the subtable-overflow fallback path (where `fixLookupOverFlows` is re-tried with `SubTableIndex` set) the "LookupList offset overflowed" wording would be wrong, so it's skipped there.
- Conditioned the advice on `have_uharfbuzz`: only suggest installing uharfbuzz when it isn't already installed; otherwise suggest reducing the number of lookups.

Added tests for all three cases: the not-installed branch (mentions uharfbuzz), the installed branch (asserts it does *not* say "install uharfbuzz"), and a subtable overflow with `SubTableIndex` set (asserts no LookupList message is logged).

anthrotype:
if you rebase or merge origin/main, the flaky tests should fix themselves

Tests/ttLib/tables/otTables_test.py
```diff
@@ -639,6 +639,89 @@ def test_fixLookupOverFlows_returns_false_without_progress():
...
+        ext.ExtSubTable = otTables.SinglePos()
+        lookup = otTables.Lookup()
+        lookup.LookupType = otTables.ExtensionPos.LookupType
+        lookup.SubTable = [ext]
+        lookups.append(lookup)
+    gpos.table.LookupList.Lookup = lookups
+    return font
+
+
+def test_fixLookupOverFlows_all_extension_logs_error(caplog, monkeypatch):
```
anthrotype: nit: this and the following test could be parametrized on `have_uharfbuzz`

AmitMY: Done — parametrized both on `have_uharfbuzz`. Also rebased on origin/main, CI should be green now. Thanks!

anthrotype (approved):

PR #4111, opened 2026-06-22
Pin a single head.modified timestamp across TTCollection.save

### LINKED ISSUE #4110: TTC v2 roundtrip tests are flaky and may depend on table compilation ~~state~~ time
The TTC v2 roundtrip tests added in #4100 assert exact TTC header bytes, including font offsets and DSIG offset. These values depend on the compiled size of every shared table, which can differ between raw (from the reader) and recompiled (from a decompiled table object).

This seems to have caused flaky CI failures on an unrelated PR #4109 across 4 re-run attempts, with different tests failing each time (different pytest-randomly seeds). 

- [attempt 2, Python 3.12](https://github.com/fonttools/fonttools/actions/runs/27945824838/job/82693458224): test_roundtrip_ttc_dsig_decompile[False]: `At index 31 diff: b'p' != b'8'`
- [attempt 2, Python 3.14 Cython](https://github.com/fonttools/fonttools/actions/runs/27945824838/job/82693458205): test_save_ttc_v2_dsig[False]: `At index 31 diff: b't' != b'<'`
- [attempt 3, Python 3.13](https://github.com/fonttools/fonttools/actions/runs/27945824838/job/82695352763): test_roundtrip_ttc_dsig[False]: `At index 31 diff: b'p' != b'8'`
- [attempt 3, Windows 3.14](https://github.com/fonttools/fonttools/actions/runs/27945824838/job/82695352942): test_save_ttc_v2_dsig_none[True]: `At index 1 diff: b'\x02' != b'\x00'`

At the fourth attempt they all passed:

Image

Byte 31 is the low byte of the `ulDsigOffset` field in the TTC v2 header; all three Linux failures show the DSIG offset landing 0x38 (56) bytes higher than expected.

I was unable to reproduce locally (macOS, Linux ARM64 and x86_64 containers with Python 3.10/3.12, tox, full requirements.txt, same random seeds), so I don't know the exact cause. The consistent +56 byte offset shift across failures suggests a specific table recompiling to a different size, but I couldn't identify which one or what triggers it.

The tests should verify the structural properties of the v2 header (version, font count, DSIG tag/length, DSIG data content) without asserting exact byte offsets that depend on table compilation state.

UPDATE: this has nothing to do with table compilation state, but it's a timing race, see below.

anthrotype:
Saving a TTC with `recalcTimestamp=True` (default) restamps each font's head with `timestampNow()` (1-second granularity); if the clock ticks between two fonts' `head.compile()` calls their heads diverge, no longer dedup, and the saved size shifts by a whole head table (+56 bytes), which causes the flaky TTC v2 roundtrip test failures in https://github.com/fonttools/fonttools/issues/4110.

Stamp one timestamp across the collection so the heads stay shareable.

Fixes https://github.com/fonttools/fonttools/issues/4110

### DIFF
--- a/Lib/fontTools/ttLib/ttCollection.py
+++ b/Lib/fontTools/ttLib/ttCollection.py
@@ -2,10 +2,49 @@ from fontTools.ttLib.ttFont import TTFont
 from io import BytesIO
 import struct
 import logging
+from contextlib import contextmanager
 from fontTools.ttLib.sfnt import TTC_V1, TTC_V2, readTTCHeader, writeTTCHeader
 from fontTools.ttLib.tables.D_S_I_G_ import table_D_S_I_G_
+from fontTools.misc.timeTools import timestampNow
 
 log = logging.getLogger(__name__)
+
+
+@contextmanager
+def _sharedModifiedTimestamp(fonts):
+    """Pin a single ``head.modified`` timestamp across an iterable of TTFonts.
+
+    When several fonts are saved together with ``recalcTimestamp=True``, each
+    one restamps its own ``head`` table via ``timestampNow()``, whose
+    granularity is one second. If the wall clock ticks between two fonts'
+    ``head.compile()`` calls they get different ``modified`` values, which
+    defeats byte-identical table sharing (``shareTables=True``) and makes the
+    saved size non-deterministic.
+
+    This reads ``timestampNow()`` once (so ``SOURCE_DATE_EPOCH`` is still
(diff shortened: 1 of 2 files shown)


behdad:
Wow. Thanks for tracking down. Lg.

behdad (approved):

anthrotype:
yeah this was hard to track down, I couldn't reproduce it locally because my machine is too fast 🤣

PR #4113, opened 2026-06-24
[ttLib] add EBSC — Embedded Bitmap Scaling Table

TakWolf:
There may be some rough spots that require adjustment:

```python
XMLAttrs = dict[str, str]
XMLContent = list[str | tuple[str, XMLAttrs, "XMLContent"]]
```

Additional, the typing of the base class seems incorrect, I'm not sure

```python
class DefaultTable:
    def toXML(
        self, writer: XMLWriter, ttFont: TTFont, **kwargs: dict[str, Any]  // <- this
    ) -> None:
       pass

    def fromXML(
        self, name: str, attrs: dict[str, str], content: str,   // <- is this str?
        ttFont: TTFont
    ) -> None:
        pass

```


behdad:
Lgtm in general.

anthrotype:
Thanks! I will fix black in a follow up commit (github doesn't let me push to your PR branch)

anthrotype:
I made a small follow-up in 05a3477 on main to reuse the shared XML content aliases from your #4126, apply the latest Black formatting, and reject invalid sbitLineMetrics direction values.

Issue #4118, opened 2026-07-02
[ttLib] `OS/2` table should compat with the `bhed` table


TakWolf:
The commit https://github.com/fonttools/fonttools/commit/41af9f64813b77ac426ae5cbb5fd47dbbd630803 has already added support for Apple's embedded bitmap tables `bloc`, `bdat` and `bhed`. 

According to the FontForge documentation https://fontforge.org/docs/techref/bitmaponlysfnt.html#apple

> The ‘[head](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6head.html)’ table is replaced by a ‘[bhed](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6bhed.html)’ table which is byte for byte identical

This means that in some cases, only `bhed` will exist and `head` will not. In such cases, if an `OS/2` table is also present, compatibility with this scenario needs to be handled.

```python
    def compile(self, ttFont):
        self.updateFirstAndLastCharIndex(ttFont)
        panose = self.panose
        head = ttFont["head"]    # <--- Here should be made compatible.
        if (self.fsSelection & 1) and not (head.macStyle & 1 << 1):
            log.warning(
                "fsSelection bit 0 (italic) and "
                "head table macStyle bit 1 (italic) should match"
            )
        if (self.fsSelection & 1 << 5) and not (head.macStyle & 1):
            log.warning(
                "fsSelection bit 5 (bold) and "
                "head table macStyle bit 0 (bold) should match"
            )
        if (self.fsSelection & 1 << 6) and (self.fsSelection & 1 + (1 << 5)):
            log.warning(
                "fsSelection bit 6 (regular) is set, "
                "bits 0 (italic) and 5 (bold) must be clear"
            )
        if self.version < 4 and self.fsSelection & 0b1110000000:
            log.warning(
                "fsSelection bits 7, 8 and 9 are only defined in "
                "OS/2 table version 4 and up: version %s",
                self.version,
            )
        self.panose = sstruct.pack(panoseFormat, self.panose)
        if self.version == 0:
            data = sstruct.pack(OS2_format_0, self)
        elif self.version == 1:
            data = sstruct.pack(OS2_format_1, self)
        elif self.version in (2, 3, 4):
            data = sstruct.pack(OS2_format_2, self)
        elif self.version == 5:
            d = self.__dict__.copy()
            d["usLowerOpticalPointSize"] = round(self.usLowerOpticalPointSize * 20)
[...]

PR #4119, opened 2026-07-02
[ttLib] Make the OS/2 table compatible with the bhed table.

TakWolf:
Fix issue https://github.com/fonttools/fonttools/issues/4118


Lib/fontTools/ttLib/tables/O_S_2f_2.py
```diff
@@ -268,6 +259,22 @@ def updateFirstAndLastCharIndex(self, ttFont):
             self.usFirstCharIndex = min(0xFFFF, minCode)
             self.usLastCharIndex = min(0xFFFF, maxCode)
 
+    def checkHeadMacStyle(self, tag, ttFont):
```
behdad: Shouldn't this method be prefixed with `_`?

TakWolf: Fixed.

PR #4121, opened 2026-07-02
[test-data] Set glyphDataFormat=0 in VARC test fonts

behdad:
These variable-composite fonts store their outlines in an ordinary glyf table, so glyphDataFormat should be 0, not 1 (glyphDataFormat=1 is only for cubic glyf outlines). Head-only change; VARC/glyf/gvar bytes are untouched.


khaledhosny:
> (glyphDataFormat=1 is only for cubic glyf outlines)

Is this still the case with the GLYF table?

behdad:
> > (glyphDataFormat=1 is only for cubic glyf outlines)
> 
> Is this still the case with the GLYF table?

Err, no. Cleaning that up is in the beyond-64k branch hopefully, which I plan to get back to in a week or so.

PR #4123, opened 2026-07-06
[misc.roundTools] Add types and documentation

Hoolean:
This PR adds types and documentation for the rounding utility functions. The main goal is to have well-typed returns from `otRound` and `noRound` to keep types alive during OpenType-y arithmetic, but I've added types and documentation for the extra utility functions too for completeness sake, and a collating page for Sphinx.

For ease-of-review, I have tried to avoid all semantic changes; the one exception is the declaration of a type alias at runtime, which should be safe.

### DIFF
--- a/Doc/source/misc/index.rst
+++ b/Doc/source/misc/index.rst
@@ -32,6 +32,7 @@ be more generally useful.
    psCharStrings
    psLib
    psOperators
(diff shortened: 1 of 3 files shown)


Hoolean:
Thanks Behdad!

PR #4124, opened 2026-07-11
[misc.filesystem] contain OSFS paths within the filesystem root

insaf021:
OSFS._abs resolved root/rel_path without verifying the result stayed inside root, so an untrusted UFO glyphs/contents.plist entry or .ufoz member of the form ../../x escaped the filesystem root (arbitrary read for a directory UFO, and file creation via the ZipFS temp mirror for a zip UFO); reject any path that resolves outside root, as pyfilesystem2's own OSFS does.

### DIFF
--- a/Lib/fontTools/misc/filesystem/_errors.py
+++ b/Lib/fontTools/misc/filesystem/_errors.py
@@ -55,7 +55,7 @@ class ResourceReadOnly(ResourceError):
 
 
 class IllegalBackReference(ValueError):
+    # Named after fs.errors.IllegalBackReference (which is also a ValueError and
+    # not an FSError) so that code catching the upstream error keeps working.
     def __init__(self, path):
-        super().__init__(
-            f"path {path!r} contains back-references outside of filesystem"
-        )
+        super().__init__(f"path {path!r} resolves outside of the filesystem root")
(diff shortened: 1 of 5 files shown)


insaf021:
any update?

anthrotype:
Thanks for this, and sorry it sat for a while. I could reproduce an arbitrary file read through the public API (`UFOReader(ufo).getGlyphSet().getGLIF("A")` on a UFO whose contents.plist maps a glyph to `../../secret.txt`). Also the zip variant where building `ReadZipFS`'s directory mirror creates files outside the temp dir.

Rather than another review roundtrip I pushed two follow-up commits directly to this branch, hope you don't mind.

The first fixes a bug: `_directory` cached the `TempFS` before populating it, so when the population loop raised on some member (e.g. the new check firing on an escaping name) the half-built mirror stayed cached, and the next call silently saw an incomplete archive instead of failing again. It could bite any caller that catches the ValueError and carries on, or any code path that touches the same ReadZipFS again. It now publishes the mirror only once population completes, and rejects escaping member names up front so the error names the offending zip entry.

The second commit simplifies the containment check to `is_relative_to` and rewords the error message, which claimed back-references for paths that contain none (e.g. a symlink pointing out of the root). Because `OSFS._abs()` was already resolving symlinks (via `.resolve()`), your check ends up being stricter than pyfilesystem2, which only normalizes textually and will happily follow a symlink out of the root. I think stricter is right for untrusted input. A symlink in a hostile UFO is the same attack, so I kept that and pinned it with a test.

I'd like this in the 4.64.0 release I'm about to cut so I'll merge. Feel free to comment or follow up with another PR.

PR #4125, opened 2026-07-11
Replace private `glob.glob1()` with `glob.glob(..., root_dir=...)`

mgorny:
Replace the use of undocumented `glob.glob1()` function that was removed in Python 3.15 with its equivalent `glob.glob(..., root_dir=...)`.  This fixes compatibility with Python 3.15 (where the function is removed), and works all the way down to Python 3.10.

See: https://docs.python.org/3.13/whatsnew/3.13.html#new-deprecations


justvanrossum (approved):

mgorny:
Thanks!

PR #4128, opened 2026-07-14
sanitise glyph name in EBDT/CBDT extfile export

insaf021:
`ttx -z extfile` on a font whose 'post' table names a glyph `../../../x`:

    open(".../out/bitmaps/strike0/../../../x.png", "wb")

_writeExtFileImageData (shared by EBDT and CBDT) builds the export filename by
concatenating the glyph name onto the strike folder, so a crafted name walks out
of the bitmaps directory and the image bytes get written wherever the path
resolves. Glyph names are arbitrary Pascal strings read from the font, so
dumping an untrusted bitmap font with this option is an arbitrary file write.

The fix keeps only the final path component of the name before joining. Real
glyph names carry no separators, so their output is unchanged.

### DIFF
--- a/Lib/fontTools/ttLib/tables/E_B_D_T_.py
+++ b/Lib/fontTools/ttLib/tables/E_B_D_T_.py
@@ -18,6 +18,13 @@ from .BitmapGlyphMetrics import (
 from . import DefaultTable
 import itertools
 import os
+import re
+from urllib.parse import quote
+
+# a glyph name is usable as a filename as-is unless it contains a path
+# separator (on any platform, not just this one) or the '%' that marks an
+# encoded name
+_UNSAFE_IN_FILENAME = re.compile(r"[%/\\]")
 import struct
 import logging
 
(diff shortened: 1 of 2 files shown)


insaf021:
any update?

anthrotype:
good catch again, thanks. Glyph names are arbitrary strings from the post table, so running `ttx -z extfile` on an untrusted font could write outside the output directory.

I pushed one follow-up commit to your branch. `os.path.basename` blocks the traversal but maps distinct names onto one file: `a/b` and `b` both write `strike0/b.bin`, the second silently overwriting the first, and both `extfileimagedata` entries then point at the same image. A name that previously crashed with `FileNotFoundError` woudl silently corrupt the export instead. We can escape such names like URLs (using `urllib.parse.quote`) instead, with a leading % so an escaped name can never collide with a plain one. Normal glyph names contain no `/` or `%` and are kept as-is, so existing exports keep their filenames.

PR #4131, opened 2026-07-14
[instancer] Don't crash on CFF2 fonts without a VariationStore

### LINKED ISSUE #4130: [instancer] instantiateCFF2 crashes with AttributeError on CFF2 fonts without a VariationStore
`instantiateCFF2` intends to early-out when the CFF2 table has no VariationStore (`Lib/fontTools/varLib/instancer/__init__.py:676-684`):

```python
    varStore = topDict.VarStore.otVarStore
    if not varStore:
        ...
        return
```

but the guard never runs: for a CFF2 table without a VarStore, `topDict.VarStore` itself raises `AttributeError` (cffLib top dicts only grow the attribute when the operator is present), so instancing *any* font that pairs a static CFF2 table with fvar crashes:

```
$ fonttools varLib.instancer --output=out.otf StaticCFF2WithFvar.otf wght=300:600
...
  File ".../fontTools/varLib/instancer/__init__.py", line 678, in instantiateCFF2
    varStore = topDict.VarStore.otVarStore
  File ".../fontTools/cffLib/__init__.py", line 2507, in __getattr__
    raise AttributeError(name)
AttributeError: VarStore
```

Such fonts are easy to produce (e.g. `fontBuilder.setupCFF2` with non-variable charstrings plus `setupFvar`; HVAR/GSUB variations don't require CFF2 blends). `varStore = getattr(topDict, "VarStore", None) and topDict.VarStore.otVarStore` (or a `hasattr` guard) fixes it.

Found while building test fonts for harfbuzz/harfbuzz#6068; sibling of #4129 but a distinct code path.

behdad:
`instantiateCFF2` accessed `topDict.VarStore` before its no-VarStore early-out could run, raising `AttributeError` for any font pairing a static CFF2 table with fvar (variations coming from HVAR etc.).

Test: fully instancing legitimately deletes the VarStore from the table; instancing that result again used to reproduce the crash.

Fixes #4130

### DIFF
--- a/Lib/fontTools/varLib/instancer/__init__.py
+++ b/Lib/fontTools/varLib/instancer/__init__.py
@@ -675,7 +675,9 @@ def instantiateCFF2(
 
     cff = varfont["CFF2"].cff
     topDict = cff.topDictIndex[0]
-    varStore = topDict.VarStore.otVarStore
+    varStore = getattr(topDict, "VarStore", None)
+    if varStore is not None:
+        varStore = varStore.otVarStore
     if not varStore:
         if downgrade:
             from fontTools.cffLib.CFF2ToCFF import convertCFF2ToCFF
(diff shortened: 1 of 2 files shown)


behdad:
@copilot resolve the merge conflicts in this pull request

PR #4133, opened 2026-07-16
Avoid the undocumented `glob.glob1` API, removed in Python 3.15

musicinmybrain:
See https://docs.python.org/3.15/whatsnew/3.15.html#glob.

See also https://docs.python.org/3.15/library/glob.html#glob.glob, which notes that the `root_dir` keyword was added in Python 3.10. Since that’s the minimum version supported by `fonttools`, we can rely on it being available.


justvanrossum:
Duplicate of #4125, which I've just merged.

musicinmybrain:
> Duplicate of #4125, which I've just merged.

Great! Thanks, @mgorny.

PR #4136, opened 2026-07-26
[cmap] Don't drop subtables in unsupported formats

LuShadowX:
`cmap_format_unknown` keeps the raw subtable data so that a subtable in a format we can't read still gets written back out as-is. That doesn't work for a font read from binary, because the subtable has no `cmap` attribute: reading it goes through `__getattr__` -> `ensureDecompiled()`, which clears `self.data` and then raises `AttributeError: cmap`. `table__c_m_a_p.compile()` reads `table.cmap`, so once the table has been loaded the font can't be saved, and the data is gone afterwards.

```pycon
>>> font = TTFont("Tests/ttLib/tables/data/aots/cmap10_font1.otf")
>>> font["cmap"]
>>> font.save("out.otf")
AttributeError: cmap

>>> TTFont("Tests/ttLib/tables/data/aots/cmap10_font1.otf", lazy=False).saveXML("out.ttx")
TypeError: object of type 'NoneType' has no len()
```

There is nothing to decompile for these subtables, so give them an empty `cmap` (as `fromXML()` already does) and leave `self.data` alone.


justvanrossum (approved):
LGTM.

LuShadowX:
Thanks for the review!

anthrotype:
Good catch on the lost data, thanks.

However it introduced a RecursionError because the base `__getattr__` is a retry loop that only terminates because `ensureDecompiled()` sets `self.data = None`, so the retry takes the first branch:

```python
def getattr(self, attr):
    if self.data is None:
        raise AttributeError(attr)
    self.ensureDecompiled()
    return getattr(self, attr)   # retry
```

Your override kept the data, so the retry could loop forever. I fixed by also overriding `__getattr__` to raise AttributeError directly (there is nothing to decompile here, so the lazy path isn't needed)

LuShadowX:
Thanks for merging, and for catching the recursion — I had not spotted that the base `__getattr__` relies on `ensureDecompiled()` clearing `self.data` to terminate. Overriding it to raise directly is the right call here.

PR #4137, opened 2026-07-27
[varLib.instancer] Instance the BASE table

Endika:
`varLib.instancer` leaves the BASE table untouched. A font whose BASE carries variations keeps its Format 3 baseline coordinates, their VariationIndex device tables and the whole ItemVariationStore after instancing, so the instance still uses the default master's baselines, and after a full pin it also ends up carrying variation data with no `fvar` behind it.

Minimal reproduction: take a variable font whose `latn`/`romn` baseline is -120 at the default and -100 at `wght=1.0`, then pin `wght=700`. Before this change the BaseCoord is still `Format 3, Coordinate -120` with its device table and VarStore in place; after it, it is `Format 1, Coordinate -100` and the table has been downgraded to version 1.0.

BASE is now instanced next to GDEF and GPOS. Each BaseCoord goes through `MutatorMerger`, which BaseCoord can share with CaretValue because both have the same Format 3 shape, and then the device varidxes are remapped onto the optimised store, or the store is dropped and the table downgraded when no regions are left.

The `TODO(anthrotype)` in `instantiateOTL` mentioned JSTF and BASE; I left it in place for JSTF, which this does not touch.

One note on the tests: neither feaLib nor varLib can produce a variable BASE table today, so the fixture builds one directly with `OnlineVarStoreBuilder`. The cases cover partial instancing along each axis and full pins at the four corners. Eight of the nine fail before the change. The full test suite passes, and `black` and `mypy` are clean.

### DIFF
--- a/Lib/fontTools/varLib/instancer/__init__.py
+++ b/Lib/fontTools/varLib/instancer/__init__.py
@@ -1318,8 +1318,36 @@ def instantiateItemVariationStore(
     return defaultDeltas
 
 
+def _instantiateBASE(varfont, axisLimits):
+    if "BASE" not in varfont:
+        return
+
+    base = varfont["BASE"].table
+    varStore = getattr(base, "VarStore", None)
+    if varStore is None:
+        return
+
+    log.info("Instantiating BASE table")
+
+    fvarAxes = varfont["fvar"].axes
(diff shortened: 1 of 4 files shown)


behdad:
LGTM. Thanks.

PR #4139, opened 2026-07-28
[misc.xmlWriter] escape the CDATA terminator in writecdata

insaf021:
Dumping a font whose `SVG ` document contains `]]>`, then reading the dump back:

    tables present: ['GlyphOrder', 'SVG ', 'name']
    smuggled name records: [(6, 'INJECTED-PSNAME')]

The input font has no `name` table. `writecdata` pastes the string straight
between `<![CDATA[` and `]]>`, so a document carrying its own terminator closes
the section early and everything after it is read as markup. All three callers
(`SVG `, `bgcl`, `Debg`) hand it data taken verbatim from the font, and
`json.dumps` in the latter two escapes quotes and backslashes but not `]]>`, so
a crafted font decides what its own TTX says and a dump/inspect/recompile pass
rebuilds a font holding tables that were never visible in the dump.

Splitting the section around the terminator leaves the text identical on
re-read, so existing dumps only gain the split.

### DIFF
--- a/Lib/fontTools/misc/xmlWriter.py
+++ b/Lib/fontTools/misc/xmlWriter.py
@@ -93,6 +93,9 @@ class XMLWriter(object):
 
     def writecdata(self, string):
         """Writes text in a CDATA section."""
+        # a CDATA section can't hold "]]>", so close it and open a new one around
+        # the ">"; the two sections read back as the original text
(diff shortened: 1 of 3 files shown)


insaf021:
any update?

anthrotype:
looks good, I had missed this one, thanks for the ping

PR #4141, opened 2026-07-30
[colorLib] Validate glyphs referenced by PaintGlyph/PaintColrGlyph exist in glyphMap (#2629)

### LINKED ISSUE #2629: [colorLib] Do sanity checking to ensure listed glyphs are actually present?
The following line raises a non-descript `KeyError` when the glyph in question doesn't actually exist in `glyphMap`:

https://github.com/fonttools/fonttools/blob/42e4d66184be1acda05bed1896da6d61c084835e/Lib/fontTools/colorLib/builder.py#L609

There should probably be some sanity checking somewhere or early erroring out so fontmake users get a legible error message.

srijavuppala:
Fixes #2629

## Problem

`buildColrV1` already validates that *base* glyphs exist in the `glyphMap` (via `_check_base_glyphs_exist`), but glyphs referenced *inside* paints — `PaintGlyph` (format 10) and `PaintColrGlyph` (format 11) — were not checked. A reference to a glyph that isn't in the `glyphMap` therefore surfaced as an obscure failure later on rather than a clear message. As noted in #2629, this shows up e.g. as a `struct.error: ('required argument is not an integer', 'str', 'Paint', ...)` at compile time, which is hard to act on.

## Change

When a `glyphMap` is provided, `LayerListBuilder` now registers `AFTER_BUILD` callbacks for `PaintGlyph` and `PaintColrGlyph` that raise a legible `ColorLibError` naming the offending paint and glyph, e.g.:

```
PaintGlyph: glyph not found in glyphMap: 'B'
```

This reuses the existing callback mechanism already used elsewhere in the builder, and `buildColrV1`'s per-base-glyph error collection wraps it into the usual "Failed to build BaseGlyphList" report.

The check is a no-op when no `glyphMap` is passed, matching the existing behaviour of the base-glyph check, so this is backward compatible.

## Tests

- `test_buildCOLR_missing_paint_glyph_in_glyphMap_raises_clear_error`
- `test_buildCOLR_missing_paint_colr_glyph_in_glyphMap_raises_clear_error`

Both assert a clear `ColorLibError` is raised. The full `Tests/colorLib/` suite passes and `black --check` is clean. A `NEWS.rst` entry is included.

### DIFF
--- a/Lib/fontTools/colorLib/builder.py
+++ b/Lib/fontTools/colorLib/builder.py
@@ -525,7 +525,8 @@ class LayerListBuilder:
     cache: LayerReuseCache
     allowLayerReuse: bool
 
-    def __init__(self, *, allowLayerReuse=True):
+    def __init__(self, *, glyphMap=None, allowLayerReuse=True):
+        self.glyphMap = glyphMap
         self.layers = []
         if allowLayerReuse:
             self.cache = LayerReuseCache()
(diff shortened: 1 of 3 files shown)


anthrotype:
thanks, looks good. I figured we'd add the same treatment for COLRv0 ;)

PR #4143, opened 2026-08-03
[glyf] Don't crash recompiling OTB fonts with single-entry loca

### LINKED ISSUE #4120: [ttLib] Improve support for bitmap-only sfnt / X11 OTB fonts with zero-length glyf table
### Description

fontTools currently has trouble handling X11 OTB / bitmap-only sfnt fonts that use a true zero-length `glyf` table.

The OTB format is not well documented in official OpenType specifications, but FontForge documents and implements it as a separate bitmap-only sfnt variant used by X11. In this format, the outline tables do not have the same meaning as in ordinary TrueType fonts.

FontForge’s bitmap-only sfnt documentation describes X11 “OpenType Bitmap” fonts (`.otb`) as follows:

- bitmap data is stored in `EBLC` and `EBDT`
- `glyf` exists but has zero length
- `loca` exists and has one entry
- `head` is used, not `bhed`
- `maxp.numGlyphs` is set to the number of bitmap glyphs, not to the size of the `loca` table

Reference:

- https://github.com/fonttools/fonttools/issues/684
- https://fontforge.org/docs/techref/bitmaponlysfnt.html
- https://www.irif.fr/~jch//software/xfree86-bitmap-fonts.html

In this format, `glyf` is not an outline glyph container. It is only a required empty shell, while the actual glyph bitmap data lives in `EBLC` / `EBDT`.

There are two related problems in fontTools:

1. Existing OTB-style fonts with zero-length `glyf` cannot be saved normally once the `glyf` table is loaded.
2. fontTools cannot create or preserve a true zero-length `glyf` table through the normal `glyf` compiler, because empty `glyf` data is changed to `b"\0"` for OTS / Windows compatibility.

### FontForge implementation evidence

FontForge’s source code has a specific special case for X11 OTB / bitmap-only sfnt fonts.

When reading `maxp`, FontForge checks for this condition:

```c
if ( info->glyph_cnt==0 && info->glyph_length==0 && info->loca_length<=4 &&
        info->cff_length==0 && info->typ1_start==0 ) {
    /* X11 OpenType bitmap format */;
    info->onlystrikes = true;
} else if ( cnt!=info->glyph_cnt && info->loca_length!=0 ) {
    ff_post_notice(_("Bad Glyph Count"), _("Font file has bad glyph count field. maxp says: %d sizeof(loca)=>%d"), cnt, info->glyph_cnt);
    info->bad_glyph_data = true;
    if ( cnt>info->glyph_cnt )
        cnt = info->glyph_cnt;
}
```

This is from FontForge’s `fontforge/parsettf.c`.

The important part is that the X11 OTB special case requires:

```text
glyph count derived from loca: 0
glyf length: 0
loca length: <= 4
no CFF / Type1 outline data
```

So a true zero-length `glyf` table is important for FontForge’s OTB recognition path. If `glyf` is rewritten as a one-byte table, this special case no longer applies, and the font can be treated as an ordinary outline sfnt with inconsistent `maxp` / `loca` / `glyf` data.

FontForge’s OTB generation path also intentionally emits no outline glyph data:

```c
} else if ( format==ff_none && at->otbbitmaps ) {
    aborted = !dumpcffhmtx(at,sf,true);
    dumpnoglyphs(&at->gi);
}
...
if ( format==ff_none && at->otbbitmaps )
    dummyloca(at);
```

This confirms that, at least for FontForge’s X11 OTB model, `glyf` is expected to be a zero-length shell and `loca` is expected to be a dummy one-entry table.

### libxfont behavior

I also checked the X11 libxfont source code:

https://gitlab.freedesktop.org/xorg/lib/libxfont

In the FreeType backend, libxfont detects bitmap fonts as follows:

```c
face->bitmap = ((face->face->face_flags & FT_FACE_FLAG_SCALABLE) == 0);
if(!face->bitmap) {
    TT_MaxProfile *maxp;
    maxp = FT_Get_Sfnt_Table(face->face, ft_sfnt_maxp);
    if(maxp && maxp->maxContours == 0)
        face->bitmap = 1;
}
```

So libxfont treats a face as bitmap-only if either:

1. FreeType does not mark the face as scalable, or
2. the face looks scalable/SFNT-like, but `maxp.maxContours == 0`.

I did not find a direct `glyf` table length check in libxfont. However, this still shows that bitmap-only sfnt fonts need to preserve bitmap-only semantics in `maxp`, especially `maxp.maxContours == 0`.

Combined with FontForge’s parser, the desired OTB structure is:

```text
EBLC/EBDT: present
glyf length: 0
loca: one entry, usually [0]
head: present
maxp.numGlyphs: number of bitmap glyphs
maxp.maxContours: 0
```

### Issue 1: OTB-style zero-length glyf cannot be saved normally after loading

For an OTB-style font, `loca` may contain only one entry (`[0]`), while `maxp.numGlyphs`, `cmap`, and the font glyph order still contain the bitmap glyphs.

When such a `glyf` table is decompiled, `table__g_l_y_f.decompile()` iterates over `range(0, len(loca) - 1)`. If `loca` has only one entry, no glyphs are added to `self.glyphs`:

```python
    loca = ttFont["loca"]
    ...
    self.glyphs = {}
    self.glyphOrder = glyphOrder = ttFont.getGlyphOrder()
    ...
    for i in range(0, len(loca) - 1):
        ...
        self.glyphs[glyphName] = glyph
```

Later, when the font is saved and the loaded `glyf` table is compiled again, `table__g_l_y_f.compile()` iterates over `self.glyphOrder` and assumes each glyph name exists in `self.glyphs`:

```python
for glyphName in self.glyphOrder:
    glyph = self.glyphs[glyphName]
```

For OTB-style fonts this fails with `KeyError`, because the glyph order comes from the font / cmap, but the zero-length `glyf` table intentionally has no outline glyph records.

A minimal reproduction is:

```python
from fontTools.ttLib import TTFont, newTable

font = TTFont(sfntVersion="\x00\x01\x00\x00")
font.setGlyphOrder([".notdef", "A"])

font["maxp"] = newTable("maxp")
font["maxp"].numGlyphs = 2

font["loca"] = newTable("loca")
font["loca"].locations = [0]

font["glyf"] = newTable("glyf")
font["glyf"].decompile(b"", font)

print(font.getGlyphOrder())
print(list(font["glyf"].glyphs.keys()))
print(list(font["loca"]))

font.getTableData("glyf")
```

Current output / traceback:

```text
['.notdef', 'A']
[]
[0]
  File ".../fontTools/ttLib/ttFont.py", line 1282, in getTableData
    return self.tables[tag].compile(self)
  File ".../fontTools/ttLib/tables/_g_l_y_f.py", line 143, in compile
    glyph = self.glyphs[glyphName]
KeyError: '.notdef'
```

In practice, saving an existing OTB font directly may fail for this reason.

The workaround I currently have to use is to force the loaded `glyf` table to use an empty local glyph order before saving:

```python
tb_glyf = font["glyf"]
tb_glyf.glyphOrder = []
```

This works around the crash because `glyf.compile()` no longer iterates over `font.getGlyphOrder()`. However, this is only a manual compatibility workaround. It also shows that the normal `glyf` table compiler assumes outline glyph semantics that do not apply to OTB fonts.

### Issue 2: fontTools cannot create or preserve a true zero-length glyf table

Even if the compile-time `KeyError` is avoided, `table__g_l_y_f.compile()` currently changes empty `glyf` data into a one-byte table:

```python
data = b"".join(dataList)
...
if not data:
    # As a special case when all glyph in the font are empty, add a zero byte
    # to the table, so that OTS doesn’t reject it, and to make the table work
    # on Windows as well.
    # See https://github.com/khaledhosny/ots/issues/52
    data = b"\0"
return data
```

This behavior was added for ordinary TrueType fonts with empty outline glyphs, and I understand that it is useful for OTS / Windows compatibility.

However, it conflicts with the X11 OTB requirement that the `glyf` table is truly zero length.

A minimal reproduction is:

```python
from fontTools.ttLib import TTFont, newTable
from fontTools.ttLib.tables._g_l_y_f import Glyph

font = TTFont(sfntVersion="\x00\x01\x00\x00")
font.setGlyphOrder([".notdef", "space"])

font["head"] = newTable("head")
font["head"].indexToLocFormat = 0

font["maxp"] = newTable("maxp")
font["maxp"].numGlyphs = 2

font["loca"] = newTable("loca")

font["glyf"] = newTable("glyf")
font["glyf"].glyphOrder = font.getGlyphOrder()
font["glyf"].glyphs = {name: Glyph() for name in font.getGlyphOrder()}

data = font["glyf"].compile(font)

print(len(data), data)
print(list(font["loca"]))
print(font["maxp"].numGlyphs)
```

Current output:

```text
1 b'\x00'
[0, 0, 0]
2
```

For X11 OTB, the desired structure would be:

```text
glyf length: 0
loca: [0]
maxp.numGlyphs: number of bitmap glyphs
maxp.maxContours: 0
```

The zero-length `glyf` table is not just a cosmetic detail. In FontForge’s parser it is part of the condition that recognizes the X11 OTB special case. Rewriting it as a one-byte table can make the file no longer match the OTB structure and may cause other tools to treat it as an inconsistent ordinary outline sfnt.

### Current workaround

At the moment, I can only make this work by bypassing the normal `glyf` compiler with a custom table class:

```python
from fontTools.ttLib.tables import DefaultTable

class table__g_l_y_f_zero_length(DefaultTable.DefaultTable):
    def __init__(self):
        super().__init__("glyf")

    def compile(self, ttFont: TTFont) -> bytes:
        if "loca" in ttFont:
            ttFont["loca"].set([0])
        return b""
```

For existing OTB fonts, I may also need to force the loaded `glyf` table to use an empty glyph order before saving:

```python
tb_glyf = font["glyf"]
tb_glyf.glyphOrder = []
```

These workarounds suggest that the lower-level SFNT writer can handle zero-length tables correctly. The problem is specific to the normal `glyf` table compiler and its outline-font assumptions.

### Expected behavior

fontTools should provide a supported way to load, save, create, and preserve X11 OTB / bitmap-only sfnt fonts with a zero-length `glyf` table.

In particular:

- Loading and saving an existing OTB-style font should not fail with `KeyError` when `glyf` is loaded.
- A zero-length `glyf` table should be preserved when round-tripping an OTB-style font.
- A one-entry `loca` table should be preserved or supported for this format.
- `maxp.numGlyphs` should be allowed to represent the number of bitmap glyphs, even though `loca` does not contain `numGlyphs + 1` outline offsets.
- `maxp.maxContours == 0` should be preserved for bitmap-only semantics.
- The existing `b"\0"` compatibility behavior for ordinary TrueType outline fonts should not necessarily be removed globally, because it was added for OTS / Windows compatibility.

### Possible approaches

Some possible ways to support this without regressing ordinary TrueType fonts:

1. Add an explicit opt-in flag or attribute on the `glyf` table, allowing it to behave as a zero-length shell.
2. Detect X11 OTB-style fonts, for example fonts with `EBLC` / `EBDT`, zero-length `glyf`, and one-entry `loca`, and preserve this structure.
3. Make `table__g_l_y_f.compile()` tolerate a zero-length-shell mode where `self.glyphs` is empty even though `ttFont.getGlyphOrder()` is not.
4. Provide a dedicated table class or documented mechanism for OTB zero-length `glyf` shells.

### Related history

The one-byte empty `glyf` behavior was added for OTS / Windows compatibility:

- https://github.com/fonttools/fonttools/issues/899
- https://github.com/fonttools/fonttools/pull/1829
- https://github.com/khaledhosny/ots/issues/52

That behavior still seems relevant for ordinary TrueType fonts. The request here is to support the separate X11 OTB / bitmap-only sfnt case, where a true zero-length `glyf` shell is part of the implemented format used by FontForge.

WSHAPER:
Fixes the `KeyError` from `font.save()` after loading an OTB-style font (single-entry loca, zero-length glyf).

Loading and ttx-dumping these fonts already worked since #1680 / #1681; this finishes the round-trip on the compile side, preserving the empty-shell state: when `self.glyphs` is empty, `compile()` skips the `glyphOrder` traversal and emits the single-entry `loca`, leaving `maxp.numGlyphs` (which counts the bitmap glyphs in these fonts) untouched. Nothing is synthesized at decompile time, so a `glyf` table with no outline records stays distinguishable from one whose records are all empty, and a truncated `loca` in an ordinary TrueType font still fails loudly.

`maxp.recalc()` is likewise skipped for the empty shell, as it assumes `numGlyphs` matches the glyphs and there are no outlines to recalculate bounds from.

The `b"\0"` rule from #1829 is left alone. Part 2 of #4120 (true zero-length glyf) is out of scope here.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -139,6 +139,17 @@ class table__g_l_y_f(DefaultTable.DefaultTable):
         dataList = []
         recalcBBoxes = ttFont.recalcBBoxes
         boundsDone = set()
+        if not self.glyphs:
+            # Bitmap-only sfnt fonts (X11 OTB) have an empty 'glyf' table and a
+            # single-entry 'loca': the glyphs are bitmaps in EBLC/EBDT and no
+            # outline glyph records exist. Emit that empty shell instead of
+            # iterating glyphOrder (which would raise KeyError), and leave
+            # maxp.numGlyphs alone since in these fonts it counts the bitmap
+            # glyphs, not the 'loca' entries.
+            # https://github.com/fonttools/fonttools/issues/4120
+            if "loca" in ttFont:
+                ttFont["loca"].set([0])
+            return b"\0"
         for glyphName in self.glyphOrder:
             glyph = self.glyphs[glyphName]
             glyphData = glyph.compile(
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -111,6 +111,14 @@ class table__g_l_y_f(DefaultTable.DefaultTable):
             glyph = Glyph(glyphdata)
             self.glyphs[glyphName] = glyph
             pos = nextPos
+        # OTB bitmap-only sfnt fonts use a single-entry loca ([0]) with an empty
+        # glyf shell; fill placeholder glyphs so compile() doesn't KeyError.
+        # Mirrors the tolerant handling in toXML() and _l_o_c_a.py.
+        # https://github.com/fonttools/fonttools/issues/4120
+        for glyphName in glyphOrder:
+            if glyphName not in self.glyphs:
+                log.warning("glyph '%s' has no entry in glyf table", glyphName)
+                self.glyphs[glyphName] = Glyph()
         if len(data) - nextPos >= 4:
             log.warning(
                 "too much 'glyf' table data: expected %d, received %d bytes",
--- a/Lib/fontTools/ttLib/tables/_m_a_x_p.py
+++ b/Lib/fontTools/ttLib/tables/_m_a_x_p.py
@@ -46,7 +46,16 @@ class table__m_a_x_p(DefaultTable.DefaultTable):
 
     def compile(self, ttFont):
         if "glyf" in ttFont:
-            if ttFont.isLoaded("glyf") and ttFont.recalcBBoxes:
+            if (
+                ttFont.isLoaded("glyf")
+                and ttFont.recalcBBoxes
+                # bitmap-only sfnt fonts (X11 OTB) have an empty glyf table with
+                # no outline glyph records to recalculate from; recalc() also
+                # assumes numGlyphs matches the loca/glyphs, which does not
+                # hold for these fonts as numGlyphs counts the bitmap glyphs
+                # https://github.com/fonttools/fonttools/issues/4120
+                and ttFont["glyf"].glyphs
+            ):
                 self.recalc(ttFont)
         else:
             pass  # CFF
--- a/Tests/ttLib/tables/_g_l_y_f_test.py
+++ b/Tests/ttLib/tables/_g_l_y_f_test.py
@@ -381,6 +381,37 @@ class GlyfTableTest(unittest.TestCase):
         self.assertEqual(font["glyf"][".notdef"].numberOfContours, 0)
         self.assertEqual(font["glyf"]["space"].numberOfContours, 0)
 
+    def test_decompile_otb_style_single_entry_loca(self):
+        # https://github.com/fonttools/fonttools/issues/4120
+        font = TTFont()
+        glyphNames = [".notdef", "space"]
+        font.setGlyphOrder(glyphNames)
+        font["loca"] = newTable("loca")
+        font["loca"].locations = [0]
+        font["glyf"] = newTable("glyf")
+        font["glyf"].decompile(b"", font)
+        self.assertEqual(len(font["glyf"]), len(glyphNames))
+        for name in glyphNames:
+            self.assertEqual(font["glyf"][name].numberOfContours, 0)
+
+    def test_compile_after_decompile_otb_style(self):
+        # https://github.com/fonttools/fonttools/issues/4120
+        font = TTFont(sfntVersion="\x00\x01\x00\x00")
+        glyphNames = [".notdef", "space"]
+        font.setGlyphOrder(glyphNames)
+        font["head"] = newTable("head")
+        font["head"].indexToLocFormat = 0
+        font["maxp"] = newTable("maxp")
+        font["maxp"].numGlyphs = len(glyphNames)
+        font["loca"] = newTable("loca")
+        font["loca"].locations = [0]
+        font["glyf"] = newTable("glyf")
+        font["glyf"].decompile(b"", font)
+        glyfData = font["glyf"].compile(font)
+        self.assertEqual(glyfData, b"\x00")
+        self.assertEqual(list(font["loca"]), [0] * (len(glyphNames) + 1))
+        self.assertEqual(font["maxp"].numGlyphs, len(glyphNames))
+
     def test_getPhantomPoints(self):
         # https://github.com/fonttools/fonttools/issues/2295
         font = TTFont()
--- a/Tests/ttLib/tables/_g_l_y_f_test.py
+++ b/Tests/ttLib/tables/_g_l_y_f_test.py
@@ -442,6 +442,84 @@ class GlyfTableTest(unittest.TestCase):
         self.assertEqual(font["glyf"][".notdef"].numberOfContours, 0)
         self.assertEqual(font["glyf"]["space"].numberOfContours, 0)
 
+    def test_decompile_otb_style_single_entry_loca(self):
+        # https://github.com/fonttools/fonttools/issues/4120
+        # An OTB bitmap-only sfnt has an empty glyf table and a single-entry
+        # loca; decompiling it must not synthesize outline glyph records.
+        font = TTFont()
+        glyphNames = [".notdef", "space"]
+        font.setGlyphOrder(glyphNames)
+        font["loca"] = newTable("loca")
+        font["loca"].locations = [0]
+        font["glyf"] = newTable("glyf")
+        font["glyf"].decompile(b"", font)
+        self.assertEqual(len(font["glyf"].glyphs), 0)
+        self.assertIsNone(font["glyf"].get(".notdef"))
+
+    def test_compile_after_decompile_otb_style(self):
+        # https://github.com/fonttools/fonttools/issues/4120
+        # Recompiling the OTB empty shell must not raise KeyError, must keep
+        # the single-entry loca, and must leave maxp.numGlyphs (which counts
+        # the bitmap glyphs in these fonts) untouched.
+        font = TTFont(sfntVersion="\x00\x01\x00\x00")
+        glyphNames = [".notdef", "space"]
+        font.setGlyphOrder(glyphNames)
+        font["head"] = newTable("head")
+        font["head"].indexToLocFormat = 0
+        font["maxp"] = newTable("maxp")
+        font["maxp"].numGlyphs = len(glyphNames)
+        font["loca"] = newTable("loca")
+        font["loca"].locations = [0]
+        font["glyf"] = newTable("glyf")
+        font["glyf"].decompile(b"", font)
+        glyfData = font["glyf"].compile(font)
+        self.assertEqual(glyfData, b"\x00")
+        self.assertEqual(list(font["loca"]), [0])
+        self.assertEqual(font["maxp"].numGlyphs, len(glyphNames))
+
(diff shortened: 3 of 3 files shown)


TakWolf:
Thanks for working on this. However, overall I don’t agree with the approach taken in this PR.

The crash certainly needs to be fixed, but I don’t think synthesizing empty `Glyph` objects during `glyf.decompile()` is the right model for this case.

1. **The workaround is broader than the OTB case and may hide malformed font data.**

   The added loop fills every glyph missing from `self.glyphs`, regardless of why it is missing. This is not restricted to an OTB-style empty `glyf` shell with a single-entry `loca`. A truncated or otherwise malformed `loca` table in an ordinary TrueType font would receive the same treatment.

   Although warnings are emitted, the inconsistent input is then normalized by synthesizing empty glyphs. This may conceal genuine structural errors and allow later processing to continue with glyph objects that were not represented by the original `loca`/`glyf` data.

2. **It mutates the in-memory representation during decompilation.**

   An OTB-style font initially has a meaningful state such as:

   ```text
   glyphOrder = [".notdef", "A", ...]
   glyf.glyphs = {}
   loca = [0]
   ```

   This means that the font has bitmap glyphs, but the `glyf` table contains no outline glyph records.

   With this PR, loading `glyf` changes that state to:

   ```text
   glyf.glyphs = {
       ".notdef": Glyph(),
       "A": Glyph(),
       ...
   }
   ```

   These `Glyph` objects do not come from the original `glyf` table. They are synthetic placeholders created during decompilation. Their presence may affect subsequent code that uses membership in `glyf`, `len(glyf)`, glyph iteration, XML output, subsetting, or other operations to distinguish between absent outline records and existing empty outline glyphs.

3. **Saving converts the empty-shell representation into a different structure.**

   Each placeholder `Glyph()` compiles to empty data, but it still participates in the normal glyph compilation loop. As a result, `loca` is rebuilt with one offset per placeholder glyph plus the final offset:

   ```text
   [0] -> [0] * (numGlyphs + 1)
   ```

   Since the combined glyph data is empty, the existing OTS/Windows compatibility rule also changes the `glyf` table from zero bytes to `b"\0"`.

   Therefore, after one save, the original OTB-style structure is no longer preserved:

   ```text
   before:
     glyf length = 0
     loca = [0]

   after:
     glyf length = 1
     loca = [0] * (numGlyphs + 1)
   ```

   I understand that preserving a truly zero-length `glyf` is outside the scope of this PR, but synthesizing placeholder glyphs makes this structural conversion an inherent consequence of merely loading and recompiling the table.

4. **The resulting semantics are ambiguous.**

   The accurate OTB semantics are:

   > The font has bitmap glyphs, but `glyf` is an empty shell and contains no outline glyph records.

   The representation introduced by this PR instead means:

   > Every glyph has an outline glyph record, but every such record is empty.

   These states are not equivalent. “No outline glyph records exist” should remain distinguishable from “outline glyph records exist but contain no contours.”

I think the fix should preserve the empty-shell state instead. When `self.glyphs` is empty, `glyf.compile()` can skip the `glyphOrder` traversal and emit the corresponding single-entry `loca`. When `self.glyphs` is non-empty, the current strict lookup against `glyphOrder` should remain unchanged, so genuine inconsistencies in ordinary TrueType fonts are not silently normalized.

This would fix the recompilation `KeyError` without creating synthetic glyphs, without changing the in-memory meaning of the table, and without weakening the existing consistency behavior for non-empty `glyf` tables.

WSHAPER:
Thanks for the thorough review, agreed on all four points. "No outline glyph records exist" and "records exist but are empty" should remain distinguishable states.

I've reworked the PR accordingly: `decompile()` no longer synthesizes anything (the OTB shell loads with `glyphs = {}` and `glyphOrder` intact), and `compile()` now special-cases the empty shell: it skips the `glyphOrder` traversal, sets the single-entry `loca` (`[0]`), and returns without touching `maxp.numGlyphs`, which counts the bitmap glyphs in these fonts. The strict lookup for non-empty `glyf` is unchanged, so a truncated `loca` in an ordinary TrueType font still fails loudly (now also covered by a test).

One addition beyond your sketch: a full `save()` of a real OTB font (with `hmtx`) still crashed in `maxp.recalc()` via the `glyf.__len__` consistency assert, so `maxp.compile()` now skips recalc when the loaded `glyf` has no outline records. Happy to split that into a separate PR if you prefer.

The `glyf` data still compiles to the single zero byte per the existing OTS/Windows rule; a truly zero-length `glyf` (#4120 pt 2) is untouched, as you noted.

PR #4145, opened 2026-08-03
[misc.etree] don't resolve external entities in XMLParser

insaf021:
lxml 4.9.4, a UFO whose `fontinfo.plist` declares `<!ENTITY xxe SYSTEM "file:///etc/passwd">`:

    >>> info = SimpleNamespace()
    >>> UFOReader("demo.ufo").readInfo(info)
    >>> info.note
    '##\n# User Database\n# \n# Note that this file is consulted directly only when th'

`subset/svg.py` already parses OT-SVG with `resolve_entities=False`, but the other four entry points into the etree shim take lxml's default, which resolved external entities until lxml 5.0, and `setup.py` still allows `lxml >= 4.0`: `plistlib.load` (every UFO plist), both GLIF readers in `glifLib`, `SVGPath`, and `designspaceLib`'s reader. Turning the option off in the shim's `XMLParser` is what makes the lxml backend agree with ElementTree, which never resolved them, and it covers plistlib and glifLib where the parser is already built explicitly. `SVGPath` and the designspace reader passed no parser at all, so they now pass one.

Predefined and numeric character references go through libxml2's core rather than entity resolution, so valid documents keep their existing behaviour. Regression tests fail on lxml 4.9.4 before this and pass after; suite is green with and without the lxml extra.

### DIFF
--- a/Lib/fontTools/designspaceLib/__init__.py
+++ b/Lib/fontTools/designspaceLib/__init__.py
@@ -1994,7 +1994,7 @@ class BaseDocReader(LogMixin):
     def __init__(self, documentPath, documentObject):
         self.path = documentPath
         self.documentObject = documentObject
-        tree = ET.parse(self.path)
+        tree = ET.parse(self.path, parser=ET.XMLParser())
         self.root = tree.getroot()
         self.documentObject.formatVersion = self.root.attrib.get("format", "3.0")
         self._axes = []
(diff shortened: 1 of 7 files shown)


insaf021:
gentle ping

anthrotype:
Thanks. I pushed follow-up commit to confine the shim only to lxml versions that actually need it (<5.0).

Also, the tests wrapped the parse in `except Exception: return`, which passes even against unpatched code (any error short-circuits the assertion). I narrowed to catch `ParseError`.

PR #4147, opened 2026-08-10
Raise TTLibError instead of AssertionError on a truncated table entry

eeshsaxena:
### Problem

`SFNTDirectoryEntry.loadData()` (in `sfnt.py`) validates the read with a bare assertion:

```python
def loadData(self, file):
    file.seek(self.offset)
    data = file.read(self.length)
    assert len(data) == self.length
```

A corrupt or truncated table directory entry whose `offset`/`length` runs past the end of the file makes `file.read()` return fewer bytes than `self.length`, so loading a malformed font raises a bare `AssertionError` (with no message) instead of `TTLibError`:

```python
import io
data = bytearray(open('some.ttf','rb').read())
data[5] ^= 0xFF                 # corrupt the table count / directory
f = TTFont(io.BytesIO(bytes(data)), lazy=True)
[f[t] for t in f.keys()]        # AssertionError
```

Two problems: the `AssertionError` isn't `TTLibError` (so callers guarding font loading don't catch it), and under `python -O` the assert is stripped entirely, so the short data is returned silently.

### Fix

Raise a `TTLibError` with a descriptive message when the read comes up short.

### Test

Added `test_loadData_truncated_raises_TTLibError` in `Tests/ttLib/sfnt_test.py`. It fails with a bare `AssertionError` before this change and passes after. `Tests/ttLib/sfnt_test.py` and `ttFont_test.py` pass (33 tests).

### DIFF
--- a/Lib/fontTools/ttLib/sfnt.py
+++ b/Lib/fontTools/ttLib/sfnt.py
@@ -519,7 +519,14 @@ class DirectoryEntry(object):
     def loadData(self, file):
         file.seek(self.offset)
         data = file.read(self.length)
-        assert len(data) == self.length
+        if len(data) != self.length:
+            # A corrupt or truncated table directory entry can point past the
+            # end of the file; raise a TTLibError instead of a bare assertion
+            # (which is also silently skipped under `python -O`).
+            raise TTLibError(
+                "unexpected end of table data: expected %d bytes but got %d "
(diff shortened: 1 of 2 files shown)


anthrotype:
Thanks. I took the liberty to push two small follow-ups: an end-to-end regression test using an actually truncated font to complement your focused unit test, and the offending table tag displayed in the error message to make it more actionable.

PR #4151, opened 2026-08-14
Raise TTLibError for a truncated or out-of-bounds cmap subtable header

eeshsaxena:
Following up on #4147 (truncated table entries), here's another spot where a malformed font surfaces a raw `struct.error` instead of `TTLibError`, this time in the `cmap` loader.

`table__c_m_a_p.decompile` reads the table header, the subtable directory, and each subtable's format/length via `struct.unpack` on slices of `data`. A malformed font can make the table shorter than 4 bytes, promise more subtables than the directory holds, or give a subtable an offset past the end of the data, and any of those hands `struct.unpack` a short buffer, e.g. `struct.error: unpack requires a buffer of 4 bytes`.

I added length checks around the header reads so they raise `TTLibError` instead. This covers the subtable-header parsing; deeply malformed subtable *bodies* can still raise from the individual format decompilers, which is a separate surface.

Added a test in _c_m_a_p_test.py (too-short table, truncated directory, out-of-bounds offset); it raises struct.error on main and passes with the change, and the existing cmap tests still pass. Found it by fuzzing font loading with mutated system fonts.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_c_m_a_p.py
+++ b/Lib/fontTools/ttLib/tables/_c_m_a_p.py
@@ -1,6 +1,6 @@
 from fontTools.misc.textTools import bytesjoin, safeEval, readHex
 from fontTools.misc.encodingTools import getEncoding
-from fontTools.ttLib import getSearchRange
+from fontTools.ttLib import getSearchRange, TTLibError
 from fontTools.unicode import Unicode
 from . import DefaultTable
 import sys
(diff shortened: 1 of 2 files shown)


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`Raise TTLibError for a truncated or out-of-bounds cmap subtable header`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining fonttools!

anthrotype:
Thanks for this, and sorry for the slow reply. The change looks good, and I'd like to widen it slightly before it goes in.

The `offset < 0` check sent me digging, and it turns out to be papering over a much older bug (confirmed by Just minutes ago). `subtableOffset` is `Offset32`, unsigned in both the OpenType and Apple specs, but line 162 has unpacked it signed (`">HHl"`, lowercase `l`) since the original 1999 import (see #4152 for the archaeology). If you're up for it, fixing that here would be neatest: `">HHL"` on the read, plus the matching `struct.pack(">HHl", ...)` at line 235 for symmetry. After that `offset < 0` is dead code and `offset + 4 > len(data)` covers the case on its own. (The compile-side change is mostly cosmetic, just so the read and write agree).

Perhaps that case deserves a test of its own (`subtableOffset = 0xFFFFFFFF`) to pin down the unpack format itself.

Another thing I noticed is the slice handed to `decompileHeader` a few lines further down, `data[offset : offset + int(length)]`, which the new guards don't cover. Two ways it can go wrong:

- the slice is shorter than `length` and `assert len(data) == length` inside `decompileHeader` fires, `AssertionError` disappears under `python -O`, leaving `self.length` bogus (same bug class as #4147).
- `length` is too small for that format's header struct, so the `struct.unpack` at the top of `decompileHeader` gets a short buffer. The length itself depends on the format: `< 6` for formats 0/2/4/6 (`">HHH"`), `< 16` for 12/13 (`">HHLLL"`), `< 10` for 14 (`">HLL"`). `length == 0` is already skipped by the existing zero-length branch, so those narrow bands are what remains exposed.

Both are header parsing rather than subtable bodies, so they're inside what the PR title promises.

Test nits, all minor:

- The three cases share one `assertRaises` in a plain loop, so when it fails only the first case is reported. `subTest` fixes that.
- No `match=` on the messages, so a case could pass by tripping the wrong guard.
- The messages themselves carry no expected/got numbers, unlike the ones on #4147, worth aligning the two.

eeshsaxena:
Thanks for digging into the signed-unpack history, that makes sense. Happy to widen it here: I'll switch the read to >HHL with the matching pack, add the 0xFFFFFFFF case, and guard the short-length slice into decompileHeader, plus split the tests into subTests with match=.

anthrotype:
Thanks! I pushed one commit on top rather than sending you round a third time.
Since the numbers in `_CMAP_SUBTABLE_HEADER_SIZE` already lived in each subtable's `decompileHeader`, I moved the struct format string onto a`headerFormat` class attribute and let decompile `calcsize` it.

`cmap_format_unknown` has to set it to `""` explicitly otherwise it'd inherit ">HHH" from the base class and formats 8 and 10 would incorrectly pick up a 6-byte minimum (added a test for that).

anthrotype:
One more on top. The assert in `cmap_format_12_or_13.decompileHeader` was the last header check that wasn't a TTLibError, and -O drops it, so a subtable whose nGroups disagrees with its length would either came back empty or die of IndexError later. It now fails with a proper error.

PR #4156, opened 2026-08-20
Make the glyf table more dict-like

jenskutilek:
When I do something like this:

```py
for glyph_name in self.font["glyf"].keys():
    print(glyph_name)
```
`ruff` will tell me to _Use `key in dict` instead of `key in dict.keys()`_ ([SIM118](https://docs.astral.sh/ruff/rules/in-dict-keys/)).

Following this advice results in a traceback:

```txt
File "iter_glyphs.py", line 5, in <module>
    for name in f["glyf"]:
                ~^^^^^^^^
  File "Lib/fontTools/ttLib/tables/_g_l_y_f.py", line 344, in __getitem__
    glyph = self.glyphs[glyphName]
            ~~~~~~~~~~~^^^^^^^^^^^
KeyError: 0
```

This PR adds the `__iter__`, `items` and `values` methods to the `glyf` table to make it more dict-like.

### DIFF
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -320,6 +320,15 @@ class table__g_l_y_f(DefaultTable.DefaultTable):
     def keys(self):
         return self.glyphs.keys()
 
+    def __iter__(self):
+        return self.glyphs.__iter__()
+
+    def values(self):
+        return self.glyphs.values()
+
+    def items(self):
+        return self.glyphs.items()
+
     def has_key(self, glyphName):
         return glyphName in self.glyphs
 
--- a/Lib/fontTools/ttLib/tables/_g_l_y_f.py
+++ b/Lib/fontTools/ttLib/tables/_g_l_y_f.py
@@ -1,6 +1,7 @@
 """_g_l_y_f.py -- Converter classes for the 'glyf' table."""
 
 from collections import namedtuple
+from collections.abc import ItemsView, ValuesView
 from fontTools.misc import sstruct
 from fontTools import ttLib
 from fontTools import version
@@ -324,10 +325,10 @@ class table__g_l_y_f(DefaultTable.DefaultTable):
         return self.glyphs.__iter__()
 
     def values(self):
-        return self.glyphs.values()
+        return ValuesView(self)
 
     def items(self):
-        return self.glyphs.items()
+        return ItemsView(self)
 
     def has_key(self, glyphName):
         return glyphName in self.glyphs
--- a/Tests/ttLib/tables/_g_l_y_f_test.py
+++ b/Tests/ttLib/tables/_g_l_y_f_test.py
@@ -23,6 +23,7 @@ from fontTools.ttLib.tables._g_l_y_f import (
 from fontTools.ttLib.tables import ttProgram
 import sys
 import array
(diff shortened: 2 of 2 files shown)


anthrotype:
And who told ruff that glyf is a dict?! technically it's not (we don't subclass dict or Mapping), so ruff is probably pattern-matching on any `for x in obj.keys()`.

That said.. the class already has `keys`, `get`, `__getitem__` etc. so it quacks like a Mapping in all but protocol, so  anyone writing `for name in glyf` hits the same trap. So I'm in favour of completing the protocol.

However, your `values()` and `items()` methods as written bypass lazy expansion and return the raw dict view, whereas `__getitem__` and `get()` both call `glyph.expand(self)`. I think the new methods should also expand for consistency, otherwise, e.g.:

```py
font = TTFont("MyFont.ttf")
for name, glyph in font["glyf"].items():
    print(name, glyph.numberOfContours)
# AttributeError: 'Glyph' object has no attribute 'numberOfContours'
```

You could return the [`collections.abc` view classes ](https://docs.python.org/3/library/collections.abc.html#collections.abc.ValuesView) which route each access through `__getitem__` so glyphs expand lazily as they are consumed:

```py
from collections.abc import ItemsView, ValuesView

    def values(self):
        return ValuesView(self)

    def items(self):
        return ItemsView(self)
```

That's the same pattern `collections.abc.Mapping` itself uses to implement `items()`/`values()`. The views are re-iterable and sized, so your existing isinstance assertions pass unchanged, and iteration stays lazy (no eager decompile of the whole table).

Ideally you also add a test that iterates `items()`/`values()` on a lazily loaded font.

jenskutilek:
> And who told ruff that glyf is a dict?!

Apparently it just matches the method name, `keys`.

anthrotype:
Thanks, the `ValuesView`/`ItemsView` change looks good.

The new lazy tests pass against the old implementation too, though: building the right-hand side of `assert items == [(".notdef", glyf[".notdef"]), ...]` calls `glyf[name]` on every glyph, which expands the same `Glyph` objects already held in `items`, so the later `numberOfContours` check never sees an unexpanded glyph. Just move that assert above the equality check in both tests:

```diff
@@ -977,6 +977,7 @@
         glyf = font["glyf"]
         assert isinstance(glyf.items(), ItemsView)
         items = list(glyf.items())
+        assert items[0][1].numberOfContours == 2  # .notdef
         assert items == [
             (".notdef", glyf[".notdef"]),
             (".null", glyf[".null"]),
@@ -985,7 +986,6 @@
             ("period", glyf["period"]),
             ("ellipsis", glyf["ellipsis"]),
         ]
-        assert items[0][1].numberOfContours == 2  # .notdef
 
     def test_iter(self):
         glyf = newTable("glyf")
@@ -1023,6 +1023,7 @@
         glyf = font["glyf"]
         assert isinstance(glyf.values(), ValuesView)
         glyphs = list(glyf.values())
+        assert glyphs[0].numberOfContours == 2  # .notdef
         assert glyphs == [
             glyf[".notdef"],
             glyf[".null"],
@@ -1031,7 +1032,6 @@
             glyf["period"],
             glyf["ellipsis"],
         ]
-        assert glyphs[0].numberOfContours == 2  # .notdef
```

anthrotype (approved):
LGTM to merge after the test nit above

jenskutilek:
Good catch! Fixed.

PR #4162, opened 2026-08-25
[subset] Fully prune VARC auxiliary data

behdad:
The VARC subsetter already compacted component variation indices, axis-index entries, conditions, MultiVarData, and sparse regions. This completes auxiliary pruning by:

- recursively collecting and remapping Format 2 VarIdx references in condition formats 3 through 5
- pruning conditions before the MultiVarStore, so discarded conditions do not retain variation data
- removing empty MultiVarStore, ConditionList, and AxisIndicesList structures
- adding structural tests for variation records, nested conditions, and empty auxiliary data

VarComponent decoding remains centralized in the existing decoded object model.

Validation:
- 117 relevant tests passed, 6 skipped
- RCJK Hangul and Hanzi subsets reload and draw successfully with HarfBuzz hb-draw-all

### DIFF
--- a/Lib/fontTools/subset/__init__.py
+++ b/Lib/fontTools/subset/__init__.py
@@ -2840,27 +2840,24 @@ def closure_glyphs(self, s):
 def prune_post_subset(self, font, options):
     table = self.table
 
-    store = table.MultiVarStore
-    if store is not None:
-        usedVarIdxes = set()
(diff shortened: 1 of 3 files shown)


behdad:
HarfBuzz counterpart: https://github.com/harfbuzz/harfbuzz/pull/6210

anthrotype (approved):
LGTM

Issue #1824, opened 2020-02-09
[fealib] lookup flag ignored if it's declared before a script statement


typoman:
In the following lookup, the lookup flag will be ignored which is different from FDK. In FDK it's not ignored.
```fea
    lookup rlig_arab_0 {
        lookupflag RightToLeft IgnoreMarks;
        script arab;
        sub arAlef arLamPF.i arLamPF.m arHehPF.f by arAllah;
    } rlig_arab_0;
```
But if I move the flag after the script statement, it will be taken:
```fea
    lookup rlig_arab_0 {
        script arab;
        lookupflag RightToLeft IgnoreMarks;
        sub arAlef arLamPF.i arLamPF.m arHehPF.f by arAllah;
    } rlig_arab_0;
```
This is not a major issue as long as the user knows about this difference. I had a situation that I couldn't find an explanation why my lookup flag is ignored in fontmake but not in makeotf.

khaledhosny:
`script` and `language` are not allowed in standalone lookups per the spec.

khaledhosny:
Assuming this is a nested lookup block (i.e. inside a feature block), the [spec says](https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#4.d):
> The lookupflag attribute stays the same until explicitly changed, until a lookup reference statement is encountered that changes it, until the script is changed, or until the end of the feature.

So we need a complete test, it might be that the script is not changing (the current script is `arab` already), makeotf not actually following the spec.

typoman:
Thank you Khaled, this was a nested lookup as you described. I can send you the UFO file to your email for your testing.

khaledhosny:
> I can send you the UFO file to your email for your testing.

Yes, please.

typoman:
I just send you the file to your email, please let me know if you've received it or not.

khaledhosny:
Sorry for the delay, I got the files, and I checked what makeotf is doing, but got distracted. Will make a PR soon.

khaledhosny:
https://github.com/fonttools/fonttools/pull/1883 should fix the output to match makeotf's, but note that the first rlig lookup will still have its lookup flags reset (in both makeotf and feaLib) because the current script is different from the new one. Since all lookups in your rlig feature use the same script, you should set it outside the lookup blocks instead (also `RightToLeft` has effect only on cursive attachment lookups, it is ignored by layout engines for all other lookups, so you don’t need to set it).

[...]

Issue #613, opened 2016-05-31
feaLib not correctly applying script and language statements


jamesgk:
I don't think feaLib correctly applies script and language statements.

This can be seen in the feaLib test for spec 4.h, which should generate a "liga" feature that is split so as to only apply some substitutions to certain scripts/languages. The expected ttx (which has a single liga feature with all substitutions) is inconsistent with the type of output makeotf generates with similar input.

https://github.com/behdad/fonttools/blob/master/Lib/fontTools/feaLib/testdata/spec4h1.fea
https://github.com/behdad/fonttools/blob/master/Lib/fontTools/feaLib/testdata/spec4h1.ttx

behdad:
@brawer Sascha, this looks rather serious.  Can you please fix it this week?  Thanks.

jamesgk:
For reference, here's what makeotf is doing differently for the linked feature source:

```diff
--- a/Lib/fontTools/feaLib/testdata/spec4h1.ttx
+++ b/Lib/fontTools/feaLib/testdata/spec4h1.ttx
@@ -12,7 +12,7 @@
             <ReqFeatureIndex value="65535"/>
             <!-- FeatureCount=2 -->
             <FeatureIndex index="0" value="0"/>
-            <FeatureIndex index="1" value="1"/>
+            <FeatureIndex index="1" value="5"/>
           </DefaultLangSys>
           <!-- LangSysCount=0 -->
         </Script>
@@ -23,8 +23,8 @@
           <DefaultLangSys>
             <ReqFeatureIndex value="65535"/>
             <!-- FeatureCount=2 -->
-            <FeatureIndex index="0" value="0"/>
-            <FeatureIndex index="1" value="1"/>
+            <FeatureIndex index="0" value="1"/>
+            <FeatureIndex index="1" value="6"/>
           </DefaultLangSys>
           <!-- LangSysCount=0 -->
         </Script>
@@ -35,8 +35,8 @@
           <DefaultLangSys>
             <ReqFeatureIndex value="65535"/>
             <!-- FeatureCount=2 -->
-            <FeatureIndex index="0" value="0"/>
-            <FeatureIndex index="1" value="1"/>
+            <FeatureIndex index="0" value="2"/>
+            <FeatureIndex index="1" value="7"/>
           </DefaultLangSys>
           <!-- LangSysCount=2 -->
           <LangSysRecord index="0">
@@ -44,8 +44,8 @@
             <LangSys>
[...]

PR #4169, opened 2026-09-04
[feaLib] Fix tracking for redundant script statements

### LINKED ISSUE #2522: [feaLib] "language dflt;" gets put into DFLT script
foo.fea:
```afdko
languagesystem latn dflt;

feature liga {
    script latn;
    sub a by x;
} liga;

feature liga {
    script latn;
    language dflt;
    sub b by y;
} liga;
```

```
$ shape foo.ttf 'abc'
[x=0+1073|b=1+1255|c=2+975]
```

What?

```xml
    <ScriptList>
      <!-- ScriptCount=2 -->
      <ScriptRecord index="0">
        <ScriptTag value="DFLT"/>
        <Script>
          <DefaultLangSys>
            <ReqFeatureIndex value="65535"/>
            <!-- FeatureCount=1 -->
            <FeatureIndex index="0" value="0"/>
          </DefaultLangSys>
          <!-- LangSysCount=0 -->
        </Script>
      </ScriptRecord>
      <ScriptRecord index="1">
        <ScriptTag value="latn"/>
        <Script>
          <DefaultLangSys>
            <ReqFeatureIndex value="65535"/>
            <!-- FeatureCount=1 -->
            <FeatureIndex index="0" value="1"/>
          </DefaultLangSys>
          <!-- LangSysCount=0 -->
        </Script>
      </ScriptRecord>
    </ScriptList>
    <FeatureList>
      <!-- FeatureCount=2 -->
      <FeatureRecord index="0">
        <FeatureTag value="liga"/>
        <Feature>
          <!-- LookupCount=1 -->
          <LookupListIndex index="0" value="1"/>
        </Feature>
      </FeatureRecord>
      <FeatureRecord index="1">
        <FeatureTag value="liga"/>
        <Feature>
          <!-- LookupCount=1 -->
          <LookupListIndex index="0" value="0"/>
        </Feature>
      </FeatureRecord>
    </FeatureList>
    <LookupList>
      <!-- LookupCount=2 -->
      <Lookup index="0">
        <LookupType value="1"/>
        <LookupFlag value="0"/>
        <!-- SubTableCount=1 -->
        <SingleSubst index="0">
          <Substitution in="a" out="x"/>
        </SingleSubst>
      </Lookup>
      <Lookup index="1">
        <LookupType value="1"/>
        <LookupFlag value="0"/>
        <!-- SubTableCount=1 -->
        <SingleSubst index="0">
          <Substitution in="b" out="y"/>
        </SingleSubst>
      </Lookup>
    </LookupList>
```

That's right, even though we asked for it to go into `latn/dflt`, it goes into `DFLT/dflt`! I don't know if this is a feaLib builder bug, or a bug in the AFDKO spec, but it certainly doesn't seem right.

(This is significant because ufo2ft feature writers write into `<script>/dflt`.)

cmyr:
This bit of behaviour has been a major source of headaches both here as well as in fea-rs; for a bit of history, see:

- https://github.com/fonttools/fonttools/issues/613
- https://github.com/fonttools/fonttools/issues/1824
- https://github.com/fonttools/fonttools/pull/1883
- https://github.com/fonttools/fonttools/issues/2522 (still open)
- https://github.com/fonttools/fonttools/issues/3748
- https://github.com/googlefonts/fontc/pull/959
- https://github.com/googlefonts/fontc/pull/1595

Digging into some remaining funny diffs has uncovered that both implementations were still not quite correct.

Here the issue was that start_feature() hardcoded self.script_ = "DFLT", and set_script() returned early without assigning self.script_ when the script matched the declared language systems.

Given,

```fea
languagesystem latn dflt;
feature derp {
    script latn;
    language dflt;
    sub a by b;
} derp;
```

`script_` stayed "DFLT", 'sub a by b' ended up in DFLT/dflt, and the font got no latn ScriptRecord at all.

The solution is to track the current script and language explicitly instead. start_feature() now seeds them from the _first declared language system_ (rather than DFLT) and set_script() guards its early return on the _current script and language_ rather than on the set of declared language systems.

The early return exists so that a script statement naming the already-current script does not reset the lookupflag (#1824); that is preserved, and now applies to the first declared language system as well.

This fix is directly based on the logic in makeotf. A separate patch in fea-rs will implement the same behaviour.

- fixes #2522

### DIFF
--- a/Lib/fontTools/feaLib/builder.py
+++ b/Lib/fontTools/feaLib/builder.py
@@ -128,6 +128,7 @@ class Builder(object):
             self.scalar_builder = VariableScalarBuilder.from_ttf(font)
         self.default_language_systems_ = set()
         self.script_ = None
+        self.language_ = None
         self.lookupflag_ = 0
         self.lookupflag_markFilterSet_ = None
         self.use_extension_ = False
@@ -1083,7 +1084,10 @@ class Builder(object):
                 location,
             )
         self.language_systems = self.get_default_language_systems_()
-        self.script_ = "DFLT"
+        # The current script and language start out as the first declared
+        # language system, sorted by (script, language) tag. Kind of odd,
+        # but matches makeotf (and fea-rs matches us)
+        self.script_, self.language_ = min(self.language_systems)
         self.cur_lookup_ = None
         self.cur_feature_name_ = name
         self.lookupflag_ = 0
@@ -1097,6 +1101,8 @@ class Builder(object):
         assert self.cur_feature_name_ is not None
         self.cur_feature_name_ = None
         self.language_systems = None
+        self.script_ = None
+        self.language_ = None
         self.cur_lookup_ = None
         self.lookupflag_ = 0
         self.lookupflag_markFilterSet_ = None
@@ -1169,6 +1175,7 @@ class Builder(object):
             cur_lookups = self.features_.get(key, [])
             self.features_[key] = [x for x in cur_lookups if x not in lookups]
         self.language_systems = frozenset([(self.script_, language)])
+        self.language_ = language
 
         if required:
             key = (self.script_, language)
@@ -1237,8 +1244,13 @@ class Builder(object):
                 "Script statements are not allowed " "within standalone lookup blocks",
                 location,
             )
-        if self.language_systems == {(script, "dflt")}:
-            # Nothing to do.
+        if script == self.script_ and self.language_ == "dflt":
+            # A script statement naming the already-current script still narrows
+            # the language systems that the rules after it are registered under,
+            # and still terminates the current lookup; it differs from the normal
+            # path only in leaving script_ and the lookupflag alone, see
+            # https://github.com/fonttools/fonttools/issues/1824.
+            self.set_language(location, "dflt", include_default=True, required=False)
             return
         self.cur_lookup_ = None
         self.script_ = script
--- a/Tests/feaLib/builder_test.py
+++ b/Tests/feaLib/builder_test.py
@@ -88,6 +88,8 @@ class BuilderTest(unittest.TestCase):
         contextual_inline_format_4
         chain_context_multi_subst_class
         duplicate_language_stmt
+        script_language_tracking script_language_tracking_DFLT
+        script_language_tracking_multi script_language_tracking_redundant
         CursivePosSubtable
         MarkBasePosSubtable
         MarkLigPosSubtable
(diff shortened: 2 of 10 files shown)


Lib/fontTools/feaLib/builder.py
```diff
@@ -1237,8 +1244,8 @@ def set_script(self, location, script):
                 "Script statements are not allowed " "within standalone lookup blocks",
                 location,
             )
-        if self.language_systems == {(script, "dflt")}:
-            # Nothing to do.
+        if script == self.script_ and self.language_ == "dflt":
+            # Nothing to do
             return
```
anthrotype, line 1254: When this statement matches the lexicographically first language system, the early return here leaves `self.language_systems` containing every globally declared system.

The specification says that rules before the first script or language statement apply to all declared language systems, but 

> once the first script or language statement occurs, subsequent lookups and rules are registered only within the currently specified script and language.

https://github.com/adobe-type-tools/afdko/blob/a843a0a87d9db0ea62d5ce719900acf5749c143e/docs/OpenTypeFeatureFileSpecification.md?plain=1#L1154-L1165

E.g. with the following FEA

```fea
languagesystem latn dflt;
languagesystem cyrl dflt;

feature test {
  lookupflag IgnoreMarks;
  script cyrl;
  sub A by B;
} test;
```

... this PR registers the rule under both `cyrl` and `latn` scripts. But since `script cyrl` precedes the feature's first rule, the lookup should be registered only under cyrl/dflt.  I also checked makeotf 5.0.1 and it registers only cyrl/dflt while preserving `IgnoreMarks`.

I think the redundant path can do

```python
if script == self.script_ and self.language_ == "dflt":
    self.set_language(
        location, "dflt", include_default=True, required=False
    )
    return
```

That method already narrows the registration set and terminates the current lookup without resetting lookupflag.

Then we should also add a regression test with the example above.

anthrotype: There is a second regression case worth covering, where a rule precedes the redundant script:

```fea
languagesystem latn dflt;
languagesystem cyrl dflt;

feature test {
  sub A by B;
  script cyrl;
  sub C by D;
} test;
```

At the current PR head, both substitutions are combined into one SingleSub lookup registered under both language systems. Thus latn/dflt incorrectly receives C -> D.

Per the specification, the expected result would be:

- latn/dflt: A -> B
- cyrl/dflt: A -> B and C -> D

Using the same lookup type on both sides of `script cyrl` is intentional: it verifies that the script statement (through the proposed `set_language(...)` that clears `cur_lookup_`) stops subsequent rules from being appended to the lookup that was already registered globally.

anthrotype: Thanks, the first case is fixed with a301d9532ea6f089b3e537e65d30f8cc050790a4 but the rule-before-script case still fails because this path does not clear `cur_lookup_`.

At a301d953, the example from my follow-up still produces one SingleSub lookup containing both A -> B and C -> D, registered under _both_ scripts.

cmyr: yea, looking at that second case now :)

anthrotype: My proposed set_language would handle both cases, no? The same method is already called at the end of set_script

cmyr: this is such a rabbit hole.

there's an example that fails with your case, because of the unconditional clear of cur_lookup:

```fea
languagesystem latn dflt;
feature test {
    sub A by B;
    script latn;
    sub C by D;
} test;
```

here you would create two lookups, but we only want one? I think?

anthrotype: Yes we want one. How about


```python
if script == self.script_ and self.language_ == "dflt":
    if self.language_systems != {(script, "dflt")}:
        self.set_language(
            location, "dflt", include_default=True, required=False
        )
    return
```

cmyr: ... but handling this very dumb edge case introduces meaningful cognitive complexity, so I'm happy to keep splitting it.

anthrotype: Fair enough. The unconditional set_language is simpler, and the extra lookup is harmless. SGTM.

anthrotype (approved):

PR #4171, opened 2026-09-07
[otlLib] Fix overflow handling when building contextual lookups

anthrotype:
Set the standalone lookup writer’s metadata so offset overflows raise OTLOffsetOverflowError instead of AttributeError, allowing another contextual format to be tried. Fixes a regression introduced by #3439.

### DIFF
--- a/Lib/fontTools/otlLib/builder.py
+++ b/Lib/fontTools/otlLib/builder.py
@@ -392,7 +392,11 @@ class ChainContextualBuilder(LookupBuilder):
         # We need to make a copy here because compiling
         # modifies the subtable (finalizing formats etc.)
         table = self.buildLookup_(copy.deepcopy(subtables))
-        w = OTTableWriter()
+        w = OTTableWriter(tableTag=self.table)
+        # This standalone lookup has no LookupList parent to set its metadata.
+        # Overflow reporting still needs a lookup name and index.
+        w.name = "Lookup"
+        w.repeatIndex = 0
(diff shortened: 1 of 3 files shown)


anthrotype:
I pushed a second commit that expands the scope of this PR a bit. Fixing the AttributeError wasn't enough for the font I was testing with: with the overflow now reported correctly, the builder just moves on to "All candidates overflowed", because two of its chaining lookups (7,325 and 4,933 rules) are too big for any single subtable. Format 2 overflows inside the subtable (the ChainPosClassSet offset ends up around 67k) or can't be used at all where the input classes overlap; and format 3 needs one subtable per rule, so the lookup's own Offset16 subtable array overflows (promoting to extension doesn't help since that doesn't shrink the array).

So when every candidate format fails with `OTLOffsetOverflowError`, we now bisect the ruleset into two consecutive halves and retry the format selection on each half, in the same lookup and preserving rule order, the chaining decision and flags. Each half gets its own format, and a single rule that still overflows raises as before. Nothing changes for rulesets that already fit, the splitting is only ever tried on overflow.

I considered doing this in `otTables.fixSubTableOverFlows` next to the other splitters, but it wouldn't work here: the otlLib builder raises before the GSUB/GPOS table is ever packed, the packer only sees one already-chosen format so it can't re-encode a run of format 3 subtables as format 2, and there's no valid packer-level fix for a Lookup with too many subtables anyway.

PR #4174, opened 2026-09-08
Type Hint Expansions

vibbix:
## Background
As part of my work for my [site](https://beta.mtechnic.me), I have been developing a custom font that emits glyphs directly from bitmap sources (converted to `COLRv[0|1]` and `ot-svg`). Each font size captured represents a new glyph, thats subbed in via `gsub`. One of the limits of `fonttools` is that in order to use the features, I need to typically supply a `designspace` document, but because of the cowboy way of handling it, I needed direct access to being able to setup `hvar` `/vvar` / `mvar`, which is currently limited. This bit of work just upstream a bunch of my type annotations work I've added to help along.

[stack mode link](https://github.com/vibbix/fonttools/pull/1)

## Changes
  - type hints for `fontBuilder`
  - `addOpenTypeFeaturesFromString` in `feaLib/builder.py`
  - `timeTools.py`
  - `svgPathPen.py`
  - misc. settings in `subset`
  - for `scaleUpem`
- added missing `GlyphOrder` for `__setitem__` in `ttFont`
- added more type hints to `cmap`, `glyf`, `kern`, and `name`
- for `varLib` added a bunch of type hints to improve checking for `opsz` cases

## Follow ups
```python
    @overload
    def __getitem__(self, tag: Literal["BASE"]) -> B_A_S_E_.table_B_A_S_E_: ...
    @overload
    def __getitem__(self, tag: Literal["CBDT"]) -> C_B_D_T_.table_C_B_D_T_: ...
```
1. I'm thinking of following the pattern set by `ttFont.py` above, and applying it to `newTable(...)` and `__setitem__`
2. Where I added type hints is pretty local to what I've been working on (variable fonts, kern hints, etc.), but I have some other I've run into that I'll add in the future.

# Related
I have some other tickets similar to this for `skia-pathops`, `uharfbuzz`, and `monobit` outstanding as part of this work.

---

## Metrics

| stop | mypy errors | `fontTools` typed | `ttLib/tables` typed |
|---|---|---|---|
| `main` @ 48a0cf93b | 765 | 22.1% | 14.0% |
| **this PR** | 816 (+51) | **23.7%** | **15.6%** |

## Followups
 - [https://github.com/fonttools/fonttools/pull/4175]
 - [https://github.com/fonttools/fonttools/pull/4176]

PR #4175, opened 2026-09-09
sstruct-driven table field declarations

vibbix:
# Background
See #4174 - Unfortunately, without `GH Stack` this is a bit hard to read. But the idea here is to extend what you're doing already with the table generator for `__getitem__`, but do it for all fields that require `sstruct` for OT tables.

[stack mode link](https://github.com/vibbix/fonttools/pull/2)

## Metrics
| stop | mypy errors | `fontTools` typed | `ttLib/tables` typed |
|---|---|---|---|
| `main` @ 48a0cf93b | 765 | 22.1% | 14.0% |
| #4174 | 816 | 23.7% | 15.6% |
| **this PR** | 807 (−9) | **25.4%** | **20.5%** |

## Followups 
 - #4176


justvanrossum:
In today's world with properties and type annotations, sstruct would not have been designed the way it is.

I wonder if we could turn the approach around: instead of using a complex script to insert type annotations, use type annotations to derive the [s]struct formats at runtime?

vibbix:
> In today's world with properties and type annotations, sstruct would not have been designed the way it is.

> I wonder if we could turn the approach around: instead of using a complex script to insert type annotations, use type annotations to derive the [s]struct formats at runtime?

I was looking into something similar for [monobit](https://github.com/robhagemans/monobit/issues/57) here.

The constraint ultimately is "do you want to add a build-time dependency on a library that solves this wonderfully(like [`construct`](https://construct.readthedocs.io/en/latest/) / [`construct-typing`](https://github.com/timrid/construct-typing)), or do you want to roll your own `ctypes` based parser that has no external dependencies?"

I have been toying around with it, but the hardest part with binary struct unpacking it dealing with dynamically-sized C objects. Fonts are a worst case scenario for this, as they are chockfull of nested data structures, indirect references, and quirky versioning.

For this work, I tried to use a similar strategy as was used in the preexisting script (AST injection), and not rock the boat. I also didn't want to burden everyone with a massive refactor of ultra-low-level parsing logic without discussing a sound strategy. 

In the end, the work i submitted was mostly about bringing up to spec the part of my own font pipeline that required typing so as that I can automatically create a type-safe binding in my project, without going overboard / recklessly changing logic without approval. I stuck as close as I can to "just type annotations, and minor cleanup".

anthrotype:
Just's inversion (annotations as the source, format strings derived) is a different discussion and worth keeping in its own issue. That's not this PR though.

Same skip as #4174.

PR #4180, opened 2026-09-12
[bezierTools] Bound curve intersection recursion

### LINKED ISSUE #4166: `curveCurveIntersections()` segfaults with extreme finite coordinates
### Summary

`curveCurveIntersections()` terminates the interpreter when given two quadratic curves containing only finite floats with extreme magnitudes.

### Versions

fontTools 4.63.0, CPython 3.12.3, Ubuntu 24.04 x86_64, glibc 2.39

### Reproducer

```python
from fontTools.misc.bezierTools import curveCurveIntersections

extent = 1e308

curveCurveIntersections(
    ((extent, -extent), (0.0, 0.0), (-extent, extent)),
    ((0.0, 0.0), (1.0, 1.0), (2.0, 0.0)),
)
```

```console
Segmentation fault (core dumped)
```

### ASan result

I built fontTools 4.63.0 from source with Clang 18 using `-fsanitize=address,undefined -fno-omit-frame-pointer`.
ASan reports a native stack overflow and shows the same generated Cython recursion frame repeated until the stack is exhausted:

```text
ERROR: AddressSanitizer: stack-overflow
    #16 _curve_curve_intersections_t  Lib/fontTools/misc/bezierTools.c:25678
    #19 _curve_curve_intersections_t  Lib/fontTools/misc/bezierTools.c:26356
    #22 _curve_curve_intersections_t  Lib/fontTools/misc/bezierTools.c:26356
    #25 _curve_curve_intersections_t  Lib/fontTools/misc/bezierTools.c:26356
    ...
SUMMARY: AddressSanitizer: stack-overflow
```

---

I found this while fuzzing Python C extension modules.
I expected the function to raise a Python exception rather than terminate the process.

oodadoudou:
`curveCurveIntersections()` subdivides until both curve bounds meet a fixed absolute-area tolerance. With finite coordinates at extreme magnitudes, subdivision can overflow to non-finite control points before that tolerance is reached, causing the Cython path to exhaust the native stack.

Bound the subdivision depth and return the midpoint of the current parameter ranges when the requested precision cannot be reached. The same approximation is used immediately if subdivision overflows, avoiding an exponential walk through non-finite bounds. Regular curve inputs still converge through the existing area check.

The regression uses the issue reproducer and verifies that an approximate intersection is returned by the compiled Cython extension.

Fixes #4166.


justvanrossum:
Wouldn’t it be better to return a less precise answer than to raise an exception?

oodadoudou:
Yes, that is better. I changed the guard to return the midpoint of the current parameter ranges, matching the function's existing approximation behavior. I also added the same fallback when subdivision overflows to non-finite control points; otherwise those bounds create a large recursive branch tree before the depth guard is reached. The regression now verifies that the extreme input returns an approximate intersection, including through the rebuilt Cython extension.

PR #4182, opened 2026-09-12
[subset] Preserve input flavor by default

### LINKED ISSUE #3630: pyftsubset: WOFF2 files keep .woff2 extension despite being saved as .otf/.ttf
If you don't provide a `--flavor`, a WOFF2 file will be saved as an OTF/TTF file, but still with the .woff2 file extension.

This causes unintended false alarms:

```
Input font: 373412 bytes: Literata-VariableFont_opsz,wght.woff2
Subset font: 599880 bytes: Literata-VariableFont_opsz,wght.subset.woff2
```

(The font seems ~250K bigger than before subsetting)

oodadoudou:
`pyftsubset` currently clears `TTFont.flavor` when `--flavor` is omitted. Subsetting WOFF or WOFF2 input therefore writes uncompressed sfnt data under a `.woff` or `.woff2` filename.

Use the loaded font's flavor as the default, while continuing to honor an explicit `--flavor`. Cover both WOFF formats with the default output naming path.

Fixes #3630.


anthrotype:
thank you. However, once flavor is inherited, `--flavor` needs a way to opt back out to uncompressed sfnt. Right now the only way is an awkward empty string `--flavor=` with an explicit `--output-file` (otherwise the default name ends with `.subset.`, a trailing dot). 

We should also accept a `--flavor=none` and name the default output `.ttf` or `.otf` from sfntVersion.

anthrotype:
> We should also accept a --flavor=none and name the default output .ttf or .otf from sfntVersion.

Done in https://github.com/fonttools/fonttools/pull/4182/commits/3c24603bd6d2da4d5d575433650c8676ea70a057
I didn't add ttf/otf as flavor names, since that would look like we convert between those (which we don't)

PR #4188, opened 2026-09-16
[gvar] read numbered axes from the axisCount element

anthrotype:
... not from fvar's presence.

Minor follow-up to #4187. In gvar.fromXML, deciding whether the axis names are numbers or tags from `"fvar" not in ttFont` makes the result depend on the order the table elements appear in the TTX: with gvar before fvar, "wght" is fed to safeEval and dies with a ValueError (it round-tripped byte-identical before #4187). This only affects importXML where there's no SFNTReader and fvar isn't in the font until its element has been read. fontTools always writes fvar first so a round trip is fine too, but a hand-edited or third-party TTX can have any order.

This PR sets a flag when the `<axisCount>` element is seen instead, which is only written for the no-fvar form and lives inside the gvar element itself, so it doesn't matter what other tables have been loaded yet.

The first commit adds two tests, one for the numbered-axes form and an importXML round trip of varc-6868.ttf with its gvar moved before fvar; the latter fails on current main and passes with the second commit.


anthrotype:
repro here: https://github.com/fonttools/fonttools/actions/runs/35085723408/job/104759950118?pr=4188#step:5:3716

behdad:
Thanks Cosimo. LGTM.

PR #4190, opened 2026-09-16
[instancer] Fix VARC axis remapping and preserve internal variations

behdad:
Removing an unrelated axis from a VARC font could leave stale axis indices in nested component conditions. Remap those references and reject pinning or restricting VARC-referenced axes consistently, including during avar2 partial instancing.

Preserve component-internal variations by skipping avar2 culling based on font-level reachable ranges: component overrides can reach outside those ranges. Add 26 regression cases covering unrelated-axis pins and range restrictions, nested conditions, hidden component axes, serialization round trips, and unsupported requests. Document the supported scope.

Validation: `PYTHONPATH=Lib python -m pytest Tests/varLib/instancer Tests/ttLib/ttGlyphSet_test.py -q` — 357 passed, 3 overlap tests skipped because `pathops` is unavailable. Black checks passed for the changed Python files. Negated-condition remapping is verified through serialized axis references, independently of the drawing fix in #4191.

Matching HarfBuzz implementation and regression cases: https://github.com/harfbuzz/harfbuzz/pull/6265.


anthrotype (approved):
Code looks good. You may want to add this to the tests https://github.com/fonttools/fonttools/pull/4193

PR #4192, opened 2026-09-16
Update to Unicode 18

behdad:
Update the bundled script, script-extension, block, and bidi-mirroring tables from the final [Unicode 18.0.0 data](https://www.unicode.org/Public/18.0.0/ucd/). This adds Jurchen, Proto-Cuneiform, and Seal, seven new blocks, and five new mirroring pairs, with regression coverage for the new data and script APIs.

Keep the existing `unicodedata2` requirement until its Unicode 18 release is available on PyPI.

Validation:
- Compared script, script-extension, and block lookups for all 1,114,112 code points against the official data; also verified every script alias and mirroring mapping.
- Unicode, subset, and merge tests: 161 passed, 13 skipped because optional dependencies are unavailable.
- Black and `git diff --check` passed.


anthrotype:
thanks. I'm working on updating fonttools/unicodedata2 as well

Issue #3742, opened 2025-01-17
numerical robustness of bezier splitting


cmyr:
When splitting a bezier curve in two, we expect the start point of the first new curve to be identical to the original start point, and we expect the end point of the second new curve to be identical to the end of the original curve.

This is not always the case with `splitCubicAtT`, because floating point. For instance, when splitting the curve `((-103, -231), (-61, -240), (-31.009, -245), (6, -245))` at `0.386637` the second new curve ends at the point `(5.999999999999998, -244.99999999999997)` instead of (6, -245).

This seems easy enough to fix, though, by just manually setting these two points to their expected values before returning.

I think this is probably worth fixing; I'll put together an illustrative patch.

behdad:
It's a pain indeed.

anthrotype:
https://github.com/fonttools/fonttools/pull/3743 only fixed it for the pure-python splitCubicAtT function. We should do something equivalent for the cythonized variant splitCubicAtTC and for the quadratic versions of the split functions as well.

behdad:
> We should do something equivalent for the cythonized variant splitCubicAtTC

That version is highly optimized for speed. I was hoping we can skip it, but I'm not against it.

PR #4194, opened 2026-09-17
[bezierTools] Preserve exact endpoints in splitQuadraticAtT and splitCubicAtTC

ihsandeniz:
Follow-up to #3743, addressing the remaining items from #3742:

> @anthrotype: #3743 only fixed it for the pure-python `splitCubicAtT` function. We should do something equivalent for the cythonized variant `splitCubicAtTC` and for the quadratic versions of the split functions as well.

Same fixup as #3743: the first segment is forced to start at the first control point and the last segment to end at the last one.

### Which functions are actually affected

Rather than guess from the names, I checked every `split*` function in `bezierTools` against 30000 random curves each (random control points in ±300, random `t`), asserting that the first segment starts at the original start point and the last ends at the original end point:

| function | before | after |
|---|---:|---:|
| `splitQuadraticAtT` | 27036 / 30000 deviate | **0** |
| `splitCubicAtTC` | 29260 / 30000 | **0** |
| `splitCubicAtT` | 0 | 0 — fixed by #3743 |
| `splitCubicIntoTwoAtTC` | 0 | 0 — unaffected, left alone |
| `splitLine` | 0 | 0 |

The issue example reproduces on `splitCubicAtTC` exactly as it did on `splitCubicAtT`:

```python
>>> seg = (-103-231j, -61-240j, -31.009-245j, 6-245j)
>>> list(splitCubicAtTC(*seg, 0.386637))[-1][-1]
(5.999999999999993-245j)      # before
(6-245j)                      # after
```

**Two functions are affected that were not in the issue's scope:** `splitQuadratic` and `splitCubic` (the `where`/`isHorizontal` pair) deviate in 10959 and 12787 of the same 30000 trials. I left them untouched, since they were not mentioned and their contract is less obvious to me — they can return a single unsplit segment, and I would rather not change that behaviour unasked. Happy to extend this PR to cover them, or open a separate one, whichever you prefer.

### On `splitCubicAtTC` being a generator

`splitCubicAtTC` yields rather than returns, so the `split[0]` / `split[-1]` approach from #3743 does not transfer directly. Collecting into a list would have been the small change, but that makes the function eager, and #3742 notes this variant is *"highly optimized for speed"*. Instead it holds a single segment back, so iteration stays lazy and memory stays O(1).

That is not free, and since the speed of this function was raised explicitly in the issue, here are the numbers rather than a reassurance. Both variants compiled into the same `.so` (added the old body temporarily under a second name so the comparison is apples-to-apples), median of 9 runs × 20000 calls:

| | before | after | |
|---|---|---|---|
| 1 split | 1.002 µs | 1.222 µs | +22% |
| 8 splits | 1.803 µs | 1.963 µs | +8.9% |

The relative cost is worst for the smallest call, where the extra `next()` is a large share of very little work. If that is too much for this function, the alternative is to fix up only `split[-1]` and leave the first point alone (the start point never deviated in any of the 30000 trials — the error accumulates toward the end), which removes the `next()` and most of the overhead. I did not do that on my own because it makes the two cubic variants behave differently from each other.

### Verification

* `pytest Tests` → 4593 passed, 0 failed, with the Cython extensions built (`COMPILED = True`)
* `pytest Tests/misc/bezierTools_test.py` → 28 passed; doctests in `bezierTools.py` → 15 passed
* Both the pure-Python and the compiled path were checked: the deviation counts above are from the compiled build
* Three regression tests added, following `test_splitCubicAtT_robustness`

Tested on Linux / CPython 3.12 only.

If I have misjudged the tradeoff on the generator, or the endpoint contract is deliberately looser than I assumed, please say so before merging.


anthrotype:
Thanks. I pushed three commits to your branch: folded _splitCubicAtTC into splitCubicAtTC so the fixup is one conditional instead of holding a segment back (as fast as main on a Qu2CuPen run over Times New Roman), and extended the same fix to splitQuadratic and splitCubic. I dropped the start fixups too (first point can't move: t1 is 0.0 for the first segment, so d1 == d == pt1 exactly).

PR #4196, opened 2026-09-18
Require Python 3.11+ and add explicit support for 3.15

### LINKED ISSUE #4183: About minimum Python version
Hello

It has been some time since the minimum supported Python version was last raised, and the current minimum, Python 3.10, is nearing [end of life][pyv].

So I propose raising the minimum version to **at least 3.11** before 2026-10-01.

If we follow [NEP 29][NEP29] as stated in the README (<https://github.com/fonttools/fonttools/blob/e74dcee6b032586c1b15f8481bd341bc9e3b4a0b/README.rst?plain=1#L23>), the minimum version should be changed to 3.12.

Considering that [NEP 29][NEP29] has been superseded by [SPEC 0][SPEC0], following [SPEC 0][SPEC0] would raise the minimum version to 3.13 on 2026-10-02. That seems too high, so I think the README should **replace** [NEP 29][NEP29] with Python's official [Supported Versions][pyv].

[pyv]: https://devguide.python.org/versions/#versions
[NEP29]: https://numpy.org/neps/nep-0029-deprecation_policy.html
[SPEC0]: https://scientific-python.org/specs/spec-0000/

anthrotype:
Fixes #4183


tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 minversion = 3.0
-envlist = lint, py3{10,11,12,13,14}-cov, htmlcov
+envlist = lint, py3{11,12,13,14}-cov, htmlcov
```
jenskutilek: How about also adding 3.15 for tox?

.github/workflows/test.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       fail-fast: false
       matrix:
         # Oldest supported, 'Stable' supported, Newest supported
-        python-version: ["3.10", "3.12", "3.13", "3.14"]
+        python-version: ["3.11", "3.12", "3.13", "3.14"]
```
jenskutilek: How about also adding 3.15 for tox?

anthrotype: actions/setup-python ships 3.15.0-rc.2, we would need `allow-prereleases: true`.. Maybe another time

tox.ini
anthrotype: Done. But it won't be exercised on CI until we add it to the matrix. We can do when stable is out. Have you tried yet?

jenskutilek: I haven't tried with fontTools, but my other packages have been running 3.15 with tox and on GitHub CI for a while. I'm using `astral-sh/setup-uv` to set up the Python env though.

anthrotype: we don't need uv, setup-python can install 3.15rc2, it's just that it's a prerelease and we have to opt-in to those. I'd rather wait for it to be stable (1 October I think)

anthrotype: it turned out easier than I thought, so I enabled 3.15 testing (on Ubuntu only for now, kept mac and windows on 3.14 for the 'latest', we can move them later)

PR #4198, opened 2026-09-18
fix: `<a>` in README

op200:
图片
图片

See the image: There was a leftover underline at the bottom of the RST `<a>` tag, so I switched to compact HTML instead, and the protruding underline disappeared.


khaledhosny:
This is not an HTML file. HTML markup is invalid in reStructuredText files

op200:
> This is not an HTML file. HTML markup is invalid in reStructuredText files

I see. Thank you for your answer

anthrotype:
`.. raw:: html` is actually valid docutils, but GitHub's underline glitch is not a reason to put raw HTML in the project README

